PR #1062, opened 2020-08-14
Integrate TextFeaturizer to automl

### LINKED ISSUE #908: Integrate Text Featurization Component to automl
Part 2 of text support epic #367 is to integrate the text featurization component into the pipeline generation process and automl search, to expose the functionality to users.

### LINKED ISSUE #978: Text featurizer: include the nlp primitives with evalml by default
**Background**
#907 tracked adding the text featurizer component. We merged this two weeks ago, for the 0.11.0 release. This added `featuretools` and `nlp_primitives` as core dependencies.

An issue which became readily apparent is that `nlp_primitives` has `tensorflow` and `tensorflow-hub` as dependencies. The `tensorflow` pip package is a huge download, ~500MB, and installing it caused bugs on Windows (#957). Further, our text featurizer doesn't even use the one primitive (universal sentence encoder) which depends on `tensorflow`, so installing `tensorflow` is totally unnecessary.

Today I merged a fix for #957 which made `featuretools` and `nlp_primitives` optional dependencies, and moved them to `test-requirements.txt`. This is fine in the short-term because the text featurizer component isn't used in automl yet.

**Challenge**
Here's a few things we know: 1) we'd like `featuretools` to become a core dependency for evalml (this is fine and mostly independent from this issue). 2) we don't want `tensorflow` to be a core dependency for evalml

So, the challenge is: what's the mechanism to allow us to use the nlp primitives from the `nlp_primitives` package, without also pulling in all of the requirements for all the primitives?

There's an additional challenge for `featuretools`: what happens when you install all the nlp primitives but the dependencies are missing for some? Ideally `featuretools` would print a warning and skip them, but we don't know what the current behavior is. However that's out of scope for this ticket.

**Options**
I discussed with @rwedge today and we came up with a good idea:
* Update the nlp primitives package so that `pip install nlp_primitives` excludes `tensorflow` and `pip install nlp_primitives[complete]` includes `tensorflow`
* Moving forward, if more primitives are added to `nlp_primitives` and they have dependencies which are large, we include them in `nlp_primitives[complete]` but not `nlp_primitives`
* Update featuretools to have `nlp_primitives[complete]` as a pip requirement
* In evalml, list `nlp_primitives` as a core requirement to support the text featurizer. But not `nlp_primitives[complete]`

There are a few disadvantages with this. One is that if we don't keep an eye on `nlp_primitives` and a large/unneeded dependency is added, we'll be back to square one. In theory this is true for any of our dependencies, but in practice we don't have many and this situation of an additional 500MB dependency hasn't come up before.

An alternative would be for us to copy the specific nlp primitives it needs into the evalml repo, as needed. This is simpler but has the drawback that the evalml text featurizer features may not match the featuretools `nlp_primitives` features, and this could lead to future confusion and bugs. However if we want to add the text featurizer to automl but the underlying issue with `nlp_primitives` hasn't been resolved, this may be our only option until the dust settles.

Another option would be to delete the universal sentence encoder from `nlp_primitives`, or to move it to a separate package.

**Plan**
Once we're ready to add the text featurizer to automl, let's reasses. If we have to copy the nlp primitives into our repo, we can do so. And if we're ready to make the necessary changes to `nlp_primitives` we should do that instead.

eccabay:
Closes #908 and fixes #978 by integrating the `TextFeaturizer` component into `automl` and adding `nlp_primitives` to `core-requirements.txt`. The component lives between the `Imputer` and the `OneHotEncoder` in dynamically generated pipeline graphs. 

Performance tests are not at a point where graphs can be generated, but I will post those here once available.


evalml/automl/automl_search.py
```diff
@@ -438,6 +438,7 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
             max_iterations=self.max_iterations,
             allowed_pipelines=self.allowed_pipelines,
             tuner_class=self.tuner_class,
+            text_columns=text_columns,
```
dsherry, line 441: Let's use [woodwork](https://woodwork.alteryx.com/en/stable/start.html) for this!

```
text_column_vals = X.select('natural_language')
text_columns = text_columns.to_pandas().columns
```

evalml/automl/automl_search.py
```diff
@@ -389,9 +386,14 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
                 X = pd.DataFrame(X)
             X = ww.DataTable(X)
 
+        text_column_vals = X.select('natural_language')
```
eccabay, line 389: This produces a warning if there are no text columns in `X`, open to any suggestions on how to suppress that!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1458,3 +1458,26 @@ def test_input_not_woodwork_logs_warning(mock_fit, mock_score, caplog, X_y_binar
...
+         'col_2': ['do you hear the people sing? Singing the songs of angry men\n\tIt is the music of a people who will NOT be slaves again!',
+                   'I dreamed a dream in days gone by, when hope was high and life worth living',
+                   'Red, the blood of angry men - black, the dark of ages past',
+                   'do you hear the people sing? Singing the songs of angry men\n\tIt is the music of a people who will NOT be slaves again!',
+                   'Red, the blood of angry men - black, the dark of ages past',
+                   'It was red and yellow and green and brown and scarlet and black and ochre and peach and ruby and olive and violet and fawn...']
+         })
+    y = [0, 1, 1, 0, 1, 0]
+    automl = AutoMLSearch(problem_type='binary')
+    automl.search(X, y, data_checks='disabled')
```
freddyaboulton, line 1482: Can you please add a comment explaining why the data checks are disabled? Is it because there isn't enough data for 3-fold cv?

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -30,6 +31,16 @@ def test_iterative_algorithm_allowed_pipelines(logistic_regression_binary_pipeli
     assert algo.allowed_pipelines == allowed_pipelines
 
 
+class MockEstimator(Estimator):
```
freddyaboulton, line 34: We have a fixture for a mock classifier called `dummy_classifier_estimator_class` would that work in this tests?

freddyaboulton (approved):
@eccabay Looks good to me!

evalml/automl/automl_search.py
dsherry: Good catch! I'll file that in woodwork. I don't think there should be a warning in this case; we can discuss on that ticket.

For this PR, you can add:
```python
warnings.filterwarnings('ignore', 'The following selectors were not present in your DataTable')
```

freddyaboulton: I think this would this work?

```python
from woodwork.logical_types import NaturalLanguage

 [col_name for col_name, ltype in X.logical_types.items() if ltype == NaturalLanguage]
```

We should probably file an issue with woodwork though.

dsherry: Filed https://github.com/FeatureLabs/woodwork/issues/322 for this warning

evalml/automl/automl_search.py
```diff
@@ -416,7 +418,7 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
             logger.info("Generating pipelines to search over...")
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
-            self.allowed_pipelines = [make_pipeline(X, y, estimator, self.problem_type) for estimator in allowed_estimators]
+            self.allowed_pipelines = [make_pipeline(X, y, estimator, self.problem_type, text_columns=text_columns) for estimator in allowed_estimators]
```
dsherry, line 421: Looks good. I will make a note for the woodwork integration that we should have `make_pipeline` work with datatables instead of pandas types, so that we don't need to pass `text_columns` in here!

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -22,12 +23,9 @@ def __init__(self, text_columns=None, random_state=0, **kwargs):
             random_state (int, np.random.RandomState): Seed for the random number generator.
 
         """
-        self._ft = import_or_raise("featuretools", error_msg="Package featuretools is not installed. Please install using `pip install featuretools[nlp_primitives].`")
-        self._nlp_primitives = import_or_raise("nlp_primitives", error_msg="Package nlp_primitives is not installed. Please install using `pip install featuretools[nlp_primitives].`")
-        self._trans = [self._nlp_primitives.DiversityScore,
-                       self._nlp_primitives.MeanCharactersPerWord,
-                       self._nlp_primitives.PartOfSpeechCount,
```
dsherry, line 29: We removed `PartOfSpeechCount` because it was slow, right?

evalml/pipelines/utils.py
```diff
@@ -50,6 +52,9 @@ def _get_preprocessing_components(X, y, problem_type, estimator_class):
 
     pp_components.append(Imputer)
 
+    if text_columns:
+        pp_components.append(TextFeaturizer)
```
dsherry, line 56: 👍  sweet! Yep, after the imputer, but before encoding. Great.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1458,3 +1458,26 @@ def test_input_not_woodwork_logs_warning(mock_fit, mock_score, caplog, X_y_binar
...
+    assert automl.rankings['pipeline_name'][1:].str.contains('Text').all()
```
dsherry, line 1483: Heh nice

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -130,3 +141,16 @@ def test_iterative_algorithm_results(ensembling_value, dummy_binary_pipeline_cla
...
+    class MockTextClassificationPipeline(BinaryClassificationPipeline):
+        estimator = MockEstimator
+        component_graph = [TextFeaturizer, MockEstimator]
+
+    algo = IterativeAlgorithm(allowed_pipelines=[MockTextClassificationPipeline], text_columns=['text_col_1', 'text_col_2'])
+    pipeline = algo.next_batch()[0]
+    expected_params = {'text_columns': ['text_col_1', 'text_col_2']}
+    assert pipeline.parameters['Text Featurization Component'] == expected_params
+    assert isinstance(pipeline.component_graph[0], TextFeaturizer)
+    assert pipeline.component_graph[0]._all_text_columns == ['text_col_1', 'text_col_2']
```
dsherry, line 156: Wonderful!

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -130,3 +141,16 @@ def test_iterative_algorithm_results(ensembling_value, dummy_binary_pipeline_cla
...
+def test_iterative_algorithm_instantiates_text():
+    class MockTextClassificationPipeline(BinaryClassificationPipeline):
+        estimator = MockEstimator
+        component_graph = [TextFeaturizer, MockEstimator]
+
+    algo = IterativeAlgorithm(allowed_pipelines=[MockTextClassificationPipeline], text_columns=['text_col_1', 'text_col_2'])
+    pipeline = algo.next_batch()[0]
+    expected_params = {'text_columns': ['text_col_1', 'text_col_2']}
+    assert pipeline.parameters['Text Featurization Component'] == expected_params
+    assert isinstance(pipeline.component_graph[0], TextFeaturizer)
```
dsherry, line 155: Hmm, wouldn't `pipeline.component_graph[0]` be the imputer in this case, and you want `pipeline.component_graph[1]`? What am I missing?

evalml/tests/component_tests/test_text_featurizer.py
```diff
@@ -38,12 +35,9 @@ def test_featurizer_only_text(text_df):
                               'MEAN_CHARACTERS_PER_WORD(col_2)',
                               'POLARITY_SCORE(col_1)',
                               'POLARITY_SCORE(col_2)'])
-    for i in range(15):
-        expected_col_names.add(f'PART_OF_SPEECH_COUNT(col_1)[{i}]')
-        expected_col_names.add(f'PART_OF_SPEECH_COUNT(col_2)[{i}]')
     X_t = tf.transform(X)
     assert set(X_t.columns) == expected_col_names
-    assert len(X_t.columns) == 40
+    assert len(X_t.columns) == 10
```
dsherry, line 40: Wow, so part of speech was creating 30 features on this dataset?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -251,6 +253,27 @@ def test_make_pipeline_numpy_input(problem_type):
...
+
+    multiclass_pipeline = make_pipeline(X, y, LogisticRegressionClassifier, ProblemTypes.MULTICLASS, text_columns=['text'])
+    assert isinstance(multiclass_pipeline, type(MulticlassClassificationPipeline))
+    assert multiclass_pipeline.component_graph == [Imputer, TextFeaturizer, OneHotEncoder, StandardScaler, LogisticRegressionClassifier]
+    assert multiclass_pipeline.custom_hyperparameters is None
+
+    regression_pipeline = make_pipeline(X, y, RandomForestRegressor, ProblemTypes.REGRESSION, text_columns=['text'])
+    assert isinstance(regression_pipeline, type(RegressionPipeline))
+    assert regression_pipeline.component_graph == [Imputer, TextFeaturizer, OneHotEncoder, RandomForestRegressor]
+    assert regression_pipeline.custom_hyperparameters is None
```
dsherry, line 274: Awesome, looks good

dsherry (approved):
@eccabay this is🔥 ! Really great test coverage.

All that's blocking IMO is:
* I left a suggestion for how to suppress the woodwork warning in automl search text feature selection. 
* I left one question about `test_iterative_algorithm_instantiates_text`
* Left comment about adding `feature_types` to breaking changes
* Perf test results

evalml/automl/automl_search.py
gsheni: Yes, that should work. We may fix this due to the issue Dylan [raised](https://github.com/FeatureLabs/woodwork/issues/322).

evalml/automl/automl_search.py
```diff
@@ -438,6 +440,7 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
             max_iterations=self.max_iterations,
             allowed_pipelines=self.allowed_pipelines,
             tuner_class=self.tuner_class,
+            text_columns=text_columns,
```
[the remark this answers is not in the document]
dsherry: @freddyaboulton you're right, we could delegate this to the fit method of the text transformers. However, I think this is better. This PR as it stands will use the woodwork datatable to detect and select text features. If users don't like the default selections, they can override and pass a new datatable into `search`. But if we had the text featurizer do the detection, users wouldn't be able to override as easily. This is one of the big value adds of the datatables for us--it provides users with a clear way to control how automl search treats each feature, and to change the behavior!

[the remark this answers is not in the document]
dsherry: Long-term, we'd like all pipelines, components and utils to be able to accept woodwork datatables (#1288). At that point, we can delete this `text_columns` plumbing in favor of the datatables

evalml/automl/automl_search.py
dsherry: @freddyaboulton oops I had written back but didn't submit my comments until after you commented 😆 

I think we can just suppress the warning, yeah?

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
eccabay: Yep! It took significantly longer than any of the other primitives without adding any noticeable performance.

evalml/tests/automl_tests/test_iterative_algorithm.py
eccabay: The pipeline in this test isn't dynamically generated, it's defined a few lines above as only having the TextFeaturizer and an estimator

evalml/tests/component_tests/test_text_featurizer.py
eccabay: Yep, one feature for each part of speech

evalml/tests/automl_tests/test_iterative_algorithm.py
dsherry: Ah right, thanks!

eccabay:
Performance testing doc is [here](https://alteryx.quip.com/a2BbAaSwWoUx/TextFeaturizer-Integration-into-AutoML)

dsherry:
@eccabay thanks, I just reviewed the perf tests. Looks great!

dsherry:
@eccabay to fix the `check_dependencies_updated_linux` failure, I think you gotta update `evalml/tests/latest_dependency_versions.txt` on your branch--it looks like moving featuretools/nlp_primitives to core added `distributed`, and that's currently on our watch-list in the circleci config for the job. We could also take it off the watch list, but for now let's just add it to the deps versions, yeah?

PR #1134, opened 2020-09-02
Implement stacked ensemble classes

### LINKED ISSUE #1129: Implement stacked ensemble classifier and regressor classes
This issue tracks implementing the stacked ensemble classifier and regressor classes.

Epic design doc here: https://alteryx.quip.com/wcJbAoWJVpAN/Ensembling-Design-Document

angela97lin:
Closes #1129 

@dsherry I did some digging around to see what would be necessary to support passing in our estimators to scikit-learn directly, rather than `estimator._component_obj`. It required at least:
- implementing a `get_params(deep)`
- adding `classes_` and `estimator_type`

These additions were fine, but the underlying implementation for concatenating the results from each of the estimators assumed that the `predict` method for each estimator returned an `np.array` and errors out since we pass in `pd.DataFrame`. As such, I think there would need to be more work done to enable passing in our estimators and decided to opt for just passing in `_component.obj`, just as we've done for our other components.

---

Update: This PR also includes adding a utility method to wrap our pipeline / component classes in a simple scikit-learn estimator instance. Currently, `check_estimator(obj)` (from [documentation on rolling your own estimator](https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator) breaks, so the returned estimators are probably not fully scikit-learn compliant, but they're sufficient in this case. Will put up an issue for this once this is merged :)

I tried making the objects scikit-learn compliant, but it is probably not really plausible without some big API changes on our end. I ran into a lot of pickling issues, since dynamic classes can't be pickled using `pickle` and there are tests in `check_estimator` for this.

It seems like the wrapper works fine if `n_jobs=1`, and weird behavior happens otherwise (https://app.circleci.com/pipelines/github/alteryx/evalml/6037/workflows/ebe774d2-77be-42cb-8716-d73b9f382130/jobs/64121). Hence, for the sake of pushing this PR forward, I'll try to investigate this issue and as a fallback, just use `n_jobs=1` for now.

Potentially relevant issue: https://github.com/scikit-learn-contrib/skope-rules/issues/18
---


freddyaboulton (dismissed):
@angela97lin This looks good! I have some suggestions for improving the api of `__init__` I'd like to discuss (I think these could be a way to make use `EnsembleBase` 😉  )

evalml/pipelines/components/estimators/estimator.py
```diff
@@ -71,4 +71,4 @@ def feature_importance(self):
             raise MethodPropertyNotFoundError("Estimator requires a feature_importance property or a component_obj that implements feature_importances_")
 
     def __eq__(self, other):
-        return self.supported_problem_types == other.supported_problem_types and super().__eq__(other)
+        return super().__eq__(other) and self.supported_problem_types == other.supported_problem_types
```
angela97lin, line 74: Small unrelated-ish change but necessary because if `other` is not an Estimator then `supported_problem_types` will raise an AttributeError.

evalml/pipelines/components/utils.py
```diff
@@ -103,3 +108,54 @@ def handle_component_class(component_class):
...
+            return WrappedSKRegressor(evalml_obj)
+        elif evalml_obj.problem_type == ProblemTypes.BINARY or evalml_obj.problem_type == ProblemTypes.MULTICLASS:
+            return WrappedSKClassifier(evalml_obj)
+    else:
+        # EvalML Estimator
+        if evalml_obj.supported_problem_types == [ProblemTypes.REGRESSION]:
+            return WrappedSKRegressor(evalml_obj)
+        elif evalml_obj.supported_problem_types == [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+            return WrappedSKClassifier(evalml_obj)
+    raise ValueError("Could not wrap EvalML object in scikit-learn wrapper.")
```
dsherry, line 161: Two notes:
1. You can replace `is_pipeline` with `isinstance(evalml_obj, PipelineBase)`
2. This code is great but augh we're still being chased by this issue of estimators not having an explicit problem type!

evalml/tests/component_tests/test_components.py
```diff
@@ -776,11 +796,22 @@ def test_serialization_protocol(mock_cloudpickle_dump, tmpdir):
...
+    except EnsembleMissingPipelinesError:
+        if estimator_class == StackedEnsembleClassifier:
+            estimator = estimator_class(input_pipelines=[logistic_regression_binary_pipeline_class(parameters={})])
+        elif estimator_class == StackedEnsembleRegressor:
+            estimator = estimator_class(input_pipelines=[linear_regression_pipeline_class(parameters={})])
     if estimator._component_obj is None:
         pytest.skip(f"Skipping {estimator_class} because does not have component object.")
-    params = estimator._component_obj.get_params()
+    if estimator_class.model_family == ModelFamily.ENSEMBLE:
+        params = estimator._parameters
```
dsherry, line 812: Why `_parameters` and not `parameters` here?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1150,6 +1150,71 @@ def test_pipeline_not_fitted_error(mock_fit, problem_type, X_y_binary, X_y_multi
...
+    if problem_type == ProblemTypes.BINARY:
+        X, y = X_y_binary
+        base_pipeline_class = BinaryClassificationPipeline
+        input_pipelines = [make_pipeline_from_components([classifier], problem_type) for classifier in stackable_classifiers]
+        parameters = {
+            "Stacked Ensemble Classifier": {
+                "input_pipelines": input_pipelines
+            }
+        }
+        graph = ['Simple Imputer', 'Stacked Ensemble Classifier']
```
dsherry, line 1169: What you have here is great, just a note there may be ways to make this less verbose if you use the problem type to generate some of it:
```python
if problem_type == BINARY:
    X, y = X_y_binary
    base_pipeline_class = BinaryClassificationPipeline
    stacking_component_name = StackedEnsembleClassifier.name
    stackables = stackable_classifiers
elif ...

input_pipelines = [make_pipeline_from_components([obj], problem_type) for obj in stackables]
parameters = {
    stacking_component_name: {...}
}
...
```

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1094,3 +1094,49 @@ def test_pipeline_not_fitted_error(mock_fit, problem_type, X_y_binary, X_y_multi
...
+                "estimators": [RandomForestRegressor()],
+                "final_estimator": RandomForestRegressor(),
+            }
+        }
+    pipeline = StackedPipeline(parameters)
+    pipeline.fit(X, y)
+    assert not np.isnan(pipeline.predict(X)).values.any()
+
+    if problem_type == ProblemTypes.BINARY or problem_type == ProblemTypes.MULTICLASS:
+        assert not np.isnan(pipeline.predict_proba(X)).values.any()
```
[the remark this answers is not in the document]
dsherry: Sounds good. I don't think its necessary for the unit test in the first case to assert that the non-ensemble pipeline does not also get non-perfect score -- that's fine! The ensembling should only give a slight improvement anyways.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1150,6 +1150,71 @@ def test_pipeline_not_fitted_error(mock_fit, problem_type, X_y_binary, X_y_multi
...
+                "input_pipelines": input_pipelines
+            }
+        }
+        graph = ['Simple Imputer', 'Stacked Ensemble Classifier']
+        comparison_pipeline_class = logistic_regression_binary_pipeline_class
+        objective = 'Log Loss Binary'
+    elif problem_type == ProblemTypes.MULTICLASS:
+        X, y = X_y_multi
+        base_pipeline_class = MulticlassClassificationPipeline
+        input_pipelines = [make_pipeline_from_components([classifier], problem_type) for classifier in stackable_classifiers]
```
dsherry, line 1175: So this makes a pipeline with a single component, which is the estimator? Where the list of estimators from `stackable_classifiers`/`stackable_regressors` includes all accessible estimators? Cool :)

dsherry (approved):
@angela97lin great work, this looks real solid!

The only things I had which were blocking were:
* Remove line added to `TextFeaturizer`
* Docstring for WrappedSKClassifier/other in utils
* File bug in epic for the multi-processing error you saw related to `n_jobs`

Everything else was just small stuff. Right on!

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -0,0 +1,85 @@
...
+    def default_parameters(cls):
+        """Returns the default parameters for stacked ensemble classes.
+
+         Returns:
+             dict: default parameters for this component.
+        """
+        return {'final_estimator': None,
+                'cv': None,
+                'n_jobs': 1
+                }
```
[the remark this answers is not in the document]
dsherry: Ok, that makes sense

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Yup! Maybe overkill but it was my way of checking that any of our accessible classifiers/regressors are allowed as inputs to the stacked ensemble component ahaha

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -0,0 +1,85 @@
...
+    """Stacked Ensemble Base Class."""
+    model_family = ModelFamily.ENSEMBLE
+    _stacking_estimator_class = None
+    _default_final_estimator = None
+
+    def __init__(self, input_pipelines=None, final_estimator=None, cv=None, n_jobs=None, random_state=0, **kwargs):
+        """Stacked ensemble base class.
+
+        Arguments:
+            input_pipelines (list(PipelineBase or subclass)): List of PipelineBase objects to use as the base estimators.
```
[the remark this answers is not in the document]
angela97lin: Right now we're passing in objects / instances of PipelineBase subclasses :o

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -0,0 +1,85 @@
...
+            final_estimator (Estimator or subclass): The estimator used to combine the base estimators.
+            cv (int, cross-validation generator or an iterable): Determines the cross-validation splitting strategy used to train final_estimator.
+                For int/None inputs, if the estimator is a classifier and y is either binary or multiclass, StratifiedKFold is used. In all other cases, KFold is used.
+                Possible inputs for cv are:
+                - None: 5-fold cross validation
+                - int: the number of folds in a (Stratified) KFold
+                - An scikit-learn cross-validation generator object
+                - An iterable yielding (train, test) splits
+            n_jobs (int or None): Non-negative integer describing level of parallelism used for pipelines.
+                None and 1 are equivalent. If set to -1, all CPUs are used. For n_jobs below -1, (n_cpus + 1 + n_jobs) are used.
```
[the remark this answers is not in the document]
angela97lin: Haha yup: https://scikit-learn.org/stable/glossary.html#term-n-jobs

evalml/pipelines/components/utils.py
angela97lin: Hmm, doing so will cause a circular dependency (having to import `PipelineBase`) but I do like this better. For now, will put the import statement in the method. (

dsherry: Ah good point on the circular dep. Now that you mention it, since this is supposed to work for pipelines and components, it should go in `evalml/pipelines/utils.py`. Then the circular dep wouldn't happen either :)

dsherry: I don't think you updated the impl but that's ok!

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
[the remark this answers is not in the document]
dsherry: Ah, got it, my bad.

evalml/pipelines/components/utils.py
angela97lin: @dsherry Oh, I believe the circular dependency still happens if we put it in `evalml/pipelines/utils.py` since the method is being used in a component (`StackedEnsembleBase`). I think this is another case where we should think about separating pipelines from components to avoid the problem!

Also, I currently see on main that `from evalml.pipelines.pipeline_base import PipelineBase` is called in `scikit_learn_wrapped_estimator`. Is that different from what you see? :o

dsherry: Ah, interesting, right... this component now knows about pipelines. Ok, got it, thanks.

Oh lol I was looking for it in this PR. Got it.

angela97lin: Yup! Definitely a bit odd. This works for now but could be indicative of some API considerations we'll need to make.

And cool, glad that everything was in place! Odd that Github doesn't mark it as outdated but I guess the change was not exactly to this block 😂

PR #1150, opened 2020-09-08
Updated partial dependence methods to support non-numeric columns

### LINKED ISSUE #1125: Support non-numeric columns in one-way Partial Dependency
### Non-Numeric Columns and one-way Partial Dependency

- As a developer, I want to calculate the one-way partial dependency for each feature in my holdout data. I want to better understand the relation between each of the features.
- In many cases, my dataset contains non-numeric columns, like categoricals, datetimes, and naturallanguage. As a ML scientist, I want AutoML search to handle these data types when calculating partial dependency.

## Example
```python
import evalml
from evalml.model_understanding.graphs import partial_dependence
from featuretools.demo import load_retail
from evalml.model_family import ModelFamily, handle_model_family

X_y = load_retail(nrows=1000, return_single_table=True)
X = X_y.drop(columns=['total'])
y = X_y['total']

automl = AutoMLSearch(problem_type="regression", objective="auto", max_pipelines=5)
X_train, X_holdout, y_train, y_holdout = evalml.preprocessing.split_data(X, y, 
                                                                         test_size=0.2,
                                                                         regression=True)
automl.search(X_train, y_train)
for x in automl.rankings['id']:
    pipeline = automl.get_pipeline(x)
    if pipeline.model_family != ModelFamily.BASELINE:
        # quantity is a numeric field
        # But cannot calculate partial dependency since  some other columns in X are non-numeric
        try:
            data = partial_dependence(pipeline, X_holdout, feature='quantity')
        except Exception as e:
            print(e)
```

## Errors
```
Invalid cat_features[6] = 9 value: index must be < 9.
could not convert string to float: 'C536391'
could not convert string to float: 'C536391'
This XGBRegressor instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator.
```

angela97lin:
Closes #1125 

This PR enables support of non-numeric columns in our partial dependence methods by transforming the input before passing it to the scikit-learn method.


evalml/model_understanding/graphs.py
```diff
@@ -426,24 +426,22 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
+    if isinstance(pipeline, evalml.pipelines.ClassificationPipeline):
+        pipeline._estimator_type = "classifier"
+    elif isinstance(pipeline, evalml.pipelines.RegressionPipeline):
+        pipeline._estimator_type = "regressor"
+    pipeline.feature_importances_ = pipeline.feature_importance
+    avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
+
+    # Delete scikit-learn attributes that were temporarily set
+    del pipeline._estimator_type
+    del pipeline.feature_importances_
```
dsherry, line 444: Very cool!

One note: if `sk_partial_dependence` throws, we'll never reach this code. You could use `finally` though:
```python
try:
    avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
finally:
    del pipeline._estimator_type
    del pipeline.feature_importances_
```

This will allow the exception to still propagate up the stack because we didn't add an `except` clause, but it will guarantee that the code in `finally` runs before the exception continues on!

Not critical, but it would be cool to add this sometime.

dsherry (commented):
@angela97lin I took a quick look, sorry I missed your comments yesterday, looks great, I'm glad pipelines worked here! I left one idea for an improvement.

PR #1194, opened 2020-09-17
Infer Problem type 

### LINKED ISSUE #826: Automl: infer problem type from target data
Hello!

I wrote this tiny helper function that tests if a target variable is multiclass or binary for classification problems. Max thought it might be helpful if we added it to EvalML.

```python
def get_problem_type(input_problem_type, target_data):
    """helper function to determine if classification problem is binary or multiclass dependent on target variable values."""
    if input_problem_type == "classification":
        values = target_data.value_counts()
        if values.size == 2:
            return ProblemTypes.BINARY
        elif values.size > 2:
            return ProblemTypes.MULTICLASS
        else:
            raise ValueError(
                "The target field contains less than two unique values. It cannot be used for modeling."
            )
    return ProblemTypes.REGRESSION
```

bchen1116:
fix #826 

Create a method to infer problem type given the target data. Design doc for this is [here](https://alteryx.quip.com/Fr2IAMv2utid/AutoML-Infer-problem-type-from-target-data)

Updated api docs [here](https://evalml.alteryx.com/en/bc_826_targets/api_reference.html#problem-types)


angela97lin (dismissed):
Looking good, made some suggestions on returning `ProblemTypes` enum instead.

Also, we should revert the line break deletions, as they're updating the docs and making things look funky.

bchen1116:
Docs [here](https://evalml.alteryx.com/en/bc_826_targets/user_guide/automl.html#Using-custom-pipelines)

docs/source/user_guide/automl.ipynb
```diff
@@ -71,6 +71,35 @@
...
+    "\n",
+    "y = pd.Series([0, 1, 1, 0, 1, 1])\n",
+    "detect_problem_type(y)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This function can return the predicted problem type as a ProblemType enum, choosing from ProblemType.BINARY, ProblemType.MULTICLASS, and ProblemType.REGRESSION. If the target data is invalid (for instance when there is only 1 unique label), the function will throw an error instead."
```
dsherry, line 100: I suggest you move this up to the first sentence, before the code example, and make some minor tweaks to the text to simplify: So in total:
> EvalML includes a simple method `detect_problem_type` which returns the machine learning problem type deemed most useful for modeling the specified target.

I like what you said about the return types and the error! I think that's great stuff for the API docs. The user guide is intended as a quick high-level dive into using evalml.

evalml/problem_types/utils.py
```diff
@@ -19,3 +23,31 @@ def handle_problem_types(problem_type):
...
+        >>> problem_type = detect_problem_type(y)
+        >>> assert problem_type == ProblemTypes.BINARY
+    """
+    y = pd.Series(y).dropna()
+    num_classes = y.nunique()
+    if num_classes < 2:
+        raise ValueError("Less than 2 classes detected! Target unusable for modeling")
+    elif num_classes == 2:
+        return ProblemTypes.BINARY
+    else:
```
dsherry, line 49: Nit-pick: delete this `else`, not necessary since you `raise`/`return` in the `if`/`elif`:
```python
if num_classes < 2:
    raise ValueError("Less than 2 classes detected! Target unusable for modeling")
if num_classes == 2:
    return ProblemTypes.BINARY
if y.dtype in numeric_dtypes:
    return ProblemTypes.REGRESSION
return ProblemTypes.MULTICLASS
```

This keeps the control flow flat and easy to follow!

evalml/tests/problem_type_tests/test_problem_types.py
```diff
@@ -31,6 +37,78 @@ def test_handle_incorrect_type():
...
+    y_values = pd.Series([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, np.nan])
+    y_float = pd.Series([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11.0])
+
+    assert detect_problem_type(y_values) == ProblemTypes.REGRESSION
+    assert detect_problem_type(y_float) == ProblemTypes.REGRESSION
+
+
+def test_nan_none_na():
+    y_none = pd.Series([None])
+    y_pdna = pd.Series([pd.NA])
```
dsherry, line 98: Cool that you kept this and added nan as another case!

dsherry (approved):
👍 🚀

angela97lin (approved):
LGTM! :shipit:

Issue #1162, opened 2020-09-11
Util: create pipeline instance from list of component instances


dsherry:
Having a pipeline creation method would make it easy for users to define an API using component instances they've already created, without having to learn about how to extend our pipeline classes.

Example:
```python
def make_pipeline(component_instances, problem_type, pipeline_name=None):
    if not isinstance(Estimator, component_instances[-1]):
        raise ...
    pipeline_parameters = {inst.name: inst.parameters for inst in component_instances}
    class TemplatedPipeline(BinaryClassificationPipeline)
        custom_name = pipeline_name
    return TemplatedPipeline(pipeline_parameters)
```

One challenge: I think we'd need users to provide the `problem_type`, unless there's some smart way to infer that from the components. Our components aren't currently provided with a `problem_type`, and our component inheritance structure doesn't have problem-type-specific classes like we do for pipelines.

(@kmax12 @freddyaboulton filing this RE our recent discussion)

PR #1204, opened 2020-09-21
Refined `make_pipeline_from_components` implementation

christopherbunn:
Continuing from the initial PR (#1176 ), additional work for #1162.

- Added fixes so that `make_pipeline_from_components` shows up in the API documentation
- Updated function implementation so that `component_graph` isn't directly set
- Added additional test to show that two pipelines with the same components will have the same predictions  
- Added example in API reference

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -123,7 +123,6 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
         >>> components = [Imputer(), StandardScaler(), CatBoostClassifier()]
         >>> pipeline = make_pipeline_from_components(components, problem_type="binary")
         >>> pipeline.describe()
-        >>> assert pipeline.component_graph == components
 
     """
     if not isinstance(component_instances[-1], Estimator):
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -120,7 +120,7 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
         Pipeline instance with component instances and specified estimator
 
     Example:
-        >>> components = [Imputer(), StandardScaler(), CatBoostClassifier()]
+        >>> components = [Imputer(), StandardScaler(), LogisticRegressionClassifier()]
         >>> pipeline = make_pipeline_from_components(components, problem_type="binary")
         >>> pipeline.describe()
 
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -119,6 +119,12 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
     Returns:
         Pipeline instance with component instances and specified estimator
 
+    Example:
+        >>> components = [Imputer(), StandardScaler(), CatBoostClassifier()]
+        >>> pipeline = make_pipeline_from_components(components, problem_type="binary")
+        >>> pipeline.describe()
+        >>> assert pipeline.components_graph == components
+
     """
     if not isinstance(component_instances[-1], Estimator):
         raise ValueError("Pipeline needs to have an estimator at the last position of the component list")
@@ -129,7 +135,4 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
     class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
         custom_name = pipeline_name
         component_graph = [c.__class__ for c in component_instances]
-
-    pipeline_instance = TemplatedPipeline({})
-    pipeline_instance.component_graph = component_instances
-    return pipeline_instance
+    return TemplatedPipeline({c.name: c.parameters for c in component_instances})
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -7,14 +7,16 @@ from .multiclass_classification_pipeline import (
 from .regression_pipeline import RegressionPipeline
 
 from evalml.model_family import ModelFamily
-from evalml.pipelines.components import (
+from evalml.pipelines.components import (  # noqa: F401
     CatBoostClassifier,
     CatBoostRegressor,
+    ComponentBase,
     DateTimeFeaturizer,
     DropNullColumns,
     Estimator,
     Imputer,
     OneHotEncoder,
+    RandomForestClassifier,
     StandardScaler
 )
 from evalml.pipelines.components.utils import get_estimators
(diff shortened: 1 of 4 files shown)


evalml/pipelines/utils.py
```diff
@@ -119,6 +119,11 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
     Returns:
         Pipeline instance with component instances and specified estimator
 
+    Example:
+        >>> components = [Imputer(), StandardScaler(), Estimator()]
```
dsherry, line 123: I think you should say `RandomForestClassifier()` here, otherwise this example isn't runnable because `Estimator()` is abstract and will throw. Actually I thought we had docstrings get run during our docs, strange that this didn't fail!
```python
In [2]: evalml.pipelines.components.Estimator()
<ipython-input-2-81c4f09910a6> in <module>
----> 1 evalml.pipelines.components.Estimator()

TypeError: Can't instantiate abstract class Estimator with abstract methods model_family, name, supported_problem_types
```

evalml/pipelines/utils.py
```diff
@@ -129,7 +134,4 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
     class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
         custom_name = pipeline_name
         component_graph = [c.__class__ for c in component_instances]
-
-    pipeline_instance = TemplatedPipeline({})
-    pipeline_instance.component_graph = component_instances
-    return pipeline_instance
+    return TemplatedPipeline({c.name: c.parameters for c in component_instances})
```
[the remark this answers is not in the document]
dsherry: @freddyaboulton good idea on the fitting. Unfortunately I don't think we can have this method return a fitted pipeline. Our pipeline classes do more than just wrap the components at this point. We do things like encoding classification labels to `[0, 1, 2, ...]` to prevent modeling errors.

A test case which would break in this case:
```python
X, y = X_y_binary
# convert each 0 or 1 int value to a str
y = np.array(list(map(str, y)), dtype='str')
...
```

In other words, there are structures on the pipeline which need the training data. There are ways around this: we could try to represent the target encoding as a component in the graph just like any other. The thing is, we'd need to define a decode component on the output, and that would get hairy, so we opted for making our pipeline "smart" enough to encode the target to ints for classification targets.

So, let's update the docs for this method to say that it returns an untrained pipeline instance, even if the components are already trained.

Presumably, if someone can train the components, they should be able to get access to the training data anyways, so this shouldn't be a big deal.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -280,6 +281,19 @@ class DummyEstimator(Estimator):
     expected_parameters = {'Dummy!': {'bar': 'baz'}}
     assert pipeline.parameters == expected_parameters
 
+    X, y = X_y_binary
+    automl = AutoMLSearch(problem_type='binary', objective="Log Loss Binary", max_pipelines=5)
+    automl.search(X, y)
```
dsherry, line 286: @christopherbunn why use automl in this unit test? You can just create a pipeline class to use:
```python
class TestPipeline(BinaryClassificationPipeline):
    component_graph = [Imputer, OneHotEncoder, LogisticRegressionClassifier]

pipeline = TestPipeline(parameters={})
...
```
Or, use one already defined in our test fixtures, like [this one](https://github.com/alteryx/evalml/blob/main/evalml/tests/conftest.py#L233).

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -280,6 +281,19 @@ class DummyEstimator(Estimator):
...
+    best_pipeline = automl.best_pipeline
+    new_pipeline = make_pipeline_from_components(best_pipeline.component_graph, ProblemTypes.BINARY)
+    assert best_pipeline.describe() == new_pipeline.describe()
+    best_pipeline.fit(X, y)
+    best_predictions = best_pipeline.predict(X)
+    new_pipeline.fit(X, y)
+    new_predictions = new_pipeline.predict(X)
+    assert np.array_equal(best_predictions, new_predictions)
+    assert np.array_equal(best_pipeline.feature_importance, new_pipeline.feature_importance)
```
dsherry, line 295: Cool that you're comparing the predictions! Can you also compare both static and instance pipeline attributes: `name`, `parameters`, `summary`, `component_graph` (can just compare each element with `isinstance` or something), `describe`

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -244,7 +245,7 @@ def test_make_pipeline_problem_type_mismatch():
         make_pipeline(pd.DataFrame(), pd.Series(), Transformer, ProblemTypes.MULTICLASS)
 
 
-def test_make_pipeline_from_components():
+def test_make_pipeline_from_components(X_y_binary):
```
dsherry, line 248: @christopherbunn this test is great! What are some other cases you could cover? Here's a few thoughts on edge cases:
def make_pipeline_from_components(component_instances, problem_type, custom_name=None):
* What if `component_instances` contains class references instead of instances? Or generally things which aren't evalml component instances?
* What if `problem_type` is invalid?
* What if `custom_name` is not a string?

In all of those cases, I'd hope we'd get some sort of exception when we try to define the pipeline template class. As it stands, because our `PipelineBaseMeta` metaclass is pretty simple and doesn't check these fields, I bet no error comes up for some, so perhaps `make_pipeline_from_components` should error out instead.

dsherry (approved):
Thanks for chasing this down!

I liked your discussion of using fitted components. I left my recommendation on what we should do, which is to punt on having this method create a fitted pipeline from fitted components. I also left unit test comments. Otherwise, good stuff!

evalml/tests/pipeline_tests/test_pipelines.py
christopherbunn: Hmm, I think the intent was to originally compare fitted pipelines, but then I guess that gets fitted separately anyways. I'll update this.

evalml/pipelines/utils.py
christopherbunn: Ah yeah this did fail here. I'm having an interesting problem with this end estimator in the docs. `Estimator` doesn't work for the reasons you stated, but linting keeps catching `RandomForestClassifier` as an "unused" import since it isn't used in the code itself. I tried `Catboost` but that would fail when there during the core dependencies check. 

I think the best solution for now is to import `RandomForestClassifier` and put in a `#noqa: F401` after this line.

PR #1215, opened 2020-09-23
Fix SimpleImputer error which occurs when all features are bool type

### LINKED ISSUE #1166: Boolean column to Imputer breaks if it is the only categorical column
If only a boolean column is passed to a SimpleImputer, or if a boolean column is the only non-numeric column passed to an Imputer, it breaks:

Repro: the following code fails with a `ValueError: SimpleImputer does not support data with dtype bool. Please provide either a numeric array (with a floating point or integer dtype) or categorical data represented either as an array with integer dtype or an array of string values with an object dtype.` check.
```
X = pd.DataFrame({
        "bool col with nan": pd.Series([True, np.nan, False, np.nan, True], dtype='bool'),
    })
imputer = Imputer()
imputer.fit(X, y)
```

However, as soon as we add another column (object, category, with or without np.nan), the imputer works fine:
```
X = pd.DataFrame({
        "object with nan": ["b", "b", np.nan, "c", np.nan],
        "bool col with nan": pd.Series([True, np.nan, False, np.nan, True], dtype='bool'),
    })
imputer = Imputer()
imputer.fit(X, y)
```

The stack trace is as follows:
```
<ipython-input-36-10afc6697657> in <module>
     23     })
     24 imputer = Imputer()
---> 25 imputer.fit(X, y)
     26 # lp.fit(X, y)

     12         @wraps(method)
     13         def _set_fit(self, X, y=None):
---> 14             return_value = method(self, X, y)
     15             self._is_fitted = True
     16             return return_value

~/Desktop/evalml/evalml/pipelines/components/transformers/imputers/imputer.py in fit(self, X, y)
     78         if len(X_categorical.columns) > 0:
     79             import pdb; pdb.set_trace()
---> 80             self._categorical_imputer.fit(X_categorical, y)
     81             self._categorical_cols = X_categorical.columns
     82         return self

     12         @wraps(method)
     13         def _set_fit(self, X, y=None):
---> 14             return_value = method(self, X, y)
     15             self._is_fitted = True
     16             return return_value

~/Desktop/evalml/evalml/pipelines/components/transformers/imputers/simple_imputer.py in fit(self, X, y)
     47         X = X.fillna(value=np.nan)
     48 
---> 49         self._component_obj.fit(X, y)
     50         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
     51         return self

~/Desktop/venv/lib/python3.7/site-packages/sklearn/impute/_base.py in fit(self, X, y)
    266         self : SimpleImputer
    267         """
--> 268         X = self._validate_input(X)
    269         super()._fit_indicator(X)
    270 

~/Desktop/venv/lib/python3.7/site-packages/sklearn/impute/_base.py in _validate_input(self, X)
    249                              "categorical data represented either as an array "
    250                              "with integer dtype or an array of string values "
--> 251                              "with an object dtype.".format(X.dtype))
    252 
    253         return X

ValueError: SimpleImputer does not support data with dtype bool. Please provide either a numeric array (with a floating point or integer dtype) or categorical data represented either as an array with integer dtype or an array of string values with an object dtype.
```

Not quite sure what we can do, given this looks like a check on the scikit-learn side (via `_validate_input`); we could pass in a fake column to appease this check?

christopherbunn:
Fixed issues with all-bool inputs for `Imputer` and `SimpleImputer` by internally converting to the category datatype and reconverting to bools at the end. 

Resolves #1166

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -39,6 +39,7 @@ Release Notes
         * Updated calls to ``Woodwork``'s ``to_pandas()`` to ``to_series()`` and ``to_dataframe()`` :pr:`1428`
         * Fixed bug in OHE where column names were not guaranteed to be unique :pr:`1349`
         * Fixed bug with percent improvement of ``ExpVariance`` objective on data with highly skewed target :pr:`1467`
+        * Added conversion of all bool to categories internally for imputers :pr:`1215`
     * Changes
         * Changed ``OutliersDataCheck`` to return the list of columns, rather than rows, that contain outliers :pr:`1377`
         * Simplified and cleaned output for Code Generation :pr:`1371`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -39,7 +39,7 @@ Release Notes
         * Updated calls to ``Woodwork``'s ``to_pandas()`` to ``to_series()`` and ``to_dataframe()`` :pr:`1428`
         * Fixed bug in OHE where column names were not guaranteed to be unique :pr:`1349`
         * Fixed bug with percent improvement of ``ExpVariance`` objective on data with highly skewed target :pr:`1467`
-        * Added conversion of all bool to categories internally for imputers :pr:`1215`
+        * Fix SimpleImputer error which occurs when all features are bool type :pr:`1215`
     * Changes
         * Changed ``OutliersDataCheck`` to return the list of columns, rather than rows, that contain outliers :pr:`1377`
         * Simplified and cleaned output for Code Generation :pr:`1371`
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -71,7 +71,6 @@ class Imputer(Transformer):
 
         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
         X_copy = X.copy()
-
         X_null_dropped = X_copy.drop(self._all_null_cols, axis=1, errors='ignore')
 
         X_numerics = X_null_dropped[[col for col in numeric_cols if col not in self._all_null_cols]]
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -72,9 +72,6 @@ class Imputer(Transformer):
         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
         X_copy = X.copy()
 
-        if (X.dtypes == bool).all():
-            X_copy = X_copy.astype('category')
-
         X_null_dropped = X_copy.drop(self._all_null_cols, axis=1, errors='ignore')
 
         X_numerics = X_null_dropped[[col for col in numeric_cols if col not in self._all_null_cols]]
@@ -103,10 +100,6 @@ class Imputer(Transformer):
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
         X_null_dropped = X.copy()
-
-        if (X.dtypes == bool).all():
-            return X
-
         X_null_dropped.drop(self._all_null_cols, inplace=True, axis=1, errors='ignore')
         X_null_dropped.reset_index(inplace=True, drop=True)
         if X_null_dropped.empty:
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -71,6 +71,10 @@ class Imputer(Transformer):
 
         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
         X_copy = X.copy()
+
+        if (X.dtypes == bool).all():
+            X_copy = X_copy.astype('category')
+
         X_null_dropped = X_copy.drop(self._all_null_cols, axis=1, errors='ignore')
 
         X_numerics = X_null_dropped[[col for col in numeric_cols if col not in self._all_null_cols]]
@@ -99,6 +103,11 @@ class Imputer(Transformer):
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
         X_null_dropped = X.copy()
+
+        all_bool = (X.dtypes == bool).all()
+        if all_bool:
+            X_null_dropped = X_null_dropped.astype('category')
+
         X_null_dropped.drop(self._all_null_cols, inplace=True, axis=1, errors='ignore')
         X_null_dropped.reset_index(inplace=True, drop=True)
         if X_null_dropped.empty:
@@ -111,5 +120,8 @@ class Imputer(Transformer):
         if self._categorical_cols is not None and len(self._categorical_cols) > 0:
             X_categorical = X_null_dropped[self._categorical_cols]
             X_null_dropped[X_categorical.columns] = self._categorical_imputer.transform(X_categorical)
+
+        if all_bool:
+            X_null_dropped = X_null_dropped.astype(bool)
 
         return X_null_dropped
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -104,9 +104,8 @@ class Imputer(Transformer):
 
         X_null_dropped = X.copy()
 
-        all_bool = (X.dtypes == bool).all()
-        if all_bool:
-            X_null_dropped = X_null_dropped.astype('category')
+        if (X.dtypes == bool).all():
+            return X
 
         X_null_dropped.drop(self._all_null_cols, inplace=True, axis=1, errors='ignore')
         X_null_dropped.reset_index(inplace=True, drop=True)
@@ -120,8 +119,5 @@ class Imputer(Transformer):
         if self._categorical_cols is not None and len(self._categorical_cols) > 0:
             X_categorical = X_null_dropped[self._categorical_cols]
             X_null_dropped[X_categorical.columns] = self._categorical_imputer.transform(X_categorical)
-
-        if all_bool:
-            X_null_dropped = X_null_dropped.astype(bool)
 
         return X_null_dropped
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -48,6 +48,7 @@ class SimpleImputer(Transformer):
         X = _convert_to_woodwork_structure(X)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
+        # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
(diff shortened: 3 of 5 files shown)


jeremyliweishih (approved):
cool looks good!

evalml/tests/component_tests/test_imputer.py
```diff
@@ -286,3 +286,26 @@ def test_imputer_with_none():
...
+    assert_frame_equal(X_expected_arr, X_t)
+
+    X_multi = pd.DataFrame({
+        "bool with nan": pd.Series([True, np.nan, False, np.nan, False]),
+        "bool no nan": pd.Series([False, False, False, False, True]),
+    }, dtype=bool)
+    imputer = Imputer()
+    imputer.fit(X_multi, y)
+    X_multi_expected_arr = pd.DataFrame({
+        "bool with nan": pd.Series([True, True, False, True, False]),
```
dsherry, line 307: Why were nans here filled with `True` rather than `False`? Your example data had one `True` and two `False`. I thought the default categorical strategy of `most_frequent` would fill nans with `False`.

evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -67,6 +67,10 @@ def fit(self, X, y=None):
 
         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
         X_copy = X.copy()
+
+        if (X.dtypes == bool).all():
+            X_copy = X_copy.astype('category')
```
dsherry, line 72: @angela97lin heads up these changes will cause conflicts for #1423 (update all components to accept woodwork datatables). I think they should be pretty straightforward to address--when you're resolving the conflicts you could call `dt.set_logical_types(...)` to do this transformation, or you could keep doing the transformation with pandas. Great that @christopherbunn is adding test coverage for this case, that'll help!

dsherry (approved):
@christopherbunn thank you, looks good! I left one question about the behavior of the imputer for boolean inputs, may be a bug there.

evalml/pipelines/components/transformers/imputers/imputer.py
dsherry: Ah, #1423 was just merged!

@christopherbunn if you need a hand updating this, please ping @angela97lin or I

angela97lin: Oo yup, just saw this. Yes yes, please reach out if you need help fixing the merge conflicts!

christopherbunn:
@dsherry Thanks for the review! I looked at the test case a bit deeper and part of the reason why that passes is that np.nan actually gets casted to True when the data type is set to bool. As an example, this snippet below

```
    X_multi = pd.DataFrame({
        "bool with nan": pd.Series([True, np.nan, False, np.nan, False], dtype=bool),
        "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
    })
``` 
is actually equal to
```
    pd.DataFrame({
        "bool with nan": pd.Series([True, True, False, True, False], dtype=bool),
        "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
    })
```
Because of this, I'm not actually sure if its possible to impute a bool-only df like `X_multi` since it will always be filled with values that are either True or False. `X_multi` could exist with True, False, and np.nan values but only if the dtype is `Categorical` from the beginning. In that case, the existing code would already work.

I think that resolving this issue should instead focus on just catching this value error. The current HEAD of this PR will just return all-bool dfs as is without imputing it and I added a new case to test this behavior. Thoughts @angela97lin and @dsherry?

angela97lin:
@christopherbunn Yup, I'd agree that resolving this issue to just catch the ValueError sounds good to me. It's a limitation or design choice made by pandas, and one I came across as well while implementing Woodwork stuff. Now that we have woodwork integration in place for components, users can use Woodwork structures, which use the new nullable pd.NA, and this would be converted to a category (or pandas "object") column, which you've mentioned already works. (Maybe this is a worthwhile test to add for the imputer specifically?)

dsherry:
Discussed with @angela97lin @christopherbunn . Recommendations:
* For `SimpleImputer`, if we get this error, catch the error and return the original data
* Unit tests should cover both pandas as input and woodwork as input

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -47,6 +47,10 @@ def fit(self, X, y=None):
         """
         X = _convert_to_woodwork_structure(X)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
+
+        if (X.dtypes == bool).all():
```
dsherry, line 51: [Suggested alternative:](https://pandas.pydata.org/pandas-docs/version/0.25.3/reference/api/pandas.api.types.is_bool_dtype.html) `pd.ptypes.is_bool_dtype(X).all()`

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -70,6 +74,8 @@ def transform(self, X, y=None):
         # Convert None to np.nan, since None cannot be properly handled
         X = X.fillna(value=np.nan)
 
+        if (X.dtypes == bool).all():
```
dsherry, line 77: Since this is kinda confusing, perhaps add a comment saying "bool dtype doesn't support nans, and sklearn errors if all cols are bool, so return early"

evalml/tests/component_tests/test_imputer.py
```diff
@@ -291,3 +291,44 @@ def test_imputer_with_none():
     imputer = Imputer()
     transformed = imputer.fit_transform(X, y)
     assert_frame_equal(transformed, expected, check_dtype=False)
+
+
+@pytest.mark.parametrize("data_type", ['pd', 'ww'])
+def test_imputer_all_bool(data_type):
```
dsherry, line 297: Since `Imputer` uses `SimpleImputer` under the hood, and you're putting the fix into `SimpleImputer`, I think the coverage of `SimpleImputer` is enough and you can delete this copy of the unit test.

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -102,6 +103,45 @@ def test_simple_imputer_col_with_non_numeric():
...
+    assert_frame_equal(X_expected_arr, X_t)
+
+    X = pd.DataFrame([True, np.nan, False, np.nan, True], dtype=object)
+    X_expected_arr = pd.DataFrame([True, True, False, True, True], dtype='category')
+    if data_type == 'ww':
+        X = ww.DataTable(X)
+    imputer = SimpleImputer()
+    imputer.fit(X, y)
+    X_t = imputer.transform(X)
+    assert_frame_equal(X_expected_arr, X_t)
```
dsherry, line 126: This test is great!

I think the first two cases (i.e. above this line) are fully covered by the last case, right? I suggest you delete the first two cases, above this line, and keep the last one below.

docs/source/release_notes.rst
```diff
@@ -25,6 +25,7 @@ Release Notes
         * Updated enum classes to show possible enum values as attributes :pr:`1391`
         * Updated calls to ``Woodwork``'s ``to_pandas()`` to ``to_series()`` and ``to_dataframe()`` :pr:`1428`
         * Fixed bug in OHE where column names were not guaranteed to be unique :pr:`1349`
+        * Added conversion of all bool to categories internally for imputers :pr:`1215`
```
dsherry, line 28: Suggest "Fix SimpleImputer error which occurs when all features are bool type"

dsherry (approved):
I left a few suggestions which we should resolve before merging but LGTM!

freddyaboulton (approved):
@christopherbunn I think this is fantastic! Thanks for making this change. I think your unit test for the `Imputer`  shows that `SimpleImputer` and `Imputer` convert dtypes between input and output differently (even though `Imputer` calls `SimpleImputer` under the hood). 

I don't think that should block merge (maybe that behavior existed before this change) so maybe the best thing is to address @dsherry 's comments (and any others that come up) and file an issue to continue the discussion.

evalml/tests/component_tests/test_imputer.py
christopherbunn: Yep, my understanding is that while `Imputer` does use `SimpleImputer` under the hood, the output of `Imputer` is different as it recasts everything to the proper type. I don't see an issue with changing this in the future.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
christopherbunn: It might not have the same behavior, as it checks the individual data type rather than the series/dataframe type. As such, this would return True for types that are able to be handled by the imputer just fine (such as category). 

image

evalml/tests/component_tests/test_simple_imputer.py
christopherbunn: I'd argue for keeping all of the test cases. The first test case triggers the type conversion while the second one checks for proper boolean impution. The third one is more of an extension of the second one but it also combines a complete boolean column to make sure that the types are being processed properly.

evalml/tests/component_tests/test_imputer.py
```diff
@@ -291,3 +291,44 @@ def test_imputer_with_none():
...
+    X_multi_expected_arr = pd.DataFrame({
+        "bool with nan": pd.Series([True, False, False, False, False], dtype=object),
+        "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
+    })
+    if data_type == 'ww':
+        X_multi = ww.DataTable(X_multi)
+    imputer = Imputer()
+    imputer.fit(X_multi, y)
+    X_multi_t = imputer.transform(X_multi)
+    assert_frame_equal(X_multi_expected_arr, X_multi_t)
```
dsherry, line 334: Unit tests discussed with @christopherbunn 
1. All cols are bool-type
`X = pd.DataFrame([True, True, False, True, True], dtype=bool)`
2. Imputation strategies work for bool type
`X = pd.DataFrame([True, np.nan, False, np.nan, True], dtype=object)`
`X = pd.DataFrame([True, np.nan, False, np.nan, False], dtype=object)`
3. Multi-type, with at least one bool typed col
```python
    X_multi_expected_arr = pd.DataFrame({
        "bool with nan": pd.Series([True, False, False, False, False], dtype=object),
        "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
    })
```

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -70,6 +74,9 @@ def transform(self, X, y=None):
         # Convert None to np.nan, since None cannot be properly handled
         X = X.fillna(value=np.nan)
 
+        # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
```
dsherry, line 77: @christopherbunn should this comment be in `fit` too?

dsherry (approved):
@christopherbunn thanks for chasing this down!

PR #1218, opened 2020-09-23
Add `__str__` and `__repr__` for components and pipelines

### LINKED ISSUE #474: Add `__str__`+`__repr__` for `ComponentBase`+`PipelineBase`
- __str__
    - Printing a pipeline or component instance should quickly identify which pipeline or component it is.
    - Components
        - Use ComponentBase.name
     - Pipelines
        - Use PipelineBase.summary
- __repr__
    - Calling repr() on a pipeline or component should produce useful output for debugging.
    - Components
        - print Component and parameters like python expression
        - ComponentBase(param1=1, param2=2)
    - Pipelines
            - print python expression with parameters dict and objective
            - LogisticRegressionPipeline(parameters={...}, objective='precision
                - parameters = {'LGC': {'C':1.5}, 'SI':{'Impute Strategy': 'mean}}

```
// component
def __str__(self):
        return self.name 

// pipeline
def __str__(self):
        return self.name

// component
def __repr__(self):
        rpr = f"{(type(self).__name__)}("
        for parameter, value in self.parameters.items():
             rpr = rpr + f"{parameter}={value},"
        rpr = + ')'
        return rpr
    
// pipeline
def __repr__(self):
        rpr = f"{(type(self).__name__)}(parameters={"
        for component, parameters in self.parameters.items():
            rpr = rpr + f"'\n{component}':{"
            for parameter, value in parameters:
                rpr = rpr + f"\n{parameter}={value},"
            rpr = rpr + "},"
         rpr = + ')'
         return rpr
```

eccabay:
Closes #474

### DIFF
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -10,7 +10,8 @@ from evalml.utils import (
     classproperty,
     get_logger,
     get_random_state,
-    log_subtitle
+    log_subtitle,
+    safe_repr
 )
 
 logger = get_logger(__file__)
@@ -156,13 +157,15 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
         return self.name
 
     def __repr__(self):
-        rpr = f"{(type(self).__name__)}("
-        for parameter, value in self.parameters.items():
-            if type(value) == str:
-                rpr = rpr + f"{parameter}='{value}',"
-            elif value == float('inf') or value == float('-inf'):
-                rpr = rpr + f"{parameter}=float('{value}'),"
-            else:
-                rpr = rpr + f"{parameter}={value},"
-        rpr = rpr + ')'
-        return rpr
+        # rpr = f"{(type(self).__name__)}("
+        # for parameter, value in self.parameters.items():
+        #     if type(value) == str:
+        #         rpr = rpr + f"{parameter}='{value}',"
+        #     elif value == float('inf') or value == float('-inf'):
+        #         rpr = rpr + f"{parameter}=float('{value}'),"
+        #     else:
+        #         rpr = rpr + f"{parameter}={value},"
+        # rpr = rpr + ')'
+        # return rpr
+        parameters_repr = ', '.join([f'{key}={safe_repr(value)}' for key, value in self.parameters.items()])
+        return f'{(type(self).__name__)}({parameters_repr})'
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -158,6 +158,6 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
     def __repr__(self):
         rpr = f"{(type(self).__name__)}("
         for parameter, value in self.parameters.items():
-             rpr = rpr + f"{parameter}={value},"
+            rpr = rpr + f"{parameter}={value},"
         rpr = rpr + ')'
         return rpr
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -157,15 +157,5 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
         return self.name
 
     def __repr__(self):
-        # rpr = f"{(type(self).__name__)}("
-        # for parameter, value in self.parameters.items():
-        #     if type(value) == str:
-        #         rpr = rpr + f"{parameter}='{value}',"
-        #     elif value == float('inf') or value == float('-inf'):
-        #         rpr = rpr + f"{parameter}=float('{value}'),"
-        #     else:
-        #         rpr = rpr + f"{parameter}={value},"
-        # rpr = rpr + ')'
-        # return rpr
         parameters_repr = ', '.join([f'{key}={safe_repr(value)}' for key, value in self.parameters.items()])
         return f'{(type(self).__name__)}({parameters_repr})'
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -151,3 +151,13 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
             if getattr(self, attribute) != getattr(other, attribute):
                 return False
         return True
+
+    def __str__(self):
+        return self.name
+
+    def __repr__(self):
+        rpr = f"{(type(self).__name__)}("
+        for parameter, value in self.parameters.items():
+             rpr = rpr + f"{parameter}={value},"
+        rpr = rpr + ')'
+        return rpr
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -158,6 +158,11 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
     def __repr__(self):
         rpr = f"{(type(self).__name__)}("
         for parameter, value in self.parameters.items():
-            rpr = rpr + f"{parameter}={value},"
+            if type(value) == str:
+                rpr = rpr + f"{parameter}='{value}',"
+            elif value == float('inf') or value == float('-inf'):
+                rpr = rpr + f"{parameter}=float('{value}'),"
+            else:
+                rpr = rpr + f"{parameter}={value},"
         rpr = rpr + ')'
         return rpr
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -27,7 +27,8 @@ from evalml.utils import (
     import_or_raise,
     jupyter_check,
     log_subtitle,
-    log_title
+    log_title,
+    safe_repr
 )
 
 logger = get_logger(__file__)
@@ -492,16 +493,9 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         return self.name
 
     def __repr__(self):
-        rpr = f"{(type(self).__name__)}(parameters={{"
-        for component, parameters in self.parameters.items():
-            rpr = rpr + f"'{component}':{{"
-            for parameter, value in parameters.items():
-                if type(value) == str:
-                    rpr = rpr + f"'{parameter}': '{value}', "
-                elif value == float('inf') or value == float('-inf'):
-                    rpr = rpr + f"'{parameter}': float('{value}'), "
-                else:
-                    rpr = rpr + f"'{parameter}': {value}, "
-            rpr = rpr + "}, "
-        rpr = rpr + '})'
-        return rpr
+
+        def repr_component(parameters):
+            return ', '.join([f"'{key}': {safe_repr(value)}" for key, value in parameters.items()])
+
+        parameters_repr = ' '.join([f"'{component}':{{{repr_component(parameters)}}}," for component, parameters in self.parameters.items()])
+        return f'{(type(self).__name__)}(parameters={{{parameters_repr}}})'
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -498,6 +498,8 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
             for parameter, value in parameters.items():
                 if type(value) == str:
                     rpr = rpr + f"'{parameter}': '{value}', "
+                elif value == float('inf') or value == float('-inf'):
+                    rpr = rpr + f"'{parameter}': float('{value}'), "
                 else:
                     rpr = rpr + f"'{parameter}': {value}, "
             rpr = rpr + "}, "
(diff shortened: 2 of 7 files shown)


evalml/pipelines/components/component_base.py
```diff
@@ -151,3 +151,18 @@ def __eq__(self, other):
...
+        rpr = f"{(type(self).__name__)}("
+        for parameter, value in self.parameters.items():
+            if type(value) == str:
+                rpr = rpr + f"{parameter}='{value}',"
+            elif value == float('inf') or value == float('-inf'):
+                rpr = rpr + f"{parameter}=float('{value}'),"
+            else:
+                rpr = rpr + f"{parameter}={value},"
+        rpr = rpr + ')'
+        return rpr
```
dsherry, line 168: @eccabay good call looking into `float('inf')`!! Another one is `float('nan')`. You've hit on [what appears to be a funny bug and/or design flaw with python's `float` type](https://bugs.python.org/issue22951): `repr(float('inf'))` comes back as `'inf'`, which... is just messed up! 🤔   Same for `float('nan')`. I think they did this so that `float(repr(float('inf')))` is equivalent to `float('inf')`, and same for nan.

The same is true for `np.inf` and `np.nan`, by the way:
```python
In [6]: parameters = {'int': 42, 'string': 'string', 'float': 3.14159,
                      'inf': float('inf'), 'np inf': np.inf, 'np nan': np.nan}

In [7]: repr(parameters)
Out[7]: "{'int': 42, 'string': 'string', 'float': 3.14159, 'inf': inf, 'np inf': inf, 'np nan': nan}"
```

---
So, what do we do? It would be really nice if we could have what comes out of `repr(component)` be something we could paste into a terminal in order to get an identical instance.

I like the approach you're taking. I played with it for a bit and here's what I got:
```python
# define this in gen_utils so we can use it for the pipeline repr too
def safe_repr(value):
    if isinstance(value, float):
        if pd.isna(value):
            return 'np.nan'
        if np.isinf(value):
            return f'float({repr(value)})'
    return repr(value)

# then the component repr:
def __repr__(self):
    parameters_repr = ', '.join([f'{key}={safe_repr(value)}' for key, value in self.parameters.items()])
    return f'{self.name}({parameters_repr})'
```

Here's a test I did of the `safe_repr`, which shows that it outputs the string we need in order to avoid some ugly bugs:
```
In [1]: [safe_repr(el) for el in ['string', float('nan'), float('-inf'), float('inf'), np.nan, -np.inf, np.inf]]
Out[1]:
["'string'", 'np.nan', 'float(-inf)', 'float(inf)', 'np.nan', 'float(-inf)', 'float(inf)']
```

What do you think of that? Weird stuff!

evalml/pipelines/pipeline_base.py
```diff
@@ -487,3 +487,21 @@ def __eq__(self, other):
...
+            for parameter, value in parameters.items():
+                if type(value) == str:
+                    rpr = rpr + f"'{parameter}': '{value}', "
+                elif value == float('inf') or value == float('-inf'):
+                    rpr = rpr + f"'{parameter}': float('{value}'), "
+                else:
+                    rpr = rpr + f"'{parameter}': {value}, "
+            rpr = rpr + "}, "
+        rpr = rpr + '})'
+        return rpr
```
dsherry, line 507: Same comment as for component repr, let's flatten this out, and great thinking trying nan/inf types!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1270,3 +1270,48 @@ def test_pipeline_equality_different_fitted_data(problem_type, X_y_binary, X_y_m
...
+
+    pipeline = MockPipeline(parameters={})
+    print(repr(pipeline))
+    assert eval(repr(pipeline)) == pipeline
+
+    pipeline_with_parameters = MockPipeline(parameters={'Imputer': {'numeric_fill_value': 42}})
+    assert eval(repr(pipeline_with_parameters)) == pipeline_with_parameters
+
+    pipeline_with_inf_parameters = MockPipeline(parameters={'Imputer': {'numeric_fill_value': float('inf')}})
+    assert eval(repr(pipeline_with_inf_parameters)) == pipeline_with_inf_parameters
```
dsherry, line 1317: This test is pretty cool! Great thinking enforcing this invariant on `repr`

Two things:
1. RE our discussion of nan types at the top, it would be ideal if we had a similar test to this one which checks all those edge cases. Example: define a mock component which has a few input parameters, and pass in `float('nan')`/`np.nan` and the native/np infs.

2. There are [security concerns with using python's `eval`](https://stackoverflow.com/questions/3513292/python-make-eval-safe). It turns out if you can compromise what goes into `eval` you can do a lot of bad stuff. To avoid this entirely, my recommendation is to avoid using `eval` and just check the string output matches what we expect. I know, this is test code so it feels kinda silly, lol--[related reading](https://rules.sonarsource.com/python/type/Security%20Hotspot/RSPEC-1523) if you're interested.

dsherry (changes requested):
@eccabay great work on this! There's some seriously tricky stuff going on here for a feature which looks so simple from the outside. It's really cool we're doing this :)

I left a suggested change on the impl along with some discussion. I also left a couple items on testing. Should be all set after that IMO.

evalml/pipelines/components/component_base.py
```diff
@@ -151,3 +152,20 @@ def __eq__(self, other):
...
+        # rpr = f"{(type(self).__name__)}("
+        # for parameter, value in self.parameters.items():
+        #     if type(value) == str:
+        #         rpr = rpr + f"{parameter}='{value}',"
+        #     elif value == float('inf') or value == float('-inf'):
+        #         rpr = rpr + f"{parameter}=float('{value}'),"
+        #     else:
+        #         rpr = rpr + f"{parameter}={value},"
+        # rpr = rpr + ')'
+        # return rpr
```
dsherry, line 169: Please delete this dead code

evalml/utils/gen_utils.py
```diff
@@ -232,3 +233,12 @@ def jupyter_check():
         return True
     except NameError:
         return False
+
+
+def safe_repr(value):
```
dsherry, line 238: Please add a docstring!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1270,3 +1270,54 @@ def test_pipeline_equality_different_fitted_data(problem_type, X_y_binary, X_y_m
...
+    assert repr(pipeline_with_parameters) == expected_repr
+
+    pipeline_with_inf_parameters = MockPipeline(parameters={'Imputer': {'numeric_fill_value': float('inf'), 'categorical_fill_value': np.inf}})
+    expected_repr = f"MockPipeline(parameters={{'Imputer':{{'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': float('inf'), 'numeric_fill_value': float('inf')}}, '{final_estimator}':{{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}},}})"
+    assert repr(pipeline_with_inf_parameters) == expected_repr
+
+    pipeline_with_nan_parameters = MockPipeline(parameters={'Imputer': {'numeric_fill_value': float('nan'), 'categorical_fill_value': np.nan}})
+    expected_repr = f"MockPipeline(parameters={{'Imputer':{{'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': np.nan, 'numeric_fill_value': np.nan}}, '{final_estimator}':{{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}},}})"
+    assert repr(pipeline_with_nan_parameters) == expected_repr
```
dsherry, line 1323: Looks great! I left a comment on the impl about how to get rid of the double curly braces here.

evalml/pipelines/pipeline_base.py
```diff
@@ -487,3 +488,14 @@ def __eq__(self, other):
...
+    def __str__(self):
+        return self.name
+
+    def __repr__(self):
+
+        def repr_component(parameters):
+            return ', '.join([f"'{key}': {safe_repr(value)}" for key, value in parameters.items()])
+
+        parameters_repr = ' '.join([f"'{component}':{{{repr_component(parameters)}}}," for component, parameters in self.parameters.items()])
+        return f'{(type(self).__name__)}(parameters={{{parameters_repr}}})'
```
dsherry, line 501: I believe [you want double curly braces instead of triple](https://stackoverflow.com/questions/5466451/how-can-i-print-literal-curly-brace-characters-in-python-string-and-also-use-fo) here.

dsherry (approved):
Sweet!! Approved pending a couple small things:
* Remove extra curly braces from pipeline repr
* Add docstring for `safe_repr`

evalml/pipelines/pipeline_base.py
eccabay: Since the parameters is a dictionary and contains an expression, I do need all three! The first two are for literal curly braces and the third is for the formatting.

dsherry: Hm interesting, but then the output (taken from the unit test) ends up as
```python
MockPipeline(parameters={{'Imputer':{{'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': None, 'numeric_fill_value': None}}, '{final_estimator}':{{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}},}})
```

But ideally I think it should be
```python
MockPipeline(parameters={'Imputer': {'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': None, 'numeric_fill_value': None}, 'final_estimator': {'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}})
```

Do you agree? The second output could be evaluated in the python repl and turned into an object. I think the first would fail.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1267,3 +1267,54 @@ def test_pipeline_equality_different_fitted_data(problem_type, X_y_binary, X_y_m
...
+        final_estimator = 'Random Forest Classifier'
+    else:
+        final_estimator = 'Random Forest Regressor'
+
+    class MockPipeline(pipeline_class):
+        name = "Mock Pipeline"
+        component_graph = ['Imputer', final_estimator]
+
+    pipeline = MockPipeline(parameters={})
+    expected_repr = f"MockPipeline(parameters={{'Imputer':{{'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': None, 'numeric_fill_value': None}}, '{final_estimator}':{{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}},}})"
```
dsherry, line 1307: @eccabay I just noticed: can we replace `'{final_estimator}':` with `'final_estimator':`?

dsherry: Doesn't block merge though IMO! Just a detail we should probably chase down

evalml/pipelines/pipeline_base.py
dsherry: @eccabay but, I think its fine to merge this and we can circle back. It adds value even if there's a couple details we may wanna discuss further :)

eccabay: The double curly braces in the unit tests are once again for f-string formatting! If you actually print the repr, you get the second code block you pasted there, and calling eval on the expected_repr produces the correct object.

dsherry: @eccabay hmm could you paste an example? I don't follow yet. I thought that `__repr__` returns a string, which requires no further formatting, and that's the end of it 😂  In other words, I thought we should output the second snippet, because that can be pasted into the python REPL and evaluated, but we're currently outputting the first.

Again, not blocking merge, I'm just trying to make sure we all agree on desired behavior.

evalml/tests/pipeline_tests/test_pipelines.py
eccabay: Unfortunately the way these tests are written, we can't. The string `__repr__` prints out uses the name of the estimator, so checking for string equality will fail. If I replace `{final_estimator}` with `final_estimator` and try `assert eval(repr(pipeline)) == eval(expected_repr)`, that will pass, but that will reintroduce `eval` to the code.

The alternative of having `__repr__` output `final_estimator` instead of the name of said estimator feels unnecessarily clunky.

dsherry: Oh! My bad, I didn't notice this was a format string until just now. Got it, makes sense. Thanks

eccabay: Wonderful, glad we're on the same page :sweat_smile:. Does this also clear up our other discussion?

evalml/pipelines/pipeline_base.py
dsherry: @eccabay ok, I think I understand now. So the string I copied was a format string, which is why you need the double curlys there.

I just did a test locally and everything looks great

Screen Shot 2020-09-29 at 10 40 39 AM

Realizing the unit tests were using a format string was the key point I was missing, lol. Thanks! 😆

PR #1222, opened 2020-09-24
Update documentation about objectives in AutoML

### LINKED ISSUE #1211: Clarify what objective in AutoML is used for
As a follow-up to #1173, we should make more clear that the objective passed to AutoMLSearch is used to rank and choose pipelines, but not used to optimize individual pipelines during `fit`.

angela97lin:
Closes #1211 by clarifying in AutoMLSearch docstring and AutoML docs what the objective parameter is used for.

[Updated API reference](https://evalml.alteryx.com/en/1211_objective_docs/user_guide/automl.html) and [updated automl guide](https://evalml.alteryx.com/en/1211_objective_docs/user_guide/automl.html)

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -102,10 +102,10 @@ class AutoMLSearch:
             problem_type (str or ProblemTypes): Choice of 'regression', 'binary', or 'multiclass', depending on the desired problem type.
 
             objective (str, ObjectiveBase): The objective to optimize for. Used to select and rank models, but not for optimizing each model during fit-time.
-                When set to auto, chooses:
-                    - LogLossBinary for binary classification problems,
-                    - LogLossMulticlass for multiclass classification problems, and
-                    - R2 for regression problems.
+                When set to 'auto', chooses:
+                - LogLossBinary for binary classification problems,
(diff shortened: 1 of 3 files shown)


evalml/automl/automl_search.py
```diff
@@ -101,10 +101,12 @@ def __init__(self,
         Arguments:
             problem_type (str or ProblemTypes): Choice of 'regression', 'binary', or 'multiclass', depending on the desired problem type.
 
-            objective (str, ObjectiveBase): The objective to optimize for. When set to auto, chooses:
-                LogLossBinary for binary classification problems,
-                LogLossMulticlass for multiclass classification problems, and
-                R2 for regression problems.
+            objective (str, ObjectiveBase): The objective to optimize for. Used to select and rank models, but not for optimizing each model during fit-time.
```
freddyaboulton, line 104: not-blocking: I think it would be better to use `pipelines` instead of `model` to match the language in the user guide. Also, I think `propose` is better than `select` hehe

freddyaboulton (approved):

evalml/automl/automl_search.py
angela97lin: LOL okie dokie doke, will update!

PR #1223, opened 2020-09-24
Add DecisionTreeClassifier and DecisionTreeRegressor classes

### LINKED ISSUE #1196: Add sklearn DecisionTreeRegressor and DecisionTreeClassifier estimators
https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeRegressor.html

Params we should consider exposing to automl: max_depth, criterion, max_features.

angela97lin:
Closes #1196 by adding DecisionTreeClassifier and DecisionTreeRegressor

- [x] Add DecisionTree* classes (+API reference)

TODO in later PR: Add DecisionTree* to AutoML and do perf testing

### DIFF
--- a/evalml/model_family/model_family.py
+++ b/evalml/model_family/model_family.py
@@ -29,7 +29,7 @@ class ModelFamily(Enum):
         return "ModelFamily." + self.name
 
     def is_tree_estimator(self):
-        """Checks whether the estimator's model family uses tree ensembles."""
+        """Checks whether the estimator's model family uses trees."""
         tree_estimators = {self.CATBOOST, self.EXTRA_TREES, self.RANDOM_FOREST,
-                           self.XGBOOST, self.LIGHTGBM}
+                           self.DECISION_TREE, self.XGBOOST, self.LIGHTGBM}
         return self in tree_estimators
--- a/evalml/model_family/model_family.py
+++ b/evalml/model_family/model_family.py
@@ -9,6 +9,7 @@ class ModelFamily(Enum):
     LINEAR_MODEL = 'linear_model'
     CATBOOST = 'catboost'
     EXTRA_TREES = 'extra_trees'
+    DECISION_TREE = 'decision_tree'
     BASELINE = 'baseline'
     NONE = 'none'
 
@@ -19,6 +20,7 @@ class ModelFamily(Enum):
                              ModelFamily.LINEAR_MODEL.name: "Linear",
                              ModelFamily.CATBOOST.name: "CatBoost",
                              ModelFamily.EXTRA_TREES.name: "Extra Trees",
+                             ModelFamily.DECISION_TREE.name: "Decision Tree",
                              ModelFamily.BASELINE.name: "Baseline",
                              ModelFamily.NONE.name: "None"}
         return model_family_dict[self.name]
--- a/evalml/pipelines/components/estimators/classifiers/decision_tree_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/decision_tree_classifier.py
@@ -1,0 +1,39 @@
+from sklearn.tree import DecisionTreeClassifier as SKDecisionTreeClassifier
+from skopt.space import Integer
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+
+
+class DecisionTreeClassifier(Estimator):
+    """Decision Tree Classifier."""
+    name = "Decision Tree Classifier."
+    hyperparameter_ranges = {
+        "criterion": ["gini", "entropy"],
+        "max_features": ["auto", "sqrt", "log2"],
+        "max_depth": Integer(4, 10)
+    }
+    model_family = ModelFamily.DECISION_TREE
+    supported_problem_types = [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]
+
+    def __init__(self,
+                 criterion="gini",
+                 max_features="auto",
+                 max_depth=6,
+                 min_samples_split=2,
+                 min_weight_fraction_leaf=0.0,
+                 random_state=0,
+                 **kwargs):
+        parameters = {"criterion": criterion,
+                      "max_features": max_features,
+                      "max_depth": max_depth,
+                      "min_samples_split": min_samples_split,
+                      "min_weight_fraction_leaf": min_weight_fraction_leaf}
+        parameters.update(kwargs)
+
+        dt_classifier = SKDecisionTreeClassifier(random_state=random_state,
+                                                 **parameters)
+        super().__init__(parameters=parameters,
+                         component_obj=dt_classifier,
+                         random_state=random_state)
--- a/evalml/pipelines/components/estimators/classifiers/rf_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/rf_classifier.py
@@ -27,7 +27,3 @@ class RandomForestClassifier(Estimator):
         super().__init__(parameters=parameters,
                          component_obj=rf_classifier,
                          random_state=random_state)
-
-    @property
-    def feature_importance(self):
-        return self._component_obj.feature_importances_
--- a/evalml/pipelines/components/estimators/regressors/decision_tree_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/decision_tree_regressor.py
@@ -1,0 +1,39 @@
+from sklearn.tree import DecisionTreeRegressor as SKDecisionTreeRegressor
+from skopt.space import Integer
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+
+
+class DecisionTreeRegressor(Estimator):
+    """Decision Tree Regressor."""
+    name = "Decision Tree Regressor."
+    hyperparameter_ranges = {
+        "criterion": ["mse", "friedman_mse", "mae"],
+        "max_features": ["auto", "sqrt", "log2"],
+        "max_depth": Integer(4, 10)
+    }
+    model_family = ModelFamily.DECISION_TREE
+    supported_problem_types = [ProblemTypes.REGRESSION]
+
+    def __init__(self,
+                 criterion="mse",
+                 max_features="auto",
+                 max_depth=6,
+                 min_samples_split=2,
+                 min_weight_fraction_leaf=0.0,
+                 random_state=0,
+                 **kwargs):
+        parameters = {"criterion": criterion,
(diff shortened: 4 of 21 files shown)


evalml/model_family/model_family.py
```diff
@@ -9,6 +9,7 @@ class ModelFamily(Enum):
     LINEAR_MODEL = 'linear_model'
     CATBOOST = 'catboost'
     EXTRA_TREES = 'extra_trees'
+    DECISION_TREE = 'decision_tree'
```
jeremyliweishih, line 12: Maybe we can move towards generalizing model families - I see that we have `is_tree_estimator` below and I think it could be a good idea to put all tree based models together. Likewise with all gradient boosted machines! We should file an issue if we like that idea.

evalml/pipelines/components/estimators/classifiers/decision_tree_classifier.py
```diff
@@ -0,0 +1,39 @@
...
+from evalml.problem_types import ProblemTypes
+
+
+class DecisionTreeClassifier(Estimator):
+    """Decision Tree Classifier."""
+    name = "Decision Tree Classifier."
+    hyperparameter_ranges = {
+        "criterion": ["gini", "entropy"],
+        "max_features": ["auto", "sqrt", "log2"],
+        "max_depth": Integer(4, 10)
```
jeremyliweishih, line 15: Is `max_depth` values just a placeholder for now until we do perf testing? It seems a little low off the top of my head.

evalml/tests/component_tests/test_decision_tree_regressor.py
```diff
@@ -0,0 +1,43 @@
...
+    y_pred_sk = sk_clf.predict(X)
+
+    clf = DecisionTreeRegressor()
+    clf.fit(X, y)
+    y_pred = clf.predict(X)
+
+    np.testing.assert_almost_equal(y_pred, y_pred_sk, decimal=5)
+
+
+def test_feature_importance(X_y_binary):
```
jeremyliweishih, line 32: Extra test from the classifier here.

jeremyliweishih (approved):
LGTM. Just a couple questions and an extra test case but nothing blocking. Very cool!

evalml/model_family/model_family.py
angela97lin: Yup, this thought also crossed my mind while I was adding this and I agree! It's a bit tricky though, since we currently rely on `ModelFamily` to determine when we're dealing with `XGBoost` and `CatBoost` which both have to be handled differently in quite a few places 🤔

evalml/tests/component_tests/test_decision_tree_classifier.py
```diff
@@ -0,0 +1,64 @@
...
+from evalml.pipelines import DecisionTreeClassifier
+from evalml.problem_types import ProblemTypes
+
+
+def test_model_family():
+    assert DecisionTreeClassifier.model_family == ModelFamily.DECISION_TREE
+
+
+def test_problem_types():
+    assert ProblemTypes.BINARY in DecisionTreeClassifier.supported_problem_types
```
bchen1116, line 14: Maybe something like this would be cleaner?
``` 
assert set(DecisionTreeClassifier.supported_problem_types) == {ProblemTypes.MULTICLASS, ProblemTypes.BINARY}
```

evalml/tests/component_tests/test_decision_tree_classifier.py
```diff
@@ -0,0 +1,64 @@
...
+    assert ProblemTypes.MULTICLASS in DecisionTreeClassifier.supported_problem_types
+    assert len(DecisionTreeClassifier.supported_problem_types) == 2
+
+
+def test_fit_predict_binary(X_y_binary):
```
bchen1116, line 19: According to the documentation, "DecisionTreeClassifier is capable of both binary (where the labels are [-1, 1]) classification and multiclass (where the labels are [0, …, K-1]) classification."

Might be useful to check that these results for the binary test (where labels are [0, 1]) are equal to what the results would be for labels [-1, 1]

evalml/pipelines/components/estimators/classifiers/decision_tree_classifier.py
angela97lin: Ya I'm using the same value from our ExtraTrees components for now, don't know if there's anything better until we do some perf testing 😎

evalml/tests/component_tests/test_decision_tree_classifier.py
```diff
@@ -0,0 +1,64 @@
...
+
+    clf = DecisionTreeClassifier()
+    sk_clf = SKDecisionTreeClassifier(max_depth=6, max_features='auto', random_state=0)
+    sk_clf.fit(X, y)
+    sk_feature_importance = sk_clf.feature_importances_
+
+    clf.fit(X, y)
+    feature_importance = clf.feature_importance
+
+    np.testing.assert_almost_equal(sk_feature_importance, feature_importance, decimal=5)
```
bchen1116, line 64: Would be good to add a test for categorical data, as SKLearn's DTC fails with this currently

evalml/utils/gen_utils.py
```diff
@@ -165,7 +165,7 @@ def _get_subclasses(base_class):
     return subclasses
 
 
-_not_used_in_automl = {'BaselineClassifier', 'BaselineRegressor',
+_not_used_in_automl = {'BaselineClassifier', 'BaselineRegressor', 'DecisionTreeClassifier', 'DecisionTreeRegressor',
```
bchen1116, line 168: Nice!

evalml/tests/component_tests/test_decision_tree_regressor.py
angela97lin: oops, tyty

bchen1116 (changes requested):
Left some notes about potential tests to add. Decision Tree doesn't currently work with categorical data, so adding coverage of that would be important!

evalml/tests/component_tests/test_decision_tree_classifier.py
angela97lin: This is a good suggestion! I think we currently limit what we allow for targets to [0, 1] for binary classification and [0, …, K-1] for multiclass though, so we don't necessarily need to guarantee what happens when the user breaks that. (Best docs I can think of to show is this: https://evalml.featurelabs.com/en/latest/user_guide/components.html#Expectations-for-Custom-Classification-Components)

bchen1116: Gotcha, sounds good. Thanks!

evalml/pipelines/components/estimators/regressors/decision_tree_regressor.py
```diff
@@ -0,0 +1,39 @@
+from sklearn.tree import DecisionTreeRegressor as SKDecisionTreeRegressor
+from skopt.space import Integer
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+
+
+class DecisionTreeRegressor(Estimator):
```
freddyaboulton, line 9: If you don't mind, I'll file an issue for exposing the `tree_` property of these estimators. Looks like sklearn has cool features to visualize these models that we could leverage in our `model_understanding` module.

freddyaboulton (approved):
@angela97lin Looks good to me! Thanks for updating `is_tree_estimator` in model family and for excluding these from automl. I look forward to looking at the perf test results!

evalml/pipelines/components/estimators/regressors/decision_tree_regressor.py
angela97lin: Yes!! Please do. I saw this and was really tempted to just add it to this PR but filing a separate issue is definitely a better idea 😁

evalml/tests/component_tests/test_decision_tree_classifier.py
angela97lin: Hmm, thinking about this some more: I can see the flip side of adding a test for categorical data if the DecisionTree estimators *should* be able to handle categorical data (to check that it's actually handling those columns correctly), but am more unclear about the value of adding a test to check whether or not categorical data breaks? @bchen1116 @freddyaboulton Maybe you guys can convince me otherwise haha

evalml/pipelines/components/estimators/classifiers/rf_classifier.py
```diff
@@ -27,7 +27,3 @@ def __init__(self, n_estimators=100, max_depth=6, n_jobs=-1, random_state=0, **k
         super().__init__(parameters=parameters,
                          component_obj=rf_classifier,
                          random_state=random_state)
-
```
angela97lin, line 30: Unrelated but I believe these can be cleaned up.

evalml/tests/component_tests/test_decision_tree_classifier.py
bchen1116: Yeah, I think it would be much more useful to add support for categorical data, and then add tests for it to make sure the behavior is as expected!

angela97lin: Mhm, but given that DecisionTree estimators don't support categorical data, I don't think the tests are necessary / adding support for categorical data within the estimator is necessary?

freddyaboulton: Yea I agree that adding that test would make more sense if these estimators could natively handle categorical features. The thought is that if users use these components outside the context of AutoML "they know what they're doing " and it's on them to encode columns that need encoding. In the context of AutoML, we take care of adding encoders for categorical features in `make_pipeline`. However, I think there is value in expanding the test coverage for `make_pipeline` - It is using an outdated list of estimators. It would be good to add coverage for Catboost, LightGBM (these should not need encoders because they handle categorical features natively), ExtraTrees, DecisionTrees, etc.

angela97lin: Yup, I agree, `make_pipeline` and its testing should be updated. The testing for ` make_pipeline` is pretty sloppy right now (I'm likely to be the person to blame for that 😂) and should probably dynamically get the available estimators so it doesn't get outdated so quickly. Will file something to update the method if necessary and add cleaner testing. Thanks guys!

bchen1116 (approved):
LGTM!

evalml/pipelines/components/estimators/regressors/decision_tree_regressor.py
freddyaboulton: Filed #1239!

angela97lin: Thanks!!

PR #1226, opened 2020-09-24
Internal target check to ensure no class missing from train/val

### LINKED ISSUE #760: Ensure no classes are missing from training/validation splits during data splitting
**Problem**
It's possible that random splitting could produce a dataset where a class is missing from the training and/or validation data, either during CV splitting or train-validation (TVH) splitting. This is more likely to occur for a minority class when there's a class imbalance.

**Proposed Solutions**
Detection: A basic protection is to check after data splitting and see if there's any missing classes on either side, and if so log a warning or error. This could be done easily.

Retry: If we detect this situation, we could also try the random data splitting again, up to a max number of times (perhaps 10).

Class-Balanced Sampling [i.i.d.](https://en.wikipedia.org/wiki/Independent_and_identically_distributed_random_variables) distribution: We could also write our own data splitting implementation which ensures that the distribution of labels across different splits is as similar as possible. We could do this by splitting each unique label's rows into two splits, then unifying the two sets across all the label types.

bchen1116:
fix #760 

Throws an error if the data split results in missing target values in either the train or validation sets


evalml/automl/automl_search.py
```diff
@@ -607,10 +607,18 @@ def _compute_cv_scores(self, pipeline, X, y):
         start = time.time()
         cv_data = []
         logger.info("\tStarting cross validation")
+        warnings.filterwarnings("ignore", lineno=665)
         for i, (train, test) in enumerate(self.data_split.split(X, y)):
             logger.debug(f"\t\tTraining and scoring on fold {i}")
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
+            if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
```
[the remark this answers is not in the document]
freddyaboulton: What do you mean by `TrainingValidationSplit` isn't meant to pass this check? Isn't that the case for `StratefiedKFold` too. I think it's good if this fails for `TrainingValidationSplit`.

evalml/automl/automl_search.py
```diff
@@ -607,10 +607,18 @@ def _compute_cv_scores(self, pipeline, X, y):
...
+                diff_train = set(y).difference(set(y_train))
+                diff_test = set(y).difference(set(y_test))
+                diff_string = f"Missing target values in the training set: {diff_train}. " if diff_train else ""
```
freddyaboulton, line 618: Might be a good idea to specify the missing values are introduced after the data is split. As a user, I'd be confused about this message if I new my labels had values for each class.

evalml/automl/automl_search.py
```diff
@@ -607,10 +607,18 @@ def _compute_cv_scores(self, pipeline, X, y):
         start = time.time()
         cv_data = []
         logger.info("\tStarting cross validation")
+        warnings.filterwarnings("ignore", lineno=665)
```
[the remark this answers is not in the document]
freddyaboulton: I'm skeptical of `lineno` because it will be impossible to maintain as the size of this file changes. Maybe we don't worry about suppressing warnings?

FWIW, I'm still seeing the warning when I run this code anyways.

image

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1080,3 +1080,54 @@ def test_max_batches_must_be_non_negative(max_batches):
...
+    y[0] = 0
+
+    automl = AutoMLSearch(problem_type='multiclass')
+    with pytest.raises(Exception, match="Missing target values"):
+        automl.search(X, y)
+    with pytest.raises(Exception, match="Missing target values"):
+        automl.search(X, y, data_checks="disabled")
+
+    y[1] = 2
+    with pytest.raises(Exception, match=r"(\{\d?\}.+\{\d?\})|(\{.+\,.+\})"):
```
freddyaboulton, line 1118: not-blocking: Add a comment explaining the regexp? Is there a reason why we can't specify the entire match string without a regexp?

freddyaboulton (approved):
@bchen1116 I have some minor suggestions to improve the implementation and a question regarding your comment about changing this for `TrainingValidationSplit` - otherwise looks great!

evalml/automl/automl_search.py
```diff
@@ -607,10 +607,18 @@ def _compute_cv_scores(self, pipeline, X, y):
...
         for i, (train, test) in enumerate(self.data_split.split(X, y)):
             logger.debug(f"\t\tTraining and scoring on fold {i}")
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
+            if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+                diff_train = set(y).difference(set(y_train))
```
freddyaboulton, line 616: I think we should use `np.setdiff1d` - it will be faster than `set.difference` since these are already series. Might need to change the `if diff_train` to get this to work though. 

image

evalml/tests/automl_tests/test_automl.py
bchen1116: Added a comment in the test. Since the data split has no random seed provided, the splits change each time, so this regex just checks that 2 elements are missing.

bchen1116:
@freddyaboulton about my comment for `TrainingValidationSplit`. This internal data check ensures that we need each class to occur in each train/validation split. This makes sense and is easy to enforce with `StratifiedKFold` since as long as we have > 3 occurances of each target class, we shouldn't run into this error. However, for `TrainingValidationSplit`, we don't give it a `stratify` argument, which means that we might not pass this check even with more instances of a target class. I'm just curious if we should include some default `stratify` arg, otherwise the user might run into this error when they use a really large dataset?

evalml/automl/automl_search.py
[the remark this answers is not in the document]
freddyaboulton: @bchen1116 and I discussed offline and we think it's ok if this error is raised with a `TrainingValidationSplit` since it means either the training or validation set has only one class.

PR #1228, opened 2020-09-25
Add skopt.space.Categorical documentation and tests for component hyperparameter range 

### LINKED ISSUE #352: Pipeline/component doc: we accept list or Categorical as hyperparameter defs
We currently use skopt's `Real` and `Integer` wrappers in hyperparameter ranges. Perhaps we should use [`Categorical`](https://scikit-optimize.github.io/space.m.html) too. I think our API can trivially support both--we just need to update the doc and add test cases, and perhaps change the default usage to `Categorical`. This is more of a style suggestion really.

bchen1116:
fix #352 

Update documentation to include option for either list or `skopt.space.Categorical` in the hyperparameter definition. Include unit tests.

### DIFF
--- a/docs/source/user_guide/components.ipynb
+++ b/docs/source/user_guide/components.ipynb
@@ -395,7 +395,7 @@
    "metadata": {},
    "source": [
     "### Hyperparameter Ranges for AutoML\n",
-    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter. Both lists and `skopt.space.Categorical` values are accepted. \n",
+    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter. Both lists and `skopt.space.Categorical` values are accepted for categorical spaces. \n",
     "\n",
     "AutoML will perform a search over the allowed ranges for each parameter to select models which produce optimal performance within those ranges. AutoML gets the allowed ranges for each component from the component's `hyperparameter_ranges` class attribute. Any component parameter you add an entry for in `hyperparameter_ranges` will be included in the AutoML search. If parameters are omitted, AutoML will use the default value in all pipelines. "
    ]
--- a/docs/source/user_guide/components.ipynb
+++ b/docs/source/user_guide/components.ipynb
@@ -395,7 +395,7 @@
    "metadata": {},
    "source": [
     "### Hyperparameter Ranges for AutoML\n",
-    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter.\n",
+    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter. Both lists and `skopt.space.Categorical` values are accepted. \n",
     "\n",
     "AutoML will perform a search over the allowed ranges for each parameter to select models which produce optimal performance within those ranges. AutoML gets the allowed ranges for each component from the component's `hyperparameter_ranges` class attribute. Any component parameter you add an entry for in `hyperparameter_ranges` will be included in the AutoML search. If parameters are omitted, AutoML will use the default value in all pipelines. "
    ]
@@ -407,6 +407,7 @@
    "outputs": [],
    "source": [
     "from sklearn.linear_model import LinearRegression as SKLinearRegression\n",
+    "from skopt.space import Categorical\n",
     "\n",
     "from evalml.model_family import ModelFamily\n",
     "from evalml.pipelines.components.estimators import Estimator\n",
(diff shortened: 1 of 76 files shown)


docs/source/user_guide/pipelines.ipynb
```diff
@@ -125,7 +125,9 @@
    "metadata": {},
    "source": [
     "## Override Component Hyperparameter Ranges\n",
-    "To specify custom hyperparameter ranges, set the `custom_hyperparameters` property to be a dictionary where each key-value pair consists of a parameter name and range. AutoML will use this dictionary to override the hyperparameter ranges collected from each component in the component graph."
+    "To specify custom hyperparameter ranges, set the `custom_hyperparameters` property to be a dictionary where each key-value pair consists of a parameter name and range. AutoML will use this dictionary to override the hyperparameter ranges collected from each component in the component graph.\n",
+    "\n",
+    "These hyperparameter ranges can be passed in as lists or as `skopt.space.Categorical` values."
```
freddyaboulton, line 130: I think we should clarify that only categorical spaces can be specified by either a list or `skopt.space.Categorical`? For other hyperparameters we use either `Real` or `Integer` and I don't think those can be specified by a list.

evalml/tests/component_tests/test_components.py
```diff
@@ -852,3 +853,57 @@ def test_component_equality_all_components(component_class):
     parameters = component.parameters
     equal_component = component_class(**parameters)
     assert component == equal_component
+
+
+def test_categorical_hyperparameters(X_y_binary):
```
freddyaboulton, line 858: I think we can condense these two tests into one test with `pytest.mark.parametrize`. I also think there is value in testing that a component that that uses `Categorical` can be used in automl search since that is where they hyperparameter ranges are used.

freddyaboulton (approved):
@bchen1116 This looks good! I think it would be good to write a test where `AutoMLSearch` searches over pipelines that use a component that defines some hyperparameter with `Categorical` before merging. (I think we could subclass `LogisticRegressionClassifier` and change the range for `penalty` for example)

docs/source/user_guide/components.ipynb
```diff
@@ -395,7 +395,7 @@
    "metadata": {},
    "source": [
     "### Hyperparameter Ranges for AutoML\n",
-    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter.\n",
+    "`hyperparameter_ranges` is a dictionary mapping the parameter name (str) to an allowed range ([SkOpt Space](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space)) for that parameter. Both lists and `skopt.space.Categorical` values are accepted. \n",
```
angela97lin, line 398: I think similar to @freddyaboulton's comment, we should clarify that only categorical spaces can be specified by either a list or skopt.space.Categorical?

angela97lin (approved):
Looks good! I agree with @freddyaboulton's comments about testing / clarification but otherwise 👍

evalml/tests/component_tests/test_components.py
```diff
@@ -852,3 +853,36 @@ def test_component_equality_all_components(component_class):
     parameters = component.parameters
     equal_component = component_class(**parameters)
     assert component == equal_component
+
+
+@pytest.mark.parametrize("categorical", [{
```
freddyaboulton, line 858: Nice!

freddyaboulton (approved):
@bchen1116 Thanks for making these changes!

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -606,3 +611,18 @@ def test_automl_allowed_pipelines_search(mock_fit, mock_score, dummy_binary_pipe
...
+    class CustomPipeline(MulticlassClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Standard Scaler', 'Logistic Regression Classifier']
+        custom_hyperparameters = {
+            'Simple Imputer': {
+                'impute_strategy': Categorical(['mean', 'most_frequent'])
+            }
+        }
+
+    automl = AutoMLSearch(problem_type="multiclass", allowed_pipelines=[CustomPipeline])
+    automl.search(X, y)
```
dsherry, line 628: @bchen1116 just a tip, for tests which don't check scoring, we usually mock fit/score so that we save time in the unit test. See some of the other tests in this file
```python
def test_categorical_hyperparam(mock_fit, mock_score, X_y_multi):
```

evalml/tests/component_tests/test_components.py
```diff
@@ -852,3 +853,36 @@ def test_component_equality_all_components(component_class):
...
+
+        def __init__(self, agg_type, category="green"):
+            parameters = {"type": agg_type, "categories": category}
+            est = MockEstimator()
+            super().__init__(parameters=parameters,
+                             component_obj=est,
+                             random_state=0)
+
+    assert MockComponent(agg_type="mean").fit(X, y)
+    assert MockComponent(agg_type="moat", category="blue").fit(X, y)
```
dsherry, line 888: @bchen1116 why call `fit` here? What is the goal of this test?

dsherry (commented):
@bchen1116 thanks, great! I left a couple minor test suggestions.

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: Oh I see. So patching fit/score without calling `mock_fit` or `mock_score` still makes it run faster?

evalml/tests/component_tests/test_components.py
bchen1116: The goal was to make sure the component works, but since the component uses an estimator, I called `fit` to satisfy codecov

PR #1231, opened 2020-09-28
Replace `._transform` with public method to evaluate all the preprocessing steps of a pipeline

### LINKED ISSUE #1225: Public method to evaluate all the preprocessing steps of a pipeline
We currently define `PipelineBase._transform`--let's make that public so its easy for users to get the data which is passed to the estimator. And of course we should also add good test coverage to ensure the column names are set correctly, all the features are computed as expected.

An initial idea for a name: `compute_estimator_features`

Sidenote, once we add this, it'll be another thing we should think about as part of the DAG project #273

angela97lin:
Closes #1225

API reference here: https://evalml.alteryx.com/en/1225_public_pipeline_eval_method/generated/methods/evalml.pipelines.PipelineBase.compute_estimator_features.html#evalml.pipelines.PipelineBase.compute_estimator_features

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -14,6 +14,7 @@ Release Notes
     * Changes
         * Added `allow_writing_files` as a named argument to CatBoost estimators. :pr:`1202`
         * Added `solver` and `multi_class` as named arguments to LogisticRegressionClassifier :pr:`1202`
+        * Replaced pipeline's `.transform` method with public method to evaluate all the preprocessing steps of a pipeline :pr:`1231`
     * Documentation Changes
         * Included description of how to access the component instances and features for pipeline user guide :pr:`1163`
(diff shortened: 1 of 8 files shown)


bchen1116 (approved):
LGTM

docs/source/release_notes.rst
```diff
@@ -17,6 +17,7 @@ Release Notes
     * Changes
         * Added `allow_writing_files` as a named argument to CatBoost estimators. :pr:`1202`
         * Added `solver` and `multi_class` as named arguments to LogisticRegressionClassifier :pr:`1202`
+        * Replaced pipeline's `.transform` method with public method to evaluate all the preprocessing steps of a pipeline :pr:`1231`
```
freddyaboulton, line 20: Maybe we should add a line to the breaking changes section as well? I agree that changing a private api doesn't normally constitute a breaking change but some users may be relying on `_transform` from old releases?

freddyaboulton (approved):
@angela97lin Thanks for doing this! The api ref looks great.

PR #1232, opened 2020-09-28
Wrap call to scikit-learn's partial dependence method in a try/finally block

angela97lin:
From @dsherry's comment in #1150, wrap call to scikit-learn's partial dependence method in a try/finally block so that we still delete the set attributes even if call to scikit-learn's partial dependence method fails or errors out.

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -437,11 +437,12 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
     elif isinstance(pipeline, evalml.pipelines.RegressionPipeline):
         pipeline._estimator_type = "regressor"
     pipeline.feature_importances_ = pipeline.feature_importance
-    avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
-
-    # Delete scikit-learn attributes that were temporarily set
-    del pipeline._estimator_type
-    del pipeline.feature_importances_
+    try:
+        avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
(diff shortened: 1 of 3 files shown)


jeremyliweishih (approved):
👍

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -689,6 +689,20 @@ def test_partial_dependence_problem_types(problem_type, X_y_binary, X_y_multi, X
         pipeline.feature_importances_
 
 
+@patch('evalml.model_understanding.graphs.sk_partial_dependence')
+def test_partial_dependence_error_still_deletes_attributes(mock_part_dep, X_y_binary, logistic_regression_binary_pipeline_class):
```
bchen1116, line 693: Nice test!

bchen1116 (approved):
LGTM!

evalml/model_understanding/graphs.py
```diff
@@ -437,11 +437,12 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
     elif isinstance(pipeline, evalml.pipelines.RegressionPipeline):
         pipeline._estimator_type = "regressor"
     pipeline.feature_importances_ = pipeline.feature_importance
-    avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
-
-    # Delete scikit-learn attributes that were temporarily set
-    del pipeline._estimator_type
-    del pipeline.feature_importances_
+    try:
```
freddyaboulton, line 440: If we have to repeat this pattern in the future, it'd be cool to use a context manager!

freddyaboulton (approved):
@angela97lin I think this looks great!

PR #1244, opened 2020-09-29
Compute percent-better-than-baseline for all objectives

### LINKED ISSUE #1184: Add percent_better_than_baseline on a per metric basis
## Problem
- As a user of evaml, I want to understand how much better my pipeline models did when compared to a naive, baseline approach. The `percent_better_than_baseline` value in `automl.results` allows me to do this. 
- However, this the value outputted in `percent_better_than_baseline` is only for the optimization metric. It does not tell me how the baseline pipeline did for a non-optimization metric. 
  - For example, let's say I was optimizating for **F1**, but wanted to know how much better the pipelines did for **Precision** compared to the baseline model. 

## Proposal
- Change the automl results dictionary for each pipeline, where its `percent_better_than_baseline` from a float point number to a list of dictionaries. 
  - Each dictionary would be 1 objective, and the percent change compared to baseline. 
```json
[
        {
            "objective": "Log Loss Binary",
            "baselinePercentChange": 98.99224806201549
        },
        {
            "objective": "Accuracy Binary",
            "baselinePercentChange": 53.96825396825397
        },
        {
            "objective": "Balanced Accuracy Binary",
            "baselinePercentChange": 92.0
        },
        {
            "objective": "F1",
            "baselinePercentChange": 100.0
        },
    ]
```

freddyaboulton:
### Pull Request Description
Fix #1184

-----

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,7 +1,7 @@
 import copy
 import time
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 import cloudpickle
 import numpy as np
@@ -31,6 +31,7 @@ from evalml.exceptions import (
     PipelineScoreError
 )
 from evalml.objectives import (
+    get_all_objective_names,
     get_core_objectives,
     get_non_core_objectives,
     get_objective
@@ -594,13 +595,12 @@ class AutoMLSearch:
 
     @staticmethod
     def _get_mean_cv_scores_for_all_objectives(cv_data):
-        scores = {}
+        scores = defaultdict(int)
+        objective_names = set([name.lower() for name in get_all_objective_names()])
         n_folds = len(cv_data)
         for fold_data in cv_data:
             for field, value in fold_data['all_objective_scores'].items():
-                if field not in {'# Training', '# Testing'}:
-                    if field not in scores:
-                        scores[field] = 0
+                if field.lower() in objective_names:
                     scores[field] += value
         return {objective_name: score / n_folds for objective_name, score in scores.items()}
 
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -602,7 +602,7 @@ class AutoMLSearch:
             for field, value in fold_data['all_objective_scores'].items():
                 if field.lower() in objective_names:
                     scores[field] += value
-        return {objective_name: score / n_folds for objective_name, score in scores.items()}
+        return {objective_name: float(score) / n_folds for objective_name, score in scores.items()}
 
     def _compute_cv_scores(self, pipeline, X, y):
         start = time.time()
@@ -675,7 +675,6 @@ class AutoMLSearch:
             if isinstance(cv_pipeline, BinaryClassificationPipeline) and cv_pipeline.threshold is not None:
                 evaluation_entry['binary_classification_threshold'] = cv_pipeline.threshold
             cv_data.append(evaluation_entry)
-
         training_time = time.time() - start
         cv_scores = pd.Series([fold['score'] for fold in cv_data])
         cv_score_mean = cv_scores.mean()
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -222,7 +222,7 @@ class AutoMLSearch:
         self.allowed_model_families = allowed_model_families
         self._automl_algorithm = None
         self._start = None
-        self._baseline_cv_scores = None
+        self._baseline_cv_scores = {}
 
         if _max_batches is not None and _max_batches <= 0:
             raise ValueError(f"Parameter max batches must be None or non-negative. Received {_max_batches}.")
@@ -688,8 +688,10 @@ class AutoMLSearch:
         mean_cv_all_objectives = self._get_mean_cv_scores_for_all_objectives(cv_data)
         for obj_name in mean_cv_all_objectives:
             objective_class = get_objective(obj_name)
+            # In the event add_to_rankings is called before search _baseline_cv_scores will be empty so we will return
+            # nan for the base score.
             percent_better = objective_class.calculate_percent_difference(mean_cv_all_objectives[obj_name],
-                                                                          self._baseline_cv_scores[obj_name])
+                                                                          self._baseline_cv_scores.get(obj_name, np.nan))
             percent_better_than_baseline[obj_name] = percent_better
 
         # calculate high_variance_cv
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -221,7 +221,7 @@ class AutoMLSearch:
         self.allowed_model_families = allowed_model_families
         self._automl_algorithm = None
         self._start = None
-        self._baseline_cv_score = None
+        self._baseline_cv_scores = None
 
         if _max_batches is not None and _max_batches <= 0:
             raise ValueError(f"Parameter max batches must be None or non-negative. Received {_max_batches}.")
@@ -579,7 +579,7 @@ class AutoMLSearch:
                                 self._start)
 
                 baseline_results = self._compute_cv_scores(baseline, X, y)
-                self._baseline_cv_score = baseline_results["cv_score_mean"]
+                self._baseline_cv_scores = self._get_mean_cv_scores_for_all_objectives(baseline_results["cv_data"])
                 self._add_result(trained_pipeline=baseline,
                                  parameters=baseline.parameters,
                                  training_time=baseline_results['training_time'],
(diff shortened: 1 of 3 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -511,11 +512,24 @@ def test_automl_serialization(X_y_binary, tmpdir):
...
     automl.save(path)
     loaded_automl = automl.load(path)
+
     for i in range(num_max_iterations):
         assert automl.get_pipeline(i).__class__ == loaded_automl.get_pipeline(i).__class__
         assert automl.get_pipeline(i).parameters == loaded_automl.get_pipeline(i).parameters
-        assert automl.results == loaded_automl.results
-        pd.testing.assert_frame_equal(automl.rankings, loaded_automl.rankings)
+
+        for id_, pipeline_results in automl.results['pipeline_results'].items():
```
freddyaboulton, line 520: Had to modify this check because `np.nan != np.nan`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1063,6 +1079,88 @@ class Pipeline2(DummyPipeline):
...
+        pipeline_results = automl.results['pipeline_results'][1]
+        assert pipeline_results["percent_better_than_baseline_all_objectives"] == answer
+        assert pipeline_results['percent_better_than_baseline'] == pipeline_results["percent_better_than_baseline_all_objectives"][automl.objective.name]
+
+
+@pytest.mark.parametrize("fold_scores", [[2, 4, 6], [np.nan, 4, 6]])
+@patch("evalml.pipelines.ModeBaselineBinaryPipeline.score", return_value={'Log Loss Binary': 1, 'F1': 1})
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_percent_better_than_baseline_scores_different_folds(mock_fit,
```
freddyaboulton, line 1132: I noticed that all our other unit tests assume the same score is returned on each fold so I decided to add one where the score is different for each fold.

evalml/automl/automl_search.py
```diff
@@ -665,7 +677,15 @@ def _compute_cv_scores(self, pipeline, X, y):
...
         cv_score = cv_scores.mean()
-        percent_better = self.objective.calculate_percent_difference(cv_score, self._baseline_cv_score)
+
+        percent_better_than_baseline = {}
+        mean_cv_all_objectives = self._get_mean_cv_scores_for_all_objectives(cv_data)
+        for obj_name in mean_cv_all_objectives:
+            objective_class = get_objective(obj_name)
+            percent_better = objective_class.calculate_percent_difference(mean_cv_all_objectives[obj_name],
+                                                                          self._baseline_cv_scores[obj_name])
+            percent_better_than_baseline[obj_name] = percent_better
```
freddyaboulton, line 687: @gsheni This is what it would look like in `automl.results` :

```python
'percent_better_than_baseline_all_objectives': {'Log Loss Binary': 98.9276200826952,
    'MCC Binary': nan,
    'AUC': 98.17222663713376,
    'Precision': nan,
    'F1': nan,
    'Balanced Accuracy Binary': 91.62898962401862,
    'Accuracy Binary': 53.49893478518168}
```

This is slightly different from what you posted in the original issue but I think it's more human-readable. Happy to change though.

gsheni: - The reasoning for a list of dictionaries is that it would allow for adding more information in the future.
- For example, EvalML could calculate a boolean that looks at the baseline and pipeline score to determine if its actually better (even if `baselinePercentChange` is NaN).
- However, this is a major sticking point, and the end-user could get the scores, make the comparison, and calculate the boolean. 
 
```json
'percent_better_than_baseline_all_objectives': [
   {'objective' : 'Log Loss Binary',
    'baselinePercentChange': 98.9276200826952,
    'isBetterThanBaseline': True},
   {'objective' : 'MCC Binary',
    'baselinePercentChange': nan,
    'isBetterThanBaseline': True}...}
```

evalml/automl/automl_search.py
```diff
@@ -585,6 +586,17 @@ def _add_baseline_pipelines(self, X, y):
...
+    @staticmethod
+    def _get_mean_cv_scores_for_all_objectives(cv_data):
+        scores = defaultdict(int)
+        objective_names = set([name.lower() for name in get_all_objective_names()])
+        n_folds = len(cv_data)
+        for fold_data in cv_data:
+            for field, value in fold_data['all_objective_scores'].items():
+                if field.lower() in objective_names:
+                    scores[field] += value
+        return {objective_name: score / n_folds for objective_name, score in scores.items()}
```
dsherry, line 598: Cool. If `score` is `int` this won't carry the remainder, so for safety please do `float(score) / n_folds`

evalml/automl/automl_search.py
dsherry: Looks good to me!

Yep. Users can certainly easily compute that boolean if they want to.

evalml/automl/automl_search.py
```diff
@@ -686,7 +706,8 @@ def _add_result(self, trained_pipeline, parameters, training_time, cv_data, cv_s
             "high_variance_cv": high_variance_cv,
             "training_time": training_time,
             "cv_data": cv_data,
-            "percent_better_than_baseline": percent_better,
+            "percent_better_than_baseline_all_objectives": percent_better_than_baseline,
+            "percent_better_than_baseline": percent_better_than_baseline[self.objective.name],
```
dsherry, line 710: @freddyaboulton so in `cv_data`, do we have separate fields for the primary objective vs the additional ones? Let's just make sure we're consistent across our code. Relatedly, right now, does `percent_better_than_baseline_all_objectives` also include the primary objective?

I think this would be cleanest if `percent_better_than_baseline` were simply a dict here in `results`, and then in `full_rankings` when we make the rankings leaderboard, we extract that to make the column. But if that's not what we do in `cv_data`, let's punt on that. What do you think?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -511,11 +512,24 @@ def test_automl_serialization(X_y_binary, tmpdir):
...
     for i in range(num_max_iterations):
         assert automl.get_pipeline(i).__class__ == loaded_automl.get_pipeline(i).__class__
         assert automl.get_pipeline(i).parameters == loaded_automl.get_pipeline(i).parameters
-        assert automl.results == loaded_automl.results
-        pd.testing.assert_frame_equal(automl.rankings, loaded_automl.rankings)
+
+        for id_, pipeline_results in automl.results['pipeline_results'].items():
+            loaded_ = loaded_automl.results['pipeline_results'][id_]
+            for name in pipeline_results:
+                # Use np to check percent_better_than_baseline because of (possible) nans
```
dsherry, line 523: Thanks for the comment, helpful here!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1063,6 +1079,88 @@ class Pipeline2(DummyPipeline):
...
+        problem_type = problem_type_enum
+
+        def fit(self, *args, **kwargs):
+            """Mocking fit"""
+
+    core_objectives = get_core_objectives(problem_type)
+    mock_scores = {obj.name: i for i, obj in enumerate(core_objectives)}
+    mock_baseline_scores = {obj.name: i + 1 for i, obj in enumerate(core_objectives)}
+    answer = {obj.name: obj.calculate_percent_difference(mock_scores[obj.name],
+                                                         mock_baseline_scores[obj.name]) for obj in core_objectives}
```
dsherry, line 1112: Nice

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1063,6 +1079,88 @@ class Pipeline2(DummyPipeline):
...
+
+    mock_score_1 = MagicMock(return_value=mock_scores)
+    DummyPipeline.score = mock_score_1
+
+    automl = AutoMLSearch(problem_type=problem_type, max_iterations=2,
+                          allowed_pipelines=[DummyPipeline], objective="auto")
+
+    with patch(baseline_pipeline_class + ".score", return_value=mock_baseline_scores):
+        automl.search(X, y, data_checks=None)
```
dsherry, line 1121: Should we mock pipeline fit and score here too to reduce runtime?

evalml/tests/automl_tests/test_automl.py
dsherry: Great!

dsherry (approved):
@freddyaboulton looks awesome!!

I left a discussion about whether we still keep `percent_better_than_baseline` for the primary objective separate in `results`. Let's resolve that and then merge. I didn't have a great conclusion there--thinking about it now, let's leave it the way you have it now, just wanted to make you aware of that topic.

evalml/automl/automl_search.py
freddyaboulton: @dsherry I just checked and `cv_data` has a field for the primary objective (`score`) and for all objectives (`all_objective_scores` - this includes the primary objective as well). I think that means that `percent_better_than_baseline` and `percent_better_than_baseline_all_objectives` follow the same pattern we use for scores. 

Maybe we keep it like this in this PR and I file an issue for consolidating scores and percent better into a single field in the results?

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Good suggestion! I'll mock fit for the baseline pipelines since that's the only mock missing

evalml/automl/automl_search.py
dsherry: That sounds good!

PR #1247, opened 2020-09-30
Build conda on merge to main

ryan-s:
### Pull Request Description
Add in a conda package build as part of the ci when a PR is merged into main.  This PR adds in a small python script to change the recipe to use the local build (conda_config.py). Once we have that, we clone the feedstock and start to replicate what the python and shell scripts were doing do to issues on circleCI.  The end result is a build that passes or fails and produces no artifacts.

-----

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   windows: circleci/windows@2.2.0
+  shellcheck: circleci/shellcheck@2.0.0
 
 executors:
   python:
@@ -47,8 +48,42 @@ commands:
             # "!" negates return code. exit nonzero if any of these deps are found
             ! pip freeze | grep -E "xgboost|catboost|lightgbm|plotly|ipywidgets"
             exit $?
+  build_pkg:
+    steps:
+      - run: |
+            git clone https://github.com/conda-forge/evalml-core-feedstock
+            source test_python/bin/activate
+            mkdir evalml-core-feedstock/evalml
+            cp -r `ls -A | grep -v "evalml-core-feedstock"` ./evalml-core-feedstock/evalml/
+            python .circleci/conda_config.py "$(python setup.py --version)"
+            cd evalml-core-feedstock
+            echo "$DOCKER_HUB_PASS" | docker login -u psalter --password-stdin
+            export DOCKER_CONTAINERID="$(docker run -td condaforge/linux-anvil-comp7)"
+            echo "Created container ${DOCKER_CONTAINERID}"
+            chmod -R 777 ./
+            docker cp . ${DOCKER_CONTAINERID}:/home/conda/feedstock_root/
+            docker cp ./recipe/. ${DOCKER_CONTAINERID}:/home/conda/recipe_root/
+            echo "COMMITING UPDATED IMAGE"
+            docker commit ${DOCKER_CONTAINERID} psalter/build:latest
+            docker stop ${DOCKER_CONTAINERID}
+            export CONFIG=linux_64_
+            export UPLOAD_PACKAGES=False
+            export HOST_USER_ID=$(id -u)
+            export FEEDSTOCK_NAME=evalml-core-feedstock
+            docker run -t -e CONFIG -e HOST_USER_ID -e UPLOAD_PACKAGES -e GIT_BRANCH -e UPLOAD_ON_BRANCH -e CI -e FEEDSTOCK_NAME -e CPU_COUNT -e BINSTAR_TOKEN -e FEEDSTOCK_TOKEN -e STAGING_BINSTAR_TOKEN psalter/build:latest bash /home/conda/feedstock_root/.scripts/build_steps.sh
 
 jobs:
+  build_conda_pkg:
+    working_directory: ~/evalml/
+    executor:
+      name: python
+      python_version: "3.6"
+    steps:
+      - checkout
+      - install_dependencies_test
+      - setup_remote_docker:
+          version: 19.03.12
+      - build_pkg
   lint_test:
       parameters:
         python_version:
@@ -298,3 +333,12 @@ workflows:
                 core_dependencies: [false, true]
                 codecov: [true]
             name: "linux python << matrix.python_version >> unit tests, core dependencies << matrix.core_dependencies >>, codecov << matrix.codecov >>"
+  build_conda_pkg:
+    jobs:
+      - shellcheck/check
+      - build_conda_pkg:
+          requires:
+            - shellcheck/check
+          filters:
+            branches:
+              only: main
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,3 +7,4 @@ pydata-sphinx-theme==0.3.1
 Sphinx==2.0.1; python_version>'3.4'
 nbconvert==5.5.0
 nbsphinx==0.4.2
+
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -7,3 +7,4 @@ codecov==2.1.0
 category_encoders>=2.0.0
 featuretools
 nlp_primitives>=1.0.0
+PyYAML==5.3.1
--- a/.circleci/conda_config.py
+++ b/.circleci/conda_config.py
@@ -1,0 +1,40 @@
+import yaml, argparse, os, pathlib
+
+
+class CondaDumper(yaml.Dumper):
+
+    def increase_indent(self, flow=False, indentless=False):
+        return super(CondaDumper, self).increase_indent(flow, False)
+
+
+def write_conda_recipe(version):
+    """
+    Writes the EvalML recipe to build a conda package based on the version
+    Args:
+        version: The version of EvalML we are building with this feedstock
+
+    Returns:
+        None: Side effect of overwriting the existing meta.yaml in the feedstock
+    """
+
+    recipe_file_path = os.path.join(os.path.join(os.path.join(os.getcwd(), 'evalml-core-feedstock'), 'recipe'), 'meta'
+                                                                                                                '.yaml')
+    with open(recipe_file_path, 'rb') as config_file:
+        # Toss out the first line that declares the version since its not supported YAML syntax
+        next(config_file)
+        config = yaml.safe_load(config_file)
+        # Path to the evalml repository on the docker container.  Since we are doing a local build this is
+        # the target we are copying this directory to.
+        recipe_path = str(pathlib.Path('..', 'feedstock_root', 'evalml'))
+        config['source'] = {'path': recipe_path}
+        config['package']['version'] = version
+
+    with open(recipe_file_path, 'w') as recipe:
+        yaml.dump(config, recipe, default_flow_style=False, sort_keys=False, Dumper=CondaDumper)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Configure conda for local build. Run from the feedstock root")
+    parser.add_argument('version', help='The version of EvalML being built')
+    args = parser.parse_args()
+    write_conda_recipe(args.version)
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -14,6 +14,7 @@ Release Notes
         * Added DecisionTree estimators to API Reference :pr:`1246`
         * Changed class inheritance display to flow vertically :pr:`1248`
     * Testing Changes
+        * Added a test to check conda build after merge to main :pr:`1247`
 
 
 **v0.14.1 Sep. 29, 2020**


.circleci/config.yml
```diff
@@ -47,8 +48,42 @@ commands:
...
+  build_pkg:
+    steps:
+      - run: |
+            git clone https://github.com/conda-forge/evalml-core-feedstock
+            source test_python/bin/activate
+            mkdir evalml-core-feedstock/evalml
+            cp -r `ls -A | grep -v "evalml-core-feedstock"` ./evalml-core-feedstock/evalml/
+            python conda_config.py "$(python setup.py --version)"
+            cd evalml-core-feedstock
+            echo "$DOCKER_HUB_PASS" | docker login -u psalter --password-stdin
```
freddyaboulton, line 60: We have manually check out this container and run `docker run` ourselves as opposed to running `run_docker_build.sh` in the feedstock repo because circle-ci doesn't let you mount volumes to the docker image. I think this is ok because we're not straying far from the contents of that script. @ryan-s and I discussed ways of tightening this up in the future too!

.circleci/config.yml
```diff
@@ -298,3 +333,12 @@ workflows:
...
             name: "linux python << matrix.python_version >> unit tests, core dependencies << matrix.core_dependencies >>, codecov << matrix.codecov >>"
+  build_conda_pkg:
+    jobs:
+      - shellcheck/check
+      - build_conda_pkg:
+          requires:
+            - shellcheck/check
+          filters:
+            branches:
+              only: main
```
freddyaboulton, line 344: This means that this will run after a PR gets merged to `main`.

We created this check to catch breaking changes to our conda recipe (like new dependencies or dependency updates) before release day and having it run after a merge accomplishes that without having to slow down development or consume too many ci resources.

freddyaboulton (approved):
@ryan-s this is awesome! I left some comments to explain to the team what's going on (feel free to correct me where I'm mistaken hehe).

We ran this check on circle ci with [main](https://app.circleci.com/pipelines/github/alteryx/evalml/6259/workflows/fe07b6f1-813d-4ccd-b70b-7ba1ffe3b996/jobs/67031) and a [purposely broken branch](https://app.circleci.com/pipelines/github/alteryx/evalml/6260/workflows/0b49c30d-a446-4329-8513-9bfe4b9de95e/jobs/67046) and the check is working as expected!

dev-requirements.txt
```diff
@@ -7,3 +7,4 @@ pydata-sphinx-theme==0.3.1
 Sphinx==2.0.1; python_version>'3.4'
 nbconvert==5.5.0
 nbsphinx==0.4.2
+PyYAML==5.3.1
```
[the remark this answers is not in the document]
ryan-s: Moved

.circleci/config.yml
angela97lin: very cool!

dsherry: SGTM!

.circleci/config.yml
dsherry: @ryan-s @freddyaboulton is this env var `DOCKER_HUB_PASS` set by circleci? Did you have to enter it?

.circleci/config.yml
```diff
@@ -47,8 +48,42 @@ commands:
...
+            docker cp . ${DOCKER_CONTAINERID}:/home/conda/feedstock_root/
+            docker cp ./recipe/. ${DOCKER_CONTAINERID}:/home/conda/recipe_root/
+            echo "COMMITING UPDATED IMAGE"
+            docker commit ${DOCKER_CONTAINERID} psalter/build:latest
+            docker stop ${DOCKER_CONTAINERID}
+            export CONFIG=linux_64_
+            export UPLOAD_PACKAGES=False
+            export HOST_USER_ID=$(id -u)
+            export FEEDSTOCK_NAME=evalml-core-feedstock
+            docker run -t -e CONFIG -e HOST_USER_ID -e UPLOAD_PACKAGES -e GIT_BRANCH -e UPLOAD_ON_BRANCH -e CI -e FEEDSTOCK_NAME -e CPU_COUNT -e BINSTAR_TOKEN -e FEEDSTOCK_TOKEN -e STAGING_BINSTAR_TOKEN psalter/build:latest bash /home/conda/feedstock_root/.scripts/build_steps.sh
```
dsherry, line 73: If I'm following right, this command will have a non-zero return code if there's an error? Meaning the CI job will get flagged as failed?

Will the error output appear in the CI job, or do we have to do something else to get the error output? It would be amazing to have the output appear in the CI job, because otherwise its gonna be hard to debug.

.circleci/config.yml
```diff
@@ -47,8 +48,42 @@ commands:
...
+  build_conda_pkg:
+    working_directory: ~/evalml/
+    executor:
+      name: python
+      python_version: "3.6"
+    steps:
+      - checkout
+      - install_dependencies_test
+      - setup_remote_docker:
+          version: 19.03.12
```
dsherry, line 85: What is this version of? Is this the version of docker itself? And is `setup_remote_docker` a circleci predefined macro, or something else?

.circleci/config.yml
```diff
@@ -47,8 +48,42 @@ commands:
...
+            export HOST_USER_ID=$(id -u)
+            export FEEDSTOCK_NAME=evalml-core-feedstock
+            docker run -t -e CONFIG -e HOST_USER_ID -e UPLOAD_PACKAGES -e GIT_BRANCH -e UPLOAD_ON_BRANCH -e CI -e FEEDSTOCK_NAME -e CPU_COUNT -e BINSTAR_TOKEN -e FEEDSTOCK_TOKEN -e STAGING_BINSTAR_TOKEN psalter/build:latest bash /home/conda/feedstock_root/.scripts/build_steps.sh
 
 jobs:
+  build_conda_pkg:
+    working_directory: ~/evalml/
+    executor:
+      name: python
+      python_version: "3.6"
```
dsherry, line 80: @ryan-s @freddyaboulton why python 3.6? Should we do this for all our supported python versions, 3.6, 3.7 and 3.8? Or just the latest version, 3.8?

.circleci/config.yml
```diff
@@ -298,3 +333,12 @@ workflows:
                 core_dependencies: [false, true]
                 codecov: [true]
             name: "linux python << matrix.python_version >> unit tests, core dependencies << matrix.core_dependencies >>, codecov << matrix.codecov >>"
+  build_conda_pkg:
+    jobs:
+      - shellcheck/check
+      - build_conda_pkg:
+          requires:
+            - shellcheck/check
```
dsherry, line 341: Why is this listed as required when it just ran beforehand

conda_config.py
```diff
@@ -0,0 +1,38 @@
...
+
+    with open(recipe_file_path, 'w') as recipe:
+        yaml.dump(config, recipe, default_flow_style=False, sort_keys=False, Dumper=CondaDumper)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Configure conda for local build. Run from the feedstock root")
+    parser.add_argument('version', help='The version of EvalML being built')
+    args = parser.parse_args()
+    write_conda_recipe(args.version)
```
dsherry, line 38: @ryan-s this is super cool!!

Please move this file to a different location instead of the top-level directory. Should it go in `.circleci/`? That would be my first inclination.

conda_config.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def write_conda_recipe(version):
+    """
+    Writes the EvalML recipe to build a conda package based on the version
+    Args:
+        version: The version of EvalML we are building with this feedstock
+
+    Returns:
+        None: Side effect of overwriting the existing meta.yaml in the feedstock
```
dsherry, line 17: Amazing, thank you for documenting.

conda_config.py
```diff
@@ -0,0 +1,38 @@
+import yaml, argparse, os
+
+
+class CondaDumper(yaml.Dumper):
+
+    def increase_indent(self, flow=False, indentless=False):
+        return super(CondaDumper, self).increase_indent(flow, False)
```
dsherry, line 7: Why is this necessary? Is this just a style thing, or is it a functional thing?

test-requirements.txt
```diff
@@ -7,3 +7,4 @@ codecov==2.1.0
 category_encoders>=2.0.0
 featuretools
 nlp_primitives>=1.0.0
+PyYAML==5.3.1
```
dsherry, line 10: 👍

conda_config.py
```diff
@@ -0,0 +1,38 @@
...
+        None: Side effect of overwriting the existing meta.yaml in the feedstock
+    """
+
+    recipe_file_path = os.path.join(os.path.join(os.path.join(os.getcwd(), 'evalml-core-feedstock'), 'recipe'), 'meta'
+                                                                                                                '.yaml')
+    with open(recipe_file_path, 'rb') as config_file:
+        # Toss out the first line that declares the version since its not supported YAML syntax
+        next(config_file)
+        config = yaml.safe_load(config_file)
+        recipe_path = f'../feedstock_root/evalml'
```
dsherry, line 26: @ryan-s could you please [use `pathlib.Path`](https://docs.python.org/3/library/pathlib.html) here instead? For hypothetical windows support haha

```python
recipe_path = str(pathlib.Path('..', 'feedstock_root', 'evalml'))
```

Also what is this path? Like, this is intended to be the path on the docker container? A comment explaining this would be super helpful to future readers.

dsherry (approved):
@ryan-s this rocks!!! I'm so excited you're adding this! It's gonna make our release process a lot simpler and less likely to break conda. Woohoo!!

So, I take it we'll still need to do the actual conda release manually, right?

@freddyaboulton once this is in, could you update release.md to reflect this?

@ryan-s I left a bunch of questions, and two comments to address before merge:
* Please move that new file `conda_config.py` to somewhere other than the top-level directory -- I think it should go in `.circleci/`. Hmm, on reflection, since its python code, perhaps it should go in a new folder `evalml/tests/circleci/`. Your call there.
* I left a comment about using `Pathlib`--let's resolve that, either by making the change or by you telling me its unnecessary lol

🚢

.circleci/config.yml
freddyaboulton: The error appears in the CI job!

.circleci/config.yml
freddyaboulton: I think this sets the version of python in the circleci worker. The docker run command will actually build the package in python 3.7.8 - this is the same as what happens in the CI of our feedstock repo. Unfortunately, there isn't an out-of-the-box way to build with different python versions but I will add this to the future work section of our design doc for this issue.

freddyaboulton:
> @ryan-s this rocks!!! I'm so excited you're adding this! It's gonna make our release process a lot simpler and less likely to break conda. Woohoo!!
> 
> So, I take it we'll still need to do the actual conda release manually, right?
> 
> @freddyaboulton once this is in, could you update release.md to reflect this?
> 
> @ryan-s I left a bunch of questions, and two comments to address before merge:
> 
> * Please move that new file `conda_config.py` to somewhere other than the top-level directory -- I think it should go in `.circleci/`. Hmm, on reflection, since its python code, perhaps it should go in a new folder `evalml/tests/circleci/`. Your call there.
> * I left a comment about using `Pathlib`--let's resolve that, either by making the change or by you telling me its unnecessary lol
> 
> 🚢

Yes, we still need to manually edit the recipe file in the feedstock if there are dependency changes in a release but now we can skip the "Test conda version before releasing on PyPI" section of the release process! I'll update once this is in.

.circleci/config.yml
dsherry: Ah got it. So we'll only build and run our tests on 3.7.8 for now. No problem, this is a great starting point. Thanks!

.circleci/config.yml
ryan-s: @dsherry I entered it in circleci. Not sure if we have a vault somewhere to share the credentials?

.circleci/config.yml
ryan-s: Correct, this is the latest version of docker that circleci supports and it is predefined by them.

.circleci/config.yml
ryan-s: Correct, I just needed something to install pyyaml into so that we can update the recipe.  

I think the v2 of this would be to create our own container(s) based off of condaforge/linux-anvil-comp7 where we can move the existing functionality and any new improvements into.

conda_config.py
ryan-s: Unfortunately it is functional.  The existing recipe is actually a Jinja2 template masquerading as a yaml file.  Whatever they are using to parse that template requires that we have the increased indents even though the YAML spec does not.

ryan-s:
> @ryan-s this rocks!!! I'm so excited you're adding this! It's gonna make our release process a lot simpler and less likely to break conda. Woohoo!!
> 
> So, I take it we'll still need to do the actual conda release manually, right?
> 
> @freddyaboulton once this is in, could you update release.md to reflect this?
> 
> @ryan-s I left a bunch of questions, and two comments to address before merge:
> 
> * Please move that new file `conda_config.py` to somewhere other than the top-level directory -- I think it should go in `.circleci/`. Hmm, on reflection, since its python code, perhaps it should go in a new folder `evalml/tests/circleci/`. Your call there.
> * I left a comment about using `Pathlib`--let's resolve that, either by making the change or by you telling me its unnecessary lol
> 
> 🚢

As of right now yes, but in theory it wouldn't be too hard to copy the artifact out of the container and push it.  You would just need to do the inverse of what we do when we copy our feedstock into the container.

Issue #1229, opened 2020-09-25
Integrate Woodwork DataTables into EvalML


angela97lin:
This Epic tracks all issues related to integrating [Woodwork DataTables](https://github.com/FeatureLabs/woodwork) into EvalML.

Notes here: https://alteryx.quip.com/XedaAs9HXc9f/EvalML-Woodwork-Table-Integration
Design doc here: https://alteryx.quip.com/Oa4aA6h0Rf2k/Woodwork-DataTable-Integration-Design-Document

Implementation Plan:

* Update AutoML: 1 week
    * Update AutoMLSearch to support WW DataTables. Maintain support for pandas input.
    * AutoML will still pass pandas DFs to the pipelines/components
* Add documentation about what types we expect categorical, numeric and text to be set to in order for automl to treat features correctly, as well warn users what may happen if users pass in pandas DataFrames / numpy arrays. 2 days

* Update pipelines and all components to accept WW DataTables: 1 week
    * Can (and should) be done in groups
* Update AutoML to pass datatables to every pipeline/component instead of pandas DF. 1 day
* Update components from selecting using pandas dtypes to selecting using DataTable logical types instead. 3 days
* Update documentation to use DataTables in all examples. 2 days
* Update all graph methods, util methods which accept data, to support DataTables. 1 week

Total: 3 weeks 3 days, then 1 more week to do graphing/utils.

*Key Dates*
Oct release is Tues October 27th.

*Goal*
Complete all by Fri Nov 6th (4 weeks). Ship in Nov release.

*Stretch Goal*
Everything except graphing/utils done by the October release (Oct 27th).

angela97lin:
Implementation Plan:

dsherry:
@angela97lin : I think we should drop new datatable types which we won't support initially. Then we can add support for those on a case-by-case basis. So:
```python
dt = ... # woodwork datatable as input to automl search, for example
numeric_features = dt.select('numeric')
categorical_features = dt.select('categorical')
natural_language_features = dt.select('natural_language')
# then, in order to drop unsupported types, from here on out, don't use
# the other features which may still be held in the original datatable
```

A notable example: the fraud dataset uses [lat/long](https://evalml.alteryx.com/en/stable/demos/fraud.html). Currently, it passes that in as two independent floats. Short-term, we should just drop. I suppose we could encode to two independent floats, but I doubt that'll perform well. Long-term, we'll want to use featuretools' latlong-compatible primitives to build features for that type.

That make sense?

[...]

PR #1249, opened 2020-09-30
Allow user to filter which features to encode for OneHotEncoder

### LINKED ISSUE #1237: OneHotEncoder: allow user to filter which features to encode
**Background**
Currently, `OneHotEncoder` will encode all features passed in during `fit` which are categorical. That's great; simple and easy to understand.

**Proposal**
Suppose a user wants to apply the encoder to a subset of the features. Today, they could do this by filtering which features they pass to the call to `fit`. But it could be easier to support a `features_to_encode` list input, which specifies the feature names to include in the encoding, provided they're categorical features.

Error cases to handle: user specifies a feature which doesn't exist, or feature isn't categorical type.

Open question: what happens to other categoricals? Do they get passed through un-encoded? That seems like the right thing to do, but we should consider.

**Implementation Suggestion**
The first thing `OneHotEncoder.fit` currently does is get all the categorical features by calling `self._get_cat_cols(X_t)`. We should now instead take the intersection of that output with `features_to_encode`.

angela97lin:
Closes #1237.

I chose to raise a ValueError if any column in `features_to_encode` does not exist in the input DataFrame.

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -87,8 +87,9 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
 
         if self.features_to_encode is None:
             self.features_to_encode = self._get_cat_cols(X_t)
-        if not set(self.features_to_encode).issubset(list(X.columns)):
-            raise ValueError()
+        invalid_features = [col for col in self.features_to_encode if col not in list(X.columns)]
+        if len(invalid_features) > 0:
+            raise ValueError("Could not find and encode {} in input data.".format(', '.join(invalid_features)))
 
         if self.parameters['handle_missing'] == "as_category":
             X_t[self.features_to_encode] = X_t[self.features_to_encode].replace(np.nan, "nan")
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -20,6 +20,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
 
     def __init__(self,
                  top_n=10,
+                 features_to_encode=None,
                  categories=None,
                  drop=None,
                  handle_unknown="ignore",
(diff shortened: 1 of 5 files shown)


evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -112,7 +121,7 @@ def fit(self, X, y=None):
         self._encoder = SKOneHotEncoder(categories=categories,
                                         drop=self.parameters['drop'],
                                         handle_unknown=self.parameters['handle_unknown'])
-        self._encoder.fit(X_t[self._cols_to_encode])
+        self._encoder.fit(X_t[self.features_to_encode])
```
freddyaboulton, line 124: Right now the encoder is being fit on different data than what it would transform. In transform, we filter to the columns that are categorical. This creates a bug:

```python
from evalml.pipelines import OneHotEncoder
import pandas as pd
df = pd.DataFrame({"a": [1, 2, 3], "b": ["c", "d", "c"]})
ohe = OneHotEncoder(features_to_encode=["a", "b"])
ohe.fit_transform(df)
```

```python
ValueError: The number of features in X is different to the number of features of the fitted data. The fitted data had 2 features and the X has 1 features.
```

I think there are two solutions:
1. Only fit on columns that are in `self.features_to_encode` and categorical
2. Fit/transform on whatever the user passes in as `features_to_encode` if not None. 

I prefer 2 (see my comment in one of your tests as to why)

freddyaboulton (changes requested):
@angela97lin I think this is great! Right now, we fit and transform on different columns which can cause bugs. I'm a fan of letting users fit/transform on the columns they pass in as `features_to_encode` to get around this but open to other solutions.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: Ah, good catch! Let me update this to encode `features_to_encode` even if they're numeric. I think that makes sense too given that if a user is explicitly passing us the column name then they want it to be one-hot encoded even if its numeric. (arguably, maybe that user should be setting the dtype to categorical / object? :P)

freddyaboulton: Yea I agree that the best solution is to let the user tell us by typing their columns appropriately but that's hairy with pandas. Hopefully #1229 can help with this!

freddyaboulton (approved):
@angela97lin Thanks this looks great!

PR #1250, opened 2020-09-30
allow `add_to_rankings` to work before search

### LINKED ISSUE #900: Allow `add_to_rankings()` to be called before search is ran
Currently `add_to_rankings()` raises an error when search has not been run. This is because our cross validation splits are currently initialized in `search()`. The following test cases showcases this:

```
def test_add_to_rankings_no_search(mock_fit, mock_score, dummy_binary_pipeline_class, X_y):
    X, y = X_y
    automl = AutoMLSearch(problem_type='binary', max_pipelines=1, allowed_pipelines=[dummy_binary_pipeline_class])

    mock_score.return_value = {'Log Loss Binary': 0.1234}
    test_pipeline = dummy_binary_pipeline_class(parameters={})
    automl.add_to_rankings(test_pipeline, X, y)

    assert len(automl.rankings) == 1
    assert 0.1234 in automl.rankings['score'].values
```

We need to think about where to keep initialization of cross validation and how to keep it consistent in `add_to_rankings()`

bchen1116:
fix #900 

Implementation to allow `add_to_rankings()` to be called before AutoMLSearch().search().

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -337,6 +337,22 @@ class AutoMLSearch:
             else:
                 leading_char = ""
 
+    def _set_data_split(self, X):
+        """Sets the data split method for AutoMLSearch
+
+        Arguments:
+            X (DataFrame): Input dataframe to split
+        """
+        if self.problem_type == ProblemTypes.REGRESSION:
+            default_data_split = KFold(n_splits=3, random_state=self.random_state)
+        elif self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+            default_data_split = StratifiedKFold(n_splits=3, random_state=self.random_state)
+
+        if X.shape[0] > self._LARGE_DATA_ROW_THRESHOLD:
+            default_data_split = TrainingValidationSplit(test_size=self._LARGE_DATA_PERCENT_VALIDATION)
+
+        self.data_split = self.data_split or default_data_split
+
     def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_plot=True):
         """Find the best pipeline for the data set.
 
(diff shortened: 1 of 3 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -714,8 +714,45 @@ def test_add_to_rankings_no_search(mock_fit, mock_score, dummy_binary_pipeline_c
...
+    assert automl.data_split is None
+
+    automl.add_to_rankings(test_pipeline, X, y)
+    assert isinstance(automl.data_split, TrainingValidationSplit)
+    assert len(automl.rankings) == 1
+    assert 0.1234 in automl.rankings['score'].values
+
+
+@patch('evalml.pipelines.RegressionPipeline.score')
+@patch('evalml.pipelines.RegressionPipeline.fit')
```
dsherry, line 743: No need to mock `fit` if we're not calling `automl.search`

Do we have separate coverage already for calling `add_to_rankings` after `search` has been called? It would be nice to add that to this test since you're already mocking `fit` :)

evalml/automl/automl_search.py
```diff
@@ -798,8 +805,7 @@ def add_to_rankings(self, pipeline, X, y):
         if not isinstance(y, pd.Series):
             y = pd.Series(y)
 
-        if not self.has_searched:
-            raise RuntimeError("Please run automl search before calling `add_to_rankings()`")
+        self._set_data_split(X)
```
dsherry, line 808: Looks great!

dsherry (approved):
@bchen1116 awesome! I left one test comment but once that's outta the way, 🚢

evalml/tests/automl_tests/test_automl.py
bchen1116: There are tests that call `search` after `add_to_rankings`. I'll remove the `fit` mocks

jeremyliweishih: Would also be nice to make sure `automl.search` behaves properly if `add_to_rankings` is called first as well.

PR #1252, opened 2020-10-01
Add Plot for Prediction vs Actual for Regression Problems

### LINKED ISSUE #772: Add predicted vs actual plot (regression)
It would be helpful to add a predicted vs actual plot util method for regression pipelines.

Options:
* Simplest: scatter plot. This is where we should start
* Bin along each axis to produce a 2d heatmap
* In the future, this can be extended to order by datetime for timeseries regression problems

For basic regression this can be a scatter plot. Once we add timeseries support this can be extended to be ordered by time.

Others will want to use evalml who will be making this sort of plot, and rendering the results in separate UIs. If we separate the data generation functionality from the plotly plotting, we could recycle the data generation between those two teams. However, the requirements for those two use-cases may differ significantly. So my current plan for us here is simply add a plotly util, and once we've all learned and discussed mutual requirements, we can circle back and add a common data function.

bchen1116:
fix #772 

Added `outlier_threshold` for simple outlier detection

Updated Model_Understanding doc [here](https://evalml.alteryx.com/en/bc_772_plot/user_guide/model_understanding.html#Predicted-Vs-Actual-Values-Graph-for-Regression-Problems)

Regular documentation [here](https://evalml.alteryx.com/en/bc_772_plot/generated/evalml.model_understanding.graph_prediction_vs_actual.html#evalml.model_understanding.graph_prediction_vs_actual)

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -495,11 +495,14 @@ def _get_prediction_vs_actual_data(y_true, y_pred, outlier_threshold):
     data = pd.concat([pd.Series(predictions),
                       pd.Series(actual)], axis=1)
     data.columns = ['prediction', 'actual']
-    data['outlier'] = np.where((abs(data['prediction'] - data['actual']) >= outlier_threshold), "#ffff00", "#0000ff")
+    if outlier_threshold:
+        data['outlier'] = np.where((abs(data['prediction'] - data['actual']) >= outlier_threshold), "#ffff00", "#0000ff")
+    else:
+        data['outlier'] = '#0000ff'
     return data
 
 
-def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=100):
+def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=None):
     """Generate a scatter plot comparing the true and predicted values. Used for regression plotting
 
     Arguments:
@@ -507,7 +510,7 @@ def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=100):
         y_pred (pd.Series): The predicted values outputted by the regression model.
         outlier_threshold (int, float): A positive threshold for what is considered an outlier value. This value is compared to the absolute difference
                                  between each value of y_true and y_pred. Values within this threshold will be blue, otherwise they will be yellow.
-                                 Defaults to 100
+                                 Defaults to None
 
     Returns:
         plotly.Figure representing the predicted vs. actual values graph
@@ -517,7 +520,7 @@ def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=100):
     if jupyter_check():
         import_or_raise("ipywidgets", warning=True)
 
-    if outlier_threshold <= 0:
+    if outlier_threshold and outlier_threshold <= 0:
         raise ValueError(f"Threshold must be positive! Provided threshold is {outlier_threshold}")
 
     if not isinstance(y_true, pd.Series):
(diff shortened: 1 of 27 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -487,3 +486,55 @@ def _calculate_axis_range(arr):
...
+
+
+def _get_prediction_vs_actual_data(y_true, y_pred, outlier_threshold):
+    """Helper method to help calculate the y_true and y_pred dataframe, with a column for outliers"""
+    predictions = y_pred.reset_index(drop=True)
+    actual = y_true.reset_index(drop=True)
+    data = pd.concat([pd.Series(predictions),
+                      pd.Series(actual)], axis=1)
+    data.columns = ['prediction', 'actual']
+    data['outlier'] = np.where((abs(data['prediction'] - data['actual']) >= outlier_threshold), "#ffff00", "#0000ff")
```
bchen1116, line 498: needed to encode colors as hex values

evalml/model_understanding/graphs.py
```diff
@@ -487,3 +486,55 @@ def _calculate_axis_range(arr):
...
+    return data
+
+
+def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=100):
+    """Generate a scatter plot comparing the true and predicted values. Used for regression plotting
+
+    Arguments:
+        y_true (pd.Series): The real target values of the data
+        y_pred (pd.Series): The predicted values outputted by the regression model.
+        outlier_threshold (int): A positive threshold for what is considered an outlier value. This value is compared to the absolute difference
```
angela97lin, line 508: Would be useful if `outlier_threshold` was not restricted to int and could be a float instead.

angela97lin (dismissed):
Looking good but I have two immediate comments:

- It would be helpful to post an image of what this looks like or add it to the model_understanding docs and link to that instead!
- I see in the original issue (#772) that we want to support regression and timeseries data. Not sure what the updated requirements were but if this PR adds just for regression (which is fine), could you please file a separate issue to make sure timeseries data doesn't get dropped?

evalml/model_understanding/graphs.py
```diff
@@ -487,3 +486,57 @@ def _calculate_axis_range(arr):
...
+    df = _get_prediction_vs_actual_data(y_true, y_pred, outlier_threshold)
+    data = []
+    title = 'Predicted vs Actual Values Scatter Plot'
+    layout = _go.Layout(title={'text': title},
+                        xaxis={'title': 'Prediction', 'range': _calculate_axis_range(df['prediction'])},
+                        yaxis={'title': 'Actual', 'range': _calculate_axis_range(df['actual'])})
+
+    for color, outlier_group in df.groupby('outlier'):
+        name = "< outlier_threshold" if color == "#0000ff" else ">= outlier_threshold"
+        data.append(_go.Scatter(x=outlier_group['prediction'],
```
freddyaboulton, line 537: Maybe we should add a line at `y = x` to help users interpret the plot?

evalml/model_understanding/graphs.py
```diff
@@ -487,3 +486,57 @@ def _calculate_axis_range(arr):
...
+    predictions = y_pred.reset_index(drop=True)
+    actual = y_true.reset_index(drop=True)
+    data = pd.concat([pd.Series(predictions),
+                      pd.Series(actual)], axis=1)
+    data.columns = ['prediction', 'actual']
+    data['outlier'] = np.where((abs(data['prediction'] - data['actual']) >= outlier_threshold), "#ffff00", "#0000ff")
+    return data
+
+
+def graph_prediction_vs_actual(y_true, y_pred, outlier_threshold=100):
```
freddyaboulton, line 502: Maybe we should not provide a default value for the `outlier_threshold`? What constitutes a large difference depends on the units of the target variable which we can't know without user input.

bchen1116: Mm good point. I can default it as `None` then

freddyaboulton (approved):
@bchen1116 This looks great!

bchen1116:
Submitted issue #1258 to handle adding plot for timeseries.

PR #1253, opened 2020-10-01
Integrate ensemble methods in AutoML

### LINKED ISSUE #1130: Integrate ensembling into AutoML
Epic design doc here: https://alteryx.quip.com/wcJbAoWJVpAN/Ensembling-Design-Document

This issue also tracks running and writing up perf tests before merging.

angela97lin:
Closes #1130


evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -69,7 +83,8 @@ def add_result(self, score_to_minimize, pipeline):
             score_to_minimize (float): The score obtained by this pipeline on the primary objective, converted so that lower values indicate better pipelines.
             pipeline (PipelineBase): The trained pipeline object which was used to compute the score.
         """
-        super().add_result(score_to_minimize, pipeline)
+        if pipeline.model_family != ModelFamily.ENSEMBLE:
+            super().add_result(score_to_minimize, pipeline)
```
angela97lin, line 87: We don't want to add results to tuner for stacked ensemble

evalml/automl/data_splitters/training_validation_split.py
```diff
@@ -5,7 +5,7 @@
 class TrainingValidationSplit(BaseCrossValidator):
     """Split the training data into training and validation sets"""
 
-    def __init__(self, test_size=None, train_size=None, shuffle=False, stratify=None, random_state=0):
+    def __init__(self, test_size=None, train_size=None, shuffle=True, stratify=None, random_state=0):
```
angela97lin, line 8: Unrelated but setting default to True for shuffling :)

evalml/automl/automl_search.py
```diff
@@ -681,6 +686,52 @@ def _compute_cv_scores(self, pipeline, X, y):
         logger.info(f"\tFinished cross validation - mean {self.objective.name}: {cv_score_mean:.3f}")
         return {'cv_data': cv_data, 'training_time': training_time, 'cv_scores': cv_scores, 'cv_score_mean': cv_score_mean}
 
+    def _compute_ensemble_scores(self, pipeline, X, y):
+        start = time.time()
+        cv_data = []
+        X.reset_index(drop=True, inplace=True)
```
angela97lin, line 692: I'm still trying to figure out why this line is necessary. Works without locally, but breaks when running perf tests without. 🤔

angela97lin: edit: testing a few things, maybe this won't be necessary

evalml/automl/automl_algorithm/iterative_algorithm.py
dsherry: Got it 👍  I think its totally fine if we don't add tuners for ensemblers for the time being.

Long-term, it would be great to treat ensembles just like any other pipeline, meaning they have a tuner. For the stacked ensembler, I suppose we could search the final estimator, and the parameters of that estimator (not currently exposed, which is fine). I wonder if there are other algorithm parameters we'll want to expose.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -52,6 +54,18 @@ def next_batch(self):
         if self._batch_number == 0:
             next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}))
                           for pipeline_class in self.allowed_pipelines]
+
+        # One after training all pipelines one round
```
dsherry, line 58: 👍

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -52,6 +54,18 @@ def next_batch(self):
...
+        elif (len(self._first_batch_results) > 1 and
+              self._batch_number != 1 and
+              (self._batch_number) % (len(self._first_batch_results) + 1) == 0):
+            input_pipelines = []
+            for i in range(len(self._first_batch_results)):
+                pipeline_class = self._first_batch_results[i][1]
```
dsherry, line 64: @angela97lin so this ensembler is running after `1 + len(self._first_batch_results)` batches, correct? Where the "1" is for the first batch, and then we run `len(self._first_batch_results)` more batches to do deeper tuning on specific pipelines.

The reason I ask is, it appears we're allowing that number of batches to pass before adding the stacked ensembler. But, the `input_pipelines` list is constructed using only the pipelines from the first batch.

I think we should either a) run stacking immediately after the first batch, using the input list you've got here, or b) run stacking after `1 + len(self._first_batch_results)` batches, using an input list with the latest results. Both are cool starting points, but given that stacking takes a while to run, I'm a fan of option b, because it'll occur later in the process, and theoretically will also be more accurate using the non-first-batch pipelines!

Regarding how to implement that: we haven't had to do this kind of operation in the automl algo to date! I think we'll need a new datastructure (another list like `_first_batch_results`? something else) to make this easy to do. Let's chat about it! :)

evalml/automl/automl_search.py
```diff
@@ -609,6 +616,8 @@ def _compute_cv_scores(self, pipeline, X, y):
         cv_data = []
         logger.info("\tStarting cross validation")
         for i, (train, test) in enumerate(self.data_split.split(X, y)):
+            if pipeline.model_family == ModelFamily.ENSEMBLE and i > 0:
```
angela97lin, line 619: Don't run all folds for ensemble methods, only run first to use the same data split.

dsherry:
@angela97lin I just reviewed your perf test doc. Looks great!! I think we're ready to merge this from a perf testing perspective.

I do have one detail we should follow up on. Now that the lightgbm bugfix has been merged (#1302 ), we should either a) exclude lightgbm from the stacked ensembler in the automl algo code, or b) run the perf tests again, with lightgbm included, and verify there aren’t any new errors.

Other than that, I'm ready to approve this PR once its ready!

angela97lin:
@dsherry Okay! I kicked off a test last night. Seems like there aren't errors now, but I am getting a lot of warnings based on the update:

image

Are we curious and want to wait for tests to complete to see how performance changes with LightGBM? Or is that a nice to have, but not necessary for merging?

evalml/automl/automl_search.py
```diff
@@ -159,6 +161,7 @@ def __init__(self,
         self.data_split = data_split
         self.verbose = verbose
         self.optimize_thresholds = optimize_thresholds
+        self.ensembling = ensembling
```
jeremyliweishih, line 164: Same thing here with docstrings.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -95,18 +96,37 @@ def test_iterative_algorithm_results(dummy_binary_pipeline_classes):
...
+            next_batch = algo.next_batch()
+            assert len(next_batch) == algo.pipelines_per_batch
+
+            cls = dummy_binary_pipeline_classes[(algo.batch_number - 2) % len(dummy_binary_pipeline_classes)]
+            assert [p.__class__ for p in next_batch] == [cls] * len(next_batch)
+            assert algo.pipeline_number == last_pipeline_number + len(next_batch)
+            last_pipeline_number = algo.pipeline_number
+            assert algo.batch_number == last_batch_number + 1
+            last_batch_number = algo.batch_number
+            print([p.parameters for p in next_batch])
```
jeremyliweishih, line 113: extra print

jeremyliweishih (approved):
LGTM on my end - just a couple things to clean up.

evalml/tests/automl_tests/test_iterative_algorithm.py
angela97lin: I think @dsherry originally added the print statement cause it added clarity, so just keeping that in here :D

evalml/automl/automl_search.py
freddyaboulton: Is this for performance reasons? Would we get rid of this if we implemented our own stacked ensembler? Might be helpful to add a comment here.

evalml/automl/automl_search.py
```diff
@@ -88,6 +89,7 @@ def __init__(self,
                  tuner_class=None,
                  verbose=True,
                  optimize_thresholds=False,
+                 ensembling=True,
```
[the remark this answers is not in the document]
dsherry: @angela97lin and I just discussed and yes we'll leave this default `False` for the time being. Its not that adding stacking dramatically increases the runtime, its that in order for stacking to be effective we have to run several dozen pipelines. But our automl algo is currently set up to complete quickly on CPU-limited systems (desktop).

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1251,7 +1251,7 @@ class DummyPipeline(dummy_binary_pipeline_class):
     np.testing.assert_equal(pipeline_results["percent_better_than_baseline_all_objectives"]['F1'], answer)
 
 
-@pytest.mark.parametrize("max_batches", [None, 1, 5, 8, 9, 10, 12])
+@pytest.mark.parametrize("max_batches", [None, 1, 5, 8, 9, 10, 12, 20])
 @patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.8})
 @patch('evalml.pipelines.BinaryClassificationPipeline.fit')
 def test_max_batches_works(mock_pipeline_fit, mock_score, max_batches, X_y_binary):
```
freddyaboulton, line 1257: I think we should add a test that checks whether the ensembler is fit the correct number of times for different values of max_iterations. For example, I ran max_pipelines=49 and max_pipelines=50 and the ensembler only ran for max_pipelines=50. That's expected but might be good to document that in a test.

evalml/automl/automl_search.py
dsherry: Oooh. This is slick.

dsherry: Can we add a log message in this case? Just so its extra clear that we don't support CV for ensemble methods, and will override default or user-specified behavior

Making a mental note that we should add this to the user guide for stacking

evalml/pipelines/utils.py
```diff
@@ -143,3 +145,10 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
         custom_name = pipeline_name
         component_graph = [c.__class__ for c in component_instances]
     return TemplatedPipeline({c.name: c.parameters for c in component_instances})
+
+
+def _make_stacked_ensemble_pipeline(input_pipelines, problem_type):
```
dsherry, line 150: I know this is private but it would be great to have a docstring!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1268,16 +1270,20 @@ def test_max_batches_works(mock_pipeline_fit, mock_score, max_batches, X_y_binar
...
-
     assert automl._automl_algorithm.batch_number == max_batches
-    # We add 1 to pipeline_number because _automl_algorithm does not know about the baseline
     assert automl._automl_algorithm.pipeline_number + 1 == n_automl_pipelines
     assert len(automl.results["pipeline_results"]) == n_results
-    assert automl.rankings.shape[0] == min(1 + len(automl.allowed_pipelines), n_results)
+    if num_ensemble_batches == 0:
+        assert automl.rankings.shape[0] == min(1 + len(automl.allowed_pipelines), n_results)  # add one for baseline
+    else:
+        assert automl.rankings.shape[0] == min(2 + len(automl.allowed_pipelines), n_results)  # add two for baseline and stacked ensemble
```
dsherry, line 1286: Do we have a check somewhere that asserts stacking has been run? I.e. look at the search order, grab the last pipeline, and assert its a stacked ensembler?

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -95,18 +96,37 @@ def test_iterative_algorithm_results(dummy_binary_pipeline_classes):
...
-        last_batch_number = algo.batch_number
-        print([p.parameters for p in next_batch])
-        all_parameters.extend([p.parameters for p in next_batch])
-        scores = -np.arange(0, len(next_batch))
-        for score, pipeline in zip(scores, next_batch):
-            algo.add_result(score, pipeline)
-    assert any([p != dummy_binary_pipeline_classes[0]({}).parameters for p in all_parameters])
+
+    for i in range(1, 5):
+        for _ in range(len(dummy_binary_pipeline_classes)):
```
dsherry, line 103: This is totally a style nit-pick / me just confusing myself, haha, but why the double-for here?

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -95,18 +96,37 @@ def test_iterative_algorithm_results(dummy_binary_pipeline_classes):
...
+            next_batch = algo.next_batch()
+            assert len(next_batch) == 1
+            assert algo.batch_number == last_batch_number + 1
+            last_batch_number = algo.batch_number
+            assert algo.pipeline_number == last_pipeline_number + 1
+            last_pipeline_number = algo.pipeline_number
+            scores = np.arange(0, len(next_batch))
+            for score, pipeline in zip(scores, next_batch):
+                algo.add_result(score, pipeline)
+            assert pipeline.model_family == ModelFamily.ENSEMBLE
```
dsherry, line 132: Awesome!

evalml/tests/automl_tests/test_automl.py
dsherry: Ah I see we do that in the iterative algo tests, nice

freddyaboulton (approved):
@angela97lin I think this is great! I have some questions (mainly for my understanding lol)

dsherry (approved):
@angela97lin so exciting that we're ready to 🚢  this, nice going!

As discussed, let's change the automl default to `ensembling=False`. Other than that, all that's blocking: one comment in iterative algo, log message in `_compute_cv_scores`

It would be great to add the missing docstrings as well, and I left a couple other suggestions.

evalml/tests/automl_tests/test_iterative_algorithm.py
angela97lin: Ah, the for i in range(1, 5) is every cycle of batches (goes through every pipeline class) so it tests that the ensemble is called at the end of every cycle. And then the inner `for` loop is for each pipeline class. Really confusing stuff 😂

evalml/automl/automl_search.py
```diff
@@ -425,8 +430,14 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
...
-            self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self._max_batches - 1))
-
+            if self.ensembling:
+                ensemble_nth_batch = len(self.allowed_pipelines) + 1
+                num_ensemble_batches = (self._max_batches - 1) // ensemble_nth_batch
+                self.max_iterations = (1 + len(self.allowed_pipelines) +
+                                       self._pipelines_per_batch * (self._max_batches - 1 - num_ensemble_batches) +
+                                       num_ensemble_batches)
+            else:
+                self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self._max_batches - 1))
```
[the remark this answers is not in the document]
dsherry: (🤦  it's [`max_pipelines` which we're about to delete](https://github.com/alteryx/evalml/pull/1264), not `max_iterations`)

evalml/automl/automl_search.py
angela97lin: Yup @freddyaboulton it's for performance reasons--we don't do CV on the ensemble (since it internally does CV). Will add a comment + log a message!

evalml/tests/automl_tests/test_automl.py
angela97lin: Ah yes, I added max_batches=20 because I was testing with a subset. You're right, 49 and 50 are good numbers since right now, we have 8 pipeline classes, so the stacked ensemble runs for the first time after Baseline (1) + 8 pipelines in first batch (8) + 8 *5 for each pipeline class tuning round.

angela97lin: (Gonna try making this dynamic since this number will update when we add more estimators 🤔 ) Thanks for the great suggestion!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -69,10 +88,23 @@ def add_result(self, score_to_minimize, pipeline):
             score_to_minimize (float): The score obtained by this pipeline on the primary objective, converted so that lower values indicate better pipelines.
             pipeline (PipelineBase): The trained pipeline object which was used to compute the score.
         """
-        super().add_result(score_to_minimize, pipeline)
+        if pipeline.model_family != ModelFamily.ENSEMBLE:
+            super().add_result(score_to_minimize, pipeline)
         if self.batch_number == 1:
             self._first_batch_results.append((score_to_minimize, pipeline.__class__))
 
+        if pipeline.model_family not in self._best_pipeline_params and score_to_minimize is not None:
```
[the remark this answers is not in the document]
angela97lin: Haha I think this came out of the `test_pipelines_in_batch_return_none` test which checks what if the pipeline returns None 😅  When None is returned, we get an `TypeError: '<' not supported between instances of 'NoneType' and 'int'`; np.nan doesn't throw this error since you can compare (though `score_to_minimize < current_best_score` will always be False so we never update the value to np.nan)

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -52,6 +58,19 @@ def next_batch(self):
...
+              (self._batch_number) % (len(self._first_batch_results) + 1) == 0):
+            input_pipelines = []
+            for pipeline_dict in self._best_pipeline_params.values():
+                pipeline_class = pipeline_dict['pipeline_class']
+                pipeline_params = pipeline_dict['parameters']
+                input_pipelines.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, pipeline_params)))
+            ensemble = _make_stacked_ensemble_pipeline(input_pipelines, input_pipelines[0].problem_type)
+            next_batch.append(ensemble)
         else:
             idx = (self._batch_number - 1) % len(self._first_batch_results)
```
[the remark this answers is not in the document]
angela97lin: Ahh, really good catch. We have to update it if `self.ensembling` is True:
```           
num_pipeline_classes = (len(self._first_batch_results) + 1) if self.ensembling else len(self._first_batch_results) 
idx = (self._batch_number - 1) % num_pipeline_classes
```

angela97lin:
@dsherry @freddyaboulton Thanks for all of the great feedback! I've addressed all of the comments you guys made, so this PR should be good to merge :D If there's anything to follow up on, we can open a separate PR after!

PR #1254, opened 2020-10-01
Add HighVarianceCVDataCheck

### LINKED ISSUE #1117: Replace `high_variance_cv` warning in `AutoMLSearch` to use a DataCheck
Currently if our cross validation scores have high variance - we issue a warning and append a column to AutoML results. This can be updated to use a DataCheck instead.

jeremyliweishih:
Fixes #1117.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -345,12 +345,12 @@ class AutoMLSearch:
             X (DataFrame): Input dataframe to split
         """
         if self.problem_type == ProblemTypes.REGRESSION:
-            default_data_split = KFold(n_splits=3, random_state=self.random_state)
+            default_data_split = KFold(n_splits=3, random_state=self.random_state, shuffle=True)
         elif self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
-            default_data_split = StratifiedKFold(n_splits=3, random_state=self.random_state)
+            default_data_split = StratifiedKFold(n_splits=3, random_state=self.random_state, shuffle=True)
 
         if X.shape[0] > self._LARGE_DATA_ROW_THRESHOLD:
-            default_data_split = TrainingValidationSplit(test_size=self._LARGE_DATA_PERCENT_VALIDATION)
+            default_data_split = TrainingValidationSplit(test_size=self._LARGE_DATA_PERCENT_VALIDATION, shuffle=True)
 
         self.data_split = self.data_split or default_data_split
 
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -337,6 +337,22 @@ class AutoMLSearch:
             else:
                 leading_char = ""
 
+    def _set_data_split(self, X):
+        """Sets the data split method for AutoMLSearch
+
+        Arguments:
+            X (DataFrame): Input dataframe to split
+        """
+        if self.problem_type == ProblemTypes.REGRESSION:
+            default_data_split = KFold(n_splits=3, random_state=self.random_state)
+        elif self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+            default_data_split = StratifiedKFold(n_splits=3, random_state=self.random_state)
+
+        if X.shape[0] > self._LARGE_DATA_ROW_THRESHOLD:
+            default_data_split = TrainingValidationSplit(test_size=self._LARGE_DATA_PERCENT_VALIDATION)
+
+        self.data_split = self.data_split or default_data_split
+
     def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_plot=True):
         """Find the best pipeline for the data set.
 
@@ -373,16 +389,7 @@ class AutoMLSearch:
         if not isinstance(y, pd.Series):
             y = pd.Series(y)
 
-        # Set the default data splitter
-        if self.problem_type == ProblemTypes.REGRESSION:
-            default_data_split = KFold(n_splits=3, random_state=self.random_state)
-        elif self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
-            default_data_split = StratifiedKFold(n_splits=3, random_state=self.random_state)
-
-        if X.shape[0] > self._LARGE_DATA_ROW_THRESHOLD:
-            default_data_split = TrainingValidationSplit(test_size=self._LARGE_DATA_PERCENT_VALIDATION)
-
-        self.data_split = self.data_split or default_data_split
+        self._set_data_split(X)
 
         data_checks = self._validate_data_checks(data_checks)
         data_check_results = data_checks.validate(X, y)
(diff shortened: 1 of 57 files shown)


evalml/automl/automl_search.py
```diff
@@ -763,10 +766,6 @@ def describe_pipeline(self, pipeline_id, return_dict=False):
         logger.info("Total training time (including CV): %.1f seconds" % pipeline_results["training_time"])
         log_subtitle(logger, "Cross Validation", underline="-")
 
-        if pipeline_results["high_variance_cv"]:
```
jeremyliweishih, line 766: I moved the logging behavior from `describe_pipeline` to `_add_result` but otherwise kept the same usage of `high_variance_cv` within rankings etc.. I feel like it is more appropriate to notify during the search process and not just when describing a pipeline. Happy to discuss further!

evalml/automl/automl_search.py
```diff
@@ -673,16 +673,19 @@ def _compute_cv_scores(self, pipeline, X, y):
...
-        # if the coefficient of variance is greater than .2
-        with warnings.catch_warnings():
-            warnings.simplefilter('ignore')
-            high_variance_cv = (cv_scores.std() / cv_scores.mean()) > .2
 
         pipeline_name = trained_pipeline.name
         pipeline_summary = trained_pipeline.summary
         pipeline_id = len(self._results['pipeline_results'])
 
+        high_variance_cv_check = HighVarianceCVDataCheck(threshold=0.2)
```
freddyaboulton, line 681: Should users be allowed to configure this threshold now that this is a DataCheck and we let users configure other data checks?

evalml/data_checks/high_variance_cv_data_check.py
```diff
@@ -0,0 +1,45 @@
...
+        Example:
+            >>> cv_scores = pd.Series([0, 1, 1, 1])
+            >>> check = HighVarianceCVDataCheck(threshold=0.10)
+            >>> assert check.validate("LogisticRegressionPipeline", cv_scores) == [DataCheckWarning("High variance (variance >= 0.1) within cross validation scores. LogisticRegressionPipeline may not perform as estimated on unseen data.", "HighVarianceCVDataCheck")]
+        """
+        if not isinstance(cv_scores, pd.Series):
+            cv_scores = pd.Series(cv_scores)
+
+        messages = []
+        high_variance_cv = (cv_scores.std() / cv_scores.mean()) > self.threshold
```
freddyaboulton, line 40: Just noticed this but `cv_scores.mean()` can be negative for `LeadScoring` and `CostBenefitMatrix` - should we take abs?

evalml/automl/automl_search.py
freddyaboulton: I think this makes sense!

evalml/automl/automl_search.py
```diff
@@ -673,16 +673,19 @@ def _compute_cv_scores(self, pipeline, X, y):
...
+        high_variance_cv_check_results = high_variance_cv_check.validate(pipeline_name=pipeline_name, cv_scores=cv_scores)
+        high_variance_cv = False
+
+        if len(high_variance_cv_check_results) > 0:
```
freddyaboulton, line 685: nit-pick: `if high_variance_cv_check_results`

evalml/data_checks/high_variance_cv_data_check.py
```diff
@@ -0,0 +1,45 @@
...
+        # if there are items that occur less than the threshold, add them to the list of messages
+        if high_variance_cv:
+            warning_msg = f"High variance (variance >= {self.threshold}) within cross validation scores. {pipeline_name} may not perform as estimated on unseen data."
```
freddyaboulton, line 43: Perhaps we should change the message to `High coefficient of variation ... ` because the threshold is not really measuring the variance. Turns out the standard deviation divided by the mean is called the [coefficient of variation](https://www.usablestats.com/lessons/sdcv#:~:text=Coefficient%20of%20Variation%20(CV),15%20or%2015%25.).

freddyaboulton (approved):
@jeremyliweishih I think this looks great! I have some minor comments and a question about whether we should let users parametrize this check but I think the implementation is solid.

evalml/automl/automl_search.py
jeremyliweishih: good thought! Perhaps we can turn this on and off depending on if `data_checks = auto` vs. `data_checks = disabled`. But I don't think it'll fit within the existing API for parameterizing data checks as all those data checks run before search is called whereas this check is called during search. I like the idea but we would need to think about what API changes to make to `AutoMLSearch.search()`.

evalml/data_checks/high_variance_cv_data_check.py
jeremyliweishih: good catch - will make the changes 😄

evalml/tests/automl_tests/test_automl.py
```diff
@@ -73,7 +73,7 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type):
         assert isinstance(results['pipeline_summary'], str)
         assert isinstance(results['parameters'], dict)
         assert isinstance(results['score'], float)
-        assert isinstance(results['high_variance_cv'], np.bool_)
+        assert isinstance(results['high_variance_cv'], bool) or isinstance(results['high_variance_cv'], DataCheckWarning)
```
bchen1116, line 76: Isn't `high_variance_cv` always a bool in automl search?

bchen1116 (approved):
LGTM!

jeremyliweishih:
@dsherry codecov was green before merging master but it's not reporting now. Could you help merge this in? Thanks!

evalml/automl/automl_search.py
```diff
@@ -694,16 +694,18 @@ def _add_result(self, trained_pipeline, parameters, training_time, cv_data, cv_s
...
         pipeline_name = trained_pipeline.name
         pipeline_summary = trained_pipeline.summary
         pipeline_id = len(self._results['pipeline_results'])
 
+        high_variance_cv_check = HighVarianceCVDataCheck(threshold=0.2)
+        high_variance_cv_check_results = high_variance_cv_check.validate(pipeline_name=pipeline_name, cv_scores=cv_scores)
+        high_variance_cv = False
+
+        if high_variance_cv_check_results:
+            logger.warning(high_variance_cv_check_results[0])
```
dsherry, line 706: @jeremyliweishih does this show up in the console in a well-formatted way? I've noticed that `str(check)` doesn't look great. You may have to call `.message`

evalml/automl/automl_search.py
dsherry: 👍  yep @freddyaboulton I agree this should be configurable/disable-able.

This PR is essentially porting existing behavior into a new API (data checks). I'll file an issue now to track making this configurable.

evalml/automl/automl_search.py
dsherry: Agreed!

dsherry (approved):
@jeremyliweishih looks great! Left one comment about how the warning appears -- I think you have to use the `message` accessor. We should add unit test coverage for that.

PR #1256, opened 2020-10-01
Add Test Coverage for IPython

### LINKED ISSUE #1187: Add test coverage for IPython
Add test coverage for tests that require IPython, such as `jupyter_check` in `gen_utils.py`, as well as the various graphing methods that would benefit from this check. This would also improve codecov.

bchen1116:
fix #1187

### DIFF
--- a/evalml/tests/utils_tests/test_gen_utils.py
+++ b/evalml/tests/utils_tests/test_gen_utils.py
@@ -1,5 +1,5 @@
 import inspect
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 
 import numpy as np
 import pytest
@@ -13,7 +13,8 @@ from evalml.utils.gen_utils import (
     get_importable_subclasses,
     get_random_seed,
     get_random_state,
-    import_or_raise
+    import_or_raise,
+    jupyter_check
 )
 
 
@@ -214,3 +215,25 @@ def test_check_random_state_equality():
     rs_1.set_state(tuple(['MT19937', np.array([1] * 624), 0, 1, 0.1]))
     rs_2.set_state(tuple(['MT19937', np.array([1] * 624), 1, 1, 0.1]))
     assert not check_random_state_equality(rs_1, rs_2)
+
+
+@patch('evalml.utils.gen_utils.import_or_raise')
+def test_jupyter_check_errors(mock_import_or_raise):
+    mock_import_or_raise.side_effect = ImportError
+    assert not jupyter_check()
+
+    mock_import_or_raise.side_effect = Exception
+    assert not jupyter_check()
+
+
+@patch('evalml.utils.gen_utils.import_or_raise')
+@patch('IPython.core.getipython.get_ipython')
+def test_jupyter_check(mock_ipython, mock_import_or_raise):
+    mock_import_or_raise.return_value = MagicMock()
+    mock_ipython.return_value = True
+    assert jupyter_check()
+
+    mock_ipython.return_value = False
+    assert not jupyter_check()
+    mock_ipython.return_value = None
+    assert not jupyter_check()
--- a/evalml/tests/utils_tests/test_gen_utils.py
+++ b/evalml/tests/utils_tests/test_gen_utils.py
@@ -231,9 +231,12 @@ def test_jupyter_check_errors(mock_import_or_raise):
 def test_jupyter_check(mock_ipython, mock_import_or_raise):
     mock_import_or_raise.return_value = MagicMock()
     mock_ipython.return_value = True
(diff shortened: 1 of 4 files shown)


evalml/utils/gen_utils.py
```diff
@@ -233,9 +233,13 @@ def jupyter_check():
...
     try:
-        get_ipython()
-        return True
-    except NameError:
+        ipy = import_or_raise("IPython")
+        if (ipy.core.getipython.get_ipython()):
+            return True
+        return False
+    except ImportError:
+        return False
```
dsherry, line 241: No need for `except ImportError` if you already have `except Exception`

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -214,3 +215,28 @@ def test_check_random_state_equality():
...
+def test_jupyter_check_errors(mock_import_or_raise):
+    mock_import_or_raise.side_effect = ImportError
+    assert not jupyter_check()
+
+    mock_import_or_raise.side_effect = Exception
+    assert not jupyter_check()
+
+
+@patch('evalml.utils.gen_utils.import_or_raise')
+@patch('IPython.core.getipython.get_ipython')
```
dsherry, line 230: Delete this line, since the module is coming from `import_or_raise`

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -214,3 +215,28 @@ def test_check_random_state_equality():
...
+def test_jupyter_check(mock_ipython, mock_import_or_raise):
+    mock_import_or_raise.return_value = MagicMock()
+    mock_ipython.return_value = True
+    mock_import_or_raise.core.getipython.get_ipython().return_value = True
```
dsherry, line 234: I think this line needs to be changed to something like
```python
mock_import_or_raise.core.get_ipython.get_ipython = MagicMock()
mock_import_or_raise.core.get_ipython.get_ipython.return_value = True
```
At least I think that's on the right track, even if it doesn't work! Try digging into how to set attributes on mocks and that should get you on the right track.

evalml/automl/automl_search.py
```diff
@@ -377,11 +377,7 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
             self
         """
         # don't show iteration plot outside of a jupyter notebook
-        if show_iteration_plot:
-            try:
-                get_ipython
-            except NameError:
-                show_iteration_plot = False
+        show_iteration_plot = jupyter_check()
```
bchen1116, line 380: Simplified this since it checks for jupyter

test-requirements.txt
```diff
@@ -2,6 +2,7 @@ pytest==4.4.1
 pytest-xdist==1.26.1
 pytest-cov==2.6.1
 nbval==0.9.3
+IPython>=3.2.1
```
bchen1116, line 5: Added to `test-requirements.txt` rather than `requirements.txt` since `nbval` has an `IPython` dependency, causing the `core-dependencies=True` test to fail. @dsherry preferred it to be in `requirements.txt`, but this change made the tests pass. Input on this is welcome.

jeremyliweishih (approved):
very nice! Since `nbval` requires `IPython` I do agree on the move of `IPython` to the requirements. Good work!

test-requirements.txt
freddyaboulton: I think this is ok!

evalml/utils/gen_utils.py
```diff
@@ -233,9 +233,11 @@ def jupyter_check():
         Boolean: True if Ipython, False otherwise
     """
     try:
-        get_ipython()
-        return True
-    except NameError:
+        ipy = import_or_raise("IPython")
+        if (ipy.core.getipython.get_ipython()):
```
freddyaboulton, line 237: nit-pick: I think this can just be `return ipy.core.getipython.get_ipython()`

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -214,3 +215,23 @@ def test_check_random_state_equality():
...
+    assert not jupyter_check()
+
+    mock_import_or_raise.side_effect = Exception
+    assert not jupyter_check()
+
+
+@patch('evalml.utils.gen_utils.import_or_raise')
+def test_jupyter_check(mock_import_or_raise):
+    mock_import_or_raise.return_value = MagicMock()
+    mock_import_or_raise().core.getipython.get_ipython.return_value = True
```
freddyaboulton, line 232: 👏

freddyaboulton (approved):

PR #1265, opened 2020-10-06
Always shuffle data in default automl data split strategies

### LINKED ISSUE #1259: Poor performance on diamond dataset
**Problem**
Automl yields models with negative R2.

**Repro**
Dataset [here](https://alteryx0-my.sharepoint.com/:x:/g/personal/dylan_sherry_alteryx_com/EVJqRYjya5JPmr-xlHDEM08BC4cdo4wgP_WfVubuoGpgaw?e=HeMaw4).
```python
import evalml
import pandas as pd
import numpy as np
df = pd.read_csv('stones_encoded_small.csv')
y = df.pop('total_sales_price')
automl = evalml.automl.AutoMLSearch(problem_type='regression')
automl.search(df, y)
```
Data checks will fail due to highly null / single-value columns. You can disable them with `data_checks='disabled'`. Or, to address them and continue:
```python
cols_to_drop = ['culet_condition', 'fancy_color_dominant_color', 'fancy_color_intensity', 'fancy_color_overtone', 'fancy_color_secondary_color', 'fluor_color', 'image_file_url', 'diamond_id', 'currency_code', 'currency_symbol', 'fancy_color_dominant_color', 'fancy_color_intensity', 'fancy_color_overtone', 'fancy_color_secondary_color', 'has_sarineloupe']
df.drop(columns=cols_to_drop, inplace=True)
automl = evalml.automl.AutoMLSearch(problem_type='regression')
automl.search(df, y)
```
The results are highly similar either way: negative R2 values for all models, i.e. the models can't produce meaningful results.

Switching the metric to MSE and MAE yields similarly poor models.

**Discussion**
My first suspicion is that the features aren't getting the right type. When I look at the dtypes inferred by pandas, I see many are set as `float64` but only have a few unique values, i.e. they should be set as categorical. I gave that a shot but it didn't seem to change the model results, so there's more to the story.

dsherry:
Fix #1259

### DIFF
--- a/evalml/automl/data_splitters/training_validation_split.py
+++ b/evalml/automl/data_splitters/training_validation_split.py
@@ -5,7 +5,7 @@ from sklearn.model_selection._split import BaseCrossValidator
 class TrainingValidationSplit(BaseCrossValidator):
     """Split the training data into training and validation sets"""
 
-    def __init__(self, test_size=None, train_size=None, shuffle=True, stratify=None, random_state=0):
+    def __init__(self, test_size=None, train_size=None, shuffle=False, stratify=None, random_state=0):
         """Create a TrainingValidation instance
 
         Arguments:
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -645,6 +645,32 @@ def test_large_dataset_split_size(mock_fit, mock_score):
     assert automl.data_split.test_size == (automl._LARGE_DATA_PERCENT_VALIDATION)
 
 
+def test_data_split_shuffle():
+    # this test checks that the default data split strategy should shuffle data. it creates a target which
+    # increases monotonically from 0 to n-1.
+    #
+    # if shuffle is enabled, the baseline model, which predicts the mean of the training data, should accurately
+    # predict the mean of the validation data, because the training split in each CV fold will contain a mix of
+    # values from across the target range, thus yielding an R^2 of close to 0.
+    #
+    # if shuffle is disabled, the mean value learned on each CV fold's training data will be incredible inaccurate,
+    # thus yielding an R^2 well below 0.
+
+    n = 100000
+    X = pd.DataFrame({'col_0': np.random.random(n)})
+    y = pd.Series(np.arange(n), name='target')
+    automl = AutoMLSearch(problem_type='regression',
+                          max_time=1,
+                          max_pipelines=1)
+    automl.search(X, y)
+    assert automl.results['search_order'] == [0]
+    assert len(automl.results['pipeline_results'][0]['cv_data']) == 3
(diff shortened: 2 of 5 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -645,6 +645,32 @@ def generate_fake_dataset(rows):
...
+    # increases monotonically from 0 to n-1.
+    #
+    # if shuffle is enabled, the baseline model, which predicts the mean of the training data, should accurately
+    # predict the mean of the validation data, because the training split in each CV fold will contain a mix of
+    # values from across the target range, thus yielding an R^2 of close to 0.
+    #
+    # if shuffle is disabled, the mean value learned on each CV fold's training data will be incredible inaccurate,
+    # thus yielding an R^2 well below 0.
+
+    n = 100000
```
freddyaboulton, line 659: If this test doesn't take too long it might be worth running it with n=1000 so that the KFold CV is used?

freddyaboulton (approved):
@dsherry This looks good to me!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -645,6 +645,32 @@ def generate_fake_dataset(rows):
     assert automl.data_split.test_size == (automl._LARGE_DATA_PERCENT_VALIDATION)
 
 
+def test_data_split_shuffle():
```
jeremyliweishih, line 648: great test!

jeremyliweishih (approved):
LGTM

evalml/automl/data_splitters/training_validation_split.py
```diff
@@ -5,7 +5,7 @@
 class TrainingValidationSplit(BaseCrossValidator):
     """Split the training data into training and validation sets"""
 
-    def __init__(self, test_size=None, train_size=None, shuffle=True, stratify=None, random_state=0):
+    def __init__(self, test_size=None, train_size=None, shuffle=False, stratify=None, random_state=0):
```
gsheni, line 8: Should this be `shuffle=True`?

angela97lin (approved):
Great, LGTM!

evalml/automl/data_splitters/training_validation_split.py
dsherry: @gsheni I updated this to be the same as the other sklearn-defined CV methods, which have `shuffle=False` by default. Then, in the automl code which calls this, I set `shuffle=True`. I just did it to be consistent with sklearn.

evalml/tests/automl_tests/test_automl.py
dsherry: Ah shoot you're right. I'll track this down. I upped the number because I wasn't reproing it as reliably at lower values.

PR #1266, opened 2020-10-06
Add RUS to balance cross validation for classification problems

### LINKED ISSUE #973: Balanced sampling for classification problems

jeremyliweishih:
Fixes #973.


evalml/automl/automl_search.py
```diff
@@ -612,13 +613,21 @@ def _compute_cv_scores(self, pipeline, X, y):
             logger.debug(f"\t\tTraining and scoring on fold {i}")
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
+
             if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+                try:
+                    rus = RandomUnderSampler(random_state=self.random_state, replacement=False)
+                    X_train, y_train = rus.fit_resample(X_train, y_train)
+                except Exception:
```
gsheni, line 621: Is it possible to catch a specific Exception, rather than blanket exception?

evalml/automl/automl_search.py
```diff
@@ -612,13 +613,21 @@ def _compute_cv_scores(self, pipeline, X, y):
             logger.debug(f"\t\tTraining and scoring on fold {i}")
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
+
             if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+                try:
+                    rus = RandomUnderSampler(sampling_strategy='not minority' ,random_state=self.random_state, replacement=False)
```
jeremyliweishih: https://imbalanced-learn.org/stable/under_sampling.html#prototype-selection

core-requirements.txt
```diff
@@ -10,3 +10,4 @@ psutil>=5.6.3
 requirements-parser>=0.2.0
 shap>=0.35.0
 texttable>=1.6.2
+imbalanced-learn>=0.7.0
```
dsherry, line 13: @jeremyliweishih what is this?

evalml/automl/automl_search.py
```diff
@@ -612,13 +613,21 @@ def _compute_cv_scores(self, pipeline, X, y):
             logger.debug(f"\t\tTraining and scoring on fold {i}")
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
+
             if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
+                try:
+                    rus = RandomUnderSampler(sampling_strategy='not minority', random_state=self.random_state, replacement=False)
+                    X_train, y_train = rus.fit_resample(X_train, y_train)
+                except ValueError:
+                    logger.warning(f"\t\t\tFold {i}: could not downsample to balance classification")
```
dsherry, line 622: Under what circumstances would this error be raised?

jeremyliweishih: I ran into it when there wasn't enough samples for each class. Maybe I can put a datacheck before this but I only went for a warning for now.

core-requirements.txt
jeremyliweishih: The package that I'm using to down sample and balance classes! Using selection under sampling to do so: https://imbalanced-learn.readthedocs.io/en/stable/under_sampling.html#prototype-selection

dsherry:
@jeremyliweishih given that we discussed this yesterday and have more perf testing to do, I pushed this back to draft.

evalml/automl/automl_search.py
jeremyliweishih: @bchen1116 @dsherry @rpeck 

It seems imblearn does give an error if there are not enough samples. I think there are parameters to manage the number of samples as well.

dsherry:
Closing in favor of other PRs for #973

PR #1270, opened 2020-10-07
Add PCA component

### LINKED ISSUE #1262: Add PCA Component
Add a [PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA) component for dimensionality reduction. First of potentially multiple dimensionality reduction components.

eccabay:
Closes #1262 by introducing a PCA component for dimensionality reduction.

As an initial sanity test, I ran AutoML on the MNIST dataset with and without the PCA component. The original dataset had 784 sparse features, after running PCA this was reduced to 154 features.

Training with current pipelines:
Screen Shot 2020-10-07 at 10 10 32 AM

Training with the same pipelines but including the PCA component with default parameters:
Screen Shot 2020-10-07 at 8 48 07 AM


evalml/pipelines/components/transformers/dimensionality_reduction/pca.py
```diff
@@ -0,0 +1,60 @@
...
+
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils.gen_utils import is_all_numeric
+
+
+class PCA(Transformer):
+    """Reduces the number of features by using Principal Component Analysis"""
+    name = 'PCA Transformer'
+    hyperparameter_ranges = {
+        "variance": Real(0.25, 1)}
```
freddyaboulton, line 13: It's really cool we can parametrize it like this!

evalml/pipelines/components/transformers/dimensionality_reduction/pca.py
```diff
@@ -0,0 +1,60 @@
...
+        return self
+
+    def transform(self, X, y=None):
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+        if not is_all_numeric(X):
+            raise ValueError("PCA input must be numeric")
+
+        X_t = self._component_obj.transform(X)
+        return pd.DataFrame(X_t, index=X.index)
```
freddyaboulton, line 51: not-blocking: Right now, the dataframe doesn't have any column names. Maybe we should name them `[f"component_{i}" for i in range(X.shape[i])]` so that users who call `get_estimator_features` on a pipeline that has a PCA component has a better idea of what happened to their data.

freddyaboulton (approved):
@eccabay I think this is really cool!

Since this component changes the input data fed to the estimator, it has an impact on our model understanding methods.

As it stands now, `graph_partial_dependence` will error if the pipeline has a PCA component and the `explain_predictions*` functions don't show original column names (which is not a bug but maybe we can look into making them display the original columns).

Once this gets merged, I'll file an issue to track that!

evalml/pipelines/components/transformers/dimensionality_reduction/pca.py
```diff
@@ -0,0 +1,60 @@
...
+            raise ValueError("PCA input must be numeric")
+
+        X_t = self._component_obj.transform(X)
+        return pd.DataFrame(X_t, index=X.index, columns=[f"component_{i}" for i in range(X_t.shape[1])])
+
+    def fit_transform(self, X, y=None):
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+        if not is_all_numeric(X):
+            raise ValueError("PCA input must be numeric")
```
angela97lin, line 57: Maybe I'm being too pedantic here but could we say "PCA input must be _all_ numeric" / make more of an emphasis that _all_ values have to be numeric 😂

evalml/tests/component_tests/test_pca.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_pca_numeric():
+    X = pd.DataFrame([[3, 0, 1, 6],
+                      [1, 2, 1, 6],
+                      [10, 2, 1, 6],
+                      [10, 2, 2, 5],
+                      [6, 2, 2, 5]])
+    pca = PCA()
+    expected_X_t = pd.DataFrame([[3.176246, 1.282616],
```
angela97lin, line 16: Just curious, how did you get these expected values? Wondering if there's a way to write / store these values in a way that will make more sense than just a 3d list of floats :O

evalml/tests/component_tests/test_components.py
```diff
@@ -521,6 +524,10 @@ def test_transformer_transform_output_type(X_y_binary):
             if isinstance(component, SelectColumns):
                 assert transform_output.shape == (X.shape[0], 0)
                 assert isinstance(transform_output.columns, pd.Index)
+            elif isinstance(component, PCA):
+                assert transform_output.shape[0] == X.shape[0]
+                assert transform_output.shape[1] < X.shape[1]
```
angela97lin, line 529: Oh cool, I don't know the algorithm but we'll always end up with less columns? There's never the case where `transform_output.shape[1] = X.shape[1]`? :o

angela97lin (approved):
This is fantastic! Just left minor comments / things I'm curious about but LGTM!

evalml/tests/component_tests/test_pca.py
eccabay: Hm, I'm not sure if there's a better way to write these numbers! PCA centers features and projects them down to a smaller vector space, so there's no simple or easy to anticipate relationship between the input and output features. Very open to any suggestions though!

evalml/tests/component_tests/test_components.py
eccabay: Ah, good point! It is possible, since the default is to keep the features that represent *at least* 95% of the variance, which could end up being all features in some cases. I'll update this tot be `<=` instead.

eccabay:
Wrote up a slightly more in-depth summary of my testing with MNIST [here](https://alteryx.quip.com/JNabAHYfXc2w/Initial-PCA-performance-validation)

Issue #561, opened 2020-04-01
Validate component and pipeline fields at class definition


jeremyliweishih:
As #360 gets merged and more custom functionality gets added to automl users will be defining more and more pipelines. I think it is best that pipeline structure, pipeline components, and problem types are validated on definition as they errors would only surface on instantiation. If users are using pipelines with automl these errors would surface during model search which is confusing.  

Likewise, currently pipeline and component required fields are check on instantiation (as part of an ABC) instead of checked on definition.

I've done some brainstorming in the branch ` js_test_decorators` where I explored using a class decorator to check for required fields (this can be extended to check the fields themselves). We can use this solution to also enable properties we need for these classes such as returning copies of class variables.

dsherry:
Incomplete list of some things we could validate on components at definition-time:
* supported_problem_types
* hyperparameter_ranges: ensure that each range contains the default from `default_parameters` (@bchen1116 and I discussed this on #767 )

Incomplete list of some things we could validate on pipelines at definition-time:
* component_graph: ensure all entries are valid components. After pipelines as DAGs (#273 ), check for cycles etc.
* custom_hyperparameters: same as for components

PR #1284, opened 2020-10-09
Freeze Hyperparameters for AutoMLSearch

### LINKED ISSUE #767: Ability to freeze specific hyperparameters
**Description**
There are cases where we'll want to fix certain pipeline parameters to a predetermined value, thus excluding them from the search.

Today, in order to do this, we'd have to define a new pipeline subclass which updates `custom_hyperparameters` to set only one allowed value for the parameters being held constant. Then we'd have to exclude the original pipeline from automl and add the new subclass to automl. This should work fine but feels heavy-handed and hard to use.

**Options**
* Don't do anything; the way to do this is described above
* Same but add a convenience method to wrap the above
* A better way to support this would be to allow users to specify which parameters they want to freeze as an input to the automl search. We wouldn't need to create any subclasses or re-register any pipelines with automl. Instead, we could upgrade our use of the `Tuner` class to be able to exclude the specified parameters from the search space.

**Proposal**
The last option feels more sustainable. What would go into that:
* Add a `pipelines_freeze_parameters` parameter (name TBD) to `AutoSearchBase.__init__`. This could accept a map with partial parameters dicts for each pipeline:
```
{
    'Logistic Regression Pipeline': {
        'Simple Imputer': {
            'imputation_strategy': 'constant',
            'fill_value': 42
        }
    },
    ... # could include params from other pipelines
}
```
* When `AutoSearchBase` creates the tuners for each pipeline during init, it will pass that pipeline's frozen parameter dict to the tuner class
* Update `Tuner` to have an optional init arg `freeze_parameters`. If specified, when the tuner creates the search space, it will exclude the frozen parameters from the space. But when the tuner proposes a new set of parameters to evaluate, the proposal will include the frozen value(s).

bchen1116:
fix #767 

Right now, the implementation assumes that the user includes the 'default' values of the components in the hyperparameter dict. We catch the error that AutoML throws and warn the users if this default value isn't in the provided hyperparameter ranges. Hyperparameter validation would be tracked through [this issue](https://github.com/alteryx/evalml/issues/561). 

Since we have no hyperparameter validation as of yet, if a user creates a new pipeline with EvalML components, redefines a  hyperparameter range that doesn't include the default value, and calls `AutoMLSearch` on this pipeline, `AutoMLSearch` will raise the error that the default values aren't in the search space. This will ideally be addressed in the issue above. Additionally, since we don't do hyperparam validation, we don't catch/throw an error if a user misspells a component name, which will allow `AutoMLSearch` to run with no errors

```python
custom_hyperparameters = {
        # misspell `Imputer'
        "Inpute": {
            "numeric_impute_strategy": ["mean"]
        }
 }
# will run with no errors
make_pipelines(X, y, estimator, 'binary', custom_hyperparameters)
# will run and compute with original `Imputer` hyperparameters, and quietly return the results
AutoMLSearch(problem_type='binary', allowed_pipelines=[make_pipelines]).search(X, y) 
```

This can be fixed once hyperparameter validation is implemented in the before-mentioned issue. 

Docs [here](https://evalml.alteryx.com/en/bc_767_hyperparameters/user_guide/automl.html#Limiting-the-Hyperparameter-Search-Space)


evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -52,6 +52,11 @@ def add_result(self, score_to_minimize, pipeline):
             score_to_minimize (float): The score obtained by this pipeline on the primary objective, converted so that lower values indicate better pipelines.
             pipeline (PipelineBase): The trained pipeline object which was used to compute the score.
         """
+        if self.batch_number == 1:
```
bchen1116, line 55: We use this try/except on batch 1 because this batch contains the pipelines with components initialized on their default args. If this is outside of the search space (hyperparameter ranges) for the components, then we receive a `ValueError`, which we then use to raise a more interpretable `ValueError`.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -52,6 +52,14 @@ def add_result(self, score_to_minimize, pipeline):
...
+            try:
+                self._tuners[pipeline.name].add(pipeline.parameters, score_to_minimize)
+            except ValueError as e:
+                if 'is not within the bounds of the space' in str(e):
```
bchen1116, line 59: Only throw the hyperparameter range error if we receive that error, otherwise we can throw the original error

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -52,6 +52,14 @@ def add_result(self, score_to_minimize, pipeline):
...
+                    raise ValueError("Default parameters for components not in the hyperparameter ranges: {}".format(e))
```
dsherry, line 60: Does this describe which pipeline this problem occurs for? It would be helpful to include the `pipeline.name` in this message.

docs/source/user_guide/automl.ipynb
```diff
@@ -148,6 +148,47 @@
     "automl_custom = evalml.automl.AutoMLSearch(problem_type='multiclass', allowed_pipelines=[CustomMulticlassClassificationPipeline])"
    ]
   },
+  {
```
dsherry, line 151: @bchen1116 could we move this content below "Get best pipeline" and above "Access raw results"?

docs/source/user_guide/automl.ipynb
```diff
@@ -148,6 +148,47 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Limiting the Hyperparameter Search Space\n",
```
dsherry, line 155: I suggest we call this "Limiting the AutoML Search Space"

docs/source/user_guide/automl.ipynb
```diff
@@ -148,6 +148,47 @@
...
+    "custom_hyperparameters = {'Imputer': {\n",
+    "    'numeric_impute_strategy': ['mean', 'median']\n",
+    "}}\n",
+    "\n",
+    "# example of setting parameter to just one value\n",
+    "\"\"\"\n",
+    "custom_hyperparameters = {'Imputer': {\n",
+    "    'numeric_impute_strategy': ['mean']\n",
+    "}}\n",
+    "\"\"\"\n",
```
dsherry, line 183: I think we should include this code un-commented. We could also delete it, but I like that you've included this example too.

dsherry: If you want the multi-value hyperparams to be the ones used in your search code below, then please move this definition to the top, so that the multi-value is defined last and is used in the search

evalml/pipelines/utils.py
```diff
@@ -103,7 +105,11 @@ def make_pipeline(X, y, estimator, problem_type, text_columns=None):
     preprocessing_components = _get_preprocessing_components(X, y, problem_type, text_columns, estimator)
     complete_component_graph = preprocessing_components + [estimator]
 
-    hyperparameters = None
+    if custom_hyperparameters:
+        if not isinstance(custom_hyperparameters, dict):
```
dsherry, line 109: Style nitt-pick:  flatten
```python
if custom_hyperparameters and not isinstance(custom_hyperparameters, dict)
```

dsherry (approved):
@bchen1116 looks great! I left some suggestions on the docs content, but LGTM :D

dsherry:
@bchen1116 RE the bug you ran into with the docs here: I filed as #1367 

We already know one workaround: create a woodwork `DataTable` before calling `make_pipelines`, so that the datetime feature has the right type in the pandas `DataFrame` passed to `make_pipelines` (`datetime64` instead of `object`). It occurs to me another workaround for you in this PR would be to use a dataset which doesn't have a datetime feature 😆

PR #1293, opened 2020-10-09
Clean up lacking codecov for __main__.py and add `evalml info` to docs

### LINKED ISSUE #796: Add/ensure test coverage for CLI commands
#763 updated our logger to use python's logging util. During that transition, we changed testing calls to `capsys` to `caplog` instead. Since our logger is technically logging to both a log file as well as to stdout, it'd be nice to ensure / add test coverage to make sure that our CLI commands are indeed still printing to stdout as well as to the logger.

More broadly, is there a way to test that our logger is actually outputting to several destinations (file + stdout)?

angela97lin:
Closes #796 
Updated docs here: https://evalml.alteryx.com/en/796_cli_check/user_guide/utilities.html
- Fixes broken CLI `evalml info`. Try pip installing evalml and running `evalml info`--it currently breaks. Then, try using `pip install git+ssh://git@github.com/alteryx/evalml.git@796_cli_check` to install in fresh environment, cli should work now :)
- Removes lines of code because it doesn't affect functionality and were previously missed by codecov.
- I think current coverage of `evalml info` to the logger output suffices, capsys and caplog don't play really well with each other 😬


setup.py
```diff
@@ -22,5 +22,8 @@
         ]
     },
     data_files=[('evalml/demos/data', ['evalml/demos/data/fraud_transactions.csv.tar.gz', 'evalml/demos/data/churn.csv']),
-                ("evalml/tests/data", ['evalml/tests/data/tips.csv', 'evalml/tests/data/titanic.csv'])]
+                ('evalml/tests/data', ['evalml/tests/data/tips.csv', 'evalml/tests/data/titanic.csv'])],
+    package_data = {
```
angela97lin, line 26: We need to add `core-requirements.txt` to package_data so that it is included in the binary distribution and available via pip installation.

freddyaboulton: This would install the `core-requirements.txt` file at the `site-packages` level instead of `site-packages/evalml` . It seems kind of weird that we would read a  file that is technically not in our package and it could also conflict with other files called `core-requirements.txt`. 

I think I would prefer the hardcode-a-list-of-requirements approach I think you mentioned at stand-up the other day.

image

Also, I think we should check if the `../` syntax works on windows?

evalml/__main__.py
```diff
@@ -14,7 +14,3 @@ def info():
 
 
 cli.add_command(info)
-
-
-if __name__ == '__main__':
-    cli()
```
dsherry, line 20: Hmm, so this isn't needed for click to work?

setup.py
dsherry: @angela97lin @freddyaboulton ah, I see what you're talking about. The current impl of the `evalml info` CLI happens to read in `core-requirements.txt` in order to do its thing.

I suggest we change the impl instead:
* In `cli_utils.py` delete `get_core_requirements()`
* And update `print_deps()`:
```python
def print_deps(dependencies):
    logger.info("\nINSTALLED VERSIONS")
    logger.info("------------------")
    installed_packages = get_installed_packages()
    packages_to_log = []
    for package in installed_packages:
        packages_to_log.append((x, installed_packages[x]))
        logger.info("{package}: {version}".format(package=package, version=installed_packages[package]))
```

Will that work?

Also. good point about the pathing! So I think that `data_files` and `package_data` are used during package creation, meaning they need to work on whatever platform we build source dist / wheels on. We currently do this on linux. So, we're ok for now :)

dsherry (changes requested):
@angela97lin thanks for looking at this and for adding that guide!

I left a suggestion on the impl. Basically, I wonder if instead of packaging `core-requirements.txt`, we can just change the impl to log out the entirety of `pip freeze`. Let's conclude that conversation one way or another, and then I'll happily approve.

setup.py
angela97lin: @dsherry @freddyaboulton 

Hm, is this what we want though? By removing `get_core_requirements()` / just calling ` get_installed_packages()`, we're getting *all* of the packages the user has installed, regardless of if it is an evalml dependency or not. I'd argue then that this is not much more useful than `pip freeze`. If there's no good workaround, I'm okay with just having a hardcoded list as [Featuretools has] (https://github.com/FeatureLabs/featuretools/blob/main/featuretools/utils/cli_utils.py). Thoughts?

docs/source/user_guide/utilities.ipynb
```diff
@@ -4,14 +4,37 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## System Information\n",
+    "\n",
+    "EvalML provides a command-line interface (CLI) tool prints the version of EvalML and core dependencies installed, as well as some basic system information. To use this tool, just run `evalml info` in your shell or terminal. This could be useful for debugging purposes or tracking down any version-related issues."
+   ]
```
dsherry, line 28: @angela97lin I think there's a bug somewhere -- the output shows a stack trace in the docs

angela97lin: My bad, I was testing out the suggestion you made. Reverted to working state here (https://evalml.alteryx.com/en/796_cli_check/user_guide/utilities.html) until we discuss.

dsherry (approved):
Discussed with @angela97lin . Plan:
* Merge this PR, since it adds missing code coverage
* File issue to track removing `core-requirements.txt` from our package, and update the env CLI to list all installed packages (i.e. `pip freeze`)

angela97lin:
@dsherry Filed https://github.com/alteryx/evalml/issues/1328, so this should be good to merge!

PR #1300, opened 2020-10-13
Add `Component Graph` class

### LINKED ISSUE #1277: Add `ComponentGraph` class
* Component Graph API:
    * Define graph of components
    * Ability to add a new node to the graph
    * Ability to add one graph to another graph (example: building ensembling method)
    * Show a pictorial representation of the graph (support pipeline.graph() )

eccabay:
Closes #1277 as the first step towards implementing pipelines as DAGs. A full description of the purpose of this class is in the design doc [here](https://alteryx.quip.com/YLX3ABQcGUkM/Pipelines-as-DAGs-Design-Doc).

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -58,7 +58,8 @@ class ComponentGraph:
             if isinstance(component_class, ComponentBase):
                 if component_parameters != {}:
                     raise ValueError(f"Attempting to instantiate component {component_name} with parameters {component_parameters} when component already instantiated")
-                continue
+                else:
+                    continue
 
             component_class = handle_component_class(component_class)
 
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -161,10 +161,11 @@ class ComponentGraph:
         return self._component_graph.predecessors(component_name)
 
     def graph(self, name, graph_format):
-        """Generate an image representing the pipeline graph
+        """Generate an image representing the component graph
 
         Arguments:
-            filepath (str, optional): Path to where the graph should be saved. If set to None (as by default), the graph will not be saved.
+            name (str): Name of the graph
+            graph_format (str): file format to save the graph in
 
         Returns:
             graphviz.Digraph: Graph object that can be directly displayed in Jupyter notebooks.
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -1,0 +1,177 @@
+import graphviz
+import networkx as nx
+from networkx.algorithms.dag import topological_sort
+
+from evalml.pipelines.components.utils import handle_component_class
+
+class ComponentGraph:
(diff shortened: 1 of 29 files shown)


evalml/pipelines/component_graph.py
```diff
@@ -0,0 +1,213 @@
...
+            component_name (str): Name of the child component to look up
+
+        Returns:
+            iterator of parent component names
+        """
+        if component_name not in self.component_names.keys():
+            raise ValueError(f'Component {component_name} is not in the graph')
+        return self._component_graph.predecessors(component_name)
+
+    def graph(self, name, graph_format):
```
eccabay, line 163: This function is intended to be called by the `PipelineBase.graph()` function, as there are a few filepath handling things there that I've left to live within the pipeline function.

eccabay:
At the moment, `ComponentGraph.graph()` can be called before or after instantiation. Before instantiation, this looks like:
Screen Shot 2020-10-13 at 11 03 40 AM
After instantiation, the parameters are included as with the currently existing `PipelineBase.graph()` function:
Screen Shot 2020-10-13 at 11 04 52 AM

evalml/pipelines/component_graph.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def instantiate(self, parameters):
+        """Instantiates all components within the graph using the given parameters
+
+        Arguments:
+            parameters (dict): Dictionary with component names as keys and dictionary of that component's parameters as values.
+                               An empty dictionary {} implies using all default values for component parameters.
+        """
+        for component_name, component_class in self.component_names.items():
+            if isinstance(component_class, ComponentBase):
```
freddyaboulton, line 57: Maybe we should continue here rather than throwing a ValueError? Right now, we let users define graphs with a mix classes and instances and by raising we don't let them instantiate the classes.

```python
cg = ComponentGraph() 
cg.add_node('OneHot', OneHotEncoder())
cg.add_node("OneHot2", OneHotEncoder)
cg.instantiate({"OneHot2": {'top_n': 5}})
```

freddyaboulton (approved):
@eccabay This is awesome! I have some questions on the implementation but this looks good to me!

evalml/pipelines/component_graph.py
eccabay: Oh, this is a really interesting point I overlooked! The original intention was to disallow constructing graphs with instances, and that way the graph is "static" until instantiated within the Pipeline class that contains it. However, I can see the added benefit of flexibility in allowing this mix of classes and instances, so I'll update this!

evalml/pipelines/component_graph.py
```diff
@@ -0,0 +1,214 @@
...
+                raise ValueError(f'Cannot instantiate already instantiated component {component_name}')
+
+            try:
```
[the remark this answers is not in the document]
eccabay: 🤦 You're totally right!

eccabay:
This will need a significant amount of rework before merging, pending design decisions with @kmax12 and @dsherry.

Issue #1267, opened 2020-10-06
LightGBM bug: LightGBMError: Check failed: (num_leaves) > (1) at /Users/runner/work/1/s/python-package/compile/src/io/config_auto.cpp, line 318 .


angela97lin:
When running AutoMLSearch on one of the perf test datasets, I ran into this error (after many pipelines) had been trained.
```
df = pd.read_csv('AP_Endometrium_Prostate_1.csv')
y = df['class']
X = df.drop('class', axis=1)
automl = AutoMLSearch(problem_type='binary', max_pipelines=100)
automl.search(X, y)
```

```
---------------------------------------------------------------------------
LightGBMError                             Traceback (most recent call last)
<ipython-input-5-53f318d8717f> in <module>
      1 automl = AutoMLSearch(problem_type='binary', max_pipelines=100)
----> 2 automl.search(X, y)

~/Desktop/evalml/evalml/automl/automl_search.py in search(self, X, y, data_checks, feature_types, show_iteration_plot)
    482                 score_to_minimize = -score if self.objective.greater_is_better else score
    483                 current_batch_pipeline_scores.append(score_to_minimize)
--> 484                 self._automl_algorithm.add_result(score_to_minimize, pipeline)
    485 
    486                 if search_iteration_plot:

~/Desktop/evalml/evalml/automl/automl_search.py in _evaluate(self, pipeline, X, y)
    731         self._add_result(trained_pipeline=pipeline,
    732                          parameters=parameters,
--> 733                          training_time=evaluation_results['training_time'],
    734                          cv_data=evaluation_results['cv_data'],
    735                          cv_scores=evaluation_results['cv_scores'])

~/Desktop/evalml/evalml/automl/automl_search.py in _compute_cv_scores(self, pipeline, X, y)
    651                     logger.info(f"\t\t\tFold {i}: The scores for these objectives will be replaced with nan.")
    652                     logger.info(f"\t\t\tFold {i}: Please check {logger.handlers[1].baseFilename} for the current hyperparameters and stack trace.")
--> 653                     logger.debug(f"\t\t\tFold {i}: Hyperparameters:\n\t{pipeline.hyperparameters}")
    654                     logger.debug(f"\t\t\tFold {i}: Exception during automl search: {str(e)}")
    655                     nan_scores = {objective: np.nan for objective in e.exceptions}

~/Desktop/evalml/evalml/automl/automl_search.py in _compute_cv_scores(self, pipeline, X, y)
[...]

PR #1302, opened 2020-10-13
Fix 'RF' error for LightGBM Classifier

### LINKED ISSUE #1251: LightGBM errors out during AutoMLSearch
Repro:
```
automl = AutoMLSearch(problem_type='binary', max_pipelines=30)
automl.search(X, y)
```

At pipeline 15/30, I get:
```
(15/30) LightGBM Classifier w/ Imputer           Elapsed:00:11
	Starting cross validation
			Fold 0: Encountered an error.
			Fold 0: All scores will be replaced with nan.
			Fold 0: Please check /Users/angela.lin/Desktop/Jupyter/evalml_debug.log for the current hyperparameters and stack trace.
			Fold 1: Encountered an error.
			Fold 1: All scores will be replaced with nan.
			Fold 1: Please check /Users/angela.lin/Desktop/Jupyter/evalml_debug.log for the current hyperparameters and stack trace.
			Fold 2: Encountered an error.
			Fold 2: All scores will be replaced with nan.
			Fold 2: Please check /Users/angela.lin/Desktop/Jupyter/evalml_debug.log for the current hyperparameters and stack trace.
```

The logs produce (across all folds, I just chose to copy/paste one):
```
10/01/2020 01:47:41 PM - DEBUG - automl_search.py: 			Fold 0: Exception during automl search: Check failed: config->bagging_freq > 0 && config->bagging_fraction < 1.0f && config->bagging_fraction > 0.0f at /Users/runner/work/1/s/python-package/compile/src/boosting/rf.hpp, line 35 .
```

bchen1116:
fix #1251 and #1267

Add in `bagging_freq` and `bagging_fraction` parameters for LightGBM classifier
Set `num_leaves` hyperparameter to start at 2 rather than 1 since the LightGBM's expecting a value > 1


evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -40,9 +40,15 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
...
                       "min_child_samples": min_child_samples,
-                      "n_jobs": n_jobs}
+                      "n_jobs": n_jobs,
+                      "bagging_freq": bagging_freq,
+                      "bagging_fraction": bagging_fraction}
         parameters.update(kwargs)
 
+        # if the boosting type is random forest, we want to change the bagging_freq to 1 so that we avoid errors
+        if boosting_type == "rf" and not bagging_freq:
+            parameters.update({'bagging_freq': 1})
```
bchen1116, line 50: I'm updating the bagging frequency here, but I think it would be nice to update the user with this. Can I add logging/printing to the component itself?

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -21,7 +21,7 @@ class LightGBMClassifier(Estimator):
         "boosting_type": ["gbdt", "dart", "goss", "rf"],
         "n_estimators": Integer(10, 100),
         "max_depth": Integer(0, 10),
-        "num_leaves": Integer(1, 100),
+        "num_leaves": Integer(2, 100),
```
dsherry, line 24: Ah, I see. So 1/100th of the time, we were setting `num_leaves` to 1 and the estimator was failing.

It would be great if we had coverage where we fit/transform/predict each component with the min and max allowed values from the defined hyperparameter ranges. Not required for this PR, but... let's keep it in mind :)

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -30,7 +30,7 @@ class LightGBMClassifier(Estimator):
     SEED_MIN = 0
     SEED_MAX = SEED_BOUNDS.max_bound
 
-    def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, max_depth=0, num_leaves=31, min_child_samples=20, n_jobs=-1, random_state=0, **kwargs):
+    def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, max_depth=0, num_leaves=31, min_child_samples=20, n_jobs=-1, random_state=0, bagging_fraction=0.9, bagging_freq=0, **kwargs):
```
dsherry, line 33: Why default `bagging_freq` to 0? Won't that cause the bug when `boosting_type="rf"`? What default does lightgbm choose for this parameter?

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
dsherry: @bchen1116 see my comment above. Perhaps we should prevent automl from recommending this case when `boosting_type="rf"` and `bagging_freq=0`. And then, for direct usage of the component, we could either raise an error when this combination of parameters is used, or let lightgbm raise the error. How ugly is the error?

evalml/tests/component_tests/test_lgbm_classifier.py
```diff
@@ -281,3 +282,38 @@ def test_binary_label_encoding(mock_fit, mock_predict, X_y_binary):
     assert_series_equal(y_arg, y_numeric)
 
     clf.predict(X)
+
+
+def test_binary_rf_not_defaults(X_y_binary):
```
dsherry, line 287: Do we have coverage somewhere for calling `fit` with the default parameters? If I follow correctly, that was one of the cases which was erroring. Specifically, with `boosting_type="rf"` and `bagging_freq=0`

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -40,9 +40,15 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
                       "max_depth": max_depth,
                       "num_leaves": num_leaves,
                       "min_child_samples": min_child_samples,
-                      "n_jobs": n_jobs}
+                      "n_jobs": n_jobs,
+                      "bagging_freq": bagging_freq,
+                      "bagging_fraction": bagging_fraction}
```
dsherry, line 45: @bchen1116 could you please explain why adding these two parameters fixed the bug?

dsherry (approved):
@bchen1116 thanks for digging into this! I understand 1 of the 2 bug fixes. I left a comment asking for an explanation of the 2nd, just so I can follow along. I also left a couple questions/suggestions about how to set up the new default parameters. Approved pending resolution of those conversations.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: LightGBM defaults to 0 for `bagging_freq`. Users can set it to 1 and change `bagging_fraction` if they want to speed up computation and randomly select data for other boosting types, but it's required to be 1 for `boosting_type=rf` (along with `0 < bagging_fraction < 1.0`).

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: As some background, LightGBM has 4 boosting types: "gbdt", "dart", "goss", "rf". `Bagging_freq` refers to the frequency of bagging, where it bags every `bagging_freq = k` iterations (0 means it doesn't bag). `bagging_fraction` refers to the amount of data randomly selected without resampling (1 means select all, 0 means none). This can help speed up the training process. 

The default `bagging_freq` that LightGBM sets is 0, which works with `gbdt`, `dart`, and `goss`. However, for `rf`, since its random forest, LightGBM requires that it uses bagging, which means `bagging_freq` must be 1 and `bagging_fraction` must be set to be below 1.0. By adding those two parameters and changing `bagging_freq` when the `boosting_type=rf`, we do a simple fix to avoid this bug.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: Yeah, that makes sense! I'll keep that in mind

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
dsherry: Thanks for the clear explanation! That makes sense.

Can we tweak the comment you left on line 48:
> if the boosting type is random forest, bagging is required by lightgbm, so we set bagging_freq to 1 in order to avoid errors

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
dsherry: Resolved by comment above.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
dsherry: Got it. This looks good. Is 0.9 the default `bagging_fraction` in lightgbm?

dsherry (approved):
@bchen1116 looks great!

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: @dsherry it defaults to 1.0

bchen1116:
LightGBM doesn't have smart defaults for `bagging_freq` and `bagging_fraction`, so when `boosting_type=rf`, we have to manually set `bagging_fraction < 1.0` with `bagging_freq=0`. We choose `bagging_fraction = 0.9` arbitrarily, so this value can be updated whenever necessary.

dsherry:
Thanks @bchen1116 . Yep, agreed, we may be able to find a better default for the value of 0.9 for `bagging_fraction`. Having this value for now is preferable to having more "magic" behavior where the actual value passed to lightgbm changes.

PR #1306, opened 2020-10-14
Code Generation for Pipelines and Components

### LINKED ISSUE #1028: Code generation for pipelines
Add the ability for a pipeline to generate code (as a string) which can be copy-pasted to retrain and recreate the pipeline.

Initially, we should only support python code. However, in the future we could generate code for other languages (c, java, etc.). We should keep that in mind as we design the APIs.

Some third-party modeling libraries will support this easily and some won't. Let's focus on the easier ones first.

Example: suppose we've defined a binary classification pipeline `LogisticRegressionPipeline` which uses missing value imputation, one-hot encoding and then a logreg estimator. We'd add a method which generates something like the following code as a string:
```python
from evalml import ... # fill in with the right names
def train_pipeline(X, y):
    class CodeGeneratedPipeline(BinaryClassificationPipeline):
        component_graph = [Imputer, OneHotEncoder, LogisticRegressionClassifier]
        custom_name = 'Logistic Regression Pipeline'
    parameters = {'Imputer': {'numeric_impute_strategy': 'mean', ...}, ...}
    pipeline = CodeGeneratedPipeline(parameters=parameters)
    pipeline.fit(X, y)
    return pipeline
```

bchen1116:
fix #1028 

Design Doc [here](https://alteryx.quip.com/WABbAD1QHEE1/Code-Generation-Design-Document)

Component docs [here](https://evalml.alteryx.com/en/bc_code_gen/user_guide/components.html#Generate-Component-Code)
Pipeline docs [here](https://evalml.alteryx.com/en/bc_code_gen/user_guide/pipelines.html#Generate-Code)
Api docs [here](https://evalml.alteryx.com/en/bc_code_gen/api_reference.html#pipeline-utils) and [here](https://evalml.alteryx.com/en/bc_code_gen/generated/evalml.pipelines.components.utils.generate_component_code.html#evalml.pipelines.components.utils.generate_component_code)

Requirements:
- [x] Functions to generate pipeline or component code as a string so users can copy and use it elsewhere
- [x] Documentation for usage
- [x] Support for custom components


evalml/pipelines/components/utils.py
```diff
@@ -217,3 +217,34 @@ def scikit_learn_wrapped_estimator(evalml_obj):
         elif evalml_obj.supported_problem_types == [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
             return WrappedSKClassifier(evalml_obj)
     raise ValueError("Could not wrap EvalML object in scikit-learn wrapper.")
+
+
+def generate_component_code(element):
+    """Creates and returns a string that contains the Python imports and code required for running the EvalML component.
```
dsherry, line 223: Nice description!

evalml/pipelines/components/utils.py
```diff
@@ -217,3 +217,34 @@ def scikit_learn_wrapped_estimator(evalml_obj):
...
+    component_names = [c.name for c in all_components()]
+    # hold the imports needed and add code to end
+    code_strings = []
+    base_string = "\n"
+
+    if not isinstance(element, ComponentBase):
+        raise ValueError("Element must be a component instance, received {}".format(type(element)))
+
+    if element.__class__.name in component_names:
+        code_strings.append("from {} import {}".format(element.__class__.__module__, element.__class__.__name__))
```
dsherry, line 241: Love this!

Can we instead say `if element.__class__ in all_components()`?

evalml/pipelines/components/utils.py
```diff
@@ -217,3 +217,34 @@ def scikit_learn_wrapped_estimator(evalml_obj):
...
+    component_parameters = element.parameters
+    name = element.name[0].lower() + element.name[1:].replace(' ', '')
+    base_string += "{0} = {1}(**{2})" \
+                   .format(name,
+                           element.__class__.__name__,
+                           component_parameters)
+
+    code_strings.append(base_string)
+    return "\n".join(code_strings).lstrip("\n")
```
dsherry, line 250: Why is the `lstrip` necessary?

evalml/pipelines/utils.py
```diff
@@ -147,6 +148,55 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+    """
+    component_names = [c.name for c in all_components()]
+    # hold the imports needed and add code to end
+    code_strings = []
+    if not isinstance(element, PipelineBase):
+        raise ValueError("Element must be a pipeline instance, received {}".format(type(element)))
+
+    component_graph = [com.name for com in element.component_graph]
+    custom_components = [(com.__class__.__name__, com.name) for com in element.component_graph if com.name not in component_names]
+    import_strings = [com.__class__.__name__ for com in element.component_graph if com.name in component_names]
```
dsherry, line 169: As with the component code, let's use `com.__class__` instead of `com.__class__.__name__` when we need to do these comparisons

evalml/pipelines/utils.py
```diff
@@ -147,6 +148,55 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+    if import_strings:
+        code_strings.append("from evalml.pipelines.components import (\n\t{}\n)".format(",\n\t".join(import_strings)))
+    code_strings.append("from {} import {}".format(element.__class__.__bases__[0].__module__, element.__class__.__bases__[0].__name__))
+    # check for other attributes associated with pipeline (ie name, custom_hyperparameters)
+    pipeline_list = []
+    for k, v in sorted(list(filter(lambda item: item[0][0] != '_', element.__class__.__dict__.items())), key=lambda x: x[0]):
```
dsherry, line 175: Fancy!

evalml/pipelines/utils.py
```diff
@@ -147,6 +148,55 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+        if k != 'component_graph':
+            pipeline_list += ["{} = '{}'".format(k, v)] if isinstance(v, str) else ["{} = {}".format(k, v)]
```
dsherry, line 177: Style nit-pick: flatten
```python
if k == 'component_graph':
    continue
```

dsherry: Also let's use `append` since we're adding a single-element list

evalml/pipelines/utils.py
```diff
@@ -147,6 +148,55 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+                  "\tcomponent_graph = {2}\n" \
+                  "{3}" \
+                  "\nparameters = {4}\n" \
+                  "pipeline = {0}(parameters)" \
+                  .format(element.__class__.__name__,
+                          element.__class__.__bases__[0].__name__,
+                          component_graph,
+                          pipeline_string,
+                          element.parameters)
+    # remove quotes from custom components
```
dsherry, line 191: Why is this necessary? Can we catch it further up in the code and avoid having to do a removal here?

dsherry (approved):
@bchen1116 this is awesome!!! Fancy implementation for the pipeline method, wow, I like it. Means we won't have to update every time we add a new pipeline attribute.

I left a few comments but nothing major. I like the API!

I do have one consideration: for non-evalml component definitions, perhaps we should include the entire definition in the generated code. Like [in your docs example](https://evalml.alteryx.com/en/bc_code_gen/user_guide/components.html#Generate-Component-Code), the generated code would include `MyDropNullColumns`. What do you think? My reasoning is that if its not included in evalml, we should include it in the code gen.

dsherry:
Followed up with @bchen1116 , we're gonna pass on including custom component source in the generated code, because the implementation for handling imports would be complex. We'll wait to get a user request / feedback before adding the ability to include custom components in the generated code output.

evalml/pipelines/utils.py
```diff
@@ -147,6 +148,55 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+    pipeline_string = "\t" + "\n\t".join(pipeline_list) + "\n" if len(pipeline_list) else ""
+    # create the base string for the pipeline
+    base_string = "\nclass {0}({1}):\n" \
+                  "\tcomponent_graph = {2}\n" \
+                  "{3}" \
+                  "\nparameters = {4}\n" \
+                  "pipeline = {0}(parameters)" \
+                  .format(element.__class__.__name__,
+                          element.__class__.__bases__[0].__name__,
+                          component_graph,
```
dsherry, line 188: Suggestion to remove the need for the "remove quotes from custom components" code below
```python
component_graph = [com.name for com in element.component_graph]
component_graph_string = ', '.join([com.name for com in element.component_graph if com in custom_components else "'{}'".format(com.name)])
```

evalml/tests/component_tests/test_components.py
```diff
@@ -1023,3 +1024,50 @@ def __init__(self, agg_type, category="green"):
 
     assert MockComponent(agg_type="mean").fit(X, y)
     assert MockComponent(agg_type="moat", category="blue").fit(X, y)
+
+
+def test_generate_code_errors():
+    with pytest.raises(ValueError, match="Element must be a component instance"):
+        generate_component_code(make_pipeline_from_components([RandomForestClassifier()], ProblemTypes.BINARY))
```
dsherry, line 1031: 👍

dsherry (approved):
🚢 ✨

PR #1320, opened 2020-10-19
Added max_batches as a public parameter

### LINKED ISSUE #1294: Show AutoML Batching to Users
- [x] Move batching as a public parameter for AutoMLSearch
- [x] Update the search progress log output for batching
- [ ] Update docs to reflect batching and how that works

christopherbunn:
Updated search progress look (for `max_batches=2`)
image

Resolves #1294

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -12,7 +12,7 @@ Release Notes
         * Added ``HighVarianceCVDataCheck`` and replaced synonymous warning in ``AutoMLSearch`` :pr:`1254`
         * Added ``PCA Transformer`` component for dimensionality reduction :pr:`1270`
         * Updated ``AutoMLSearch`` to support ``Woodwork`` data structures :pr:`1299`
-        * Added ``max_batches`` parameter to `AutoMLSearch` :pr:`1320`
+        * Make ``max_batches`` argument to ``AutoMLSearch.search`` public :pr:`1320`
     * Fixes
         * Fixed ML performance issue with ordered datasets: always shuffle data in automl's default CV splits :pr:`1265`
         * Fixed broken ``evalml info`` CLI command :pr:`1293`
@@ -51,11 +51,7 @@ Release Notes
         * Renamed ``LabelLeakageDataCheck`` to ``TargetLeakageDataCheck`` :pr:`1319`
         * ``max_pipelines`` parameter has been removed from ``AutoMLSearch``. Please use ``max_iterations`` instead. :pr:`1264`
         * ``AutoMLSearch.search()`` will now log a warning if the input is not a ``Woodwork`` data structure (``pandas``, ``numpy``) :pr:`1299`
-
-.. warning::
-
-    **Breaking Changes**
-        * ``__max_batches`` will be deprecated in favor of ``max_batches`` as it is now public :pr:`1320`
+        * Make ``max_batches`` argument to ``AutoMLSearch.search`` public :pr:`1320`
 
 **v0.14.1 Sep. 29, 2020**
     * Enhancements
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -12,6 +12,7 @@ Release Notes
         * Added ``HighVarianceCVDataCheck`` and replaced synonymous warning in ``AutoMLSearch`` :pr:`1254`
         * Added ``PCA Transformer`` component for dimensionality reduction :pr:`1270`
         * Updated ``AutoMLSearch`` to support ``Woodwork`` data structures :pr:`1299`
+        * Added ``max_batches`` parameter to `AutoMLSearch` :pr:`1320`
     * Fixes
         * Fixed ML performance issue with ordered datasets: always shuffle data in automl's default CV splits :pr:`1265`
         * Fixed broken ``evalml info`` CLI command :pr:`1293`
@@ -51,6 +52,10 @@ Release Notes
         * ``max_pipelines`` parameter has been removed from ``AutoMLSearch``. Please use ``max_iterations`` instead. :pr:`1264`
         * ``AutoMLSearch.search()`` will now log a warning if the input is not a ``Woodwork`` data structure (``pandas``, ``numpy``) :pr:`1299`
 
+.. warning::
+
+    **Breaking Changes**
+        * ``__max_batches`` will be deprecated in favor of ``max_batches`` as it is now public :pr:`1320`
 
 **v0.14.1 Sep. 29, 2020**
     * Enhancements
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -481,6 +481,8 @@ class AutoMLSearch:
 
         current_batch_pipelines = []
         current_batch_pipeline_scores = []
+        if self.max_batches:
+            self.current_batch = 0
         while self._check_stopping_condition(self._start):
             try:
                 if len(current_batch_pipelines) == 0:
@@ -489,6 +491,8 @@ class AutoMLSearch:
                             raise AutoMLSearchException(f"All pipelines in the current AutoML batch produced a score of np.nan on the primary objective {self.objective}.")
                         current_batch_pipelines = self._automl_algorithm.next_batch()
                         current_batch_pipeline_scores = []
+                        if self.max_batches:
+                            self.current_batch += 1
                     except StopIteration:
                         logger.info('AutoML Algorithm out of recommendations, ending')
                         break
@@ -504,7 +508,10 @@ class AutoMLSearch:
                     desc = desc[:self._MAX_NAME_LEN - 3] + "..."
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
-                update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
+                if self.max_batches:
+                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start, self.current_batch)
+                else:
+                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
 
                 evaluation_results = self._evaluate(pipeline, X, y)
                 score = evaluation_results['cv_score_mean']
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -596,7 +596,7 @@ class AutoMLSearch:
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
                 update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations,
-                                self._start, 'Baseline')
+                                self._start, '1')
 
                 baseline_results = self._compute_cv_scores(baseline, X, y)
                 self._baseline_cv_scores = self._get_mean_cv_scores_for_all_objectives(baseline_results["cv_data"])
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -596,7 +596,7 @@ class AutoMLSearch:
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
                 update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations,
-                                self._start, '1')
+                                self._start, 1)
 
                 baseline_results = self._compute_cv_scores(baseline, X, y)
                 self._baseline_cv_scores = self._get_mean_cv_scores_for_all_objectives(baseline_results["cv_data"])
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -89,8 +89,7 @@ class AutoMLSearch:
                  verbose=True,
                  optimize_thresholds=False,
                  ensembling=False,
-                 max_batches=None,
-                 _max_batches=None):
+                 max_batches=None):
         """Automated pipeline search
 
         Arguments:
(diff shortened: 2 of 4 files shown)


christopherbunn:
See screenshot above for changes to the progress output. 

I updated the docstrings, but any suggestions for additional documentation edits are very welcome.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1260,6 +1261,10 @@ def test_max_batches_works(mock_pipeline_fit, mock_score, max_batches, X_y_binar
     assert len(automl.results["pipeline_results"]) == n_results
     assert automl.rankings.shape[0] == min(1 + len(automl.allowed_pipelines), n_results)
     assert automl.full_rankings.shape[0] == n_results
+    if check_output:
+        assert f"Searching up to {max_batches} batches for a total of {n_automl_pipelines} pipelines." in caplog.text
+        for i in range(1, max_batches):
+            assert f"({i}: " in caplog.text
```
freddyaboulton, line 1267: :+1:

freddyaboulton (approved):
@christopherbunn Looks great! 

In terms of formatting the progress - maybe we can do something like `Batch 1: Iteration 5/14` to make it self explanatory? I think it might be a good idea to display the batch number regardless of whether the user specified `max_batches`. Whatever we decide, the formatting for the baseline pipeline iteration should be the same as the other iterations. It looks different in the screenshot you posted.

I'm curious what others think. I think what you have now is great too.

christopherbunn:
> In terms of formatting the progress - maybe we can do something like `Batch 1: Iteration 5/14` to make it self explanatory?

@freddyaboulton I think that's a good idea. I was worried that It would look a bit off formatting-wise but just prototyping it

> the formatting for the baseline pipeline iteration should be the same as the other iterations

My reasoning for not having the same formatting is that I think the baseline is not officially considered as part of any batches. That being said, maybe we can mark it as a "Baseline" pipeline or something like that. Alternatively, it could be "Batch 0"? The formatting would look a bit more consistent but I think it could potentially be confusing as well 
  
Taking both of these things into account, the new output could look like this:

image

freddyaboulton:
@christopherbunn I think the output looks good! I am ok with referring to the baseline as `Baseline` instead of `Batch 1` or `Batch 0` but my thought for labeling it as `Batch 1` is that the baseline will get run if the user specifies `max_batches=1` so it's included in that sense.

evalml/automl/automl_search.py
```diff
@@ -86,6 +86,7 @@ def __init__(self,
                  tuner_class=None,
                  verbose=True,
                  optimize_thresholds=False,
+                 max_batches=None,
                  _max_batches=None):
```
dsherry, line 90: @christopherbunn because we named `_max_batches` as a "private" input arg, let's delete it in this PR in favor of `max_batches`. This is why we kept it private, to test it out before renaming to make it public!

docs/source/release_notes.rst
```diff
@@ -33,9 +34,9 @@ Release Notes
 .. warning::
 
     **Breaking Changes**
+        * ``__max_batches`` will be deprecated in favor of ``max_batches`` as it is now public :pr:`1320`
```
dsherry, line 37: Please update to:

> Make ``max_batches`` argument to `AutoMLSearch.search` public

docs/source/release_notes.rst
```diff
@@ -11,6 +11,7 @@ Release Notes
         * Added percent-better-than-baseline for all objectives to automl.results :pr:`1244`
         * Added ``HighVarianceCVDataCheck`` and replaced synonymous warning in ``AutoMLSearch`` :pr:`1254`
         * Added `PCA Transformer` component for dimensionality reduction :pr:`1270`
+        * Added ``max_batches`` parameter to `AutoMLSearch` :pr:`1320`
```
dsherry, line 14: We added this previously, so can you change this to

> Make `max_batches` argument to `AutoMLSearch.search` public

evalml/automl/automl_search.py
```diff
@@ -185,13 +188,25 @@ def __init__(self,
         else:
             raise TypeError("max_time must be a float, int, or string. Received a {}.".format(type(max_time)))
 
+        if _max_batches:
+            if not max_batches:
+                max_batches = _max_batches
+            logger.warning("`_max_batches` will be deprecated in the next release. Use `max_batches` instead.")
```
dsherry, line 194: RE comments above, let's delete this, and `_max_batches`, entirely, in favor of `max_batches`

evalml/automl/automl_search.py
```diff
@@ -185,13 +188,25 @@ def __init__(self,
...
+
+        if max_batches is not None and max_batches <= 0:
+            raise ValueError(f"Parameter max batches must be None or non-negative. Received {max_batches}.")
+        self.max_batches = max_batches
+        # This is the default value for IterativeAlgorithm - setting this explicitly makes sure that
+        # the behavior of max_batches does not break if IterativeAlgorithm is changed.
+        self._pipelines_per_batch = 5
```
dsherry, line 201: Could we add `_pipelines_per_batch` as an automl search parameter as well? I do think we should keep that one private while we're getting parallel support in place, and then once we're confident we wanna keep it we can consider making it public. But it'll be nice to have external control over this for perf testing purposes. Ok to do this separately, or file a separate issue.

evalml/automl/automl_search.py
```diff
@@ -435,7 +443,9 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
         logger.info("Optimizing for %s. " % self.objective.name)
         logger.info("{} score is better.\n".format('Greater' if self.objective.greater_is_better else 'Lower'))
 
-        if self.max_iterations is not None:
+        if self.max_batches is not None:
+            logger.info(f"Searching up to {self.max_batches} batches for a total of {self.max_iterations} pipelines. ")
```
dsherry, line 447: 👍

evalml/automl/automl_search.py
```diff
@@ -475,7 +489,10 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
                     desc = desc[:self._MAX_NAME_LEN - 3] + "..."
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
-                update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
+                if self.max_batches:
+                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start, self.current_batch)
+                else:
+                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
```
dsherry, line 495: Please get the current batch number from `AutoMLAlgorithm` instead:
```python
self._automl_algorithm.batch_number
```

evalml/utils/logger.py
```diff
@@ -80,13 +80,17 @@ def update_pipeline(logger, pipeline_name, current_iteration, max_iterations, st
...
-        status_update_format = "({current_iteration}/{max_iterations}) {pipeline_name} Elapsed:{time_elapsed}"
-        format_params = {'max_iterations': max_iterations, 'current_iteration': current_iteration}
+        if current_batch:
+            status_update_format = "({current_batch}: {current_iteration}/{max_iterations}) {pipeline_name} Elapsed:{time_elapsed}"
+        else:
+            status_update_format = "({current_iteration}/{max_iterations}) {pipeline_name} Elapsed:{time_elapsed}"
+        format_params = {'current_batch': current_batch, 'max_iterations': max_iterations, 'current_iteration': current_iteration}
     else:
         status_update_format = "{pipeline_name} Elapsed: {time_elapsed}"
         format_params = {}
```
dsherry, line 96: Let's clean this method up:
```python
elapsed_time = time_elapsed(start_time)
if current_batch is not None:
    logger.info(f"({current_batch}: {current_iteration}/{max_iterations}) {pipeline_name} Elapsed:{elapsed_time}")
elif max_iterations is not None:
    logger.info(f"{current_iteration}/{max_iterations}) {pipeline_name} Elapsed:{elapsed_time}")
else:
    logger.info(f"{pipeline_name} Elapsed: {elapsed_time}")
```

dsherry (changes requested):
@christopherbunn looks good! I left a few comments.

evalml/automl/automl_search.py
christopherbunn: Sure. I'll file this as a separate issue.

christopherbunn:
> @christopherbunn I think the output looks good! I am ok with referring to the baseline as `Baseline` instead of `Batch 1` or `Batch 0` but my thought for labeling it as `Batch 1` is that the baseline will get run if the user specifies `max_batches=1` so it's included in that sense.

Ah I see, that makes sense. In that case, I'll count the baseline pipeline as Batch 1.

evalml/automl/automl_search.py
```diff
@@ -591,7 +596,7 @@ def _add_baseline_pipelines(self, X, y):
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
                 update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations,
-                                self._start)
+                                self._start, '1')
```
bchen1116, line 599: Why is `'1'` passed in as a string instead of an int?

bchen1116 (approved):
LGTM! Just left one question

evalml/automl/automl_search.py
dsherry: Thanks!

evalml/automl/automl_search.py
```diff
@@ -489,7 +491,10 @@ def search(self, X, y, data_checks="auto", feature_types=None, show_iteration_pl
                     desc = desc[:self._MAX_NAME_LEN - 3] + "..."
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
-                update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
+                if self.max_batches:
+                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start, self._automl_algorithm.batch_number)
```
dsherry, line 495: @christopherbunn I see we're passing this batch number directly to the output. Does it start at 0, or at 1? RE someone else's comment, I think the batch should start at 1 in the log output.

evalml/automl/automl_search.py
dsherry: Hmm yeah I think its fine to pass as an int or a string, right? Its the current batch number, going into a format string in `update_pipeline`

Actually, @christopherbunn, can we please use `self._automl_algorithm.batch_number` here instead? The automl algorithm is responsible for keeping track of the batch number. In this case it'll [start at 0](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/automl_algorithm.py#L38)

evalml/automl/automl_search.py
dsherry: Aha, I read further. Here's what I suggest:
* Everywhere we call `update_pipeline`, pass the batch number in using `self._automl_algorithm.batch_number`. This will be 0 for the first batch
* In `update_pipeline`, update the format string to print `{current_batch+1}` instead of `{current_batch}`

Sound good?

We should also add unit test coverage which checks that the right batch number prints out. My suggestion:
* Set `max_iterations = len(get_estimators(self.problem_type))+1 + _pipelines_per_batch*2` to run 3 batches
* Mock pipeline fit/score to avoid long runtime
* Run search, and using the `caplog` fixture, assert that `Batch 1` appears `len(get_estimators(self.problem_type)) + 1` times, and `Batch 2` and `Batch 3` each appear `_pipelines_per_batch` times.

dsherry (approved):
@christopherbunn great!

I left a couple comments about updating the batch number so it starts at 1 instead of 0 in the display output, and always using `self._automl_algorithm.batch_number` to get the batch number. I also left a suggestion for a test to check the log output is correct.

Otherwise, looks good!

evalml/automl/automl_search.py
christopherbunn: The test case sounds good, I'll add that in. 

> In update_pipeline, update the format string to print {current_batch+1} instead of {current_batch}

@dsherry because the pipeline is counted as its own separate batch, the number of batches would actually be off by one since `batch_number` is incremented by 1 when `next_batch()` is called. Since `next_batch()` is called for the first batch before the output is seen, Batch 1 actually shows up as Batch 2 with `{current_batch+1}`.

I could:
1.  Keep `{current_batch+1}` and move where we increment batch number to when the results are reported back
2. Have the baseline pipeline show as batch 0 and use `self._automl_algorithm.batch_number`  
3. Override the output for the baseline (and not use `self._automl_algorithm.batch_number`) and show it as either batch 1, "baseline",  or "base"

Imo 1 has the potential to get messy since we're modifying `IterativeAlgorithm` so I want to avoid that. I'm leaning towards 2 although 3 seems fine with me as well. If we wanted "base" or "baseline" for 3, then I'll add another case for `update_pipeline()`

dsherry: Ah got it. Thanks.

I'll throw one more option on the pile:
2b. Label the baseline pipeline as batch 1 instead of batch 0. Use `self._automl_algorithm.batch_number` everywhere else, meaning the 1st batch shows up as batch 1, etc.

I think I agree with what you had here, and we should keep the baseline pipeline displayed as "batch 1" along with the others.

Sound good?

christopherbunn: 2b works for me, I'll manually set the baseline as batch 1 then. Thanks!

PR #1321, opened 2020-10-20
Passing random state to pipelines created by IterativeAlgorithm next_batch

### LINKED ISSUE #1181: AutoML: pass random_state to the created pipeline instances
**Problem**
I was digging through the code with @piguy314159265 and [we noticed that `IterativeAlgorithm.next_batch` doesn't initialize pipeline instances with a `random_state` value.](https://github.com/FeatureLabs/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L60)

**Symptom**
In short, artificially decreased variance in the modeling results.

Suppose we call automl search two times, the first time with the default `random_state` of 0, and the second time with a custom value, like 42. The two calls will generate different data splits, and the automl tuners (after the first batch) could select different pipelines and generate different parameters for those pipelines. That's fine. But, for the first batch at least, the pipelines which use default parameters will run *with the same random seed both times*.

Most users wouldn't notice this. However, it means some of our performance testing results may have higher variance than our results to date have reported. Our perf tests are supposed to run each trial with a different random seed. After this issue is fixed, the pipelines will also run with a different random seed.

**Fix**
[Update IterativeAlgorithm](https://github.com/FeatureLabs/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L60) to pass down the random seed when creating pipeline instances:
```python
pipeline_class(parameters=self._transform_parameters(pipeline_class, proposed_parameters), random_state=self._random_state)
```

Once we do this, we should do some performance testing to see what the impact is. My recommendation: 20 trials each on the small datasets yaml. Note we also have to resolve [a similar issue in the perf test code](https://github.com/FeatureLabs/evalml_looking_glass/issues/98) before we can test this.

freddyaboulton:
### Pull Request Description
Fixes #1181 . Also noticed a bug where `PipelineBase.clone` uses random_state = 0 as the default - which means the random state was unexpectedly being changed before fit/score.

-----

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -58,7 +58,7 @@ class IterativeAlgorithm(AutoMLAlgorithm):
 
         next_batch = []
         if self._batch_number == 0:
-            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}))
+            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_state=self.random_state)
                           for pipeline_class in self.allowed_pipelines]
 
         # One after training all pipelines one round
@@ -79,7 +79,8 @@ class IterativeAlgorithm(AutoMLAlgorithm):
             pipeline_class = self._first_batch_results[idx][1]
             for i in range(self.pipelines_per_batch):
                 proposed_parameters = self._tuners[pipeline_class.name].propose()
-                next_batch.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, proposed_parameters)))
+                pl_parameters = self._transform_parameters(pipeline_class, proposed_parameters)
+                next_batch.append(pipeline_class(parameters=pl_parameters, random_state=self.random_state))
         self._pipeline_number += len(next_batch)
         self._batch_number += 1
         return next_batch
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -651,7 +651,7 @@ class AutoMLSearch:
                 y_threshold_tuning = None
                 if self.optimize_thresholds and self.objective.problem_type == ProblemTypes.BINARY and self.objective.can_optimize_threshold:
                     X_train, X_threshold_tuning, y_train, y_threshold_tuning = train_test_split(X_train, y_train, test_size=0.2, random_state=self.random_state)
-                cv_pipeline = pipeline.clone()
+                cv_pipeline = pipeline.clone(pipeline.random_state)
                 logger.debug(f"\t\t\tFold {i}: starting training")
                 cv_pipeline.fit(X_train, y_train)
                 logger.debug(f"\t\t\tFold {i}: finished training")
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -1462,6 +1462,7 @@ def test_input_not_woodwork_logs_warning(mock_fit, mock_score, caplog, X_y_binar
     assert "`y` passed was not a DataColumn. EvalML will try to convert the input as a Woodwork DataTable and types will be inferred. To control this behavior, please pass in a Woodwork DataTable instead." in caplog.text
 
 
+<<<<<<< HEAD
 @patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.8})
 @patch('evalml.pipelines.BinaryClassificationPipeline.fit')
 def test_pipelines_per_batch(mock_fit, mock_score, X_y_binary):
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -1503,14 +1503,16 @@ def test_automl_respects_random_state(mock_fit, mock_score, X_y_binary, dummy_cl
     class DummyPipeline(BinaryClassificationPipeline):
         component_graph = [dummy_classifier_estimator_class]
         num_pipelines_different_seed = 0
+        num_pipelines_init = 0
 
         def __init__(self, parameters, random_state):
             random_state = get_random_state(random_state)
             is_diff_random_state = not check_random_state_equality(random_state, expected_random_state)
+            self.__class__.num_pipelines_init += 1
             self.__class__.num_pipelines_different_seed += is_diff_random_state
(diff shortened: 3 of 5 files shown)


jeremyliweishih (commented):
lgtm!

jeremyliweishih (approved):
lgtm!

angela97lin (approved):
Looks good! Based on the original issue, it looks like we still want to do perf testing to see how this changes results--will this be attached to this PR or done separately?

freddyaboulton:
> Looks good! Based on the original issue, it looks like we still want to do perf testing to see how this changes results--will this be attached to this PR or done separately?

Waiting on [looking glass issue 98](https://github.com/FeatureLabs/evalml_looking_glass/issues/98) for perf tests.

angela97lin:
> > Looks good! Based on the original issue, it looks like we still want to do perf testing to see how this changes results--will this be attached to this PR or done separately?
> 
> Waiting on [looking glass issue 98](https://github.com/FeatureLabs/evalml_looking_glass/issues/98) for perf tests but I think we can do them in this PR!

Sweet, thanks @freddyaboulton! 😊

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -58,7 +58,7 @@ def next_batch(self):
 
         next_batch = []
         if self._batch_number == 0:
-            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}))
+            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_state=self.random_state)
```
dsherry, line 61: Yep!

evalml/automl/automl_search.py
```diff
@@ -644,7 +644,7 @@ def _compute_cv_scores(self, pipeline, X, y):
                 y_threshold_tuning = None
                 if self.optimize_thresholds and self.objective.problem_type == ProblemTypes.BINARY and self.objective.can_optimize_threshold:
                     X_train, X_threshold_tuning, y_train, y_threshold_tuning = train_test_split(X_train, y_train, test_size=0.2, random_state=self.random_state)
-                cv_pipeline = pipeline.clone()
+                cv_pipeline = pipeline.clone(pipeline.random_state)
```
dsherry, line 647: Yep

RE your comment in standup this week: @eccabay and I chose this behavior because we felt copying random_state under the hood was dirty. However, in light of our recent conversation about #1324 (component/pipeline determinism), perhaps we should revisit this, and have `clone` copy `random_state` by default. Do you have thoughts @freddyaboulton @eccabay ? We can move the discussion to an issue and resolve there.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -106,6 +107,7 @@ def test_iterative_algorithm_results(ensembling_value, dummy_binary_pipeline_cla
             num_pipelines_classes = (len(dummy_binary_pipeline_classes) + 1) if ensembling_value else len(dummy_binary_pipeline_classes)
             cls = dummy_binary_pipeline_classes[(algo.batch_number - 2) % num_pipelines_classes]
             assert [p.__class__ for p in next_batch] == [cls] * len(next_batch)
+            assert all(check_random_state_equality(p.random_state, algo.random_state) for p in next_batch)
```
dsherry, line 110: Very nice

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1450,3 +1452,26 @@ def test_input_not_woodwork_logs_warning(mock_fit, mock_score, caplog, X_y_binar
...
+        def __init__(self, parameters, random_state):
+            random_state = get_random_state(random_state)
+            is_diff_random_state = not check_random_state_equality(random_state, expected_random_state)
+            self.__class__.num_pipelines_different_seed += is_diff_random_state
+            super().__init__(parameters, random_state)
+
+    automl = AutoMLSearch(problem_type="binary", allowed_pipelines=[DummyPipeline],
+                          random_state=expected_random_state)
+    automl.search(X, y)
+    assert DummyPipeline.num_pipelines_different_seed == 0
```
dsherry, line 1477: Wow this test is genius 🤯 

Could you add `max_iterations=6` or something in the constructor just to be explicit that we're running more than 1 pipeline? I guess if we were really being paranoid, we'd wanna assert that `DummyPipeline` was initialized more than once too--you could keep a `num_pipelines` alongside `num_pipelines_different_seed`.

dsherry (approved):
@freddyaboulton this PR rocks!! 🤘 🚀  That unit test is amazing ✨

I left a comment about `clone` and random state but its not important for this PR since you included random state in there. Wanna file an issue and we can discuss there?

evalml/automl/automl_search.py
freddyaboulton: Great idea! I filed #1340 .

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Great idea! Nothing wrong with being paranoid 😄

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1450,3 +1452,28 @@ def test_input_not_woodwork_logs_warning(mock_fit, mock_score, caplog, X_y_binar
...
+            random_state = get_random_state(random_state)
+            is_diff_random_state = not check_random_state_equality(random_state, expected_random_state)
+            self.__class__.num_pipelines_init += 1
+            self.__class__.num_pipelines_different_seed += is_diff_random_state
+            super().__init__(parameters, random_state)
+
+    automl = AutoMLSearch(problem_type="binary", allowed_pipelines=[DummyPipeline],
+                          random_state=expected_random_state, max_iterations=10)
+    automl.search(X, y)
+    assert DummyPipeline.num_pipelines_different_seed == 0 and DummyPipeline.num_pipelines_init
```
dsherry, line 1479: Great!

dsherry:
@freddyaboulton any reason not to merge this?

freddyaboulton:
@dsherry I was waiting on perf test PR 149 because I was under the impression perf tests were required for this feature!

freddyaboulton:
@dsherry And I discussed and we agree that perf tests are not needed for this feature because there will be no noticeable change in behavior to users (maybe if they are highly overfitting would they notice a change). We'll run the perf tests with different random states for each run in the future but that shouldn't block this fix.

PR #1322, opened 2020-10-20
Documentation: fix install error for Python 3.6

### LINKED ISSUE #1242: Error installing evalml (shap) when using python 3.6
While testing the latest release, I tried creating a new virtualenv and installing evalml but ran into the following error:

```
Collecting shap>=0.35.0 (from evalml)
  Cache entry deserialization failed, entry ignored
  Using cached https://files.pythonhosted.org/packages/d2/17/37ee6c79cafbd9bb7423b54e55ea90beec66aa7638664d607bcc28de0bae/shap-0.36.0.tar.gz
No files/directories in /private/var/folders/pq/zlxt5w357570xxt07x_splf40000gn/T/pip-build-7o0aqjos/shap/pip-egg-info (from PKG-INFO)
```
Repro steps:

```
python3 -m venv test_env
source test_env/bin/activate
pip install evalml
```

I was running `Python 3.6.0`.

Related issue: https://github.com/slundberg/shap/issues/830

The solutions in that issue [here](https://github.com/slundberg/shap/issues/830#issuecomment-537382789) or [here](https://github.com/slundberg/shap/issues/830#issuecomment-692601174) solved the error but could still be worth looking into.

bchen1116:
Fix #1242 

Doc is [here](https://evalml.alteryx.com/en/bc_1242_shap/install.html)

Add documentation for installing EvalML for python 3.6. Avoids the `shap` error


docs/source/install.ipynb
```diff
@@ -16,6 +16,8 @@
     "pip install evalml\n",
     "```\n",
     "\n",
+    "For Python version 3.6, ensure that versions for `pip >= 18.0` and `setuptools > 40.0.0` in order to install evalml\n",
```
jeremyliweishih, line 19: Maybe: "For Python versions 3.6.X you must install `pip >= 18.0` and `setuptools > 40.0.0` in order to install evalml"

jeremyliweishih (approved):
Looks good. Just a comment on wording.

freddyaboulton (approved):
Looks good to me!

docs/source/install.ipynb
```diff
@@ -16,6 +16,8 @@
     "pip install evalml\n",
     "```\n",
     "\n",
+    "For Python versions 3.6.X you must install `pip >= 18.0` and `setuptools > 40.0.0` in order to install evalml.\n",
```
dsherry, line 19: What happens when a user runs `pip install evalml` but has pip below this version, or setuptools below this version? Should we update our requirements with this?

dsherry: (We don't currently list pip or setuptools as dependencies)

bchen1116: If the user is using 3.6, they run into the `shap` error that caused this issue. I figured adding it into the install docs would be easiest since `pip` and `setuptools` aren't in our requirements docs

angela97lin: Not sure if this is 100% relevant, but I'd like point out that creating a virtual env with 3.6 directly means we don't run into this issue:
```
python3.6 -m pip install virtualenv
python3.6 -m virtualenv my-venv --python=python3.6
source my-venv/bin/activate
python3.6 -m pip install shap (or -r requirements.txt in my case)
```

This works fine. Could be because of the versions of pip and setuptools that are installed with the virtualenv but didn't dig too deeply.

bchen1116: Hmm yeah that's interesting. Looking further into it, when I used @angela97lin's approach, `pip == 20.2.3` and `setuptools == 50.3.1`, as opposed to using pyenv to create a virtualenv for Python3. I think both approaches are valid since the versions of `pip` and `setuptools` are high enough. I think with the current message that I added, it should be enough to allow users to install `evalml` without issues. @dsherry Do you think these two should be added to our dependencies, or would leaving/updating the message be enough?

dsherry: @bchen1116 so as I just wrote on the issue, when I create a pyenv with `3.6.10` and run `pip install .`, it passes. But that fails with `3.6.0`.

Can we update this to say "For older Python versions (3.6.0 in particular), you must install `pip>=18.0` and `setuptools>=40.0.0` before installing evalml."

Great find!

dsherry (approved):
Approved with one comment. Thanks!

PR #1324, opened 2020-10-20
Make OHE deterministic when top_n < no. of categories

### LINKED ISSUE #1279: OHE not deterministic when fit multiple times with top_n < number categories
Repro:

```python
from evalml.pipelines.components import OneHotEncoder
import pandas as pd

df = pd.DataFrame({"categories": ["cat_1"] * 5 + ["cat_2"] * 4 + ["cat_3"]*3 + ["cat_4"]*3 + ["cat_5"]*3})
ohe = OneHotEncoder(top_n=4, random_state=5)
df1 = ohe.fit_transform(df)
df2 = ohe.fit_transform(df)
pd.testing.assert_frame_equal(df1, df2)
```

Users would expect the OHE fit on the same data to produce the same results. This can cause the results of pipelines fit multiple times on the same data to not be deterministic because the data coming out of the OHE is slightly different each time.

I think part of the cause is that every time we call `sample`  in the OHE fit method we mutate the random state.

freddyaboulton:
### Pull Request Description
Fixes #1279

-----

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -112,8 +112,9 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
                 if top_n is None or len(value_counts) <= top_n:
                     unique_values = value_counts.index.tolist()
                 else:
-                    self.random_state.set_state(self._initial_state)
-                    value_counts = value_counts.sample(frac=1, random_state=self.random_state)
+                    new_random_state = np.random.RandomState()
+                    new_random_state.set_state(self._initial_state)
+                    value_counts = value_counts.sample(frac=1, random_state=new_random_state)
                     value_counts = value_counts.sort_values([col], ascending=False, kind='mergesort')
                     unique_values = value_counts.head(top_n).index.tolist()
                 unique_values = np.sort(unique_values)
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -69,6 +69,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
         super().__init__(parameters=parameters,
                          component_obj=None,
                          random_state=random_state)
+        self._initial_state = self.random_state.get_state()
 
     @staticmethod
     def _get_cat_cols(X):
@@ -111,6 +112,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
                 if top_n is None or len(value_counts) <= top_n:
                     unique_values = value_counts.index.tolist()
                 else:
+                    self.random_state.set_state(self._initial_state)
                     value_counts = value_counts.sample(frac=1, random_state=self.random_state)
                     value_counts = value_counts.sort_values([col], ascending=False, kind='mergesort')
                     unique_values = value_counts.head(top_n).index.tolist()
--- a/evalml/tests/component_tests/test_one_hot_encoder.py
+++ b/evalml/tests/component_tests/test_one_hot_encoder.py
@@ -430,7 +430,8 @@ def test_ohe_features_to_encode_no_col_names():
 
 
 def test_ohe_top_n_categories_always_the_same():
-    df = pd.DataFrame({"categories": ["cat_1"] * 5 + ["cat_2"] * 4 + ["cat_3"] * 3 + ["cat_4"] * 3 + ["cat_5"] * 3})
+    df = pd.DataFrame({"categories": ["cat_1"] * 5 + ["cat_2"] * 4 + ["cat_3"] * 3 + ["cat_4"] * 3 + ["cat_5"] * 3,
+                       "numbers": range(18)})
 
     def check_df_equality(random_state):
         ohe = OneHotEncoder(top_n=4, random_state=random_state)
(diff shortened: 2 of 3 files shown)


evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -111,6 +112,7 @@ def fit(self, X, y=None):
                 if top_n is None or len(value_counts) <= top_n:
                     unique_values = value_counts.index.tolist()
                 else:
+                    self.random_state.set_state(self._initial_state)
```
dsherry, line 115: @freddyaboulton could you please explain why this change is necessary?

Is the culprit the call to `value_counts.sample` below? Is this change masking a pandas bug? If so, could we do
```python
# patch non-determinism bug in pandas sample
rs_pandas_patch = np.random_state.RandomState()
rs_pandas_patch.set_state(self._initial_state)
value_counts = value_counts.sample(frac=1, random_state=rs_pandas_patch)
```

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -428,3 +427,17 @@ def test_ohe_features_to_encode_no_col_names():
...
+                       "numbers": range(18)})
+
+    def check_df_equality(random_state):
+        ohe = OneHotEncoder(top_n=4, random_state=random_state)
+        df1 = ohe.fit_transform(df)
+        df2 = ohe.fit_transform(df)
+        pd.testing.assert_frame_equal(df1, df2)
+
+    check_df_equality(5)
+    check_df_equality(get_random_state(5))
```
dsherry, line 443: Test looks good! Have you verified this was failing before the fix was included?

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
freddyaboulton: @dsherry Yes, the culprit is the call to sample! The problem is that the internal state of `self.random_state` gets modified after each call to sample. This is not a pandas bug but expected behavior of `np.random.RandomState`. I believe my fix and your code snippet are equivalent (explicitly setting the state of `self.random_state` to what it was when the OHE was initialized).

evalml/tests/component_tests/test_one_hot_encoder.py
freddyaboulton: Thanks! Yes, this test fails on `main`!

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
dsherry: Ah! Got it. Yep, thanks for that. I think what threw me off was that this call to set state was pretty far into the nested `if`/`else` tree. As a reader that had me thinking this code must *only* happen in this specific block, but really it could happen at a higher level.

(Aside: because python passes by reference, we should avoid calling `self.random_state.set_state`. That'll update the state for the random state obj which is shared outside of this object.)

@freddyaboulton I left a long writeup of my thoughts below. Can we jump on a call tomorrow and discuss what to do?

---

So when I read the reproducer on #1279 a couple weeks back, I wasn't reading carefully, and I thought that the bug was that this would fail:
```python
df1 = OneHotEncoder(top_n=4, random_state=5).fit_transform(df)
df2 = OneHotEncoder(top_n=4, random_state=5).fit_transform(df)
pd.testing.assert_frame_equal(df1, df2)
```

However returning to that reproducer, I see that it was in fact
```python
ohe = OneHotEncoder(top_n=4, random_state=5)
df1 = ohe.fit_transform(df)
df2 = ohe.fit_transform(df)
pd.testing.assert_frame_equal(df1, df2)
```

The way I see it, there's two options here, for `fit`, and for other component methods like `transform`/`predict`:
1. Calling `fit` multiple times in a row will produce different behavior (because one RNG is held by the object)
2. Calling `fit` multiple times in a row will *not* produce different behavior.

I believe this is a design decision (which we haven't made yet!) rather than a bug. I think option 2 is a more non-dev-user-centric opinion, and option 1 is more developer-centric, i.e. easier for us to write, haha.

My gut is telling me to argue that the first snippet above should pass (as it currently does I believe?), and the second snippet *should always fail*, i.e. option 1. My rationale for that argument is that, as the caller, if you create a single object, which initializes the RNG, and then call `fit` on that object twice in a row, you should expect the RNG will be called two sets of times, once per call to `fit`. And that necessarily would produce different behavior the second time.

However, I could see the other side as well, option 2. Its much simpler behavior to say that whenever you set the random state, be it a seed or `np.random.RandomState`, no matter how many times you call `fit`/`transform`/`predict`, you always get the same output. It appears this is sklearn's opinion as well, because the following passes:
```python
import sklearn
import numpy as np
import pandas as pd

lr1 = sklearn.linear_model.LogisticRegression(random_state=np.random.RandomState(5))
state1 = lr1.random_state.get_state()
lr1.fit(X, y)
state2 = lr1.random_state.get_state()
np.testing.assert_array_equal(state1[1], state2[1])
```

Worth noting option 2 is harder to build, and if we decide that's the invariant we want to enforce, we should think about how to enforce that behavior across *all* our components, not just the OHE.

freddyaboulton:
@dsherry @bchen1116 This is good for a second review. Once this is merged, I'll file an issue for tracking the "global" fix we discussed this afternoon.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -111,7 +112,9 @@ def fit(self, X, y=None):
                 if top_n is None or len(value_counts) <= top_n:
                     unique_values = value_counts.index.tolist()
                 else:
-                    value_counts = value_counts.sample(frac=1, random_state=self.random_state)
+                    new_random_state = np.random.RandomState()
```
freddyaboulton, line 115: Decided to define `new_random_state` here rather than at the top of the function because this is the only place it's needed. Happy to change it though!

dsherry: This looks fine to me!

dsherry (approved):
👍 🚢

PR #1326, opened 2020-10-21
Add stacked ensemble to __init__ file for pipelines

angela97lin:
Running into an issue importing here: https://github.com/FeatureLabs/evalml_looking_glass/pull/147

I believe it has to do with the missing import in the `evalml/pipelines/__init__` file since fixing that locally makes the tests run.


dsherry (approved):
Thanks!

Issue #1239, opened 2020-09-28
Visualize Decision Trees


freddyaboulton:
As an evalml user, I'd like to be able to visualize a decision tree classifier or regressor to understand and debug my features and estimator.

While this issue tracks visualizing a single tree, we can explore visualizing an ensemble of trees as well - like Catboost, RandomForest - in the future.

freddyaboulton:
I was poking around sklearn and saw that they have utils for exporting trees to text or a graphviz data format. Since we already have graphviz as one of our requirements we could easily visualize the graph like we visualize the pipeline component graph.

However it would be nice if we could extract the tree structure to a friendlier data format than either text or graphviz because it would allow users to visualize the tree with their own plotting library, like plotly. This follows the pattern we have for our other model understanding methods where we first get the data to visualize and then visualize it but it's a higher lift than using the sklearn utils. 

What's do you guys think is in scope for this release @dsherry @tyler3991 ?

dsherry:
@freddyaboulton good point!

I think there's value in building an intermediate data format now, because evalml users can use that to do a custom visualization in their UI/application. So that would be my preference.

Whoever picks this up: you can choose either path, a) build graphviz method or b) build a graphviz method and build an intermediate data function. But if you choose A please file something separately for the intermediate data function!

PR #1331, opened 2020-10-21
Add Linear Discriminant Analysis component

### LINKED ISSUE #1314: Add LDA Component
Add [Linear Discriminant Analysis](https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html) component for dimensionality reduction. LDA is a classifier as well as a preprocessing component, but this first implementation should keep it as solely a transformer only usable on classification problems.

eccabay:
Closes #1314 

The number of features remaining after transform is called is required to be <= `min(n_classes-1, n_features)`. Because of this, an LDA component would only add value for Multiclass Classification problems, especially those with many target classes. This leads me to wonder if it's worth it to include in evalml, but I'm still leaning towards doing so because it has the potential to significantly decrease training time for those cases. 

Since the only real benefit is for multiclass problems, the performance tests as they stand now are not a good measure of LDA's benefits. In order to get some sort of results, I performed the same test using MNIST as I did for the [PCA component.](https://alteryx.quip.com/JNabAHYfXc2w/Initial-PCA-performance-validation) 

Current AutoML search ran with these results:
Screen Shot 2020-11-10 at 10 06 54 AM

Custom Pipelines that added the LDA component ran with these results:
Screen Shot 2020-11-10 at 10 08 06 AM

The 4-fold decrease in training time is spectacular to see. The performance does decrease noticeably, but the number of features is reduced from 784 down to 9. For comparison, the PCA component in its equivalent test had 152 features and reduced the performance to 95% accuracy.

I'm very interested in others' thoughts as to whether adding this component is worth it.


evalml/pipelines/components/transformers/dimensionality_reduction/lda.py
```diff
@@ -0,0 +1,51 @@
+import pandas as pd
+from sklearn.discriminant_analysis import LinearDiscriminantAnalysis as SkLDA
+
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils.gen_utils import is_all_numeric
+
+
+class LDA(Transformer):
```
freddyaboulton, line 8: I'd vote for changing the name to `LinearDiscriminantAnalysis` because LDA could be Latent Dirichlet Allocation.

evalml/pipelines/components/transformers/dimensionality_reduction/lda.py
```diff
@@ -0,0 +1,51 @@
...
+    name = 'Linear Discriminant Analysis Transformer'
+    hyperparameter_ranges = {}
+
+    def __init__(self, n_components=None, random_state=0, **kwargs):
+        """Initalizes an transformer that reduces the number of features using linear discriminant analysis."
+
+        Arguments:
+            n_components (int): the number of features to maintain after computationn. Defaults to None.
+        """
+        parameters = {"n_components": n_components}
```
freddyaboulton, line 19: What happens if the user passes n_components > min(n_classes - 1, n_features)? Should we prevent that?

freddyaboulton (approved):
@eccabay I think this implementation is great! 

I'd vote to keep this if we make it an estimator instead of a transformer for the following reason:

- In the spectrum of interpretable to black box, LDA is on the interpretable side. It has a closed-form solution that rests on well-understood (maybe not by me 😆 ) statistical theory. Users may want to compare better performing black box estimators against simple estimators like LDA to gage if the performance is worth the penalty in speed. This was one of the reasons we decided to add single decision trees even though they rarely out-perform black-box estimators.

If we decide to go this route, we can treat the transform method that projects the data onto lower dimensions as a "model understanding" util much like #1239 would do for decision trees. 

In short, we would need to performance test this to get a better sense of the trade-offs but I think there's precedent for adding estimators that sacrifice performance in favor of simplicity and speed.

evalml/pipelines/components/transformers/dimensionality_reduction/lda.py
eccabay: Scikit-learn throws its own error here, it would probably make sense to catch it ourselves first!

bchen1116 (approved):
Looks good! I left a few comments, but nothing blocking

dsherry:
> I'd vote to keep this if we make it an estimator instead of a transformer for the following reason:
> 
> * In the spectrum of interpretable to black box, LDA is on the interpretable side. It has a closed-form solution that rests on well-understood (maybe not by me 😆 ) statistical theory. Users may want to compare better performing black box estimators against simple estimators like LDA to gage if the performance is worth the penalty in speed. This was one of the reasons we decided to add single decision trees even though they rarely out-perform black-box estimators.
> 
> If we decide to go this route, we can treat the transform method that projects the data onto lower dimensions as a "model understanding" util much like #1239 would do for decision trees.
> 
> In short, we would need to performance test this to get a better sense of the trade-offs but I think there's precedent for adding estimators that sacrifice performance in favor of simplicity and speed.

@freddyaboulton this is a really great point. Both approaches can be useful. My suggestion: merge this PR to get the transformer in (not added to automl just yet), then file an issue to build an LDA classifier.

PR #1333, opened 2020-10-21
Add ClassImbalanceDataCheck to DefaultDataChecks

### LINKED ISSUE #1276: Add class imbalance data check to automl
**Background**
In #971 we [implemented a data check for detecting class imbalance](https://github.com/alteryx/evalml/pull/1135/files), but we didn't add it to `DefaultDataChecks`.

**Proposal**
* Add `ClassImbalanceDataCheck` to `DefaultDataChecks`
* It currently has a warning for <10% representation. We should also have that data check include an error if the number of instances of any class is less than `2 * n_cv_folds`, because if not we'll get modeling errors. Please read the discussion in #971.
* Let's make sure our test coverage for this includes both binary and multiclass cases, categorical/int dtypes.

**Testing**
A great dataset to test with is [the abalone dataset](https://www.kaggle.com/rodolfomendes/abalone-dataset), with target "Rings", because that target has a bunch of classes which are heavily imbalanced

```python
df['Rings'].value_counts()
```
outputs
```python
9     689
10    634
8     568
11    487
7     391
12    267
6     259
13    203
14    126
5     115
15    103
16     67
17     58
4      57
18     42
19     32
20     26
3      15
21     14
23      9
22      6
24      2
27      2
1       1
25      1
2       1
26      1
29      1
Name: Rings, dtype: int64
```

bchen1116:
fix #1276 

- Throw error when the number of instances of any label is less than `2 * num_cv_folds`
(We default to `num_cv_folds == 3` since that is what our data split techniques use, although we leave it as an input parameter for the user)
- Add ClassImbalanceDataCheck to DefaultDataChecks

** This datacheck is only used in DefaultDataChecks if the problem type is classification, not regression

### DIFF
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -1,23 +1,28 @@
 import pandas as pd
 
 from .data_check import DataCheck
-from .data_check_message import DataCheckWarning
+from .data_check_message import DataCheckError, DataCheckWarning
 
 
 class ClassImbalanceDataCheck(DataCheck):
     """Checks if any target labels are imbalanced beyond a threshold. Use for classification problems"""
 
-    def __init__(self, threshold=0.1):
-        """Check if any of the features are likely to be ID columns.
+    def __init__(self, threshold=0.1, num_cv_folds=3):
+        """Check if any of the target labels are imbalanced, or if the number of values for each target
+           are below 2 times the number of cv folds
 
         Arguments:
             threshold (float): The minimum threshold allowed for class imbalance before a warning is raised.
                 A perfectly balanced dataset would have a threshold of (1/n_classes), ie 0.50 for binary classes.
                 Defaults to 0.10
+            num_cv_folds (int): The number of cross-validation folds. Must be positive. Choose 0 to ignore this warning.
         """
         if threshold <= 0 or threshold > 0.5:
             raise ValueError("Provided threshold {} is not within the range (0, 0.5]".format(threshold))
         self.threshold = threshold
+        if num_cv_folds < 0:
+            raise ValueError("Provided number of CV folds {} is less than 0".format(num_cv_folds))
+        self.cv_folds = num_cv_folds * 2
 
     def validate(self, X, y):
         """Checks if any target labels are imbalanced beyond a threshold for binary and multiclass problems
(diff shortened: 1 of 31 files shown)


bchen1116:
Using this datacheck on the Abalone dataset (with y = abalone['Rings']), we get the following datacheck error messages:
Screen Shot 2020-10-22 at 11 15 08 AM

We see the datacheck works as expected on this imbalanced dataset

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -28,17 +33,25 @@ def validate(self, X, y):
...
             y = pd.Series(y)
         messages = []
+        fold_counts = y.value_counts(normalize=False)
+        # search for targets that occur less than twice the number of cv folds first
+        below_threshold_folds = fold_counts.where(fold_counts < self.cv_folds).dropna()
+        if len(below_threshold_folds):
+            warning_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
+            messages.append(DataCheckError(warning_msg.format(self.cv_folds, below_threshold_folds.index.tolist()), self.name))
+
         counts = y.value_counts(normalize=True)
```
freddyaboulton, line 55: Not-blocking: I think we can set `counts = fold_counts/fold_counts.sum()`?

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -28,17 +33,25 @@ def validate(self, X, y):
...
+            >>> assert target_check.validate(X, y) == [DataCheckError("The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "ClassImbalanceDataCheck"), DataCheckWarning("The following labels fall below 10% of the target: [0]", "ClassImbalanceDataCheck")]
         """
         if not isinstance(y, pd.Series):
             y = pd.Series(y)
         messages = []
+        fold_counts = y.value_counts(normalize=False)
+        # search for targets that occur less than twice the number of cv folds first
+        below_threshold_folds = fold_counts.where(fold_counts < self.cv_folds).dropna()
+        if len(below_threshold_folds):
+            warning_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
```
freddyaboulton, line 52: Helpful error message! But it might be confusing if the user passes in a custom 4 fold CV and see `2 * number of cross folds = 6` in the message. Is there a plan to modify AutoMLSearch to init this class with the number of folds in the cv? Or maybe we make this more generic like "The number of instances of these targets is too small for cross validation? Just thinking out loud - this is fine to merge and we can refine if we get user feedback 😅

freddyaboulton (approved):
@bchen1116 Looks good!

evalml/data_checks/class_imbalance_data_check.py
bchen1116: @freddyaboulton Mmmm that's fair, I didn't think about that. I think it would make sense to modify such that we can pass `num_folds` arg from AutoMLSearch to the ClassImbalanceDataCheck itself. For instance, if the user passed in a custom data-splitter with > 6 folds, this data check now does nothing, and if they use the default data checks, it could pass that while failing to catch the error. I think that can be filed as a separate issue, unless it's better to implement it into this PR?

bchen1116: @freddyaboulton filed that as an issue [here](https://github.com/alteryx/evalml/issues/1345). I'll merge this PR and we can address the `num_folds` arg in AutoMLSearch afterwards

dsherry: Good point @freddyaboulton and thank you for filing @bchen1116 ! Yep sounds good. I agree this message can be improved.

PR #1334, opened 2020-10-21
Add codecov yaml, round coverage up instead of down

dsherry:
I think this will help with some of the codecov precision issues we've been seeing lately. If not, we can continue to tweak.

[The default codecov.io precision is 2](https://docs.codecov.io/docs/codecov-yaml#default-yaml) -- this just changes to [round](https://docs.codecov.io/docs/codecovyml-reference#coverageround) up instead of down when computing the codecov percentage.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,3 @@
+coverage:
+  round: up
+  precision: 2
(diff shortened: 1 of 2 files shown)


angela97lin (approved):
Yay, nice! Looks like the default precision is 2 (https://docs.codecov.io/docs/codecov-yaml), but not sure if we have to override everything once we want to customize. Either way, good to specify! LGTM ⚓

dsherry:
@angela97lin thanks! Yep I figured it was better to be explicit about that, for once we forget that we added this 😂

freddyaboulton (approved):
Thanks @dsherry !

PR #1335, opened 2020-10-21
Include editable installs in all install make targets

dsherry:
Our contributing guide asks people to run `make installdeps-dev`, but that doesn't include `pip install -e .`, which is important! This gave @rpeck some trouble getting the dev env working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,6 @@ commands:
       - run: |
             virtualenv test_python -q
             source test_python/bin/activate
-            make installdeps
             make installdeps-dev
   install_dependencies_core:
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -34,7 +34,6 @@ commands:
       - run: |
             virtualenv test_python -q
             source test_python/bin/activate
-            make installdeps
```
dsherry, line 37: No longer necessary now that `make installdeps-dev` does editable install

angela97lin (approved):
LGTM :shipit:

jeremyliweishih (approved):

bchen1116 (approved):
LGTM!

PR #1337, opened 2020-10-22
Updated _evaluate_pipelines to consolidate side effects

### LINKED ISSUE #1295: Refactor AutoML Search for Parallel Workers
- [ ] Define an evaluation method (name can be whatever) which takes pipeline+data input, outputs scores/results/CV data etc., with no side-effects
- [ ] To do this: update `_evaluate` to call `_add_result` afterwards instead of inside that method, so that _evaluate has no side effects.
- [ ] May have to update `_compute_cv_scores` too

christopherbunn:
Taking on a different approach with `_evaluate` side effects by restructuring and renaming to `_evaluate_pipelines`. See [this comment](https://github.com/alteryx/evalml/issues/1295#issuecomment-713820825) for more info. 
Resolves #1295

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -727,7 +727,8 @@ class AutoMLSearch:
                     desc = desc[:self._MAX_NAME_LEN - 3] + "..."
                 desc = desc.ljust(self._MAX_NAME_LEN)
 
-                update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations, self._start)
+                update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations,
+                                self._start, 1 if baseline else self._automl_algorithm.batch_number)
 
                 evaluation_results = self._compute_cv_scores(pipeline, X, y)
                 parameters = pipeline.parameters
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -370,9 +370,6 @@ class AutoMLSearch:
                 search begins. If "disabled" or None, no data checks will be done.
                 If set to "auto", DefaultDataChecks will be done. Default value is set to "auto".
 
-            feature_types (list, optional): list of feature types, either numerical or categorical.
-                Categorical features will automatically be encoded
-
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
         """
@@ -762,7 +759,7 @@ class AutoMLSearch:
 
             except KeyboardInterrupt:
                 current_pipeline_batch = self._handle_keyboard_interrupt(pipeline, current_pipeline_batch)
-                if current_pipeline_batch == []:
+                if len(current_pipeline_batch) == 0:
                     return current_batch_pipeline_scores
 
         return current_batch_pipeline_scores
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -42,7 +42,8 @@ from evalml.pipelines import (
     BinaryClassificationPipeline,
     MeanBaselineRegressionPipeline,
     ModeBaselineBinaryPipeline,
-    ModeBaselineMulticlassPipeline
+    ModeBaselineMulticlassPipeline,
+    PipelineBase
 )
 from evalml.pipelines.components.utils import get_estimators
 from evalml.pipelines.utils import make_pipeline
@@ -368,7 +369,7 @@ class AutoMLSearch:
 
             data_checks (DataChecks, list(Datacheck), str, None): A collection of data checks to run before
                 automl search. If data checks produce any errors, an exception will be thrown before the
-                search begins. If "disabled" or None, no data checks will be done.
+                search begins. If "disabled" or None, `no `data checks will be done.
                 If set to "auto", DefaultDataChecks will be done. Default value is set to "auto".
 
         Returns:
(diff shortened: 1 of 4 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -1058,32 +1058,40 @@ def test_catch_keyboard_interrupt(mock_fit, mock_score, mock_input,
...
+@patch('evalml.automl.AutoMLSearch._evaluate_pipelines')
+def test_pipelines_in_batch_return_nan(mock_evaluate_pipelines, mock_next_batch, X_y_binary, dummy_binary_pipeline_class):
     X, y = X_y_binary
-    mock_evaluate.side_effect = [{'cv_score_mean': 0}, {'cv_score_mean': np.nan}]
+    mock_evaluate_pipelines.return_value = [0, np.nan]
     mock_next_batch.side_effect = [[dummy_binary_pipeline_class(parameters={}), dummy_binary_pipeline_class(parameters={})]]
     automl = AutoMLSearch(problem_type='binary', allowed_pipelines=[dummy_binary_pipeline_class])
-    automl.search(X, y)
+    # Mock rankings so `best_pipeline` setting does not error out
+    with patch('evalml.automl.AutoMLSearch.rankings', new_callable=PropertyMock) as mock_rankings:
```
freddyaboulton, line 1068: Why did you change `mock_evaluate_pipelines` from `.side_effect` to `.return_value`?

So we need this this new PropertyMock because you're mocking `_evaluate_pipelines` which is what calls `_add_result` (which is what populates the data for the ranking table)? 

I wonder if we even need this code block in the test. I think the point of the test is to verify that the search ends when it encounters a batch of all nan (which is what happens in the code block that doesn't have the property mock). 

This would be fine to merge as is but I'm guessing we can simplify this test a bit without losing any coverage.

evalml/automl/automl_search.py
```diff
@@ -748,18 +702,64 @@ def _add_result(self, trained_pipeline, parameters, training_time, cv_data, cv_s
...
+                if not baseline and not add_single_pipeline:
+                    self._automl_algorithm.add_result(score_to_minimize, pipeline)
+
+                if search_iteration_plot:
+                    search_iteration_plot.update()
+
+                if add_single_pipeline:
+                    add_single_pipeline = False
+
+            except KeyboardInterrupt:
```
freddyaboulton, line 757: From what I can tell there isn't any change to the keyboard interrupt feature! What do you think @christopherbunn ?

freddyaboulton (approved):
@christopherbunn Looks great!

evalml/tests/automl_tests/test_automl.py
christopherbunn: > Why did you change mock_evaluate_pipelines from .side_effect to .return_value?

Good question. I changed it to `.return_value` as I needed the mock `_evaluate_pipelines` to return the entire list of pipeline scores. Passing in a list to `.side_effect` cause it to iterate through the list and only return each element. 

RE: the code block, you're right about the reason why we have to populate the ranking table manually. My impression of the intent of this section was to show that having a `np.nan` score as one of the pipeline results won't result in raising the AutoMLSearchException. If this isn't a necessary check, then I'm good with cutting out this section.

evalml/automl/automl_search.py
christopherbunn: I think technically if the user terminates the search while the next batch is being generated, it wouldn't get caught by this KeyboardInterrupt. In practice, getting the next batch takes so little time it's very unlikely that this will occur.

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Maybe we can delete that code block then? The second code block returns a Nan in the second batch and the search doesn't terminate then but we can check that explicitly with `assert mock_evaluate_pipelines.call_count == 3`.

This isn't blocking merge but would be nice to remove code that provides redundant coverage.

evalml/automl/automl_search.py
```diff
@@ -588,33 +569,9 @@ def _add_baseline_pipelines(self, X, y):
...
-                                 training_time=baseline_results['training_time'],
-                                 cv_data=baseline_results['cv_data'],
-                                 cv_scores=baseline_results['cv_scores'])
-            except KeyboardInterrupt:
-                pipelines = self._handle_keyboard_interrupt(baseline, pipelines)
-                if not pipelines:
-                    return True
-
+        scores = self._evaluate_pipelines(pipelines, X, y, baseline=True)
+        if scores == []:
```
dsherry, line 573: `return len(scores) > 0` ?

evalml/automl/automl_search.py
```diff
@@ -425,6 +427,7 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
         if self.max_batches and self.max_iterations is None:
+            self.show_batch_output = True
```
dsherry, line 430: @christopherbunn what's this? Why do we need it?

dsherry (approved):
@christopherbunn LGTM!

I left one question about `show_batch_output`. I also left a comment about deleting an old docstring. Otherwise, nothing blocking

evalml/automl/automl_search.py
christopherbunn: Currently, we show the batch number only if the user specifies a number of `max_batches`. Since we use batching internally even if only `max_iterations` is specified, there isn't really a clean way to infer whether or not we want to show the number of batches other than setting a variable at the beginning.

dsherry:
@christopherbunn and I saw intermittent failures in [the linux CI tests on his branch](https://app.circleci.com/pipelines/github/alteryx/evalml?branch=1295_%60_evaluate%60_changes). Since they were coming from pipeline tests and this PR only changes automl code, we conclude those failures aren't introduced by this PR. Will keep debugging.

PR #1342, opened 2020-10-23
Hide LightGBM warnings

### LINKED ISSUE #1330: LightGBM emits warnings for subsample_freq and subsample
Following the bug fix from https://github.com/alteryx/evalml/pull/1302 (thanks @bchen1116!), I ran the perf tests again for #1253. LightGBM now emits warnings about subsample_freq and subsample:

image

bchen1116:
fix #1330


evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -40,21 +40,23 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
...
-                      "bagging_fraction": bagging_fraction}
+                      "n_jobs": n_jobs}
         parameters.update(kwargs)
-
+        lg_parameters = copy.copy(parameters)
         # when boosting type is random forest (rf), LightGBM requires bagging_freq == 1 and  0 < bagging_fraction < 1.0
-        if boosting_type == "rf" and not bagging_freq:
-            parameters.update({'bagging_freq': 1})
+        if boosting_type == "rf":
+            # bagging_fraction could be passed in as a kwarg, which we wouldn't want to overwrite
```
freddyaboulton, line 48: I think we should override bagging_fraction if it's equal to 1 or 0 or else we end up back at #1251 ?

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -40,21 +40,23 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
...
+            if 'bagging_fraction' not in lg_parameters.keys():
```
freddyaboulton, line 49: What you have now is great but I thought this was easier to understand
```python
updates = {'bagging_freq': 1, 'subsample': None, 'subsample_freq': None}
if 'bagging_fraction' not in lg_parameters:
    updates['bagging_fraction'] = 0.9
lg_parameters.update(updates)
```

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -30,7 +30,7 @@ class LightGBMClassifier(Estimator):
     SEED_MIN = 0
     SEED_MAX = SEED_BOUNDS.max_bound
 
-    def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, max_depth=0, num_leaves=31, min_child_samples=20, n_jobs=-1, random_state=0, bagging_fraction=0.9, bagging_freq=0, **kwargs):
+    def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, max_depth=0, num_leaves=31, min_child_samples=20, n_jobs=-1, random_state=0, **kwargs):
```
freddyaboulton, line 33: How come we're getting rid of the defaults for these parameters? I think the warnings would still be suppressed if they're included.

freddyaboulton (approved):
@bchen1116 Looks great ! I left a couple of questions.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: @freddyaboulton yea you're right, i think. If I just override `subsample` and `subsample_freq` as None, it should work

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -44,17 +44,19 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
                       "bagging_freq": bagging_freq,
                       "bagging_fraction": bagging_fraction}
         parameters.update(kwargs)
-
+        lg_parameters = copy.copy(parameters)
+        updates = {'subsample': None, 'subsample_freq': None}
```
dsherry, line 48: Wait, so do these need to be `None` if `boosting_type != "rf"? As it stands, they will always be `None` going into lightgbm.

If not:
```python
if boosting_type == "rf":
    lg_parameters.update({'subsample': None, 'subsample_freq': None, 'bagging_freq': 1})
```
And pass that to lightgbm.

Either way, it would be nice to have this impl use 2 dicts instead of 3, harder to make buggy in the future :)

dsherry (approved):
@bchen1116 great! Two considerations:

1. Can we add unit test coverage which expects that there's no warning raised? For reference I think we do this in a few places in our tests, including [in the elasticnet classifier tests](https://github.com/alteryx/evalml/blob/main/evalml/tests/component_tests/test_en_classifier.py#L107)

2. Were we seeing this warning in *any* usage of our lightgbm estimator? Or was it just for specific components? If its the latter, it would be great to have unit test coverage with a set of parameters which previously resulted in a warning.

I also left a comment on the impl

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: If we pass `bagging_freq` or `bagging_fraction` into LightGBM, we need to set `subsample` to None to avoid the warning messages

dsherry: Ah got it. I see their docs say subsample and subsample_freq are aliases for  bagging_fraction and bagging_freq.

In that case what about
```python
if boosting_type == "rf":
    lg_parameters.update({'bagging_freq': 1})
# avoid lightgbm warnings having to do with parameter aliases
if bagging_freq is not None or bagging_fraction is not None:
    lg_parameters.update({'subsample': None, 'subsample_freq': None})
```
or something along those lines?

PR #1346, opened 2020-10-26
Fix 'Invalid Value Encountered' error during AutomLSearch

### LINKED ISSUE #1343: LightGBM "invalid value encountered" warning during fit
Repro: load breast cancer dataset and run automl. As is currently visible [on the start page in the docs](https://evalml.alteryx.com/en/latest/start.html) 😅
> invalid value encountered in double_scalars

Tagging @bchen1116 as this could be related to recent lightgbm changes

bchen1116:
Fix #1343 

This is a highVarianceDataCheck error, rather than a LightGBM Error

Old docs [here](https://evalml.alteryx.com/en/latest/start.html) and new docs [here](https://evalml.alteryx.com/en/bc_1343_value/start.html)

Note: The LightGBM warning messages are being addressed in [this pr](https://github.com/alteryx/evalml/pull/1342)

### DIFF
--- a/evalml/data_checks/high_variance_cv_data_check.py
+++ b/evalml/data_checks/high_variance_cv_data_check.py
@@ -37,7 +37,7 @@ class HighVarianceCVDataCheck(DataCheck):
             cv_scores = pd.Series(cv_scores)
 
         messages = []
-        if cv_scores.mean() == 0 and cv_scores.std() == 0:
+        if cv_scores.mean() == 0:
             high_variance_cv = 0
         else:
             high_variance_cv = abs(cv_scores.std() / cv_scores.mean()) > self.threshold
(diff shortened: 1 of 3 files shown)


evalml/data_checks/high_variance_cv_data_check.py
```diff
@@ -37,7 +37,10 @@ def validate(self, pipeline_name, cv_scores):
             cv_scores = pd.Series(cv_scores)
 
         messages = []
-        high_variance_cv = abs(cv_scores.std() / cv_scores.mean()) > self.threshold
+        if cv_scores.mean() == 0 and cv_scores.std() == 0:
```
jeremyliweishih, line 40: Would this work if we only check if the mean is equal to 0?

jeremyliweishih (approved):
lgtm

evalml/data_checks/high_variance_cv_data_check.py
bchen1116: mm yeah I think so. Since these are scores, if the mean is 0, the std is also 0. Thanks for pointing that out!

PR #1349, opened 2020-10-26
Modifying ohe get_feature_names so encoded columns are always unique

### LINKED ISSUE #1298: OHE doesn't always result in unique names
```
import pandas as pd
df = pd.DataFrame({"A": ["x_y"] , "A_x": ["y"]})
from evalml.pipelines.components import OneHotEncoder
OneHotEncoder().fit_transform(df)
```
gives
```
   A_x_y  A_x_y
0    1.0    1.0
```

freddyaboulton:
### Pull Request Description
Fixes #1298

-----

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import numpy as np
 import pandas as pd
 from sklearn.preprocessing import OneHotEncoder as SKOneHotEncoder
@@ -184,6 +185,16 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
             raise ValueError(f'Feature "{feature_name}" was not provided to one-hot encoder as a training feature')
         return self._encoder.categories_[index]
 
+    @staticmethod
+    def _make_name_unique(name, seen_before):
+        if name not in seen_before:
+            return name
+        i = 1
+        while name in seen_before:
+            name = f"{name}_{i}"
+            i += 1
+        return name
+
     def get_feature_names(self):
         """Return feature names for the input features after fitting.
 
@@ -201,11 +212,10 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
                     continue
 
                 # Follow sklearn naming convention but if name has been seen before
-                # then add "1" to make it unique
+                # then add an int to make it unique
                 proposed_name = f"{col}_{category}"
-
                 if proposed_name in seen_before:
-                    proposed_name = f"{col}_{category}_1"
+                    proposed_name = self._make_name_unique(proposed_name, seen_before)
                 unique_names.append(proposed_name)
                 seen_before.add(proposed_name)
         return unique_names
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -187,8 +187,9 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
     def _make_name_unique(name, seen_before):
         """Helper to make the name unique."""
         i = 1
+        name = f"{name}_{i}"
         while name in seen_before:
-            name = f"{name}_{i}"
+            name = f"{name[:name.rindex('_')]}_{i}"
             i += 1
         return name
 
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 import numpy as np
 import pandas as pd
 from sklearn.preprocessing import OneHotEncoder as SKOneHotEncoder
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -186,8 +186,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
 
     @staticmethod
     def _make_name_unique(name, seen_before):
-        if name not in seen_before:
-            return name
+        """Helper to make the name unique."""
         i = 1
         while name in seen_before:
             name = f"{name}_{i}"
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -72,7 +72,6 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
                          component_obj=None,
                          random_state=random_state)
         self._initial_state = self.random_state.get_state()
-        self.drop = drop
 
     @staticmethod
     def _get_cat_cols(X):
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -186,6 +186,11 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
     @staticmethod
     def _make_name_unique(name, seen_before):
         """Helper to make the name unique."""
+
+        if name not in seen_before:
+            return name
+
+        # Only modify the name if it has been seen before
         i = 1
         name = f"{name}_{i}"
         while name in seen_before:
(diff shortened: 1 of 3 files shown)


evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -180,10 +179,37 @@ def categories(self, feature_name):
...
+        return name
+
     def get_feature_names(self):
         """Return feature names for the input features after fitting.
 
         Returns:
             np.array: The feature names after encoding, provided in the same order as input_features.
         """
-        return self._encoder.get_feature_names(self.features_to_encode)
+        unique_names = []
```
freddyaboulton, line 197: Tested this on a df with two columns with 10,000 unique categories and top_n set to None. `timeit` reports a mean time of 5.75 ms which I think is acceptable. Especially since we usually pass in top_n and name clashes are pretty rare in my experience.

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -122,6 +122,31 @@ def test_drop():
     assert col_names == expected_col_names
 
 
+def test_drop_binary():
```
freddyaboulton, line 125: Noticed we didn't have coverage for all accepted values of drop.

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -441,3 +466,13 @@ def check_df_equality(random_state):
...
+
+
+def test_ohe_column_names_unique():
+    df = pd.DataFrame({"A": ["x_y"], "A_x": ["y"]})
+    df_transformed = OneHotEncoder().fit_transform(df)
+    assert set(df_transformed.columns) == {"A_x_y", "A_x_y_1"}
+
+    df = pd.DataFrame({"A": ["x_y", "z"], "A_x": ["y_1", "y"], "A_x_y": ["1", "y"]})
+    df_transformed = OneHotEncoder().fit_transform(df)
+    assert set(df_transformed.columns) == {"A_x_y", "A_z", "A_x_y_1", "A_x_y_1_1", "A_x_y_1_1_2", "A_x_y_y"}
```
bchen1116, line 478: This looks kinda wild, although I guess this instance shouldn't occur very often for a user.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -180,10 +178,37 @@ def categories(self, feature_name):
             raise ValueError(f'Feature "{feature_name}" was not provided to one-hot encoder as a training feature')
         return self._encoder.categories_[index]
 
+    @staticmethod
+    def _make_name_unique(name, seen_before):
+        """Helper to make the name unique."""
+        i = 1
+        while name in seen_before:
+            name = f"{name}_{i}"
```
bchen1116, line 186: Could we do something like 
```python
i = 1
name = f"{name}_{i}"
while name in seen_before:
    i += 1
    name = f"{name[:name.rindex("_")]}_{i}"
```
It's messier, but I think it makes it cleaner from a user standpoint to see
```
transformed.columns = ['X_y_1', 'X_y_2', 'X_y_3']
```
versus 
```
transformed.columns = ['X_y_1', 'X_y_1_2', 'X_y_1_2_3']
```

bchen1116 (approved):
LGTM! I left a comment on potential styling but looks good either way!

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
freddyaboulton: Great suggestion! I am not sure there is a perfect solution here but I think this is nicer than what I originally had hehe

evalml/tests/component_tests/test_one_hot_encoder.py
freddyaboulton: Yea I don't think there is a perfect solution here in the sense that any solution will slightly obfuscate the column name/category level that results in each column in the transformed dataframe (you can create an adversarial example no matter the format we use to name the new columns).

That being said, I think it's important to have unique names to avoid any potential downstream bugs that can arise from having duplicate column names. 

Maybe we can better explain how the columns are named in the docs (and explain what happens when there are collisions) so users can better trace the data through the pipeline if they are debugging.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -180,10 +178,38 @@ def categories(self, feature_name):
...
+                    if cat_index == self._encoder.drop_idx_[col_index]:
+                        continue
+
+                # Follow sklearn naming convention but if name has been seen before
+                # then add an int to make it unique
+                proposed_name = f"{col}_{category}"
+                if proposed_name in seen_before:
+                    proposed_name = self._make_name_unique(proposed_name, seen_before)
+                unique_names.append(proposed_name)
+                seen_before.add(proposed_name)
```
eccabay, line 214: Is it necessary to have these be two separate variables? It looks like they store the same information

eccabay (approved):
LGTM! Left one comment about a redundant variable

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
freddyaboulton: Good question! The `_make_name_unique` needs to repeatedly check whether we have already seen a name so a set is better than a list for that I think. The other option is to create a set from `unique_names` inside `_make_name_unique` but I figured we might as well just keep one set around at that point. What do you think? 

We definitely need to return a list because the order of the columns matters.

eccabay: I see, that makes a lot of sense. Totally fine with keeping both!

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -180,10 +178,38 @@ def categories(self, feature_name):
...
+                # Drop categories specified by the user
+                if self._encoder.drop_idx_ is not None and self._encoder.drop_idx_[col_index] is not None:
+                    if cat_index == self._encoder.drop_idx_[col_index]:
+                        continue
+
+                # Follow sklearn naming convention but if name has been seen before
+                # then add an int to make it unique
+                proposed_name = f"{col}_{category}"
+                if proposed_name in seen_before:
+                    proposed_name = self._make_name_unique(proposed_name, seen_before)
```
angela97lin, line 212: Really nit-picky comment but we can keep the logic of making a unique name in `make_name_unique()` by not checking here, and doing so in the helper instead. What I mean is here, 
```
proposed_name = self._make_name_unique(f"{col}_{category}", seen_before)
```
And then in `_make_name_unique`:
```
    def _make_name_unique(name, seen_before):
        """Helper to make the name unique."""
        i = 1
        while name in seen_before:
            name = f"{name[:name.rindex('_')]}_{i}"
            i += 1
        return name
```

Something like that? :D

angela97lin (approved):
Really awesome stuff! LGTM, left one really nit-picky comment on how we could cut down on checking for whether or not we need to make a unique name or not :)

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
freddyaboulton: Good suggestion! I had to add some minor modifications but the loop within `get_feature_names` is easier to read now.

angela97lin (commented):
Ah, one more thing: it would be good to document this behavior somewhere, so that a user who uses the OHE isn't confused ("What is A_x_y_1_2? Is it encoding col A_x_y_1 value 2? Or category 1 in A_x_y?" Docstring update is sufficent imo :)

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -180,10 +178,48 @@ def categories(self, feature_name):
...
+
     def get_feature_names(self):
-        """Return feature names for the input features after fitting.
+        """Return feature names for the categorical features after fitting.
+
+        Feature names are formatted as {column name}_{category name}. In the event of a duplicate name,
+        an integer will be added at the end of the feature name to distinguish it.
+
+        For example, consider a dataframe with a column called "A" and category "x_y" and another column
+        called "A_x" with "y". In this example, the feature names would be "A_x_y" and "A_x_y_1".
```
dsherry, line 203: Edge case: does this fix address the case where you have a feature "A_x" with category "y" and a feature "A" with category "x_y"?

freddyaboulton: I think this is the first case in `test_ohe_column_names_unique` but let me know if I'm misunderstanding!

dsherry: Cool!

PR #1352, opened 2020-10-27
Warning on null values for partial_dependency

### LINKED ISSUE #1285: Partial Dependency Fails when feature has nans
I would like to generate a Partial Dependency for 'log_tel'. However, log_tel has nans:
`df['log_tel'].isnull().sum()`
`205`

I am getting the following Partial Dependency result:
image

Next, I filled the nans with zero and verified the result:
`df['log_tel'] = df['log_tel'].fillna(0)`
`df['log_tel'].isnull().sum()`
`0`

Now, the Partial Dependency works properly:
image

bchen1116:
fix #1285 

Warns user when there are null values in the features of interest.


evalml/model_understanding/graphs.py
```diff
@@ -436,6 +436,12 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
     elif isinstance(pipeline, evalml.pipelines.RegressionPipeline):
         pipeline._estimator_type = "regressor"
     pipeline.feature_importances_ = pipeline.feature_importance
+    if isinstance(feature, int):
```
freddyaboulton, line 439: I think NaNs would still be an issue if they are present in another feature and not in the feature of interest for the partial dependence. I wonder if it's better to issue the warning if the pipeline does not have an imputer (something like `Imputer not present in pipeline, partial_dependence may have NaN values if NaN values are present in the data`)? 

I'm guessing NaNs would throw off 95% of our features and it's not sustainable to independently check for NaNs in all places. I think it's on the user to add an imputer to the pipeline if nans are possible in their data. Maybe the question should be whether or not that's communicated well in the docs. What do you think?

freddyaboulton (approved):
@bchen1116 Looks good to me!

evalml/model_understanding/graphs.py
```diff
@@ -436,6 +436,12 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
+        if X.iloc[:, feature].isnull().sum():
+            warnings.warn("There are null values in the features, which will cause NaN values in the partial dependency. Fill in these values to remove the NaN values.")
+    if isinstance(feature, str):
+        if X[feature].isnull().sum():
+            warnings.warn("There are null values in the features, which will cause NaN values in the partial dependency. Fill in these values to remove the NaN values.")
```
dsherry, line 444: Looks good! A few nit-picks:
1. Please declare a warning type! We don't have this pattern in place yet, but: add a [subclass `UserWarning`](https://docs.python.org/3/library/exceptions.html#UserWarning) in `evalml/exceptions/exceptions.py`, add a docstring there, and then include that class here in the call to `warn`, after the message. Why do this: allows users to apply the `warnings` filters [based on category](https://docs.python.org/3/library/warnings.html#overriding-the-default-filter) as well as message contents. We just ran into this same consideration with the text / datatables work, in fact!
2. "NaN values in the partial dependency" --> "NaN values in the partial dependence output" ?
3. style:
```python
if ((isinstance(feature, int) and X.iloc[:, feature].isnull().sum()) or
    (isinstance(feature, str) and X[feature].isnull().sum())):
    warnings.warn(...)
```

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -816,6 +816,18 @@ def test_partial_dependence_not_fitted(X_y_binary, logistic_regression_binary_pi
...
+def test_partial_dependence_warning(logistic_regression_binary_pipeline_class):
+    X = pd.DataFrame({'a': [2, None, 2, 2], 'b': [1, 2, 2, 1]})
+    y = pd.Series([0, 1, 0, 1])
+    pipeline = logistic_regression_binary_pipeline_class(parameters={})
+    pipeline.fit(X, y)
+    with pytest.warns(UserWarning, match="There are null values in the features, which will cause NaN values in the partial dependency"):
+        partial_dependence(pipeline, X, feature=0, grid_resolution=20)
+
+    with pytest.warns(UserWarning, match="There are null values in the features, which will cause NaN values in the partial dependency"):
+        partial_dependence(pipeline, X, feature='a', grid_resolution=20)
```
dsherry, line 828: Great

dsherry (approved):
@bchen1116 awesome, left a couple comments but other than that pls 🚢 !

PR #1358, opened 2020-10-28
Removed pandas warning when converting to Woodwork in search()

### LINKED ISSUE #1344: Pandas warning appears when using breast cancer demo dataset
Likely related to the internal conversion of input data to Woodwork tables. 
 
image

christopherbunn:
Three possible solutions: 
1. Use `copy_datatable=True` in setting the Woodwork table internally. This might have a degradation in the speed of `search()` since it's copying the entire data table, will need to run performance testing to verify. Current route that this PR goes with. 
2. [Disable the warning in pandas](https://stackoverflow.com/questions/20625582/how-to-deal-with-settingwithcopywarning-in-pandas/20627316#20627316). Another possible solution, but I'm hesitant to disable error messages just to remove the out. 
3.  Change [evalml/preprocessing/utils.py](https://github.com/alteryx/evalml/blob/4c9e5a99d420213aac21ab63a1459a626cb326bc/evalml/preprocessing/utils.py#L68-L69) to use `.loc` instead of `.iloc`. This may cause issues with indexes that are not sequential integer values (such as strings) and unique numbers but aren't zero-indexed. 

Open to hearing why we should or should not go with another approach that isn't option 1. 

Resolves #1344

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -29,6 +29,7 @@ Release Notes
         * Fix warnings thrown during AutoMLSearch in ``HighVarianceCVDataCheck`` :pr:`1346`
         * Fixed bug where TrainingValidationSplit would return invalid location indices for dataframes with a custom index :pr:`1348`
         * Fixed bug where the AutoMLSearch ``random_state`` was not being passed to the created pipelines :pr:`1321`
+        * Removed pandas warning when converting to Woodwork in ``search()`` :pr:`1358`
     * Changes
         * Allow ``add_to_rankings`` to be called before AutoMLSearch is called :pr:`1250`
         * Removed Graphviz from test-requirements to add to requirements.txt :pr:`1327`
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -386,7 +386,7 @@ class AutoMLSearch:
             logger.warning("`X` passed was not a DataTable. EvalML will try to convert the input as a Woodwork DataTable and types will be inferred. To control this behavior, please pass in a Woodwork DataTable instead.")
             if isinstance(X, np.ndarray):
                 X = pd.DataFrame(X)
-            X = ww.DataTable(X)
+            X = ww.DataTable(X, copy_dataframe=True)
 
         text_column_vals = X.select('natural_language')
         text_columns = list(text_column_vals.to_pandas().columns)


docs/source/release_notes.rst
```diff
@@ -29,6 +29,7 @@ Release Notes
         * Fix warnings thrown during AutoMLSearch in ``HighVarianceCVDataCheck`` :pr:`1346`
         * Fixed bug where TrainingValidationSplit would return invalid location indices for dataframes with a custom index :pr:`1348`
         * Fixed bug where the AutoMLSearch ``random_state`` was not being passed to the created pipelines :pr:`1321`
+        * Removed pandas warning when converting to Woodwork in ``search()`` :pr:`1358`
```
freddyaboulton, line 32: I can't reproduce this with pandas 1.1.1 and woodwork 0.0.3. 

I vote with trying to filter the warning rather than copying the dataframe if possible but this would be ok to merge since the woodwork team is fixing the issue.

thehomebrewnerd: @freddyaboulton I'm not sure there is a solution for this inside of Woodwork that does not involve copying the dataframe during DataTable initialization. There is an open issue for this that we can continue to investigate, but I'm not sure what can be done at this point.

Some additional comments can be found in the [Woodwork Issue](https://github.com/FeatureLabs/woodwork/issues/312#issuecomment-716601515)

christopherbunn:
Interesting. I am now currently unable to reproduce the error on today's main. After doing some digging, it turns out that a change in the [TextFeaturizer commit ](https://github.com/alteryx/evalml/commit/83343aa85b2bc9ec272b52c1151e35bce041def9) did something so that the error message does not appear. I am still able to replicate the error if I used the commit before and the pandas + woodwork versions that Freddy has.

dsherry:
@christopherbunn interesting! Do you know what change did it?

A few candidates at first glance:
* We did [a warning filter](https://github.com/alteryx/evalml/commit/83343aa85b2bc9ec272b52c1151e35bce041def9#diff-9031616fb13129be1f11715e1ce202570c56596aa70c945062380ab1972cd8feR22)
* [Added featuretools and nlp_primitives as core requirements](https://github.com/alteryx/evalml/commit/83343aa85b2bc9ec272b52c1151e35bce041def9#diff-e7031ce8aee6d7dc175631195661f5f893bfa3614e5f63ec93c15d2d59235667R15), maybe that changed our pandas version somehow

Happy to discuss further if you'd like

thehomebrewnerd:
Ok, this is confusing me as well. If I run this code I get the warning (note pandas only):
```
import pandas as pd

df = pd.DataFrame({
    'id': [0, 1, 2],
    'val1': [100, 200, 300],
    'val2': [10, 20, 30],
})
df_slice = df.iloc[[0, 1]]
new_vals = pd.Series([1000, 2000])
df_slice.loc[:, 'val1'] = new_vals
```

But, if I add an `import evalml` statement at the top, the warning goes away. I'm not sure what it is exactly, but it seem like something that is happening during `import evalml` is causing the warning to go away or be suppressed.

docs/source/release_notes.rst
freddyaboulton: Thanks for the context @thehomebrewnerd !

freddyaboulton:
@christopherbunn @thehomebrewnerd Looks like this line that got recently added to our top-level init file may have something to do with it?

`warnings.filterwarnings('ignore', 'The following selectors were not present in your DataTable')`

thehomebrewnerd:
A little more digging. Adding `import featuretools` instead of `import evalml` also causes the warning to disappear. I see that featuretools is now imported in `text_featurizer.py`, so maybe it is actually something in Featuretools that has caused the warning to disappear.

I get the warning even with the new `warnings.filterwarnings` line removed, and given I can replicate the behavior without evalml anywhere in the loop makes me think this is because caused by something else. Featuretools is my current guess, but I'm not sure what exactly...

thehomebrewnerd:
The suppression of this warning is being caused by Featuretools. This line in `entityset/entityset.py` is the source: https://github.com/FeatureLabs/featuretools/blob/eab027e10e9c89c906954273c7590b7714a4ccc0/featuretools/entityset/entityset.py#L24

freddyaboulton:
Great find!

christopherbunn:
Thanks @thehomebrewnerd! I'll close out this PR and issue then.

dsherry:
@thehomebrewnerd wow, thanks for digging that up! So, by including featuretools as a core dependency, we've chosen option 2 from @christopherbunn's great list in the PR description: disable the warning in pandas. Sounds good to me for the time being.

I do wonder if there's a pattern which is better for users here. Setting `pd.options.mode.chained_assignment = None` is a global configuration, meaning it applies outside of featuretools code. If someone is using featuretools as a third-party library, that means simply importing featuretools will change pandas behavior here. That seems undesirable. We have the same concern in evalml with [our warning handling](https://github.com/alteryx/evalml/blob/main/evalml/__init__.py#L16). Anyways, I'll file something separate to continue discussing this.

Thank you all!!

PR #1364, opened 2020-10-29
Don't add OneHotEncoder to pipeline for text features

dsherry:
**Problem**
Our `make_pipelines` code builds a pipeline based on the feature types present in the data. The [`_get_preprocessing_components` method](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L66) will currently add a one-hot encoder component if there are any features of type `object` or `category` in the pandas dataframe. However, in the case where the data includes text features but no categorical features, this is unnecessary.

**Fix**
Short-term (this PR): during pipeline creation, exclude text features when determining whether or not to add an encoder for categorical features.

Long-term: use woodwork to get categorical cols, not pandas! (#1229 ). The `make_pipelines` code should be given a woodwork datatable, and we should instead use woodwork's select method.

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -63,6 +63,7 @@ def _get_preprocessing_components(X, y, problem_type, text_columns, estimator_cl
 
     # DateTimeFeaturizer can create categorical columns
     categorical_cols = X.select_dtypes(include=categorical_dtypes)
+    categorical_cols.drop(text_columns, inplace=True, errors='ignore', axis=1)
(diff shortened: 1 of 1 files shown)


dsherry:
This can be closed! We [now use woodwork in `make_pipeline`](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L68) to select categorical features and to determine whether or not we need to add categorical encoding to the pipeline. Thanks @freddyaboulton for reminding me to close this :)

PR #1377, opened 2020-10-30
Change OutliersDataCheck to find outliers for columns 

### LINKED ISSUE #1313: Outlier detection data check: identify column, not row, with outliers
[Our outlier check](https://github.com/alteryx/evalml/blob/main/evalml/data_checks/outliers_data_check.py#L66) currently emits a warning for each row with an outlier in some feature. We should flip this to emit a warning for each feature (column) with an outlier in some row.

bchen1116:
fix #1313 

Return warnings for list of columns rather than rows

### DIFF
--- a/evalml/data_checks/outliers_data_check.py
+++ b/evalml/data_checks/outliers_data_check.py
@@ -49,20 +49,12 @@ class OutliersDataCheck(DataCheck):
             q1 = df.quantile(0.25)
             q3 = df.quantile(0.75)
             iqr = q3 - q1
-            lower_bound = q1 - (k * iqr)
-            upper_bound = q3 + (k * iqr)
-            return (lower_bound, upper_bound)
+            lower_bound = pd.Series(q1 - (k * iqr), name='lower_bound')
+            upper_bound = pd.Series(q3 + (k * iqr), name='upper_bound')
+            return pd.concat([lower_bound, upper_bound], axis=1)
 
-        lower_bound, upper_bound = get_IQR(X)
-        indices = set()
-        # get the columns that fall out of the bounds, which means they contain outliers
-        for idx, bound in enumerate([lower_bound, upper_bound]):
-            cols_in_range = (X >= bound.values) if idx == 0 else (X <= bound.values)
-            cols_in_range = cols_in_range.all()
-            outlier_cols = cols_in_range[~cols_in_range].keys()
-            indices.update(outlier_cols.tolist())
-        # order the columns by how they appear in the dataframe
-        indices = sorted(list(indices), key=lambda x: X.columns.tolist().index(x))
+        iqr = get_IQR(X, k=2.0)
+        has_outliers = ((X < iqr['lower_bound']) | (X > iqr['upper_bound'])).any()
         warning_msg = "Column '{}' is likely to have outlier data"
-        s = [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in indices]
-        return s
+        cols = has_outliers.index[has_outliers]
+        return [DataCheckWarning(warning_msg.format(col), self.name) for col in cols]
--- a/evalml/data_checks/outliers_data_check.py
+++ b/evalml/data_checks/outliers_data_check.py
@@ -1,5 +1,4 @@
 import pandas as pd
-from sklearn.ensemble import IsolationForest
 
 from .data_check import DataCheck
 from .data_check_message import DataCheckWarning
@@ -9,8 +8,7 @@ from evalml.utils.gen_utils import numeric_dtypes
 
 
 class OutliersDataCheck(DataCheck):
-    """Checks if there are any outliers in input data by using an Isolation Forest to obtain the anomaly score
-        of each index and then using IQR to determine score anomalies. Indices with score anomalies are considered outliers."""
+    """Checks if there are any outliers in input data by using IQR to determine score anomalies. Columns with score anomalies are considered to contain outliers."""
 
     def __init__(self, random_state=0):
         """Checks if there are any outliers in the input data.
@@ -21,24 +19,23 @@ class OutliersDataCheck(DataCheck):
         self.random_state = get_random_state(random_state)
 
     def validate(self, X, y=None):
-        """Checks if there are any outliers in a dataframe by using an Isolation Forest to obtain the anomaly score
-        of each index and then using IQR to determine score anomalies. Indices with score anomalies are considered outliers.
+        """Checks if there are any outliers in a dataframe by using IQR to determine column anomalies. Column with anomalies are considered to contain outliers.
 
         Arguments:
             X (pd.DataFrame): Features
             y: Ignored.
 
         Returns:
-            A set of indices that may have outlier data.
+            A set of columns that may have outlier data.
 
         Example:
             >>> df = pd.DataFrame({
-            ...     'x': [1, 2, 3, 40, 5],
-            ...     'y': [6, 7, 8, 990, 10],
+            ...     'x': [1, 2, 3, 4, 5],
+            ...     'y': [6, 7, 8, 9, 10],
             ...     'z': [-1, -2, -3, -1201, -4]
             ... })
             >>> outliers_check = OutliersDataCheck()
-            >>> assert outliers_check.validate(df) == [DataCheckWarning("Row '3' is likely to have outlier data", "OutliersDataCheck")]
+            >>> assert outliers_check.validate(df) == [DataCheckWarning("Column 'z' is likely to have outlier data", "OutliersDataCheck")]
         """
 
         if not isinstance(X, pd.DataFrame):
@@ -56,11 +53,16 @@ class OutliersDataCheck(DataCheck):
             upper_bound = q3 + (k * iqr)
             return (lower_bound, upper_bound)
 
-        clf = IsolationForest(random_state=self.random_state)
-        clf.fit(X)
-        scores = pd.Series(clf.decision_function(X))
-        lower_bound, upper_bound = get_IQR(scores, k=2)
-        outliers = (scores < lower_bound) | (scores > upper_bound)
-        outliers_indices = outliers[outliers].index.values.tolist()
-        warning_msg = "Row '{}' is likely to have outlier data"
-        return [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in outliers_indices]
+        lower_bound, upper_bound = get_IQR(X)
+        indices = set()
+        # get the columns that fall out of the bounds, which means they contain outliers
+        for idx, bound in enumerate([lower_bound, upper_bound]):
+            cols_in_range = (X >= bound.values) if idx == 0 else (X <= bound.values)
+            cols_in_range = cols_in_range.all()
+            outlier_cols = cols_in_range[~cols_in_range].keys()
+            indices.update(outlier_cols.tolist())
+        # order the columns by how they appear in the dataframe
+        indices = sorted(list(indices), key=lambda x: X.columns.tolist().index(x))
+        warning_msg = "Column '{}' is likely to have outlier data"
+        s = [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in indices]
+        return s
--- a/evalml/tests/data_checks_tests/test_outliers_data_check.py
+++ b/evalml/tests/data_checks_tests/test_outliers_data_check.py
@@ -19,16 +19,17 @@ def test_outliers_data_check_warnings():
     data = np.tile(a, (100, 10))
 
     X = pd.DataFrame(data=data)
-    X.iloc[3, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[25, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[55, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[72, :] = pd.Series(np.random.randn(100) * 1000)
+    X.iloc[0, 3] = 1000
+    X.iloc[3, 25] = 1000
+    X.iloc[5, 55] = 10000
+    X.iloc[10, 72] = -1000
+    X.iloc[:, 90] = 'string_values'
 
     outliers_check = OutliersDataCheck()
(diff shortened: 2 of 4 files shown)


evalml/data_checks/outliers_data_check.py
```diff
@@ -56,11 +55,15 @@ def get_IQR(df, k=2.0):
...
-        lower_bound, upper_bound = get_IQR(scores, k=2)
-        outliers = (scores < lower_bound) | (scores > upper_bound)
-        outliers_indices = outliers[outliers].index.values.tolist()
-        warning_msg = "Row '{}' is likely to have outlier data"
-        return [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in outliers_indices]
+        lower_bound, upper_bound = get_IQR(X)
+        indices = set()
+        # get the columns that fall out of the bounds, which means they contain outliers
+        for idx, bound in enumerate([lower_bound, upper_bound]):
+            boundary = ((X[bound.keys()] >= bound.values) if idx == 0 else (X[bound.keys()] <= bound.values))[bound.keys()].all()
```
freddyaboulton, line 62: I advocate for breaking this up into multiple lines to make it easier to understand how `boundary` is created. Also, I don't think you need to call `[bound.keys()]` everywhere since `X` was already filtered to the columns with numeric data types.

freddyaboulton, line 62: Before we were computing the IQR of the scores returned by an IsolationForest and now we are computing the IQR of the column values. Are those two things the same? 

One thing we can do is keep the isolation forest but then use SHAP to find the columns that contribute to the high/low scores. Here is a proof-of-concept based on one of the existing unit tests:


image

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -19,16 +19,15 @@ def test_outliers_data_check_warnings():
...
-                                          DataCheckWarning("Row '25' is likely to have outlier data", "OutliersDataCheck"),
-                                          DataCheckWarning("Row '55' is likely to have outlier data", "OutliersDataCheck"),
-                                          DataCheckWarning("Row '72' is likely to have outlier data", "OutliersDataCheck")]
+    assert outliers_check.validate(X) == [DataCheckWarning("Column '3' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '25' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '55' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '72' is likely to have outlier data", "OutliersDataCheck")]
 
 
 def test_outliers_data_check_input_formats():
```
freddyaboulton, line 33: Might be good to add a test (or modify one of these tests) to have a couple of columns with categorical/string values.

freddyaboulton (commented):
@bchen1116 This is great! I have a question about whether or not we should still keep the old behavior of using scores from an IsolationForest.

evalml/data_checks/outliers_data_check.py
```diff
@@ -56,11 +55,15 @@ def get_IQR(df, k=2.0):
...
-        scores = pd.Series(clf.decision_function(X))
-        lower_bound, upper_bound = get_IQR(scores, k=2)
-        outliers = (scores < lower_bound) | (scores > upper_bound)
-        outliers_indices = outliers[outliers].index.values.tolist()
-        warning_msg = "Row '{}' is likely to have outlier data"
-        return [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in outliers_indices]
+        lower_bound, upper_bound = get_IQR(X)
+        indices = set()
+        # get the columns that fall out of the bounds, which means they contain outliers
+        for idx, bound in enumerate([lower_bound, upper_bound]):
```
jeremyliweishih, line 61: This was a little hard to follow at first: maybe an approach like [this](https://stackoverflow.com/questions/40156469/how-to-check-if-any-value-of-a-column-is-in-a-range-in-between-two-values-in-p) (but check for outside of range and map to every column in the DF).

jeremyliweishih (approved):
Looks good! One comment for improvement.

evalml/data_checks/outliers_data_check.py
angela97lin: Don't think they're exactly the same; here's the post I had referenced while implementing via IsolationForest: https://towardsdatascience.com/isolation-forest-with-statistical-rules-4dd27dad2da9; that being said, I'm for keeping it simple via IQR only or exploring using SHAP--what @freddyaboulton outlined looks pretty cool 😲

angela97lin (commented):
Agree with @freddyaboulton's comment, could be cool to look into. Regardless of approach, let's update the docs too: https://evalml.alteryx.com/en/bc_1313_outlier/generated/evalml.data_checks.OutliersDataCheck.html?next=https%3A%2F%2Fevalml.alteryx.com%2Fen%2Fbc_1313_outlier%2Fgenerated%2Fevalml.data_checks.OutliersDataCheck.html&ticket=ST-1604333709-o5ChxGrdvwTlEh71ywOXKxHGtWpqhSTB#evalml.data_checks.OutliersDataCheck

(The text still says: "Checks if there are any outliers in input data by using an Isolation Forest to obtain the anomaly score of each index and then using IQR to determine score anomalies. Indices with score anomalies are considered outliers.")

evalml/data_checks/outliers_data_check.py
bchen1116: Yeah, they are different! I was experimenting around with SHAP since I'm not very familiar with it, and I think for now, it makes more sense to keep the simpler implementation and not use IsolationForest/SHAP. One of the tests I ran was this:
Screen Shot 2020-11-02 at 11 59 59 AM

When I included lots of other values in column 3, where half the dataset was > 500 and half the dataset was very small values, the iso/SHAP combo still classifies column 3 as an outlier, while the simpler implementation doesn't. Not sure which performance we prefer, but statistically, I'm more inclined to say the simpler implementation makes more sense. Let me know your thoughts @angela97lin @freddyaboulton

freddyaboulton: @bchen If I follow your example, column 3 is classified as an outlier for row 0 but I don't think we should compute the SHAP value for row 0 in your example because it doesn't have an isolation forest score outside the IQR. I computed the shap for row 0 in my example because I knew it would have a score outside the IQR (that was the behavior from the old unit test). Sorry if that was confusing! 

I think your solution makes sense and I think between a complex implementation and a simple implementation we should pick the simple one. That being said, I'm not sure what the best practice is for outlier detection! Maybe others have more decisive thoughts hehe 🙈

bchen1116: Ahh, I see. Yeah I misunderstood what your code was doing, but this makes sense now! Thanks for clarifying

I agree though, I think this implementation should catch values that are statistically outliers, but would love other input on whether using an IsolationForest would be preferred or not.

evalml/data_checks/outliers_data_check.py
```diff
@@ -21,24 +19,23 @@ def __init__(self, random_state=0):
...
         Returns:
-            A set of indices that may have outlier data.
+            A set of columns that may have outlier data.
 
         Example:
             >>> df = pd.DataFrame({
-            ...     'x': [1, 2, 3, 40, 5],
-            ...     'y': [6, 7, 8, 990, 10],
+            ...     'x': [1, 2, 3, 4, 5],
+            ...     'y': [6, 7, 8, 9, 10],
```
dsherry, line 34: Nice I like this simplification!

evalml/data_checks/outliers_data_check.py
```diff
@@ -21,24 +19,23 @@ def __init__(self, random_state=0):
...
             ...     'z': [-1, -2, -3, -1201, -4]
             ... })
             >>> outliers_check = OutliersDataCheck()
-            >>> assert outliers_check.validate(df) == [DataCheckWarning("Row '3' is likely to have outlier data", "OutliersDataCheck")]
+            >>> assert outliers_check.validate(df) == [DataCheckWarning("Column 'z' is likely to have outlier data", "OutliersDataCheck")]
```
dsherry, line 38: 👍

evalml/data_checks/outliers_data_check.py
```diff
@@ -56,11 +53,16 @@ def get_IQR(df, k=2.0):
...
+        # get the columns that fall out of the bounds, which means they contain outliers
+        for idx, bound in enumerate([lower_bound, upper_bound]):
+            cols_in_range = (X >= bound.values) if idx == 0 else (X <= bound.values)
+            cols_in_range = cols_in_range.all()
+            outlier_cols = cols_in_range[~cols_in_range].keys()
+            indices.update(outlier_cols.tolist())
+        # order the columns by how they appear in the dataframe
+        indices = sorted(list(indices), key=lambda x: X.columns.tolist().index(x))
+        warning_msg = "Column '{}' is likely to have outlier data"
+        s = [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in indices]
```
dsherry, line 67: @bchen1116 how does this code work? Why does this say `row_index` and not `column_name`?

evalml/data_checks/outliers_data_check.py
```diff
@@ -56,11 +53,16 @@ def get_IQR(df, k=2.0):
...
 
-        clf = IsolationForest(random_state=self.random_state)
-        clf.fit(X)
-        scores = pd.Series(clf.decision_function(X))
-        lower_bound, upper_bound = get_IQR(scores, k=2)
-        outliers = (scores < lower_bound) | (scores > upper_bound)
-        outliers_indices = outliers[outliers].index.values.tolist()
-        warning_msg = "Row '{}' is likely to have outlier data"
-        return [DataCheckWarning(warning_msg.format(row_index), self.name) for row_index in outliers_indices]
+        lower_bound, upper_bound = get_IQR(X)
```
dsherry, line 56: I know we default `k` to 2 in the signature but couldn't hurt to include `k=2` here.

Actually... I think it'd be cool if this were an init param, but fine to file separately if you don't wanna lump it into this PR

evalml/data_checks/outliers_data_check.py
dsherry: Why not save the column names directly in the impl? That would make the code simpler. Suggestion:
```python
def get_IQR(df, k=2.0):
    q1 = df.quantile(0.25)
    q3 = df.quantile(0.75)
    iqr = q3 - q1
    lower_bound = pd.Series(q1 - (k * iqr), name='lower_bound')
    upper_bound = pd.Series(q3 + (k * iqr), name='upper_bound')
    return pd.concat([lower_bound, upper_bound], axis=1)

iqr = get_IQR(X)
has_outliers = ((X < iqr['lower_bound']) & (X < iqr['upper_bound'])).any()
cols = has_outliers.index[has_outliers]
return [DataCheckWarning(warning_msg.format(col), self.name) for col in cols]
```
I love pandas bitwise operators!

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -19,16 +19,17 @@ def test_outliers_data_check_warnings():
...
 
     X = pd.DataFrame(data=data)
-    X.iloc[3, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[25, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[55, :] = pd.Series(np.random.randn(100) * 1000)
-    X.iloc[72, :] = pd.Series(np.random.randn(100) * 1000)
+    X.iloc[0, 3] = 1000
+    X.iloc[3, 25] = 1000
+    X.iloc[5, 55] = 10000
+    X.iloc[10, 72] = -1000
```
dsherry, line 25: Awesome

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -19,16 +19,17 @@ def test_outliers_data_check_warnings():
...
+    X.iloc[:, 90] = 'string_values'
```
dsherry, line 26: Did you add this just to make sure the code wouldn't crash if there's an unexpected value? That's cool

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -19,16 +19,17 @@ def test_outliers_data_check_warnings():
...
 
     outliers_check = OutliersDataCheck()
-    assert outliers_check.validate(X) == [DataCheckWarning("Row '3' is likely to have outlier data", "OutliersDataCheck"),
-                                          DataCheckWarning("Row '25' is likely to have outlier data", "OutliersDataCheck"),
-                                          DataCheckWarning("Row '55' is likely to have outlier data", "OutliersDataCheck"),
-                                          DataCheckWarning("Row '72' is likely to have outlier data", "OutliersDataCheck")]
+    assert outliers_check.validate(X) == [DataCheckWarning("Column '3' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '25' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '55' is likely to have outlier data", "OutliersDataCheck"),
+                                          DataCheckWarning("Column '72' is likely to have outlier data", "OutliersDataCheck")]
```
dsherry, line 32: Got it. Could you make the DF have non-integer col names? Will make it super clear that this is correct
```python
import string
...
n_cols = 26
a = np.arange(n_cols) * 0.01
data = np.tile(a, (100, 1))
X = pd.DataFrame(data=data, columns=[string.ascii_lowercase[i] for i in range(n_cols)])
```

dsherry (approved):
@bchen1116 looks good! I left some impl and test suggestions

evalml/data_checks/outliers_data_check.py
bchen1116: Ah, that was my mistake. I can fix, thanks for the simplification suggestion!

evalml/tests/data_checks_tests/test_outliers_data_check.py
bchen1116: yea, just to make sure it can handle string data

PR #1386, opened 2020-10-30
Add TimeSeriesRegression problem type

### LINKED ISSUE #1378: Implement problem type for time series regression
Currently we have a convention that each defined objective maps to a single problem type. In the future we will add objectives that are only valid for time series regression, but we would also want to evaluate the pipeline on "standard" regression metrics like MSE or MAE.

If we follow our existing convention, we would have to add `MSETimeSeries`, `MAETimeSeries` classes. This would lead to 2x the number of objectives with 100% duplication! 

@dsherry and I agreed that this wasn't the most maintainable solution so we agreed on the following **breaking** change: 

* Let `problem_type` in `RegressionObjective`, `BinaryClassificationObjective`, `MulticlassClassficationObjective`  be a list
* Add a method to `ObjectiveBase` called `is_defined_for_problem_type` that determines if the objective is defined for the given problem type

freddyaboulton:
### Pull Request Description
Fix #1378 .

-----

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -172,7 +172,7 @@ class AutoMLSearch:
         self.objective = self._validate_objective(objective)
         if self.data_split is not None and not issubclass(self.data_split.__class__, BaseCrossValidator):
             raise ValueError("Not a valid data splitter")
-        if self.problem_type not in objective.problem_type:
+        if not objective.is_defined_for_problem_type(self.problem_type):
             raise ValueError("Given objective {} is not compatible with a {} problem.".format(self.objective.name, self.problem_type.value))
         if additional_objectives is None:
             additional_objectives = get_core_objectives(self.problem_type)
@@ -559,7 +559,7 @@ class AutoMLSearch:
 
     def _validate_problem_type(self):
         for obj in self.additional_objectives:
-            if self.problem_type not in obj.problem_type:
+            if not obj.is_defined_for_problem_type(self.problem_type):
                 raise ValueError("Additional objective {} is not compatible with a {} problem.".format(obj.name, self.problem_type.value))
 
         for pipeline in self.allowed_pipelines or []:
@@ -652,13 +652,13 @@ class AutoMLSearch:
             try:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
-                if self.optimize_thresholds and ProblemTypes.BINARY in self.objective.problem_type and self.objective.can_optimize_threshold:
+                if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
                     X_train, X_threshold_tuning, y_train, y_threshold_tuning = train_test_split(X_train, y_train, test_size=0.2, random_state=self.random_state)
                 cv_pipeline = pipeline.clone(pipeline.random_state)
                 logger.debug(f"\t\t\tFold {i}: starting training")
                 cv_pipeline.fit(X_train, y_train)
                 logger.debug(f"\t\t\tFold {i}: finished training")
-                if ProblemTypes.BINARY in self.objective.problem_type:
+                if self.objective.is_defined_for_problem_type(ProblemTypes.BINARY):
                     cv_pipeline.threshold = 0.5
                     if self.optimize_thresholds and self.objective.can_optimize_threshold:
                         logger.debug(f"\t\t\tFold {i}: Optimizing threshold for {self.objective.name}")
@@ -808,7 +808,7 @@ class AutoMLSearch:
         log_subtitle(logger, "Training")
         logger.info("Training for {} problems.".format(pipeline.problem_type))
 
-        if self.optimize_thresholds and ProblemTypes.BINARY in self.objective.problem_type and self.objective.can_optimize_threshold:
+        if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
             logger.info("Objective to optimize binary classification pipeline thresholds for: {}".format(self.objective))
 
         logger.info("Total training time (including CV): %.1f seconds" % pipeline_results["training_time"])
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -172,7 +172,7 @@ class AutoMLSearch:
         self.objective = self._validate_objective(objective)
         if self.data_split is not None and not issubclass(self.data_split.__class__, BaseCrossValidator):
             raise ValueError("Not a valid data splitter")
-        if self.problem_type != self.objective.problem_type:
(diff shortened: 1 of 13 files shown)


evalml/tests/problem_type_tests/test_problem_types.py
```diff
@@ -16,7 +16,7 @@ def correct_problem_types():
 
 
 def test_handle_string(correct_problem_types):
-    problem_types = ['regression', ProblemTypes.MULTICLASS, 'binary']
+    problem_types = ['regression', ProblemTypes.MULTICLASS, 'binary', ProblemTypes.TIME_SERIES_REGRESSION]
```
bchen1116, line 19: how did you add `ProblemTypes.TIME_SERIES_REGRESSION` here without adding anything to `correct_problem_types`?

evalml/automl/automl_search.py
```diff
@@ -559,11 +559,11 @@ def _check_stopping_condition(self, start):
 
     def _validate_problem_type(self):
         for obj in self.additional_objectives:
-            if obj.problem_type != self.problem_type:
+            if not obj.is_defined_for_problem_type(self.problem_type):
                 raise ValueError("Additional objective {} is not compatible with a {} problem.".format(obj.name, self.problem_type.value))
 
         for pipeline in self.allowed_pipelines or []:
-            if not pipeline.problem_type == self.problem_type:
+            if pipeline.problem_type != self.problem_type:
```
angela97lin, line 566: 👍

evalml/tests/problem_type_tests/test_problem_types.py
angela97lin: Hm, I think we actually need to add to `correct_problem_types` for this test to be fully updated. Reason why it passes right now is because `zip(problem_types, correct_problem_types)` will only iterate when both lists still have items, so it only goes through `correct_problem_types` (3 elements)

evalml/objectives/regression_objective.py
```diff
@@ -9,5 +9,5 @@ class RegressionObjective(ObjectiveBase):
     problem_type (ProblemTypes): Type of problem this objective is. Set to ProblemTypes.REGRESSION.
```
angela97lin, line 9: We should update the docstring for each of the objective classes!

evalml/objectives/regression_objective.py
```diff
@@ -9,5 +9,5 @@ class RegressionObjective(ObjectiveBase):
...
     """
 
-    problem_type = ProblemTypes.REGRESSION
+    problem_types = [ProblemTypes.REGRESSION, ProblemTypes.TIME_SERIES_REGRESSION]
```
angela97lin, line 12: For my own curiosity: Are all regression objectives time-series applicable? Is there ever the case where an objective should be `ProblemTypes.TIME_SERIES_REGRESSION` and not `ProblemTypes.REGRESSION`? :o

angela97lin (approved):
LGTM! Just a few small comments about updating the test completely and the docstrings for each of the objective classes, but otherwise looks good 😁

evalml/tests/problem_type_tests/test_problem_types.py
freddyaboulton: Good catch guys! I'll update this.

evalml/objectives/regression_objective.py
freddyaboulton: I would say all of our current regression objectives would also work for a time series problem. But in the future, we will add objectives that are time-series specific, e.g:

https://en.wikipedia.org/wiki/Mean_absolute_scaled_error

evalml/objectives/regression_objective.py
freddyaboulton: Great point!

evalml/tests/problem_type_tests/test_problem_types.py
freddyaboulton: Done!

evalml/objectives/regression_objective.py
dsherry: I think this is the right call.

All our current regression objectives are valid for timeseries regression. So allowing `REGRESSION` and `TIME_SERIES_REGRESSION` is great.

When we add the first time-series-only objective, we can override `problem_types` in each of those impls to only allow `TIME_SERIES_REGRESSION`. We could also then choose to define `TimeseriesRegressionObjective` to facilitate this, but probably not necessary.

We can follow the same pattern for binary/multiclass.

evalml/problem_types/problem_types.py
```diff
@@ -11,11 +11,14 @@ class ProblemTypes(Enum):
     """Multiclass classification problem."""
     REGRESSION = 'regression'
     """Regression problem."""
+    TIME_SERIES_REGRESSION = 'time_series_regression'
```
dsherry, line 14: This is a nit-pick, but: in our docs examples we typically type out `problem_type='multiclass'` etc. Is it easier for people to type/remember `problem_type='time series regression'` or `problem_type='time_series_regression'`?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1120,16 +1122,22 @@ def test_percent_better_than_baseline_in_rankings(objective, pipeline_scores, ba
     # Ok to only use binary labels since score and fit methods are mocked
     X, y = X_y_binary
 
+    objective_type = ProblemTypes.BINARY
+    if isinstance(objective, RegressionObjective):
+        objective_type = ProblemTypes.REGRESSION
+    elif isinstance(objective, MulticlassClassificationObjective):
+        objective_type = ProblemTypes.MULTICLASS
```
dsherry, line 1129: This is fine for now. But once we add objectives which aren't specific to one problem type (like timeseries), we'll have to remember to change this. Is there anything we can do to future-proof this test?

One option: add `problem_type` to the `parametrize`. Then add
```python
if not objective.is_defined_for_problem_type(problem_type):
    pytest.skip()
```
(syntax prob wrong)

No prob if you don't get to this. I hope that in the wild, we'll always know the `problem_type` when we're working with objectives, so this doesn't worry me outside of our tests.

evalml/tests/problem_type_tests/test_problem_types.py
dsherry: 👏   great reviewing @bchen1116 @angela97lin !

dsherry (approved):
🚢 !

evalml/problem_types/problem_types.py
freddyaboulton: Good point! Avoiding the underscores seems easier to type so I changed it to that hehe.

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Made this change! Good point - it does make the test simpler! And when we add time series to automl, we just need to update the list of values that's parametrized so that's great.

Issue #1367, opened 2020-10-29
Automl fails with custom pipelines if data contains datetime feature(s)


dsherry:
#### Background
We're in a funny intermediate stage with adding support for woodwork in #1229. We've added support to automl `search`, but pipelines, components, `make_pipeline` and other utils still expect pandas `DataFrame`s.

Another fact which is important here: creating a woodwork `DataTable` will modify the types of the underlying pandas `DataFrame`

#### Problem
The following code doesn't work, borrowed from @bchen1116 's docs additions in his PR #1284 with some modifications for simplicity:
```python
import evalml
X, y = evalml.demos.load_fraud(n_rows=1000)

estimators = evalml.pipelines.components.utils.get_estimators('binary', [evalml.model_family.ModelFamily.EXTRA_TREES])
pipelines = [evalml.pipelines.utils.make_pipeline(X, y, estimator, 'binary') for estimator in estimators]
automl = evalml.automl.AutoMLSearch(problem_type='binary', allowed_pipelines=pipelines)
automl.search(X, y)
```
errors out with
```
AutoMLSearchException: All pipelines in the current AutoML batch produced a score of np.nan on the primary objective <evalml.objectives.standard_metrics.LogLossBinary object at 0x13ea47700>.
```

Reason: `make_pipeline` doesn't have access to the pandas DF. Because the datetime feature comes in with type "object" (string), `make_pipeline` won't know there's a datetime feature, and won't add the `DateTimeFeaturizer`.

Then, in automl search, because we use woodwork `DataTable`s in there right now, the datetime feature will get converted to type `datetime64`. When the pipeline is evaluated on that data, it'll break, because the `StandardScaler` will try to apply scaling to a `datetime64`-type column and will error out.

#### Workaround
The following code does work:
```python
import evalml
X, y = evalml.demos.load_fraud(n_rows=1000)
import woodwork as ww
X_dt = ww.DataTable(X)
y_dc = ww.DataColumn(y)

estimators = evalml.pipelines.components.utils.get_estimators('binary', [ModelFamily.EXTRA_TREES])
pipelines = [evalml.pipelines.utils.make_pipeline(X, y, estimator, 'binary') for estimator in estimators]
automl = evalml.automl.AutoMLSearch(problem_type='binary', allowed_pipelines=pipelines)
automl.search(X, y)
```

[...]

Issue #1289, opened 2020-10-09
Update AutoML to pass Woodwork DataTables to every pipeline/component (instead of pandas DataFrames)


angela97lin:
Design doc for epic here: https://alteryx.quip.com/Oa4aA6h0Rf2k/Woodwork-DataTable-Integration-Design-Document

angela97lin:
Note: we would need some features from Woodwork in order to make this happen. Most prominently, a way to access a subset of the Woodwork DataTable without having to convert to pandas (iloc).

PR #1393, opened 2020-11-02
Update pipelines and `make_pipelines` to accept Woodwork DataTables

### LINKED ISSUE #1288: Update pipelines and all components to accept Woodwork DataTables

angela97lin:
Addresses of #1288 and #1367 to update pipeline classes and `make_pipelines` to handle Woodwork data types. Components will be handled in a separate PR.

### DIFF
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -33,7 +33,6 @@ class ClassificationPipeline(PipelineBase):
 
         Arguments:
             X (ww.DataTable, pd.DataFrame or np.ndarray): The input training data of shape [n_samples, n_features]
-
             y (ww.DataColumn, pd.Series, np.ndarray): The target training labels of length [n_samples]
 
         Returns:
@@ -119,7 +118,7 @@ class ClassificationPipeline(PipelineBase):
 
         Arguments:
             X (ww.DataTable, pd.DataFrame or np.array): Data of shape [n_samples, n_features]
-            y (pd.Series): True labels of length [n_samples]
+            y (pd.Series, ww.DataColumn): True labels of length [n_samples]
             objectives (list): List of objectives to score
 
         Returns:
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -74,7 +74,7 @@ class ClassificationPipeline(PipelineBase):
         """Make predictions using selected features.
 
         Arguments:
-            X (pd.DataFrame or np.array): Data of shape [n_samples, n_features]
+            X (pd.DataFrame): Data of shape [n_samples, n_features]
             objective (Object or string): The objective to use to make predictions
 
         Returns:
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -4,6 +4,10 @@ from sklearn.preprocessing import LabelEncoder
 
 from evalml.objectives import get_objective
 from evalml.pipelines import PipelineBase
+from evalml.utils.gen_utils import (
+    _convert_to_woodwork_structure,
+    _convert_woodwork_types_wrapper
+)
 
 
 class ClassificationPipeline(PipelineBase):
@@ -28,18 +32,18 @@ class ClassificationPipeline(PipelineBase):
             by sorted(set(y)) and then are mapped to values between 0 and n_classes-1.
 
         Arguments:
-            X (pd.DataFrame or np.array): The input training data of shape [n_samples, n_features]
+            X (ww.DataTable, pd.DataFrame or np.ndarray): The input training data of shape [n_samples, n_features]
 
-            y (pd.Series): The target training labels of length [n_samples]
+            y (ww.DataColumn, pd.Series, np.ndarray): The target training labels of length [n_samples]
 
         Returns:
             self
 
         """
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
-        if not isinstance(y, pd.Series):
-            y = pd.Series(y)
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_pandas())
+        y = _convert_woodwork_types_wrapper(y.to_pandas())
         self._encoder.fit(y)
         y = self._encode_targets(y)
         self._fit(X, y)
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -80,8 +80,6 @@ class ClassificationPipeline(PipelineBase):
         Returns:
             pd.Series: Estimated labels
         """
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
         X_t = self.compute_estimator_features(X)
         return self.estimator.predict(X_t)
 
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -89,12 +89,14 @@ class ClassificationPipeline(PipelineBase):
         """Make predictions using selected features.
 
         Arguments:
-            X (pd.DataFrame or np.array): Data of shape [n_samples, n_features]
+            X (ww.DataTable, pd.DataFrame, or np.array): Data of shape [n_samples, n_features]
             objective (Object or string): The objective to use to make predictions
 
         Returns:
             pd.Series : Estimated labels
         """
+        X = _convert_to_woodwork_structure(X)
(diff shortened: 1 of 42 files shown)


evalml/pipelines/utils.py
```diff
@@ -99,6 +100,12 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None, te
         class: PipelineBase subclass with dynamically generated preprocessing components and specified estimator
 
     """
+    X = _convert_to_woodwork_structure(X)
+    y = _convert_to_woodwork_structure(y)
+
+    X = _convert_woodwork_types_wrapper(X.to_pandas())
+    y = _convert_woodwork_types_wrapper(y.to_pandas())
```
eccabay, line 107: I don't think this actually needs to be converted - nothing in this function seems to explicitly need numpy types. `_get_preprocessing_components` uses dtype selection, can we use woodwork for that instead? This would also simplify the text support code, if `_get_preprocessing_components` can check for text columns itself rather than the variable being passed all the way through the stack trace from `search` (as per the discussion [here](https://github.com/alteryx/evalml/pull/1062#discussion_r512944810))

Alternatively, we could file a separate issue for this

angela97lin:
Note: we want warnings for users who pass in non-Woodwork data structures. Will either update this PR or put up one shortly after (depending on where reviews are) to address this.

evalml/pipelines/utils.py
freddyaboulton: Yea I think it would be better for `_get_preprocessing_components` to have access to the woodwork table rather than a dataframe. I think we'll need to do this eventually because our logic to determine the pre-processing components should leverage the full range of logical types in ww. Might be best to do that sooner rather than later?

Also, as @eccabay mentioned, that would move us closer to being able to delete the `text_columns` parameter which I think is nice!

evalml/tests/pipeline_tests/classification_pipeline_tests/test_classification.py
```diff
@@ -47,3 +47,9 @@ def test_pipeline_has_classes_property(logistic_regression_binary_pipeline_class
 
     pipeline.fit(X, y)
     pd.testing.assert_series_equal(pd.Series(pipeline.classes_), pd.Series(answer))
+
+
+def test_woodwork_classification_pipeline(logistic_regression_binary_pipeline_class):
```
freddyaboulton, line 52: Did you mean to convert X and y to ww objects before passing to fit?

If not, then this test may be redundant? The other tests in this file are also doing the ww conversion in fit.

Same for `test_regression.py`

evalml/tests/pipeline_tests/regression_pipeline_tests/test_regression.py
```diff
@@ -1,17 +1,23 @@
...
+from evalml.demos import load_breast_cancer, load_diabetes, load_wine
 
 
-@pytest.mark.parametrize("target_type", ["categorical", "string", "bool"])
+@pytest.mark.parametrize("target_type", ["category", "string", "bool"])
 def test_invalid_targets_regression_pipeline(target_type, dummy_regression_pipeline_class):
     X, y = load_wine()
-    if target_type == "categorical":
-        y = pd.Categorical(y)
+    if target_type == "category":
```
freddyaboulton, line 10: How come we need this change?

freddyaboulton (approved):
@angela97lin I think this looks good! I am interested in your thoughts on whether we should convert back to pandas in `make_pipeline`!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -230,6 +258,38 @@ def test_make_pipeline_no_column_names(problem_type):
...
+        for problem_type in estimator_class.supported_problem_types:
+            pipeline = make_pipeline(X, y, estimator_class, problem_type, text_columns=['text'])
+            assert isinstance(pipeline, type(pipeline_class))
+            assert pipeline.custom_hyperparameters is None
+            if estimator_class.model_family == ModelFamily.LINEAR_MODEL:
+                assert pipeline.component_graph == [Imputer, TextFeaturizer, OneHotEncoder, StandardScaler, estimator_class]
+            elif estimator_class.model_family == ModelFamily.CATBOOST:
+                assert pipeline.component_graph == [Imputer, TextFeaturizer, estimator_class]
+            else:
+                assert pipeline.component_graph == [Imputer, TextFeaturizer, OneHotEncoder, estimator_class]
```
dsherry, line 290: Yes!!!! One step closer to having the `TextFeaturizer` get text columns directly from the woodwork datatable
hooray

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1143,32 +1182,49 @@ def test_get_default_parameters(logistic_regression_binary_pipeline_class):
     assert logistic_regression_binary_pipeline_class.default_parameters == expected_defaults
 
 
+@pytest.mark.parametrize("data_type", ['np', 'pd', 'ww'])
```
dsherry, line 1185: Awesome coverage!

evalml/utils/gen_utils.py
```diff
@@ -272,6 +273,25 @@ def is_all_numeric(df):
...
+def _convert_to_woodwork_structure(data):
+    """
+    Takes input data structure, and if it is not a Woodwork data structure already, will convert it to a Woodwork DataTable or DataColumn structure.
+    """
+    if isinstance(data, ww.DataTable) or isinstance(data, ww.DataColumn):
+        return data
+    # Convert numpy data structures to pandas data structures
+    if isinstance(data, list):
+        data = np.array(data)
+    if isinstance(data, pd.api.extensions.ExtensionArray) or (isinstance(data, np.ndarray) and len(data.shape) == 1):
```
dsherry, line 285: @angela97lin cool I didn't realize `ExtensionArray` was specifically 1d, but I see [the pandas docs mention it](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.extensions.ExtensionArray.html). Looks good

evalml/pipelines/classification_pipeline.py
```diff
@@ -70,42 +74,41 @@ def _predict(self, X, objective=None):
...
-            X = pd.DataFrame(X)
         X_t = self.compute_estimator_features(X)
         return self.estimator.predict(X_t)
 
     def predict(self, X, objective=None):
         """Make predictions using selected features.
 
         Arguments:
-            X (pd.DataFrame or np.array): Data of shape [n_samples, n_features]
+            X (ww.DataTable, pd.DataFrame, or np.array): Data of shape [n_samples, n_features]
```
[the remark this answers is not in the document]
dsherry: I see that we do, and that you've added more in this PR. Great!

dsherry (approved):
@angela97lin amazing!

Blocking:
* delete pdb import from pipeline base
* Resolve comment in `make_pipelines` about not needing `_convert_woodwork_types_wrapper`
* Resolve @freddyaboulton comment about `test_woodwork_classification_pipeline`

evalml/pipelines/classification_pipeline.py
```diff
@@ -28,18 +32,18 @@ def fit(self, X, y):
...
 
         """
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
-        if not isinstance(y, pd.Series):
-            y = pd.Series(y)
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_pandas())
+        y = _convert_woodwork_types_wrapper(y.to_pandas())
```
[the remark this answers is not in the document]
angela97lin: @dsherry Not quite; for now, we're passing the DataFrames to each pipeline. We'll pass the actual woodwork structure in work done for https://github.com/alteryx/evalml/issues/1289 but there are a few methods and more massaging that needs to happen on the automl side (getting woodwork to work with data splitting, primarily) before this is true.

evalml/pipelines/utils.py
angela97lin: Ah, good idea! I was planning on doing the connectors and passing woodwork through each of the methods in a separate PR but you guys are right that it's work that needs to be done at some point so better sooner than later. Updating, thanks all!

evalml/pipelines/utils.py
```diff
@@ -109,9 +116,6 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None, te
         raise ValueError(f"if custom_hyperparameters provided, must be dictionary. Received {type(custom_hyperparameters)}")
 
     hyperparameters = custom_hyperparameters
-    if not isinstance(X, pd.DataFrame):
-        X = pd.DataFrame(X)
```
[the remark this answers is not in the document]
angela97lin: Yeah seems like we always converted to pd.DataFrame before lol

PR #1396, opened 2020-11-02
Add a DelayedFeaturesTransformer

### LINKED ISSUE #1379: Add a DelayedFeaturesTransformer
Add a transformer to compute lagged features from the features in the dataset and the target variable.

Requirements:
1. Compute lagged features according to both the max_lag and gap parameters. 
2. Column names should make the feature name and delay amount clear.
3. Should return `nan` for feature/lag combos that don't have enough historical data.

Out of scope:
1. Handling irregularly spaced time series

freddyaboulton:
### Pull Request Description
Fix #1379

-----


evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,42 @@
...
+        if not isinstance(X, pd.DataFrame):
+            if y is None:
+                X = pd.DataFrame(X, columns=["target"])
+            else:
+                X = pd.DataFrame(X)
+        if y is not None and not isinstance(y, pd.Series):
+            y = pd.Series(y)
+
+        original_columns = X.columns
+        X = X.assign(**{f"{col}_delay_{t}": X[col].shift(t)
```
freddyaboulton, line 32: If the time-series is irregularly spaced, shifting may give you data that is too "far in the past", e.g. we have daily data, but for some reason, a whole month is missing. We've decided that irregularly-spaced time series are out-of-scope for this first release. 

https://alteryx.quip.com/AM04ASOaQS4v/Time-Series-November-Design-Document

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,42 @@
...
+
+from evalml.pipelines.components.transformers.transformer import Transformer
+
+
+class DelayedFeaturesTransformer(Transformer):
+    name = "Delayed Features Transformer"
+    hyperparameter_ranges = {}
+    needs_fitting = False
+
+    def __init__(self, max_delay=2, random_state=0, **kwargs):
```
freddyaboulton, line 11: In the future, we may want to add a parameter for features that should not be lagged (they don't change with time). I think proceeding without that functionality is good enough for an MVP though.

evalml/tests/component_tests/test_delayed_features_transformer.py
```diff
@@ -0,0 +1,115 @@
...
+    return X, y
+
+
+def test_delayed_features_transformer_init():
+
+    delayed_features = DelayedFeaturesTransformer(max_delay=4, random_state=1)
+    assert delayed_features.parameters == {"max_delay": 4}
+
+
+def test_lagged_feature_extractor_maxdelay3_gap1(delayed_features_data):
```
freddyaboulton, line 20: These tests could be condensed into one test with `parametrize` but I think that being explicit here makes it easier to see that the output matches what's in the design doc.

bchen1116 (approved):
LGTM!

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,56 @@
...
+        Arguments:
+            X (pd.DataFrame): Data to transform.
+            y (pd.Series, optional): Targets.
+
+        Returns:
+            pd.DataFrame: Transformed X.
+        """
+        if not isinstance(X, pd.DataFrame):
+            if y is None:
+                X = pd.DataFrame(X, columns=["target"])
```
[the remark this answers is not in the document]
freddyaboulton: In time series it's possible to fit estimators based only the target variable so we need to detect when we are in that case. 

If X is not a dataframe and y is None (no second argument was passed in), then I'm inferring that only the target variable was passed in and so I convert it to a dataframe with only one column.

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,56 @@
...
+        if y is not None and not isinstance(y, pd.Series):
+            y = pd.Series(y)
+
+        original_columns = X.columns
+        X = X.assign(**{f"{col}_delay_{t}": X[col].shift(t)
+                        for t in range(self.max_delay + 1)
+                        for col in X})
+        X.drop(columns=original_columns, inplace=True)
+
+        # Handle cases where the label was not passed
```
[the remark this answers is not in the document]
freddyaboulton: I'll fix this and add some more comments to explain what's going on based on your previous comment! ^

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,57 @@
...
+                        for t in range(self.max_delay + 1)
+                        for col in X})
+        X.drop(columns=original_columns, inplace=True)
+
+        # Handle cases where the target was passed in
+        if y is not None:
+            X = X.assign(**{f"target_delay_{t}": y.shift(t)
+                            for t in range(self.max_delay + 1)})
+
+        return X
```
dsherry, line 57: 👍  awesome, so we're adding the delayed target features to the returned X. Yep.

Note the target delay range should be `for t in range(1, self.max_delay + 1)`, to avoid `delay==0`!

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,57 @@
...
+
+from evalml.pipelines.components.transformers.transformer import Transformer
+
+
+class DelayedFeaturesTransformer(Transformer):
+    name = "Delayed Features Transformer"
+    hyperparameter_ranges = {}
+    needs_fitting = False
+
+    def __init__(self, max_delay=2, random_state=0, **kwargs):
```
dsherry, line 11: @freddyaboulton can we add parameters to control delaying the target vs delaying the features?
```python
delay_target=False,
delay_features=True
```
I think it'll be helpful to be able to control the behavior like this.

evalml/tests/component_tests/test_components.py
```diff
@@ -530,6 +531,10 @@ def test_transformer_transform_output_type(X_y_binary):
                 assert transform_output.shape[0] == X.shape[0]
                 assert transform_output.shape[1] <= X.shape[1]
                 assert isinstance(transform_output.columns, pd.Index)
+            elif isinstance(component, DelayedFeaturesTransformer):
+                # We just want to check that DelayedFeaturesTransformer outputs a DataFrame
+                # The dataframe shape and index are checked in test_delayed_features_transformer.py
+                continue
```
dsherry, line 537: 👍

evalml/tests/component_tests/test_delayed_features_transformer.py
```diff
@@ -0,0 +1,115 @@
...
+                                  "target_delay_3": y.shift(3)})
+    pd.testing.assert_frame_equal(DelayedFeaturesTransformer(max_delay=3, gap=7).fit_transform(y), answer_only_y)
+
+
+def test_lagged_feature_extractor_numpy(delayed_features_data):
+    X, y = delayed_features_data
+    X_np = X.values
+    y_np = y.values
+
+    # Example 3 from the design document
```
dsherry, line 99: I'd delete these comments, the examples stand on their own here!

evalml/tests/component_tests/test_delayed_features_transformer.py
```diff
@@ -0,0 +1,115 @@
...
+
+    pd.testing.assert_frame_equal(DelayedFeaturesTransformer(max_delay=5, gap=1).fit_transform(X, y), answer)
+
+    answer_only_y = pd.DataFrame({"target_delay_0": y,
+                                  "target_delay_1": y.shift(1),
+                                  "target_delay_2": y.shift(2),
+                                  "target_delay_3": y.shift(3),
+                                  "target_delay_4": y.shift(4),
+                                  "target_delay_5": y.shift(5)})
+    pd.testing.assert_frame_equal(DelayedFeaturesTransformer(max_delay=5, gap=1).fit_transform(y), answer_only_y)
```
dsherry, line 68: In our components we should always be able to assume that X contains features and y contains the target.

So I'd expect this call to be
```python
tf.fit_transform(None, y=y)
```
or perhaps
```python
tf.fit_transform(pd.DataFrame(), y=y)
```

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,57 @@
+import pandas as pd
+
+from evalml.pipelines.components.transformers.transformer import Transformer
+
+
+class DelayedFeaturesTransformer(Transformer):
```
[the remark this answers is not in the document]
dsherry: Also could you please add class and init docstrings?

dsherry (changes requested):
Very cool!

I left a few comments. The impl and tests look solid, just following up on details.

evalml/pipelines/components/transformers/preprocessing/time_series.py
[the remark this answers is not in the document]
freddyaboulton: Will change the name! I thought plural was appropriate because it will lag more than one feature if there are multiple input features.

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,57 @@
...
+
+        For each feature in X, it will add a column to the output dataframe for each
+        delay in the (inclusive) range [1, max_delay]. The values would correspond to the
+        value of the feature 'delay' time-steps in the past.
+
+        If y is not None, it will also compute the delayed values for the target variable.
+
+        Arguments:
+            X (pd.DataFrame): Data to transform.
+            y (pd.Series, optional): Targets.
```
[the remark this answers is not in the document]
freddyaboulton: Changed to None!

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,57 @@
...
+            else:
+                X = pd.DataFrame(X)
+        if y is not None and not isinstance(y, pd.Series):
+            y = pd.Series(y)
+
+        original_columns = X.columns
+        X = X.assign(**{f"{col}_delay_{t}": X[col].shift(t)
+                        for t in range(self.max_delay + 1)
+                        for col in X})
+        X.drop(columns=original_columns, inplace=True)
```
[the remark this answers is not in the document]
freddyaboulton: We call `.shift(0)` on the column which keeps the values the same but we append `delay_0` to the column name. We drop the original column to avoid redundant columns.

evalml/pipelines/components/transformers/preprocessing/time_series.py
freddyaboulton: The pipeline will shift the target variable by the `gap` amount to take care of target leakage. The only time this would cause a problem is when `gap=0` which I haven't personally seen in practice but it'd be a good idea to support. In that case, I'll start the target delay at `1`.

So in short:

- Add a `gap` parameter to the transformer init method.
-  If gap = 0, the features will be delayed from [0, max_delay], the target will be delayed [1, max_delay] (square brackets means the range is inclusive)
-  Else, features and targets will be delayed from [0, max_delay] (current behavior)

evalml/pipelines/components/transformers/preprocessing/time_series.py
freddyaboulton: @dsherry Should these be tuned by automl?

evalml/pipelines/components/transformers/preprocessing/time_series.py
dsherry: Ah very good point, you're right of course, I agree with your plan! Thanks.

I agree `gap=0` for timeseries is not gonna see heavy use. You'd still get the delayed features which is nice, but that's essentially saying "given todays info, and historical features, predict todays target". Not invalid per se, but not the main point of timeseries.

evalml/pipelines/components/transformers/preprocessing/time_series.py
dsherry: @freddyaboulton good question. Thinking about that, I think we should:
* Default these both to True in this PR
* Once we get start performance testing for timeseries we can determine whether or not we should make these tuneable. But my default assumption would be to leave it as non-tuneable

We need the ability to generate both delayed features and delayed target features. But for many datasets, `target[n-1]` will be strongly correlated with `target[n]`, and users may not want their models to rely so heavily on the delayed target, and will want to disable the delayed target features. Hence why I suggested we add these two parameters.

evalml/pipelines/components/transformers/preprocessing/time_series.py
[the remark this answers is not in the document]
dsherry: Thanks. Yep got it, that makes sense, I'm fine either way but prefer the singular here because I think its more in line with our other components. I also recognize its a nit-pick / opinion haha

evalml/tests/component_tests/test_delayed_features_transformer.py
freddyaboulton: Good point! I thought it'd be easier for users to avoid having to type `None` for an Arima-like problem where only the y is being modeled but it's best to be explicit and follow our convention.

evalml/pipelines/components/transformers/preprocessing/time_series.py
freddyaboulton: Great point about the previous time step being correlated with the current time step! 

I agree with your suggestion to leave the question about tuning until we have performance tests. I was thinking about the case when there are many delayed features. In that situation, not all of them will be useful for modeling the target and so we may want to intelligently not delay some of those features. There are other, and probably better ways of parametrizing that though, apart from the coarse "should we delay or not delay" decision.

dsherry: @freddyaboulton agreed. These two binary parameters are not going to be enough for us for tuning in the long-term.

evalml/pipelines/components/transformers/preprocessing/time_series.py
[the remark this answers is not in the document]
dsherry: Ah, yep I agree, we should keep the original feature names here instead of renaming them.

evalml/pipelines/components/transformers/preprocessing/time_series.py
```diff
@@ -0,0 +1,75 @@
...
+                at 1.
+            random_state (int, np.random.RandomState): Seed for the random number generator. There is no randomness
+                in this transformer.
+        """
+        self.max_delay = max_delay
+        self.delay_features = delay_features
+        self.delay_target = delay_target
+
+        # If 0, start at 1
+        self.start_delay_for_target = gap == 0
```
dsherry, line 32: Do you need to convert to int here?

dsherry (approved):
Great!

PR #1401, opened 2020-11-03
Add Target Encoder to Components

### LINKED ISSUE #1390: Target encoding
https://contrib.scikit-learn.org/category_encoders/targetencoder.html

Will need to run perf tests.

Ideally we can show key examples where target encoding outperforms one-hot encoding. There should be some kaggle datasets where target encoding outperforms one-hot.

bchen1116:
fix #1390 

Adding target encoder to components. Perf test results [are here](https://alteryx.quip.com/ly8rA0Ks3M6V/Target-Encoder-Transformer)

Docs are [here](https://evalml.alteryx.com/en/bc_1390_target/generated/evalml.pipelines.components.TargetEncoder.html#evalml.pipelines.components.TargetEncoder)

[UPDATE]
On Friday, 11/6, we decided to add additional perf tests for `TE` (discussion in the quip doc above).
Things to test:
1. Running OHE vs TE on categorical string data
2. Running OHE vs TE on categorical data that has already been converted to int/float
3. Running OHE vs TE on categorical data that has already been converted to int/float, where we convert the column type to categorical if there are less than 10 unique values in that column.

The perf test doc contains these 3 tests, ordered from 3, 1, 2.

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/target_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/target_encoder.py
@@ -1,6 +1,10 @@
+import pandas as pd
+
 from ..transformer import Transformer
 
-from evalml.pipelines.components.transformers.encoders.onehot_encoder import OneHotEncoderMeta
+from evalml.pipelines.components.transformers.encoders.onehot_encoder import (
+    OneHotEncoderMeta
+)
 from evalml.utils import import_or_raise
 
 
@@ -48,8 +52,10 @@ class TargetEncoder(Transformer, metaclass=OneHotEncoderMeta):
                          random_state=random_state)
 
     def fit(self, X, y):
-        X.reset_index(drop=True, inplace=True)
-        y.reset_index(drop=True, inplace=True)
+        if isinstance(X, pd.DataFrame):
+            X.reset_index(drop=True, inplace=True)
+        if isinstance(y, pd.Series):
+            y.reset_index(drop=True, inplace=True)
         return super().fit(X, y)
 
     def transform(self, X, y=None):
--- a/evalml/pipelines/components/transformers/encoders/target_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/target_encoder.py
@@ -1,15 +1,10 @@
 from ..transformer import Transformer
 
-from evalml.pipelines.components import ComponentBaseMeta
+from evalml.pipelines.components.transformers.encoders.onehot_encoder import OneHotEncoderMeta
 from evalml.utils import import_or_raise
 
 
-class TargetEncoderMeta(ComponentBaseMeta):
-    """A version of the ComponentBaseMeta class which includes validation on an additional target-encoder method `categories`"""
-    METHODS_TO_CHECK = ComponentBaseMeta.METHODS_TO_CHECK + ['get_feature_names']
-
-
-class TargetEncoder(Transformer, metaclass=TargetEncoderMeta):
+class TargetEncoder(Transformer, metaclass=OneHotEncoderMeta):
     """Target encoder to encode categorical data"""
     name = 'Target Encoder'
     hyperparameter_ranges = {}
@@ -26,7 +21,7 @@ class TargetEncoder(Transformer, metaclass=TargetEncoderMeta):
         Arguments:
             cols (list): Columns to encode. If None, all string columns will be encoded, otherwise only the columns provided will be encoded.
                 Defaults to None
-            smoothing (float): The smoothing factor to apply. The larger this value is, the more influence the expected value has
+            smoothing (float): The smoothing factor to apply. The larger this value is, the more influence the expected target value has
                 on the resulting target encodings. Must be strictly larger than 0. Defaults to 1.0
             handle_unknown (string): Determines how to handle unknown categories for a feature encountered. Options are 'value', 'error', nd 'return_nan'.
                 Defaults to 'value', which replaces with the target mean
@@ -53,10 +48,15 @@ class TargetEncoder(Transformer, metaclass=TargetEncoderMeta):
                          random_state=random_state)
 
     def fit(self, X, y):
+        X.reset_index(drop=True, inplace=True)
+        y.reset_index(drop=True, inplace=True)
         return super().fit(X, y)
 
     def transform(self, X, y=None):
         return super().transform(X, y)
+
+    def fit_transform(self, X, y):
+        return self.fit(X, y).transform(X)
 
     def get_feature_names(self):
         """Return feature names for the input features after fitting.
--- a/evalml/pipelines/components/transformers/encoders/target_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/target_encoder.py
@@ -1,0 +1,67 @@
+from ..transformer import Transformer
+
+from evalml.pipelines.components import ComponentBaseMeta
+from evalml.utils import import_or_raise
+
+
+class TargetEncoderMeta(ComponentBaseMeta):
+    """A version of the ComponentBaseMeta class which includes validation on an additional target-encoder method `categories`"""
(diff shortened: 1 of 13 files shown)


evalml/pipelines/components/transformers/encoders/target_encoder.py
```diff
@@ -0,0 +1,68 @@
...
+            raise ValueError("Invalid input '{}' for handle_unknown".format(handle_unknown))
+        if handle_missing not in unknown_and_missing_input_options:
+            raise ValueError("Invalid input '{}' for handle_missing".format(handle_missing))
+        if smoothing <= 0:
+            raise ValueError("Smoothing value needs to be strictly larger than 0. {} provided".format(smoothing))
+
+        category_encode = import_or_raise('category_encoders', error_msg='category_encoders not installed. Please install using `pip install category_encoders`')
+        self._target_encoder = category_encode.target_encoder.TargetEncoder(**parameters)
+        super().__init__(parameters=parameters,
+                         component_obj=self._target_encoder,
```
eccabay, line 53: I don't think you need to save the `target_encoder` as an instance variable - calling `super().__init__` and passing in the target encoder will save that object in `self._component_obj`.

evalml/tests/component_tests/test_target_encoder.py
```diff
@@ -0,0 +1,147 @@
+import numpy as np
+import pandas as pd
+import pytest
+from pytest import importorskip
+
+from evalml.exceptions import ComponentNotYetFittedError
+from evalml.pipelines.components import TargetEncoder
+
+importorskip('category_encoders', reason='Skipping test because lightgbm not installed')
```
eccabay, line 9: `lightgbm` - typo?

evalml/pipelines/pipeline_base.py
```diff
@@ -194,8 +194,8 @@ def compute_estimator_features(self, X):
         return X_t
 
     def _fit(self, X, y):
-        X_t = X
-        y_t = y
+        X_t = X.reset_index(drop=True)
```
bchen1116, line 197: Added to avoid a bug when either dataframe/series have indices that aren't aligned with each other

evalml/pipelines/pipeline_base.py
```diff
@@ -205,7 +205,8 @@ def _compute_features_during_fit(self, X, y):
         return X_t
 
     def _fit(self, X, y):
-
+        X.reset_index(inplace=True, drop=True)
+        y.reset_index(inplace=True, drop=True)
```
bchen1116, line 209: Include this code in order to avoid errors in `TargetEncoder` when used with the `Imputer`

evalml/tests/component_tests/test_target_encoder.py
```diff
@@ -0,0 +1,147 @@
...
+                               'col_2': [2, 1, 1, 1, 1],
+                               'col_3': [0.686166, 0.686166, 0.686166, 0.686166, 0.6]})
+    pd.testing.assert_frame_equal(X_t, X_expected)
+
+    encoder = TargetEncoder(smoothing=100)
+    encoder.fit(X, y)
+    X_t = encoder.transform(X)
+    X_expected = pd.DataFrame({'col_1': [1, 2, 1, 1, 2],
+                               'col_2': [2, 1, 1, 1, 1],
+                               'col_3': [0.676125, 0.676125, 0.676125, 0.676125, 0.6]})
```
freddyaboulton, line 134: Interesting! Increasing the smoothing by a factor of 10 barely changed the estimate compared to smoothing=10.

evalml/pipelines/components/transformers/encoders/target_encoder.py
```diff
@@ -0,0 +1,67 @@
+from ..transformer import Transformer
+
+from evalml.pipelines.components import ComponentBaseMeta
+from evalml.utils import import_or_raise
+
+
+class TargetEncoderMeta(ComponentBaseMeta):
```
freddyaboulton, line 7: nit-pick: I think we can reuse the OneHotEncoderMeta. If a user calls `categories` on TargetEncoder, they'll get an `AttributeError` instead of `ComponentNotYetFittedError` which is what I'd expect. 

It's a minor point but it would be nice if our encoders can share as much code as possible, in my opinion.

evalml/tests/component_tests/test_target_encoder.py
```diff
@@ -0,0 +1,147 @@
...
+    with pytest.raises(ValueError, match='Columns to be encoded can not contain null'):
+        encoder.fit(X, y)
+
+
+def test_cols():
+    X = pd.DataFrame({'col_1': [1, 2, 1, 1, 2],
+                      'col_2': ['2', '1', '1', '1', '1'],
+                      'col_3': ["a", "a", "a", "a", "a"]})
+    y = pd.Series([0, 1, 1, 1, 0])
+    encoder = TargetEncoder(cols=[])
```
freddyaboulton, line 71: Nice, I think this is the same behavior on the `OneHotEncoder`!

evalml/tests/component_tests/test_target_encoder.py
```diff
@@ -0,0 +1,147 @@
+import numpy as np
```
freddyaboulton, line 1: I think we should add some unit tests with custom indices. I am now officially paranoid 😂

evalml/pipelines/components/transformers/encoders/target_encoder.py
```diff
@@ -0,0 +1,67 @@
...
+    """A version of the ComponentBaseMeta class which includes validation on an additional target-encoder method `categories`"""
+    METHODS_TO_CHECK = ComponentBaseMeta.METHODS_TO_CHECK + ['get_feature_names']
+
+
+class TargetEncoder(Transformer, metaclass=TargetEncoderMeta):
+    """Target encoder to encode categorical data"""
+    name = 'Target Encoder'
+    hyperparameter_ranges = {}
+
+    def __init__(self,
```
freddyaboulton, line 17: Should we file an issue for adding a `top_n` like the OneHotEncoder? I like @rpeck 's suggestion to bin low-frequency categories into an "other" category.

freddyaboulton (approved):
@bchen1116 Looks good! Thanks for hunting down that weird bug hehe. I have a suggestion for how to guard against that bug without taking the extreme measure of resetting the index at the pipeline level. But maybe that is something we should consider in the future! Custom indices have shot us in the foot so many times lol.

evalml/pipelines/pipeline_base.py
bchen1116: mmm I agree. Yeah, I can make that change, thanks for the catch!

evalml/pipelines/components/transformers/encoders/target_encoder.py
bchen1116: Great catch, will fix

evalml/pipelines/components/transformers/encoders/target_encoder.py
bchen1116: I'm not sure if we want to do `top_n` for this. I remember in our discussion we talked about using `OHE` for smaller amounts of categorical values (for instance, < 100), while using `TE` for larger amounts of categorical values. I think this was something we would look into further, and I can file an issue for that exploration? Correct me if I'm wrong @dsherry @rpeck

freddyaboulton: Yea I don't think we should drop categories based on top_n like the OHE, rather that we group all the ones outside the top_n most frequent into the same category. 

Definitely something for future work! Just wanted to make sure we tracked it.

bchen1116: Submitted an issue [here](https://github.com/alteryx/evalml/issues/1440)!

Issue #1174, opened 2020-09-15
Simplify evalml's logging behavior


dsherry:
**Background**
Our current logging behavior sets up our default logging pattern: log debug-level stuff to a file, and info-level to stdout. The file location is configurable using the env var `EVALML_LOG_FILE`. [Here's our doc on this](https://evalml.alteryx.com/en/stable/user_guide/utilities.html#Configuring-Logging).

**Discussion**
I like that we have a default logging pattern. It makes it easy for users to get up and running with evalml.

However, what we have now is also too complex. I wouldn't expect a third-party python library to have its own logging behavior--I'd expect that I could control its behavior via the python `logging` package, i.e. changing the log level, creating a file logger, pretty-printing etc.

The question at hand: **How can we simplify our log behavior to make it easy for users to modify evalml's logging behavior?**

Use-cases:
1. User scripting with evalml directly in jupyter/other.
2. User wants to use evalml programmatically, i.e. integrated into their application.

I think we've tailored our logging setup for the first case, but haven't considered the second. In the second case, the parent application could be a python application, or a standalone application. In either case, we should make it easy and clear to change the behavior. One drawback of the current approach is that users need to learn about our env var setup, but it would probably be more intuitive for us to support setting the log level directly using `logging`.

Note this issue also applies to featuretools and the other Innovation Labs-related analytics projects. If we can come up with a good pattern for this, we can update those libraries to do the same.

dsherry:
#1273 was a dup. Note to whoever picks this up: please read #1273 !

dsherry:
@bchen1116 @freddyaboulton and I spent some time debugging a fun issue (filed as #1366). It took longer to debug because the debug log didn't include a stack trace. We should make it so that we log stack traces at debug-level during automl.

freddyaboulton:
> @bchen1116 @freddyaboulton and I spent some time debugging a fun issue (filed as #1366). It took longer to debug because the debug log didn't include a stack trace. We should make it so that we log stack traces at debug-level during automl.

For whoever picks this up, I believe the stacktrace is logged when there is a `PipelineScoreError` but not if the exception occurs outside scoring (fitting, splitting data, optimizing threshold).

dsherry:
A user brought this last point up in #1710 : if a model errors during automl, the message logged out directs users to look at `evalml_debug.log` for a stack trace, but then it turns out there's no stack trace visible there!

jeremyliweishih:
[...]

Issue #1047, opened 2020-08-12
Asynchronous Search API


dsherry:
Today, EvalML's API is synchronous. We define a couple callbacks which can be used during search to get status updates.

The idea with an asynchronous API would be something like this:
```python
automl = AutoMLSearch(...)
assert not automl.is_running
automl.start_search(X, y)
assert automl.is_running
...
# then, at any point, the latest results could be accessed
assert automl.is_running
print(automl.results)
print(automl.best_model)
pipeline = automl.best_model(parameters={})
pipeline.fit(X, y)
print('Holdout score: {}'.format(pipeline.score(X_holdout, y_holdout, objectives=['f1']))
...
# eventually, the search could be stopped if its still running
automl.stop_search()
assert not automl.is_running
```

PR #1403, opened 2020-11-03
Add callback for error handling in AutoMLSearch

### LINKED ISSUE #1217: AutoML search: Add callback for error handling
**Background**
We [removed the old `raise_errors` flag](https://github.com/alteryx/evalml/pull/936/files) from `AutoMLSearch.search` in July. This flag stopped automl search on the first error.

**Use-case**
Users want to have control over what happens when automl search encounters an exception during pipeline fit/score.

#1047 tracks building an asynchronous API for automl search. But doing this issue would still add value before we get to #1047.

This would [help with perf testing too](https://github.com/FeatureLabs/evalml_looking_glass/issues/123).

**Proposal**

```python
# four different automl error callbacks we could define
def silent_error(automl, exception):
    """No-op"""

def log_error(automl, exception):
    """doesn't throw, but logs. default behavior"""
    logger.error(f'AutoML search encountered an exception: {str(e)}')
automl = AutoMLSearch(automl_error_callback=log_error)
automl.search(X, y)

def raise_error(automl, exception):
    """throws"""
    logger.error(f'AutoML search got fatal exception: {str(e)}')
    raise exception

def log_and_save_error(automl, exception):
    """doesn't throw, but adds exception to a list in results"""
    logger.warning(f'AutoML search encountered an exception: {str(e)}')
    automl._results['errors'] = automl._results.get('errors', [])
    automl._results['errors'].append(e)
automl = AutoMLSearch(automl_error_callback=raise_error)
automl.search(X, y) 

def raise_and_save_error(automl, exception):
    """save exception and throw"""
    logger.warning(f'AutoML search encountered an exception: {str(e)}')
    automl._results['errors'] = [e]
    raise exception

# usage of any of the above:
automl = AutoMLSearch(automl_error_callback=my_callback)
automl.search(X, y) # may now throw, or continue, depending.
automl.results['errors']
```

@jeremyliweishih fyi

angela97lin:
Closes #1217

Questions:
- Should we subclass [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) to enforce that kwargs is passed? (https://stackoverflow.com/questions/64071488/callback-protocol-for-functions-with-different-kwargs)

If all objectives fail and `AutoMLException` is raised, we will raise since it indicates that something is likely wrong. This is the same behavior as what is currently implemented.


evalml/automl/automl_search.py
```diff
@@ -126,16 +128,19 @@ def __init__(self,
...
-                Passed three parameters: pipeline_class, parameters, and the AutoMLSearch object.
+            start_iteration_callback (callable): Function called before each pipeline training iteration.
+                Passed three parameters: The pipeline class, the pipeline parameters, and the AutoMLSearch object.
 
-            add_result_callback (callable): function called after each pipeline training iteration.
+            add_result_callback (callable): Function called after each pipeline training iteration.
                 Passed three parameters: A dictionary containing the training results for the new pipeline, an untrained_pipeline containing the parameters used during training, and the AutoMLSearch object.
 
+            error_callback (callable): Function called when `search()` errors and raises an Exception.
+                Passed two positional parameters: The Exception raised and the AutoMLSearch object. Accepts kwargs.
```
freddyaboulton, line 142: nit-pick: I think we should say "must accept kwargs"? Right now fold_num and pipeline are always passed as kwargs.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1621,3 +1628,46 @@ def __init__(self, parameters, random_state):
...
+    automl = AutoMLSearch(problem_type="binary", error_callback=raise_error_callback)
+    with pytest.raises(Exception, match="all your model are belong to us"):
+        automl.search(X, y)
+    assert "AutoMLSearch raised a fatal exception: all your model are belong to us" in caplog.text
+
+    caplog.clear()
+    automl = AutoMLSearch(problem_type="binary", error_callback=log_and_save_error_callback)
+    automl.search(X, y)
+    assert "AutoML search encountered an exception: all your model are belong to us" in caplog.text
+    for e in automl._results['errors']:
```
freddyaboulton, line 1663: nit-pick: I think we should check that that the expected number of results get stored.

evalml/automl/callbacks.py
```diff
@@ -0,0 +1,47 @@
...
+    pass
+
+
+def raise_error_callback(exception, automl, **kwargs):
+    """Raises the exception thrown by the AutoMLSearch object. Also logs the exception as an error."""
+    logger.error(f'AutoMLSearch raised a fatal exception: {str(exception)}')
+    raise exception
+
+
+def log_and_save_error_callback(exception, automl, **kwargs):
```
freddyaboulton, line 18: I think it would be more useful for users if we save the traceback instead of the exception?

Context: https://github.com/alteryx/evalml/issues/1174#issuecomment-718963326

jeremyliweishih: Having the traceback would be great for performance testing as well. Maybe we can save both in a tuple?

evalml/automl/callbacks.py
```diff
@@ -0,0 +1,47 @@
...
+    """Logs the exception thrown by the AutoMLSearch object as a warning
+        and adds the exception to the 'errors' list in AutoMLSearch object results."""
+    logger.warning(f'AutoML search encountered an exception: {str(exception)}')
+    automl._results['errors'] = automl._results.get('errors', [])
+    automl._results['errors'].append(exception)
+
+
+def raise_and_save_error_callback(exception, automl, **kwargs):
```
jeremyliweishih, line 26: Same thing about traceback here.

jeremyliweishih (approved):
Looks good! I agree with @freddyaboulton that we should add the traceback as part of saving the errors. Could be done in this PR or another but would be great for looking glass.

evalml/automl/automl_search.py
dsherry: Agreed about "must accept kwargs" and @jeremyliweishih suggestion. Would be helpful to describe what the kwargs part is for, how to use it. Will respond to `Protocol` discussion elsewhere.

evalml/automl/automl_search.py
```diff
@@ -674,22 +680,14 @@ def _compute_cv_scores(self, pipeline, X, y):
                 logger.debug(f"\t\t\tFold {i}: {self.objective.name} score: {scores[self.objective.name]:.3f}")
                 score = scores[self.objective.name]
             except Exception as e:
+                if self.error_callback is not None:
+                    self.error_callback(e, self, fold_num=i, pipeline=pipeline)
```
dsherry, line 684: Cool, so because we pass `self` back in the callback, someone could theoretically write a callback which responds to failure and calls `automl.search` again. So this is implementing one part of an asynchronous API (#1047 ). Neat!

evalml/automl/callbacks.py
```diff
@@ -0,0 +1,47 @@
+from evalml.exceptions import PipelineScoreError
+from evalml.utils.logger import get_logger
+
+logger = get_logger(__file__)
+
+
+def silent_error_callback(exception, automl, **kwargs):
+    """No-op."""
+    pass
```
dsherry, line 9: Delete `pass`? I think it messes up codecov and you can just do
```python
def silent_error_callback(exception, automl, **kwargs):
    """No-op."""
```

evalml/automl/callbacks.py
dsherry: Can't you compute the traceback given the exception? Or is it true that once you leave scope you can't get the full stack trace anymore?

I suppose either way its not expensive to compute the traceback, so I agree we should do so. Perhaps in that case the args to each callback could be: automl object, exception, traceback, kwargs

evalml/automl/callbacks.py
```diff
@@ -0,0 +1,47 @@
...
+    """Raises the exception thrown by the AutoMLSearch object. Also logs the exception as an error."""
+    logger.error(f'AutoMLSearch raised a fatal exception: {str(exception)}')
+    raise exception
+
+
+def log_and_save_error_callback(exception, automl, **kwargs):
+    """Logs the exception thrown by the AutoMLSearch object as a warning
+        and adds the exception to the 'errors' list in AutoMLSearch object results."""
+    logger.warning(f'AutoML search encountered an exception: {str(exception)}')
+    automl._results['errors'] = automl._results.get('errors', [])
```
dsherry, line 22: Ah, slick.

If I had to choose between this and adding code to `AutoMLSearch.__init__` which initializes "errors" key in automl results to an empty list, I think I'd choose the latter, simply because that way there's *always* an "errors" key in the results, even if its empty, and that feels like simpler behavior to me.

evalml/automl/callbacks.py
```diff
@@ -0,0 +1,47 @@
...
+    pipeline = kwargs.get('pipeline')
+    if isinstance(exception, PipelineScoreError):
+        logger.info(f"\t\t\tFold {fold_num}: Encountered an error scoring the following objectives: {', '.join(exception.exceptions)}.")
+        logger.info(f"\t\t\tFold {fold_num}: The scores for these objectives will be replaced with nan.")
+    else:
+        logger.info(f"\t\t\tFold {fold_num}: Encountered an error.")
+        logger.info(f"\t\t\tFold {fold_num}: All scores will be replaced with nan.")
+    logger.info(f"\t\t\tFold {fold_num}: Please check {logger.handlers[1].baseFilename} for the current hyperparameters and stack trace.")
+    logger.debug(f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.hyperparameters}")
+    logger.info(f"\t\t\tFold {fold_num}: Exception during automl search: {str(exception)}")
```
dsherry, line 47: 👍  I guess there's no easy way to avoid having to do the tabbing here. Lgtm

evalml/automl/automl_search.py
dsherry, line 142: @angela97lin let's also explain that `None` defaults to `log_error_callback`

dsherry (approved):
@angela97lin this looks great!

I left a few comments, but there was nothing major.

RE `Protocol` subclassing, I think we're ok to skip that for now, but its a good idea.

evalml/automl/automl_search.py
angela97lin: Yes!!

evalml/automl/callbacks.py
angela97lin: Thanks for the feedback guys! I updated s.t. the arguments are automl object, exception, traceback, kwargs and both the traceback and exception are logged/saved :)

PR #1411, opened 2020-11-05
make_pipeline_from_components accepts a random state

### LINKED ISSUE #1338: make_pipeline_from_components does not accept a random state
Users who call `make_pipeline_from_components` will receive a pipeline with `random_state=0` every time. Users may want to control the randomness of their pipeline so we should add this parameter.

freddyaboulton:
### Pull Request Description
Fix #1338

-----

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -72,7 +72,8 @@ class IterativeAlgorithm(AutoMLAlgorithm):
             for pipeline_dict in self._best_pipeline_info.values():
                 pipeline_class = pipeline_dict['pipeline_class']
                 pipeline_params = pipeline_dict['parameters']
-                input_pipelines.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, pipeline_params)))
+                input_pipelines.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, pipeline_params),
+                                                      random_state=self.random_state))
             ensemble = _make_stacked_ensemble_pipeline(input_pipelines, input_pipelines[0].problem_type,
                                                        random_state=self.random_state)
             next_batch.append(ensemble)
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -73,7 +73,8 @@ class IterativeAlgorithm(AutoMLAlgorithm):
                 pipeline_class = pipeline_dict['pipeline_class']
                 pipeline_params = pipeline_dict['parameters']
                 input_pipelines.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, pipeline_params)))
-            ensemble = _make_stacked_ensemble_pipeline(input_pipelines, input_pipelines[0].problem_type)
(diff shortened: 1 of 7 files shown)


evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -300,11 +301,12 @@ def test_make_pipeline_from_components(X_y_binary, logistic_regression_binary_pi
     with pytest.raises(TypeError, match="Every element of `component_instances` must be an instance of ComponentBase"):
         make_pipeline_from_components(['RandomForestClassifier'], problem_type='binary')
 
-    imp = Imputer(numeric_impute_strategy='median')
-    est = RandomForestClassifier()
-    pipeline = make_pipeline_from_components([imp, est], ProblemTypes.BINARY, custom_name='My Pipeline')
-    components_list = pipeline.component_graph
-    assert components_list == [imp, est]
+    imp = Imputer(numeric_impute_strategy='median', random_state=5)
+    est = RandomForestClassifier(random_state=7)
```
dsherry, line 305: Why 5 above and 7 here?

dsherry (approved):
@freddyaboulton LGTM! Left one question

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: No strong reason other than I wanted coverage for the case when the components the user passes in components for different random states!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -73,7 +73,8 @@ def next_batch(self):
                 pipeline_class = pipeline_dict['pipeline_class']
                 pipeline_params = pipeline_dict['parameters']
                 input_pipelines.append(pipeline_class(parameters=self._transform_parameters(pipeline_class, pipeline_params)))
-            ensemble = _make_stacked_ensemble_pipeline(input_pipelines, input_pipelines[0].problem_type)
+            ensemble = _make_stacked_ensemble_pipeline(input_pipelines, input_pipelines[0].problem_type,
+                                                       random_state=self.random_state)
```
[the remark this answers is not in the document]
angela97lin: Ooo good catch indeed. @dsherry the [sklearn stacked ensembler](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.StackingClassifier.html#sklearn.ensemble.StackingClassifier) doesn't accept `random_state` as a parameter, so its enforced by the passed input estimators. So if `random_state` is passed there then yes!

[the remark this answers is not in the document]
freddyaboulton: Great catch guys! The StackedEnsembles don't overwrite the random state of the input pipelines. The random state is only used for the CV. I added unit tests to document that behavior.

Since we want the pipelines in the stacked ensemble to have the random state of the `IterativeAlgorithm` - I now pass the random state to each pipeline as they are created (before `_make_stacked_ensemble_pipeline`).

evalml/tests/pipeline_tests/test_pipelines.py
dsherry: Ah gotcha. I was confused that they were different but then reread, got it, they don't need to be the same here, 👍 
 thanks!

PR #1388, opened 2020-10-31
Fix stacked ensemble and LightGBM errors in AutoMLSearch

angela97lin:
Closes #1376

PR #1374, opened 2020-10-30
Displaying the top_k features with largest shap value magnitudes

freddyaboulton:
### Pull Request Description
Fixes #1360

FYI @RezaAkhtar

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #1413, opened 2020-11-06
Try rounding codecov down instead of up

dsherry:
Perhaps I had this backwards previously? I hope this helps with the recent codecov flakes we've been seeing occur when people delete lines of code.

https://docs.codecov.io/docs/codecovyml-reference#coverageround

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,5 +1,5 @@
 coverage:
-  round: down
+  round: up
   precision: 1
 
 codecov:
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,3 +1,3 @@
 coverage:
-  round: up
+  round: down
   precision: 2
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,6 +1,10 @@
 coverage:
   round: up
   precision: 1
+  status:
+    project:
+      default:
+        threshold: 1%
 
 codecov:
   notify:
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,3 +1,3 @@
 coverage:
   round: down
-  precision: 2
+  precision: 1
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,3 +1,7 @@
 coverage:
   round: down
   precision: 1
+
+codecov:
+  notify:
+    wait_for_ci: true
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -14,6 +14,7 @@ Release Notes
         * Added description of CLA to contributing guide, updated description of draft PRs :pr:`1402`
     * Testing Changes
         * Removed ``category_encoders`` from test-requirements.txt :pr:`1373`
+        * Tweak codecov.io settings again to avoid flakes :pr:`1413`
 
 **v0.15.0 Oct. 29, 2020**
     * Enhancements


.github/.codecov.yml
```diff
@@ -1,3 +1,3 @@
 coverage:
-  round: up
+  round: down
```
freddyaboulton, line 2: Maybe we need to change the precision? I think rounding down is the default behavior and we were seeing issues like what is happening in #1388 before we changed to rounding up.

dsherry: @freddyaboulton ah you are correct, this is the default behavior...

freddyaboulton: And the same thing is happening in #1374 😂

.github/.codecov.yml
```diff
@@ -1,3 +1,3 @@
 coverage:
-  round: up
-  precision: 2
+  round: down
+  precision: 1
```
dsherry, line 3: @freddyaboulton ok let's try this :)

.github/.codecov.yml
dsherry: The codecov.io docs are super unclear about what "round" means:
> Which direction to you want to round the coverage value

They should have included an example... what's their algorithm for determining whether or not a job passes codecov? I think the algorithm is
* Compute coverage percentage before and after
* Take the signed difference, after - before
* Round that difference using the `round` policy, then truncate using `precision` digits after the decimal
* If the resulting number is less than 0, fail

But... I haven't found their docs which say whether or not this is true, hence further trial and error

angela97lin: And in #1388 😅

dsherry: Oops [I spoke too soon](https://docs.codecov.io/docs/frequently-asked-questions#how-is-coverage-calculated):

```
round((hits / (hits + partials + misses)) * 100, 5) = Coverage Ratio
```
According to this, rounding up means a higher coverage ratio, which I think is what we want. So I'll decrease the precision but leave this set to "up". Does that make sense?

.github/.codecov.yml
```diff
@@ -1,3 +1,7 @@
...
+
+codecov:
+  notify:
+    wait_for_ci: true
```
dsherry, line 7: [This](https://docs.codecov.io/docs/common-recipe-list#send-coverage-notifications-even-if-all-ci-statuses-have-not-completed) will theoretically only run a codecov report once CI is green. We'll see :)

freddyaboulton: No more premature +/ 8% changes in coverage? 😃

jeremyliweishih:
I think we can also try to widen the [threshold](https://docs.codecov.io/docs/codecovyml-reference#coveragestatus) and see how that goes.

.github/.codecov.yml
dsherry: @freddyaboulton I sure hope not 😆

freddyaboulton:
@jeremyliweishih I think that is also worth looking into! I don't think rounding can save us if we ever decide to delete a whole file or something like that.

.github/.codecov.yml
```diff
@@ -1,3 +1,11 @@
 coverage:
   round: up
-  precision: 2
+  precision: 1
+  status:
+    project:
+      default:
+        threshold: 1%
```
dsherry, line 7: @jeremyliweishih good point! Let's try it out. I haven't been able to find docs for `threshold` other than [the link you found](https://docs.codecov.io/docs/common-recipe-list#set-status-checks-to-block-coverage).

I saw someone [on their forum used this](https://community.codecov.io/t/threshold-confusion/88/10) and didn't complain about it breaking, beyond some sort of caching bug which is hopefully now irrelevant.

freddyaboulton (approved):
🤞

dsherry:
I just tested this out in a PR #1414 , deleted a unit test to trigger a <0.01% coverage drop, and the codecov job was marked as passing!

We may wanna decrease the threshold to 0.1% or something, but this is fine for now. Merging soon!

PR #1414, opened 2020-11-06
[DO NOT MERGE] codecov test

dsherry:
Deleting a test to see what happens with codecov. Based off branch from #1413 which is trying to change coverage thresholding/behavior

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,5 +1,5 @@
 coverage:
-  round: down
+  round: up
   precision: 1
 
 codecov:
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,3 +1,3 @@
 coverage:
-  round: up
+  round: down
   precision: 2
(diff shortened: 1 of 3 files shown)


freddyaboulton:
@dsherry I don't think test lines are counted as hits? I think we need to merge in your change to one of our already failing PRs  and give it a go.

dsherry:
@freddyaboulton the way codecov works is that our linux unit tests generate `coverage` reports, by running the tests and counting execution of each line in our codebase. Then we pass those `coverage` reports up to codecov to aggregate (sum across both unit test runs, with and without our extra dependencies and tests for those dependencies).

So for our testing here: the more unit tests we delete, the more coverage drops.

I think this PR confirms that our tweaks in #1413 worked. The [codecov report ](https://codecov.io/gh/alteryx/evalml/compare/4c8de526a79bdadeeffec2bcc7819aec84552417...f2503baca80b57b94298afe1d0aad6de106eb3e5) shows that coverage dropped by <0.01%, yet the github actions job is marked as passing. That's good!

freddyaboulton:
@dsherry Thanks for explaining!

PR #1417, opened 2020-11-09
Update docstrings from `np.array` to `np.ndarray`

### LINKED ISSUE #1394: Update documentation from np.array to np.ndarray
Nuanced and maybe unnecessary, but np.array refers to [the method used to create an array](https://numpy.org/doc/stable/reference/generated/numpy.array.html) and np.ndarray is the actual data structure.

angela97lin:
Closes #1394; np.array refers to [the method used to create an array](https://numpy.org/doc/stable/reference/generated/numpy.array.html) and np.ndarray is the actual data structure.


freddyaboulton (approved):
Nice! I think you caught them all!

dsherry (approved):
Good call! 🚢

Issue #1382, opened 2020-10-30
Add a problem_configuration parameter to AutoMLSearch


freddyaboulton:
Requirements:
1. The `problem_configuration` parameters should be passed to the pipelines proposed by the search algorithm and passed to the components that need them.

PR #1418, opened 2020-11-09
Timeseries regression pipeline

### LINKED ISSUE #1380: Add a TimeSeriesRegressionPipeline
Requirements:

1. Should have a time series regression problem type
2. `fit` should drop rows with nan values in the features (due to lagging) or rows with nan target value due to lag.
3. `predict` should return nans for the rows that have nan values in the features due to lagging
4. `score` should drop rows with nans in the predictions and target before passing to the objective

freddyaboulton:
### Pull Request Description
Fix #1380

-----

### DIFF
--- a/evalml/pipelines/components/ensemble/stacked_ensemble_regressor.py
+++ b/evalml/pipelines/components/ensemble/stacked_ensemble_regressor.py
@@ -11,7 +11,7 @@ class StackedEnsembleRegressor(StackedEnsembleBase):
     """Stacked Ensemble Regressor."""
     name = "Stacked Ensemble Regressor"
     model_family = ModelFamily.ENSEMBLE
-    supported_problem_types = [ProblemTypes.REGRESSION]
+    supported_problem_types = [ProblemTypes.REGRESSION, ProblemTypes.TIME_SERIES_REGRESSION]
     hyperparameter_ranges = {}
     _stacking_estimator_class = StackingRegressor
     _default_final_estimator = LinearRegressor
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -190,25 +190,24 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         """
         X_t = X
         for component in self.component_graph[:-1]:
-            X_t = component.transform(X_t, y)
+            X_t = component.transform(X_t, y=y)
         return X_t
 
     def _compute_features_during_fit(self, X, y):
         X_t = X
-        y_t = y
         for component in self.component_graph[:-1]:
             self.input_feature_names.update({component.name: list(pd.DataFrame(X_t))})
-            X_t = component.fit_transform(X_t, y_t)
+            X_t = component.fit_transform(X_t, y=y)
 
         self.input_feature_names.update({self.estimator.name: list(pd.DataFrame(X_t))})
 
-        return X_t, y_t
+        return X_t
 
     def _fit(self, X, y):
 
-        X_t, y_t = self._compute_features_during_fit(X, y)
+        X_t = self._compute_features_during_fit(X, y)
 
-        self.estimator.fit(X_t, y_t)
+        self.estimator.fit(X_t, y)
 
     @abstractmethod
     def fit(self, X, y):
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -75,6 +75,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
 
         self._validate_estimator_problem_type()
         self._is_fitted = False
+        self._pipeline_params = parameters.get("pipeline", {})
 
     @classproperty
     def name(cls):
@@ -309,7 +310,10 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         Returns:
             dict: Dictionary of all component parameters
         """
-        return {c.name: copy.copy(c.parameters) for c in self.component_graph if c.parameters}
+        component_parameters = {c.name: copy.copy(c.parameters) for c in self.component_graph if c.parameters}
+        if self._pipeline_params:
+            component_parameters['pipeline'] = self._pipeline_params
+        return component_parameters
 
     @classproperty
     def default_parameters(cls):
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -179,7 +179,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
             logger.info(component_string)
             component.describe(print_name=False)
 
-    def compute_estimator_features(self, X):
+    def compute_estimator_features(self, X, y=None):
         """Transforms the data by applying all pre-processing components.
 
         Arguments:
@@ -190,10 +190,10 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         """
         X_t = X
         for component in self.component_graph[:-1]:
-            X_t = component.transform(X_t)
+            X_t = component.transform(X_t, y)
         return X_t
 
-    def _fit(self, X, y):
+    def _compute_features_during_fit(self, X, y):
         X_t = X
         y_t = y
         for component in self.component_graph[:-1]:
@@ -201,6 +201,13 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
             X_t = component.fit_transform(X_t, y_t)
 
         self.input_feature_names.update({self.estimator.name: list(pd.DataFrame(X_t))})
+
+        return X_t, y_t
+
+    def _fit(self, X, y):
+
+        X_t, y_t = self._compute_features_during_fit(X, y)
+
         self.estimator.fit(X_t, y_t)
 
     @abstractmethod
@@ -228,7 +235,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         """
         X = _convert_to_woodwork_structure(X)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        X_t = self.compute_estimator_features(X)
+        X_t = self.compute_estimator_features(X, y=None)
         return self.estimator.predict(X_t)
 
     @abstractmethod
--- a/evalml/pipelines/time_series.py
+++ b/evalml/pipelines/time_series.py
@@ -1,0 +1,53 @@
+import pandas as pd
+
+from evalml.objectives import get_objective
+from evalml.pipelines.regression_pipeline import RegressionPipeline
+from evalml.problem_types import ProblemTypes
+from evalml.utils.gen_utils import any_values_are_nan, drop_nan, pad_with_nans
+
+
+class TimeSeriesRegressionPipeline(RegressionPipeline):
+    problem_type = ProblemTypes.TIME_SERIES_REGRESSION
+
+    def __init__(self, parameters, gap, max_delay, random_state=0):
+        super().__init__(parameters, random_state)
+        self.gap = gap
+        self.max_delay = max_delay
+
+    def fit(self, X, y=None):
+        if y is None:
+            assert isinstance(X, pd.Series), "When modeling a single time series, the data must be a series."
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+
+        if y is None:
+            X_t, _ = self._compute_features_during_fit(X, y=None)
+            y = X.squeeze()
+        else:
+            y = pd.Series(y)
+            X_t, _ = self._compute_features_during_fit(X, y)
+
+        y_shifted = y.shift(-self.gap)
+        X_t, y_shifted = drop_nan(X_t, y_shifted)
+        self.estimator.fit(X_t, y_shifted)
+        return self
+
+    def predict(self, X, y=None, objective=None):
+        features = self.compute_estimator_features(X, y)
+        predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
+        if any_values_are_nan(features):
+            return pad_with_nans(predictions, self.max_delay)
+        else:
+            return predictions
+
+    def score(self, X, y, objectives):
+        y_predicted = self.predict(X, y)
+        if y is None:
+            y = X.copy()
+        y_shifted = y.shift(-self.gap)
+        objectives = [get_objective(o, return_instance=True) for o in objectives]
+        y_shifted, y_predicted = drop_nan(y_shifted, y_predicted)
+        return self._score_all_objectives(X, y_shifted,
+                                          y_predicted,
+                                          y_pred_proba=None,
+                                          objectives=objectives)
--- a/evalml/pipelines/time_series.py
+++ b/evalml/pipelines/time_series.py
@@ -15,6 +15,15 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         self.max_delay = max_delay
 
     def fit(self, X, y=None):
+        """Fit a time series regression pipeline.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame or np.ndarray): The input training data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series, np.ndarray): The target training targets of length [n_samples]
+
+        Returns:
+            self
+        """
         if y is None:
             assert isinstance(X, pd.Series), "When modeling a single time series, the data must be a series."
         if not isinstance(X, pd.DataFrame):
@@ -33,6 +42,16 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         return self
 
     def predict(self, X, y=None, objective=None):
+        """Make predictions using selected features.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, or np.ndarray): Data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series, np.ndarray, None): The target training targets of length [n_samples]
+            objective (Object or string): The objective to use to make predictions
+
+        Returns:
+            pd.Series: Predicted values.
+        """
         features = self.compute_estimator_features(X, y)
         predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
         if any_values_are_nan(features):
@@ -41,6 +60,16 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
             return predictions
 
     def score(self, X, y, objectives):
+        """Evaluate model performance on current and additional objectives.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame or np.ndarray): Data of shape [n_samples, n_features]
+            y (pd.Series, ww.DataColumn): True labels of length [n_samples]
+            objectives (list): Non-empty list of objectives to score on
+
+        Returns:
+            dict: Ordered dictionary of objective scores
+        """
         y_predicted = self.predict(X, y)
         if y is None:
             y = X.copy()
--- a/evalml/pipelines/time_series.py
+++ b/evalml/pipelines/time_series.py
@@ -24,18 +24,14 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         Returns:
             self
         """
-        if y is None:
-            assert isinstance(X, pd.Series), "When modeling a single time series, the data must be a series."
         if not isinstance(X, pd.DataFrame):
             X = pd.DataFrame(X)
 
-        if y is None:
-            X_t, _ = self._compute_features_during_fit(X, y=None)
-            y = X.squeeze()
+        if X.empty:
+            X_t, _ = self._compute_features_during_fit(X=None, y=y)
         else:
             y = pd.Series(y)
             X_t, _ = self._compute_features_during_fit(X, y)
-
         y_shifted = y.shift(-self.gap)
         X_t, y_shifted = drop_nan(X_t, y_shifted)
         self.estimator.fit(X_t, y_shifted)
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -51,8 +51,8 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
             X = pd.DataFrame()
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_pandas())
-        y = _convert_woodwork_types_wrapper(y.to_pandas())
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
 
         X_t = self._compute_features_during_fit(X, y)
 
@@ -76,8 +76,8 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
             X = pd.DataFrame()
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_pandas())
-        y = _convert_woodwork_types_wrapper(y.to_pandas())
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
 
         features = self.compute_estimator_features(X, y)
         predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
@@ -98,7 +98,7 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         if X is None:
             X = pd.DataFrame()
         X = _convert_to_woodwork_structure(X)
-        X = _convert_woodwork_types_wrapper(X.to_pandas())
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
         y_predicted = self.predict(X, y)
         y_shifted = y.shift(-self.gap)
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -37,7 +37,7 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         self.max_delay = pipeline_params['max_delay']
         super().__init__(parameters, random_state)
 
-    def fit(self, X, y=None):
+    def fit(self, X, y):
         """Fit a time series regression pipeline.
 
         Arguments:
@@ -54,7 +54,6 @@ class TimeSeriesRegressionPipeline(RegressionPipeline):
         X = _convert_woodwork_types_wrapper(X.to_pandas())
         y = _convert_woodwork_types_wrapper(y.to_pandas())
 
-        y = pd.Series(y)
         X_t = self._compute_features_during_fit(X, y)
 
         y_shifted = y.shift(-self.gap)
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -6,7 +6,7 @@ from evalml.problem_types import ProblemTypes
 from evalml.utils.gen_utils import (
     _convert_to_woodwork_structure,
     _convert_woodwork_types_wrapper,
-    keep_non_nan_rows,
+    drop_rows_with_nans,
     pad_with_nans
 )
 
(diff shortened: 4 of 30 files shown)


evalml/pipelines/components/ensemble/stacked_ensemble_regressor.py
```diff
@@ -11,7 +11,7 @@ class StackedEnsembleRegressor(StackedEnsembleBase):
     """Stacked Ensemble Regressor."""
     name = "Stacked Ensemble Regressor"
     model_family = ModelFamily.ENSEMBLE
-    supported_problem_types = [ProblemTypes.REGRESSION]
+    supported_problem_types = [ProblemTypes.REGRESSION, ProblemTypes.TIME_SERIES_REGRESSION]
```
freddyaboulton, line 14: Need to make this change because of `_validate_estimator_problem_type`

evalml/pipelines/pipeline_base.py
```diff
@@ -190,17 +190,24 @@ def compute_estimator_features(self, X):
         """
         X_t = X
         for component in self.component_graph[:-1]:
-            X_t = component.transform(X_t)
+            X_t = component.transform(X_t, y)
         return X_t
 
-    def _fit(self, X, y):
+    def _compute_features_during_fit(self, X, y):
```
freddyaboulton, line 196: Need to make this small refactor because we need to make sure the Nans are dropped before the estimator is fit. I think it makes the most sense to do that in the time series pipeline as opposed to the base class.

evalml/pipelines/utils.py
```diff
@@ -81,6 +82,8 @@ def _get_pipeline_base_class(problem_type):
         return MulticlassClassificationPipeline
     elif problem_type == ProblemTypes.REGRESSION:
         return RegressionPipeline
+    elif problem_type == ProblemTypes.TIME_SERIES_REGRESSION:
```
freddyaboulton, line 85: Need to make this change so that the unit tests for make_pipeline don't error out for TIME_SERIES_REGRESSION problem types. 

I'll modify the logic of make_pipeline to handle time series when we integrate time series regression into automl search in #1382

evalml/tests/component_tests/test_utils.py
```diff
@@ -60,6 +60,9 @@ def test_scikit_learn_wrapper(X_y_binary, X_y_multi, X_y_regression):
                 num_classes = 3
             elif problem_type == ProblemTypes.REGRESSION:
                 X, y = X_y_regression
+            elif problem_type == ProblemTypes.TIME_SERIES_REGRESSION:
+                # Skipping because make_pipeline_from_components does not yet work for time series.
```
freddyaboulton, line 64: The reason is because of the pipeline-level paramters of gap and max_delay. I imagine we'll use the same pattern to specify them that we will introduce in #1382 .

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -0,0 +1,189 @@
...
+
+@pytest.fixture
+def ts_data():
+    X, y = pd.DataFrame({"features": range(101, 132)}), pd.Series(range(1, 32))
+    y.index = pd.date_range("2020-10-01", "2020-10-31")
+    X.index = pd.date_range("2020-10-01", "2020-10-31")
+    return X, y
+
+
+class MockDelayedFeatures(Transformer):
```
freddyaboulton, line 19: A lot of the complexity from time series comes from the nans that are introduced into the features. We need a component that does this to make the tests more robust. I'll delete this in favor of the bona-fide DelayedFeaturesTransformer when #1379 is merged.

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+            self
+        """
+        if y is None:
+            assert isinstance(X, pd.Series), "When modeling a single time series, the data must be a series."
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+
+        if y is None:
+            X_t, _ = self._compute_features_during_fit(X, y=None)
+            y = X.squeeze()
```
dsherry, line 34: Shouldn't `y` just remain `None` if it wasn't provided to `fit`?

evalml/pipelines/pipeline_base.py
```diff
@@ -190,17 +190,24 @@ def compute_estimator_features(self, X):
...
         X_t = X
         y_t = y
         for component in self.component_graph[:-1]:
             self.input_feature_names.update({component.name: list(pd.DataFrame(X_t))})
             X_t = component.fit_transform(X_t, y_t)
 
         self.input_feature_names.update({self.estimator.name: list(pd.DataFrame(X_t))})
+
+        return X_t, y_t
```
dsherry, line 205: Why return `y_t` here?

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            y = pd.Series(y)
```
dsherry, line 36: Can you move this out, like we're doing for X online 29? Will make it easier to normalize to dataframe when that comes in

dsherry: Ooh actually @freddyaboulton we should be converting these to woodwork now that @angela97lin merged #1393. Check out the code in the other pipeline classes. That just got merged recently.

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+            X_t, _ = self._compute_features_during_fit(X, y)
+
+        y_shifted = y.shift(-self.gap)
+        X_t, y_shifted = drop_nan(X_t, y_shifted)
```
dsherry, line 40: Nice.

Perhaps `drop_nan_rows` or `drop_rows_with_nans` would be a good name for this helper function

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            pd.Series: Predicted values.
+        """
+        features = self.compute_estimator_features(X, y)
+        predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
+        if any_values_are_nan(features):
+            return pad_with_nans(predictions, self.max_delay)
+        else:
+            return predictions
```
dsherry, line 60: Why not just always run `pad_with_nans`, and in the case where there are no nans, nothing happens? Maybe I'm not following this code right yet

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, or np.ndarray): Data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series, np.ndarray, None): The target training targets of length [n_samples]
+            objective (Object or string): The objective to use to make predictions
+
+        Returns:
+            pd.Series: Predicted values.
+        """
+        features = self.compute_estimator_features(X, y)
+        predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
```
dsherry, line 56: Should we use the `drop_nan` helper here too? Maybe I'm just confused about the names.

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+            X (ww.DataTable, pd.DataFrame or np.ndarray): Data of shape [n_samples, n_features]
+            y (pd.Series, ww.DataColumn): True labels of length [n_samples]
+            objectives (list): Non-empty list of objectives to score on
+
+        Returns:
+            dict: Ordered dictionary of objective scores
+        """
+        y_predicted = self.predict(X, y)
+        if y is None:
+            y = X.copy()
```
dsherry, line 75: Why do this? If y isn't provided, that's an error...

dsherry (changes requested):
Cool! I read the impl and left some thoughts and questions.

evalml/pipelines/time_series.py
freddyaboulton: This is for the case when there are no features and only the y is used. In #1396 we decided to use X=None for that and I just pushed that change up! So it should hopefully be clearer now.

evalml/pipelines/time_series.py
freddyaboulton: This is for the case where there are no features and only y is used but I just cleaned this up!

evalml/pipelines/pipeline_base.py
freddyaboulton: Good point! We don't need this.

evalml/pipelines/time_series.py
freddyaboulton: `predictions` will never have NaNs. I'm doing this because our design says that the data passed into this method and the predictions should always have the same number of rows. Since we drop NaN rows before calling `predict` on the estimator, I need to pad with NaNs. 

There are many ways to do this:

1.
```python
return pad_with_nans(predictions, self.max_delay if any_values_are_nan(features) else 0)
```

2.
```python
if features.shape[0] != len(predictions):
    return pad_with_nans(predictions, self.max_delay)
else:
    return predictions
```

3. 
```python
return pad_with_nans(predictions, self.max_delay if features.shape[0] != len(predictions) else 0)
```

Which do you think is best? Open to suggestions!

evalml/pipelines/time_series.py
freddyaboulton: Maybe I should rename the `drop_nan` helper `keep_non_nan_rows`. Since the target and features will have NaNs in different places in `score`, we need a way for selecting the rows that are not NaN in both of them. As far as I know, there isn't a pandas method that does this. 

In this case, we just need to drop the NaN rows from a dataframe and pandas has a built-in for it.

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
+import pandas as pd
```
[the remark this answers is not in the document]
freddyaboulton: I was thinking of adding all of the time series pipelines in one file but let's stick with the regression vs classification separation we have with our files!

evalml/pipelines/time_series.py
freddyaboulton: Definitely agree that this needs a clearer name. Let me know what you think of my suggestion below to use `keep_non_nan_rows`.

evalml/pipelines/time_series.py
```diff
@@ -0,0 +1,82 @@
...
+from evalml.objectives import get_objective
+from evalml.pipelines.regression_pipeline import RegressionPipeline
+from evalml.problem_types import ProblemTypes
+from evalml.utils.gen_utils import any_values_are_nan, drop_nan, pad_with_nans
+
+
+class TimeSeriesRegressionPipeline(RegressionPipeline):
+    problem_type = ProblemTypes.TIME_SERIES_REGRESSION
+
+    def __init__(self, parameters, gap, max_delay, random_state=0):
```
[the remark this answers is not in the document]
freddyaboulton: I was thinking of doing that once we got to the AutoML integration but it's way better to do this now. I just pushed this change up. The one thing I'd like to hear your thoughts on is whether we should raise an exception when the `pipeline` key is missing. I vote yes because it doesn't make sense to specify default arguments for these pipeline-level arguments.

evalml/pipelines/time_series.py
freddyaboulton: Thank you for reminding me! Just pushed this up.

evalml/pipelines/time_series.py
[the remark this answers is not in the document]
dsherry: Thanks!

Yeah, the timeseries pipelines need those timeseries parameters in order to function. So yes, if `gap` and `max_delay` aren't included in the `parameters` dict under the `pipeline` key, I think we should raise an exception.

Should we prefix these params with `timeseries`? I.e. `timeseries_gap`, `timeseries_max_delay`. An argument for this would be that we may add other pipeline-level params in the future.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Returns:
+            self
+        """
+        if X is None:
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_pandas())
+        y = _convert_woodwork_types_wrapper(y.to_pandas())
```
dsherry, line 55: Nice! @angela97lin does this look good to you?

evalml/pipelines/time_series.py
dsherry: Yep I like that name. Right, I don't think there's a convenient builtin in pandas for that particular use-case. I've always either used `dropna` or `isnan`.

evalml/pipelines/time_series_regression_pipeline.py
dsherry, line 55: Nice! @angela97lin does this look good to you?

angela97lin: Yup, looks good!

evalml/pipelines/time_series.py
freddyaboulton: Great! I changed the name

evalml/pipelines/time_series.py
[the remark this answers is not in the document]
freddyaboulton: Can you explain what you mean? This is how a user would instantiate a TS regression pipeline right now:

```python
    pl = Pipeline({"Delayed Feature Transformer": {"gap": gap, "max_delay": max_delay,
                                                   "delay_features": include_delayed_features,
                                                   "delay_target": include_delayed_features},
                   "pipeline": {"gap": gap, "max_delay": max_delay}})
```

I think renaming to the "pipeline" dict keys to "timeseries_gap" and "timeseries_max_delay" is bit too verbose. We can  add more pipeline level params to the "pipeline" dict as we need without conflicts.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, parameters, random_state=0):
+        """Machine learning pipeline for time series regression problems made out of transformers and a classifier.
+
+        Required Class Variables:
+            component_graph (list): List of components in order. Accepts strings or ComponentBase subclasses in the list
+
+        Arguments:
+            parameters (dict): Dictionary with component names as keys and dictionary of that component's parameters as values.
+                 An empty dictionary {} implies using all default values for component parameters. Pipeline-level
+                 parameters such as gap and max_delay must be specified as a dictionary with the "pipeline" key.
```
angela97lin, line 29: "must be specified in a dictionary with the "pipeline" key"? Hmm. This was a tad bit confusing to me--not sure best way to clear it up via wording. Maybe would be good to give an example?

evalml/utils/gen_utils.py
```diff
@@ -314,3 +314,48 @@ def _convert_woodwork_types_wrapper(pd_data):
...
+
+def any_values_are_nan(pd_data):
+    """Determine if any rows have at least one NaN.
+    Args:
+        pd_data (pd.Dataframe or pd.Series): Pandas data to check.
+
+    Returns:
+        pd.Series with bool values. Each value corresponds to if any values in the row are NaN.
+    """
+    return pd_data.isna().values.any()
```
angela97lin, line 361: Let's either specify these in the API docs or keep them private? :d

evalml/utils/gen_utils.py
```diff
@@ -314,3 +314,48 @@ def _convert_woodwork_types_wrapper(pd_data):
...
+        else:
+            return ~pd_data.isna().any(axis=1).values
+
+    mask = np.logical_and(_not_nan(pd_data_1), _not_nan(pd_data_2))
+    return pd_data_1.iloc[mask], pd_data_2.iloc[mask]
+
+
+def any_values_are_nan(pd_data):
+    """Determine if any rows have at least one NaN.
+    Args:
```
angela97lin, line 355: Args --> Arguments

evalml/utils/gen_utils.py
angela97lin, line 361: Arguably necessary/not since it's just one line? :d

evalml/utils/gen_utils.py
```diff
@@ -314,3 +314,48 @@ def _convert_woodwork_types_wrapper(pd_data):
...
+    if isinstance(pd_data, pd.Series):
+        padding = pd.Series([None] * num_to_pad)
+    else:
+        padding = pd.DataFrame({col: [None] * num_to_pad
+                                for col in pd_data.columns})
+    return pd.concat([padding, pd_data], ignore_index=True).infer_objects()
+
+
+def keep_non_nan_rows(pd_data_1, pd_data_2):
+    """Keep nows that do not have any NaNs in both pd_data_1 and pd_data_2.
```
angela97lin, line 337: Keep nows --> Keep rows

angela97lin (commented):
Looking good, just left a few docstring comments / questions for my own curiosity haha

evalml/pipelines/time_series.py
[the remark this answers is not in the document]
dsherry: Ok, I agree! Let's stick with the example you've provided here.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+            random_state (int, np.random.RandomState): The random seed/state. Defaults to 0.
+        """
+        if "pipeline" not in parameters:
+            raise ValueError("gap and max_delay parameters cannot be omitted from the parameters dict. "
+                             "Please specify them as a dictionary with the key 'pipeline'.")
+        pipeline_params = parameters.pop("pipeline")
+        self.gap = pipeline_params['gap']
+        self.max_delay = pipeline_params['max_delay']
+        super().__init__(parameters, random_state)
```
dsherry, line 38: If we do `parameters.pop` as above, the pipeline-level params won't be saved in the `parameters` attribute of the pipeline instance. Can we make it so those parameters are saved just like any other component's parameters?

angela97lin: Agreed; we should add a test to make sure `gap` and `max_delay` are in the parameters dict after initialization to check for this

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+        """Fit a time series regression pipeline.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame or np.ndarray): The input training data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series, np.ndarray): The target training targets of length [n_samples]
+
+        Returns:
+            self
+        """
+        if X is None:
```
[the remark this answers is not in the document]
freddyaboulton: Yes! It's common to build ts models based on only delayed values of the target. Like let's say you have a sequence of daily stock prices or temperature readings, you could model future values based solely on the previous values.

evalml/pipelines/time_series_regression_pipeline.py
freddyaboulton: Great point! I had to modify PipelineBase a little to get this to work (the pipeline parameters method would only read parameters from the component graph) but I'll push this up momentarily!

evalml/utils/gen_utils.py
freddyaboulton: Deleted!

freddyaboulton: Added them to the API docs! Thanks for reminding me.

evalml/pipelines/time_series_regression_pipeline.py
freddyaboulton: Good suggestion - I added an example!

evalml/pipelines/time_series_regression_pipeline.py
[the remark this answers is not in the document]
dsherry: Yep. One common model type like this is [ARIMA](https://en.wikipedia.org/wiki/Autoregressive_integrated_moving_average).

I think in practice, most timeseries pipelines will use the input features.

evalml/pipelines/time_series.py
dsherry: Ah got it, thanks, will respond on new code since this comment is now outdated

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_pandas())
+        y = _convert_woodwork_types_wrapper(y.to_pandas())
+
+        features = self.compute_estimator_features(X, y)
+        predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
+        return pad_with_nans(predictions, self.max_delay if features.isna().values.any() else 0)
```
dsherry, line 84: @freddyaboulton this looks good!

Why not say `pad_with_nans(predictions, self.max_delay)` ? I figure we should rely on the configuration rather than on properties of the data. Fundamentally, we need the shape of the output to match the number of rows in X and y.

freddyaboulton: If the user doesn't have any delayed features (either the `DelayedFeatureTransformer` is missing from the component graph or they passed `delay_features:False, 'delay_target:False` to `DelayFeatureTransformer`) then we don't need to pad the predictions.

`pad_with_nans(predictions, self.max_delay)` will always add `max_delay` number of rows with NaNs to the beginning.

evalml/pipelines/time_series_regression_pipeline.py
dsherry: > I decided to only save the pipeline-level parameters if they are present to avoid introducing breaking changes.

I.e. whatever the user passes in is what's saved on the pipeline? If so, yep I agree. We shouldn't create an empty "pipelines" entry in parameters dicts, only create if its populated

freddyaboulton: Yes that's what I meant!

evalml/pipelines/time_series_regression_pipeline.py
dsherry: Got it, thanks.

Edge cases to consider:
* `delay_features` and `delay_target` were `False`. And, due to a bug or unexpected component behavior, `features` happens to have a `nan` value somewhere, even though there were no `nans` created by the delayed featurizer.
* `delay_features` or `delay_target` were `True`. And, due to a bug or unexpected component behavior, `features` happens to not have any `nans`. Maybe this case isn't important, but it could happen if someone put an imputer after the delayed feature eng.

Is there a foolproof way to avoid these sorts of scenarios? What about
```python
n_rows_padding_pre = X.shape[0] - predictions.shape[0]
return pad_with_nans(predictions, n_rows_padding_pre)
```

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+            dict: Ordered dictionary of objective scores
+        """
+        # Only converting X for the call to _score_all_objectives
+        if X is None:
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+
+        y_predicted = self.predict(X, y)
+        y_shifted = y.shift(-self.gap)
```
dsherry, line 104: Yep, makes sense. So we shift the target down by `gap` rows, so that the target value at row `n` headed into the estimator comes originally from row `n + gap`. And the final `gap` rows are `nan`.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+
+    assert Pipeline(parameters=pl.parameters) == pl
+
+    with pytest.raises(ValueError, match="gap and max_delay parameters cannot be omitted from the parameters dict"):
+        Pipeline({})
+
+
+@pytest.mark.parametrize("only_use_y", [True, False])
+@pytest.mark.parametrize("include_delayed_features", [True, False])
+@pytest.mark.parametrize("gap,max_delay", [(1, 2), (2, 2), (7, 3), (2, 4)])
```
dsherry, line 46: Do we have coverage where `gap` is set to 0 ("given historical features and today's features, predict today's target"), or where `max_delay` is set to 0 ("given today's features, predict a future target") or 1?

I think we should add this to the tests in this file, fit/predict/score

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+@pytest.mark.parametrize("pipeline_class,estimator_name", [(TimeSeriesRegressionPipeline, "Random Forest Regressor")])
+@patch("evalml.pipelines.components.RandomForestRegressor.fit")
+def test_fit_drop_nans_before_estimator(mock_regressor_fit, pipeline_class,
+                                        estimator_name, gap, max_delay, include_delayed_features, only_use_y, ts_data):
+
+    if only_use_y and not include_delayed_features:
+        pytest.skip("Can't prevent label leakage when only the target is used without lagging.")
```
dsherry, line 53: So this would happen when `max_delay=0` and `delay_target=True`? Perhaps we should raise an error in that case?

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+
+    pl = Pipeline({"Delayed Feature Transformer": {"gap": gap, "max_delay": max_delay,
+                                                   "delay_features": include_delayed_features,
+                                                   "delay_target": include_delayed_features},
+                   "pipeline": {"gap": gap, "max_delay": max_delay}})
+
+    if only_use_y:
+        pl.fit(None, y)
+    else:
+        pl.fit(X, y)
```
dsherry, line 75: Nice

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+        pytest.skip("Can't prevent label leakage when only the target is used without lagging.")
+
+    X, y = ts_data
+
+    if include_delayed_features:
+        train_index = pd.date_range(f"2020-10-{1 + max_delay}", f"2020-10-{31-gap}")
+        expected_target = np.arange(1 + gap + max_delay, 32)
+    else:
+        train_index = pd.date_range(f"2020-10-01", f"2020-10-{31-gap}")
+        expected_target = np.arange(1 + gap, 32)
```
dsherry, line 62: This data gen is great, easy to follow

dsherry (approved):
@freddyaboulton colossal!

I like that you put parametrize in place in the unit tests in prep for the other problem types 👏  
and I love the coverage which checks the data provided to the estimator is correct.
I think we've rounded out most of the open stuff. I left a comment in `predict` about the padding -- ideally we should find a way to tighten that up. I left a couple unit test suggestions too.

It would be nice to have coverage for using a time series pipeline which happens to not use delayed feature transformer, but this isn't high priority since we know we want these features.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
freddyaboulton: Yea the problem is that all of our estimators require the X to be non-empty. If you are passing an X=None to the pipeline and you are not computing any features with a DelayedFeaturesTransformer, then our estimators will freak out. I'll raise an error in fit if we're in that case but we may need to rework this in the future when we add time-series specific estimators that compute their own features internally. In that case, we'll likely want to pass in X=None and not include a DelayFeatureTransformer in the pipeline.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
freddyaboulton: Added these combinations of gap and max_delay to the tests!

```python
(0, 0), (1, 0), (0, 2), (1, 1)
```

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -0,0 +1,110 @@
...
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+
+        X_t = self._compute_features_during_fit(X, y)
+
+        y_shifted = y.shift(-self.gap)
+        X_t, y_shifted = keep_non_nan_rows(X_t, y_shifted)
```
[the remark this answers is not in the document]
freddyaboulton: Changed! No worries I agree ehehe

evalml/pipelines/time_series_regression_pipeline.py
freddyaboulton: Yea I think that works! Good suggestion.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
dsherry: That's a good point @freddyaboulton . Sounds good, raising an error in pipeline fit makes sense, and then when we add timeseries support to `make_pipelines` for automl we'll just have to handle this edge case and not add delayed feature eng to the pipeline.

Issue #1290, opened 2020-10-09
Update components from selecting using pandas dtypes to selecting using DataTable semantic or logical types


angela97lin:
Design doc for epic here: https://alteryx.quip.com/Oa4aA6h0Rf2k/Woodwork-DataTable-Integration-Design-Document

dsherry:
Note that we'll want to pay special attention to the text featurizer component. Currently (as in #1062 ), it expects a list of text features to be passed in via the parameters dict. Now that we have woodwork in place, we should remove that behavior in favor of doing `dt.select('natural_language')` in fit and transform. For consideration: we could keep the `text_features` params input as optional, if specified would override the woodwork datatable content. Or we could simply use the datatable.

dsherry:
@angela97lin will #1423 close this issue? If not, what's left after that's merged? I just saw your comment about [the datetime featurizer](https://github.com/alteryx/evalml/pull/1423#discussion_r525388736).

angela97lin:
@dsherry I don't think in all cases (such as the datetime featurizer), so this issue tracks any dangling conversions we have yet to make, as well as (hopefully) completely getting rid of the dtype lists we have (`numeric_dtypes`,`boolean`, `numeric_and_boolean_dtypes`, etc.)

dsherry:
@angela97lin understood, thanks!

PR #1423, opened 2020-11-10
Update components to accept Woodwork inputs

angela97lin:
Continuation of work for #1288

Relevant: https://github.com/pandas-dev/pandas/issues/30038

Had to tweak `_convert_woodwork_types_wrapper` to properly handle how NaN was converted from nullable array --> pd.Series/DataFrame.

Ultimately, I updated s.t. functionality is preserved as it would be for Series/DataFrames.

A column with booleans + nulls using nullable types will be inferred as 'boolean' type. However, this causes a problem when converting to 'bool' type since NaN values are not allowed and will error out. Converting to `bool` type by manually handling will convert nan values to True--undesirable. Hence, we will convert to `object` type (and allow SimpleImputer to fill in the values!)

```
s = pd.Series([True, np.nan, False, np.nan, True], dtype='bool')
--> converts nan to True
s = pd.Series([True, np.nan, False, np.nan, True])
--> dtype = object, no conversion
```

https://stackoverflow.com/questions/36825925/expressions-with-true-and-is-true-give-different-results

https://app.circleci.com/pipelines/github/alteryx/evalml/7300/workflows/0d609f82-10ba-4cf8-84b2-01e9ecf7864f/jobs/81615


evalml/pipelines/classification_pipeline.py
```diff
@@ -41,7 +41,7 @@ def fit(self, X, y):
         """
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        # X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_woodwork_types_wrapper(y.to_series())
         self._encoder.fit(y)
         y = self._encode_targets(y)
```
angela97lin, line 47: Because we encode the targets, keeping the conversion code for `y`.

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -82,29 +80,3 @@ def default_parameters(cls):
                 'cv': None,
                 'n_jobs': 1,
                 }
-
-    def fit(self, X, y=None):
```
angela97lin, line 86: Realized this was unnecessary since it inherits from the `Estimator` class. Deleting!

evalml/pipelines/components/estimators/regressors/baseline_regressor.py
```diff
@@ -38,9 +42,10 @@ def __init__(self, strategy="mean", random_state=0, **kwargs):
     def fit(self, X, y=None):
         if y is None:
             raise ValueError("Cannot fit Baseline regressor if y is None")
-
-        if not isinstance(y, pd.Series):
-            y = pd.Series(y)
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
+        y = _convert_woodwork_types_wrapper(y.to_series())
```
angela97lin, line 48: Arguable: is it necessary to do this conversion for `X` if all we need is `X.shape`? I did here for consistency but...

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -82,17 +85,22 @@ def _get_cat_cols(X):
...
             self.features_to_encode = self._get_cat_cols(X_t)
         invalid_features = [col for col in self.features_to_encode if col not in list(X.columns)]
         if len(invalid_features) > 0:
             raise ValueError("Could not find and encode {} in input data.".format(', '.join(invalid_features)))
-
         if self.parameters['handle_missing'] == "as_category":
+            for col in self.features_to_encode:
+                if X_t[col].dtype == 'category' and pd.isna(X_t[col]).any():
+                    X_t[col] = X_t[col].cat.add_categories("nan")
+                    # X_t[col] = X_t[col].where(pd.isna(X_t[col]) == False, other='nan')
```
angela97lin, line 101: Syntax using `where`: https://stackoverflow.com/questions/36825925/expressions-with-true-and-is-true-give-different-results

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -82,17 +85,22 @@ def _get_cat_cols(X):
...
         X_t = X
 
         if self.features_to_encode is None:
             self.features_to_encode = self._get_cat_cols(X_t)
         invalid_features = [col for col in self.features_to_encode if col not in list(X.columns)]
         if len(invalid_features) > 0:
             raise ValueError("Could not find and encode {} in input data.".format(', '.join(invalid_features)))
-
         if self.parameters['handle_missing'] == "as_category":
+            for col in self.features_to_encode:
```
angela97lin, line 98: Syntax using `where`: https://stackoverflow.com/questions/36825925/expressions-with-true-and-is-true-give-different-results

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -310,7 +321,7 @@ def test_numpy_input():
     encoder = OneHotEncoder()
     encoder.fit(X)
     X_t = encoder.transform(X)
-    assert pd.DataFrame(X).equals(X_t)
+    pd.testing.assert_frame_equal(pd.DataFrame(X), X_t, check_dtype=False)
```
angela97lin, line 324: As mentioned in standup, windows will use int32 as opposed to int64, so not checking for dtype equality. https://app.circleci.com/pipelines/github/alteryx/evalml/7300/workflows/0d609f82-10ba-4cf8-84b2-01e9ecf7864f/jobs/81615

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -260,6 +267,10 @@ def test_more_top_n_unique_values_large():
     encoder = OneHotEncoder(top_n=3, random_state=random_seed)
     encoder.fit(X)
     X_t = encoder.transform(X)
+
+    # Conversion changes the resulting dataframe dtype, resulting in a different random state, so we need make the conversion here too
```
freddyaboulton, line 271: What does converting to woodwork have to do with the random state? 😨

freddyaboulton (approved):
@angela97lin I think this looks good! My main comment is that in all components we follow the formula: data -> woodwork -> pandas when in a lot of cases we don't need to go through woodwork for the component to work as expected . 

I realize this may be a premature performance optimization, but I've seen the conversion pandas to woodwork take a while and it would kill users to have to do that same conversion in every component if they passed in pandas data. I think it would be best to do the woodwork conversion on an "as-needed" basis. Thoughts?

Also, I don't think our component unit tests pass in ww data. Is that coming in a separate PR?

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: Yup, because the new nullable boolean--which is hard to test for equality--we need to grab these values and convert them to the "nan" category manually.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -69,42 +73,43 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
...
     def _encode_categories(self, X, fit=False):
-        X2 = pd.DataFrame(copy.copy(X))
-        # encode each categorical feature as an integer
-        X2.columns = np.arange(len(X2.columns))
+        """Encodes each categorical feature using ordinal encoding."""
+        X_encoded = _convert_to_woodwork_structure(X)
+        X_encoded = _rename_column_names_to_numeric(X_encoded)
+        cat_cols = list(X_encoded.select('category').columns)
+        X_encoded = _convert_woodwork_types_wrapper(X_encoded.to_dataframe())
         # necessary to wipe out column names in case any names contain symbols ([, ], <) which LightGBM cannot properly handle
```
dsherry, line 81: Should this comment move too? @bchen1116

evalml/pipelines/components/estimators/regressors/baseline_regressor.py
dsherry: Ah got it. Yep, may as well use the same code for now for consistency.

evalml/pipelines/components/estimators/regressors/catboost_regressor.py
```diff
@@ -53,11 +55,13 @@ def __init__(self, n_estimators=10, eta=0.03, max_depth=6, bootstrap_type=None,
                          random_state=random_state)
 
     def fit(self, X, y=None):
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
-        if not isinstance(y, pd.Series):
-            y = pd.Series(y)
-        cat_cols = X.select_dtypes(categorical_dtypes)
+        X = _convert_to_woodwork_structure(X)
+        cat_cols = list(X.select('category').columns)
```
dsherry, line 59: 😁 I'm so excited to say goodbye to selecting with pandas types!

bchen1116 (approved):
Things look good to me! I made a few comments for clarifications, but generally agree that if a conversion is unnecessary, we should leave it out.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
dsherry: Ah neat.

Does this work?
```python
X_t[col] = X_t[col].where(~pd.isna(X_t[col]), other='nan')
```
[The tilde is the bitwise negation operator in pandas](https://stackoverflow.com/questions/46054318/tilde-sign-in-python-dataframe)

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -82,17 +85,19 @@ def _get_cat_cols(X):
...
+                if X_t[col].dtype == 'category' and pd.isna(X_t[col]).any():
+                    X_t[col] = X_t[col].cat.add_categories("nan")
+                    X_t[col] = X_t[col].where(pd.isna(X_t[col]).isin([False]), other='nan')
```
dsherry, line 100: So this says that we'll fill anything for which `pd.isna` is true (i.e. `pd.NA`, `np.nan`, `None` I believe?) with `np.nan`?

I see we use this code in `fit` and in `transform`. Is there some way we can move this for loop code to a common private helper method, to avoid duplication? Or would that be overkill in this instance?

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -137,29 +142,32 @@ def transform(self, X, y=None):
...
         # Add the non-categorical columns, untouched
-        for col in X.columns:
+        for col in X_copy.columns:
             if col not in cat_cols:
-                X_t = pd.concat([X_t, X[col]], axis=1)
+                X_t = pd.concat([X_t, X_copy[col]], axis=1)
         # The call to pd.concat above changes the type of the index so we will manually keep it the same.
         if not X_t.empty:
-            X_t.index = X.index
+            X_t.index = X_copy.index
```
dsherry, line 166: Why refer to `X_copy` here instead of `X`?

evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -62,19 +63,22 @@ def fit(self, X, y=None):
...
+        numeric_cols = list(X.select('numeric').columns)
+
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
         self._all_null_cols = set(X.columns) - set(X.dropna(axis=1, how='all').columns)
         X_copy = X.copy()
         X_null_dropped = X_copy.drop(self._all_null_cols, axis=1, errors='ignore')
 
-        X_numerics = X_null_dropped.select_dtypes(include=numeric_dtypes)
+        X_numerics = X_null_dropped[[col for col in numeric_cols if col not in self._all_null_cols]]
```
dsherry, line 76: 👍

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
```diff
@@ -51,8 +53,9 @@ def __init__(self, features_to_extract=None, random_state=0, **kwargs):
                          random_state=random_state)
 
     def fit(self, X, y=None):
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        # to updateeeeeeee
```
dsherry, line 58: Hahaha

Can delete this comment?

evalml/pipelines/pipeline_base.py
```diff
@@ -197,16 +193,13 @@ def compute_estimator_features(self, X, y=None):
     def _compute_features_during_fit(self, X, y):
         X_t = X
         for component in self.component_graph[:-1]:
-            self.input_feature_names.update({component.name: list(pd.DataFrame(X_t))})
+            self.input_feature_names.update({component.name: list(X_t.columns)})
```
dsherry, line 196: So we're relying on the other pipeline code to have done this conversion to pandas DF for us before `PipelineBase.fit` is called?

evalml/tests/component_tests/test_one_hot_encoder.py
dsherry: Hey I'm glad `assert_frame_equal` handles this case :)

evalml/utils/gen_utils.py
```diff
@@ -219,8 +219,16 @@ def _rename_column_names_to_numeric(X):
     Returns:
         Transformed X where column names are renamed to numerical values
     """
-    name_to_col_num = dict((col, col_num) for col_num, col in enumerate(X.columns.values))
-    return X.rename(columns=name_to_col_num, inplace=False)
+    X_t = X
+    if isinstance(X, np.ndarray):
+        return X
```
dsherry, line 224: Should we normalize to dataframe here too? Or we're assuming the caller does that?

evalml/utils/gen_utils.py
```diff
@@ -219,8 +219,16 @@ def _rename_column_names_to_numeric(X):
...
+    if isinstance(X, ww.DataTable):
+        X_t = X.to_dataframe()
+    name_to_col_num = dict((col, col_num) for col_num, col in enumerate(list(X.columns)))
+    X_renamed = X_t.rename(columns=name_to_col_num, inplace=False)
+    if isinstance(X, ww.DataTable):
+        return ww.DataTable(X_renamed)
```
dsherry, line 230: Ah, got it.

One point about this case: we should copy the types from the original woodwork datatable. These types could come from the default woodwork type inference, but they could also be user-defined. We need to preserve them. Otherwise, we could get in a situation where lightgbm/xgboost are fitting on a different set of categorical features for instance vs what the user specified.

@gsheni @thehomebrewnerd @tamargrey any advice on how we can do this?

@angela97lin we should add test coverage for this case too.

evalml/utils/gen_utils.py
```diff
@@ -277,19 +285,22 @@ def _convert_to_woodwork_structure(data):
...
+    elif isinstance(ww_data, np.ndarray):
+        ww_data = pd.DataFrame(ww_data)
+
     # Convert pandas data structures to Woodwork data structures
-    if isinstance(data, pd.Series):
-        return ww.DataColumn(data)
-    return ww.DataTable(data)
+    if isinstance(ww_data, pd.Series):
+        return ww.DataColumn(ww_data)
+    return ww.DataTable(ww_data, copy_dataframe=True)
```
dsherry, line 303: Nice, I love the way this logic is organized @angela97lin ! Its easy to follow what transformations will be applied to the input data for each type of supported input.

I was reasoning through to see if we're subject to the same problem as what I highlighted in `_rename_column_names_to_numeric` above, where woodwork types aren't getting copied over. But I think we're all set here :)

Let's make sure we have solid test coverage of this method. I think we added that in a prev PR but can't remember -- @angela97lin would you mind linking to / describing that if so?

evalml/utils/gen_utils.py
```diff
@@ -300,18 +311,26 @@ def _convert_woodwork_types_wrapper(pd_data):
...
 
     Returns:
-        New pandas data structure (pd.DataFrame or pd.Series) with original data and dtypes that can be handled by numpy
+        Modified pandas data structure (pd.DataFrame or pd.Series) with original data and dtypes that can be handled by numpy
     """
     nullable_to_numpy_mapping = {pd.Int64Dtype: 'int64',
                                  pd.BooleanDtype: 'bool',
                                  pd.StringDtype: 'object'}
-
+    nullable_to_numpy_mapping_nan = {pd.Int64Dtype: 'float64',
```
dsherry, line 319: Hmm why does `Int64Dtype` map to `float64` here? Is it because `float64` supports `np.nan` but `int64` doesn't?

evalml/utils/gen_utils.py
```diff
@@ -300,18 +311,26 @@ def _convert_woodwork_types_wrapper(pd_data):
...
+                                     pd.BooleanDtype: 'object',
```
dsherry, line 320: Got it, so this is what you mentioned in the PR description [RE this thread in pandas-dev](https://github.com/pandas-dev/pandas/issues/30038)?

dsherry (approved):
@angela97lin this rocks!!! It's so cool that we've made it and have woodwork fully integrated into our components, pipelines and automl 🎊

I left a few questions, including some on the gen_utils data transformation methods (`_convert_woodwork_types_wrapper` etc). The only other thing I had is that we should make sure we have good test coverage of those methods, trying them with each of the various different types as input. I remember we added some coverage along these lines but I'm blanking on the specifics, so if you're able to respond to that that would be great :)

Otherwise, looking ready to 🚢💨!

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
bchen1116: yeah above line 78 would be good, but it can also be removed

evalml/utils/gen_utils.py
angela97lin: Yup, that's right 😅

evalml/utils/gen_utils.py
angela97lin: Yup yup! In the case where there is a pd.NA value that needs to be handled, we need to convert to potentially different types (since int64 and bool can't handle).

evalml/tests/component_tests/test_lsa.py
```diff
@@ -152,20 +152,6 @@ def test_int_col_names():
     assert X_t.dtypes.all() == np.float64
 
 
-def test_repeat_col_names():
```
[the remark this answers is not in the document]
angela97lin: Was removed because Woodwork disallows duplicate column names, so no need to test if this still works--Woodwork will throw an error instead (https://github.com/FeatureLabs/woodwork/issues/377)

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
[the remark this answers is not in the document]
angela97lin: bahahaha oops

evalml/utils/gen_utils.py
tamargrey: DataTable has a `logical_types` property that's the same format as the `logical_types` dict you pass in on init, so you could grab that with `X.logical_types`, rename the columns along with the underlying data, and then pass it in on line 230. 

But as @freddyaboulton mentioned in another comment, DataTable has a `rename` method I think you could make use of here if you want

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
angela97lin: Yes 😂 

To note though: I meant to update this to use Woodwork but on second thought, might be best to hold off. Woodwork's `select` is not currently organized by the original order of the columns. Maybe doesn't matter in this case but good to keep in mind; will update for #1290

evalml/tests/component_tests/test_one_hot_encoder.py
angela97lin: Yes, agreed! 😂

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
angela97lin: Removing, since the docstring for `_rename_column_names_to_numeric` explains :)

evalml/utils/gen_utils.py
angela97lin: Ah, you're right, we should normalize here!

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: Oo yes, I always forget this exists. Thanks, updating since this is cleaner!

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -56,10 +59,8 @@ def fit(self, X, y=None):
         return self
 
     def transform(self, X, y=None):
-
-        if not isinstance(X, pd.DataFrame):
-            X = pd.DataFrame(X)
-
+        X = _convert_to_woodwork_structure(X)
```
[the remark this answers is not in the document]
angela97lin: Agreed, though maybe arguably cleaner to have the woodwork conversion handle things, rather than having code check if its a woodwork or numpy structure and then calling the appropriate method to convert. 🤔

evalml/tests/component_tests/test_one_hot_encoder.py
angela97lin: @freddyaboulton I believe the conversion to woodwork can update some dtypes (object --> category) which then results in differences when we sample. 😓 Check this out:
image

angela97lin:
@freddyaboulton You're right that there are certainly some situations where the conversion to Woodwork structures doesn't add value here, but I think it is fine to add for consistency, and allows automl to pass the woodwork structures down to the components without worrying about the data structure. Currently, we're in this weird intermediate state where that's not true, but once work has been done in #1289 to pass the Woodwork data structures directly to the pipelines, which passes it to the components, the conversion code will basically be a no-op! Having the conversion code handled in these util methods also allows us to get rid of the `if isinstance(X, pd.DataFrame)` (or checks for Woodwork) blocks, because it's handled in a centralized place :)

I see your concern with the conversion though, and it's difficult to fully see how time consuming or how much more memory this takes. If this gets out of hand, we can revisit and optimize? 😁

evalml/pipelines/pipeline_base.py
angela97lin: Yeah... confusing where things link but, `fit` (where conversion code is) --> `_fit` --> `_compute_features_during_fit`

evalml/tests/component_tests/test_datetime_featurizer.py
```diff
@@ -77,6 +77,6 @@ def test_datetime_featurizer_no_datetime_cols():
 
 def test_datetime_featurizer_numpy_array_input():
     datetime_transformer = DateTimeFeaturizer()
-    X = np.array(['2007-02-03', '2016-06-07', '2020-05-19'], dtype='datetime64')
+    X = np.array([['2007-02-03'], ['2016-06-07'], ['2020-05-19']], dtype='datetime64')
```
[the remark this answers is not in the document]
angela97lin: Previously, we just had code that wrapped this in a `DataFrame`. Now, the conversion code checks for the dimensionality to determine whether to create a Series or DataFrame (since the same conversion code is used for both), so having a 1D numpy would have converted this to a Series instead, which isn't what we want.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: X_copy is the converted-to-woodwork/pandas version; previously, we modified/converted X to a DataFrame so we used X directly. If we use original X now, it could be a numpy object without a `.index` attribute

evalml/utils/gen_utils.py
```diff
@@ -219,8 +219,16 @@ def _rename_column_names_to_numeric(X):
     Returns:
         Transformed X where column names are renamed to numerical values
     """
-    name_to_col_num = dict((col, col_num) for col_num, col in enumerate(X.columns.values))
-    return X.rename(columns=name_to_col_num, inplace=False)
+    X_t = X
+    if isinstance(X, np.ndarray):
+        return X
+    if isinstance(X, ww.DataTable):
```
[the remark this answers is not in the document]
angela97lin: Ah, the `rename` method isn't in the latest release yet, so this will have to do for now :( 

But yes, I think you're right--updating to preserve types, since that's important!

evalml/utils/gen_utils.py
angela97lin: Just confirmed with Tamar that the `rename` method is not in the latest release so sticking with this for now, but updating to make sure the logical types are carried over, and adding test coverage. Thanks all!

dsherry:
> I see your concern with the conversion though, and it's difficult to fully see how time consuming or how much more memory this takes. If this gets out of hand, we can revisit and optimize? 😁

Yep that sounds good to me. If we can get #1289 done quickly, this won't matter, because automl search will handle the conversion to woodwork types first before the pipelines/components do their thing. And if not, and we do see problems with automl resulting from this, we can do another release quickly once #1289 is merged.

One thing to check before merging is that the unit tests don't take significantly longer to run than on `main`. If they did, that could be potential evidence that there's a problem. This is unlikely though because we don't use large datasets in our tests.

Given that we're talking about performance, @angela97lin how would you feel about running some perf tests on this? I think it would be fine to merge this and test after the fact, since we know this is an intermediate step along the way, but it would help us decide when to release.

angela97lin:
@dsherry Yup, that sounds good! I'll merge this in first, and if we don't have time to merge in the automl change before the next release, I'll run some perf tests to make sure the conversions aren't taking too long.

PR #1425, opened 2020-11-11
Add stacked ensembling to AutoML documentation

### LINKED ISSUE #1329: Add documentation about stacked ensembles in AutoML
Following #1253, we should add some documentation explaining the behavior of stacked ensembles in AutoML and add an example. When do they run? How can users access this? How can they turn it off?

angela97lin:
Closes #1329

[Docs here](https://evalml.alteryx.com/en/1329_stacked_ensemble_docs/user_guide/automl.html#Ensemble-methods:-Stacking)


docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Ensemble methods: Stacking\n",
+    "\n",
+    "[Stacking](https://en.wikipedia.org/wiki/Ensemble_learning#Stacking) is an ensemble machine learning algorithm that involves training a model to best combine the predictions of several base learning algorithms. First, each base learning algorithms is trained using the given data. The combining algorithm or meta-learner is then trained on the predictions made by those base learning algorithms to make a final prediction.\n",
+    "\n",
+    "AutoML enables stacking using the `ensembling` flag during initalization; this is set to `False` by default. The stacking ensemble pipeline runs in its own batch after a whole cycle of training has occurred (each allowed pipeline trains for one batch). Note that this means a large number of iterations may need to run before the stacking ensemble runs. It is also important to note that only the first CV fold is calculated for stacking ensembles because the model internally uses CV folds."
```
jeremyliweishih, line 330: Maybe we can highlight the note in red or something

docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
    "source": [
     "automl.results"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Ensemble methods: Stacking\n",
```
jeremyliweishih, line 326: Up to you but I think something like "Adding ensemble methods to AutoML" as a header and then "stacking" as a subheader could make it fit better with the rest of the page.

docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
...
+    "\n",
+    "[Stacking](https://en.wikipedia.org/wiki/Ensemble_learning#Stacking) is an ensemble machine learning algorithm that involves training a model to best combine the predictions of several base learning algorithms. First, each base learning algorithms is trained using the given data. The combining algorithm or meta-learner is then trained on the predictions made by those base learning algorithms to make a final prediction.\n",
```
jeremyliweishih, line 328: I think "First.., then..." rolls off the tongue a little nicer!

jeremyliweishih (approved):
LGTM just some comments!

docs/source/user_guide/automl.ipynb
freddyaboulton: Good idea! That would structure would accommodate adding new ensemble algorithms in the future!

freddyaboulton (approved):
@angela97lin This looks good to me! This is out of scope of this PR but this sentence might be confusing to users because I don't think we ever explain how the batching in AutoML works:

> The stacking ensemble pipeline runs in its own batch after a whole cycle of training has occurred (each allowed pipeline trains for one batch). Note that this means a large number of iterations may need to run before the stacking ensemble runs.

One thing we can do is add something to the AutoML printout like this if the number of iterations selected by the user is too small:

```
Ensembling set to true but the number of iterations is too small
for ensembling to run. Set it to at least ...
```

and 

```
Ensembling is set to True and max_iterations is set to ... 
so ensembling will run in batches 4, 8, 12 
```

That way users are not so confused about whether or not ensembling will run. Like I said, out of scope of this PR, but I think this would be a useful enhancement in the future!

docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
...
+   "outputs": [],
+   "source": [
+    "X, y = evalml.demos.load_breast_cancer()\n",
+    "X_dt = ww.DataTable(X)\n",
+    "y_dc = ww.DataColumn(y)\n",
+    "automl_with_ensembling = AutoMLSearch(problem_type=\"binary\",\n",
+    "                                      allowed_model_families=[ModelFamily.RANDOM_FOREST, ModelFamily.LINEAR_MODEL],\n",
+    "                                      max_iterations=20,\n",
+    "                                      ensembling=True)\n",
+    "automl_with_ensembling.search(X_dt, y_dc)"
```
bchen1116, line 348: Is it possible to add a call to the `stacked_ensembler.describe` so that users can see and better understand how this works? I think it would be helpful to include info on the other models that help make up this model's predictions

bchen1116 (approved):
LGTM! Just left one comment but not blocking

docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
...
+    "## Ensemble methods: Stacking\n",
+    "\n",
+    "[Stacking](https://en.wikipedia.org/wiki/Ensemble_learning#Stacking) is an ensemble machine learning algorithm that involves training a model to best combine the predictions of several base learning algorithms. First, each base learning algorithms is trained using the given data. The combining algorithm or meta-learner is then trained on the predictions made by those base learning algorithms to make a final prediction.\n",
+    "\n",
+    "AutoML enables stacking using the `ensembling` flag during initalization; this is set to `False` by default. The stacking ensemble pipeline runs in its own batch after a whole cycle of training has occurred (each allowed pipeline trains for one batch). Note that this means a large number of iterations may need to run before the stacking ensemble runs. It is also important to note that only the first CV fold is calculated for stacking ensembles because the model internally uses CV folds."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
```
dsherry, line 335: Weird, I expected this to increase the docs runtime, because it involves running more batches, but it looks more or less the same vs `main`. Cool!

docs/source/user_guide/automl.ipynb
```diff
@@ -318,6 +318,35 @@
...
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "X, y = evalml.demos.load_breast_cancer()\n",
+    "X_dt = ww.DataTable(X)\n",
+    "y_dc = ww.DataColumn(y)\n",
+    "automl_with_ensembling = AutoMLSearch(problem_type=\"binary\",\n",
+    "                                      allowed_model_families=[ModelFamily.RANDOM_FOREST, ModelFamily.LINEAR_MODEL],\n",
+    "                                      max_iterations=20,\n",
```
dsherry, line 346: Could we use `max_batches` instead?

docs/source/user_guide/automl.ipynb
```diff
@@ -336,7 +365,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.4"
+   "version": "3.7.5"
```
dsherry, line 368: I think thanks to @freddyaboulton's additions you can run `make lint-fix` locally and this will get cleaned up!!

dsherry (approved):
@angela97lin looks good to me!

I only had one change request, which was to use `max_batches=5` (or 4? forgetting) instead of `max_iterations=20`

angela97lin:
@freddyaboulton Great suggestion!! I filed https://github.com/alteryx/evalml/issues/1461 to track this 😁

PR #1431, opened 2020-11-12
Adding a script to standardize the notebook versions.

### LINKED ISSUE #1419: Docs notebooks: "sanitize" python version to make it easier to change notebooks
Its annoying that every time we change the docs notebooks locally, that'll update [the "version" listed in the notebook](https://github.com/alteryx/evalml/blob/main/docs/source/index.ipynb?short_path=ca0f4ab#L103). We should find a way to make this easier. We could create a "sanitizer" script or something which runs as part of `lint-fix`, that would be helpful.

freddyaboulton:
### Pull Request Description
Fixes #1419 

This is a test to show that circle ci will fail when the notebook versions don't match: https://app.circleci.com/pipelines/github/alteryx/evalml/7250/workflows/8b47906b-6b9f-4f95-b0ee-c2f78f2a6d9c/jobs/80895

This is what the output looks like when lint fails:

image

Then you can fix with lint-fix:
image

-----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,13 @@ clean:
 
 .PHONY: lint
 lint:
-	flake8 evalml && isort --check-only --recursive evalml
+	flake8 evalml && isort --check-only --recursive evalml && python docs/notebook_version_standardizer.py check-versions
 
 .PHONY: lint-fix
 lint-fix:
 	autopep8 --in-place --recursive --max-line-length=100 --select="E225,E222,E303,E261,E241,E302,E203,E128,E231,E251,E271,E127,E126,E301,W291,W293,E226,E306,E221" evalml
 	isort --recursive evalml
+	python docs/notebook_version_standardizer.py standardize
 
 .PHONY: test
 test:
--- a/docs/notebook_version_standardizer.py
+++ b/docs/notebook_version_standardizer.py
@@ -1,0 +1,77 @@
+import click
+import json
+import os
+
+DOCS_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "source")
+
+
+def _get_python_version(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        version = source['metadata']['language_info']['version']
+    return version
+
+
+def _standardize_python_version(notebook, desired_version='3.7.4'):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        source['metadata']['language_info']['version'] = desired_version
+    json.dump(source, open(notebook, "w"), indent=1)
+
+
+def _get_ipython_notebooks(docs_source):
+    directories_to_skip = ["_templates", "generated", ".ipynb_checkpoints"]
+    notebooks = []
+    for root, _, filenames in os.walk(docs_source):
+        if any(dir_ in root for dir_ in directories_to_skip):
+            continue
+        for filename in filenames:
+            if filename.endswith('.ipynb'):
+                notebooks.append(os.path.join(root, filename))
+    return notebooks
+
+
+def _get_notebooks_with_different_versions(notebooks, desired_version='3.7.4'):
+    different_versions = []
+    for notebook in notebooks:
+        version = _get_python_version(notebook)
+        if version != desired_version:
+            different_versions.append(notebook)
+    return different_versions
+
+
+def _standardize_versions(notebooks, desired_version='3.7.4'):
+    for notebook in notebooks:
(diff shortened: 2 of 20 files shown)


docs/notebook_version_standardizer.py
```diff
@@ -0,0 +1,77 @@
...
+    notebooks = _get_ipython_notebooks(DOCS_PATH)
+    different_versions = _get_notebooks_with_different_versions(notebooks, desired_version)
+    if different_versions:
+        _standardize_versions(different_versions, desired_version)
+        different_versions = ['\t' + notebook for notebook in different_versions]
+        different_versions = "\n".join(different_versions)
+        click.echo(f"Set the notebook version to {desired_version} for:\n {different_versions}")
+
+
+if __name__ == '__main__':
```
angela97lin, line 76: Ooo is this necessary because we're calling `python docs/notebook_version_standardizer.py check-versions`? Jw cause I know we had something similar for our CLI commands and then removed them :d

angela97lin (approved):
Amazing, thanks for this!! LGTM 👍 🚢

Makefile
```diff
@@ -7,12 +7,13 @@ clean:
...
 .PHONY: lint
 lint:
-	flake8 evalml && isort --check-only --recursive evalml
+	flake8 evalml && isort --check-only --recursive evalml && python docs/notebook_version_standardizer.py check-versions
 
 .PHONY: lint-fix
 lint-fix:
 	autopep8 --in-place --recursive --max-line-length=100 --select="E225,E222,E303,E261,E241,E302,E203,E128,E231,E251,E271,E127,E126,E301,W291,W293,E226,E306,E221" evalml
 	isort --recursive evalml
+	python docs/notebook_version_standardizer.py standardize
```
dsherry, line 16: Very cool!!

docs/notebook_version_standardizer.py
dsherry: Ah yeah, @angela97lin I think we had to remove that for codecov, right, before we loosened the coverage thresholding? I'm blanking on whether there was another reason... fundamentally, if this works it works!

docs/notebook_version_standardizer.py
```diff
@@ -0,0 +1,77 @@
...
+        _standardize_python_version(notebook, desired_version)
+
+
+@click.group()
+def cli():
+    """no-op"""
+
+
+@cli.command()
+@click.option('--desired-version', default='3.7.4', help='python version that all notebooks should match')
```
dsherry, line 54: Perhaps we should standardize to the latest 3.8? I guess it doesn't really matter lol, unless there's some ugly version-dependent docs bug which pops up haha. I remember we surveyed the group and most people were running different python versions locally so that doesn't help us decide here 😆

docs/notebook_version_standardizer.py
```diff
@@ -0,0 +1,77 @@
...
+def check_versions(desired_version):
+    notebooks = _get_ipython_notebooks(DOCS_PATH)
+    different_versions = _get_notebooks_with_different_versions(notebooks, desired_version)
+    if different_versions:
+        different_versions = ['\t' + notebook for notebook in different_versions]
+        different_versions = "\n".join(different_versions)
+        raise SystemExit(f"The following notebooks don't match {desired_version}:\n {different_versions}")
```
dsherry, line 61: Could you please add something about how to fix? Like "please run `make lint-fix` to fix". Because this will now show up in the lint job failures.

dsherry (approved):
This rocks!! I'm gonna use this all the time locally. Plus I think this means lint will fail if this standardization hasn't been run... neat! Thank you 🙏  I left one comment about having the error message mention `make lint-fix`

bchen1116 (approved):
O this is awesome! LGTM

docs/notebook_version_standardizer.py
freddyaboulton: I'm going for 3.8.2 since that was the version of 3.8 that was in the notebooks before this PR. The good thing is that we can pass in the desired version in the makefile in the future

docs/notebook_version_standardizer.py
freddyaboulton: Done!

docs/notebook_version_standardizer.py
freddyaboulton: I think we don't need it for the cli because the setup.py file sets the entry point as __main__.py:cli()? Without this line, the cli won't run if you do `python notebook_version_sanitizer.py`

PR #1433, opened 2020-11-13
Implement Engine API

### LINKED ISSUE #1296: Add Search Engine API
Preserve existing sequential implementation
- [ ] Create structures for engines 
- [ ] Implement base class + SequentialEngine
- [ ] Add in docs for defining your own execution engine
- [ ] Throughly unit + perf test implementation to ensure consistency

christopherbunn:
Resolves #1296

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,8 +1,6 @@
 import copy
-import sys
 import time
-import traceback
-from collections import OrderedDict, defaultdict
+from collections import defaultdict
 
 import cloudpickle
 import numpy as np
@@ -13,6 +11,7 @@ from .pipeline_search_plots import PipelineSearchPlots
 
 from evalml.automl.automl_algorithm import IterativeAlgorithm
 from evalml.automl.callbacks import log_error_callback
+from evalml.automl.engines import SequentialEngine
 from evalml.automl.utils import (
     get_default_primary_search_objective,
     make_data_splitter
@@ -24,19 +23,13 @@ from evalml.data_checks import (
     EmptyDataChecks,
     HighVarianceCVDataCheck
 )
-from evalml.exceptions import (
-    AutoMLSearchException,
-    PipelineNotFoundError,
-    PipelineScoreError
-)
-from evalml.model_family import ModelFamily
+from evalml.exceptions import AutoMLSearchException, PipelineNotFoundError
 from evalml.objectives import (
     get_core_objectives,
     get_non_core_objectives,
     get_objective
 )
 from evalml.pipelines import (
-    BinaryClassificationPipeline,
     MeanBaselineRegressionPipeline,
     ModeBaselineBinaryPipeline,
     ModeBaselineMulticlassPipeline,
@@ -59,8 +52,7 @@ from evalml.utils.logger import (
     get_logger,
     log_subtitle,
     log_title,
-    time_elapsed,
-    update_pipeline
+    time_elapsed
 )
 
 logger = get_logger(__file__)
@@ -277,6 +269,7 @@ class AutoMLSearch:
                                                    n_splits=3, shuffle=True, random_state=self.random_seed)
         self.data_splitter = self.data_splitter or default_data_splitter
         self.pipeline_parameters = pipeline_parameters if pipeline_parameters is not None else {}
+        self.search_iteration_plot = None
 
     def _validate_objective(self, objective):
         non_core_objectives = get_non_core_objectives()
@@ -367,33 +360,7 @@ class AutoMLSearch:
         else:
             return DataChecks(data_checks)
 
-    def _handle_keyboard_interrupt(self, pipeline, current_batch_pipelines):
-        """Presents a prompt to the user asking if they want to stop the search.
-
-        Arguments:
-            pipeline (PipelineBase): Current pipeline in the search.
-            current_batch_pipelines (list): Other pipelines in the batch.
-
-        Returns:
-            list: Next pipelines to search in the batch. If the user decides to stop the search,
-                an empty list will be returned.
-        """
-        leading_char = "\n"
-        start_of_loop = time.time()
-        while True:
-            choice = input(leading_char + "Do you really want to exit search (y/n)? ").strip().lower()
-            if choice == "y":
-                logger.info("Exiting AutoMLSearch.")
-                return []
-            elif choice == "n":
-                # So that the time in this loop does not count towards the time budget (if set)
-                time_in_loop = time.time() - start_of_loop
-                self._start += time_in_loop
-                return [pipeline] + current_batch_pipelines
-            else:
-                leading_char = ""
-
-    def search(self, data_checks="auto", show_iteration_plot=True):
+    def search(self, data_checks="auto", show_iteration_plot=True, engine=None):
         """Find the best pipeline for the data set.
 
         Arguments:
@@ -407,6 +374,9 @@ class AutoMLSearch:
 
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
+
+            engine (EngineBase): The pipeline processing engine to use during AutoML search.
+                If not specified, uses `SequentialEngine` by default.
         """
         if self._searched:
             logger.info("AutoMLSearch.search() has already been run and will not run again on the same instance. Re-initialize AutoMLSearch to search again.")
@@ -493,9 +463,14 @@ class AutoMLSearch:
             pipeline_params=pipeline_params
         )
 
+        if engine is None:
+            engine = SequentialEngine()
+
         log_title(logger, "Beginning pipeline search")
         logger.info("Optimizing for %s. " % self.objective.name)
         logger.info("{} score is better.\n".format('Greater' if self.objective.greater_is_better else 'Lower'))
+        if engine:
+            logger.info(f"Using {engine.name} to process pipelines.")
 
         if self.max_batches is not None:
             logger.info(f"Searching up to {self.max_batches} batches for a total of {self.max_iterations} pipelines. ")
@@ -504,13 +479,12 @@ class AutoMLSearch:
         if self.max_time is not None:
             logger.info("Will stop searching for new pipelines after %d seconds.\n" % self.max_time)
         logger.info("Allowed model families: %s\n" % ", ".join([model.value for model in self.allowed_model_families]))
-        search_iteration_plot = None
         if self.plot:
-            search_iteration_plot = self.plot.search_iteration_plot(interactive_plot=show_iteration_plot)
+            self.search_iteration_plot = self.plot.search_iteration_plot(interactive_plot=show_iteration_plot)
 
         self._start = time.time()
 
-        should_terminate = self._add_baseline_pipelines()
+        should_terminate = self._add_baseline_pipelines(engine=engine)
         if should_terminate:
             return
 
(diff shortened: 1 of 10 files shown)


evalml/automl/engines/engine_base.py
```diff
@@ -0,0 +1,184 @@
...
+                        parameters=parameters,
+                        training_time=evaluation_result['training_time'],
+                        cv_data=evaluation_result['cv_data'],
+                        cv_scores=evaluation_result['cv_scores'])
+        logger.debug('Adding results complete')
+
+        if self.automl.search_iteration_plot:
+            self.automl.search_iteration_plot.update()
+
+    def _handle_keyboard_interrupt(self, current_batch_pipelines, pipeline=None):
```
christopherbunn, line 81: The keyboard interrupt works, but I think there's still some room to tweak this.

evalml/automl/engines/engine_base.py
```diff
@@ -0,0 +1,212 @@
...
+            elif choice == "n":
+                # So that the time in this loop does not count towards the time budget (if set)
+                time_in_loop = time.time() - start_of_loop
+                self.automl._start += time_in_loop
+                return [pipeline] + current_batch_pipelines if pipeline else current_batch_pipelines
+            else:
+                leading_char = ""
+
+    @staticmethod
+    def _compute_cv_scores(pipeline, automl, X, y):
```
christopherbunn, line 135: Since this is static, this could potentially be moved back into `automl_search` rather than stay in engines. If kept in `AutoMLSearch`, it would give it one central place for the default implementation of computing CV scores. However, keeping it in engines would make sense for reimplementations of `_compute_cv_scores`. Thoughts?

evalml/automl/engines/sequential_engine.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Arguments:
+            pipeline_batch (list(PipelineBase)): A batch of pipelines to be fitted and evaluated.
+            result_callback (callable): Function called once the pipeline is finished evaluating to store the results. If None, results will only be returned.
+            log_pipeline (bool): If True, log the pipeline and relevant information before evaluation.
+
+        Returns:
+            list(PipelineBase), list(dict), list(PipelineBase): A list of evaluated pipelines, the results of the evaluated pipelines, and the remaining pipelines in the batch that were not evaluated. 
+        """
+        super().evaluate_batch()
```
christopherbunn, line 21: Calling these dataset checks aren't super intuitive. Should I split them out into another function (like `check_search_info` or something like that)?

christopherbunn (commented):
Left a few points I wanted to discuss a bit, open to hearing thoughts! I'm also looking for more direct engine test cases to cover, as I'm not trying to overlap too much with the existing test cases for AutoML search.

evalml/automl/engines/engine_base.py
```diff
@@ -0,0 +1,212 @@
...
+    def load_data(self, X, y):
+        """Loads the data to fit the pipeline on. Required to run `_compute_cv_scores`."""
+        self.X = X
+        self.y = y
+
+    def load_search(self, search_obj):
+        """Loads the current AutoML state. This includes information such as the search parameters and objectives, which is required to run `_compute_cv_scores`."""
+        self.automl = search_obj
+
+    @abstractmethod
```
freddyaboulton, line 45: How come we need both `evaluate_batch` and `evaluate_pipeline`? I think it would be nice if there was only one `evaluate_batch` that lived in the base class that called `evaluate_pipeline` on each pipeline in the batch. I think having only one `evaluate_batch` implementation will help with maintainability in the long run:
- It'll help us keep the `KeyboardInterrupt` in only one place as opposed to every engine implementation. 
- It will also help us keep the `if self.X is None` checks in one place as well

I think the trade-off is that this means that `evaluate_pipeline` should always return a "future"-like object. We would have to create an `EngineResult` base class to create a single api between the results of all the engines. 

So it's not without a cost but it could be worth it to reduce the code duplication between the `evaluate_batch` and `evaluate` of different engines.

freddyaboulton (commented):
@christopherbunn Looks good! Not finished but I have a couple of high-level questions and comments about the design so far.

jeremyliweishih (commented):
Left some discussion points!

evalml/automl/engines/engine_base.py
christopherbunn: The behavior is more or less the same. However, in the context of engines, the mechanism of emptying out the batch on termination feels unnecessary. I might put some more thought in what a cleaner implementation would look like but this one currently works as-is.

evalml/automl/engines/engine_base.py
christopherbunn: This version of `_compute_cv_scores` shouldn't have any side effects. Dask is actually able to use this as-is since there isn't any side effects. Let's try moving `_compute_cv_scores` to `AutoMLSearch` then.

evalml/automl/automl_search.py
```diff
@@ -740,62 +626,40 @@ def _add_result(self, trained_pipeline, parameters, training_time, cv_data, cv_s
...
-
-                if not add_single_pipeline:
-                    update_pipeline(logger, desc, len(self._results['pipeline_results']) + 1, self.max_iterations,
-                                    self._start, 1 if baseline else self._automl_algorithm.batch_number, self.show_batch_output)
-
-                evaluation_results = self._compute_cv_scores(pipeline, X, y)
-                parameters = pipeline.parameters
-
+        current_pipeline_batch_size = 1 if isinstance(current_pipeline_batch, PipelineBase) else len(current_pipeline_batch)
+        if engine is None:
```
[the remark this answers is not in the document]
christopherbunn: @freddyaboulton Re creating a new engine for each batch: it's definitely not the intended behavior. I moved this check into `search` and `add_to_rankings`. I also added engines as an optional parameter in `add_to_rankings`. I think it makes sense to still make the engines parameter optional and keep this check in `_evaluate_pipelines`. In the future, it would make setting `_evaluate_pipelines` as public easier.

@jeremyliweishih Re changing the default parameter: I originally had this check set here rather than as a default parameter so that in subsequent PRs we could refactor this later on to use a Dask engine if supported and move this as a fallback engine. This however is a separate refactor down the line and I wouldn't be opposed to changing the default parameter. 

In terms of adding engine as a default parameter arg, I originally wanted to do this but decided against it.
- Currently, this implementation uses the entire AutoMLSearch object in the engine to process pipelines. If the Dask Engine was part of the automl object, then Dask would not be able to serialize it since it can't serialize itself. This could be resolved by having a preprocessing step in engines itself that would remove the engines attribute, but for implementing additional engines in the future it's pretty unintuitive. 
- While a bit niche, I could see cases where we would want to use different engines for evaluating different types of pipelines. For example, for a large search I might want to use the Dask engine while if I'm adding a single pipeline with `add_to_rankings` I might use the SequentialEngine. 

I do see how passing in the engine could become tedious (especially for subsequent functions that evaluates pipelines such as `add_to_rankings`). 

Again, looking forward to discussing each one of these points further!

evalml/automl/engines/sequential_engine.py
```diff
@@ -0,0 +1,68 @@
...
+            try:
+                if not self.automl._check_stopping_condition(self.automl._start):
+                    return completed_pipelines, evaluation_results, []
+                if log_pipeline:
+                    self.log_pipeline(pipeline)
+                if self.automl.start_iteration_callback:
+                    self.automl.start_iteration_callback(pipeline.__class__, pipeline.parameters, self)
+                pipeline, evaluation_result = self._compute_cv_scores(pipeline, self.automl, self.X, self.y)
+                if result_callback:
+                    self._add_result_callback(result_callback, pipeline, evaluation_result)
```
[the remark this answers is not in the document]
christopherbunn: I originally had it so that `_add_result` is called after the batch of pipelines is returned (which I think is the design that you mentioned). However, this design would mean that we could only update the search iteration plot after every batch rather than after every pipeline like we have in main. We could update the way that the pipeline iteration plot works so that it can update at a per-pipeline level by taking in a result. However, this would become messy quickly since it would involve having to combine results from the current search results and the finished results of the current batch. I figured it would be easier to have a function in `EngineBase` that has the callback + updates the plot rather than try to change how plotting works. 

I agree with the renaming since the current implementation is updating the plot + it makes it clear where the callback is going to be called. Updated in newest commit.

evalml/automl/engines/engine_base.py
christopherbunn: I agree with the two points you made for just keeping `evaluate_batch`. My original reasoning for having a separate `evaluate_batch` and `evaluate_pipeline` function is that there would be separate overhead with processing a batch function that isn't needed in processing a single pipeline. However, in practice, there really isn't that much of a difference especially since for Dask we are distributing the dataset across workers in `load_data`. It also allows for a separate method for processing for the baseline search pipeline and `add_to_rankings`, although this is moot since we could just specify a "batch" with one pipeline. I'll take a pass at refactoring so that we just have `evaluate_batch`.

I also agree on the `EngineResult` base class. We're currently returning three things, with one of them being a big dictionary with the numerical results. It might also make sending back error messages (such as in the case of search termination) more explicit to automl search.

evalml/automl/engines/engine_base.py
```diff
@@ -0,0 +1,212 @@
...
+        """This class represents an "engine" for AutoML, which handles the evaluation of a list of pipelines generated from an AutoML search.
+
+        To use this interface, you must define an `evaluate_batch` method and an `evaluate_pipeline` method.
+        """
+        self.name = "Base Engine"
+        self.X = None
+        self.y = None
+        self.automl = None
+
+    def load_data(self, X, y):
```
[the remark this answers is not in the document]
christopherbunn: I had this as a separate function so that we could reuse an engine across multiple searches with different datasets. This is useful for engines like Dask, where this function is overridden and used to distribute the dataset to all workers. 

We could change `__init__` to take in an initial X and y and call `load_data` internally. Thoughts?

christopherbunn:
Thanks for the review @jeremyliweishih and @freddyaboulton! I pushed some of the quick fixes but I'll re-request once I have the bigger ones in.

evalml/automl/automl_search.py
```diff
@@ -442,7 +404,7 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
         if self._data_check_results["errors"]:
             raise ValueError("Data checks raised some warnings and/or errors. Please see `self.data_check_results` for more information or pass data_checks='disabled' to search() to disable data checking.")
         if self.allowed_pipelines is None:
-            logger.info("Generating pipelines to search over...")
+            logger.debug("Generating pipelines to search over...")
```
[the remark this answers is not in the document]
christopherbunn: Oops, that was a mistake. I'll revert it.

evalml/automl/automl_search.py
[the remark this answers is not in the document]
freddyaboulton: @christopherbunn Agreed that keeping engine as an optional parameter to `_evaluate_pipelines` is a good idea!

christopherbunn:
@freddyaboulton and @jeremyliweishih I took in some of your suggestions. Notable changes include:
- Removing `evaluate_pipeline` and just using `evaluate_batch`. I might rename `evaluate_batch` to `evaluate_pipelines` in the future.
- Introducing the `EngineResult` class.  At the end of `evaluate_batch`, it should return an `EngineResult` with the processed pipelines, results, and processed pipelines. It also indicates whether or not there was early stopping. Is it worthwhile to separate out stopping because of the search condition vs. a keyboard interrupt in this context?
- Updated handle keyboard interrupt implementation so that it handles prompting the user and updating the timing. 

I'm still working on adding tests for `EngineResult`, but I'm looking for feedback on the implementation for now 😄

dsherry:
I'm picking this up! Closing in favor of #1838 which is based on this PR. Not ready for review yet but soon 🤞

Issue #1430, opened 2020-11-12
Data checks: JSON-friendly message fmt, include a type enum and affected column names


dsherry:
This will make it easy to programmatically identify which feature or features have a particular error or warning.

BoopBoopBeepBoop:
Slight expansion: to enable other programmatic formatting & internationalization, it would be nice if EvalML returned the un-formatted information in addition to the message that it returns right now. Something like:

```json
{
  "message": "Warning: too many null values present in column 'foobar'",
  "code": "TOO_MANY_NULLS",
  "detail": {
    "level": "warning",
    "columns": ["foobar"]
  }
}
```

This would make it possible to perform language independent formatting on the results, by tying the `code` back to internationalization strings. And having metadata available in the detail allows that information to be optionally formatted back into the messages that are presented in a structured fashion.

dsherry:
@BoopBoopBeepBoop thanks. I like that proposal.

Would you expect "code" to be an enum, or just a string defined somewhere inside the data check?

One tweak I may suggest is to keep the structure flat:
```python
{
  "message": "Warning: too many null values present in column 'foobar'",
  "code": "TOO_MANY_NULLS",
  "level": "warning",
  "columns": ["foobar"]
}
```

Here we're talking in terms of JSON, but setting that aside I think each of these fields should be attached to the `DataCheckWarning`/`DataCheckError` object.

BoopBoopBeepBoop:
That's fair - I had added those elements there as in my head `detail` was "essentially a map". However, those aren't great examples, because they're probably present for all checks ...

Would there be any situations where our messages should include additional metadata that is essentially present on a "per-message" basis? Examples I've seen on this before would be calling out specific values as examples, formatting dynamic lower/upper bounds in messages, or similar. That is usually where you benefit from having a `detail` (or pick a different name) map

BoopBoopBeepBoop:
> Would you expect "code" to be an enum, or just a string defined somewhere inside the data check?

I would just expect it to be stable - where it lives is 🤷

Cmancuso:
> > Would you expect "code" to be an enum, or just a string defined somewhere inside the data check?
> 
> I would just expect it to be stable - where it lives is 🤷

If it's all the same I think I'd prefer an enum. We're currently defining our own strings to determine which checks to run as part of the job, and it would be nice to have an enum consistent to EvalML to leverage there in stead of creating our own. 

[...]

PR #1444, opened 2020-11-17
Update data checks to return DataCheckResults object

### LINKED ISSUE #1325: Data checks: make it easier to process results
Currently, the data check messages are returned as a list of `DataCheckMessage` objects, which can either be errors or warnings.

This achieves the goal of delivering the data check errors and warnings to users. But the current return format makes it hard to work with the results--we should update the API to make it easy.

Use-cases we should bear in mind here:
* User wants to get a list of all errors, or a list of all warnings, but not both
* User wants to know if there's been at least one error or warning (perhaps redundant with the above)
* User wants to combine multiple errors into a single string message, to raise an exception / display

angela97lin:
Closes #1325 

- `validate` always returns a DataCheckResults object now; no messages means returning `{DataCheckMessageType.WARNING: [], DataCheckMessageType.ERROR: []}`
- to use `DataCheckMessageType` as key?


jeremyliweishih (approved):
LGTM - very nice

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -33,29 +34,33 @@ def validate(self, X, y):
...
 
         Example:
             >>> X = pd.DataFrame({})
             >>> y = pd.Series([0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
             >>> target_check = ClassImbalanceDataCheck(threshold=0.10)
-            >>> assert target_check.validate(X, y) == [DataCheckError("The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "ClassImbalanceDataCheck"), DataCheckWarning("The following labels fall below 10% of the target: [0]", "ClassImbalanceDataCheck")]
+            >>> assert target_check.validate(X, y) == {DataCheckMessageType.ERROR: [DataCheckError("The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "ClassImbalanceDataCheck")],\
+                                                       DataCheckMessageType.WARNING: [DataCheckWarning("The following labels fall below 10% of the target: [0]", "ClassImbalanceDataCheck")]}
         """
+        messages = {
```
freddyaboulton, line 47: I think the downside of using `DataCheckMessageType` as the keys is that the user needs to import the enum in order to look at the warnings and results. I think that can be annoying, especially if the user is running automl and trying to access `data_checks_results`:

1. Creating a `DataCheckResults` class.
2. Using string values like "warning" and "error" as keys in the dict you have here.

The benefits of 1 is that we can expose the warnings and errors as instance properties and the user doesn't have to import another class. We can also have an api for checking if there are warnings or errors as opposed to `if self._data_check_results[DataCheckMessageType.ERROR]`. The downside is that it's another class but our docs already show how to get data from the data checks results.

The benefit of 2 is that it'd be a small change but would definitely lead to typos lol.

My vote is for 1 I think it'd be fine to keep it as-is for now too. I saw that you left `to use DataCheckMessageType as key?` in the PR description so wanted to offer my thoughts lol.

freddyaboulton (approved):
@angela97lin Looks good! I left a comment about how defining a class could make things easier/cleaner in the long run.

bchen1116 (approved):
LGTM!

evalml/data_checks/class_imbalance_data_check.py
angela97lin: Ooo I like this suggestion a lot. I agree, the reason why I left the comment about using `DataCheckMessageType` as key was because as I was updating the code, I too felt the inconvenience / frustration of having to import the enum everywhere, but as @dsherry had mentioned, since we already have the enums in place we might as well use them.

That being said, I like the idea of creating a separate `DataCheckResults` class, and having `warnings` and `errors` as attributes 🤔 That way, the user doesn't need to directly type in the keys as strings, and any typos will result in an `AttributeError` instead.

evalml/data_checks/data_checks.py
```diff
@@ -81,13 +82,17 @@ def validate(self, X, y=None):
...
-        messages = []
+        messages = {
+            DataCheckMessageType.WARNING: [],
+            DataCheckMessageType.ERROR: []
+        }
         for data_check in self.data_checks:
             messages_new = data_check.validate(X, y)
-            messages.extend(messages_new)
+            messages[DataCheckMessageType.WARNING].extend(messages_new[DataCheckMessageType.WARNING])
+            messages[DataCheckMessageType.ERROR].extend(messages_new[DataCheckMessageType.ERROR])
```
dsherry, line 95: 👍

dsherry (approved):
@angela97lin LGTM, 🚢 ! I suggest you merge this and then start #1430 because #1430 builds on this work.

angela97lin:
Closing in favor of #1448

PR #1447, opened 2020-11-19
Reduce number of CircleCI workers to 4

christopherbunn:
To reduce memory utilization but retain some speedup, this PR reduces the number of CircleCI workers to 4.


dsherry (approved):
👍  yep! 🚢

PR #1448, opened 2020-11-19
Update data checks to return a dictionary

### LINKED ISSUE #1325: Data checks: make it easier to process results

angela97lin:
Continuation of #1444, closes #1325 

Rather than returning the `DataCheckMessage` classes themselves, we now return the dictionary version so that validate returns:
```
{ "errors": [{"message": "message here", "data_check_name": "name here", "level": "error"}],
 "warnings": [#same here]
}
```


evalml/data_checks/no_variance_data_check.py
```diff
@@ -50,32 +51,34 @@ def validate(self, X, y):
...
         for name in unique_counts:
             message = self._check_for_errors(name, unique_counts[name], any_nulls[name])
-
             if message:
-                messages.append(message)
-
+                if message.message_type == DataCheckMessageType.ERROR:
+                    messages["errors"].append(message.to_dict())
+                elif message.message_type == DataCheckMessageType.WARNING:
+                    messages["warnings"].append(message.to_dict())
```
dsherry, line 74: What about
```python
for name in unique_counts:
    message = self._check_for_errors(name, unique_counts[name], any_nulls[name])
    if not message:
        continue
    messages[message.message_type].append(message.to_dict())
```
---
And, is there any way we can make this easier to implement? What if we defined this helper in `DataCheck`
```python
@staticmethod
def add_message(message, messages):
    if message.message_type not in messages:
        messages[message.message_type] = []
    messages[message.message_type].append(message.to_dict())
```
and then here in the no-variance data check:
```python
for name in unique_counts:
    message = self._check_for_errors(name, unique_counts[name], any_nulls[name])
    DataCheck.add_message(message, messages)
```

docs/source/release_notes.rst
```diff
@@ -24,7 +24,8 @@ Release Notes
     * Changes
         * Changed ``OutliersDataCheck`` to return the list of columns, rather than rows, that contain outliers :pr:`1377`
         * Simplified and cleaned output for Code Generation :pr:`1371`
-        * Reverted changes from :pr:`1337` :pr:`1409`
+        * Reverted changes from :pr:`1337` :pr:`1409``
```
dsherry, line 27: Typo, shouldn't add another tick mark at the end of this line!

dsherry (approved):
LGTM! Left one suggestion but not blocking

Issue #1427, opened 2020-11-11
Use wrapper to do Woodwork conversion


angela97lin:
The Woodwork integration is requiring a lot of conversions to Woodwork data structures and then back to pandas to pass along to third party libraries, causing a lot of the same code to be sprinkled around:
```
        X = _convert_to_woodwork_structure(X)
        X = _convert_woodwork_types_wrapper(X.to_pandas())
```
 While this is fine for the main integration portion, perhaps we should try to encapsulate this behavior in a wrapper. This would make it slightly harder to debug (as wrappers/metaclasses do), but could save on a lot of duplicate code and the possibility of forgetting a call when necessary.

angela97lin:
Closing this. I think during the integration, I realized that it is maybe not that useful to have a wrapper, especially since sometimes we want to have lines of logic between these two calls (check what columns are categorical, grab those cols, etc.). These two lines aren't too much real estate, and I think having it for some simple Transformers and not for others that need more logic in between would ruin the whole point of having a wrapper that we can just use.

If we decide this could actually be very useful, happy to reopen and discuss, but this just seems like a premature / unnecessary improvement for now.

Issue #1292, opened 2020-10-09
Update all graph methods, util methods which accept data, to support DataTables


angela97lin:
Design doc for epic here: https://alteryx.quip.com/Oa4aA6h0Rf2k/Woodwork-DataTable-Integration-Design-Document

PR #1450, opened 2020-11-20
Update AutoML to pass Woodwork DataTables to every pipeline/component (instead of pandas DataFrames)

### LINKED ISSUE #1289: Update AutoML to pass Woodwork DataTables to every pipeline/component (instead of pandas DataFrames)

angela97lin:
Closes #1289

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -696,7 +696,6 @@ class AutoMLSearch:
                 logger.debug(f"\t\t\tFold {i}: {self.objective.name} score: {scores[self.objective.name]:.3f}")
                 score = scores[self.objective.name]
             except Exception as e:
-                raise e
 
                 if self.error_callback is not None:
                     self.error_callback(exception=e, traceback=traceback.format_tb(sys.exc_info()[2]), automl=self,
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -412,7 +412,6 @@ class AutoMLSearch:
 
         # X = _convert_woodwork_types_wrapper(X.to_dataframe())
         # y = _convert_woodwork_types_wrapper(y.to_series())
-
         self._set_data_split(X)
 
         data_checks = self._validate_data_checks(data_checks)
@@ -649,7 +648,6 @@ class AutoMLSearch:
         start = time.time()
         cv_data = []
         logger.info("\tStarting cross validation")
-        # import pdb; pdb.set_trace()
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
         X_pd = _convert_woodwork_types_wrapper(X.to_dataframe())
@@ -698,6 +696,8 @@ class AutoMLSearch:
                 logger.debug(f"\t\t\tFold {i}: {self.objective.name} score: {scores[self.objective.name]:.3f}")
                 score = scores[self.objective.name]
             except Exception as e:
+                raise e
+
                 if self.error_callback is not None:
                     self.error_callback(exception=e, traceback=traceback.format_tb(sys.exc_info()[2]), automl=self,
                                         fold_num=i, pipeline=pipeline)
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -696,6 +696,7 @@ class AutoMLSearch:
                 logger.debug(f"\t\t\tFold {i}: {self.objective.name} score: {scores[self.objective.name]:.3f}")
                 score = scores[self.objective.name]
             except Exception as e:
+                raise e
 
                 if self.error_callback is not None:
                     self.error_callback(exception=e, traceback=traceback.format_tb(sys.exc_info()[2]), automl=self,
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -410,13 +410,13 @@ class AutoMLSearch:
             logger.warning("`y` passed was not a DataColumn. EvalML will try to convert the input as a Woodwork DataTable and types will be inferred. To control this behavior, please pass in a Woodwork DataTable instead.")
             y = _convert_to_woodwork_structure(y)
 
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        y = _convert_woodwork_types_wrapper(y.to_series())
+        # X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        # y = _convert_woodwork_types_wrapper(y.to_series())
 
         self._set_data_split(X)
 
         data_checks = self._validate_data_checks(data_checks)
-        data_check_results = data_checks.validate(X, y)
+        data_check_results = data_checks.validate(_convert_woodwork_types_wrapper(X.to_dataframe()), _convert_woodwork_types_wrapper(y.to_series()))
         if data_check_results:
             self._data_check_results = data_check_results
             for message in self._data_check_results:
@@ -649,7 +649,14 @@ class AutoMLSearch:
         start = time.time()
         cv_data = []
         logger.info("\tStarting cross validation")
-        for i, (train, test) in enumerate(self.data_split.split(X, y)):
+        # import pdb; pdb.set_trace()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X_pd = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y_pd = _convert_woodwork_types_wrapper(y.to_series())
+
+        for i, (train, test) in enumerate(self.data_split.split(X_pd, y_pd)):
+
             if pipeline.model_family == ModelFamily.ENSEMBLE and i > 0:
                 # Stacked ensembles do CV internally, so we do not run CV here for performance reasons.
                 logger.debug(f"Skipping fold {i} because CV for stacked ensembles is not supported.")
@@ -658,8 +665,8 @@ class AutoMLSearch:
             X_train, X_test = X.iloc[train], X.iloc[test]
             y_train, y_test = y.iloc[train], y.iloc[test]
             if self.problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
-                diff_train = set(np.setdiff1d(y, y_train))
-                diff_test = set(np.setdiff1d(y, y_test))
+                diff_train = set(np.setdiff1d(y_pd, y_train.to_series()))
+                diff_test = set(np.setdiff1d(y_pd, y_test.to_series()))
                 diff_string = f"Missing target values in the training set after data split: {diff_train}. " if diff_train else ""
                 diff_string += f"Missing target values in the test set after data split: {diff_test}." if diff_test else ""
                 if diff_string:
@@ -706,8 +713,8 @@ class AutoMLSearch:
             ordered_scores = OrderedDict()
             ordered_scores.update({self.objective.name: score})
             ordered_scores.update(scores)
-            ordered_scores.update({"# Training": len(y_train)})
-            ordered_scores.update({"# Testing": len(y_test)})
+            ordered_scores.update({"# Training": y_train.shape[0]})
+            ordered_scores.update({"# Testing": y_test.shape[0]})
 
             evaluation_entry = {"all_objective_scores": ordered_scores, "score": score, 'binary_classification_threshold': None}
             if isinstance(cv_pipeline, BinaryClassificationPipeline) and cv_pipeline.threshold is not None:
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -28,7 +28,6 @@ from evalml.data_checks import (
     EmptyDataChecks,
     HighVarianceCVDataCheck
 )
-from evalml.data_checks.data_check_message_type import DataCheckMessageType
 from evalml.exceptions import (
     AutoMLSearchException,
     PipelineNotFoundError,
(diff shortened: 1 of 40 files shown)


evalml/automl/automl_search.py
```diff
@@ -409,20 +409,18 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
...
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        y = _convert_woodwork_types_wrapper(y.to_series())
-
+        # X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        # y = _convert_woodwork_types_wrapper(y.to_series())
         self._set_data_split(X)
 
         data_checks = self._validate_data_checks(data_checks)
-        self._data_check_results = data_checks.validate(X, y)
+        self._data_check_results = data_checks.validate(_convert_woodwork_types_wrapper(X.to_dataframe()), _convert_woodwork_types_wrapper(y.to_series()))
```
angela97lin, line 417: Data checks currently don't support Woodwork types (to do in #1292 or later PR), so passing pandas to data checks.

evalml/automl/automl_search.py
```diff
@@ -645,7 +641,13 @@ def _compute_cv_scores(self, pipeline, X, y):
         start = time.time()
         cv_data = []
         logger.info("\tStarting cross validation")
-        for i, (train, test) in enumerate(self.data_split.split(X, y)):
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
```
angela97lin, line 645: In case `add_to_rankings` is called (not from `search`). From search, this should basically be a no-op, as the conversion to woodwork already happened.

evalml/tests/objective_tests/test_fraud_detection.py
```diff
@@ -32,7 +33,7 @@ def test_fraud_objective_function_amount_col(X_y_binary):
                           fraud_payout_percentage=.75,
                           amount_col="this column does not exist")
     y_predicted = pd.Series([.1, .5, .5])
-    y_true = [True, False, True]
+    y_true = pd.Series([True, False, True])
```
angela97lin, line 36: Updated these to Series (per our docstrings) :d Alternative is to check for lists and handle 🤷

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1715,3 +1715,15 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
     assert len(automl._results['errors']) == 1  # Raises exception at first error
     for e in automl._results['errors']:
         assert str(e) == msg
+
+
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_woodwork_passed_to_pipelines(mock_fit, mock_score, dummy_binary_pipeline_class, X_y_binary):
```
angela97lin, line 1722: Test that checks that woodwork is passed to pipelines :)

evalml/objectives/objective_base.py
```diff
@@ -73,28 +76,39 @@ def _standardize_input_type(y_in):
...
         Arguments:
-            y_predicted (pd.Series): Predicted values of length [n_samples]
+            y_predicted (pd.Series, pd.DataFrame, ww.DataColumn, or ww.DataTable): Predicted values of length [n_samples]
             y_true (pd.Series): Actual class labels of length [n_samples]
 
         Returns:
             None
         """
-        if len(y_predicted) != len(y_true):
+        is_multidimensional = False
```
angela97lin, line 93: If score needs predict_proba, we can have a multidimensional array to check, so updating. Not sure why this wasn't caught previously 🤔

evalml/objectives/lead_scoring.py
```diff
@@ -26,18 +24,15 @@ def objective_function(self, y_true, y_predicted, X=None):
...
             Returns:
-                float: profit per lead
+                float: Profit per lead
         """
-        if not isinstance(y_predicted, pd.Series):
-            y_predicted = pd.Series(y_predicted)
-
-        if not isinstance(y_true, pd.Series):
-            y_true = pd.Series(y_true)
+        y_true = self._standardize_input_type(y_true)
```
angela97lin, line 34: Updating to reuse our helper method (which handles DataColumns too)

evalml/automl/automl_search.py
```diff
@@ -645,7 +641,13 @@ def _compute_cv_scores(self, pipeline, X, y):
...
+
+        X_pd = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y_pd = _convert_woodwork_types_wrapper(y.to_series())
+        for i, (train, test) in enumerate(self.data_split.split(X_pd, y_pd)):
```
freddyaboulton, line 649: Do we need to convert to pandas? I think data tables are splitable:

image

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1715,3 +1715,15 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
...
+    X, y = X_y_binary
+    mock_score.return_value = {'Log Loss Binary': 1.0}
+    automl = AutoMLSearch(problem_type='binary', max_batches=4)
```
freddyaboulton, line 1725: Might as well check `mock_score` while we're at it?

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Do we check anywhere that passing in woodwork does not result in a conversion and that user-defined types are preserved throughout? Might be worth to do this on automl search on small scale: the only pipeline allowed uses mock components, and we only do two iterations.

evalml/objectives/fraud_cost.py
```diff
@@ -56,20 +61,19 @@ def objective_function(self, y_true, y_predicted, X):
...
             y_predicted (pd.Series): Predicted fraud labels
-             y_true (pd.Series): True fraud labels
-            X (pd.DataFrame): pd.Dataframe with transaction amounts
+            y_true (pd.Series): True fraud labels
+            X (ww.DataTable, pd.DataFrame): Data with transaction amounts
 
         Returns:
             float: Amount lost to fraud per transaction
         """
+        if isinstance(X, ww.DataTable):
```
freddyaboulton, line 70: Nice, I see we don't do any conversions if X is a pandas dataframe.

freddyaboulton (approved):
@angela97lin I think this is great! I noticed none of out automl tests use ww inputs. Not sure if the right thing to do is update all of our tests to use ww, add some ww-specific unit tests, or both but I think we should do at least one before merge lol. In particular, I think we need to be careful that the user-defined types are defined throughout.

evalml/automl/automl_search.py
angela97lin: @freddyaboulton KFolds works, but I don't think the other ones do (Stratified):

I think because of the attributes on/missing from DataTables, scikit-learn's `check_array` fails:

```
evalml/automl/automl_search.py:649: in _compute_cv_scores
../venv/lib/python3.7/site-packages/sklearn/model_selection/_split.py:735: in split
    y = check_array(y, ensure_2d=False, dtype=None)
../venv/lib/python3.7/site-packages/sklearn/utils/validation.py:72: in inner_f
    return f(**kwargs)
../venv/lib/python3.7/site-packages/sklearn/utils/validation.py:648: in check_array
    n_samples = _num_samples(array)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = array(<DataColumn: None (Physical Type = Int64) (Logical Type = WholeNumber) (Semantic Tags = {'numeric'})>,
      dtype=object)

    def _num_samples(x):
        """Return number of samples in array-like x."""
        message = 'Expected sequence or array-like, got %s' % type(x)
        if hasattr(x, 'fit') and callable(x.fit):
            # Don't get num_samples from an ensembles length!
            raise TypeError(message)

        if not hasattr(x, '__len__') and not hasattr(x, 'shape'):
            if hasattr(x, '__array__'):
                x = np.asarray(x)
            else:
                raise TypeError(message)

        if hasattr(x, 'shape') and x.shape is not None:
            if len(x.shape) == 0:
                raise TypeError("Singleton array %r cannot be considered"
>                               " a valid collection." % x)
E               TypeError: Singleton array array(<DataColumn: None (Physical Type = Int64) (Logical Type = WholeNumber) (Semantic Tags = {'numeric'})>,
E                     dtype=object) cannot be considered a valid collection.

../venv/lib/python3.7/site-packages/sklearn/utils/validation.py:196: TypeError
```

freddyaboulton: Ah thanks for that @angela97lin ! I forgot about the stratified split 😂 

The issue is that [deep-inside sklearn](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf39bbdacd6ed713c00724f8f871d60370/sklearn/utils/validation.py#L598), it calls `np.asarray` on the data column. This produces a weird result that doesn't have a meaningful shape. Totally cool to keep this as-is! Just wanted to make sure we weren't calling `to_series` if we didn't need to!

evalml/tests/automl_tests/test_automl.py
angela97lin: Yup, added `test_automl_woodwork_user_types_preserved` for this :D

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1715,3 +1715,65 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
     assert len(automl._results['errors']) == 1  # Raises exception at first error
     for e in automl._results['errors']:
         assert str(e) == msg
+
+
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_woodwork_user_types_preserved(mock_fit, mock_score, X_y_binary):
```
freddyaboulton, line 1722: 👍

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1715,3 +1715,65 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
...
+    X, y = X_y_binary
+    X = pd.DataFrame(X)
+    new_col = np.zeros(len(X))
+    new_col[:int(len(new_col) / 2)] = 1
+    X['cat col'] = pd.Series(new_col)
+    X = ww.DataTable(X, semantic_tags={'cat col': 'category'}, logical_types={'cat col': ww.logical_types.Categorical})
```
dsherry, line 1728: Is it necessary to pass in the semantic/logical info here? Woodwork should infer that "cat col" is binary/categorical, yes?

angela97lin: @dsherry Without passing it in directly, I think it infers the 'category' semantic tag but also numeric/categorical logic tags. Thus, I purposely set this to be what woodwork would *not* infer, and check that the user-specified edits are preserved :)

thehomebrewnerd: @angela97lin One Woodwork side note here - you can pass in the LogicalType specifiers as camel-case or snake-case strings as well, so you don't need to use the full `ww.logical_types.Categorical` syntax, but you could just pass `"Categorical"` instead - might keep the code a tad bit cleaner.

angela97lin: @thehomebrewnerd Oo awesome, thank you for this! Indeed, would make things cleaner--will update! 😁

thehomebrewnerd: The string syntax also helps in situations where you would otherwise have to `import ww` or similar in a file just for the sake of setting the types, and lets you eliminate that import.

evalml/objectives/fraud_cost.py
```diff
@@ -36,12 +39,14 @@ def decision_function(self, ypred_proba, threshold=0.0, X=None):
...
             ypred_proba (pd.Series): Predicted probablities
-            X (pd.DataFrame): Dataframe containing transaction amount
             threshold (float): Dollar threshold to determine if transaction is fraud
+            X (ww.DataTable, pd.DataFrame): Data containing transaction amounts
 
         Returns:
             pd.Series: pd.Series of predicted fraud labels using X and threshold
         """
+        if isinstance(X, ww.DataTable):
+            X = X.to_dataframe()
```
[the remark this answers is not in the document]
angela97lin: Yup, agreed. For the sake of getting the impl in, but I filed https://github.com/alteryx/evalml/issues/1427 along this line of thought 😁

evalml/objectives/objective_base.py
angela97lin: Ooo, interesting. I see, since in a lot of cases we use X_y_binary which results in numpy input, maybe our tests previously missed this?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1715,3 +1715,65 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
...
+def test_automl_woodwork_passed_to_pipelines_binary(mock_fit, mock_score, X_y_binary):
+    X, y = X_y_binary
+    X = pd.DataFrame(X)
+    mock_score.return_value = {'Log Loss Binary': 1.0}
+    automl = AutoMLSearch(problem_type='binary', max_batches=5)
+    automl.search(X, y)
+    for arg in mock_fit.call_args[0]:
+        assert isinstance(arg, (ww.DataTable, ww.DataColumn))
+    for arg in mock_score.call_args[0]:
+        assert isinstance(arg, (ww.DataTable, ww.DataColumn))
```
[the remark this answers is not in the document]
angela97lin: Haha I guess so! Originally, this was just to test that the woodwork instances were passed and `test_automl_woodwork_user_types_preserved` checked types are preserved but yes, in checking that the types are preserved, we're also checking that ww is passed. Removing!

evalml/objectives/fraud_cost.py
```diff
@@ -56,20 +61,19 @@ def objective_function(self, y_true, y_predicted, X):
...
+            X = _convert_woodwork_types_wrapper(X.to_dataframe())
         if not isinstance(X, pd.DataFrame):
             X = pd.DataFrame(X)
-
-        if not isinstance(y_predicted, pd.Series):
-            y_predicted = pd.Series(y_predicted)
-
-        if not isinstance(y_true, pd.Series):
-            y_true = pd.Series(y_true)
+        y_true = self._standardize_input_type(y_true)
```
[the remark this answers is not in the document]
angela97lin: Ah, it wasn't previously because `_standardize_input_type` only handled 1D inputs. Updated so this is no longer the case, and centralizing our standardizing logic in `_standardize_input_type` :)

evalml/objectives/objective_base.py
```diff
@@ -73,28 +76,39 @@ def _standardize_input_type(y_in):
...
+        if is_multidimensional:
+            if np.isnan(y_predicted).any().any() or np.isinf(y_predicted).any().any():
+                raise ValueError("y_predicted contains NaN or infinity")
+            if self.score_needs_proba and np.any([(y_predicted < 0) | (y_predicted > 1)]).any():
+                raise ValueError("y_predicted contains probability estimates not within [0, 1]")
+        else:
+            if np.isnan(y_predicted).any() or np.isinf(y_predicted).any():
+                raise ValueError("y_predicted contains NaN or infinity")
+            if self.score_needs_proba and np.any([(y_predicted < 0) | (y_predicted > 1)]):
+                raise ValueError("y_predicted contains probability estimates not within [0, 1]")
```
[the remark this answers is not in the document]
angela97lin: Woah, neat! `flatten` cleans this up a lot, thank you for the suggestion!! 😁

evalml/tests/objective_tests/test_fraud_detection.py
angela97lin: Just filed https://github.com/alteryx/evalml/issues/1456 for this!

evalml/objectives/fraud_cost.py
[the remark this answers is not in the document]
dsherry: Ah right, thank you!

Issue #1422, opened 2020-11-10
Investigate if binary classification problems still require numerical {0, 1} targets


angela97lin:
Now that we've implemented LabelEncoder, do we still need the check in `InvalidTargetDataCheck`? If not, we should delete it and update docs accordingly.

---

Is the last point still true? Now that we have the label encoder in the classification pipeline, I think this use case will work fine. If I'm right, can we file something to either delete this from `InvalidTargetDataCheck`, or change it to a warning instead of an error? Certainly, providing something which looks suspicious could indicate the user has provided incorrect data for some reason, which makes me think showing a warning is a good idea rather than just deleting.

_Originally posted by @dsherry in https://github.com/alteryx/evalml/pull/1412#discussion_r520073251_

freddyaboulton:
We technically don't need that part of the check anymore because @dsherry is right that the label encoder will automatically encode the values into [0, n_classes - 1]. I think we should change from error to warning because I'd be suspicious if someone meant to use a numeric column that doesn't have [0, 1] as the target value. It's possible they're using a numeric column that only happens to have two unique values in the sample but that doesn't correspond to any "classes" at all. 

```python
from evalml.pipelines import BinaryClassificationPipeline
from evalml.demos import load_breast_cancer

class DemoPipeline(BinaryClassificationPipeline):
    component_graph = ["Logistic Regression Classifier"]
    
X, y = load_breast_cancer()

y_ints = y.map({"malignant": 4.89, "benign": 8.78})

pl = DemoPipeline({})

pl.fit(X, y_ints)
```

```python
pl.predict_proba(X)
```

4.89 | 8.78
-- | --
1.000000 | 3.933512e-15
1.000000 | 2.638231e-08
1.000000 | 2.157034e-07
0.151240 | 8.487604e-01
0.999996 | 4.049651e-06

angela97lin:
@freddyaboulton You're absolutely right! I ran this locally and got the same results as you. Just realized that since I updated #1451 to convert the data check to return a warning instead of an error, this issue is resolved.

PR #1451, opened 2020-11-20
Update data check message to include a "code" and "details" fields

### LINKED ISSUE #1430: Data checks: JSON-friendly message fmt, include a type enum and affected column names

angela97lin:
Closes #1430

Also addresses #1422 by converting the check for binary classification targets != {0, 1} to return a warning instead of an error.

### DIFF
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -33,29 +33,33 @@ class ClassImbalanceDataCheck(DataCheck):
             y: Target labels to check for imbalanced data.
 
         Returns:
-            list (DataCheckWarning, DataCheckError): list with DataCheckWarnings if imbalance in classes is less than the threshold,
-                                                     and DataCheckErrors if the number of values for each target is below 2 * num_cv_folds.
+            dict: Dictionary with DataCheckWarnings if imbalance in classes is less than the threshold,
+                  and DataCheckErrors if the number of values for each target is below 2 * num_cv_folds.
 
         Example:
             >>> X = pd.DataFrame({})
             >>> y = pd.Series([0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
             >>> target_check = ClassImbalanceDataCheck(threshold=0.10)
-            >>> assert target_check.validate(X, y) == [DataCheckError("The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "ClassImbalanceDataCheck"), DataCheckWarning("The following labels fall below 10% of the target: [0]", "ClassImbalanceDataCheck")]
+        >>> assert target_check.validate(X, y) == {"errors": [{"message": "The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "data_check_name": "ClassImbalanceDataCheck", "level": "error"}],\
+                                                       "warnings": [{"message": "The following labels fall below 10% of the target: [0]", "data_check_name": "ClassImbalanceDataCheck", "level": "warning"}]}
         """
+        messages = {
+            "warnings": [],
+            "errors": []
+        }
         if not isinstance(y, pd.Series):
             y = pd.Series(y)
-        messages = []
         fold_counts = y.value_counts(normalize=False)
         # search for targets that occur less than twice the number of cv folds first
         below_threshold_folds = fold_counts.where(fold_counts < self.cv_folds).dropna()
         if len(below_threshold_folds):
             error_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
-            messages.append(DataCheckError(error_msg.format(self.cv_folds, below_threshold_folds.index.tolist()), self.name))
+            DataCheck._add_message(DataCheckError(error_msg.format(self.cv_folds, below_threshold_folds.index.tolist()), self.name), messages)
 
         counts = fold_counts / fold_counts.sum()
         below_threshold = counts.where(counts < self.threshold).dropna()
         # if there are items that occur less than the threshold, add them to the list of messages
         if len(below_threshold):
             warning_msg = "The following labels fall below {:.0f}% of the target: {}"
-            messages.append(DataCheckWarning(warning_msg.format(self.threshold * 100, below_threshold.index.tolist()), self.name))
+            DataCheck._add_message(DataCheckWarning(warning_msg.format(self.threshold * 100, below_threshold.index.tolist()), self.name), messages)
         return messages
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -47,12 +47,12 @@ class ClassImbalanceDataCheck(DataCheck):
         >>> assert target_check.validate(X, y) == {"errors": [{"message": "The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]",\
                                                                "data_check_name": "ClassImbalanceDataCheck",\
                                                                "level": "error",\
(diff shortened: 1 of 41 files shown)


evalml/data_checks/high_variance_cv_data_check.py
```diff
@@ -41,12 +46,17 @@ def validate(self, pipeline_name, cv_scores):
         if not isinstance(cv_scores, pd.Series):
             cv_scores = pd.Series(cv_scores)
 
+        variance = 0
         if cv_scores.mean() == 0:
-            high_variance_cv = 0
+            high_variance_cv = False
```
angela97lin, line 51: Same logic, but I think intention is more clear.

evalml/data_checks/high_variance_cv_data_check.py
```diff
@@ -41,12 +46,17 @@ def validate(self, pipeline_name, cv_scores):
...
+            variance = abs(cv_scores.std() / cv_scores.mean())
             high_variance_cv = abs(cv_scores.std() / cv_scores.mean()) > self.threshold
         # if there are items that occur less than the threshold, add them to the list of messages
         if high_variance_cv:
             warning_msg = f"High coefficient of variation (cv >= {self.threshold}) within cross validation scores. {pipeline_name} may not perform as estimated on unseen data."
-            DataCheck._add_message(DataCheckWarning(warning_msg, self.name), messages)
+            DataCheck._add_message(DataCheckWarning(message=warning_msg,
+                                                    data_check_name=self.name,
+                                                    message_code=DataCheckMessageCode.HIGH_VARIANCE,
+                                                    details={"variance": variance, "pipeline_name": pipeline_name}), messages)
```
dsherry, line 61: Cool!

evalml/data_checks/data_check_message.py
```diff
@@ -25,30 +29,27 @@ def __eq__(self, other):
...
+        message_dict = {
             "message": self.message,
-            "data_check_name": self.data_check_name
+            "data_check_name": self.data_check_name,
+            "code": self.message_code,
+            "level": self.message_type.value,
         }
+        if self.details is not None:
+            message_dict.update({"details": self.details})
+        return message_dict
```
dsherry, line 45: 👍

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -53,13 +65,21 @@ def validate(self, X, y):
...
         below_threshold = counts.where(counts < self.threshold).dropna()
         # if there are items that occur less than the threshold, add them to the list of messages
         if len(below_threshold):
+            below_threshold_values = below_threshold.index.tolist()
             warning_msg = "The following labels fall below {:.0f}% of the target: {}"
-            DataCheck._add_message(DataCheckWarning(warning_msg.format(self.threshold * 100, below_threshold.index.tolist()), self.name), messages)
+            DataCheck._add_message(DataCheckWarning(message=warning_msg.format(self.threshold * 100, below_threshold_values),
+                                                    data_check_name=self.name,
+                                                    message_code=DataCheckMessageCode.CLASS_IMBALANCE_BELOW_THRESHOLD,
+                                                    details={"target_values": below_threshold_values}), messages)
```
dsherry, line 84: Nice, this `details` field is going to be helpful

dsherry (approved):
🚢 ! 😁 

No blocking comments

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -40,8 +44,16 @@ def validate(self, X, y):
             >>> X = pd.DataFrame({})
             >>> y = pd.Series([0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
             >>> target_check = ClassImbalanceDataCheck(threshold=0.10)
-        >>> assert target_check.validate(X, y) == {"errors": [{"message": "The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]", "data_check_name": "ClassImbalanceDataCheck", "level": "error"}],\
-                                                       "warnings": [{"message": "The following labels fall below 10% of the target: [0]", "data_check_name": "ClassImbalanceDataCheck", "level": "warning"}]}
+        >>> assert target_check.validate(X, y) == {"errors": [{"message": "The number of instances of these targets is less than 2 * the number of cross folds = 6 instances: [0]",\
```
freddyaboulton, line 47: This is a great format! My only comment is that Enums are not actually json-serializable and the issue specifically mentions a "JSON-friendly" message format. I think doing `.name` on the code will be really useful for users! 

image

freddyaboulton (approved):
Thanks @angela97lin ! I think this is great. My only comment is that enums are not actually json serializable so we should convert to it to string in `validate`!

dsherry:
Oh shoot, that's a super great point @freddyaboulton , that python enums aren't JSON serializable. Yes, let's include the stringified version of the enum. (@angela97lin )

angela97lin:
@freddyaboulton @dsherry Ah, really good point--Will update this to include the stringified versions instead. Thanks for catching this!

evalml/data_checks/outliers_data_check.py
```diff
@@ -61,5 +66,8 @@ def get_IQR(df, k=2.0):
         has_outliers = ((X < iqr['lower_bound']) | (X > iqr['upper_bound'])).any()
         warning_msg = "Column '{}' is likely to have outlier data"
```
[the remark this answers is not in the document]
angela97lin: @dsherry Good point! I was curious why this data check and not, then, updating all of our data checks to follow this pattern... but for certain data checks, we have different thresholds or percentages that we also report (ex: % null) while there isn't information that's lost here. This may make our outputs less consistent, but I agree that it could be easier to see one data check for all the outliers, so I'll update this one :)

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -42,19 +47,34 @@ def validate(self, X, y):
...
+            messages["errors"].append(DataCheckError(message="Target does not have two unique values which is not supported for binary classification",
+                                                     data_check_name=self.name,
+                                                     message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_UNIQUE_VALUES,
+                                                     details={"target_values": unique_values}).to_dict())
 
         if len(value_counts) == 2 and y.dtype in numeric_and_boolean_dtypes:
-            unique_values = value_counts.index.tolist()
             if set(unique_values) != set([0, 1]):
-                messages["errors"].append(DataCheckError("Numerical binary classification target classes must be [0, 1], got [{}] instead".format(", ".join([str(val) for val in unique_values])), self.name).to_dict())
+                messages["warnings"].append(DataCheckWarning(message="Numerical binary classification target classes must be [0, 1], got [{}] instead".format(", ".join([str(val) for val in unique_values])),
```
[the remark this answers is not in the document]
angela97lin: @dsherry Right, but that's why we have `problem_type` passed in for the initialization of `InvalidTargetDataCheck`, right? Or are you concerned if the user accidentally passes in the wrong problem type?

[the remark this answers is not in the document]
dsherry: Yep, that's what I was thinking, if someone accidentally ran binary classification with a regression target. One quick way around this would be to show the number of uniques and the counts of the 100 most frequent uniques instead of the counts of all the uniques!

No need to add this to this PR heh. But this would be great to file as a performance enhancement.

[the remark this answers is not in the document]
angela97lin: Ah, got it! Filed #1460 :)

PR #1459, opened 2020-11-23
Add LightGBM Regressor

### LINKED ISSUE #296: Add LightGBM regression component
[Documentation](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMRegressor.html)

Mentioned in #293. #295 tracks LightGBM classification pipeline

bchen1116:
fix #296 

Perf tests [here](https://alteryx.quip.com/HTTuArJve1uW/LightGBM-Regressor-Perf-Tests)
Documentation [here](https://evalml.alteryx.com/en/bc_296_lgbm_regressor/generated/evalml.pipelines.components.LightGBMRegressor.html#evalml.pipelines.components.LightGBMRegressor)


evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
```diff
@@ -0,0 +1,96 @@
...
+    """LightGBM Regressor"""
+    name = "LightGBM Regressor"
+    hyperparameter_ranges = {
+        "learning_rate": Real(0.000001, 1),
+        "boosting_type": ["gbdt", "dart", "goss", "rf"],
+        "n_estimators": Integer(10, 100),
+        "max_depth": Integer(0, 10),
+        "num_leaves": Integer(2, 100),
+        "min_child_samples": Integer(1, 100),
+        "bagging_fraction": Real(0.000001, 1),
```
dsherry, line 28: Is 0 an invalid setting for `bagging_fraction`?

evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class LightGBMRegressor(Estimator):
+    """LightGBM Regressor"""
+    name = "LightGBM Regressor"
+    hyperparameter_ranges = {
+        "learning_rate": Real(0.000001, 1),
+        "boosting_type": ["gbdt", "dart", "goss", "rf"],
+        "n_estimators": Integer(10, 100),
+        "max_depth": Integer(0, 10),
```
dsherry, line 25: @bchen1116 do you have any sense of what the fit time of lightgbm is like when max_depth or n_estimators is near the max value? We've observed larger values for similar parameters for other tree-based models can cause long fit times with no benefit in accuracy. I do see that these ranges align with some of our other tree-based estimators, including [catboost](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/estimators/regressors/catboost_regressor.py#L26) which we've had problems with in the past, so that's good.

evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
```diff
@@ -0,0 +1,96 @@
...
+        "bagging_fraction": Real(0.000001, 1),
+        "bagging_freq": Integer(0, 1)
+    }
+    model_family = ModelFamily.LIGHTGBM
+    supported_problem_types = [ProblemTypes.REGRESSION]
+
+    SEED_MIN = 0
+    SEED_MAX = SEED_BOUNDS.max_bound
+
+    def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, max_depth=0, num_leaves=31, min_child_samples=20, n_jobs=-1, random_state=0, bagging_fraction=0.9, bagging_freq=0, **kwargs):
```
dsherry, line 37: @bchen1116 any particular reason for choosing these default values? It could be interesting to do some scanning on these parameters and see if they should be tweaked. `max_depth=0` jumps out to me as being too low--the default [in lightgbm is -1](https://lightgbm.readthedocs.io/en/latest/Parameters.html#max_depth) i.e. "no limit", but I don't think we should do that.

evalml/tests/component_tests/test_lgbm_regressor.py
```diff
@@ -0,0 +1,225 @@
...
+    X2 = pd.DataFrame(X)
+    X2.columns = np.arange(len(X2.columns))
+    clf = LightGBMRegressor()
+    clf.fit(X, y)
+    arg_X = mock_fit.call_args[0][0]
+    np.testing.assert_array_equal(arg_X, X2)
+
+    clf.predict(X)
+    arg_X = mock_predict.call_args[0][0]
+    np.testing.assert_array_equal(arg_X, X2)
```
dsherry, line 121: @bchen1116 could you remind me what this unit test is for?

dsherry (approved):
@bchen1116 nice going! Great work debugging that weird looking glass error, haha.

I left some questions about the parameters, and some comments on the perf test doc. I think it would be cool to change the default `max_depth` to something greater than 0, and see how that impacts the accuracy in the first batch. It would be cool to try perf testing a couple other tweaks too but not required.

evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
bchen1116: Yep, LightGBM requires it to be strictly larger than 0 when provided

evalml/tests/component_tests/test_lgbm_regressor.py
bchen1116: Mm good call, I had originally put this in as a test for no categorical data, but a few tests prior to this covers that instance. I'll fix this!

evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
bchen1116: I am not sure. I can run some perf tests monday to experiment with these though!

bchen1116: Perf tests are in the updated doc, and parameter defaults have been changed!

bchen1116:
Perf docs updated with new tests that determine the idea `n_estimators` and `max_depth` values.

freddyaboulton (approved):
@bchen1116 This is great! The performance tests to see the impact of different default parameter values are awesome. 

There's a lot of overlap between this class and the lgbm classifier. It'd be nice if we could share at least the logic for overriding the parameters in init and `_encode_categories`. What do you think? Not blocking and probably best to do in another issue anyways if you agree.

bchen1116:
@freddyaboulton definitely agree! It would be nice to run some perf tests on the LGBM Classifier and see if we can find better default values, especially since we have much more test data for classification. I think that is a good separate issue, which I've just submitted [here](https://github.com/alteryx/evalml/issues/1564)

PR #1473, opened 2020-11-26
Update woodwork to >=0.0.5 for core-requirements.txt

angela97lin:
Because Woodwork updated their API from `to_pandas` to `to_series`/`to_dataframe` in 0.0.5, we should update our core-requirements s.t. woodwork >= 0.0.5. Woodwork users for any versions before will run into issues with our library!

@rpeck FYI :)


freddyaboulton (approved):
Looks good!

PR #1476, opened 2020-11-30
Update `detect_problem_types` implementation

### LINKED ISSUE #1469: Update `numeric_dtypes`
Update the values that are considered numeric types in `gen_utils.numeric_dtypes`. Woodwork supports many additional numeric dtypes that aren't captured by the `numeric_dtypes` array we currently have in EvalML. 

Repro code:
```python
import woodwork as ww
from evalml.utils.gen_utils import numeric_dtypes

data = ww.demo.load_retail(nrows=100)
print(data.types)
for column in data.columns:    
    print(f'"{column}" is numeric: {data[column].dtype in numeric_dtypes}')
```

Result that is displayed:
image

We see that column `quantity` is of type `Int64`, but that isn't in `numeric_dtypes`. 

*`numeric_dtypes` is used in `detect_problem_type`, which is called in the Looking Glass code for `analyze_metadata.py`.

bchen1116:
fix #1469 

Updated this implementation to catch Int64 dtypes. Previously, using `analyze_metadata` in looking_glass, if the target data was `Int64`, the problem_type would be classified as multiclass as long as there were `>2` unique values. 

After using `is_numeric_dtype`:
image

Since we drop NaN data, classifying Boolean (nullable) as a numeric dtype is ok, since we'll catch this binary case before determining if it is regression or multiclass.

### DIFF
--- a/evalml/problem_types/utils.py
+++ b/evalml/problem_types/utils.py
@@ -1,8 +1,7 @@
 import pandas as pd
+from pandas.api.types import is_numeric_dtype
 
 from .problem_types import ProblemTypes
-
-from pandas.api.types import is_numeric_dtype
 
 
 def handle_problem_types(problem_type):
--- a/evalml/problem_types/utils.py
+++ b/evalml/problem_types/utils.py
@@ -2,7 +2,7 @@ import pandas as pd
 
 from .problem_types import ProblemTypes
 
-from evalml.utils.gen_utils import numeric_dtypes
+from pandas.api.types import is_numeric_dtype
 
 
 def handle_problem_types(problem_type):
@@ -46,7 +46,7 @@ def detect_problem_type(y):
         raise ValueError("Less than 2 classes detected! Target unusable for modeling")
     if num_classes == 2:
         return ProblemTypes.BINARY
-    if y.dtype in numeric_dtypes:
+    if is_numeric_dtype(y.dtype):
         if (num_classes > 10):
             return ProblemTypes.REGRESSION
     return ProblemTypes.MULTICLASS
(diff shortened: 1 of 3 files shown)


jeremyliweishih (approved):
good catch! LGTM

freddyaboulton (approved):

evalml/problem_types/utils.py
```diff
@@ -46,7 +45,7 @@ def detect_problem_type(y):
         raise ValueError("Less than 2 classes detected! Target unusable for modeling")
     if num_classes == 2:
         return ProblemTypes.BINARY
-    if y.dtype in numeric_dtypes:
+    if is_numeric_dtype(y.dtype):
```
dsherry, line 48: Got it, thanks @bchen1116 !

I think the ultimate goal is to update this (and all our utilities) to standardize to woodwork (#1229 ) and then check if the "numeric" semantic tag has been applied to the target. (@angela97lin FYI)

evalml/tests/problem_type_tests/test_problem_types.py
```diff
@@ -95,6 +95,14 @@ def test_detect_problem_type_regression():
     assert detect_problem_type(y_float) == ProblemTypes.REGRESSION
 
 
+def test_numeric_extensions():
+    y_Int64 = pd.Series([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], dtype='Int64')
+    y_Int64_null = pd.Series([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, None], dtype='Int64')
```
dsherry, line 100: Does this change if you specify `np.nan` instead of `None`? If numpy/pandas parse these to the same thing, no need to add test coverage; if they're different, we should add coverage.

bchen1116: Since I use `pd.dropna()` in the beginning, all NaN/None/null values are treated the same.

Issue #1383, opened 2020-10-30
Run performance tests on time series regression problems


freddyaboulton:

freddyaboulton:
Results in this [document](https://alteryx.quip.com/KZ9YAa5N9rQb/Time-Series-Regression-Performance-Tests)!

PR #1479, opened 2020-11-30
DateTimeFeaturizer encodes features as ints

### LINKED ISSUE #1477: DateTimeFeaturizer should return integer-valued features
The `DateTimeFeaturizer` returns string-valued features for month and day of week. These are useful features but the string values necessitate having an encoder downstream in the pipeline. The `DateTimeFeaturizer` should handle the encoding and provide a mapping so users know how the encoding happened. 

Proposed output:

image

New api method:
```python
dt = DateTimeFeaturizer()
dt.fit(X, y)
dt.get_feature_names()
```

```python
{"Date_month: ["January", ..., "December"],
 "Date_day_of_week: ["Monday, ..., "Sunday"]
}

### PR DESCRIPTION
**Pull request by freddyaboulton**
### Pull Request Description
Fixes #1477 

-----

### DIFF
--- a/docs/source/user_guide/pipelines.ipynb
+++ b/docs/source/user_guide/pipelines.ipynb
@@ -387,7 +387,7 @@
     "\n",
     "class CustomPipeline(MulticlassClassificationPipeline):\n",
     "    name = \"Custom Pipeline\"\n",
-    "    component_graph = ['Imputer', MyDropNullColumns, 'One Hot Encoder', 'Random Forest Classifier']\n",
+    "    component_graph = ['Imputer', MyDropNullColumns, 'DateTime Featurization Component', 'One Hot Encoder', 'Random Forest Classifier']\n",
     "    custom_hyperparameters={\n",
     "        \"Imputer\": {\n",
     "            \"numeric_impute_strategy\": ['mean', 'median']\n",
@@ -401,8 +401,8 @@
     "# This string can then be pasted into a separate window and run, although since the pipeline has custom component `MyDropNullColumns`, \n",
     "#      the code for that component must also be included\n",
     "\n",
-    "from evalml.demos import load_breast_cancer\n",
-    "X, y = load_breast_cancer()\n",
+    "from evalml.demos import load_fraud\n",
+    "X, y = load_fraud()\n",
     "exec(code)\n",
     "pipeline.fit(X, y)"
    ]
--- a/docs/source/user_guide/pipelines.ipynb
+++ b/docs/source/user_guide/pipelines.ipynb
@@ -387,7 +387,7 @@
     "\n",
     "class CustomPipeline(MulticlassClassificationPipeline):\n",
     "    name = \"Custom Pipeline\"\n",
-    "    component_graph = ['Imputer', MyDropNullColumns, 'DateTime Featurization Component', 'One Hot Encoder', 'Random Forest Classifier']\n",
+    "    component_graph = ['Imputer', MyDropNullColumns, 'One Hot Encoder', 'Random Forest Classifier']\n",
     "    custom_hyperparameters={\n",
     "        \"Imputer\": {\n",
     "            \"numeric_impute_strategy\": ['mean', 'median']\n",
--- a/docs/source/user_guide/pipelines.ipynb
+++ b/docs/source/user_guide/pipelines.ipynb
@@ -401,8 +401,8 @@
     "# This string can then be pasted into a separate window and run, although since the pipeline has custom component `MyDropNullColumns`, \n",
     "#      the code for that component must also be included\n",
     "\n",
-    "from evalml.demos import load_fraud\n",
-    "X, y = load_fraud()\n",
+    "from evalml.demos import load_breast_cancer\n",
+    "X, y = load_breast_cancer()\n",
     "exec(code)\n",
     "pipeline.fit(X, y)"
    ]
--- a/evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
@@ -6,7 +6,7 @@ from evalml.utils.gen_utils import (
 )
 
 
-def _extract_year(col):
+def _extract_year(col, encode_as_categories=False):
     return col.dt.year, None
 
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py [hunk 1]
```diff
@@ -7,19 +7,33 @@
```
freddyaboulton, line 7: This implementation takes 150 ms to create features for 150k rows. I think that's reasonable.

image

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
```diff
@@ -74,7 +89,21 @@ def transform(self, X, y=None):
...
             for feature in features_to_extract:
-                X_t[f"{col_name}_{feature}"] = self._function_mappings[feature](X_t[col_name])
+                name = f"{col_name}_{feature}"
+                features, categories = self._function_mappings[feature](X_t[col_name])
+                X_t[name] = features
+                if categories:
+                    self._categories[name] = categories
         return X_t.drop(self._date_time_col_names, axis=1)
+
+    def get_feature_names(self):
```
freddyaboulton, line 102: I named it `get_feature_names` to match the OHE.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -105,6 +105,8 @@ def score(self, X, y, objectives):
             X = pd.DataFrame()
         X = _convert_to_woodwork_structure(X)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
+        y = _convert_woodwork_types_wrapper(y.to_series())
```
freddyaboulton, line 109: I noticed this was missing 😬

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
```diff
@@ -7,19 +7,33 @@
 
 
 def _extract_year(col):
-    return col.dt.year
+    return col.dt.year, None
```
freddyaboulton, line 10: I debated whether or not we should encode the years starting at 0 but that sounds like a performance test? I can file an issue but it'd be blocked by #1383

ParthivNaresh: Would encoding years take away contextual information about the datetime later on in the pipeline? January will always be mapped to 0 but 1981 being mapped to 0 would take information wouldn't it? Or maybe I'm misunderstanding!

ParthivNaresh:
Looks great! I don't know if you want a test checking the get_feature_names output with an input that has just year and hour datetime information because function mapping return None for those. Not really a big deal anyways

ParthivNaresh (approved):

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
freddyaboulton: I had a thought that linear models would work best with scaled features. Now that I think about it again, we automatically add a standard scaler for linear models at least in AutoML so maybe we can keep as is for now!

angela97lin (commented):
Hmm, I'm curious and don't know too much about the pros / cons of encoding dates as ints vs categories as they are right now. I assume for time series regression problems, it makes sense to encode as int, as there's valuable information about how close two dates are (1 and 2 are much closer than 1 and 30). But would it also be useful to use categories for day of the week, where the magnitude might not necessarily matter? 🤔

freddyaboulton:
@angela97lin and I just chatted about her comment about whether or not day-of-week should be treated as a category. We agreed that we shouldn't do that by default because it forces the user to have a ohe downstream (or else some estimators would crash, like xgboost) which could lead to an explosion of columns (if there are many delayed day-of-week features, which is likely). These extra columns may not increase performance and would have a negative impact on downstream modules, like permutation importance in model understanding.

For the short-term, we'll add a `encode_as_categories` parameter to init that defaults to False. In the long term, Automl should be "smart enough" try different encodings for the same columns per estimator.

docs/source/user_guide/pipelines.ipynb
```diff
@@ -387,7 +387,7 @@
     "\n",
     "class CustomPipeline(MulticlassClassificationPipeline):\n",
     "    name = \"Custom Pipeline\"\n",
-    "    component_graph = ['Imputer', MyDropNullColumns, 'DateTime Featurization Component', 'One Hot Encoder', 'Random Forest Classifier']\n",
+    "    component_graph = ['Imputer', MyDropNullColumns, 'One Hot Encoder', 'Random Forest Classifier']\n",
```
freddyaboulton, line 390: While #1495 gets fixed.

docs/source/user_guide/pipelines.ipynb
```diff
@@ -401,8 +401,8 @@
     "# This string can then be pasted into a separate window and run, although since the pipeline has custom component `MyDropNullColumns`, \n",
     "#      the code for that component must also be included\n",
     "\n",
-    "from evalml.demos import load_fraud\n",
-    "X, y = load_fraud()\n",
+    "from evalml.demos import load_breast_cancer\n",
```
bchen1116, line 404: Was there a reason why this was changed? Run time?

bchen1116 (approved):
LGTM!

docs/source/user_guide/pipelines.ipynb
freddyaboulton: Not run time! Since I removed the date time featurizer, we can't use fraud because there's a datetime column. An alternative is to drop the datetime column and not switch datasets but I felt that could distract from the code gen.

bchen1116: Got it, thanks!

kmax12:
im jumping in a little late to the convo here, but in case this nuance wasn't discussed, wanted to chime in.

one of the reasons we might encode the day of the week or month as a category, rather than int is to not confuse the the estimator of the cyclic nature of the variable e.g january (0) and december (11) are actually only one month a part. A linear estimator can learn that january and december are close in time when encode as 0 and 11. 

on the other hand, the numeric encoding makes it easier in other cases for the classifier to learn that january and febuary are close in time compared to the categorical approach. 

all that being said, I'm not sure what the the best default option is for us. mainly documenting the situation. 

here's SO post with some more discussion: https://datascience.stackexchange.com/questions/17759/encoding-features-like-month-and-hour-as-categorial-or-numeric

PR #1480, opened 2020-11-30
Update allow list for deps update bot

dsherry:
Include woodwork, featuretools etc.


angela97lin (approved):
LGTM! 😁 Will be useful to use everything from core-requirements.txt, but hopefully doesn't trigger toooo often?

evalml/tests/dependency_update_check/make_deps_diff.sh
```diff
@@ -0,0 +1,3 @@
+allow_list=$(cat core-requirements.txt requirements.txt | grep -oE "^[a-zA-Z0-9\-_]+" | paste -d "|" -s -)
```
freddyaboulton, line 1: The grep picks up both scikit-learn and scikit-optimize as just "scikit". Is that ok?

image

freddyaboulton (approved):
@dsherry This looks great! Does updating the circle-ci check update the github dependency bot? (I don't really know how it works hehe).

I have a question about the grep but it's mainly just for my understanding.

jeremyliweishih:
@freddyaboulton the updates to the bot are under the changes to `.github/workflows/dependency_check.yml`. Basically it just outputs the file and the bot detects a change in the branch and puts up a PR.

freddyaboulton:
@jeremyliweishih Thanks! I get it now 😅

evalml/tests/dependency_update_check/make_deps_diff.sh
dsherry: @freddyaboulton yes because we're using `grep -E` which uses a regex, and "scikit" will match both sklearn and skopt.

PR #1481, opened 2020-11-30
Update data checks to accept Woodwork data structures

### LINKED ISSUE #1465: Update all data checks to support Woodwork

angela97lin:
Closes #1465

- Should we update *all* tests to use Woodwork? ex: we currently usually test pandas data structures, and then have one test to check for other inputs (lists, arrays, etc.). I added Woodwork to that test in most cases, but is that against the principle of having Woodwork be the main expected input?

From discussion with @dsherry and @freddyaboulton we agreed that as long as we have some test that makes sure Woodwork inputs work as well as pandas/numpy, that suffices in most cases where we're not using any Woodwork metadata.

- In some tests, I removed checks for 1D lists or arrays. These would get converted to DataColumns, then to Series, and would not have the appropriate attributes/methods that our method impl were looking for. This was previously not as much of an issue because we just did a check and converted our input to DataFrames (regardless of 1D or not). This update is more limiting, but more consistent--the user must pass in a 2D structure for `X`. Would love to hear suggestions/thoughts though!
- Is it possible to use `woodwork_wrapper` in multiple places (outside of data checks)? May be difficult given different parameters across different methods.


evalml/tests/data_checks_tests/test_id_columns_data_check.py
```diff
@@ -125,21 +126,17 @@ def test_id_cols_data_check_input_formats():
     # test empty pd.DataFrame
     assert id_cols_check.validate(pd.DataFrame()) == {"warnings": [], "errors": []}
 
-    #  test list
-    assert id_cols_check.validate([1, 2, 3, 4, 5]) == {
```
freddyaboulton, line 129: I think it's fine to remove this test case in favor of passing in a 2d dataframe or data table. But I wonder why we even supported it in the beginning? Was the goal to also check if a single column (or maybe the target values) were id?

evalml/utils/gen_utils.py
```diff
@@ -378,3 +378,13 @@ def _not_nan(pd_data):
 
     mask = np.logical_and(_not_nan(pd_data_1), _not_nan(pd_data_2))
     return pd_data_1.iloc[mask], pd_data_2.iloc[mask]
+
+
+def woodwork_wrapper(func):
```
freddyaboulton, line 383: I think this would also work for pipelines/estimators but I haven't tested it. I think another alternative is to have this be a function rather than a wrapper.

So we would go from 
```python
@woodwork_wrapper
def my_method(self, X, y, ...):
```
to 
```python
def my_method(self, X, y, ...):
    X, y = woodwork_wrapper(X, y)
```

I think this would be the most general (since it can apply to more than just methods that take `self`) and also a bit less mysterious. For example, in `HighlyNullDataCheck`, our docstrings say we accept pd DataFrames, but then we call 

```python
```

and `to_dataframe()` is not a dataframe method. So someone reading the code wight be confused as to how it works. 

I think this implementation is great (and it potentially works for more than just data checks) but I wanted to propose an alternative!

freddyaboulton (approved):
@angela97lin Thanks for doing this! 

Should we update AutoMLSearch to not [convert to pandas](https://github.com/alteryx/evalml/blob/1465_data_checks_ww/evalml/automl/automl_search.py#L434) before running data checks?

You raise a good point about whether we should use ww as the primary data type in our tests! I think this applies to components/pipelines and not just data checks. 

I'd be in favor of continuing our pattern of using pandas as the "primary" data type and having one (or a couple) unit tests with a `parametrize` that checks that the feature works as expected given all possible data types. The reason being that (I think) most of our features don't leverage the logical/semantic types in the table so I don't see a lot of value in  updating all of our tests.

That being said, I think there are some places where we don't follow the pattern I just mentioned so we should update those tests. And features that leverage logical/semantic types should be tested against lots of different inferred and user-defined types. (I think that's the case but just being explicit).

Curious to hear what others think. Certainly happy to brainstorm on a call or something together.

evalml/utils/gen_utils.py
bchen1116: I agree here, I think if users were to look at our source code, they might be confused why we can make the `to_dataframe()` or `to_series()` calls.

bchen1116 (approved):
Looks great! I agree with @freddyaboulton in keeping the pandas tests for the reasons he mentioned. Also agree that the `@woodwork_wrapper` would be more understandable from the user's perspective as a call inside the method instead of as a function wrapper.

dsherry:
@angela97lin RE your question in the description:
> Should we update all tests to use Woodwork? ex: we currently usually test pandas data structures, and then have one test to check for other inputs (lists, arrays, etc.). I added Woodwork to that test in most cases, but is that against the principle of having Woodwork be the main expected input?

Yes I think we should make that update! And we should also have separate tests which ensure that if a pandas dataframe is provided to the data checks, it gets converted to woodwork correctly.

evalml/utils/gen_utils.py
dsherry: @freddyaboulton and @bchen1116 thanks for this!

When @angela97lin and I discussed this recently, we liked the decorator approach. However your opinions here are making me think there's not much difference between doing this in a decorator and doing it in a helper method, except that the decorator adds some complexity by hiding the conversion. In light of that, I'd be in favor of using a helper over a decorator. @angela97lin what do you think?

evalml/tests/data_checks_tests/test_id_columns_data_check.py
dsherry: Yep I think this was just checking that a 1d vector would get treated just like a 2d dataframe with one column, right? I agree its fine to use 2d input here instead, although I do think it would be nice to have a test case with 1d input too.

evalml/utils/gen_utils.py
angela97lin: Thanks for the suggestions! I agree, even when implementing the wrapper, it seemed a bit silly :d I like the idea of not using the wrapper.

In response to all of this: is an additional helper method really necessary? Or it is it okay to just continue with adding
`X = _convert_to_woodwork_structure(X)` and `y = _convert_to_woodwork_structure(y)` to methods that need it? 

@freddyaboulton I've thought about a helper that takes in both `X` and `y` before in hopes that it'd be a bit cleaner, but at least for pipelines/components/other methods, it wasn't always the case that we had both inputs (X and y), and it seemed more clear to me that each input was explicitly transformed separately and then their types converted via `_convert_woodwork_types_wrapper` in a parallel manner right after, so I decided to keep them separate. What are your thoughts on this?

freddyaboulton: @angela97lin I see what you mean! I think it's fine to keep the pattern we have where we call `_convert_to_woodwork_structure` and `_convert_woodwork_types_wrapper` for `X` and `y` separately. I don't think it's too verbose and it's explicit about what's happening. 

What I was suggesting is converting the wrapper you have now into a helper that guarantees that both `X` and `y` are either ww types or None. That's basically what your wrapper is doing now and I think we could use it for components, pipelines, any method. As a case study, here's what `ComponentBase.fit` looks like now:

```python
def fit(self, X, y=None):
    if y is not None:
    try:
        self._component_obj.fit(X, y)
        return self
    except AttributeError:
        ....
```
With the "conversion wrapper" it'd be:

```python
def fit(self, X, y=None):
    X, y = convert_to_woodwork(X, y)
    if y is not None:
    try:
        self._component_obj.fit(X, y)
        return self
    except AttributeError:
```

It's very similar but I think it makes our conventions clearer: When data comes into evalml it should be converted to woodwork. Whether it gets turned back to pandas is up to the logic of the component/pipeline/function.

Let me know what you think! Like I said, I like the pattern we had before and we don't have to let this discussion hold up the merge.

evalml/tests/data_checks_tests/test_id_columns_data_check.py
angela97lin: Yeah, I think so. That worked "fine" because we converted any input we got into a DataFrame. Current updated logic will see that this is 1D and update to a DataColumn, which will then convert back to a Series. Issue is, Series objects don't have some methods available, breaking our logic so I'm removing this. If we see it fit / useful to provide 1D inputs, we can add back? I didn't think it was worth cluttering the logic with extra dimension checks/different logic 🤔

evalml/utils/gen_utils.py
angela97lin: @freddyaboulton Ah, I think I see what you mean. I think for now, it'll be okay to keep as is. There are times where we don't necessarily use/need X/y, and this could make it a little clearer when something is ignored. If it gets unwieldy later we can revisit and refactor. This discussion was great though, thanks a lot 😁

dsherry (commented):
@angela97lin looks good! 👍

PR #1478, opened 2020-11-30
Remove `copy_dataframe` parameter, update requirements with release of Woodwork v0.0.6

angela97lin:
- Remove use of `copy_dataframe` parameter which was removed in 0.0.6
- Update Woodwork requirement to >=0.0.6

PR #1484, opened 2020-12-01
Pin woodwork version

dsherry:
In order to avoid breaking changes (#1478 ), let's pin the woodwork version in each evalml release, so that `pip install evalml` always produces a working installation.


jeremyliweishih (approved):
Good idea!

angela97lin (approved):
Yup, sounds good 👍 This'll help @jeff-hernandez too, who was running into this issue a couple of times!

freddyaboulton (approved):
Thanks @dsherry !

PR #1487, opened 2020-12-01
Add script to generate github markdown for release notes

dsherry:
Usage:
* Make release branch, update version, update release notes
* Run `tools/format_format_release_notes.sh` to print out the markdown to paste into the release PR and github release

It was only a matter of time 😂  https://xkcd.com/1319/

### DIFF
--- a/tools/format_release_notes.sh
+++ b/tools/format_release_notes.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+if [ -z "$1" ] || [ "$1" = "-h" ]
+  then
+    echo -e 'Grab the "Future Releases" section from the release notes and format it in markdown for GitHub releases
+
+Usage: tools/format_release_notes.sh 0.16.1'
+    exit 0
+fi
+new_evalml_version=$1
+content_raw=$(cat docs/source/release_notes.rst |  sed -n -e '/^\*\*Future Releases\*\*/,/^\*\*v/p;/^\*\*v/q' | sed '/^\.\. warning::$/d' | tail -n +2 | sed '$ d' | awk 'NF')
+date_formatted=$(date '+%b. %-d, %Y')
+content_markdown=$(echo -n "${content_raw}" | sed 's/^        \* /- /g' | sed 's/^    \* /### /g' | sed 's/    \*\*Breaking Changes\*\*/### Breaking Changes/g' | sed -E 's/:pr:\`([0-9]+)\`/#\1/g')
+full_markdown=$(echo -e "# v${new_evalml_version} ${date_formatted}\n${content_markdown}")
+echo -e "${full_markdown}"
--- a/tools/format_release_notes.sh
+++ b/tools/format_release_notes.sh
@@ -8,7 +8,7 @@ Usage: tools/format_release_notes.sh'
     exit 0
 fi
 evalml_version=$(sed -n -E "s/.*version=\'([0-9A-Za-z\.\_]+)\'.*/\1/p" setup.py)
-content_raw=$(sed -n -e "/^\*\*v0.16.1/,/^\*\*v0.16.0/p;/^\*\*v0.16.0/q" docs/source/release_notes.rst | sed '/^\.\. warning::$/d' | tail -n +2 | sed '$ d' | awk 'NF')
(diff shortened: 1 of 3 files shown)


tools/format_release_notes.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [ "$1" = "-h" ]
+  then
+    echo -e 'Grab the "Future Releases" section from the release notes and format it in markdown for GitHub releases
+
+Usage: tools/format_release_notes.sh'
+    exit 0
+fi
+evalml_version=$(sed -n -E "s/.*version=\'([0-9A-Za-z\.\_]+)\'.*/\1/p" setup.py)
+content_raw=$(sed -n -e "/^\*\*v0.16.1/,/^\*\*v0.16.0/p;/^\*\*v0.16.0/q" docs/source/release_notes.rst | sed '/^\.\. warning::$/d' | tail -n +2 | sed '$ d' | awk 'NF')
```
angela97lin, line 11: This is like black magic to me 😂 But I tried to use it and it seems to always output v0.16.1--is that because this line hardcodes it?

jeremyliweishih: This could be:

```
prev_version = $(grep -o -E -m 1 'v[0-9]\.[0-9]+\.[0-9]' docs/source/release_notes.rst)
sed -n -e "/^\*\*Future Releases/,/^\*\*${prev_version}/p;/^\*\*${prev_version}/q" docs/source/release_notes.rst | sed '/^\.\. warning::$/d' | tail -n +2 | sed '$ d' | awk 'NF'
```

I checked the grep command but not sure if I substituted the previous version into sed properly 😄

jeremyliweishih (dismissed):
Looks good but needs to grab between Future Releases and the latest version.

tools/format_release_notes.sh
dsherry: Very good point! Thanks for catching me here, I forgot that this was hard-coded in 🤦 

Yep, could grep for prev version. I opted for separating out the before vs the after.

tools/format_release_notes.sh
```diff
@@ -0,0 +1,17 @@
...
+
+Note: this was written to work properly on mac. It should work on linux if -r is used instead of -E for sed.'
+    exit 0
+fi
+evalml_version=$(sed -n -E "s/.*version=\'([0-9A-Za-z\.\_]+)\'.*/\1/p" setup.py)
+content_raw=$(sed -n -E '/^\*\*v[A-Za-z0-9\., ]+\*\*$/,$p' docs/source/release_notes.rst | tail -n +2 | sed -E '/\*\*v[A-Za-z0-9\., ]+\*\*$/q' | sed '$ d' | awk 'NF')
+date_formatted=$(date '+%b. %-d, %Y')
+content_markdown=$(echo -n "${content_raw}" | sed 's/^        \* /- /g' | sed 's/^    \* /### /g' | sed 's/    \*\*Breaking Changes\*\*/### Breaking Changes/g' | sed -E "s/:pr:\`([0-9]+)\`/#\1/g")
+full_markdown=$(echo -e "# v${evalml_version} ${date_formatted}\n${content_markdown}")
+echo -e "${full_markdown}"
```
dsherry, line 17: @angela97lin @jeremyliweishih I think I fixed the hard-coding here, and I broke things up so it should be a bit easier to follow! It was pretty soupy before.

@angela97lin you mentioned this is like magic--I agree, unix utilities are beautiful magic ✨  quick breakdown of what's going on:
* `var=$(...)`: evaluating commands in sub-shells to save the output.
* `sed` is a super handy find-replace utility. I used it on line 12 to define a regex group and extract the matching version number as `evalml_version`
* Then by including "p" and "q" commands with `sed` on line 13, we're telling it to print everything after a match or to print everything up to a match, respectively. That lets us get all the lines between the last release header (i.e. the one about to go out) and the release header before that.
* `tail`, `sed '$ d'` and `awk 'NF'` are just hacks to remove leading/trailing lines and whitespace
* The usage of `date` below gets the date in the exact human-readable format we use it in our release notes.
* The markdown part just uses a sequence of `sed` commands to replace specific line prefixes with their markdown counterparts, and then to convert our pr tags to use a hash for markdown.
* Finally we put it all together in `full_markdown` and output the result at the end!

freddyaboulton (approved):
Very cool @dsherry ! I just tested it out locally and it works as expected! 

Thanks for explaining what was happening with all of the `sed` and `awk` commands.

PR #1490, opened 2020-12-02
Fix windows unit test CI

dsherry:
Thanks @freddyaboulton for the tips in slack! He mentioned:
> I think the issue is that the latest release of shap has numba as a dependency but numba released version 0.52.0 this afternoon which I guess can't be built on windows with pip

My suggested fix: install `numba` using conda, and mention this on the install page in the "Windows" section.


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  windows: circleci/windows@2.2.0
+  windows: circleci/windows@2.4.0
```
dsherry, line 4: This wasn't needed for this PR, I just figured we should increment

freddyaboulton (approved):
@dsherry Thanks for the fix! The docs change look great to me.

PR #1483, opened 2020-12-01
Plot for time series regression

freddyaboulton:
### Pull Request Description
Fixes #1258

Example with a year's worth of (synthetic) data:
image

Example with a month's worth of data:
image

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #1494, opened 2020-12-02
Fix custom index bug in get_prediction_vs_actual_over_time_data

freddyaboulton:
### Pull Request Description
After merging #1483 I realized that `get_prediction_vs_actual_over_time_data` does not support a custom index! This fixes that.

-----

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -561,25 +561,28 @@ def get_prediction_vs_actual_over_time_data(pipeline, X, y, dates):
     """Get the data needed for the prediction_vs_actual_over_time plot.
 
     Arguments:
-        pipeline (PipelineBase): Fitted time series regression pipeline.
+        pipeline (TimeSeriesRegressionPipeline): Fitted time series regression pipeline.
         X (pd.DataFrame): Features used to generate new predictions.
         y (pd.Series): Target values to compare predictions against.
+        dates (pd.Series): Dates corresponding to target values and predictions.
 
     Returns:
        pd.DataFrame
     """
-    return pd.DataFrame({"dates": dates,
-                         "target": y,
-                         "prediction": pipeline.predict(X, y)})
+    prediction = pipeline.predict(X, y)
+    return pd.DataFrame({"dates": dates.reset_index(drop=True),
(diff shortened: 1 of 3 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -561,25 +561,28 @@ def get_prediction_vs_actual_over_time_data(pipeline, X, y, dates):
     """Get the data needed for the prediction_vs_actual_over_time plot.
 
     Arguments:
-        pipeline (PipelineBase): Fitted time series regression pipeline.
+        pipeline (TimeSeriesRegressionPipeline): Fitted time series regression pipeline.
```
dsherry, line 564: Ah good call!

evalml/model_understanding/graphs.py
```diff
@@ -561,25 +561,28 @@ def get_prediction_vs_actual_over_time_data(pipeline, X, y, dates):
...
     Returns:
        pd.DataFrame
     """
-    return pd.DataFrame({"dates": dates,
-                         "target": y,
-                         "prediction": pipeline.predict(X, y)})
+    prediction = pipeline.predict(X, y)
+    return pd.DataFrame({"dates": dates.reset_index(drop=True),
+                         "target": y.reset_index(drop=True),
+                         "prediction": prediction.reset_index(drop=True)})
```
dsherry, line 575: @freddyaboulton hmm, why is it necessary to drop the index here? I guess I'd expect the index here to be the same as it was for X and y. In some cases the index could actually be the dates.

I'm assuming we're doing this here in order for the graph method to work. If so, could we move these `reset_index` calls into the graph method instead?

dsherry (approved):
LGTM! Left one suggestion on the impl

evalml/model_understanding/graphs.py
freddyaboulton: The problem is that if any of these series (target, prediction, or dates) have a different index (which is possible if "target" is coming from a data split) then the dataframe will have NaNs where the indices don't match. In that case, you couldn't reset the index to the graph method.

dsherry: Ok that sounds good to me. I thought the preds and target would have the same shape though?

PR #1496, opened 2020-12-02
Add Time Series Baseline Regression Components and Pipelines

### LINKED ISSUE #1482: Baseline Pipeline for Time Series Regression problems
As an evalml user, I'd like AutoMLSearch to compare my time series pipelines against a time-series-specific baseline (predicting the mean is not a reasonable strategy). This would make the "percent better" calculations meaningful and would help me decide how well my pipelines are doing.

A valid strategy we should implement is the [naive forecast method](https://en.wikipedia.org/wiki/Forecasting#Na.C3.AFve_approach), which predicts the current target as the previous target.

One thing we should keep in mind is that our time series regression pipelines output NaNs for the first `max_delay` number of rows. We should do the same to make sure the pipelines are evaluated on the same data (this is a minor detail since `max_delay` is typically small compared to the size of the data).

jeremyliweishih:
Fixes #1482.

- [x] Add regressor
- [x] Add pipeline
- [x] Add to automl


evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,74 @@
...
+            X = pd.DataFrame()
+        if y is None:
+            raise ValueError("Cannot predict Time Series Baseline Regressor if y is None")
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
+        y = _convert_woodwork_types_wrapper(y.to_series())
+
+        first = y.iloc[0]
+        y_hat = y.shift(periods=1)
```
jeremyliweishih, line 61: Naive forecasting where y_t = y_(t-1) but I keep the 0th index the same (y_0 = y_0).

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,74 @@
...
+    def predict(self, X, y=None):
+        if X is None:
+            X = pd.DataFrame()
+        if y is None:
+            raise ValueError("Cannot predict Time Series Baseline Regressor if y is None")
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
+        y = _convert_woodwork_types_wrapper(y.to_series())
+
```
jeremyliweishih, line 59: Naive forecasting where y_t = y_(t-1) but I keep the 0th index the same (y_0 = y_0).

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_algorithms.py
```diff
@@ -70,7 +72,10 @@ def test_value_errors_raised(mock_tree_explainer, pipeline, exception, match):
         pytest.importorskip("catboost", "Skipping test because catboost is not installed.")
 
     with pytest.raises(exception, match=match):
-        _ = _compute_shap_values(pipeline({}), pd.DataFrame(np.random.random((2, 16))))
+        if pipeline == TimeSeriesBaselineRegressionPipeline:
```
freddyaboulton, line 75: Thanks for updating this test! Up to you but I would prefer to skip the if/else since you can specify pipeline parameters for any pipeline class:

```python
    with pytest.raises(exception, match=match):
        _ = _compute_shap_values(pipeline({"pipeline": {"gap": 0, "max_delay": 0}}), pd.DataFrame(np.random.random((2, 16))))
```

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -7,14 +7,6 @@
 from evalml.pipelines import TimeSeriesRegressionPipeline
 
 
-@pytest.fixture
```
freddyaboulton, line 10: 👍

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1221,8 +1233,12 @@ def fit(self, *args, **kwargs):
     mock_score_1 = MagicMock(return_value=mock_scores)
     DummyPipeline.score = mock_score_1
 
-    automl = AutoMLSearch(problem_type=problem_type, max_iterations=2,
-                          allowed_pipelines=[DummyPipeline], objective="auto")
+    if problem_type == "time series regression":
```
freddyaboulton, line 1236: I think we can eliminate the if/else here too. My concern is having to maintain the if/else when time series binary and multiclass get added. 

The good thing about the if is that it makes it clear in which cases the `problem_configuration` is used. Maybe we can add a comment to the test saying something like "Specifying problem_configuration for all problem types for conciseness"?

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -105,6 +111,8 @@ def score(self, X, y, objectives):
             X = pd.DataFrame()
         X = _convert_to_woodwork_structure(X)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
```
freddyaboulton, line 114: Yep. Had to do this in #1479 as well !

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,74 @@
...
+    """Time series regressor that predicts using the naive forecasting approach.
+
+    This is useful as a simple baseline regressor for time series problems
+    """
+    name = "Time Series Baseline Regressor"
+    hyperparameter_ranges = {}
+    model_family = ModelFamily.BASELINE
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, random_state=0, **kwargs):
```
freddyaboulton, line 23: We should take in `gap` here so that we know to shift by `1` in that case.

evalml/pipelines/regression/timeseries_baseline_regression.py
```diff
@@ -0,0 +1,7 @@
+from evalml.pipelines import TimeSeriesRegressionPipeline
+
+
+class TimeSeriesBaselineRegressionPipeline(TimeSeriesRegressionPipeline):
+    """Baseline Pipeline for time series regression problems."""
+    _name = "Time Series Baseline Regression Pipeline"
+    component_graph = ["Delayed Feature Transformer", "Time Series Baseline Regressor"]
```
freddyaboulton, line 7: I don't think we need the Delayed Feature Transformer in this case?

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,74 @@
...
+        if X is None:
+            X = pd.DataFrame()
+        if y is None:
+            raise ValueError("Cannot predict Time Series Baseline Regressor if y is None")
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_to_woodwork_structure(y)
+        y = _convert_woodwork_types_wrapper(y.to_series())
+
+        first = y.iloc[0]
```
freddyaboulton, line 60: I don't think we should do any shifting here and `predict` should basically be a no-op that return the `y` that is input. 

The `score` method in `TimeSeriesRegressionPipeline` will shift the target by `-gap` which will ensure the target is `gap` time periods ahead of the prediction. Right now, with this `shift` they would be offset by `gap + 1` which I don't think is intended. Let's say the user has dates October 1-8 and the gap is 3. I think the desired pairs of prediction and target dates should be:

| Prediction Date      | Target Date |
| ----------- | ----------- |
| 2020-10-01      | 2020-10-04       |
| 2020-10-02   | 2020-10-05        |
| 2020-10-03 | 2020-10-06|
| 2020-10-04 | 2020-10-07|
| 2020-10-05 | 2020-10-08|

We need to be mindful of the `gap=0` case, cause we should shift by `1` in this `predict` method to avoid perfect overlap in the pipeline `score` method.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -86,7 +88,11 @@ def predict(self, X, y=None, objective=None):
         y = _convert_woodwork_types_wrapper(y.to_series())
 
         features = self.compute_estimator_features(X, y)
-        predictions = self.estimator.predict(features.dropna(axis=0, how="any"))
+
+        if 'y' in inspect.signature(self.estimator.predict).parameters:
```
freddyaboulton, line 92: I see. So right now the baseline regressor for time series is the only one that takes `y` in the predict method. This is great but can you please add a comment explaining why we need the call to inspect? I wonder if there's some other mechanism we can use here?

evalml/automl/automl_search.py
```diff
@@ -591,9 +592,10 @@ def _add_baseline_pipelines(self, X, y):
             baseline = ModeBaselineBinaryPipeline(parameters={})
         elif self.problem_type == ProblemTypes.MULTICLASS:
             baseline = ModeBaselineMulticlassPipeline(parameters={})
-        else:
+        elif self.problem_type == ProblemTypes.REGRESSION:
             baseline = MeanBaselineRegressionPipeline(parameters={})
-
+        else:
+            baseline = TimeSeriesBaselineRegressionPipeline(parameters={"pipeline": {"gap": 0, "max_delay": 0}})
```
freddyaboulton, line 598: I think we should use the problem configuration parameters passed by the user to ensure the shifting between targets and predictions is accurate.

freddyaboulton (changes requested):
@jeremyliweishih Thank you so much! This is great.

I think this is almost good to merge but I think we need to tweak the implementation of the regressor a bit to ensure the predictions and targets are offset by the right amount by the time we get to pipeline `score`.

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
jeremyliweishih: @freddyaboulton  got it! will make the changes.

evalml/automl/automl_search.py
jeremyliweishih: right, thanks!

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,73 @@
...
+    name = "Time Series Baseline Regressor"
+    hyperparameter_ranges = {}
+    model_family = ModelFamily.BASELINE
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, gap=1, random_state=0, **kwargs):
+        """Baseline time series regressor that predicts using the naive forecasting approach.
+
+        Arguments:
+            random_state (int, np.random.RandomState): seed for the random number generator
```
angela97lin, line 27: Add `gap` to arguments?

evalml/utils/gen_utils.py
```diff
@@ -169,7 +169,7 @@ def _get_subclasses(base_class):
     return subclasses
 
 
-_not_used_in_automl = {'BaselineClassifier', 'BaselineRegressor',
+_not_used_in_automl = {'BaselineClassifier', 'BaselineRegressor', 'TimeSeriesBaselineRegressor',
```
angela97lin, line 172: We want to add `TimeSeriesBaselineRegressionPipeline` too right?

evalml/pipelines/components/estimators/regressors/timeseries_baseline_regressor.py
```diff
@@ -0,0 +1,73 @@
...
+    """Time series regressor that predicts using the naive forecasting approach.
+
+    This is useful as a simple baseline regressor for time series problems
+    """
+    name = "Time Series Baseline Regressor"
+    hyperparameter_ranges = {}
+    model_family = ModelFamily.BASELINE
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, gap=1, random_state=0, **kwargs):
```
angela97lin, line 23: Also, I'm assuming it doesn't make sense if gap is negative (or 0?)? Can we add a check for that and raise / warn if that's true?

evalml/pipelines/regression/timeseries_baseline_regression.py
```diff
@@ -0,0 +1,36 @@
...
+            pd.Series: Predicted values.
+        """
+        if X is None:
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+
+        features = self.compute_estimator_features(X, y)
```
freddyaboulton, line 34: not-blocking:

Since we don't really need to call `compute_estimator_features`, I vote we get rid of it.

```python
predictions = self.estimator.predict(None, y)
return pad_with_nans(predictions, max(0, X.shape[0] - predictions.shape[0]))
```

What do you think? I like passing `None` to `estimator.predict` just to be clear that we only need the targets. That's the only reason this class needs it's own implementation of `fit`.

evalml/tests/pipeline_tests/regression_pipeline_tests/test_timeseries_baseline_regression.py
```diff
@@ -0,0 +1,21 @@
...
+def test_time_series_baseline(ts_data):
+    X, y = ts_data
+
+    clf = TimeSeriesBaselineRegressionPipeline(parameters={"pipeline": {"gap": 1, "max_delay": 1}})
+    clf.fit(X, y)
+
+    np.testing.assert_allclose(clf.predict(X, y), y)
+
+
+def test_time_series_baseline_no_X(ts_data):
```
freddyaboulton, line 15: Can you please add a test that verifies that the predictions and targets are offset by the right amount in `score` for both the `gap=0` and `gap > 0` cases? This can be similar to `test_score_drop_nans` in `test_time_series_pipeline`.

evalml/pipelines/regression/timeseries_baseline_regression.py
```diff
@@ -0,0 +1,36 @@
...
+    """Baseline Pipeline for time series regression problems."""
+    _name = "Time Series Baseline Regression Pipeline"
+    component_graph = ["Time Series Baseline Regressor"]
+
+    def predict(self, X, y=None, objective=None):
+        """Make predictions using selected features.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, or np.ndarray): Data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series, np.ndarray, None): The target training targets of length [n_samples]
```
freddyaboulton, line 21: not-blocking: Maybe we shouldn't set `y` to `None` by default and update the docstring since `None` isn't allowed?

freddyaboulton (approved):
@jeremyliweishih I think this is great! Thank you so much for helping on this. Before merging, I suggest we add a test that verifies that the predictions and targets are offset by the right amount in the baseline pipeline (just cause that's a pretty big deal and I want to make sure we have something in place to guard against regressions in future refactorings). I think we should also add the baseline pipeline and estimator to the api ref?

Other than that, I left some comments that don't need to be resolved before merging!

PR #1505, opened 2020-12-03
Reduce Unit Test Memory

### LINKED ISSUE #1438: Unit tests use up to 20GB of memory on circleci
**Problem**
While debugging intermittent test failures on PR #1410, @christopherbunn and I measured memory usage of the unit tests on circleci and found a complete end-to-end run can use up to 20GB at peak.

That's way more than I would have expected... the question is, why?

**Observations**
We ssh'ed into a circleci box running on `main` and ran the following [using `memory-profiler`](https://pypi.org/project/memory-profiler/):
```python
mprof run  --include-children pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v
```
Which created the following plot, visible with `mprof plot`:

I ran this twice and got a similar plot, so the results appear to be consistent across runs.

This is dangerously close to the max memory allowed on the circleci worker size we're using. That's why we started looking into this -- on #1410, we saw that the memory usage went 5GB higher for some reason.

freddyaboulton:
### Pull Request Description
Fixes #1438

Only 10 gb on circle-ci! This is half of what it was before when we had `n 8` in the pytest config and about 4 gb less than the current footprint with `n 4`

This is also faster I think. I recall seeing times around 270-300 seconds before making this change.

1438-latest

-----


evalml/tests/conftest.py
```diff
@@ -272,24 +272,48 @@ def regression_core_objectives():
...
         if (set(supported_problem_types) == {ProblemTypes.REGRESSION, ProblemTypes.TIME_SERIES_REGRESSION} and
             estimator_class.model_family not in _nonstackable_model_families and
                 estimator_class.model_family != ModelFamily.ENSEMBLE):
-            stackable_regressors.append(estimator_class())
+            stackable_regressors.append(helper_functions.safe_init_with_njobs_1(estimator_class))
     return stackable_regressors
+
+
+@pytest.fixture
+def helper_functions():
```
freddyaboulton, line 299: Not sure if this actually belongs in our utils module since it is only used for tests.

Makefile
```diff
@@ -22,15 +22,15 @@ test:
...
 
 .PHONY: circleci-test-minimal-deps
 circleci-test-minimal-deps:
-	pytest evalml/ -n 4 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v --has-minimal-dependencies
+	pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v --has-minimal-dependencies
 
 .PHONY: win-circleci-test
 win-circleci-test:
-	pytest evalml/ -n 4 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v
+	pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v
```
dsherry, line 33: Oh amazing!! I wasn't expecting we could pump this back up.

jeremyliweishih (approved):
nice.

evalml/tests/conftest.py
dsherry: @freddyaboulton good point, yep test seems right!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -978,6 +981,7 @@ def test_score_with_objective_that_requires_predict_proba(mock_predict, dummy_re
         clf = dummy_regression_pipeline_class(parameters={})
         clf.fit(X, y)
         clf.score(X, y, ['precision', 'auc'])
+    # Why don't we use pytest.raises here?
```
dsherry, line 984: @freddyaboulton delete this comment?

As to the question: I don't think there's any particular reason! `pytest.raises` would work fine here. I guess we're checking for multiple substrings in the error message, but I think there's a way to access the exception with `pytest.raises`...

evalml/tests/conftest.py
```diff
@@ -272,24 +272,48 @@ def regression_core_objectives():
...
+    class Helpers:
+        @staticmethod
+        def safe_init_with_njobs_1(component_class):
```
dsherry, line 302: `safe_init_component_with_njobs_1`?

evalml/tests/conftest.py
```diff
@@ -272,24 +272,48 @@ def regression_core_objectives():
...
+
+        @staticmethod
+        def safe_init_pipeline_with_njobs_1(pipeline_class):
+            try:
+                estimator = pipeline_class.component_graph[-1]
+                estimator_name = estimator if isinstance(estimator, str) else estimator.name
+                pl = pipeline_class({estimator_name: {'n_jobs': 1}})
+            except ValueError:
+                pl = pipeline_class({})
+            return pl
```
dsherry, line 317: Looks good. It could mess with call counts if someone were mocking / introspecting on the component or pipeline constructor or something, but I don't think any of our current tests do that so nbd.

So we're not using this helper in places where we pass other parameters into the component or pipeline?

dsherry (approved):
@freddyaboulton that's an amazing result!! Well done 😁 

Even if we don't know the precise root cause of the memory issues, I think your results mean we can now say with certainty that the memory issues we saw were caused by our usage of the `n_jobs` parameter.

I didn't have any big changes to suggest. One question though: do we still have some test coverage of running with `n_jobs=-1` or `n_jobs>1`? I just wanna make sure we still have coverage for that case. Ideally, we'd have a test which ensures you can fit each estimator with `n_jobs=-1`, and then another test which ensures `n_jobs` is threaded through from automl down into the components (I think we have that coverage in place).

Defining a separate job to handle testing parallelism is an option available to us. Although certainly we don't need to do that in this PR. @christopherbunn and I were discussing the possibility of defining a separate "parallelization" integration test job for the parallel work. I think for now he's going to try adding the dask client fixture to our existing unit test job.

angela97lin (approved):
Awesome debug work on this 🤩

evalml/tests/pipeline_tests/test_pipelines.py
dsherry: @angela97lin oooh, great detective work! So it was for codecov, interesting. We've relaxed the limits for codecov a bit since then, so this may work now 🤷‍♂️

evalml/tests/conftest.py
freddyaboulton: Yep exactly. We have some tests that iterate over all estimators/pipelines, init with default parameters, and verify some functionality so we use these helpers to just change the `n_jobs` since it's not critical to what the test is verifying.

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: Of course I was confused about my own code 😅  I will delete the comment and keep as-is. Thanks for digging into this @angela97lin !

evalml/tests/conftest.py
freddyaboulton: Done! Good call.

freddyaboulton:
@dsherry I couldn't find any unit tests that explicitly check that the right value of n_jobs is passed to the pipelines so I added one for AutoMLSearch and IterativeAlgorithm. Good call! We have coverage with n_jobs set to -1 and 1 for all estimators that accept n_jobs so I'm confident about that.

I think the idea you and @christopherbunn have about defining a separate testing job for parallelism is great! I think we need to be careful about having parallelism across and within tests at the same time.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1568,6 +1573,39 @@ def test_iterative_algorithm_pipeline_hyperparameters_make_pipeline(mock_fit, mo
...
+        component_graph = [MockEstimatorWithNJobs]
+
+    automl = AutoMLSearch(problem_type='binary', n_jobs=3, max_batches=2,
+                          allowed_pipelines=[Pipeline1, Pipeline2, dummy_binary_pipeline_class])
+    automl.search(X, y)
+    for parameters in automl.full_rankings.parameters:
+        if "Mock Classifier with njobs" in parameters:
+            assert parameters["Mock Classifier with njobs"]["n_jobs"] == 3
+        else:
+            assert all("n_jobs" not in component_params for component_params in parameters.values())
```
dsherry, line 1606: @freddyaboulton looks great!! Thanks for adding

PR #1510, opened 2020-12-07
1463 mean absolute percentage error

### LINKED ISSUE #1463: Add Mean Absolute Percentage Error (MAPE)
For timeseries.

https://en.wikipedia.org/wiki/Mean_absolute_percentage_error

@freddyaboulton

ParthivNaresh:
Fixes #1463

### DIFF
--- a/evalml/objectives/standard_metrics.py
+++ b/evalml/objectives/standard_metrics.py
@@ -345,6 +345,17 @@ class MAE(RegressionObjective):
         return metrics.mean_absolute_error(y_true, y_predicted)
 
 
+class MAPE(RegressionObjective):
+    """Mean absolute percentage error for time series regression."""
+    name = "Mean Absolute Percentage Error"
+    greater_is_better = False
+    score_needs_proba = False
+    perfect_score = 0.0
+
+    def objective_function(self, y_true, y_predicted, X=None):
+        return (np.abs((y_true - y_predicted) / y_true))[y_true != 0].mean() * 100
+
+
 class MSE(RegressionObjective):
     """Mean squared error for regression."""
     name = "MSE"
--- a/evalml/objectives/standard_metrics.py
+++ b/evalml/objectives/standard_metrics.py
@@ -9,6 +9,7 @@ from .multiclass_classification_objective import (
     MulticlassClassificationObjective
 )
 from .regression_objective import RegressionObjective
+from .time_series_regression_objective import TimeSeriesRegressionObjective
 
 
 class AccuracyBinary(BinaryClassificationObjective):
@@ -345,15 +346,20 @@ class MAE(RegressionObjective):
         return metrics.mean_absolute_error(y_true, y_predicted)
 
 
-class MAPE(RegressionObjective):
-    """Mean absolute percentage error for time series regression."""
+class MAPE(TimeSeriesRegressionObjective):
+    """Mean absolute percentage error for time series regression. Scaled by 100 to return a percentage.
+
+    Only valid for nonzero inputs. Otherwise, will throw a ValueError
+    """
     name = "Mean Absolute Percentage Error"
     greater_is_better = False
-    score_needs_proba = False
     perfect_score = 0.0
 
     def objective_function(self, y_true, y_predicted, X=None):
-        return (np.abs((y_true - y_predicted) / y_true))[y_true != 0].mean() * 100
+        if (y_true == 0).any():
+            raise ValueError("Mean Absolute Percentage Error cannot be used when "
+                             "targets contain the value 0.")
+        return (np.abs((y_true - y_predicted) / y_true)).mean() * 100
 
 
 class MSE(RegressionObjective):
--- a/evalml/objectives/time_series_regression_objective.py
+++ b/evalml/objectives/time_series_regression_objective.py
@@ -1,0 +1,13 @@
+from .regression_objective import RegressionObjective
+
+from evalml.problem_types import ProblemTypes
+
+
+class TimeSeriesRegressionObjective(RegressionObjective):
+    """Base class for all time series regression objectives.
+
+    problem_types (list(ProblemType)): List of problem types that this objective is defined for.
+        Set to [ProblemTypes.TIME_SERIES_REGRESSION]
+    """
+
+    problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
--- a/evalml/tests/dependency_update_check/latest_dependency_versions.txt
+++ b/evalml/tests/dependency_update_check/latest_dependency_versions.txt
@@ -8,8 +8,8 @@ ipywidgets==7.5.1
 lightgbm==3.1.0
 nlp-primitives==1.1.0
 numpy==1.19.4
-pandas==1.1.4
-plotly==4.13.0
+pandas==1.1.5
+plotly==4.14.0
 psutil==5.7.3
 requirements-parser==0.2.0
 scikit-learn==0.23.2
--- a/evalml/tests/dependency_update_check/latest_dependency_versions.txt
+++ b/evalml/tests/dependency_update_check/latest_dependency_versions.txt
@@ -5,7 +5,7 @@ colorama==0.4.4
 featuretools==0.22.0
 graphviz==0.15
 ipywidgets==7.5.1
-lightgbm==3.1.0
+lightgbm==3.1.1
 nlp-primitives==1.1.0
 numpy==1.19.4
 pandas==1.1.5
--- a/evalml/tests/objective_tests/test_standard_metrics.py
+++ b/evalml/tests/objective_tests/test_standard_metrics.py
@@ -7,6 +7,7 @@ from sklearn.metrics import matthews_corrcoef as sk_matthews_corrcoef
 
 from evalml.objectives import (
     F1,
+    MAPE,
     MSE,
     AccuracyBinary,
     AccuracyMulticlass,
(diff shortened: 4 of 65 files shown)


evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +345,17 @@ def objective_function(self, y_true, y_predicted, X=None):
...
 
 
+class MAPE(RegressionObjective):
+    """Mean absolute percentage error for time series regression."""
+    name = "Mean Absolute Percentage Error"
+    greater_is_better = False
+    score_needs_proba = False
+    perfect_score = 0.0
+
+    def objective_function(self, y_true, y_predicted, X=None):
```
freddyaboulton, line 355: Maybe we should just raise a ValueError if any `y_true` are 0? My concern is that 0 can be a meaningful target value (like stock returns) and users may be mislead that those records are being used in the evaluation when they are not.

evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +345,17 @@ def objective_function(self, y_true, y_predicted, X=None):
         return metrics.mean_absolute_error(y_true, y_predicted)
 
 
+class MAPE(RegressionObjective):
+    """Mean absolute percentage error for time series regression."""
+    name = "Mean Absolute Percentage Error"
+    greater_is_better = False
+    score_needs_proba = False
```
freddyaboulton, line 352: nit-pick: You don't need to specify `score_needs_proba` since that is specified in `RegressionObjective`

evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +345,17 @@ def objective_function(self, y_true, y_predicted, X=None):
         return metrics.mean_absolute_error(y_true, y_predicted)
 
 
+class MAPE(RegressionObjective):
```
freddyaboulton, line 348: I think we should add to the api reference too

freddyaboulton (approved):
@ParthivNaresh This looks great! I think we should resolve the discussion about what to do with division by 0 before merging but other than that I think it's good to merge! 

I'm ok with holding off with making this the default objective for time series regression problems because of its [noted limitations](https://en.wikipedia.org/wiki/Mean_absolute_percentage_error#Issues), namely the bias in selecting estimators that under-estimate the target.

I think for the default objective for automl we should consider these two alternatives:
1. [Mean Absolute Scaled Error](https://en.wikipedia.org/wiki/Mean_absolute_scaled_error)
2. [Log forecast to Actual](https://poseidon01.ssrn.com/delivery.php?ID=919102102125114084027127082015014024125011038043027056100114112024102116028092091119017122037004103000001023067093098125105096001014042028065078087113084005126022037011055081071017115026000087023008123083066070022120086113103100120020126116115009101&EXT=pdf)

evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +345,17 @@ def objective_function(self, y_true, y_predicted, X=None):
...
+    """Mean absolute percentage error for time series regression."""
+    name = "Mean Absolute Percentage Error"
+    greater_is_better = False
+    score_needs_proba = False
+    perfect_score = 0.0
+
+    def objective_function(self, y_true, y_predicted, X=None):
+        return (np.abs((y_true - y_predicted) / y_true))[y_true != 0].mean() * 100
```
dsherry, line 356: @ParthivNaresh nice, looks good!

Sklearn 0.24.2 (a current release candidate) [has a MAPE impl](https://github.com/scikit-learn/scikit-learn/blob/4773f3e39/sklearn/metrics/_regression.py#L197) ([docs](https://scikit-learn.org/dev/modules/generated/sklearn.metrics.mean_absolute_percentage_error.html)).

I like [their use of `max(y_true, eps)`](https://github.com/scikit-learn/scikit-learn/blob/4773f3e39/sklearn/metrics/_regression.py#L261) in the denominator. I think we should do that instead of excluding rows where `y_true` is exactly 0.

I also liked that they had weight support. We don't have a pattern for this yet from other objectives, and its not a requirement for this PR, but would be nice to have.

evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +345,17 @@ def objective_function(self, y_true, y_predicted, X=None):
         return metrics.mean_absolute_error(y_true, y_predicted)
 
 
+class MAPE(RegressionObjective):
+    """Mean absolute percentage error for time series regression."""
```
dsherry, line 349: Should we mention that we've scaled this by 100 so that its a percentage? I've seen it defined both with or without the 100 factor in the past.

evalml/tests/objective_tests/test_standard_metrics.py
```diff
@@ -437,6 +438,23 @@ def test_mcc_catches_warnings():
...
+
+    s2_actual = np.array([1, 2, 3, 2, 1, 2, 1, 0])
+    s2_predicted = np.array([0, 2, 2, 1, 3, 2, 0, 1])
+
+    s3_actual = np.array([1.5, 2.5, -1.5, 0, 1, 3, 2.4, -3.2])
+    s3_predicted = np.array([1.5, 1.5, -1, 1, 1.1, -3, 2, -3])
+
+    assert obj.score(s1_actual, s1_predicted) == pytest.approx(3.5 / 6 * 100)
+    assert obj.score(s2_actual, s2_predicted) == pytest.approx(4.83333333 / 7 * 100)
+    assert obj.score(s3_actual, s3_predicted) == pytest.approx(3.0625 / 7 * 100)
```
dsherry, line 455: @ParthivNaresh nice 👍  I was able to walk through by hand and reproduce the math for the first example you have here. Only thought here is that if you could simplify the numbers a bit, make the arrays shorter,  make some of those decimals more round etc., it would help make this math easier to follow.

evalml/objectives/standard_metrics.py
dsherry, line 348: @freddyaboulton did we decide not to define `TimeseriesRegressionObjective`? Or did we simply not decide yet, haha, since this is the first objective explicitly defined for timeseries? I can't remember where we left that conversation.

dsherry (approved):
@ParthivNaresh looks great!! The only blocking comments I had were about using `eps` in the impl, and starting a thread with @freddyaboulton about whether or not we should define `TimeseriesRegressionObjective` right now

evalml/objectives/standard_metrics.py
ParthivNaresh: @dsherry That's an interesting approach, but wouldn't that result in an extremely high value in the overall calculation? @freddyaboulton

freddyaboulton: Good point @dsherry! sklearn's definition of `eps` is way too small but even if we used our own I'm worried that there'll always be one use case where that value is "wrong".

I prefer treating this how we treat RootMeanSquaredLogError or MeanSquaredLogError. Those objectives error out if passed targets with any negative values. They are also kept out of the core objectives because the user only uses them when they are defined for their particular dataset. Maybe we should add MAPE to `get_non_core_objectives`

image


image

evalml/objectives/standard_metrics.py
freddyaboulton: Previously, our objectives were only defined for a single problem type. We agreed to change this to a list of valid problem types and that our existing regression objectives should work for both regression and time series regression problems.

So long story short, we didn't define a `TimeSeriesRegressionObjective` because we didn't have any objectives that should only work for time series. We should make MAPE a `TimeSeriesRegressionObjective` in that case :) Good call.

dsherry: Ah got it, thanks @freddyaboulton 

So @ParthivNaresh , I think what we should do here is:
* Define `TimeseriesRegressionObjective` which subclasses `RegressionObjective` but allows only `ProblemTypes.TIME_SERIES_REGRESSION`
* Have `MAPE` subclass `TimeseriesRegressionObjective`

That sound good?

evalml/objectives/standard_metrics.py
dsherry: > sklearn's definition of eps is way too small

Hmm, its an arbitrary threshold, right? Setting it to min numerical precision makes sense to me because any differences smaller than that (more or less) cannot be trusted.

It seems we have a few options for what to do when the target contains values equal to or super close to zero (options not all mutually exclusive):
1. Have the objective error out.
2. Have the objective round the zero value up to a small eps in the denominator.
3. Have automl error out during init if MAPE was selected as the automl objective. Similar treatment for RMSLE/MSLE if the target has negative values.

What behavior is most useful to users?

@freddyaboulton @ParthivNaresh I don't have a strong preference between 1 and 2--if you're both advocating for 1, let's do that in this PR. However to really nail that question I think we should also file 3 as an enhancement and address it separately.

freddyaboulton: I think the problem with the sklearn approach is that it will divide by eps whenever y=0 which will cause the score to blow up.

image

I don't think scores that large are useful to users, especially if they have a dataset where y=0 occurs frequently. I like option 1 because I prefer letting the user know at the beginning that this objective isn't great for problems with y=0 (rather than having them comb through or impl trying to reason why their objective is ~1e14). 

You bring up a good point about AutoMLSearch! We should definite file that. Currently we never allow RMSLE and MSLE as the primary objective but I like the idea of only disallowing it if the target has negative values. Similar for MAPE. That sounds like it could be a data check!

dsherry: @freddyaboulton agreed! And ooh yeah definitely, perhaps that could be part of the "InvalidTargetsDataCheck"

freddyaboulton: Filed #1521 !

evalml/objectives/standard_metrics.py
ParthivNaresh: @dsherry Sounds great!

evalml/objectives/standard_metrics.py
```diff
@@ -345,6 +346,22 @@ def objective_function(self, y_true, y_predicted, X=None):
...
+    Only valid for nonzero inputs. Otherwise, will throw a ValueError
+    """
+    name = "Mean Absolute Percentage Error"
+    greater_is_better = False
+    perfect_score = 0.0
+
+    def objective_function(self, y_true, y_predicted, X=None):
+        if (y_true == 0).any():
+            raise ValueError("Mean Absolute Percentage Error cannot be used when "
+                             "targets contain the value 0.")
```
dsherry, line 361: 👍  cool!

evalml/objectives/time_series_regression_objective.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class TimeSeriesRegressionObjective(RegressionObjective):
+    """Base class for all time series regression objectives.
+
+    problem_types (list(ProblemType)): List of problem types that this objective is defined for.
+        Set to [ProblemTypes.TIME_SERIES_REGRESSION]
+    """
+
+    problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
```
dsherry, line 13: @ParthivNaresh looks great thanks

evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -5,7 +5,7 @@ colorama==0.4.4
 featuretools==0.22.0
 graphviz==0.15
 ipywidgets==7.5.1
-lightgbm==3.1.0
+lightgbm==3.1.1
```
dsherry, line 8: Ah @ParthivNaresh for future reference, if the dependency update check fails on your PR, the dependency checker bot which runs every hour will get a separate PR up in with this update. And then once that's merged you can rebase/update your PR and you should be all set.

dsherry: We keep those things separate so that we can watch the dependency update step pass/fail CI, independent from any other changes like those being made in this PR. This helps us catch breaking changes made by the latest version of our dependency libs.

PR #1515, opened 2020-12-07
Add multicollinearity data check

### LINKED ISSUE #811: Data check to detect multicollinearity
Following the Data Checks API project, all pre-existing guardrails, including `detect_label_leakage` were converted to use the new Data Checks API (now `DetectLabelLeakageDataCheck`). However, that data check only works for numerical columns.

Some previous work had been done with #184 and #141 as guardrails. It would be useful to use what was started in those PRs and add data checks to address correlation / multilinearity for categorical columns.

Older relevant issues / PRs:
https://github.com/FeatureLabs/evalml/issues/26

angela97lin:
Resolves #811 by adding an initial version of a multicollinearity data check.

I switch between "correlation" and "multicollinear" because mutual information right now is technically just checking for correlation but open to thoughts about consistency 🤷

### DIFF
--- a/evalml/data_checks/id_columns_data_check.py
+++ b/evalml/data_checks/id_columns_data_check.py
@@ -32,7 +32,6 @@ class IDColumnsDataCheck(DataCheck):
 
         Arguments:
             X (ww.DataTable, pd.DataFrame, np.ndarray): The input features to check
-            threshold (float): The probability threshold to be considered an ID column. Defaults to 1.0
 
         Returns:
             dict: A dictionary of features with column name or index and their probability of being ID columns
--- a/evalml/data_checks/multicollinearity_data_check.py
+++ b/evalml/data_checks/multicollinearity_data_check.py
@@ -28,7 +28,7 @@ class MulticollinearityDataCheck(DataCheck):
 
         Returns:
             dict: dict with a DataCheckWarning if there are any potentially multicollinear columns.
-            
+
         """
         messages = {
             "warnings": [],
--- a/evalml/data_checks/multicollinearity_data_check.py
+++ b/evalml/data_checks/multicollinearity_data_check.py
@@ -1,0 +1,39 @@
+
+from evalml.data_checks import (
+    DataCheck,
+    DataCheckMessageCode,
+    DataCheckWarning
+)
+from evalml.utils.gen_utils import (
+    _convert_to_woodwork_structure,
+    _convert_woodwork_types_wrapper
+)
+
+
+class MulticollinearityDataCheck(DataCheck):
+    """Check if any set features are likely to be multicollinear."""
+
+    def __init__(self, threshold=1.0):
+        """Check if any set of features are likely to be multicollinear.
+
+        Arguments:
+            threshold (float): The threshold to be considered. Defaults to 1.0.
+        """
+        if threshold < 0 or threshold > 1:
+            raise ValueError("threshold must be a float between 0 and 1, inclusive.")
+        self.threshold = threshold
+
+    def validate(self, X, y=None):
+        """Check if any set of features are likely to be multicollinear.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, np.ndarray): The input features to check
+
+        Returns:
+            dict: A dictionary of features with column name or index and their probability of being ID columns
+        """
+        messages = {
+            "warnings": [],
+            "errors": []
+        }
+        return messages
--- a/evalml/data_checks/multicollinearity_data_check.py
+++ b/evalml/data_checks/multicollinearity_data_check.py
@@ -27,7 +27,8 @@ class MulticollinearityDataCheck(DataCheck):
             X (ww.DataTable, pd.DataFrame, np.ndarray): The input features to check
 
         Returns:
-            dict: A dictionary of features with column name or index and their probability of being ID columns
+            dict: dict with a DataCheckWarning if there are any potentially multicollinear columns.
+            
         """
         messages = {
             "warnings": [],
--- a/evalml/data_checks/multicollinearity_data_check.py
+++ b/evalml/data_checks/multicollinearity_data_check.py
@@ -37,7 +37,6 @@ class MulticollinearityDataCheck(DataCheck):
 
         X = _convert_to_woodwork_structure(X)
         mutual_info_df = X.mutual_information()
-        # zero or one case?
         if mutual_info_df.empty:
             return messages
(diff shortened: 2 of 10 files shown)


evalml/data_checks/id_columns_data_check.py
```diff
@@ -32,7 +32,6 @@ def validate(self, X, y=None):
 
         Arguments:
             X (ww.DataTable, pd.DataFrame, np.ndarray): The input features to check
-            threshold (float): The probability threshold to be considered an ID column. Defaults to 1.0
```
angela97lin, line 35: Not related but removing extraneous argument in docstring :d

evalml/tests/data_checks_tests/test_multicollinearity_data_check.py
```diff
@@ -0,0 +1,79 @@
...
+    assert multi_check.validate(X) == {
+        "warnings": [DataCheckWarning(message="Columns are likely to be correlated: [('col_1', 'col_4'), ('col_3', 'col_5'), ('col_3', 'col_6'), ('col_5', 'col_6'), ('col_1', 'col_2'), ('col_2', 'col_4')]",
+                                      data_check_name=multi_data_check_name,
+                                      message_code=DataCheckMessageCode.IS_MULTICOLLINEAR,
+                                      details={'columns': [('col_1', 'col_4'), ('col_3', 'col_5'), ('col_3', 'col_6'), ('col_5', 'col_6'), ('col_1', 'col_2'), ('col_2', 'col_4')]}).to_dict()],
+        "errors": []
+    }
+
+
+def test_multicollinearity_data_check_input_formats():
```
angela97lin, line 75: Would have liked adding ww test here but made more sense to reuse data from `test_multicollinearity_nonnumeric_cols`

evalml/data_checks/multicollinearity_data_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+        """
+        messages = {
+            "warnings": [],
+            "errors": []
+        }
+
+        X = _convert_to_woodwork_structure(X)
+        mutual_info_df = X.mutual_information()
+        # zero or one case?
```
angela97lin, line 40: oops, will address

bchen1116 (approved):
LGTM! Might be nice to add a `np.array` as a test to see whether the output columns are understandable, but not necessary.

jeremyliweishih (approved):
Current implementation looks good - just a couple questions,

1. Is this added to AutoML by default? I'm a bit worried about performance across larger datasets but maybe the WW team would know more about it.

2. Did we go with WW's implementation for convenience? I remember us looking at a VIF implementation in the past. I tried researching but couldn't find if there was a difference in using NMIF vs. just correlation as well.

jeremyliweishih:
> I switch between "correlation" and "multicollinear" because mutual information right now is technically just checking for correlation but open to thoughts about consistency 🤷

I think all user facing things should be "multicollinear" but usage of correlation or correlated in the code should be fine!

angela97lin:
@jeremyliweishih Yup, currently not adding to default data checks because of what you mentioned, but open to discussing if we should or not. Would be nice to get some sort of clustering/faster algo first before doing so.

And WW's normalized mutual information method handles categoricals, which is nice (and I believe something VIF doesn't), and is also what PS is currently using so this makes the most sense as a first version :)

tyler3991 (approved):
LGTM! Nice!!

evalml/data_checks/multicollinearity_data_check.py
```diff
@@ -0,0 +1,50 @@
...
+    DataCheckMessageCode,
+    DataCheckWarning
+)
+from evalml.utils.gen_utils import _convert_to_woodwork_structure
+
+
+class MulticollinearityDataCheck(DataCheck):
+    """Check if any set features are likely to be multicollinear."""
+
+    def __init__(self, threshold=1.0):
```
freddyaboulton, line 13: Maybe we should change the default threshold? I think [ww uses](https://github.com/alteryx/woodwork/blob/main/woodwork/datatable.py#L1000) sklearns [normalized_mutal_information](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.normalized_mutual_info_score.html) which is scaled between (0, 1) so this only issues a warning if columns are perfectly correlated.

freddyaboulton (approved):
@angela97lin Looks good! Maybe we should benchmark this to know where we are now in terms of performance. Obviously, that shouldn't block this merge :)

evalml/data_checks/multicollinearity_data_check.py
dsherry: Agreed! Let's lower this a bit. What was the PS team doing?

evalml/data_checks/multicollinearity_data_check.py
```diff
@@ -0,0 +1,50 @@
...
+            "warnings": [],
+            "errors": []
+        }
+
+        X = _convert_to_woodwork_structure(X)
+        mutual_info_df = X.mutual_information()
+        if mutual_info_df.empty:
+            return messages
+        above_threshold = mutual_info_df.loc[mutual_info_df['mutual_info'] >= self.threshold]
+        correlated_cols = [(col_1, col_2) for col_1, col_2 in zip(above_threshold['column_1'], above_threshold['column_2'])]
```
dsherry, line 43: Will this remove duplicate pairs, i.e. `('col_A', 'col_B')` and `('col_B', 'col_A')` ? Ideally we should find a way to do that.

dsherry (approved):
@angela97lin LGTM! Had a couple comments. We should definitely lower the default threshold.

evalml/data_checks/multicollinearity_data_check.py
angela97lin: Luckily, `mutual_information` from Woodwork handles this; that's why we only see one pair in the expected test results 😁

evalml/data_checks/multicollinearity_data_check.py
angela97lin: @dsherry @freddyaboulton I'll poke around or ask, but my understanding was that they just display the matrix, and don't use a threshold to determine whether things are correlated. I'll lower the threshold to 0.9--kind of arbitrary, but hard to determine without more testing 🤔

dsherry: @angela97lin ah yeah that makes sense. Sure, seems fine to me.

PR #1524, opened 2020-12-08
Codegen support for bool

### LINKED ISSUE #1495: Code generation can't handle booleans in the parameters dict
Repro:

```python
from evalml.pipelines.utils import generate_pipeline_code
from evalml.pipelines import MulticlassClassificationPipeline
from evalml.pipelines.components import Transformer

    name = "I have a bool parameter"

    def __init__(self, do_something=False, random_state=0):
        super().__init__(parameters={'do_something': do_something},
                         component_obj=None,
                         random_state=random_state)

        
class CustomPipeline(MulticlassClassificationPipeline):
    name = "Custom Pipeline"
    component_graph = [MyTransformer,'Random Forest Classifier']

pipeline_instance = CustomPipeline(parameters={})
code = generate_pipeline_code(pipeline_instance)
print(code)


exec(code)
```

```python
from evalml.pipelines.multiclass_classification_pipeline import MulticlassClassificationPipeline

class CustomPipeline(MulticlassClassificationPipeline):
	component_graph = [
		MyTransformer,
		'Random Forest Classifier'
	]
	name = 'Custom Pipeline'

parameters = {
	"I have a bool parameter": {
		"do_something": false
	},
	"Random Forest Classifier": {
		"n_estimators": 100,
		"max_depth": 6,
		"n_jobs": -1
	}
}
pipeline = CustomPipeline(parameters)
```

```
NameError                                 Traceback (most recent call last)
<ipython-input-5-0e2afc7a82c0> in <module>
     21 
     22 
---> 23 exec(code)

<string> in <module>

NameError: name 'false' is not defined
```

bchen1116:
fix #1495 

Add support for pythonic types that JSON doesn't normally support, like boolean and None values. 

I didn't add in support/tests for Python objects (like Datetime) or tuples (which get converted to lists by `json.dumps` and remains a list after `json.loads`) since these seem to be edge cases and are unlikely to be params in pipelines/pipeline hyperparameters. Certainly down to discuss this if it is otherwise important to have support for these/other cases!

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -172,7 +172,7 @@ def _json_dict(dictionary):
         else:
             try:
                 json.dumps(v)
-            except:
+            except TypeError:
                 raise TypeError(f"Value {v} of type {type(v)} cannot be JSON-serialized")
 
 
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -174,7 +174,7 @@ def generate_pipeline_code(element):
         Does not include code for custom component implementation.
     """
     # hold the imports needed and add code to end
-    code_strings = []
+    code_strings = ['import json']
     if not isinstance(element, PipelineBase):
         raise ValueError("Element must be a pipeline instance, received {}".format(type(element)))
 
@@ -192,13 +192,13 @@ def generate_pipeline_code(element):
     base_string = "\nclass {0}({1}):\n" \
                   "\tcomponent_graph = [\n\t\t{2}\n\t]\n" \
                   "{3}" \
-                  "\nparameters = {4}\n" \
+                  "\nparameters = json.loads(\"\"\"{4}\"\"\")\n" \
                   "pipeline = {0}(parameters)" \
                   .format(element.__class__.__name__,
                           element.__class__.__bases__[0].__name__,
                           component_graph_string,
                           pipeline_string,
-                          json.dumps(element.parameters, indent='\t').replace('null', 'None'))
+                          json.dumps(element.parameters, indent='\t'))
     code_strings.append(base_string)
     return "\n".join(code_strings)
 
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -163,6 +163,19 @@ def make_pipeline_from_components(component_instances, problem_type, custom_name
     return TemplatedPipeline({c.name: c.parameters for c in component_instances}, random_state=random_state)
 
 
+def _json_dict(dictionary):
+    """Helper method to make sure all elements in the provided dictionary are JSON-serializable
+    """
+    for v in dictionary.values():
+        if isinstance(v, dict):
+            _json_dict(v)
+        else:
+            try:
+                json.dumps(v)
+            except:
+                raise TypeError(f"Value {v} of type {type(v)} cannot be JSON-serialized")
+
+
 def generate_pipeline_code(element):
     """Creates and returns a string that contains the Python imports and code required for running the EvalML pipeline.
 
@@ -188,6 +201,7 @@ def generate_pipeline_code(element):
         pipeline_list += ["{} = '{}'".format(k, v)] if isinstance(v, str) else ["{} = {}".format(k, v)]
 
     pipeline_string = "\t" + "\n\t".join(pipeline_list) + "\n" if len(pipeline_list) else ""
+    _json_dict(element.parameters)
     # create the base string for the pipeline
     base_string = "\nclass {0}({1}):\n" \
                   "\tcomponent_graph = [\n\t\t{2}\n\t]\n" \
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,7 +9,7 @@ Release Notes
     * Fixes
         * Fix Windows CI jobs: install ``numba`` via conda, required for ``shap`` :pr:`1490`
         * Added custom-index support for `reset-index-get_prediction_vs_actual_over_time_data` :pr:`1494`
-        * Fix ``generate_pipeline_code`` to account for boolean and None differences between Python and Json :pr:``
+        * Fix ``generate_pipeline_code`` to account for boolean and None differences between Python and JSON :pr:`1524`
     * Changes
         * Update circleci badge to apply to ``main`` :pr:`1489`
         * Added script to generate github markdown for releases :pr:`1487`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,6 +9,7 @@ Release Notes
     * Fixes
         * Fix Windows CI jobs: install ``numba`` via conda, required for ``shap`` :pr:`1490`
         * Added custom-index support for `reset-index-get_prediction_vs_actual_over_time_data` :pr:`1494`
+        * Fix ``generate_pipeline_code`` to account for boolean and None differences between Python and Json :pr:``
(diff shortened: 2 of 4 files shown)


evalml/pipelines/utils.py
```diff
@@ -192,13 +192,13 @@ def generate_pipeline_code(element):
     base_string = "\nclass {0}({1}):\n" \
                   "\tcomponent_graph = [\n\t\t{2}\n\t]\n" \
                   "{3}" \
-                  "\nparameters = {4}\n" \
+                  "\nparameters = json.loads(\"\"\"{4}\"\"\")\n" \
```
bchen1116, line 195: Adding in `json.loads` to convert the JSON back to Python. This handles both JSON null to Python None and JSON "false"/"true" to Pythonic booleans.

evalml/pipelines/utils.py
```diff
@@ -192,13 +192,13 @@ def generate_pipeline_code(element):
...
                   "pipeline = {0}(parameters)" \
                   .format(element.__class__.__name__,
                           element.__class__.__bases__[0].__name__,
                           component_graph_string,
                           pipeline_string,
-                          json.dumps(element.parameters, indent='\t').replace('null', 'None'))
+                          json.dumps(element.parameters, indent='\t'))
```
freddyaboulton, line 201: not-blocking as it doesn't relate to this fix but it does relate to your question about which types we should handle. I think we should at least raise an actionable user-friendly exception if for whatever reason the parameters are not json serializable. Right now, the user gets a messy stack trace. 

For example, I used a numpy array to specify the `features_to_extract` parameter  instead of a list in the code gen demo. This is a corner-case for sure but the date time featurizer will work if you use a numpy array instead of a list.

```python
pipeline_instance = CustomPipeline(parameters={"Imputer": {"numeric_impute_strategy": "median"},
                                               "DateTime Featurization Component": {"features_to_extract": np.array(["year"])}})
code = generate_pipeline_code(pipeline_instance)
exec(code)
```

Stack trace:
```
~/sources/evalml/evalml/pipelines/utils.py in generate_pipeline_code(element)
    199                           component_graph_string,
    200                           pipeline_string,
--> 201                           json.dumps(element.parameters, indent='\t'))
    202     code_strings.append(base_string)
    203     return "\n".join(code_strings)

~/miniconda3/envs/evalml/lib/python3.8/json/__init__.py in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    232     if cls is None:
    233         cls = JSONEncoder
--> 234     return cls(
    235         skipkeys=skipkeys, ensure_ascii=ensure_ascii,
    236         check_circular=check_circular, allow_nan=allow_nan, indent=indent,

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in encode(self, o)
    199         chunks = self.iterencode(o, _one_shot=True)
    200         if not isinstance(chunks, (list, tuple)):
--> 201             chunks = list(chunks)
    202         return ''.join(chunks)
    203 

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in _iterencode(o, _current_indent_level)
    429             yield from _iterencode_list(o, _current_indent_level)
    430         elif isinstance(o, dict):
--> 431             yield from _iterencode_dict(o, _current_indent_level)
    432         else:
    433             if markers is not None:

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in _iterencode_dict(dct, _current_indent_level)
    403                 else:
    404                     chunks = _iterencode(value, _current_indent_level)
--> 405                 yield from chunks
    406         if newline_indent is not None:
    407             _current_indent_level -= 1

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in _iterencode_dict(dct, _current_indent_level)
    403                 else:
    404                     chunks = _iterencode(value, _current_indent_level)
--> 405                 yield from chunks
    406         if newline_indent is not None:
    407             _current_indent_level -= 1

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in _iterencode(o, _current_indent_level)
    436                     raise ValueError("Circular reference detected")
    437                 markers[markerid] = o
--> 438             o = _default(o)
    439             yield from _iterencode(o, _current_indent_level)
    440             if markers is not None:

~/miniconda3/envs/evalml/lib/python3.8/json/encoder.py in default(self, o)
    177 
    178         """
--> 179         raise TypeError(f'Object of type {o.__class__.__name__} '
    180                         f'is not JSON serializable')
    181 

TypeError: Object of type ndarray is not JSON serializable
```

evalml/pipelines/utils.py
freddyaboulton: Cool! I think all of our current components default parameters are json serializable but maybe we should encode that in a test? Just to make sure future components have this convention.

freddyaboulton (approved):
@bchen1116 I think this is great! Thanks for making the fix. I left a couple of non-blocking comments about testing and handling non-json-serializable types.

evalml/pipelines/utils.py
bchen1116: Great point! I'll add some code to make this more user-friendly

evalml/pipelines/utils.py
bchen1116: @freddyaboulton Do you mean as a test for all components?

bchen1116: Filed a new [issue here](https://github.com/alteryx/evalml/issues/1526) to track adding these tests

evalml/pipelines/utils.py
```diff
@@ -174,7 +187,7 @@ def generate_pipeline_code(element):
         Does not include code for custom component implementation.
     """
     # hold the imports needed and add code to end
-    code_strings = []
+    code_strings = ['import json']
```
jeremyliweishih, line 190: nice foresight 👍

jeremyliweishih (approved):
Looks good !

evalml/pipelines/utils.py
```diff
@@ -188,17 +201,18 @@ def generate_pipeline_code(element):
         pipeline_list += ["{} = '{}'".format(k, v)] if isinstance(v, str) else ["{} = {}".format(k, v)]
 
     pipeline_string = "\t" + "\n\t".join(pipeline_list) + "\n" if len(pipeline_list) else ""
+    _json_dict(element.parameters)
```
dsherry, line 204: @bchen1116 could we cut out the middleman by deleting `_json_dict` and doing the following in `generate_pipeline_code`:
```python
try:
    ret = json.dumps(element.parameters, indent='\t')
except TypeError:
    raise TypeError(f"Value {v} of type {type(v)} cannot be JSON-serialized")
```

evalml/pipelines/utils.py
```diff
@@ -163,6 +163,19 @@ class TemplatedPipeline(_get_pipeline_base_class(problem_type)):
...
+    """Helper method to make sure all elements in the provided dictionary are JSON-serializable
+    """
+    for v in dictionary.values():
+        if isinstance(v, dict):
+            _json_dict(v)
+        else:
+            try:
+                json.dumps(v)
+            except TypeError:
+                raise TypeError(f"Value {v} of type {type(v)} cannot be JSON-serialized")
```
dsherry, line 176: @bchen1116 for what types does this error occur? Should we add test coverage of those cases?

bchen1116: It errors out for lots of python-specific types. I can add more coverage of that in the tests

PR #1527, opened 2020-12-08
Provide more information to users about ensembling behavior

### LINKED ISSUE #1461: Provide more information to users about ensembling behavior
@freddyaboulton made a really great suggestion about making the ensembling behavior more visible and usable to users in https://github.com/alteryx/evalml/pull/1425#pullrequestreview-529200735 :

@angela97lin This looks good to me! This is out of scope of this PR but this sentence might be confusing to users because I don't think we ever explain how the batching in AutoML works:

> The stacking ensemble pipeline runs in its own batch after a whole cycle of training has occurred (each allowed pipeline trains for one batch). Note that this means a large number of iterations may need to run before the stacking ensemble runs.

One thing we can do is add something to the AutoML printout like this if the number of iterations selected by the user is too small:

```
Ensembling set to true but the number of iterations is too small
for ensembling to run. Set it to at least ...
```

and 

```
Ensembling is set to True and max_iterations is set to ... 
so ensembling will run in batches 4, 8, 12 
```

That way users are not so confused about whether or not ensembling will run. 

This would be really helpful for users who want to use ensembling but are confused as to why they do not see an ensembling model trained.

angela97lin:
Closes #1461

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -453,6 +453,16 @@ class AutoMLSearch:
         if run_ensembling and len(self.allowed_pipelines) == 1:
             logger.warning("Ensembling was set to True, but the number of unique pipelines is one, so ensembling will not run.")
             run_ensembling = False
+        if run_ensembling and self.max_iterations is not None:
+            # Baseline + first batch + each pipeline iteration (5 is current default pipelines_per_batch) + 1
+            if self.max_iterations < (1 + self.allowed_pipelines + self.allowed_pipelines * self._pipelines_per_batch + 1):
+                run_ensembling = False
+                logger.warning("Ensembling was set to True, but not enough max iterations before first ensemble runs, so ensembling will not run.")
+        if run_ensembling and self.max_batches and self.max_iterations is None:
+            # Baseline + first batch + each pipeline iteration (5 is current default pipelines_per_batch) + 1
+            if self.max_batches < len(self.allowed_pipelines) + 1:
+                run_ensembling = False
+                logger.warning("Ensembling was set to True, but not enough max batches before first ensemble runs, so ensembling will not run.")
 
         if self.max_batches and self.max_iterations is None:
             self.show_batch_output = True
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -459,6 +459,11 @@ class AutoMLSearch:
             if run_ensembling:
                 ensemble_nth_batch = len(self.allowed_pipelines) + 1
                 num_ensemble_batches = (self.max_batches - 1) // ensemble_nth_batch
+                if num_ensemble_batches == 0:
+                    logger.warning("Not enough batches to run ensembling")
+                else:
+                    logger.info(f"Ensemble run every {ensemble_nth_batch} batches")
+                    
                 self.max_iterations = (1 + len(self.allowed_pipelines) +
                                        self._pipelines_per_batch * (self.max_batches - 1 - num_ensemble_batches) +
                                        num_ensemble_batches)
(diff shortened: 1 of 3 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -1311,6 +1311,8 @@ def test_max_iteration_works_with_stacked_ensemble(mock_pipeline_fit, mock_score
         assert not pipeline_names.str.contains('Ensemble').any()
     elif use_ensembling:
         assert pipeline_names.str.contains('Ensemble').any()
+        ensemble_nth_batch = len(automl.allowed_pipelines) + 1
+        assert f"Ensembling will run every {ensemble_nth_batch} batches."
```
freddyaboulton, line 1315: I think this is a typo since it always evaluates to True?

Currently, we don't output a message of when ensembling will run if the user specifies `max_iterations` instead of `max_batches`. Maybe we can say "Ensembling will run at the kth iteration and every j iterations after that"?

freddyaboulton (approved):
@angela97lin Looks great! I have one minor comment about letting the user know when ensembling will run if they specify a large enough number of iterations instead of batches. I think we only tell them anything if the number of iterations is too small.

I think we should also add coverage for the "Ensembling will run every k batches." message. Maybe in `test_max_batches_works` ?

evalml/automl/automl_search.py
```diff
@@ -452,14 +452,26 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
 
         run_ensembling = self.ensembling
         if run_ensembling and len(self.allowed_pipelines) == 1:
-            logger.warning("Ensembling was set to True, but the number of unique pipelines is one, so ensembling will not run.")
+            logger.warning("Ensembling is set to True, but the number of unique pipelines is one, so ensembling will not run.")
             run_ensembling = False
 
+        if run_ensembling and self.max_iterations is not None:
+            # Baseline + first batch + each pipeline iteration + 1
+            first_ensembling_iteration = (1 + len(self.allowed_pipelines) + len(self.allowed_pipelines) * self._pipelines_per_batch + 1)
```
bchen1116, line 460: nit-pick: Can this simplify to `(2 + len(self.allowed_pipelines) + len(self.allowed_pipelines) * self._pipelines_per_batch)`?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1380,15 +1382,41 @@ def test_automl_one_allowed_pipeline_ensembling_disabled(mock_pipeline_fit, mock
...
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_max_batches_less_than_ensembling_disabled(mock_pipeline_fit, mock_score, X_y_binary, caplog):
+    X, y = X_y_binary
+    automl = AutoMLSearch(problem_type="binary", max_batches=2, allowed_model_families=[ModelFamily.LINEAR_MODEL], ensembling=True)
+    automl.search(X, y, data_checks=None)
+    first_ensemble_batch = 1 + len(automl.allowed_pipelines) + 1  # First batch + each pipeline batch
+    assert f"Ensembling is set to True, but max_batches is too small, so ensembling will not run. Set max_batches >= {first_ensemble_batch} to run ensembling." in caplog.text
+
+    pipeline_names = automl.rankings['pipeline_name']
+    assert not pipeline_names.str.contains('Ensemble').any()
```
bchen1116, line 1419: Nice!

bchen1116 (approved):
Looks good to me. I like the test @freddyaboulton pointed out. Left one small suggestion, but nothing blocking.

evalml/automl/automl_search.py
angela97lin: Haha I kept the two 1's separate so it'd be more clear where the math is coming from! But lmk if it doesn't help #_#

evalml/tests/automl_tests/test_automl.py
angela97lin: thanks for catching this hehe, updated!

Issue #1517, opened 2020-12-07
Allow X to be None in Time Series Pipelines


jeremyliweishih:
In previous EvalML pipelines we have always made the assumption that `X` cannot be `None`. However, with the addition of the `TimeSeriesBaselineRegressor` and plans to add more estimators that rely only on `y` (such as ARIMA, PROPHET) we can revise this assumption for TimeSeriesRegressionPipelines. 

This issue keeps track of:

1. Removing the `ValueError` from `TimeSeriesRegressionPipeline.fit` and accommodating for an empty `X`
2. Removing the overridden `fit` method in `TimeSeriesBaseLineRegressionPipeline` as that was put in as a patch to solve this problem

freddyaboulton:
Thanks for filing @jeremyliweishih ! Yea I think we should remove the exception from `TimeSeriesRegressionPipeline.fit`! For context, here is the discussion where we decided to add it in the first place. 

https://github.com/alteryx/evalml/pull/1418#discussion_r522397434

PR #1528, opened 2020-12-08
Classification Pipelines for Time Series

### LINKED ISSUE #1474: Binary and Multiclass pipelines for time series

freddyaboulton:
### Pull Request Description
Fixes #1474

I'll add to AutoML in a separate PR since this is a big diff.

Most of the changes come from adding time series binary and time series multiclass problems types to the list of supported problem types for each existing classifier. I would ignore the diff in the estimators and estimator tests and focus on the diff in `pipeline_base_meta.py`, `time_series_classification_pipelines.py` and `test_time_series_pipeline.py`

-----


evalml/pipelines/pipeline_base_meta.py
```diff
@@ -24,5 +25,9 @@ def _check_for_fit(self, X=None, y=None):
             elif y is None:
                 return method(self, X)
             else:
+                # For time series classification pipelines, predict will take X, y, objective
```
freddyaboulton, line 28: The problem is that `TimeSeriesBinaryClassification.predict` accepts `X`, `y` and `objective`. All other pipeline classes accept `X` and `y` (`TimeSeriesRegressionPipeline`) or `X` and `objective` (`BinaryClassificationPipeline`, `MulticlassClassificationPipeline`) or `X` (`ClassificationPipeline`).

I think adding this extra check for the number of arguments in `predict` is the lowest-touch way of handling the complexity (as opposed to adding another metaclass) especially since this class is not user-facing and we don't expect users to define their own pipeline base classes. 

That being said, maybe we can refactor this metaclass definition to make it a little clearer what's going on?

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+            y (ww.DataColumn, pd.Series, np.ndarray): The target training targets of length [n_samples]
+
+        Returns:
+            self
+        """
+        X, y = self._convert_to_woodwork(X, y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+        self._encoder.fit(y)
+        y = self._encode_targets(y)
```
freddyaboulton, line 61: Throughout, we need to encode targets before computing features because we allow the user to pass in string-valued targets. If there is a `DelayedFeatureTransformer` in the pipeline, these string values will get added to the feature matrix and if there isn't a `OneHotEncoder`, the estimator will crash. 

Long story short, this is a repeat of #1477 . I think it would be better if we pass this complexity to the `DelayedFeatureTransformer` and have it encode the target before delaying but we should not let that block this PR. Happy to file something for updating the `DelayedFeatureTransformer`!

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+                               "that either 1) you passed in X=None to fit and don't have a DelayFeatureTransformer "
+                               "in your pipeline or 2) you do have a DelayFeatureTransformer but gap=0 and max_delay=0. "
+                               "Please add a DelayFeatureTransformer or change the values of gap and max_delay")
+
+        y_shifted = y.shift(-self.gap)
+        X_t, y_shifted = drop_rows_with_nans(X_t, y_shifted)
+        self.estimator.fit(X_t, y_shifted)
+        return self
+
+    def _predict(self, X, y, objective=None, pad=False):
```
freddyaboulton, line 75: I have added a `pad` parameter to `_predict` because in `predict` we want to return un-padded predictions or else `self._decode_targets` will error out. However in `_compute_predictions` (called by `score`) we want to return padded predictions (so that predictions, predicted probabilities, and the shifted target all have the same length. This makes dropping nans much easier).

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+        X, y = self._convert_to_woodwork(X, y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+        objectives = [get_objective(o, return_instance=True) for o in objectives]
+
+        y_encoded = self._encode_targets(y)
+        y_shifted = y_encoded.shift(-self.gap)
+        y_pred, y_pred_proba = self._compute_predictions(X, y, objectives)
+        non_nan_mask = _get_rows_without_nans(y_shifted, y_pred, y_pred_proba)
+        if y_pred is not None:
```
freddyaboulton, line 150: This is a bit ugly but depending on the objectives that are passed in, `y_pred` or `y_pred_proba` can be `None` so we can only drop nans if possible. 

I've thought about avoiding this complexity by having `_compute_predictions` return un-padded predictions so that we don't have to worry about dropping anything. However, that doesn't really work because the predictions and `y_shifted` will have different number of nans depending on the values of `gap` and the number of delayed features computed in the pipeline. The least complex way of knowing what the common non-nan subset is by looking at the nans in the predictions and targets.

In short, we may be able to hide this complexity in a helper function but I don't think we can avoid it.

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+        y_encoded = self._encode_targets(y)
+        features = self.compute_estimator_features(X, y_encoded)
+        features_no_nan = features.dropna(axis=0, how="any")
+
+        if objective is not None:
+            objective = get_objective(objective, return_instance=True)
+            if not objective.is_defined_for_problem_type(self.problem_type):
+                raise ValueError(f"Objective {objective.name} is not defined for time series binary classification.")
+
+        if self.threshold is None:
```
freddyaboulton, line 182: This looks a bit different from `BinaryClassificationPipeline` so that I only have to call `pad_with_nans` once.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -191,3 +228,159 @@ class Pipeline(pipeline_class):
...
+        # NaNs are expected because of padding due to max_delay
+        assert set(pl.predict(X, y).dropna().unique()) == expected_unique_values
+    pl.score(X, y, objectives)
+
+
+@patch('evalml.pipelines.TimeSeriesClassificationPipeline._decode_targets')
+@patch('evalml.objectives.BinaryClassificationObjective.decision_function')
+@patch('evalml.pipelines.components.Estimator.predict_proba', return_value=pd.DataFrame({0: [1.]}))
+@patch('evalml.pipelines.components.Estimator.predict', return_value=pd.Series([1.]))
+def test_binary_classification_predictions_thresholded_properly(mock_predict, mock_predict_proba,
```
freddyaboulton, line 326: This is to achieve 100% coverage on the binary classification `_predict` method.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -191,3 +228,159 @@ class Pipeline(pipeline_class):
...
+@pytest.mark.parametrize("pipeline_class,objectives", [(TimeSeriesBinaryClassificationPipeline, ["MCC Binary"]),
+                                                       (TimeSeriesBinaryClassificationPipeline, ["Log Loss Binary"]),
+                                                       (TimeSeriesBinaryClassificationPipeline, ["MCC Binary", "Log Loss Binary"]),
+                                                       (TimeSeriesMulticlassClassificationPipeline, ["MCC Multiclass"]),
+                                                       (TimeSeriesMulticlassClassificationPipeline, ["Log Loss Multiclass"]),
+                                                       (TimeSeriesMulticlassClassificationPipeline, ["MCC Multiclass", "Log Loss Multiclass"]),
+                                                       (TimeSeriesRegressionPipeline, ['R2']),
+                                                       (TimeSeriesRegressionPipeline, ['R2', "Mean Absolute Percentage Error"])])
+@pytest.mark.parametrize("use_ww", [True, False])
+def test_score_works(pipeline_class, objectives, use_ww, X_y_binary, X_y_multi, X_y_regression):
```
freddyaboulton, line 285: I think it's good to have a unit test where we pass data through a minimal time series pipeline and make sure everything works. I'm setting n_jobs=1 here and the test takes 2.5 seconds to run so I think it's ok.

evalml/pipelines/pipeline_base_meta.py
jeremyliweishih: I think checking # of arguments is a good idea. Maybe we can just refactor it into the usual if else logic? 

```
             if not self._is_fitted:
                raise PipelineNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
            elif X is None and y is None:
                return method(self)
            elif y is None:
                return method(self, X)
            # For time series classification pipelines, predict will take X, y, objective
            elif len(inspect.getfullargspec(method).args) == 4:
                 return method(self, X, y, objective)
            else:
                # For other pipelines, predict will take X, y or X, objective
                return method(self, X, y)

File: evalml/pipelines/time_series_classifcation_pipelines.py [hunk 8]
```diff
@@ -0,0 +1,204 @@
...
+    drop_rows_with_nans,
+    pad_with_nans
+)
+
+
+class TimeSeriesClassificationPipeline(ClassificationPipeline):
+    """Pipeline base class for time series classifcation problems."""
+
+    def __init__(self, parameters, random_state=0):
+        """Machine learning pipeline for time series regression problems made out of transformers and a classifier.
```
jeremyliweishih, line 19: `classification` instead of `regression`

evalml/pipelines/time_series_classifcation_pipelines.py
jeremyliweishih: Agree with you here! I like the idea of filing an issue and updating the `DelayedFeatureTransformer` as well as removing  encoding in the pipelines.

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        X, y = self._convert_to_woodwork(X, y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+        self._encoder.fit(y)
+        y = self._encode_targets(y)
+
+        X_t = self._compute_features_during_fit(X, y)
+        if X_t.empty:
+            raise RuntimeError("Pipeline computed empty features during call to .fit. This means "
```
jeremyliweishih, line 65: @freddyaboulton do we want behavior here to be the same as the current regression pipelines or should it be like after the fixes for #1517? I think this is fine for now but if we want to change behavior in the future we should add that info to #1517.

jeremyliweishih (approved):
LGTM @freddyaboulton! Just left some questions and comments.

evalml/pipelines/time_series_classifcation_pipelines.py
freddyaboulton: Good catch!

evalml/pipelines/time_series_classifcation_pipelines.py
freddyaboulton: Good point! I think we should get this merged and then fix #1517 for all ts pipelines at once. I don't want the pipelines to have differing behavior if we make the changes needed for #1517 here and not simultaneously for regression.

evalml/pipelines/pipeline_base_meta.py
freddyaboulton: Done!

evalml/utils/gen_utils.py
```diff
@@ -353,11 +354,39 @@ def pad_with_nans(pd_data, num_to_pad):
...
+        if pd_data is None:
+            return np.array([True])
+        if isinstance(pd_data, pd.Series):
+            return ~pd_data.isna().values
+        elif isinstance(pd_data, pd.DataFrame):
+            return ~pd_data.isna().any(axis=1).values
+        else:
+            return pd_data
+
+    mask = reduce(lambda a, b: np.logical_and(_not_nan(a), _not_nan(b)), data)
```
bchen1116, line 388: This is cool! So this means that if the second Dataframe/Series in the sequence had a nan value in the last column, then the last value in the mask will be False, like below? 
image

Might be nice to update the return to `mask where each entry is True if and only if all corresponding entries in that index in data are non-nan` or something of the sort. I wasn't sure what this was supposed to do

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -126,6 +145,14 @@ class Pipeline(pipeline_class):
...
+    def mock_predict(df):
+        return pd.Series(range(200, 200 + df.shape[0]))
+
+    if isinstance(pl, TimeSeriesRegressionPipeline):
+        mock_regressor_predict.side_effect = mock_predict
+    else:
+        mock_classifier_predict.side_effect = mock_predict
+
     if only_use_y:
         pl.fit(None, y)
```
bchen1116, line 157: How does this work? Looking at the code for `fit()`, it seems like X is necessary to pass in, since you do call `estimator.fit()` on the computed features of X?

bchen1116 (approved):
Great work, Freddy! The logic for when to keep track of adding padding versus not is definitely confusing, but everything looks good to me! I left a couple questions and a comment, but solid implementation and test coverage!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
freddyaboulton: When `X` is `None` it's assumed that the pipeline will create some features from the `y` before passing to the estimator (typically `DelayedFeatureTransformer`). We're getting rid of this assumption in #1517, however, because some (future) estimators will naturally support the `X=None` case.

evalml/utils/gen_utils.py
freddyaboulton: Yes exactly. Great suggestion on the docstring!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
bchen1116: Gotcha, that makes sense. Thanks!

dsherry (commented):
@freddyaboulton I left a few comments, got halfway through the impl so far :)

evalml/pipelines/pipeline_base_meta.py
```diff
@@ -15,14 +16,18 @@ def check_for_fit(cls, method):
...
+        def _check_for_fit(self, X=None, y=None, objective=None):
             klass = type(self).__name__
             if not self._is_fitted:
                 raise PipelineNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
             elif X is None and y is None:
                 return method(self)
             elif y is None:
                 return method(self, X)
+            # For time series classification pipelines, predict will take X, y, objective
+            elif len(inspect.getfullargspec(method).args) == 4:
```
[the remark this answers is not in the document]
freddyaboulton: Positional arguments! So it includes any arguments that don't have `*` or `**` in front of their name.

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+            y (ww.DataColumn, pd.Series, np.ndarray, None): The target training targets of length [n_samples]
+            objective (Object or string): The objective to use to make predictions
+
+        Returns:
+            pd.Series: Predicted values.
+        """
+        X, y = self._convert_to_woodwork(X, y)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = _convert_woodwork_types_wrapper(y.to_series())
+        n_features = max(len(y), X.shape[0])
```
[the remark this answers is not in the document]
freddyaboulton: Yes!

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
+import pandas as pd
```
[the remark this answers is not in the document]
freddyaboulton: Great catch! 😬

evalml/pipelines/classification_pipeline.py
```diff
@@ -50,7 +50,7 @@ def fit(self, X, y):
     def _encode_targets(self, y):
         """Converts target values from their original values to integer values that can be processed."""
         try:
-            return pd.Series(self._encoder.transform(y))
+            return pd.Series(self._encoder.transform(y), index=y.index)
```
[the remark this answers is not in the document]
freddyaboulton: I know LightGBM and Catboost also use encoders internally. I'll make a note to see if we need to do something similar there and if so file an issue!

Great thinking!

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+        """Machine learning pipeline for time series classification problems made out of transformers and a classifier.
+
+        Required Class Variables:
+            component_graph (list): List of components in order. Accepts strings or ComponentBase subclasses in the list
+
+        Arguments:
+            parameters (dict): Dictionary with component names as keys and dictionary of that component's parameters as values.
+                 An empty dictionary {} implies using all default values for component parameters. Pipeline-level
+                 parameters such as gap and max_delay must be specified with the "pipeline" key. For example:
+                 Pipeline(parameters={"pipeline": {"max_delay": 4, "gap": 2}}).
```
angela97lin, line 28: Great example, nice and clear :)

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+            random_state (int, np.random.RandomState): The random seed/state. Defaults to 0.
+        """
+        if "pipeline" not in parameters:
```
angela97lin, line 31: Really nit-picky, but what if a user passes in a "pipeline" key mapping to a dictionary with "gap" or "max_delay" missing? Ex: {pipeline: {"gap":1}} I think we get a KeyError in the following lines, right?

evalml/pipelines/time_series_classifcation_pipelines.py
```diff
@@ -0,0 +1,204 @@
...
+        self.max_delay = pipeline_params['max_delay']
+        super().__init__(parameters, random_state)
+
+    @staticmethod
+    def _convert_to_woodwork(X, y):
+        if X is None:
+            X = pd.DataFrame()
+        X = _convert_to_woodwork_structure(X)
+        y = _convert_to_woodwork_structure(y)
+        return X, y
```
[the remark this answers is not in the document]
angela97lin: Hm... could be, but the conversion to an empty DataFrame if X is None is a time series specific thing, no? In other cases, we'd probably want to raise something if the user passes None as X?

evalml/pipelines/time_series_classifcation_pipelines.py
freddyaboulton: When the dust settles and we build more features, I want to see if there is a way to refactor these classes to avoid duplication between classification and time series classification pipelines. 

It's not apparent right now because the "time-series-specific" code gets sprinkled in between the "non-time-series" code so it's not easy to just isolate the time series code into a helper function/mix-in yet. But maybe a new design will be clearer when this one becomes more stable and we test it!

evalml/pipelines/pipeline_base_meta.py
```diff
@@ -15,14 +16,18 @@ def check_for_fit(cls, method):
...
+                return method(self, X, y, objective)
+            # For other pipelines, predict will take X, y or X, objective
```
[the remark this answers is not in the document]
freddyaboulton: Yea it's correct but confusing 😅 .

For non-timeseries classification pipelines, the arguments will be `X` and `objective` so the `y` in this `else` is actually the `objective` the user passed in!

I left this (comment)[https://github.com/alteryx/evalml/pull/1528#discussion_r540318731] earlier  but I guess it got overwritten when I pushed new changes:

> The problem is that TimeSeriesBinaryClassification.predict accepts X, y and objective. All other pipeline classes accept X and y (TimeSeriesRegressionPipeline) or X and objective (BinaryClassificationPipeline, MulticlassClassificationPipeline) or X (ClassificationPipeline).
> 
> I think adding this extra check for the number of arguments in predict is the lowest-touch way of handling the complexity (as opposed to adding another metaclass) especially since this class is not user-facing and we don't expect users to define their own pipeline base classes.
> 
> That being said, maybe we can refactor this metaclass definition to make it a little clearer what's going on?

evalml/pipelines/classification_pipeline.py
angela97lin, line 53: Curious, did this break / how did you catch this? If so, could be good to add a test for it if there isn't one already 😬

angela97lin (approved):
Left a few nit-picky comments but otherwise looks good!

evalml/pipelines/time_series_classifcation_pipelines.py
freddyaboulton: Yep you're right!

evalml/pipelines/classification_pipeline.py
freddyaboulton: The time series tests caught this! We pass a `y` with a custom objective there.

evalml/pipelines/time_series_classifcation_pipelines.py
freddyaboulton: We'll have to fix the same thing in regression pipelines so I'll just file an issue and change it for both after this is merged!

PR #1538, opened 2020-12-09
Fix XGBoost dependency

### LINKED ISSUE #1536: Unpin XGBoost Version upon XGBoost package update
Now that the XGBoost versioning issue has [been merged](https://github.com/dmlc/xgboost/pull/6482), they will be updating the pip package to version `1.3.0.post0`. This issue tracks removing our XGBoost version limitations once the new package is released.

bchen1116:
fix #1536 

New package has been published to pypi, so we are removing the version limitations

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
         environment:
-          OMP_NUM_THREADS: 16
+          OMP_NUM_THREADS: 64
 
 commands:
   install_dependencies:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
         environment:
-          OMP_NUM_THREADS: 8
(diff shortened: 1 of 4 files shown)


dsherry:
@bchen1116 I have a few ideas for how to debug that docs timeout:
1. Right away my hunch is that xgboost is taking a lot longer to run in the new version. To confirm this, you could try timing xgboost estimator `fit` on the current xgboost version we use, and then on the new version, and then comparing the timing. I'd start here.

2. It could also help to know which notebook is responsible for the timeout. To do that my instinct would be to try to reproduce the timing increase locally. I.e., run the following, before vs after this PR (don't forget to update your xgboost version to match!), and see if you can measure an overall time difference.
```bash
cd docs/source
date
make -C docs/ html
date
```
Better yet, I suppose you could try running the "start" notebook first, or any notebook which runs automl, and see if you can measure a runtime difference before vs after.

However if you're feeling stuck, feel free to close this and write a summary on the issue with a note saying we'll circle back later.

bchen1116:
Commented observations on the orignal post [here](https://github.com/alteryx/evalml/issues/1536#issuecomment-744555175). Closing this PR for now, as I am unsure what the issue is. If anyone has suggestions, I'd be happy to discuss more!

PR #1540, opened 2020-12-10
Update dependencies to fix `ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes'` errors

### LINKED ISSUE #1519: Update scikit-optimize requirements to be >=0.8.1 and scikit-learn>=0.23.1
@gsheni ran into issues with the following package versions:
```
numpy == 1.19.2
pandas == 1.1.0
scikit-learn == 0.23
scikit-optimize == 0.8
```

Following https://github.com/alteryx/evalml/issues/1121, it kooks like scikit-opt needs to be >=0.8.1 (https://github.com/scikit-optimize/scikit-optimize/issues/902) to fully address these dependencies.

From Slack, Gaurav had to upgrade to:
```
numpy == 1.19.4
pandas == 1.1.4
scikit-learn == 0.23.2
scikit-optimize == 0.8.1
```

Note: with current installation, we also get this error: 
```
ERROR: featuretools 0.22.0 has requirement pandas!=1.1.0,!=1.1.1,<2.0.0,>=0.24.1, but you'll have pandas 1.1.0 which is incompatible.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
woodwork 0.0.6 requires numpy>=1.19.1, but you have numpy 1.16.4 which is incompatible.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
woodwork 0.0.6 requires pandas>=1.1.0, but you have pandas 0.25.0 which is incompatible.
```

angela97lin:
Closes #1519 

Repro steps:
1. Create new virtualenv with no packages installed. I used python 3.7.6
2. Install the following:
numpy>=1.19.1
pandas>1.1.0
scikit-learn==0.23
scikit-optimize==0.8
3. Run `pip install -r requirements.txt` and install pytest
4. Run tests

This should raise the following:
```
================================================================= ERRORS =================================================================
_____________________________________________________ ERROR collecting test session ______________________________________________________
../../.pyenv/versions/3.7.6/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:953: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:953: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:967: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:677: in _load_unlocked
    ???
<frozen importlib._bootstrap_external>:728: in exec_module
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
evalml/__init__.py:9: in <module>
    import evalml.pipelines
evalml/pipelines/__init__.py:1: in <module>
    from .components import (
evalml/pipelines/components/__init__.py:2: in <module>
    from .estimators import (
evalml/pipelines/components/estimators/__init__.py:2: in <module>
    from .classifiers import (LogisticRegressionClassifier,
evalml/pipelines/components/estimators/classifiers/__init__.py:1: in <module>
    from .logistic_regression import LogisticRegressionClassifier
evalml/pipelines/components/estimators/classifiers/logistic_regression.py:3: in <module>
    from skopt.space import Real
../../.pyenv/versions/3.7.6/lib/python3.7/site-packages/skopt/__init__.py:55: in <module>
    from .searchcv import BayesSearchCV
../../.pyenv/versions/3.7.6/lib/python3.7/site-packages/skopt/searchcv.py:16: in <module>
    from sklearn.utils.fixes import MaskedArray
E   ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes' (/Users/angela.lin/.pyenv/versions/3.7.6/lib/python3.7/site-packages/sklearn/utils/fixes.py)
======================================================== short test summary info =========================================================
ERROR  - ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes' (/Users/angela.lin/.pyenv/versions/3.7.6/lib/python3....
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================================ 1 error in 3.06s ============================================================
```

Upgrading `scikit-learn==0.23.1` and `scikit-optimize==0.8.1` fixes this :)

--

This PR also tracks updating numpy and pandas versions. Since we rely on woodwork, we'll get errors (see issue) with our current minimal versions.

### DIFF
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -1,8 +1,8 @@
 numpy>=1.16.4
 pandas>=0.25.0
 scipy>=1.2.1
-scikit-learn>=0.23
-scikit-optimize>=0.8
+scikit-learn>=0.23.1
+scikit-optimize>=0.8.1
 colorama
 cloudpickle>=0.2.2
 click>=7.0.0
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -1,5 +1,5 @@
-numpy>=1.16.4
-pandas>=0.25.0
+numpy>=1.19.1
+pandas>=1.1.0
 scipy>=1.2.1
 scikit-learn>=0.23.1
 scikit-optimize>=0.8.1
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -20,6 +20,7 @@ Release Notes
     * Changes
         * Update circleci badge to apply to ``main`` :pr:`1489`
         * Added script to generate github markdown for releases :pr:`1487`
+        * Updated dependencies to fix ``ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes'`` error and to address Woodwork and Featuretool dependencies :pr:`1540`
     * Documentation Changes
     * Testing Changes
         * Set ``n_jobs=1`` in most unit tests to reduce memory :pr:`1505`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -27,6 +27,12 @@ Release Notes
     * Testing Changes
         * Set ``n_jobs=1`` in most unit tests to reduce memory :pr:`1505`
 
+.. warning::
+
+    **Breaking Changes**
+        * Updated minimal dependencies: ``numpy>=1.19.1``, ``pandas>=1.1.0``, ``scikit-learn>=0.23.1``, ``scikit-optimize>=0.8.1``
+
+
 
 **v0.16.1 Dec. 1, 2020**
     * Enhancements


jeremyliweishih (approved):
I think this looks good! Adding a warning to our release notes would be helpful especially since we're jumping quite a bit on our min versions.

core-requirements.txt
```diff
@@ -1,8 +1,8 @@
-numpy>=1.16.4
-pandas>=0.25.0
+numpy>=1.19.1
+pandas>=1.1.0
```
angela97lin, line 2: Woodwork/Featuretool minimal requirements

core-requirements.txt
```diff
@@ -1,8 +1,8 @@
-numpy>=1.16.4
-pandas>=0.25.0
+numpy>=1.19.1
```
freddyaboulton, line 1: If this is for ww and featuretools should we delete numpy and pandas from our core-requirements.txt?

freddyaboulton (approved):
Thanks for fixing this @angela97lin !

bchen1116 (approved):
LGTM! Agree with Freddy, if numpy and pandas are already on WW/FT requirements, do we need them on our requirements.txt as well?

core-requirements.txt
dsherry: Interesting point... We do rely on numpy a lot so I like that we place an explicit range on the version here. Also sometimes pip can do screwy things with ordering, so its safest to list every direct dependency in the reqs files.

dsherry, line 1: @angela97lin what happens if you keep this PR as-is, but delete this change so that older numpy versions are still allowed? Do the other changes still solve the problem, or is updating our min numpy also required?

dsherry: Also we can always push back on / override woodwork and featuretools if we feel their requirements are overzealous

core-requirements.txt
```diff
@@ -1,8 +1,8 @@
...
+pandas>=1.1.0
 scipy>=1.2.1
-scikit-learn>=0.23
-scikit-optimize>=0.8
+scikit-learn>=0.23.1
```
dsherry, line 4: @angela97lin so the problem was specifically in sklearn 0.23.0?

dsherry (approved):
@angela97lin looks good! I left some questions. Could we try not updating the min numpy version and see what happens?

I'm also curious why pandas 1.1.0 is required for woodwork (?) -- let's follow up with them on that. I wonder if they can lower the min pandas version. The more permissive we can make these ranges, the more likely people will be to use our stuff heh (certainly at the risk of bugs)

core-requirements.txt
angela97lin: > @angela97lin so the problem was specifically in sklearn 0.23.0?

Combination of scikit-learn 0.23 and scikit-optimize 0.8 and numpy: https://github.com/scikit-optimize/scikit-optimize/issues/902

core-requirements.txt
angela97lin: @dsherry I posted in the original issue, but we get this error if we try to specifically install a version less than these because of Woodwork requirements:
```
ERROR: featuretools 0.22.0 has requirement pandas!=1.1.0,!=1.1.1,<2.0.0,>=0.24.1, but you'll have pandas 1.1.0 which is incompatible.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
woodwork 0.0.6 requires numpy>=1.19.1, but you have numpy 1.16.4 which is incompatible.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
woodwork 0.0.6 requires pandas>=1.1.0, but you have pandas 0.25.0 which is incompatible.
```

angela97lin:
@dsherry Could be wrong but I don't think Woodwork can handle 0.25, because they rely so heavily on the new nullable structures (pd.NA, the new nullable float that is coming soon or was recently introduced in Pandas 1.X 🤔

dsherry (approved):
@angela97lin thanks for the details :) I filed [an issue in woodwork](https://github.com/alteryx/woodwork/issues/457) to track lowering the numpy/pandas limits to match featuretools and evalml.

Yes I see that `pd.NA` was [first added in pandas 1.0.0](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html#experimental-new-features), so perhaps the pandas version needs to stay at something like `pandas>=1.1.0`. I still think the numpy version limit in woodwork is too high and that we can lower it.

Let's merge this PR, and once woodwork puts out a release with lower pip reqs for numpy and/or pandas, we can update.

core-requirements.txt
freddyaboulton: Yea I agree listing numpy explicitly is nice but if we list a minimum version below woodwork's (or any of our dependencies) pip will install the minimum version needed for woodwork anyways. So I'm worried we'll have to keep updating our numpy dependency to match woodwork's to avoid bugs like this in the future.

dsherry: @freddyaboulton yep for sure. I hope that updates to the versions of core libs like numpy will be infrequent between our projects. Let's wait and see, and if we get feedback on this or have to make many more updates we can easily delete it.

freddyaboulton: Sounds good @dsherry !

PR #1543, opened 2020-12-10
Integrate ComponentGraphs into Pipelines

### LINKED ISSUE #1278: Update `PipelineBase` to use `ComponentGraph` object
* Pipeline API:
    * Make it easy to define graphs during pipeline class definition
    * Ability to access the graph as a static field on the pipeline class
    * Ability to access each component instance
    * Ability for pipelines to exist without an estimator as the last step
    * Ability to show a visualization of the graphs
    * Ability to access list of estimators, if any — to replace PipelineBase.estimator (https://github.com/alteryx/evalml/blob/main/evalml/pipelines/pipeline_base.py#L68)
    * Pipelines are no longer required to have an estimator
* Backwards compatibility: support old evalml pipeline class definitions

eccabay:
Closes #1278 

This definitely needs more tests but at this point I'm not sure what would be the best to add, so I'm very open to suggestions! 

[The most up to date design doc](https://alteryx.quip.com/OQ0aAKq8RIPc/Pipelines-as-DAGs-Final-Design)

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -124,10 +124,7 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         parameters = {}
         if self._pipeline_params:
             parameters['pipeline'] = self._pipeline_params
-        if isinstance(pipeline_class.component_graph, list):
-            component_graph = [handle_component_class(c) for c in pipeline_class.component_graph]
-        else:
-            component_graph = [handle_component_class(c[0]) for c in pipeline_class.component_graph.values()]
+        component_graph = [handle_component_class(c) for c in pipeline_class.linearized_component_graph]
         for component_class in component_graph:
             component_parameters = proposed_parameters.get(component_class.name, {})
             init_params = inspect.signature(component_class.__init__).parameters
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -124,7 +124,10 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         parameters = {}
         if self._pipeline_params:
             parameters['pipeline'] = self._pipeline_params
-        component_graph = [handle_component_class(c) for c in pipeline_class.component_graph]
+        if isinstance(pipeline_class.component_graph, list):
+            component_graph = [handle_component_class(c) for c in pipeline_class.component_graph]
+        else:
+            component_graph = [handle_component_class(c[0]) for c in pipeline_class.component_graph.values()]
         for component_class in component_graph:
             component_parameters = proposed_parameters.get(component_class.name, {})
             init_params = inspect.signature(component_class.__init__).parameters
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -1,5 +1,6 @@
 import networkx as nx
 import pandas as pd
+import woodwork as ww
 from networkx.algorithms.dag import topological_sort
 from networkx.exception import NetworkXUnfeasible
 
@@ -96,8 +97,17 @@ class ComponentGraph:
             X (pd.DataFrame): The input training data of shape [n_samples, n_features]
             y (pd.Series): The target training data of length [n_samples]
         """
-        self._compute_features(self.compute_order[:-1], X, y, fit=True)
-        return self
+        if len(self.compute_order) <= 1:
+            return X
+
+        component_outputs = self._compute_features(self.compute_order[:-1], X, y=y, fit=True)
+        final_component_inputs = []
+        for parent in self.get_parents(self.compute_order[-1]):
+            parent_output = component_outputs.get(parent, component_outputs.get(f'{parent}.x'))
+            if isinstance(parent_output, pd.Series):
+                parent_output = pd.DataFrame(parent_output, columns=[parent])
+            final_component_inputs.append(parent_output)
+        return pd.concat(final_component_inputs, axis=1)
 
     def predict(self, X):
         """Make predictions using selected features.
@@ -151,7 +161,8 @@ class ComponentGraph:
         """
         if len(component_list) == 0:
             return X
-
+        if isinstance(X, ww.DataTable):
+            X = X.to_dataframe()
         if not isinstance(X, pd.DataFrame):
             X = pd.DataFrame(X)
 
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -200,7 +200,7 @@ class ComponentGraph:
             else:
                 if fit:
                     component_instance.fit(input_x, input_y)
-                if not (fit and component_instance.name == self.get_last_component().name):  # Don't call predict on the final component during fit
+                if not (fit and component_name == self.compute_order[-1]):  # Don't call predict on the final component during fit
                     output = component_instance.predict(input_x)
                 else:
                     output = None
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -137,7 +137,7 @@ class ComponentGraph:
         if len(self.compute_order) <= 1:
             return X
 
-        component_outputs = self._compute_features(self.compute_order, X, y=y, fit=False)
+        component_outputs = self._compute_features(self.compute_order[:-1], X, y=y, fit=False)
         final_component_inputs = []
         for parent in self.get_parents(self.compute_order[-1]):
             parent_output = component_outputs.get(parent, component_outputs.get(f'{parent}.x'))
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -26,8 +26,7 @@ class ComponentGraph:
                 raise ValueError('All component information should be passed in as a list')
             component_class = handle_component_class(component_info[0])
             self.component_instances[component_name] = component_class
-        self.compute_order = []
-        self._recompute_order()
+        self.compute_order = self.generate_order(self.component_dict)
         self.input_feature_names = {}
 
     @classmethod
@@ -316,13 +315,14 @@ class ComponentGraph:
                                         for key, val in component_class.parameters.items()])  # noqa: W605
                 label = '%s |%s\l' % (component_name, parameters)  # noqa: W605
             graph.node(component_name, shape='record', label=label)
-        edges = self._get_edges()
+        edges = self._get_edges(self.component_dict)
         graph.edges(edges)
         return graph
 
-    def _get_edges(self):
+    @staticmethod
+    def _get_edges(component_dict):
         edges = []
-        for component_name, component_info in self.component_dict.items():
+        for component_name, component_info in component_dict.items():
             if len(component_info) > 1:
                 for parent in component_info[1:]:
                     if parent[-2:] == '.x' or parent[-2:] == '.y':
@@ -330,26 +330,26 @@ class ComponentGraph:
                     edges.append((parent, component_name))
         return edges
 
-    def _recompute_order(self):
+    @classmethod
+    def generate_order(cls, component_dict):
         """Regenerated the topologically sorted order of the graph"""
-        edges = self._get_edges()
-        if len(self.component_dict) == 1:
-            self.compute_order = list(self.component_dict.keys())
-            return
+        edges = cls._get_edges(component_dict)
+        if len(component_dict) == 1:
+            return list(component_dict.keys())
         if len(edges) == 0:
-            self.compute_order = []
-            return
+            return []
         digraph = nx.DiGraph()
         digraph.add_edges_from(edges)
         if not nx.is_weakly_connected(digraph):
             raise ValueError('The given graph is not completely connected')
         try:
-            self.compute_order = list(topological_sort(digraph))
+            compute_order = list(topological_sort(digraph))
         except NetworkXUnfeasible:
             raise ValueError('The given graph contains a cycle')
-        end_components = [component for component in self.compute_order if len(nx.descendants(digraph, component)) == 0]
+        end_components = [component for component in compute_order if len(nx.descendants(digraph, component)) == 0]
         if len(end_components) != 1:
             raise ValueError('The given graph has more than one final (childless) component')
+        return compute_order
 
     def __iter__(self):
         self._i = 0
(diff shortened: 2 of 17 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -280,19 +266,31 @@ def _score_all_objectives(self, X, y, y_pred, y_pred_proba, objectives):
     def model_family(cls):
         "Returns model family of this pipeline template"""
         component_graph = copy.copy(cls.component_graph)
-        return handle_component_class(component_graph[-1]).model_family
+        if isinstance(cls.component_graph, list):
+            return handle_component_class(component_graph[-1]).model_family
+        else:
+            # TODO: Does this make sense
+            return ModelFamily.ENSEMBLE
```
eccabay, line 273: I'm not sure what the best way to determine a model family for non-linear pipelines would be. The main problems:

- A non-linear pipeline can have multiple estimators, should the model family reflect all of them or just the final one?
- Since this is a class property, we only have access to the dictionary describing the pipeline, which may not be in any sort of computation order. Without the instantiated ComponentGraph object, we have no clean way of knowing which estimator is the final one

Since this is ambiguous, I've opted to bundle these all under the ensemble model family but I'm looking for input on this!

evalml/pipelines/utils.py
```diff
@@ -177,8 +177,10 @@ def generate_pipeline_code(element):
     code_strings = ['import json']
     if not isinstance(element, PipelineBase):
         raise ValueError("Element must be a pipeline instance, received {}".format(type(element)))
+    if isinstance(element.component_graph, dict):
+        raise ValueError("Code generation for nonlinear pipelines is not supported yet")
```
eccabay, line 181: I'm not sure how much work would be required for implementing code generation for nonlinear pipelines, is it worth it to file a separate issue to track this?

evalml/pipelines/pipeline_base.py
freddyaboulton: Interesting question! These are some of the places where we call `problem_type` at the pipeline level:

1. To [skip cv](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L643) for ensemble pipelines in `_compute_cv_scores` in AutoMLSearch.
2. To [set allowed_model_families](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L482) in AutoMLSearch after pipelines are created
3. To [describe](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/pipeline_base.py#L157) a pipeline.
4. To [check](https://github.com/alteryx/evalml/blob/main/evalml/tests/automl_tests/test_automl_search_classification.py#L538) which pipelines are created by automl in tests
5. To [check](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/graphs.py#L652) the input pipeline has a decision tree as a final estimator in our model understanding utils
6. To [check](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/graphs.py#L450) we don't pass in baseline pipelines for partial dependence
7. To [determine](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L115) which pipelines to add to our ensemble in IterativeAlgorithm
8. To [determine](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L101) how the pipeline scores should be saved in IterativeAlgorithm

In all of these use cases, what's intended is to get the final estimator/component model family. It's interesting we access `model_family` for the instance, not the class.  I think we can avoid the computation order problem you bring up by making this a `property` instead. 

I think we should fix this before merge. I don't think returning `ModelFamily.Ensemble` is a good idea because in our codebase, that typically means a `StackedEnsembleClassifier` or `StackedEnsembleRegressor` is in the pipeline, which isn't guaranteed to be the case here.

evalml/pipelines/utils.py
freddyaboulton: I think it should be a separate issue!

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -620,3 +621,32 @@ class CustomPipeline(MulticlassClassificationPipeline):
 
     automl = AutoMLSearch(problem_type="multiclass", allowed_pipelines=[CustomPipeline], n_jobs=1)
     automl.search(X, y)
+
+
+def test_automl_binary_nonlinear_pipeline_search(nonlinear_binary_pipeline_class, X_y_binary):
```
freddyaboulton, line 626: Since we're not mocking score or fit, can you make sure `n_jobs=1` for these estimators? I think you can add a method/modify `safe_init_pipeline_with_njobs_1` in the Helpers test fixture!

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -400,6 +420,29 @@ def test_predict(mock_predict, mock_fit, example_graph, X_y_binary):
     assert mock_fit.call_count == 3  # Only called during fit, not predict
 
 
+@patch('evalml.pipelines.components.Estimator.fit')
+@patch('evalml.pipelines.components.Estimator.predict')
+def test_predict_repeat_estimator(mock_predict, mock_fit, X_y_binary):
```
freddyaboulton, line 425: Is there a test that checks that the data from the components flows through the graph correctly? Like an estimator will call `predict` on the expected data coming from its parents.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -344,8 +344,8 @@ def test_make_pipeline_from_components(X_y_binary, logistic_regression_binary_pi
     est = RandomForestClassifier(random_state=7)
     pipeline = make_pipeline_from_components([imp, est], ProblemTypes.BINARY, custom_name='My Pipeline',
                                              random_state=15)
-    assert [c.__class__ for c in pipeline.component_graph] == [Imputer, RandomForestClassifier]
```
freddyaboulton, line 347: I think this highlights a subtle breaking change. Before `component_graph` would be populated with instances of the components after init. Now `component_graph` will always be the "static" graph definition (no instances). Is there a reason the instances are kept private as `_component_graph`?

If there is a reason to keep the instances separate from the "static" graph, I'd prefer if the instances were kept public like before. In that case, we can rename `_component_graph` to `component_instances` or something else?

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -124,7 +124,10 @@ def _transform_parameters(self, pipeline_class, proposed_parameters):
         parameters = {}
         if self._pipeline_params:
             parameters['pipeline'] = self._pipeline_params
-        component_graph = [handle_component_class(c) for c in pipeline_class.component_graph]
+        if isinstance(pipeline_class.component_graph, list):
```
freddyaboulton, line 127: I think this logic would be easier to maintain if this lived in the `PipelineBase` implementation. This can be a `linearized_component_graph` classproperty?

evalml/pipelines/component_graph.py
```diff
@@ -189,7 +200,7 @@ def _compute_features(self, component_list, X, y=None, fit=False):
             else:
                 if fit:
                     component_instance.fit(input_x, input_y)
-                if not (fit and component_instance.name == self.get_last_component().name):  # Don't call predict on the final component during fit
+                if not (fit and component_name == self.compute_order[-1]):  # Don't call predict on the final component during fit
```
freddyaboulton, line 203: How come we're not using `get_last_component().name`?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -504,7 +505,23 @@ def test_describe(caplog, logistic_regression_binary_pipeline_class):
     assert "Model Family: Linear" in out
```
freddyaboulton, line 505: We should add a test to check the model family for non linear pipelines.

evalml/pipelines/component_graph.py
```diff
@@ -96,8 +97,17 @@ def fit_features(self, X, y):
             X (pd.DataFrame): The input training data of shape [n_samples, n_features]
             y (pd.Series): The target training data of length [n_samples]
         """
-        self._compute_features(self.compute_order[:-1], X, y, fit=True)
-        return self
+        if len(self.compute_order) <= 1:
+            return X
+
+        component_outputs = self._compute_features(self.compute_order[:-1], X, y=y, fit=True)
+        final_component_inputs = []
```
freddyaboulton, line 104: Just curious, what prompted this change?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -344,8 +344,8 @@ def test_make_pipeline_from_components(X_y_binary, logistic_regression_binary_pi
     est = RandomForestClassifier(random_state=7)
```
freddyaboulton, line 344: I don't think we have a test anywhere in this file for `compute_estimator_features`. Can you add some for the linear and non-linear case please?

freddyaboulton (approved):
@eccabay Looks good! I have a couple of comments on the implementation/design. 

What's the plan for adding support for non-linear pipelines in `make_pipeline_from_components`  ?

evalml/pipelines/component_graph.py
```diff
@@ -96,8 +97,17 @@ def fit_features(self, X, y):
...
+        for parent in self.get_parents(self.compute_order[-1]):
+            parent_output = component_outputs.get(parent, component_outputs.get(f'{parent}.x'))
+            if isinstance(parent_output, pd.Series):
+                parent_output = pd.DataFrame(parent_output, columns=[parent])
+            final_component_inputs.append(parent_output)
+        return pd.concat(final_component_inputs, axis=1)
```
freddyaboulton, line 110: Is there a unit test where the input dataframe has a custom index? pd.concat introduce NaNs anywhere the indices don't perfectly match up.

evalml/pipelines/component_graph.py
eccabay: `Pipeline._compute_features_during_fit`, which calls this method, needs an `X_t` to pass to the final estimator

evalml/pipelines/component_graph.py
eccabay: `get_last_component().name` will use the generic name of the component, rather than the custom name given to that specific component. If there are multiples of the same estimator in the pipeline and one of them is the final component, it will not call predict on the non-final one.

evalml/tests/pipeline_tests/test_pipelines.py
eccabay: There is technically one right below, line 521! I can add something more explicit as well though

evalml/automl/automl_algorithm/iterative_algorithm.py
eccabay: I'm not quite sure what you mean - would `linearized_component_graph` be a boolean flag indicating if the pipeline is linear, or a list of the components, regardless of if the graph is a dictionary or list?

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: Great! Thanks for pointing that one out to me hehe

evalml/automl/automl_algorithm/iterative_algorithm.py
eccabay: If you're thinking a list of the components in a linear order, would it make sense to combine this with [your other suggestion](https://github.com/alteryx/evalml/pull/1543#discussion_r543584707) above? Some sort of property that contains the component instances in a list

freddyaboulton: I was thinking of the list of component classes, regardless if the graph is a dictionary or a list!

freddyaboulton: I think it would be tough to combine it with the other suggestion because here we don't have the pipeline instance, just the class definition so I don't think you'll be able to grab the component instances. That being said, if there's a way to combine them that I'm not seeing that would be great!

Also, `linearized_component_graph` is just the first thing that came to mind 😂  I'm sure there are a billion better names for it hehe.

eccabay: Gotcha - so this class property wouldn't necessarily guarantee computation order then, just some linear order of components that may or may not make sense

freddyaboulton: Yea in this case the order would not matter!

evalml/pipelines/component_graph.py
freddyaboulton: Thanks for explaining!

evalml/tests/pipeline_tests/test_pipelines.py
eccabay: This is a really interesting point. The challenge here is that the instances are contained within the `ComponentGraph` object, which we're keeping completely private from users. It may be possible to add a separate public reference to the instances directly from the pipeline, but I worry that between this and the `linearized_component_graph` classproperty we discussed earlier, we're adding a fair amount of redundant information that's only there to maintain the current status quo.

What do you think?

freddyaboulton: I see what you mean about wanting to keep `ComponentGraph` private from the end user. I think you're right that we shouldn't add another accessor for the instances. Maybe what we should do is just make iterating over/accessing the pipeline instead of the pipeline.component_graph the "preferred" public accessor for the component instances. The benefit is that this functionality is already built into your PR!

For example, the line we're commenting on would turn to `[c.__class__ for c in pipeline]`. 

This would require updating our [docs](https://evalml.alteryx.com/en/stable/user_guide/pipelines.html) to not access instances with `pipeline.component_graph` and our unit tests as well. 

What do you think? If you are on board, we should leave this for another issue since this diff is large already.

eccabay: I like this idea, I think it makes a lot of sense! I do agree it's better fit for a separate issue though haha, so I can file that if other people also think this would be beneficial!

evalml/tests/pipeline_tests/test_component_graph.py
eccabay: Yep, there's a mocked test `test_component_graph_evaluation_plumbing` from my previous PR. I realized I never added a test for the `ComponentGraph.input_feature_names` which also reflects the flow of data, so I've added a separate test for that as well!

evalml/pipelines/component_graph.py
eccabay: I've added one! Since a number of components reset the indices and none should be dropping any rows, the test passed without a problem but it's definitely still a good thing to look out for

evalml/tests/pipeline_tests/test_pipelines.py
eccabay: Filed issue #1571

PR #1547, opened 2020-12-11
Return trained pipeline for AutoMLSearch.best_pipeline

### LINKED ISSUE #1546: Have automl auto-fit the best pipeline on entire training data
**Background**
Today in order to get a trained pipeline out of automl you need to call `fit` on the pipeline, because automl always returns untrained copies of the pipeline:
```python
automl.search(X_train, y_train)
best_pipeline = automl.best_pipeline
best_pipeline.fit(X_train, y_train)
best_pipeline.score(X_test, y_test, objectives=['MSE'])
pipeline = automl.get_pipeline(42)
pipeline.score(X_test, y_test, objectives=['MSE'])
```

**Challenge**
We'd like to make it as easy as possible for people to run automl, select a pipeline and use that pipeline to iterate and debug, to generate insights and to deploy to production.

**Proposal**
In the short-term (i.e. this issue): have the `best_pipeline` accessor return a trained pipeline:
```python
automl.search(X_train, y_train)
best_pipeline = automl.best_pipeline
best_pipeline.score(X_test, y_test, objectives=['MSE'])
```
If automl search hasn't run yet, that accessor should error.

My recommendation for how to implement this is to update automl search to fit the best pipeline at the end and save a reference to that trained pipeline.

Don't forget to update the user guide!

There are also implications for perf testing: we should update looking glass to record the automl search time and the time to fit the best pipeline separately, because they're independent operations.

**Future**
Long-term, I'd like us to create an abstraction for holding a reference to the data outside of the call to `search`. This would allow us to do things like have `get_pipeline` return trained pipelines as well, without us having to train all the pipelines during the call to automl `search`.

bchen1116:
fix #1546 

Added quick note to [docs here](https://evalml.alteryx.com/en/bc_1546_best_pipeline/user_guide/automl.html#Get-best-pipeline)


evalml/automl/automl_search.py
```diff
@@ -530,6 +535,11 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
 
         best_pipeline = self.rankings.iloc[0]
         best_pipeline_name = best_pipeline["pipeline_name"]
+        best_pipe = self.get_pipeline(best_pipeline['id'])
+        if self._train_best_pipe:
+            self._compute_cv_scores(best_pipe, X, y, True)
+        else:
+            self._best_pipe = best_pipe
```
dsherry, line 542: @bchen1116 why call `_compute_cv_scores` here? Why not
```python
self._best_pipeline = self.get_pipeline(best_pipeline['id'])
if self._train_best_pipeline:
    self._best_pipeline.fit(X, y)
```

evalml/automl/automl_search.py
```diff
@@ -933,16 +947,15 @@ def full_rankings(self):
...
+            PipelineBase: Trained pipeline instance associated with the best automl search result.
         """
-        if not self.has_searched:
-            raise PipelineNotFoundError("automl search must be run before selecting `best_pipeline`.")
+        if not (self.has_searched and self._best_pipe):
+            raise PipelineNotFoundError("automl search must be completely run before selecting `best_pipeline`.")
 
-        best = self.rankings.iloc[0]
-        return self.get_pipeline(best["id"])
+        return self._best_pipe
```
dsherry, line 958: 👍

docs/source/release_notes.rst
```diff
@@ -14,6 +14,7 @@ Release Notes
         * Added more information to users about ensembling behavior in ``AutoMLSearch`` :pr:`1527`
         * Add woodwork support for more utility and graph methods :pr:`1544`
         * Changed ``DateTimeFeaturizer`` to encode features as int :pr:`1479`
+        * Return trained pipelines from ``AutoMLSearch.best_pipeline`` :pr:`1547`
```
kmax12, line 17: we should make sure to document this as a breaking change and let people know they can maintain current functionality by setting `train_best_pipeline=False`

evalml/automl/automl_search.py
```diff
@@ -256,6 +259,8 @@ def __init__(self,
 
         self._validate_problem_type()
         self.problem_configuration = self._validate_problem_configuration(problem_configuration)
+        self._train_best_pipe = train_best_pipeline
+        self._best_pipe = None
```
dsherry, line 263: This is total style nit-pick but if you say `pipeline` instead of `pipe` for these variable names it'll potentially make it easier to grep through the code 😅

evalml/automl/automl_search.py
```diff
@@ -947,16 +956,15 @@ def full_rankings(self):
 
     @property
     def best_pipeline(self):
-        """Returns an untrained instance of the best pipeline and parameters found during automl search.
+        """Returns a trained instance of the best pipeline and parameters found during automl search.
```
dsherry, line 959: If `self._train_best_pipeline` is true, this is correct. But lets update this description to document what happens when its false too.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1763,29 +1768,29 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
     X, y = X_y_binary
     msg = 'all your model are belong to us'
     mock_fit.side_effect = Exception(msg)
-    automl = AutoMLSearch(problem_type="binary", error_callback=None)
+    automl = AutoMLSearch(problem_type="binary", error_callback=None, train_best_pipeline=False)
```
dsherry, line 1771: Why set this to false in this unit test? I thought this unit test was completely unrelated to the `best_pipeline` accessor

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1795,7 +1800,7 @@ def test_automl_error_callback(mock_fit, mock_score, X_y_binary, caplog):
         assert str(e) == msg
 
     caplog.clear()
-    automl = AutoMLSearch(problem_type="binary", error_callback=raise_and_save_error_callback)
+    automl = AutoMLSearch(problem_type="binary", error_callback=raise_and_save_error_callback, train_best_pipeline=False)
```
dsherry, line 1803: Same question here as in `test_automl_error_callback`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1904,3 +1909,17 @@ class Pipeline2(TimeSeriesRegressionPipeline):
...
+def test_automl_best_pipeline_false(mock_fit, mock_score, X_y_binary):
+    X, y = X_y_binary
+
+    automl = AutoMLSearch(problem_type='binary', train_best_pipeline=False)
+    automl.search(X, y)
+    assert not automl.best_pipeline._is_fitted
+
+    automl = AutoMLSearch(problem_type='binary')
+    automl.search(X, y)
+    assert automl.best_pipeline._is_fitted
```
dsherry, line 1925: Sounds good.

Since `_is_fitted` is private, the future-proof way to check this here and in the other tests below would be, if you expect the pipeline to *not* be fitted:
```python
with pytest.raises(SomeException, match="not fitted"):
    automl.best_pipeline.predict(X)
```
or in the case when you expect the pipeline to be fitted:
```python
automl.best_pipeline.predict(X)
```

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -224,7 +226,7 @@ def test_optimizable_threshold_enabled(mock_fit, mock_score, mock_predict_proba,
     mock_score.assert_called()
     mock_predict_proba.assert_called()
     mock_optimize_threshold.assert_called()
-    assert automl.best_pipeline.threshold is None
+    assert automl.best_pipeline.threshold is not None
```
dsherry, line 229: @bchen1116 ah, good call. Will the best pipeline have bin class threshold optimized? We should do what is necessary to make that true.

dsherry (approved):
@bchen1116 wonderful!! Glad you got through that cryptic windows error :)

Blocking comments from me:
* Let's make sure (if not already the case) that for bin class, the best pipeline classification threshold is optimized
* Update accessor docstring
* I left comments in the unit tests about not setting `train_best_pipeline=False` in unrelated unit tests (possible I'm missing something there)
* Ideally, let's not use `_is_fitted` directly, I left a suggestion

evalml/tests/automl_tests/test_automl.py
bchen1116: The test mocks the `BinaryClassificationPipeline.fit` method, which means that when I call `best_pipeline.fit()`, it returns an exception that isn't captured and ends up breaking out of the code. This code just checks that callbacks within `_compute_cv_scores` works, so any `fit` call outside of that fails the test here.

evalml/tests/automl_tests/test_automl.py
bchen1116: Similar reason as above

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: I see. I don't think it is currently, but I can fix that. Thanks!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -176,7 +180,7 @@ def test_pipeline_fit_raises(mock_fit, X_y_binary, caplog):
     mock_fit.side_effect = Exception(msg)
     X, y = X_y_binary
 
-    automl = AutoMLSearch(problem_type='binary', max_iterations=1)
+    automl = AutoMLSearch(problem_type='binary', max_iterations=1, train_best_pipeline=False)
```
bchen1116, line 183: Don't train the best pipeline, since this test mocks the `pipeline.fit()` method and causes it to raise an exception, which we don't want to raise while fitting the best pipeline.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1115,7 +1123,7 @@ def test_error_during_train_test_split(mock_fit, mock_score, mock_train_test_spl
     X, y = X_y_binary
     mock_score.return_value = {'Log Loss Binary': 1.0}
     mock_train_test_split.side_effect = RuntimeError()
-    automl = AutoMLSearch(problem_type='binary', objective='Accuracy Binary', max_iterations=2, optimize_thresholds=True)
+    automl = AutoMLSearch(problem_type='binary', objective='Accuracy Binary', max_iterations=2, optimize_thresholds=True, train_best_pipeline=False)
```
bchen1116, line 1126: Don't train best pipeline since we use `train_test_split` to find the ideal threshold for binary classification pipelines

evalml/automl/automl_search.py
```diff
@@ -947,16 +974,15 @@ def full_rankings(self):
...
+        """If `train_best_pipeline`, returns a trained instance of the best pipeline and parameters found during automl search. Otherwise, returns an untrained pipeline instance.
 
         Returns:
-            PipelineBase: untrained pipeline instance associated with the best automl search result.
+            PipelineBase: If `train_best_pipeline`, trained pipeline instance associated with the best automl search result. Otherwise, untrained pipeline instance.
         """
-        if not self.has_searched:
-            raise PipelineNotFoundError("automl search must be run before selecting `best_pipeline`.")
+        if not (self.has_searched and self._best_pipeline):
+            raise PipelineNotFoundError("automl search must be completely run before selecting `best_pipeline`.")
```
dsherry, line 983: Ah, good call.

Nit-pick but I'd remove "completely", not sure if users will understand what that means.

evalml/tests/automl_tests/test_automl.py
dsherry: Got it. This could be a good comment to add.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1078,6 +1081,9 @@ def test_catch_keyboard_interrupt(mock_fit, mock_score, mock_input,
     automl.search(X, y)
 
     assert len(automl._results['pipeline_results']) == number_results
+    if number_results != 5 and when_to_interrupt == 1:
```
dsherry, line 1084: Why `!= 5` here?

dsherry (approved):
@bchen1116 LGTM! I filed a couple related bugs (#1572 #1573 ) and left comments about the latter. I think it would be great to find a way to avoid duplicating code in `search` and `_compute_cv_scores` RE training the best pipeline--take a shot at that and if there's no clear solution, what's in this PR now is great to merge.

evalml/automl/automl_search.py
```diff
@@ -544,6 +549,28 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
 
         best_pipeline = self.rankings.iloc[0]
         best_pipeline_name = best_pipeline["pipeline_name"]
+        self._best_pipeline = self.get_pipeline(best_pipeline['id'])
+        if self._train_best_pipeline:
+            X_train = _convert_to_woodwork_structure(X)
+            y_train = _convert_to_woodwork_structure(y)
+            X_threshold_tuning = None
+            y_threshold_tuning = None
+            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
```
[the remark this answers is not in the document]
bchen1116: @dsherry Yep! Sounds good, I can fix that. Should I hold off on adding that one line for `ProblemTypes.BINARY`? I think it's a quick fix that I can fit in this PR, but also fine to hold off and fix it in a separate PR.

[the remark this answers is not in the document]
dsherry: Up to you :) if it were me I'd look at minimizing code duplication and merge this PR, then #1573 is a quick separate PR.

evalml/tests/automl_tests/test_automl.py
bchen1116: Only the first 2 `parametrize` args fail since they result in 0 pipelines being trained. I can simplify this though!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1904,3 +1911,30 @@ class Pipeline2(TimeSeriesRegressionPipeline):
             continue
         assert result['parameters']['Delayed Feature Transformer'] == configuration
         assert result['parameters']['pipeline'] == configuration
+
+
+@patch('evalml.objectives.BinaryClassificationObjective.optimize_threshold')
```
freddyaboulton, line 1916: @bchen1116 I think we need to set `n_jobs=1` here too!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -637,9 +642,8 @@ def test_large_dataset_regression(mock_score):
         assert automl.results['pipeline_results'][pipeline_id]['score'] == automl.results['pipeline_results'][pipeline_id]['validation_score']
 
 
-@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
 @patch('evalml.pipelines.BinaryClassificationPipeline.score')
-def test_large_dataset_split_size(mock_fit, mock_score):
+def test_large_dataset_split_size(mock_score):
```
freddyaboulton, line 646: I think we need to set `n_jobs=1` here too!

bchen1116:
Wrote a quick little summary [here](https://alteryx.quip.com/WQ4bAEzBUNfI/Potential-threading-bug-found) of the bug I found during this testing, but wasn't able to reproduce it through shorter code. The bug doesn't occur in this implementation but was just a strange occurrance I found.

docs/source/user_guide/automl.ipynb
```diff
@@ -236,7 +236,8 @@
    "metadata": {},
    "source": [
     "### Get best pipeline\n",
-    "If we specifically want to get the best pipeline, there is a convenient accessor for that."
+    "If we specifically want to get the best pipeline, there is a convenient accessor for that.\n",
```
dsherry, line 239: I know this is from before your PR but nit-pick: `we` --> `you` ?

evalml/automl/automl_search.py
```diff
@@ -529,9 +534,46 @@ def search(self, X, y, data_checks="auto", show_iteration_plot=True):
...
+                X_train, X_threshold_tuning, y_train, y_threshold_tuning = train_test_split(X_train, y_train, test_size=0.2, random_state=self.random_state)
+            self._best_pipeline.fit(X_train, y_train)
+            self._best_pipeline = self._threshold_pipeline(self._best_pipeline, X_threshold_tuning, y_threshold_tuning)
         logger.info(f"Best pipeline: {best_pipeline_name}")
         logger.info(f"Best pipeline {self.objective.name}: {best_pipeline['score']:3f}")
 
+    def _threshold_pipeline(self, pipeline, X_threshold_tuning, y_threshold_tuning, fold=None):
```
dsherry, line 541: Nit-pick: call this `_tune_binary_threshold` or similar

evalml/automl/automl_search.py
```diff
@@ -522,9 +527,46 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+                if fold is not None:
+                    logger.debug(f"\t\t\tFold {fold}: Optimizing threshold for {self.objective.name}")
+                y_predict_proba = pipeline.predict_proba(X_threshold_tuning)
+                if isinstance(y_predict_proba, pd.DataFrame):
+                    y_predict_proba = y_predict_proba.iloc[:, 1]
+                else:
+                    y_predict_proba = y_predict_proba[:, 1]
+                pipeline.threshold = self.objective.optimize_threshold(y_predict_proba, y_threshold_tuning, X=X_threshold_tuning)
+                if fold is not None:
+                    logger.debug(f"\t\t\tFold {fold}: Optimal threshold found ({pipeline.threshold:.3f})")
```
dsherry, line 565: This is minor, but I suggest we move this log to `_compute_cv_scores` and delete the other one
```python
cv_pipeline = self._threshold_pipeline(cv_pipeline, X_threshold_tuning, y_threshold_tuning, fold=i)
if pipeline.threshold is not None and self.optimize_thresholds:
    logger.debug(f"\t\t\tFold {i}: Optimal threshold found ({pipeline.threshold:.3f})")
```

That way the logging code is all in `_compute_cv_scores` and this helper just has impl code.

evalml/automl/automl_search.py
```diff
@@ -522,9 +527,46 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+            X_threshold_tuning (ww DataTable): X data to tune pipeline to
+            y_threshold_tuning (ww DataColumn): Target data to tune pipeline to
+            fold (None, int): The fold of training this belongs to in AutoMLSearch. Defaults to None
+
+        Returns:
+            Trained pipeline instance
+        """
+        if self.objective.is_defined_for_problem_type(ProblemTypes.BINARY):
+            pipeline.threshold = 0.5
+            if X_threshold_tuning:
```
dsherry, line 555: Why not `if self.optimize_thresholds and self.objective.can_optimize_threshold` here like the old code did? We do need to check those things.

evalml/automl/automl_search.py
```diff
@@ -522,9 +527,46 @@ def search(self, data_checks="auto", show_iteration_plot=True):
 
         best_pipeline = self.rankings.iloc[0]
         best_pipeline_name = best_pipeline["pipeline_name"]
+        self._best_pipeline = self.get_pipeline(best_pipeline['id'])
+        if self._train_best_pipeline:
+            X_train = _convert_to_woodwork_structure(X)
+            y_train = _convert_to_woodwork_structure(y)
```
dsherry: I think you should get rid of this now that #1597 is merged, and call `self.X_train` / `self.y_train`

evalml/automl/automl_search.py
```diff
@@ -522,9 +527,46 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+            X_threshold_tuning = None
+            y_threshold_tuning = None
+            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
+                X_train, X_threshold_tuning, y_train, y_threshold_tuning = train_test_split(X_train, y_train, test_size=0.2, random_state=self.random_state)
```
dsherry: Hmm... now that #1597 is merged, we should really move the computation of this threshold tuning split out into the `AutoMLSearch` constructor.

This is fine for now. I filed #1621 to track.

evalml/automl/automl_search.py
```diff
@@ -908,16 +938,15 @@ def full_rankings(self):
 
     @property
     def best_pipeline(self):
-        """Returns an untrained instance of the best pipeline and parameters found during automl search.
+        """If `train_best_pipeline`, returns a trained instance of the best pipeline and parameters found during automl search. Otherwise, returns an untrained pipeline instance.
```
dsherry, line 941: This is good. I think it could be easier to follow if you said "Returns an untrained instance of the best pipeline and parameters found during automl search. If `train_best_pipeline` is set to False, returns an untrained pipeline instance.

evalml/automl/automl_search.py
```diff
@@ -908,16 +938,15 @@ def full_rankings(self):
...
 
         Returns:
-            PipelineBase: untrained pipeline instance associated with the best automl search result.
+            PipelineBase: If `train_best_pipeline`, trained pipeline instance associated with the best automl search result. Otherwise, untrained pipeline instance.
         """
-        if not self.has_searched:
+        if not (self.has_searched and self._best_pipeline):
```
dsherry, line 946: Nice. Idk when the search will have been marked as ran but no best pipeline would get set, but seems fine to protect against that here.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1910,6 +1917,33 @@ class Pipeline2(TimeSeriesRegressionPipeline):
...
+
+    automl = AutoMLSearch(problem_type='binary', optimize_thresholds=True, objective="Log Loss Binary", n_jobs=1)
+    automl.search(X, y)
+    automl.best_pipeline.predict(X)
+    assert automl.best_pipeline.threshold == 0.5
+
+    automl = AutoMLSearch(problem_type='binary', optimize_thresholds=True, objective="Accuracy Binary", n_jobs=1)
+    automl.search(X, y)
+    automl.best_pipeline.predict(X)
+    assert automl.best_pipeline.threshold == 0.62
```
dsherry, line 1944: This test rocks!! 🪨

dsherry (approved):
@bchen1116 nice! I left a comment about the `if` statements in the `_threshold_pipeline` helper, and some minor suggestions. Tests look great!

evalml/automl/automl_search.py
dsherry: I wrote this before you rebased/updated your PR, so LGTM!

evalml/automl/automl_search.py
bchen1116: @dsherry I did this since `X_threshold_tuning` would only be defined if `self.optimize_thresholds and self.objective.can_optimize_threshold` were both True. I can change it back, but this check seemed simpler and cleaner?

evalml/automl/automl_search.py
bchen1116: Thanks for filing!

PR #1550, opened 2020-12-14
1535 save graphs to file

### LINKED ISSUE #1535: Model understanding: save graphs to file
It would be nice if we had a uniform way to save model understanding plots to file.

This could be a `save_plot` utility which can be called on the output of each model understanding method, or it could mean adding a `filepath` argument to each model understanding graphing method and using a common impl under the hood.

Its also possible that the plotly utilities for saving figures are enough and that we should not add saving to file! Whoever picks this up should investigate that first: whats the minimal plotly code required to save a plot/graph to file, and what if any functionality is missing from that which we could wrap up in a util?

ParthivNaresh:
Fixes #1535 

Allows the user to save an image belonging to plotly.Figure, graphviz.Source, matplotlib.pyplot.Figure, or matplotlib.axes.SubplotBase to their local file system.

### DIFF
--- a/evalml/tests/utils_tests/test_gen_utils.py
+++ b/evalml/tests/utils_tests/test_gen_utils.py
@@ -516,8 +516,7 @@ def test_save_graphviz_default_format(file_name, format, interactive, fitted_tre
 
 @pytest.mark.parametrize("file_name,format,interactive",
                          [
-                             ('test_plot', 'jpeg', False),
-                             (None, 'jpeg', False)
+                             ('test_plot', 'jpeg', False)
                          ])
 def test_save_graphviz_different_format(file_name, format, interactive, fitted_tree_estimators, tmpdir):
     est_class, est_reg = fitted_tree_estimators
@@ -531,8 +530,6 @@ def test_save_graphviz_different_format(file_name, format, interactive, fitted_t
     assert os.path.exists(output_)
     assert isinstance(output_, str)
     assert os.path.basename(output_) == 'test_plot.jpeg'
-    if not file_name:
-        os.remove('test_plot.jpeg')
 
 
 @pytest.mark.parametrize("file_name,format,interactive",
--- a/evalml/tests/utils_tests/test_gen_utils.py
+++ b/evalml/tests/utils_tests/test_gen_utils.py
@@ -588,7 +588,7 @@ def test_save_seaborn_default_format(file_name, format, interactive, fitted_tree
     def setup_plt():
         data_ = [0, 1, 2, 3, 4]
         fig = sns.scatterplot(data=data_)
-        #fig_ = fig.figure
+        # fig_ = fig.figure
         return fig
 
     fig = setup_plt()
--- a/evalml/tests/utils_tests/test_gen_utils.py
+++ b/evalml/tests/utils_tests/test_gen_utils.py
@@ -1,5 +1,6 @@
 import inspect
 import os
+import shutil
 from unittest.mock import MagicMock, patch
 
 import numpy as np
@@ -7,7 +8,8 @@ import pandas as pd
 import pytest
 import woodwork as ww
 
-from evalml.pipelines import BinaryClassificationPipeline
+from evalml.model_understanding.graphs import visualize_decision_tree
+from evalml.pipelines import MulticlassClassificationPipeline
 from evalml.pipelines.components import ComponentBase
 from evalml.utils.gen_utils import (
     SEED_BOUNDS,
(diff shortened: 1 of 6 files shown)


evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -391,3 +394,197 @@ def test_convert_to_woodwork_structure():
...
+                             ('test_plot.', 'png', False),
+                             ('test_plot.png', 'jpeg', False)
+                         ])
+def test_save_seaborn_default_format(file_name, format, interactive, fitted_tree_estimators, tmpdir, has_minimal_dependencies):
+    sns = pytest.importorskip("seaborn")
+
+    def setup_plt():
+        data_ = [0, 1, 2, 3, 4]
+        fig = sns.scatterplot(data=data_)
+        # fig_ = fig.figure
```
bchen1116, line 579: Don't need this comment

evalml/utils/gen_utils.py
```diff
@@ -379,3 +380,98 @@ def _not_nan(pd_data):
...
+        extension = extension[1:].lower() if extension else None
+        if is_plotly and interactive:
+            format_ = 'html'
+        elif not extension and not interactive:
+            format_ = format
+        else:
+            format_ = extension
+        filepath = f'{path_and_name}.{format_}'
+        try:
+            # saving_folder = os.path.dirname(filepath)
```
bchen1116, line 409: unnecessary comments

evalml/utils/gen_utils.py
```diff
@@ -379,3 +380,98 @@ def _not_nan(pd_data):
...
+        interactive (bool, optional): If True and fig is of type plotly.Figure, sets the format to 'html'.
+        is_plotly (bool, optional): Check to see if the fig being passed is of type plotly.Figure.
+
+    Returns:
+        String representing the final filepath the image will be saved to.
+    """
+    if filepath:
+        filepath = str(filepath)
+        path_and_name, extension = os.path.splitext(filepath)
+        extension = extension[1:].lower() if extension else None
```
bchen1116, line 400: Although it is in the docstring, and users can choose to return the final filepath, I wanted to bring up a suggestion to somehow give a message to the user about the file saving under a different name than the user expected. 

For instance, if the user passed in `save_plot(fig, filepath="plot.png", format="png", interactive=True)` and fig was a plotly.Figure, it could save some confusion if we told them something along the lines of `Saving plot as 'plot.html' since 'interactive=True' and fig is a plotly.Figure`. 

Definitely not necessary since it's all in the docstring already, but I figure it could save some confusion, so I wanted to bring it up and see what other people thought!

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -391,3 +394,197 @@ def test_convert_to_woodwork_structure():
...
+        assert os.path.exists(output_)
+        assert isinstance(output_, str)
+        assert os.path.basename(output_) == 'test_plot.png'
+
+
+@pytest.mark.parametrize("file_name,format,interactive",
+                         [
+                             ('test_plot', 'jpeg', False)
+                         ])
+def test_save_plotly_static_different_format(file_name, format, interactive, decision_tree_classification_pipeline_class, tmpdir, has_minimal_dependencies):
```
bchen1116, line 425: Tests are great! Nit-pick, but do we need to run this and the previous test separately? Can't we combine them with `pytest.mark.parametrize`, and include another column with the expected name? In my opinion, it wouldn't seem more messy to do it that way, unless I'm missing something.

evalml/utils/gen_utils.py
```diff
@@ -379,3 +380,98 @@ def _not_nan(pd_data):
...
+    elif isinstance(fig, graphviz_.Source):
+        is_graphviz = True
+    elif isinstance(fig, plt_.Figure):
+        is_plt = True
+    elif isinstance(fig, axes_.SubplotBase):
+        is_seaborn = True
+
+    if not filepath:
+        extension = 'html' if interactive and is_plotly else format
+        filepath = os.path.join(os.getcwd(), f'test_plot.{extension}')
```
bchen1116, line 458: Can you add in the docstring that the default `filepath` is `test_plot`?

bchen1116 (approved):
Looks good! Left a few comments on potential ways to clean tests and a suggestion on letting users know when plots are saved under different names.

evalml/tests/utils_tests/test_gen_utils.py
ParthivNaresh: @bchen1116 That's true, it definitely wouldn't be more messy, I was just trying to separate the tests by function/theme for ease of retest later if necessary

evalml/utils/gen_utils.py
ParthivNaresh: @bchen1116 Good point, and just for clarity I will add the specification that the format would be set to html in the docstring.

Issue #1548, opened 2020-12-11
InvalidTargetDataCheck: make problem_type type checking smarter


angela97lin:
Right now, in `InvalidTargetDataCheck`, we have:
```
        valid_target_types = [dtype for dtype in numeric_and_boolean_dtypes + categorical_dtypes]
        if y.dtype.name not in valid_target_types:
```

This covers all of the types we handle. However, we can also take into consideration the `problem_type` attribute to be smarter about this--for example, if problem type is regression, and the target has categorical data.

dsherry:
Thanks @angela97lin ! Agreed. RE our discussion on #1551 , it would be helpful if the data check errored if
* For regression, the problem type was not "numeric"
* For binary, the problem type was "categorical" (but first we should confirm that if a categorical column has only two unique values (other than `nan`) woodwork will label it as binary, not as "categorical"
* For multiclass, the problem type was binary
* For multiclass, the problem type had a high number of unique values--perhaps set a max cap at over 5%

PR #1551, opened 2020-12-14
 Update evalml from selecting using pandas dtypes to selecting using Woodwork logical types

### LINKED ISSUE #1290: Update components from selecting using pandas dtypes to selecting using DataTable semantic or logical types

angela97lin:
Closes #1290


evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -59,25 +59,24 @@ def validate(self, X, y):
             raise ValueError("y cannot be None")
 
         y = _convert_to_woodwork_structure(y)
```
angela97lin, line 61: Moved above because we can do logic with Woodwork types, before converting to pandas to determine # of null

evalml/pipelines/components/transformers/dimensionality_reduction/lda.py
```diff
@@ -2,7 +2,11 @@
...
 from evalml.pipelines.components.transformers import Transformer
-from evalml.utils.gen_utils import is_all_numeric
+from evalml.utils.gen_utils import (
+    _convert_to_woodwork_structure,
+    _convert_woodwork_types_wrapper,
+    is_all_numeric
+)
 
 
 class LinearDiscriminantAnalysis(Transformer):
```
angela97lin, line 12: Adding support for new LDA component :)

evalml/tests/automl_tests/test_automl.py
```diff
@@ -975,9 +973,9 @@ def test_results_getter(mock_fit, mock_score, X_y_binary):
 
 @pytest.mark.parametrize("data_type", ['np', 'pd', 'ww'])
 @pytest.mark.parametrize("automl_type", [ProblemTypes.BINARY, ProblemTypes.MULTICLASS])
-@pytest.mark.parametrize("target_type", numeric_and_boolean_dtypes + categorical_dtypes + ['Int64', 'boolean'])
-def test_targets_data_types_classification(data_type, automl_type, target_type):
-    if data_type == 'np' and target_type not in numeric_and_boolean_dtypes + categorical_dtypes:
+@pytest.mark.parametrize("target_type", ['int16', 'int32', 'int64', 'float16', 'float32', 'float64', 'bool', 'category', 'object', 'Int64', 'boolean'])
```
angela97lin, line 976: Still think this is a useful test to have around!

evalml/tests/data_checks_tests/test_target_leakage_data_check.py
```diff
@@ -96,15 +100,14 @@ def test_target_leakage_data_check_input_formats():
...
     # test X as ww.DataTable, y as ww.DataColumn
     assert leakage_check.validate(ww.DataTable(X), ww.DataColumn(y)) == expected_messages
 
-    #  test y as list
+    # test y as list
     assert leakage_check.validate(X, y.values) == expected_messages
 
     # test X as np.array
-    assert leakage_check.validate(X.to_numpy(), y) == {
+    assert leakage_check.validate(X.to_numpy().astype(float), y) == {
```
angela97lin, line 110: Type info is lost here, so need to manually convert to float: https://github.com/alteryx/woodwork/issues/456

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Agreed!

freddyaboulton (approved):
@angela97lin Looks great to me! Should these components be updated in this PR?

1. [SimpleImputer](https://github.com/alteryx/evalml/blob/1290_woodwork_type_selection/evalml/pipelines/components/transformers/imputers/simple_imputer.py#L52)
2. [OneHotEncoder](https://github.com/alteryx/evalml/blob/1290_woodwork_type_selection/evalml/pipelines/components/transformers/encoders/onehot_encoder.py#L81)
3. [IDColumnsDataCheck](https://github.com/alteryx/evalml/blob/1290_woodwork_type_selection/evalml/data_checks/id_columns_data_check.py#L68)

angela97lin:
@freddyaboulton Thanks for holding me accountable! 😁 I focused on removing use cases of the variables we created in gen_utils, and these were great spots to update too.

I wasn't sure how to update `IDColumnsDataCheck` without the 'exclude' parameter; I decided to go with including what I considered as reasonable columns to check for (int and categorical), as they cover our test case use cases. 

Recently updated this PR to address OHE, SimpleImputer and the ID col check so would love it if you could take a second look 😁

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -70,13 +70,10 @@ def transform(self, X, y=None):
...
-        X = X.fillna(value=np.nan)
-
+        ww_X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(ww_X.to_dataframe())
         # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
-        if (X.dtypes == bool).all():
+        if set(ww_X.logical_types.values()) == {ww.logical_types.Boolean} and X.isnull().sum().sum() == 0:
             return X
         X_null_dropped = X.copy()
         X_null_dropped.drop(self._all_null_cols, axis=1, errors='ignore', inplace=True)
```
angela97lin, line 79: Hm. The following is specific to the `category`-typed pandas columns passed to the imputer, so I'm not sure there's much that Woodwork can do to help out, given we have to pass a pandas DF / numpy array to the component object.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -75,20 +75,17 @@ def __init__(self,
...
     @staticmethod
     def _get_cat_cols(X):
-        """Get names of 'object' or 'categorical' columns in the DataFrame."""
-        obj_cols = []
-        for idx, dtype in enumerate(X.dtypes):
-            if dtype == np.object or pd.api.types.is_categorical_dtype(dtype):
-                obj_cols.append(X.columns.values[idx])
-        return obj_cols
+        """Get names of categorical columns in the input Woodwork DataTable."""
+        return list(X.select(include=['category']).columns.keys())
```
freddyaboulton, line 79: Nice! Looks much simpler now.

evalml/data_checks/id_columns_data_check.py
```diff
@@ -57,14 +57,14 @@ def validate(self, X, y=None):
...
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
-        col_names = [col for col in X.columns.tolist()]
+        col_names = [col for col in X.columns]
         cols_named_id = [col for col in col_names if (str(col).lower() == "id")]  # columns whose name is "id"
         id_cols = {col: 0.95 for col in cols_named_id}
 
-        non_id_types = ['float16', 'float32', 'float64', 'bool']
-        X = X.select_dtypes(exclude=non_id_types)
+        X = X.select(include=['Integer', 'Categorical'])
```
freddyaboulton, line 65: I think this makes sense!

freddyaboulton (approved):
@angela97lin Thanks for addressing those extra components/data check! Looks good to merge!

bchen1116 (approved):
LGTM!

evalml/data_checks/id_columns_data_check.py
dsherry: Agreed! Much easier to understand.

Does our documentation for this data check make it clear that this is the behavior?

evalml/utils/gen_utils.py
```diff
@@ -15,11 +15,10 @@
 
 logger = get_logger(__file__)
 
-numeric_dtypes = ['int16', 'int32', 'int64', 'float16', 'float32', 'float64']
-boolean = ['bool']
-numeric_and_boolean_dtypes = numeric_dtypes + boolean
-categorical_dtypes = ['object', 'category']
-datetime_dtypes = [np.datetime64]
+numeric_ww_types = [ww.logical_types.Integer, ww.logical_types.Double]
```
dsherry, line 18: @angela97lin can we use the "Numeric" semantic tag and delete this list?

evalml/utils/gen_utils.py
```diff
@@ -15,11 +15,10 @@
...
+boolean_ww = [ww.logical_types.Boolean]
+numeric_and_boolean_ww = numeric_ww_types + boolean_ww
+categorical_ww_types = [ww.logical_types.Categorical]
```
dsherry, line 21: Same suggestion, let's delete this list and use ww's "Categorical" semantic tag

evalml/utils/gen_utils.py
```diff
@@ -267,20 +266,21 @@ def safe_repr(value):
...
 
     Returns:
-        True if all the DataFrame columns are numeric and are not missing any values, False otherwise
+        True if all the DataTable columns are numeric and are not missing any values, False otherwise
     """
-    if df.isnull().any().any():
-        return False
-    for dtype in df.dtypes:
-        if dtype not in numeric_and_boolean_dtypes:
+    for ltype in dt.logical_types.values():
```
dsherry, line 278: @angela97lin can't we use the "numeric" semantic tag to simplify this, and ignore the logical types?

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -59,25 +59,24 @@ def validate(self, X, y):
...
-        y = _convert_woodwork_types_wrapper(y.to_series())
-
-        null_rows = y.isnull()
+        valid_target_types = numeric_and_boolean_ww + categorical_ww_types
+        if y.logical_type not in numeric_and_boolean_ww + categorical_ww_types:
+            messages["errors"].append(DataCheckError(message="Target is unsupported {} Woodwork type. Valid Woodwork logical types include: {}"
```
dsherry, line 64: I'm conflicted about whether or not we should mention woodwork here. On the one hand, we use woodwork under the hood, so people are going to have to learn about it if they want to work with how evalml handles typing. On the other hand, maybe this error message is easier to understand when it doesn't mention woodwork, since some users may not realize woodwork is in the mix until they see this. I guess the question is, what message makes it easiest for users to understand the problem and make a fix?

I think a good compromise is to say `"Target is unsupported {} type. Valid Woodwork logical types include: {}"`

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -70,13 +70,10 @@ def transform(self, X, y=None):
...
-        X = _convert_to_woodwork_structure(X)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        # Convert None to np.nan, since None cannot be properly handled
-        X = X.fillna(value=np.nan)
-
+        ww_X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(ww_X.to_dataframe())
         # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
-        if (X.dtypes == bool).all():
+        if set(ww_X.logical_types.values()) == {ww.logical_types.Boolean} and X.isnull().sum().sum() == 0:
```
dsherry, line 76: @angela97lin do we need to change this to use woodwork? If I remember right, we had to put this patch in place because the sklearn simple imputer misbehaves when all cols have pandas dtype bool, right? In that case I'd argue this should remain as a pandas dtype check.

If the new code here is functionally equivalent to the old code, fine either way, just thought I'd ask.

evalml/pipelines/regression_pipeline.py
```diff
@@ -26,9 +26,9 @@ def fit(self, X, y):
         """
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
+        if y.logical_type not in numeric_ww_types:
+            raise ValueError(f"Regression pipeline cannot handle targets with logical type: {y.logical_type}")
```
dsherry, line 30: Same question here, can we just check for the numeric semantic tag

dsherry (changes requested):
@angela97lin looks great! I left a few questions/suggestions

angela97lin:
@dsherry I went through and addressed your comments! I deleted the categorical_ww_type and others in place of just selecting the correct logical type/semantic tag, but I kept `numeric_and_boolean_ww` around. Reason is, the alternative would be to check if the semantic tag was numeric or the logic type was boolean. In the case where we also want to check categoricals, we would have to check if either 'numeric' or 'category' was in the semantic tags, or if the logic type was boolean. Keeping the list of logic types to check for seemed easier to understand, but lmk if you feel differently!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -59,25 +60,24 @@ def validate(self, X, y):
             raise ValueError("y cannot be None")
 
         y = _convert_to_woodwork_structure(y)
-        y = _convert_woodwork_types_wrapper(y.to_series())
-
-        null_rows = y.isnull()
+        is_supported_type = y.logical_type in numeric_and_boolean_ww + [ww.logical_types.Categorical]
+        if not is_supported_type:
+            messages["errors"].append(DataCheckError(message="Target is unsupported {} type. Valid Woodwork logical types include: {}"
```
dsherry, line 65: @angela97lin so we're doing one type-check here for all the problem types? Is there any value to having this vary per problem type, i.e. disallow `Categorical` for regression problems?
```python
supported_types = numeric_and_boolean_ww
if is_regression(problem_type):
    supported_types += [ww.logical_types.Categorical]
if not y.logical_type in supported_types:
    messages["errors"].append(DataCheckError(message="Target is unsupported type {}. Valid Woodwork logical types include: {}".format(y.logical_type, ", ".join([t.type_string for t in supported_types]))
```

Similarly we could disallow binary for multiclass problems

evalml/data_checks/outliers_data_check.py
```diff
@@ -54,9 +53,9 @@ def validate(self, X, y=None):
         }
 
         X = _convert_to_woodwork_structure(X)
+        X = X.select('numeric')
```
dsherry, line 56: Much cleaner!

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -58,19 +58,18 @@ def validate(self, X, y):
             "warnings": [],
             "errors": []
         }
-
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        if y.logical_type not in numeric_and_boolean_ww:
+            return messages
+        X_num = X.select(include=numeric_and_boolean_ww)
```
dsherry, line 65: Is there a reason to include boolean too? Perhaps this should just be `X.select('numeric')`

evalml/data_checks/invalid_targets_data_check.py
angela97lin: @dsherry I wanted to use this PR to strictly focus on converting our logic to the WW equivalent and not change too much behavior. You raise a great point though, and I previously already filed https://github.com/alteryx/evalml/issues/1548 to track this :)

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -75,20 +75,17 @@ def __init__(self,
...
 
     def fit(self, X, y=None):
         top_n = self.parameters['top_n']
         X = _convert_to_woodwork_structure(X)
+        if self.features_to_encode is None:
+            self.features_to_encode = self._get_cat_cols(X)
```
dsherry, line 85: Did moving this code change any behavior? Or was it just cleaner this way, but functionally equivalent?

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
```diff
@@ -77,8 +76,7 @@ def __init__(self, features_to_extract=None, encode_as_categories=False, random_
 
     def fit(self, X, y=None):
         X = _convert_to_woodwork_structure(X)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        self._date_time_col_names = X.select_dtypes(include=datetime_dtypes).columns
+        self._date_time_col_names = X.select(include=["datetime"]).columns
```
dsherry, line 79: Can't do `X.select('datetime').columns` ? Just curious because we're doing that elsewhere for numeric

evalml/pipelines/regression_pipeline.py
```diff
@@ -26,9 +25,9 @@ def fit(self, X, y):
         """
         X = _convert_to_woodwork_structure(X)
         y = _convert_to_woodwork_structure(y)
+        if "numeric" not in y.semantic_tags:
```
dsherry, line 28: Is there an `is_numeric` util method in woodwork which we can use? Or is this the recommended way to do this check? @gsheni

dsherry (commented):
@angela97lin looks great! Sorry I didn't get to this before merge. I left some questions.

evalml/data_checks/invalid_targets_data_check.py
dsherry: Amazing :)

evalml/data_checks/target_leakage_data_check.py
angela97lin: I guess this should just be based on whether or not we want to / it makes sense to check for label leakage if the target is boolean :)

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: It shouldn't change behavior; the reason for moving this is because we want to take advantage of Woodwork so we want to pass in X as a Woodwork structure before converting it later :)

evalml/pipelines/regression_pipeline.py
gsheni: Yes, there is (thought it wasn't meant to be user facing).
https://github.com/alteryx/woodwork/blob/main/woodwork/datacolumn.py#L318

PR #1554, opened 2020-12-14
Partial dependence for multiclass

### LINKED ISSUE #1404: Fix partial dependence for multiclass
Not sure if this is working at the moment. This issue tracks trying our partial dependence graphing out on multiclass and determining what if anything we need to change in our code in order to support it.

Sklearn's docs mention setting `target` in [their graph method](https://scikit-learn.org/stable/modules/partial_dependence.html#partial-dependence-plots).

freddyaboulton:
### Pull Request Description
Fixes #1404 

### Example on wine dataset
image

image

-----

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -476,8 +476,7 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
     return data
 
 
-
-def graph_partial_dependence(pipeline, X, feature, grid_resolution=100):
+def graph_partial_dependence(pipeline, X, feature, class_label=None, grid_resolution=100):
     """Create an one-way partial dependence plot.
 
     Arguments:
@@ -487,6 +486,10 @@ def graph_partial_dependence(pipeline, X, feature, grid_resolution=100):
         feature (int, string): The target feature for which to create the partial dependence plot for.
             If feature is an int, it must be the index of the feature to use.
             If feature is a string, it must be a valid column name in X.
+        class_label (string, None): Name of class to plot for multiclass problems. If None, will plot
+            the partial dependence for each class. This argument does change behavior for regression or binary
+            classification pipelines. For binary classification, the partial dependence for the positive label will
+            always be displayed.
 
     Returns:
         pd.DataFrame: pd.DataFrame with averaged predictions for all points in the grid averaged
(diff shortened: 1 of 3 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -484,19 +498,39 @@ def graph_partial_dependence(pipeline, X, feature, grid_resolution=100):
...
+        class_labels = [class_label] if class_label is not None else pipeline.classes_
+        _subplots = import_or_raise("plotly.subplots", error_msg="Cannot find dependency plotly.graph_objects")
+        fig = _subplots.make_subplots(rows=1, cols=len(class_labels), subplot_titles=class_labels,
+                                      shared_xaxes=True, shared_yaxes=True)
+        for i, label in enumerate(class_labels):
+            # Plotly trace indexing begins at 1 so we add 1 to i
+            fig.add_trace(_go.Scatter(x=part_dep.loc[part_dep.class_label == label, 'feature_values'],
+                                      y=part_dep.loc[part_dep.class_label == label, 'partial_dependence'],
+                                      line=dict(width=3)),
+                          row=1, col=i + 1)
```
ParthivNaresh, line 523: @freddyaboulton  Great! Also if the dataset has a larger number of labels, it might be difficult for the user to see all the partial dependency plots in one row.
Maybe:
`_subplots.make_subplots(rows=(len(class_labels)+1) // 2, cols=2 ...etc)`
and then in line 523:
`row=(i+2) // 2` and `col=(i%2) + 1`.
Not really a big deal either way

evalml/model_understanding/graphs.py
```diff
@@ -461,11 +461,21 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
-    return pd.DataFrame({"feature_values": values[0],
-                         "partial_dependence": avg_pred[0]})
+    classes = None
+    if isinstance(pipeline, evalml.pipelines.BinaryClassificationPipeline):
+        classes = [pipeline.classes_[1]]
+    elif isinstance(pipeline, evalml.pipelines.MulticlassClassificationPipeline):
+        classes = pipeline.classes_
 
+    data = pd.DataFrame({"feature_values": np.tile(values[0], avg_pred.shape[0]),
+                         "partial_dependence": np.concatenate([pred for pred in avg_pred])})
```
ParthivNaresh, line 471: Neat!

ParthivNaresh (approved):
@freddyaboulton Everything looks great!

evalml/model_understanding/graphs.py
freddyaboulton: Sounds good! I think the only thing missing was that we should only use two columns in the case where `class_label=None` or else there would be an empty second column in the plot.

ParthivNaresh: Excellent!

evalml/model_understanding/graphs.py
```diff
@@ -462,11 +462,21 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
+    if classes is not None:
+        data['class_label'] = np.repeat(classes, len(values[0]))
```
angela97lin, line 474: Nit-pick: Since we're changing the output to return this new field in the DF, could be good to update this docstring too?

evalml/model_understanding/graphs.py
```diff
@@ -476,6 +486,10 @@ def graph_partial_dependence(pipeline, X, feature, grid_resolution=100):
         feature (int, string): The target feature for which to create the partial dependence plot for.
             If feature is an int, it must be the index of the feature to use.
             If feature is a string, it must be a valid column name in X.
+        class_label (string, None): Name of class to plot for multiclass problems. If None, will plot
```
angela97lin, line 489: Alternatively:
```
class_label (string, optional): Name of class to plot for multiclass problems. If None, will plot...; Defaults to None.

**Review by angela97lin — APPROVED** [review]
Left two nit-picky things, but otherwise LGTM!! This is great stuff 😁

**Comment by angela97lin** [comment]
Looks great locally too; just one nit-picky suggestion, if it's not too difficult: is it possible to update the trace names 🤔 
image

File: evalml/model_understanding/graphs.py (hunk 3 above)
Comment (freddyaboulton) [reply evalml/model_understanding/graphs.py:479]: Done!

File: evalml/model_understanding/graphs.py (hunk 4 above)
Comment (freddyaboulton) [reply evalml/model_understanding/graphs.py]: Done!

**Comment by freddyaboulton** [comment]
@angela97lin good catch! I just pushed this up and updated the tests for multiclass/not multiclass.

image

File: evalml/model_understanding/graphs.py [hunk 5]
```diff
@@ -485,19 +504,47 @@ def graph_partial_dependence(pipeline, X, feature, grid_resolution=100):
...
+        # Don't specify share_xaxis and share_yaxis so that we get tickmarks in each subplot
+        fig = _subplots.make_subplots(rows=rows, cols=cols, subplot_titles=class_labels)
+        for i, label in enumerate(class_labels):
+
+            # Plotly trace indexing begins at 1 so we add 1 to i
+            fig.add_trace(_go.Scatter(x=part_dep.loc[part_dep.class_label == label, 'feature_values'],
+                                      y=part_dep.loc[part_dep.class_label == label, 'partial_dependence'],
+                                      line=dict(width=3),
+                                      name=label),
+                          row=(i + 2) // 2, col=(i % 2) + 1)
```
bchen1116, line 536: Nice!

bchen1116 (approved):
LGTM! One thing I think would be cool to have is to allow users to pass in a list of classes to create these plots for in multiclass scenarios. For instance, I, as a user, decided to create a multiclass pipeline with 20 possible target classes, I might want to plot partial dependencies for a subset of the classes only, rather than for 1 or all. Certainly not blocking, but wanted to bring that suggestion up for possible discussion.

freddyaboulton:
Great suggestion @bchen1116 ! I'm on board but let's continue the discussion to #1565 since this issue/PR tracks returning the partial dependence for all the classes as opposed to just the first one in multiclass problems.

PR #1555, opened 2020-12-14
Allow users to set feature types without having to learn about woodwork directly

### LINKED ISSUE #1545: Allow users to set feature types without having to learn about woodwork directly
@angela97lin and I had a great conversation about this earlier today. I put our notes here.

**Background**
The work done for #1229 so far has made it so that we use woodwork under the hood to a) infer feature types from the inputted pandas DF, and b) allow the components, pipelines, automl code and all the rest to use those feature types in their implementations, instead of using the pandas DF dtypes. This allows users to have a clear picture of how each of their features will be treated by evalml, and makes it easy for them to tweak this per-feature as needed.

**Challenge**
We'd like to make it as easy as possible for users to have control over how each of their features will be treated by evalml. To that end, we'd like to make it so that users can use automl and the rest of evalml without being required to import woodwork and learn any details about it.

**Next Steps**
* Add a standardization helper method which accepts feature_types and outputs woodwork types. This reduces the knowledge evalml users are required to know about woodwork down to a single method call. Something like this:
```python
X, y = infer_feature_types(X, y, feature_types={...})
automl.search(X, y)
pipeline = automl.get_pipeline(42)
pipeline.fit(X, y)
```
* Update docs:
    * On start page (https://evalml.alteryx.com/en/stable/start.html), show usage with and without the type inference helper. Mention the feature_types input and that it allows you to override the type inference. Mention that we use these same inferred types under the hood, and that that's done via woodwork.
    * On the automl user guide page (https://evalml.alteryx.com/en/latest/user_guide/automl.html#AutoML-in-EvalML), add a similar section, with more detail. (I wrote up a draft [here in the doc](https://alteryx.quip.com/4DepAA8DuBWu/EvalML-Woodwork-API-Suggestion#JKJACAbv49T))

@gsheni @kmax12 FYI since we've spoken about this recently (and @gsheni wrote up a doc for this in Nov!)

angela97lin:
Closes #1545 

I currently have a function that will accept different inputs for `feature_types` depending on the input. If the input is 2d, we expect a dictionary mapping col name to woodwork logical type string. If the input is 1d, we expect a woodwork logical type or string equivalent.

This seems the easiest to users, but I'm not sure this is the cleanest impl. Should this be two separate methods instead? Should we enforce a dictionary for 1d too, and force users to pass the name of the 1d input?

Would love thoughts before pushing forward on adding more tests :d

Docs here:https://evalml.alteryx.com/en/1545_infer_feature_types/user_guide/automl.html#AutoML-in-EvalML


jeremyliweishih (approved):
I think this looks good! I like the example Dylan gave in the issue and it should be good for user experience to keep it to one method. Can you add something in the docs to show how to use this? Might also want to add this to the API reference.

dsherry:
@angela97lin I'm excited to review this! Could we please also add this to the docs? I think it should go on the start page, and in the automl guide :)

dsherry:
For others reading, here's the example usage from the issue:
```python
X, y = infer_feature_types(X, y, feature_types={...})
automl.search(X, y)
pipeline = automl.get_pipeline(42)
pipeline.fit(X, y)
```

bchen1116 (approved):
LGTM! I do think it would be nice if the types (string vs dict) were the same for 1D and 2D datatables for the sake of consistency, but in terms of ease, I think this is easier for the user. I'm good with this implementation!

angela97lin:
@dsherry I see that you were considering adding this util method to the start page as well as the automl guide. I've added it to the automl guide, but when trying to add to the start page, wondered if it would make it too clunky / if it's necessary since we want the start page to be the most minimal example possible. Our current data set for the start page / automl guide used the breast cancer data set, which is all numeric, so I changed it to the fraud data set. That being said, I would love your thoughts!

angela97lin:
Going to merge this in, since it seems like we're okay with the API, so that it's available in the next release. If there are any further comments about it / any improvements we want to make to the documents, I'd be happy to put up another PR 😁

PR #1561, opened 2020-12-15
Fix readthedocs build

### LINKED ISSUE #1520: RTD builds fail with numpy version RuntimeError
**Problem**
All RTD builds fail with this error during ipynb execution:
```python
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
```
[Here's a recent failure](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/500104/)

Marked as blocker because we can't release again until we can build docs.

dsherry:
Fix #1520


.readthedocs.yaml
```diff
@@ -19,8 +19,7 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.7
+  version: 3.8
```
dsherry, line 22: I don't think this was necessary but it seemed like a good idea

.circleci/config.yml
```diff
@@ -262,7 +268,7 @@ jobs:
       python_version: "3.8"
     steps:
       - checkout
-      - install_dependencies_dev
+      - install_dependencies_docs
```
dsherry, line 271: Did this just so that the circleci `build_docs` job would have the same deps installed by us as on RTD.

docs-requirements.txt
```diff
@@ -0,0 +1,4 @@
+pydata-sphinx-theme>=0.3.1
+Sphinx>=2.0.1
+nbconvert>=5.5.0
+nbsphinx>=0.4.2
```
dsherry, line 4: Separated this out from `dev-requirements.txt` so that `.readthedocs.yml` can install these deps without having to install the others.

I didn't confirm that this change is necessary in order to fix the bug, it just felt like a good way to reduce complexity and speed up the build, to only install the minimum required deps in each RTD build

.readthedocs.yaml
```diff
@@ -19,8 +19,9 @@ formats:
...
   install:
+    - requirements: docs/readthedocs-requirements.txt
```
dsherry, line 24: This line is the fix: install `setuptools>=45.1.0` before installing the docs requirements (sphinx etc) and our package itself.

dev-requirements.txt
```diff
@@ -1,10 +1,6 @@
 -r requirements.txt
 -r test-requirements.txt
+-r docs-requirements.txt
```
dsherry, line 3: I just moved the deleted lines into `docs-requirements.txt` below

Makefile
```diff
@@ -46,3 +46,8 @@ installdeps-test:
 installdeps-dev:
 	pip install -e . -q
 	pip install -r dev-requirements.txt -q
+
+.PHONY: installdeps-docs
+installdeps-docs:
+	pip install -e . -q
+	pip install -r docs-requirements.txt -q
```
dsherry, line 53: This is used by the circleci `build_docs` code [at the top](https://github.com/alteryx/evalml/pull/1561/files#r543735032) and isn't part of the bugfix itself.

.circleci/config.yml
freddyaboulton: If the goal is to have the same deps as RTD, shouldn't we install `setuptools>=45.1.0` before installing our package?

.readthedocs.yaml
freddyaboulton: I think so too!

freddyaboulton (approved):
@dsherry Heroic! Thanks for fixing this before our next release. Please let us know what RTD says about this. I'm curious why they were installing an older version of setuptools.

.circleci/config.yml
dsherry: @freddyaboulton I'd say my goal here was to have the circleci job have deps which are *closer* to whats used on the RTD docker image.

If we really wanted to go the extra mile, I think we should write a job which duplicates the RTD build in its entirety, including the docker image. We could write that automation ourselves, or we could see if there's [something in the community](https://github.com/dfm/rtds-action) which we can use.

Given that we're considering migrating to github actions early next year, I think we should punt on this question. But great thinking 😁

jeremyliweishih (approved):
nice!

dsherry:
RTD came back [green](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/508663/) 💹🥬🍏♻️😂, merging!

PR #1568, opened 2020-12-17
Make it easy to customize automl search data splitter

### LINKED ISSUE #1567: Make it easy to customize automl search data splitter
**Problem**
Suppose I want to change the number of folds used in automl search, or some other aspect of data splitting. `AutoMLSearch` has a `data_split` input which allows users to provide a custom data splitter.

This satisfies the immediate need for users to be able to reconfigure evalml's CV. However, it makes it more likely users could do something screwy, like forget to set `shuffle=True` (which can lead to ML performance issues) or forget to set `random_state` (which can lead to inconsistent results run-over-run).

What if someone wants to preserve the automl defaults, like using TV split for large datasets instead of CV, but make modifications to those defaults like the number of CV folds?

**Proposal**
Let's find a way to give users access to the default automl data splits, such that they can override specific parameters like `n_folds` but still benefit from our heuristics, like the large data threshold as an example. I think we should define a `make_data_split` util method which does this.

dsherry:
Fix #1567 

Usage: configure automl search to use a different number of folds (splits) in CV
```python
random_state = 42
data_split = make_data_splitter(X, y, problem_type, n_splits=5, random_state=random_state)
automl = AutoMLSearch(problem_type=problem_type, data_split=data_split, random_state=random_state, ...)
...
```

Usage: disable shuffling
```python
random_state = 42
data_split = make_data_splitter(X, y, problem_type, shuffle=False, random_state=random_state)
automl = AutoMLSearch(problem_type=problem_type, data_split=data_split, random_state=random_state, ...)
...
```

I suppose we could simply add `n_splits` and `shuffle` to `AutoMLSearch.__init__`, but I wanted to keep us flexible on this a while longer. We'll be adding more heuristics to the data splitting logic here at some point and its nice exposing that.

Hmm, perhaps we need to rename "data_split" to "data_splitter" as well.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,7 +17,7 @@ Release Notes
         * Added more information to users about ensembling behavior in ``AutoMLSearch`` :pr:`1527`
         * Add woodwork support for more utility and graph methods :pr:`1544`
         * Changed ``DateTimeFeaturizer`` to encode features as int :pr:`1479`
-        * Added Linear Descriminant Analysis transformer for dimensionality reduction :pr:`1331`
+        * Added Linear Discriminant Analysis transformer for dimensionality reduction :pr:`1331`
         * Added multiclass support for ``partial_dependence`` and ``graph_partial_dependence`` :pr:`1554`
         * Added ``TimeSeriesBinaryClassificationPipeline`` and ``TimeSeriesMulticlassClassificationPipeline`` classes :pr:`1528`
         * Added ``make_data_splitter`` method for easier automl data split customization :pr:`1568`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,7 +17,7 @@ Release Notes
         * Added more information to users about ensembling behavior in ``AutoMLSearch`` :pr:`1527`
         * Add woodwork support for more utility and graph methods :pr:`1544`
         * Changed ``DateTimeFeaturizer`` to encode features as int :pr:`1479`
-        * Added ``LDA`` component for dimensionality reduction :pr:`1331`
+        * Added Linear Descriminant Analysis transformer for dimensionality reduction :pr:`1331`
         * Added multiclass support for ``partial_dependence`` and ``graph_partial_dependence`` :pr:`1554`
         * Added ``TimeSeriesBinaryClassificationPipeline`` and ``TimeSeriesMulticlassClassificationPipeline`` classes :pr:`1528`
         * Added ``make_data_splitter`` method for easier automl data split customization :pr:`1568`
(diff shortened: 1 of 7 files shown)


bchen1116 (approved):
LGTM! 

This change means we should file a separate issue to update our implementation for `class_imbalance_data_check`, specifically the `cv_folds` arg [here](https://github.com/alteryx/evalml/blob/a48ebfb64a16688ee95047e53f4aced54755d23b/evalml/data_checks/class_imbalance_data_check.py#L17). It defaults to 3 since that's the default data split `n_folds`, but with this change, I believe we should update AutoMLSearch to pass in the required param. Not blocking this PR though!

docs/source/release_notes.rst
```diff
@@ -16,8 +16,9 @@ Release Notes
         * Added more information to users about ensembling behavior in ``AutoMLSearch`` :pr:`1527`
         * Add woodwork support for more utility and graph methods :pr:`1544`
         * Changed ``DateTimeFeaturizer`` to encode features as int :pr:`1479`
-        * Added `Linear Discriminant Analysis Transformer` component for dimensionality reduction :pr:`1331`
+        * Added ``LDA`` component for dimensionality reduction :pr:`1331`
```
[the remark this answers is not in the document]
dsherry: @bchen1116 that's a good point. I changed it for formatting purposes, because it used single-ticks instead of double-ticks. But instead I'll change the text back and remove the ticks :)

dsherry:
@bchen1116 excellent point RE the class imbalance data check! I just filed as #1570

docs/source/release_notes.rst
```diff
@@ -16,8 +16,9 @@ Release Notes
         * Added more information to users about ensembling behavior in ``AutoMLSearch`` :pr:`1527`
         * Add woodwork support for more utility and graph methods :pr:`1544`
         * Changed ``DateTimeFeaturizer`` to encode features as int :pr:`1479`
-        * Added `Linear Discriminant Analysis Transformer` component for dimensionality reduction :pr:`1331`
+        * Added Linear Descriminant Analysis transformer for dimensionality reduction :pr:`1331`
```
freddyaboulton, line 19: nit-pick: Typo?

freddyaboulton (approved):
@dsherry Looks great!

docs/source/release_notes.rst
dsherry: Ooops, thanks :)

evalml/automl/utils.py
```diff
@@ -17,3 +27,34 @@ def get_default_primary_search_objective(problem_type):
...
+
+    Arguments:
+        X (pd.DataFrame, ww.DataTable): The input training data of shape [n_samples, n_features].
+        y (pd.Series, ww.DataColumn): The target training data of length [n_samples].
+        problem_type (ProblemType): the type of machine learning problem.
+        problem_configuration (dict, None): Additional parameters needed to configure the search. For example,
+            in time series problems, values should be passed in for the gap and max_delay variables.
+        n_splits (int, None): the number of CV splits, if applicable.
+        shuffle (bool): whether or not to shuffle the data before splitting, if applicable.
+        random_state (int, np.random.RandomState): The random seed/state. Defaults to 0.
```
angela97lin, line 43: Maybe mention default value here for n_splits/shuffle/problem_configuration?

angela97lin (approved):
LGTM! 😁

Issue #785, opened 2020-05-20
Add data check to check for negative target values


angela97lin:
As raised in #728, MSLE/RMSLE is only valid when both the target and the predictions are guaranteed to be non-negative. We should add a data check that checks for negative target values and returns a warning if that's the case.

dsherry:
I'll add another case here: if any values are 0 for MAPE (timeseries), we should raise an error.

The previous comment is a bit out of date--I think we should add these checks to the `InvalidTargetDataCheck`.

PR #1574, opened 2020-12-17
1521 RMSLE MSLE MAPE Primary Objective

### LINKED ISSUE #1521: Allow RMSLE, MSLE, MAPE as the primary objective when the target supports it
Currently, users cannot specify RMSLE, MSLE by their string name if they are specified as the [primary objective](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L260). 

The issue these objectives are not allowed as the primary objective is because they are only valid for certain datasets. In this case, y > 0 for all y. We should allow them to be primary objectives but we should validate that the targets during `search`. 

Perhaps this can be incorporated into `InvalidTargetDataCheck`?

[Context](https://github.com/alteryx/evalml/pull/1510#discussion_r537754091)

ParthivNaresh:
Fixes #1521 

-----


dsherry (changes requested):
@ParthivNaresh looking good! I had some details we should sort out in the impl, plus a couple recommendations for the unit tests.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -407,6 +408,69 @@ def test_automl_bad_data_check_parameter_type():
...
+                                                      objective=None)])
+    assert automl.data_check_results == {"warnings": [], "errors": []}
+
+
+@pytest.mark.parametrize("problem_type", [ProblemTypes.REGRESSION, ProblemTypes.BINARY, ProblemTypes.MULTICLASS])
+def test_automl_invalid_objectives_for_mape(problem_type):
+    X = pd.DataFrame([1, 2, 3], [2, 3, 4])
+    y = pd.Series([1, 2, 3])
+    with pytest.raises(ValueError, match="Given objective Mean Absolute Percentage Error is not compatible"):
+        AutoMLSearch(X, y, problem_type=problem_type, max_iterations=1, objective='Mean Absolute Percentage Error')
```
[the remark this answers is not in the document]
ParthivNaresh: Gotcha, I'll fold it in with `test_main_objective_problem_type_mismatch`

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -556,10 +556,12 @@ def test_get_permutation_importance_regression(X_y_regression, linear_regression
     pipeline = linear_regression_pipeline_class(parameters={"Linear Regressor": {"n_jobs": 1}},
                                                 random_state=np.random.RandomState(42))
     pipeline.fit(X, y)
+
     for objective in regression_core_objectives:
-        permutation_importance = calculate_permutation_importance(pipeline, X, y, objective)
-        assert list(permutation_importance.columns) == ["feature", "importance"]
-        assert not permutation_importance.isnull().all().all()
+        if objective.name not in ['Root Mean Squared Log Error', 'Mean Squared Log Error']:
```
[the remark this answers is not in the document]
ParthivNaresh: It was failing due to RMSLE and MSLE not working with the negative labels in X_y_regression but I just realized that the test case was dealing with `calculate_permutation_importance`, so it's fine for me to provide a simple nonnegative label set, thanks!

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -193,3 +196,78 @@ def test_invalid_target_data_check_n_unique():
                                   message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_UNIQUE_VALUES,
                                   details={"target_values": unique_values}).to_dict()]
     }
+
+
+@pytest.mark.parametrize("objective", ['Root Mean Squared Log Error', 'Mean Squared Log Error', 'Mean Absolute Percentage Error'])
```
freddyaboulton, line 201: Can you add some test cases where we pass an instance of the objective as opposed to the name? That's technically allowed so we should cover those cases!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -1,30 +1,32 @@
...
+    numeric_and_boolean_dtypes,
+    numeric_dtypes
 )
 
 
 class InvalidTargetDataCheck(DataCheck):
     """Checks if the target data contains missing or invalid values."""
 
-    def __init__(self, problem_type, n_unique=100):
+    def __init__(self, problem_type, objective, n_unique=100):
```
freddyaboulton, line 21: Can you please add objective to the docstring?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -412,6 +412,23 @@ def test_automl_bad_data_check_parameter_type():
...
+        return {"warnings": [], "errors": []}
+
+
+@pytest.mark.parametrize("data_checks", [DataChecks([MockDataCheckObjective],
+                                                    data_check_params={"MockDataCheckObjective": {"objective": 'R2'}})])
+def test_automl_passes_correct_objective_name_to_data_check(data_checks, X_y_regression):
+    X, y = X_y_regression
+    automl = AutoMLSearch(X, y, problem_type=ProblemTypes.REGRESSION)
+    automl.search(data_checks=data_checks)
+    assert automl._validate_data_checks(data_checks).data_checks[0].objective_name == MockDataCheckObjective("R2").objective_name
```
freddyaboulton, line 429: Can we also test that `DefaultDataChecks` is initialized with the right objective name if `data_checks='auto'`?

freddyaboulton (approved):
@ParthivNaresh Looks good! I have some small suggestions/comments :)

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -104,4 +107,12 @@ def validate(self, X, y):
                                                              message_code=DataCheckMessageCode.TARGET_BINARY_INVALID_VALUES,
                                                              details={"target_values": unique_values}).to_dict())
 
+        any_neg = not (y > 0).all() if y.dtype in numeric_dtypes else None
+        if any_neg and (self.objective_name in ['Root Mean Squared Log Error', 'Mean Squared Log Error', 'Mean Absolute Percentage Error']):
+            details = {"Count of offending values": sum(val <= 0 for val in y.values.flatten())}
+            messages["errors"].append(DataCheckError(message=f"Target has non-positive values which is not supported for {self.objective_name}",
+                                                     data_check_name=self.name,
+                                                     message_code=DataCheckMessageCode.TARGET_INCOMPATIBLE_OBJECTIVE,
+                                                     details=details).to_dict())
```
dsherry, line 116: @ParthivNaresh will this close #785? Looks like it will :)

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -104,4 +107,12 @@ def validate(self, X, y):
                                                              message_code=DataCheckMessageCode.TARGET_BINARY_INVALID_VALUES,
                                                              details={"target_values": unique_values}).to_dict())
 
+        any_neg = not (y > 0).all() if y.dtype in numeric_dtypes else None
+        if any_neg and (self.objective_name in ['Root Mean Squared Log Error', 'Mean Squared Log Error', 'Mean Absolute Percentage Error']):
```
dsherry, line 111: Perhaps instead of having a list here, we should add a `positive_only` static attribute to `ObjectiveBase`, default `False`, set to `True` for these objectives. Then here you could do
```python
if any_neg and handle_objective(self.objective_name).positive_only:
```

And we'd get future compatibility for other similar objectives we add!

What do you think? Would be fine to do this in a separate PR and/or file it.

dsherry (approved):
@ParthivNaresh looks great! I left one suggestion and one question about a related issue.

evalml/data_checks/invalid_targets_data_check.py
ParthivNaresh: @dsherry I absolutely do believe it will!

dsherry: Very cool! You can update the PR description to say "Fix #1521 fix #785" and merging will close both :)

evalml/data_checks/invalid_targets_data_check.py
ParthivNaresh: @dsherry Good idea, I'll include it here

evalml/objectives/objective_base.py
```diff
@@ -10,6 +10,8 @@
 class ObjectiveBase(ABC):
     """Base class for all objectives."""
 
+    positive_only = False
```
dsherry, line 13: @ParthivNaresh sweet!

Now that I'm looking at this next to the other attributes we've defined for `ObjectiveBase` I have a few thoughts:
* Let's follow a similar pattern to the other `ObjectiveBase` class attributes. Perhaps we don't need to make this `abstract` since the vast majority of objectives won't change it from the default.
* Whatever we settle on let's make sure a docstring appears in the docs both for a) any objectives where we set this to True and b) all other objectives
* Style nit-pick: I think we should make this the last attribute listed in the file here rather than the first, because its less important than the others.

How about this:
```python
@property
@classmethod
def positive_only(cls):
    """If True, this objective is only valid for positive data. Default False."""
    return False
```

and then in standard metrics for RMSLE/MSLE/MAPE:
```python
@property
@classmethod
def positive_only(cls):
    """If True, this objective is only valid for positive data. Default False."""
    return True
```

PR #1577, opened 2020-12-17
Update docs to include AutoMLSearch callbacks

### LINKED ISSUE #1407: Update docs to include AutoMLSearch callbacks
We should update our docs to talk about our callback arguments for `AutoMLSearch` (start_iteration_callback, add_result_callback, error_callback with #1403)

angela97lin:
Closes #1407
https://evalml.alteryx.com/en/1407_callback_docs/user_guide/automl.html#Callback-functions


jeremyliweishih (approved):
Looks good! Can you add examples for these callbacks? I think it would be helpful for users since they need to define methods that take in certain dictionary formats etc.

bchen1116 (approved):
LGTM! Agreed with Jeremy, it would be really helpful if you can include short examples for each so that users know what the expected format (and potentially expected output) looks like.

dsherry (commented):
Looks great!

PR #1578, opened 2020-12-17
Include target name for pipeline predict

### LINKED ISSUE #1493: Pipeline/Estimator predict: include the target name
If the input data included a name for the `y` input to `fit`, we should include that name in the output of `predict`.

RE discussion with @gsheni @kmax12 @angela97lin

angela97lin:
Closes #1493


evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -271,6 +271,13 @@ def test_pad_with_nans(data, num_to_pad, expected):
     _check_equality(padded, expected)
 
 
+def test_pad_with_nans_with_series_name():
```
freddyaboulton, line 274: Thanks for adding this test. I think this gives us coverage in the case where gap/max_delay != 0

evalml/pipelines/pipeline_base.py
```diff
@@ -477,7 +482,7 @@ def __eq__(self, other):
         random_state_eq = check_random_state_equality(self.random_state, other.random_state)
         if not random_state_eq:
             return False
-        attributes_to_check = ['parameters', '_is_fitted', 'component_graph', 'input_feature_names']
+        attributes_to_check = ['parameters', '_is_fitted', 'component_graph', 'input_feature_names', 'input_target_name']
```
freddyaboulton, line 485: I think it would be helpful to add a case where the input target name is different in  `test_pipeline_equality`

freddyaboulton (approved):
@angela97lin I think this looks good! I think we should add `input_target_name` to the api ref (and `input_feature_names` as well?) because it's currently not documented anywhere.

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -57,6 +57,7 @@ def fit(self, X, y):
         X, y = self._convert_to_woodwork(X, y)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_woodwork_types_wrapper(y.to_series())
+        self.input_target_name = y.name
```
chukarsten, line 60: So I've been looking at this for a bit now and I've been stressing over why grand-child class TimeSeriesClassificationPipeline and TimeSeriesRegressionPipeline don't just inherit the renaming code from the grand-parent PipelineBase class.  I guess it's because the grand-children don't invoke a private ._fit function, but I might be missing something.  Regardless, I think this code seems to accomplish what it set out to accomplish but I also think it might open up a discussion on bringing the TimeSeries pipes into compliance with the existing paradigm or deciding on.

Just to throw something out there, does it make sense for ._fit() to do it's own special fitting, be it with the component_graph, or the encoding and computing features and fitting and then have .fit() in the base class capture the name of the target and be responsible for the naming?  It seems that the behavior of .fit() should have the commonalities between the derivative classes while ._fit() should contain variations relative to that specific pipeline?

chukarsten (approved):

evalml/pipelines/time_series_classification_pipelines.py
angela97lin: @chukarsten Yeah, I hear ya, and I think your intuition is right in that our pipeline code is a bit messy. Just deleted some stuff which hopefully makes this more clear but to summarize, we currently have:
- `fit() is an abstract method of PipelineBase
- `_fit()` used by Classification/RegressionPipeline classes, where I added `input_feature_names`
- `_compute_features_during_fit()` is used by the TimeSeriesPipeline classes, where I also added `input_feature_names`

PR #1583, opened 2020-12-18
Component iteration in pipelines

### LINKED ISSUE #1571: Support iterating over pipelines for component iteration
As [discussed](https://github.com/alteryx/evalml/pull/1543#discussion_r544446628) in PR #1543, there is a slight breaking change introduced there where you can no longer iterate through component instances with `Pipeline.component_graph`. @freddyaboulton suggested we support iterating through the components using the pipeline directly, mirroring what the private `ComponentGraph` class does now.

This will involve implementing a pipeline iterator and updating the current code to reflect this change, as well as updating the docs to match.

eccabay:
Closes #1571 

~Obviously I'm not going to get this in before I leave today, but I had the time so I did what I could~

### Context since this is an old branch:
This PR addresses a breaking change made in #1543. Prior to #1543, the documented way to iterate through component instances was `[c for c in pipeline.component_graph]`. However, #1543 made it so that the `component_graph` attribute only lists the classes, not the instances. Rather than having users iterate through the private `_component_graph` attribute, we add a "public" accessor by making the pipelines iterable.

This PR accomplishes that by:
1. Adding `__iter__` and `__next__` method to pipelines (as opposed to relying on `__getitem__`)
2. Updating how we iterate through pipelines in code and docs. Let me know if I missed a spot hehe

### DIFF
--- a/evalml/tests/pipeline_tests/test_pipelines.py
+++ b/evalml/tests/pipeline_tests/test_pipelines.py
@@ -27,6 +27,7 @@ from evalml.pipelines.components import (
     DateTimeFeaturizer,
     DelayedFeatureTransformer,
     DropNullColumns,
+    ElasticNetClassifier,
     Estimator,
     Imputer,
     LinearRegressor,
@@ -2339,6 +2340,7 @@ def test_generate_code_pipeline_custom():
     assert pipeline == expected_code
 
 
+<<<<<<< HEAD
 @pytest.mark.parametrize("problem_type", [ProblemTypes.BINARY, ProblemTypes.MULTICLASS, ProblemTypes.REGRESSION,
                                           ProblemTypes.TIME_SERIES_REGRESSION, ProblemTypes.TIME_SERIES_BINARY, ProblemTypes.TIME_SERIES_MULTICLASS])
 def test_predict_has_input_target_name(problem_type, X_y_binary, X_y_multi, X_y_regression, ts_data,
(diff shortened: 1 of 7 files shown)


jeremyliweishih (approved):
really cool! looks good to me.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2323,3 +2324,93 @@ def test_predict_has_input_target_name(problem_type, X_y_binary, X_y_multi, X_y_
...
+    order_again_params = [c for c in pipeline]
+
+    assert order_params == expected_order_params
+    assert order_again_params == expected_order_params
+
+
+def test_linear_getitem(logistic_regression_binary_pipeline_class):
+    pipeline = logistic_regression_binary_pipeline_class({'One Hot Encoder': {'top_n': 4}})
+
+    assert pipeline[0] == Imputer()
```
angela97lin, line 2372: Amazing! Just to confirm, giving an invalid index throws... an IndexError? Or ValueError? Just curious since get_component throws a ValueError :p

angela97lin (approved):
Looks great! Tests are solid, just curious what happens if index is not valid :P

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: Great question, it throws an IndexError!

image

bchen1116 (approved):
This is super cool! LGTM!

PR #1585, opened 2020-12-18
Revert "1535 save graphs to file"

ParthivNaresh:
Reverts alteryx/evalml#1550

Reverting due to issues with Kaleido.


docs/source/release_notes.rst
```diff
@@ -3,7 +3,6 @@ Release Notes
 
 **Future Releases**
     * Enhancements
-        * Added ``save_plot`` that allows for saving of figures from different backends :pr:`1550`
```
dsherry, line 6: @ParthivNaresh unfortunately our release notes CI check isn't smart enough to know this is a revert 😆 

I'd suggest you move this entry to "Fixes", add a second "Revert ..." entry underneath which references PR #1550 and you should be all set. Then when you merge the PR with the fix, you can delete both of those entries

dsherry (approved):
@ParthivNaresh 👍 🚢

PR #1588, opened 2020-12-21
Add save_plot to save graphs to file

ParthivNaresh:
### Pull Request Description
Fixes and updates #1535

Created a new PR rather than revert the revert I made.
Updated to default the graphviz format to png. Wasn't an issue in the local build because graphviz.backend.FORMATS was a high enough version to support all relevant formats, but in feedstock the version was lower.
Rather than trying to fix all this now, we can table that issue and just default image output for graphviz to png as it has always been supported.
Also contains an update to DOCKER_CONTAINERID to reflect changes made in feedstock to support kaleido


.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ commands:
             python .circleci/conda_config.py "$(python setup.py --version)"
             cd evalml-core-feedstock
             echo "$DOCKERHUB_PASSWORD" | docker login -u $DOCKERHUB_USERNAME --password-stdin
-            export DOCKER_CONTAINERID="$(docker run -td condaforge/linux-anvil-comp7)"
+            export DOCKER_CONTAINERID="$(docker run -td condaforge/linux-anvil-cos7-x86_64)"
```
freddyaboulton, line 69: Yes, we need to change the docker image we use in our conda-check because because `python-kaleido` requires a centOS docker image. 

This reflects this [change](https://github.com/conda-forge/evalml-core-feedstock/pull/17) we had to make to our conda recipe.

I tested this change locally by checking out this branch and running our [conda build steps](https://github.com/alteryx/evalml/blob/main/.circleci/config.yml#L59) manually. Tests pass and the build succeeds!

freddyaboulton (approved):
@ParthivNaresh Thanks for doing this! Since this change was approved previously, I just wanted to verify the build_conda_pkg ci check would pass with the version of the feedstock.

I think we're in good shape!

PR #1590, opened 2020-12-21
Fix bug where Imputer changes index of X

### LINKED ISSUE #1422: Investigate if binary classification problems still require numerical {0, 1} targets

### LINKED ISSUE #1442: Imputer changes index on X
The Imputer calls `.reset_index()` in the `transform` method. I don't think we should do this because some downstream components/processing might (reasonably) assume that X and y have the same index, e.g. pandas groupby.

Repro:

```python
from evalml.pipelines.components import Imputer
import pandas as pd

X, y = load_diabetes()

X.index = range(1000, 1000 + X.shape[0])

X_imputed = Imputer().fit_transform(X)

pd.testing.assert_index_equal(X.index, X_imputed.index)
```

Output:
```
AssertionError                            Traceback (most recent call last)
<ipython-input-2-bd81b5490baa> in <module>
     11 X_imputed = Imputer().fit_transform(X)
     12 
---> 13 pd.testing.assert_index_equal(X.index, X_imputed.index)

    [... skipping hidden 2 frame]

AssertionError: Index are different

Index values are different (100.0 %)
[left]:  RangeIndex(start=1000, stop=1442, step=1)
[right]: RangeIndex(start=0, stop=442, step=1)
```
```

### PR DESCRIPTION
**Pull request by freddyaboulton**
### Pull Request Description
Fix #1442 

This has caused two confusing bugs already when the input data have custom indices:
1. TargetEncoder not being able to fit: https://github.com/alteryx/evalml/pull/1401#discussion_r523162379
2. DelayedFeatureTransformer adds nans when X and y have different indices (ran into this during the blitz for time series performance tests)

-----

### DIFF
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -100,16 +100,19 @@ class Imputer(Transformer):
 
         X_null_dropped = X.copy()
         X_null_dropped.drop(self._all_null_cols, inplace=True, axis=1, errors='ignore')
-        X_null_dropped.reset_index(inplace=True, drop=True)
         if X_null_dropped.empty:
             return X_null_dropped
 
         if self._numeric_cols is not None and len(self._numeric_cols) > 0:
             X_numeric = X_null_dropped[self._numeric_cols]
-            X_null_dropped[X_numeric.columns] = self._numeric_imputer.transform(X_numeric)
+            imputed = self._numeric_imputer.transform(X_numeric)
+            imputed.index = X_null_dropped.index
+            X_null_dropped[X_numeric.columns] = imputed
 
         if self._categorical_cols is not None and len(self._categorical_cols) > 0:
             X_categorical = X_null_dropped[self._categorical_cols]
-            X_null_dropped[X_categorical.columns] = self._categorical_imputer.transform(X_categorical)
+            imputed = self._categorical_imputer.transform(X_categorical)
+            imputed.index = X_null_dropped.index
+            X_null_dropped[X_categorical.columns] = imputed
 
         return X_null_dropped
--- a/evalml/tests/component_tests/test_imputer.py
+++ b/evalml/tests/component_tests/test_imputer.py
@@ -199,32 +199,30 @@ def test_imputer_empty_data(data_type):
     imputer = Imputer()
     imputer.fit(X, y)
     transformed = imputer.transform(X, y)
-    assert_frame_equal(transformed, expected, check_dtype=False)
+    assert_frame_equal(transformed, expected, check_dtype=False, check_index_type=False)
 
     imputer = Imputer()
     transformed = imputer.fit_transform(X, y)
-    assert_frame_equal(transformed, expected, check_dtype=False)
+    assert_frame_equal(transformed, expected, check_dtype=False, check_index_type=False)
 
 
-def test_imputer_resets_index():
-    X = pd.DataFrame({'input_val': np.arange(10), 'target': np.arange(10)})
+def test_imputer_does_not_reset_index():
+    X = pd.DataFrame({'input_val': np.arange(10), 'target': np.arange(10),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by angela97lin — APPROVED** [review]
Thanks, looks good! Just had a quick question for my own understanding about updating test code :d

File: evalml/tests/component_tests/test_imputer.py [hunk 1]
```diff
@@ -199,32 +199,30 @@ def test_imputer_empty_data(data_type):
     imputer = Imputer()
     imputer.fit(X, y)
     transformed = imputer.transform(X, y)
-    assert_frame_equal(transformed, expected, check_dtype=False)
+    assert_frame_equal(transformed, expected, check_dtype=False, check_index_type=False)
```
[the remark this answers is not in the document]
freddyaboulton: The index type shouldn't change! When we create the empty dataframes in the test, the index type is `Index`. What happened is that before we would always call `reset_index` so the type would change to `Int64Index`. Since we don't reset the index, the index type should be preserved as `Index`. 

I'm creating `expected` with the correct index type as opposed to setting `check_index_type=False`. I think that's clearer so thanks for the question!

evalml/tests/component_tests/test_imputer.py
```diff
@@ -206,25 +206,23 @@ def test_imputer_empty_data(data_type):
...
     imputer = Imputer()
     imputer.fit(X, y=y)
     transformed = imputer.transform(X)
     pd.testing.assert_frame_equal(transformed,
-                                  pd.DataFrame({'input_val': [1.0, 2, 3, 4, 5, 6, 7, 8, 9]},
-                                               dtype=float,
-                                               index=list(range(0, 9))))
+                                  pd.DataFrame({'input_val': [1.0, 2, 3, 4, 5, 6, 7, 8, 9],
+                                                'input_cat': pd.Categorical(['a'] * 6 + ['b'] * 3)},
+                                               index=list(range(1, 10))))
```
dsherry, line 225: Nice

evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -100,16 +100,19 @@ def transform(self, X, y=None):
...
         X_null_dropped.drop(self._all_null_cols, inplace=True, axis=1, errors='ignore')
-        X_null_dropped.reset_index(inplace=True, drop=True)
         if X_null_dropped.empty:
             return X_null_dropped
 
         if self._numeric_cols is not None and len(self._numeric_cols) > 0:
             X_numeric = X_null_dropped[self._numeric_cols]
-            X_null_dropped[X_numeric.columns] = self._numeric_imputer.transform(X_numeric)
+            imputed = self._numeric_imputer.transform(X_numeric)
+            imputed.index = X_null_dropped.index
```
dsherry, line 109: Is there ever a case where `imputed.shape != X_null_dropped.shape`? I.e., could `self._numeric_imputer.transform(X_numeric)` ever result in dropped rows?

I believe the answer is, to the best of our knowledge, no. Correct?

dsherry (commented):
Looks great! Good call. I wonder if other components need this treatment too.

evalml/pipelines/components/transformers/imputers/imputer.py
freddyaboulton: Yes, I think you're right!

Issue #1593, opened 2020-12-22
Support sklearn 0.24.0


dsherry:
#1592 is preventing us from upgrading to sklearn 0.24.0. Once that's fixed, we should be able to make the upgrade.

freddyaboulton:
For whoever picks this up, we also need to refactor `test_jupyter_graph_check` in model_understanding to support this latest release. 

The test checks that no warnings are raised but sklearn now raises a warning about a future breaking api change.

Issue #1592, opened 2020-12-22
Stacked ensembler: use same CV data splitter as the rest of automl


dsherry:
**Problem**
Currently, the stacked ensembler has its own setup for CV. `IterativeAlgorithm` calls [the `_make_stacked_ensembler` util](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L221), but doesn't currently thread through the data splitter from automl search.

[The data splitter set up by default in the stacked ensembler](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/ensemble/stacked_ensemble_base.py#L53) doesn't set `shuffle=True`, which [could lead to poor performance](https://github.com/alteryx/evalml/issues/1259) if the input dataset has an ordering. It also won't have the same settings for other parameters like `n_folds`, which isn't ideal.

Also, this difference [is preventing us from supporting sklearn 0.24.0](https://github.com/alteryx/evalml/pull/1591). Fixing this issue should allow us to support that version.

**Fix**
Let's have automl pass its data splitter through `IterativeAlgorithm` down into the stacked ensembler.

dsherry:
@angela97lin does my explanation make sense / was there a reason we chose not to do this when we were setting up stacking? :)

angela97lin:
@dsherry I think your explanation makes sense! IIRC when we were setting up stacking and were trying to make it more performant / make stacking run faster, we wanted to default to something that didn't have too many folds--hence the `self._default_cv(n_splits=3, random_state=random_state)` line where we take the default specified by scikit-learn, and hardcoded `n_splits` to 3.

angela97lin:
Dug into this a bit more, and tried to weave the data split method used by AutoML to the stacked ensemble component. However, I ran into this issue (after addressing the API updates necessary to get our `TrainingValidationSplit` class to work):
```

estimator = WrappedSKClassifier(pipeline=LogisticRegressionBinaryPipeline(parameters={'Imputer':{'categorical_impute_strategy': 'm...Logistic Regression Classifier':{'penalty': 'l2', 'C': 1.0, 'n_jobs': -1, 'multi_class': 'auto', 'solver': 'lbfgs'},}))
X =            0         1         2         3         4
0   0.965469  0.041236  0.028701  0.659165  0.213375
1   0.043831...978  0.079577
48  0.376344  0.920154  0.314640  0.180086  0.197598
49  0.682661  0.046529  0.400513  0.412513  0.751464
[...]

PR #1594, opened 2020-12-22
Pin sklearn version

dsherry:
Update-deps bot shows that upgrading to sklearn 0.24.0 breaks the stacked ensembler.

#1592 tracks the bugfix which will unblock usage of sklearn 0.24.0. Then #1593 tracks un-capping the sklearn version.


angela97lin (approved):
Thank you! 👍

jeremyliweishih (commented):
looks good!

PR #1596, opened 2020-12-22
Check two examples per class for multiclass in InvalidTargetDataCheck

### LINKED ISSUE #1595: InvalidTargetDataCheck does not support checking multiclass
- One requirement using sklearn's CV splitting is that each class have at [least 2 examples](https://github.com/scikit-learn/scikit-learn/blob/875d32fe559c51c5f9f5c33df53cae41e6a7256a/sklearn/model_selection/_split.py#L1714). 
- For a multiclass problem, this means that each class must have 2 values in the target column. 
- The **InvalidTargetDataCheck** doesn't check for this situation.

gsheni:
- Fixes #1595


dsherry (approved):
LGTM! Left one nit-pick about the error msg

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -89,6 +89,12 @@ def validate(self, X, y):
                                                      data_check_name=self.name,
                                                      message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_UNIQUE_VALUES,
                                                      details=details).to_dict())
+        if self.problem_type == ProblemTypes.MULTICLASS and value_counts.min() <= 1:
```
angela97lin, line 92: This works, but maybe it'd be more useful for users if we give them all of the classes that don't have at least 2 instances? For example if multiple classes just have one sample, this would only return one of those classes and the user would have to remove to see the other class.

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -70,6 +70,19 @@ def test_invalid_target_data_check_numeric_binary_classification_error():
...
 
+def test_invalid_target_data_check_multiclass_two_examples_per_class():
+    X = pd.DataFrame()
+    invalid_targets_check = InvalidTargetDataCheck("multiclass")
+    assert invalid_targets_check.validate(X, y=pd.Series([0, 1, 1, 2, 2])) == {
+        "warnings": [],
+        "errors": [DataCheckError(message="Target does not have two examples per class which is required for multi-classification",
+                                  data_check_name=invalid_targets_data_check_name,
+                                  message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS,
+                                  details={"least_populated_class_count": 1,
```
angela97lin, line 81: RE previous comment, I think it'd be helpful to add a test where `y=pd.Series([0, 1, 1, 2, 2, 3])` and then return in the details information about 0 or 3.

evalml/data_checks/invalid_targets_data_check.py
angela97lin, line 92: If I'm understanding this correctly, this should only trigger for values that just have one sample right? (0 samples wouldn't be a class); in that case, maybe it's not really useful to return `least_populated_class_count` since that's always 1?

angela97lin (changes requested):
Looks good, but I left some suggestions on how we could improve the output / clean up the details 👍

evalml/data_checks/invalid_targets_data_check.py
gsheni: Added.

gsheni: Yep, you're right. I was over complicating it. I removed the `least_populated_class_count` from details.

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
gsheni: Added

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -89,6 +89,12 @@ def validate(self, X, y):
...
+            details = {"least_populated_class_count": value_counts.min(), "least_populated_class_label": value_counts.idxmin()}
+            messages["errors"].append(DataCheckError(message="Target does not have two examples per class which is required for multi-classification",
```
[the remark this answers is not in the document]
gsheni: Fixed

angela97lin (approved):
Thanks for making the updates, LGTM! 😁

PR #1613, opened 2020-12-28
Addresses stacked ensemble component CV for scikit-learn 0.24 support

### LINKED ISSUE #1592: Stacked ensembler: use same CV data splitter as the rest of automl

angela97lin:
Closes #1592

### DIFF
--- a/evalml/pipelines/components/ensemble/stacked_ensemble_base.py
+++ b/evalml/pipelines/components/ensemble/stacked_ensemble_base.py
@@ -50,7 +50,7 @@ class StackedEnsembleBase(Estimator):
         if len(set([pipeline.problem_type for pipeline in input_pipelines])) > 1:
             raise ValueError("All pipelines must have the same problem type.")
 
-        cv = cv or self._default_cv(n_splits=3, random_state=random_state)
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):

PR #1616, opened 2020-12-28
Change TargetLeakageDataCheck to use Woodwork's mutual_information method

### LINKED ISSUE #927: Label leakage: use mutual information
Per @kmax12's comment in #917, in `DetectLabelLeakage` data check we could use mutual information which can support numeric and categorical data instead of pearson correlation.

https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.mutual_info_classif.html
https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.mutual_info_regression.html#sklearn.feature_selection.mutual_info_regression

bchen1116:
fix #927


evalml/data_checks/target_leakage_data_check.py
```diff
@@ -63,14 +62,17 @@ def validate(self, X, y):
         y = _convert_to_woodwork_structure(y)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_woodwork_types_wrapper(y.to_series())
-
-        if y.dtype not in numeric_and_boolean_dtypes:
+        combined = X.copy()
```
jeremyliweishih, line 65: Could we remove the overhead of converting X and y above and just convert using a combined dataframe?

jeremyliweishih (approved):
Cool looks good to me. Left a comment about removing some of the ww conversions and doing a combined one.

evalml/data_checks/target_leakage_data_check.py
bchen1116: @jeremyliweishih I tried simplifying this, but I think this might be the best way to preserve the functionality we want. In order to properly combine, we either have to have ww.Datatables or pd.Dataframes, so the first 2 '_convert_to_woodwork_structure` calls make's that process cleaner. The next two calls preserve the types that woodwork gives, while allowing us to then copy X and combine it with y, and the last call allows us to turn it back to ww. If I combined first, I would need to convert both the X and y to pandas Dataframes/Series, which could lead to longer code. Definitely free to discuss this more if I'm missing something!

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -16,7 +15,7 @@ class TargetLeakageDataCheck(DataCheck):
     def __init__(self, pct_corr_threshold=0.95):
```
angela97lin, line 15: I think based on the discussions we had where it seemed like there could be pros/cons of both methods, it'd be nice if the method used to determine correlation was something that could be set by the user via a parameter. We could still default it to use mutual information, but allow users to use the original pearson correlation method.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -63,14 +62,17 @@ def validate(self, X, y):
         y = _convert_to_woodwork_structure(y)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_woodwork_types_wrapper(y.to_series())
-
-        if y.dtype not in numeric_and_boolean_dtypes:
+        combined = X.copy()
+        combined['target'] = y
```
angela97lin, line 66: What if the original data had a column named "target"? What would happen here?

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -63,14 +62,17 @@ def validate(self, X, y):
...
+        combined = _convert_to_woodwork_structure(combined)
+        mutual_info = combined.mutual_information()
```
angela97lin, line 68: I think this is the easiest way to calculate mutual information, but it's also much more costly. Before, we were doing a O(n) implementation. Mutual information is O(n^2) which shouldn't be necessary here, given we only care about the correlation between the target and each column, and not each column with another. We should think about how to reduce this time complexity if we want to use mutual info!

angela97lin (changes requested):
Looking good, but left some comments about reducing time complexity and keeping our original pearson coefficient method via a parameter!

evalml/data_checks/target_leakage_data_check.py
bchen1116: Ah, good call! Will fix that.

evalml/data_checks/target_leakage_data_check.py
bchen1116: Great call! Will fix that.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -13,23 +13,25 @@
...
 
-    def __init__(self, pct_corr_threshold=0.95):
+    def __init__(self, pct_corr_threshold=0.95, pearson_corr=False):
         """Check if any of the features are highly correlated with the target.
 
-        Currently only supports binary and numeric targets and features.
+        If `pearson_corr=False`, supports all target and feature types. Otherwise, only supports binary with numeric and boolean dtypes.
 
         Arguments:
             pct_corr_threshold (float): The correlation threshold to be considered leakage. Defaults to 0.95.
```
gsheni, line 22: Just a heads up, mutual information values are normalized to be return between 0 (no mutual information) and 1 (perfect correlation). This is a bit different than pearson correlation values which are between -1 and 1. Might be worth mentioning this in the docs somewhere.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -63,14 +65,28 @@ def validate(self, X, y):
         y = _convert_to_woodwork_structure(y)
         X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_woodwork_types_wrapper(y.to_series())
+        if not self.pearson:
```
angela97lin, line 68: Can we extract this into its own method? We could have a `_calculate_pearson` and `_calculate_mutual_information` in this class or something.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -63,14 +65,28 @@ def validate(self, X, y):
...
+            highly_corr_cols = []
+            target = 'target'
+            while target in X.columns:
```
angela97lin, line 71: Do we need to do this? I like that we're extracting out the column from the DF and comparing it to the target but if mutual information is greater than the threshold for that [target, col], we should know what the col is without labeling the target this way, right?

angela97lin (changes requested):
Nice, this does less unnecessary computation!

Two comments:
- Can we move the impl for the two methods to their own separate methods?
- I like that we're extracting out the column from the DF and comparing it to the target but if mutual information is greater than the threshold for that [target, col], we should know what the col is without labeling the target this way, right?

evalml/data_checks/target_leakage_data_check.py
bchen1116: Thanks for the info. I'll fix that.

evalml/data_checks/target_leakage_data_check.py
bchen1116: @angela97lin I added this so that I could add the target column to the X dataframe without any column name conflicts. This way, I can iterate through the woodwork DataTable columns and easily pick out the column of interest and the target column. Was there a better way to do this?

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -13,23 +13,49 @@
 class TargetLeakageDataCheck(DataCheck):
     """Check if any of the features are highly correlated with the target."""
 
-    def __init__(self, pct_corr_threshold=0.95):
+    def __init__(self, pct_corr_threshold=0.95, pearson_corr=False):
```
chukarsten, line 16: A little nudge towards explicitness: perhaps we might consider a conditional on a string with the positional arg being named something like "corr_check_method" and the default being "pearson" with "mutual_information" being an option?  I think it's a little clear then what's happening if not Pearson correlation.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -13,23 +13,49 @@
...
+        X = X.select_dtypes(include=numeric_and_boolean_dtypes)
+        if len(X.columns) > 0:
+            highly_corr_cols = [label for label, col in X.iteritems() if abs(y.corr(col)) >= self.pct_corr_threshold]
+        return highly_corr_cols
+
+    def _calculate_mutual_information(self, X, y):
+        highly_corr_cols = []
+        target = 'target'
+        while target in X.columns:
+            target += '0'
```
chukarsten, line 43: I guess this is for my edification: is "target" always guaranteed to be in the incoming data frame?  I'm a little wary on there being a hardcoded check on it.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -13,23 +13,49 @@
...
+        combined = X.copy()
+        combined[target] = y
```
chukarsten, line 45: Is the idea here to accommodate the eventuality that the input data frame contains the target variable as well as accepting duplicate target columns from y?  It seems you could just do an `X.drop(target, errors="ignore")` if you wanted to eliminate the possibility that you have duplicate target columns.  That kind of goes back to my question about being sure that the target is named "target" though.

chukarsten (changes requested):
Left a few notes about explicitly defining the correlation check method and dropping the target var.

evalml/data_checks/target_leakage_data_check.py
bchen1116: @chukarsten, 'target' isn't guaranteed, but this code allows me to safely add the `y` data to the `X` dataframe in order to find mutual information. The default name for column with `y` would be `target`, but if `target` already exists as a column name, I wanted to find another name to allow the data to be safely added without replacing/overwriting any other columns

evalml/data_checks/target_leakage_data_check.py
bchen1116: @chukarsten this is just to make sure that the column we add doesn't interfere with the current columns the user has. If a user has a column in the `X` data already labeled as `target`, we don't want to assume that is equivalent to the `y` data, so dropping it might be detrimental

evalml/data_checks/target_leakage_data_check.py
bchen1116: @chukarsten thanks for that pointer! It definitely would be better to be more explicit here. To make it easier for the user, I'll leave the input as is, but I'll update the documentation to let users know that it uses mutual information as a default, and Pearson correlation otherwise. Having a boolean arg seems simpler for the user versus a string arg.

evalml/data_checks/target_leakage_data_check.py
angela97lin: @bchen1116 I don't think you need to append the target column to the X dataframe in order to iterate if we choose to build a dataframe at each iteration with the column of interest and the target column. Does that make sense?

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -11,25 +11,52 @@
...
         Arguments:
             pct_corr_threshold (float): The correlation threshold to be considered leakage. Defaults to 0.95.
+            pearson_corr (bool): Whether or not to use the Pearson correlation versus mutual information. Defaults to False, which uses mutual information.
 
         """
         if pct_corr_threshold < 0 or pct_corr_threshold > 1:
             raise ValueError("pct_corr_threshold must be a float between 0 and 1, inclusive.")
         self.pct_corr_threshold = pct_corr_threshold
+        self.pearson = pearson_corr
+
```
angela97lin, line 32: Do you need to pass messages here?

evalml/data_checks/target_leakage_data_check.py
bchen1116: Left a comment, but this was just how I avoided naming conflicts.

evalml/data_checks/target_leakage_data_check.py
bchen1116: I think this was commented at the wrong spot, but removed the `messages` input from `_calculate_pearson`

evalml/data_checks/target_leakage_data_check.py
bchen1116: @angela97lin changed the implementation to avoid this. Thanks for the input, I think it's much cleaner now!

evalml/data_checks/target_leakage_data_check.py
bchen1116: @chukarsten I changed the implementation for this. Thanks for the input, I think it looks much cleaner now!

chukarsten (approved):
Definitely like the changes.  Good job.

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -11,25 +13,49 @@
...
+        for col in X.columns:
+            cols_to_compare = _convert_to_woodwork_structure(pd.DataFrame({col: X[col], str(col) + "y": y}))
+            mutual_info = cols_to_compare.mutual_information()
+            if len(mutual_info) > 0 and mutual_info['mutual_info'].iloc[0] > self.pct_corr_threshold:
+                highly_corr_cols.append(col)
+        return highly_corr_cols
 
     def validate(self, X, y):
-        """Check if any of the features are highly correlated with the target.
+        """Check if any of the features are highly correlated with the target by using mutual information.
```
angela97lin, line 55: Update docstr to mention depending on bool value?

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -11,25 +13,49 @@
...
+        """Check if any of the features are highly correlated with the target by using mutual information or Pearson correlation.
 
-        Currently only supports binary and numeric targets and features.
+        If `pearson_corr=False`, this data check uses mutual information and supports all target and feature types.
+        Otherwise, it uses Pearson correlation and only supports binary with numeric and boolean dtypes.
+        Pearson correlation returns a value in [-1, 1], while mutual information returns a value in [0, 1].
 
         Arguments:
             pct_corr_threshold (float): The correlation threshold to be considered leakage. Defaults to 0.95.
+            pearson_corr (bool): Whether or not to use the Pearson correlation versus mutual information. Defaults to False, which uses mutual information.
```
angela97lin, line 27: Not blocking and just a suggestion since this works for now, but I wonder if its better to have a set of strings determining the method. I can imagine if we add a third value, we'll have to change this to support that, too. But I guess this works for now 🤷

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -11,25 +13,49 @@
...
+            if len(mutual_info) > 0 and mutual_info['mutual_info'].iloc[0] > self.pct_corr_threshold:
+                highly_corr_cols.append(col)
+        return highly_corr_cols
 
     def validate(self, X, y):
-        """Check if any of the features are highly correlated with the target.
+        """Check if any of the features are highly correlated with the target by using mutual information.
 
-        Currently only supports binary and numeric targets and features.
+        If `pearson_corr=False`, supports all target and feature types. Otherwise, only supports binary with numeric and boolean dtypes.
```
angela97lin, line 57: A bit confusing since `pearson_corr` is an argument of init and doesn't mean much here. Maybe having `pearson_corr` and `pearson` be named the same will help a bit? 🤷

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -11,25 +13,49 @@
...
+    """Check if any of the features are highly correlated with the target by using mutual information or Pearson correlation."""
 
-    def __init__(self, pct_corr_threshold=0.95):
-        """Check if any of the features are highly correlated with the target.
+    def __init__(self, pct_corr_threshold=0.95, pearson_corr=False):
+        """Check if any of the features are highly correlated with the target by using mutual information or Pearson correlation.
 
-        Currently only supports binary and numeric targets and features.
+        If `pearson_corr=False`, this data check uses mutual information and supports all target and feature types.
+        Otherwise, it uses Pearson correlation and only supports binary with numeric and boolean dtypes.
```
angela97lin, line 22: 👍

evalml/tests/data_checks_tests/test_target_leakage_data_check.py
```diff
@@ -126,3 +128,282 @@ def test_target_leakage_data_check_input_formats():
...
+
+
+def test_target_leakage_data_check_name():
+    y = pd.Series([1, 0, 1, 1])
+    X = pd.DataFrame()
+    X["target"] = y * 3
+    X["target0"] = y - 1
+    X["target00"] = y / 10
+    X["target000"] = ~y
+    X["e"] = [0, 0, 0, 0]
```
angela97lin, line 269: I see that you decided to update the impl; in that case, do we still need these tests?

angela97lin (commented):
Nice nice; left a few more comments and questions, but haven't had the chance to look at everything entirely yet :d

evalml/tests/data_checks_tests/test_target_leakage_data_check.py
bchen1116: Good catch, I'll remove these

evalml/data_checks/target_leakage_data_check.py
bchen1116: @chukarsten recommended the same thing. Will change!

PR #1619, opened 2020-12-29
1570 AutoML: pass number of CV folds to default data checks

### LINKED ISSUE #1570: AutoML: pass number of CV folds to default data checks
RE @bchen1116 point [here](https://github.com/alteryx/evalml/pull/1568#pullrequestreview-554679046):

> [We should] update our implementation for class_imbalance_data_check, specifically the cv_folds arg [here](https://github.com/alteryx/evalml/blob/a48ebfb64a16688ee95047e53f4aced54755d23b/evalml/data_checks/class_imbalance_data_check.py#L17). It defaults to 3 since that's the default data split n_folds, but with this change, I believe we should update AutoMLSearch to pass in the required param.

I completely agree! Good eye 👀 

Filing as a bug because if someone customizes the data splitter passed to automl. the class imbalance data check won't get the right `n_folds` value and therefore isn't functioning exactly as designed.

ParthivNaresh:
### Pull Request Description
Fixes #1570


evalml/automl/automl_search.py
```diff
@@ -328,7 +328,7 @@ def _validate_data_checks(self, data_checks):
             return AutoMLDataChecks(data_checks)
```
ParthivNaresh, line 328: Although this PR is for the DefaultDataChecks, I spent a while trying to figure out how we might pass it to DataChecks or AutoMLDataChecks. It becomes a bigger issue in which we have to check and see if ClassImbalanceDataCheck is one of the Data Checks, then determine what parameters were passed by the user, and if `num_cv_folds` is one of them. Then overwriting that to reflect `self.data_splitter.get_n_splits()`. There's probably an easier way I'm sure :)

evalml/tests/automl_tests/test_automl.py
```diff
@@ -407,6 +408,16 @@ def test_automl_bad_data_check_parameter_type():
...
 
 
+def test_validate_data_check_n_splits(X_y_multi):
+    X, y = X_y_multi
+    data_split = make_data_splitter(X, y, problem_type='multiclass', n_splits=4, random_state=42)
+
+    automl = AutoMLSearch(X, y, problem_type="multiclass", max_iterations=1, n_jobs=1, data_splitter=data_split)
+    automl.search(data_checks='auto')
+
+    assert automl.data_splitter.get_n_splits() == 4
```
angela97lin, line 418: Hm, I think what we want to check here is actually that the data check got n_splits = 4 to confirm that the correct number of CV folds was passed to the data check.

angela97lin (commented):
Left a comment about testing: We want to confirm that the data check got the appropriate parameter!

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: Thanks for catching that!

evalml/data_checks/default_data_checks.py
```diff
@@ -25,7 +25,7 @@ class DefaultDataChecks(DataChecks):
     _DEFAULT_DATA_CHECK_CLASSES = [HighlyNullDataCheck, IDColumnsDataCheck,
                                    TargetLeakageDataCheck, InvalidTargetDataCheck, NoVarianceDataCheck]
 
-    def __init__(self, problem_type):
+    def __init__(self, problem_type, n_splits=3):
```
freddyaboulton, line 28: Add `n_splits` to docstring please?

evalml/automl/automl_search.py
freddyaboulton: I think it's fine to only make the change for `DefaultDataChecks`! If the user is providing their own data checks, it's assumed they init them with the appropriate parameters!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -412,6 +413,15 @@ def test_automl_bad_data_check_parameter_type():
         automl.search(data_checks=[MockDataCheckErrorAndWarning])
 
 
+def test_validate_data_check_n_splits(X_y_multi):
```
freddyaboulton, line 416: Maybe a better way of testing this is to run `search(data_checks='auto')` on a dataset where one of the classes only has 7 observations.

That way we don't have to call the private `_validate_data_checks` or rely on the fact that the `ClassImbalanceDataCheck` is the last one in the list or that `cv_folds` actually corresponds to `2 * n_splits`.

freddyaboulton (approved):
@ParthivNaresh Looks good! I just have a small suggestion on how to improve the unit test. Implementation looks great :)

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: Ah I got it, so you're recommending I validate that the parameter is being passed by triggering the ClassImbalanceDataCheck error, nice!

angela97lin (approved):
Sweet, thanks for updating this! LGTM 🥳

PR #1622, opened 2020-12-29
Fix AutoMLSearch to only threshold pipelines when problem is binary classification

### LINKED ISSUE #1573: AutoML errors if objective is defined as valid for both binary and multiclass
**Problem**
Our [code in automl search](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L662) which optimizes binary classification thresholding should only run if the problem type is "binary", not for "multiclass."

**Scope**
Nobody has encountered this because a) I see from `git blame` that we updated the offending `if` statements in early Nov, and b) most (all?) of [our current objectives](https://github.com/alteryx/evalml/blob/main/evalml/objectives/standard_metrics.py) are defined for only one problem type, not both binary and multiclass. So its a minor bug.

**Repro**
I wasn't able to test this because bug #1572 gets in the way, but I'm pretty sure the following will cause a stack trace independent of #1572: 
```python
import evalml
import pandas as pd
import numpy as np
from sklearn import metrics
class ClassificationObjective(evalml.objectives.ObjectiveBase):
    problem_types = [evalml.problem_types.ProblemTypes.BINARY, evalml.problem_types.ProblemTypes.MULTICLASS]
class ClassificationAccuracy(ClassificationObjective):
    """Accuracy score for binary and multiclass classification."""
    name = "Classification Accuracy"
    greater_is_better = True
    score_needs_proba = False
    perfect_score = 1.0

    def objective_function(self, y_true, y_predicted, X=None):
        return metrics.accuracy_score(y_true, y_predicted)
X, y = evalml.demos.load_wine()
automl = evalml.automl.AutoMLSearch(problem_type="multiclass", objective=ClassificationAccuracy())
automl.search(X, y)
```

**Fix**
Currently [that code](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L662) checks `self.objective.is_defined_for_problem_type(ProblemTypes.BINARY)` in a couple places. I think it should do that but first also check `self.problem_type == ProblemTypes.BINARY`. Easy fix.

The code I linked to is in `_compute_cv_scores`; note @bchen1116 is adding similar code to `search` for fitting the best pipeline, in #1547

bchen1116:
fix #1573

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -532,7 +532,7 @@ class AutoMLSearch:
             X_threshold_tuning = None
             y_threshold_tuning = None
             X_train, y_train = self.X_train, self.y_train
-            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
+            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold and self.problem_type == ProblemTypes.BINARY:
                 X_train, X_threshold_tuning, y_train, y_threshold_tuning = split_data(X_train, y_train, self.problem_type,
                                                                                       test_size=0.2,
                                                                                       random_state=self.random_seed)
@@ -552,7 +552,7 @@ class AutoMLSearch:
         Returns:
             Trained pipeline instance
         """
-        if self.objective.is_defined_for_problem_type(ProblemTypes.BINARY):
+        if self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.problem_type == ProblemTypes.BINARY:
             pipeline.threshold = 0.5
             if X_threshold_tuning:
                 y_predict_proba = pipeline.predict_proba(X_threshold_tuning)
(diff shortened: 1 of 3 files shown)


evalml/automl/automl_search.py
```diff
@@ -532,7 +532,7 @@ def search(self, data_checks="auto", show_iteration_plot=True):
             X_threshold_tuning = None
             y_threshold_tuning = None
             X_train, y_train = self.X_train, self.y_train
-            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
+            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold and self.problem_type == ProblemTypes.BINARY:
```
dsherry, line 535: @bchen1116 can you please use the new `is_binary` helper in `evalml/problem_types/utils.py`? That way, this will be true for timeseries binary classification too

evalml/automl/automl_search.py
```diff
@@ -532,7 +532,7 @@ def search(self, data_checks="auto", show_iteration_plot=True):
             X_threshold_tuning = None
             y_threshold_tuning = None
             X_train, y_train = self.X_train, self.y_train
-            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold:
+            if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold and is_binary(self.problem_type):
```
dsherry, line 535: I just filed #1624 to track adding support here for binary timeseries pipelines. Otherwise, looks good

dsherry (approved):
LGTM! Please wait until after the Dec release to merge.

evalml/automl/automl_search.py
bchen1116: Thanks for filing!

PR #1629, opened 2020-12-31
Cap scipy<1.6.0

angela97lin:
Latest scipy update breaks main (#1628) so capping scipy so we can address separately.


jeremyliweishih (approved):
Looks good. Can you file an issue to investigate it further?

angela97lin:
Yup, filed #1630!

PR #1647, opened 2021-01-05
Stop AutoMLSearch from running multiple times

### LINKED ISSUE #1424: AutoMLSearch: calling search twice on same instance doesn't work
Not sure if this is intended behavior, but when I call `automl.search(X, y)` on the same automl object, the second search will run the baseline round and then quit because the number of iterations starts at 6 (or 1+num of iterations from first search):

image

bchen1116:
fix #1424 

- Remove `has_searched` property from `AutoMLSearch`
- Update `best_pipeline` in `add_to_rankings` as well
- No-op when user calls `AutoMLSearch.search` twice

Updated search call
image

The files changed look like I changed a lot, but I commented on the changes I made from the main branch and this branch in `automl_search.py`


evalml/automl/automl_search.py
```diff
@@ -403,143 +404,154 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
-        run_ensembling = self.ensembling
-        if run_ensembling and len(self.allowed_pipelines) == 1:
-            logger.warning("Ensembling is set to True, but the number of unique pipelines is one, so ensembling will not run.")
-            run_ensembling = False
-
-        if run_ensembling and self.max_iterations is not None:
-            # Baseline + first batch + each pipeline iteration + 1
-            first_ensembling_iteration = (1 + len(self.allowed_pipelines) + len(self.allowed_pipelines) * self._pipelines_per_batch + 1)
-            if self.max_iterations < first_ensembling_iteration:
+        if not self._searched:
```
bchen1116, line 407: Unsure why it changed this much, but I added this if statement to determine if we need to do the search, then put the rest of the search within this if statement

evalml/automl/automl_search.py
```diff
@@ -403,143 +404,154 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+                    break
+
+            elapsed_time = time_elapsed(self._start)
+            desc = f"\nSearch finished after {elapsed_time}"
+            desc = desc.ljust(self._MAX_NAME_LEN)
+            logger.info(desc)
+
+            best_pipeline = self.rankings.iloc[0]
+            best_pipeline_name = best_pipeline["pipeline_name"]
+            self._find_best_pipeline()
```
bchen1116, line 532: I find the best pipeline here

evalml/automl/automl_search.py
```diff
@@ -403,143 +404,154 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+            logger.info(f"Best pipeline: {best_pipeline_name}")
+            logger.info(f"Best pipeline {self.objective.name}: {best_pipeline['score']:3f}")
+            self._searched = True
```
bchen1116, line 535: and I set searched to be true here so we won't redo the same search

evalml/automl/automl_search.py
```diff
@@ -403,143 +404,154 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+        else:
+            logger.info("AutoMLSearch has already been run and will not run again on the same instance. Re-initialize AutoMLSearch to search again.")
 
+    def _find_best_pipeline(self):
```
bchen1116, line 539: New method to find the best pipeline and train only if the pipeline is new

evalml/automl/automl_search.py
chukarsten: Have you considered instead just returning instead of true?  Saves a little nesting of conditionals and might make the diff a little clearer.

angela97lin: Agreed with @chukarsten's comment! Makes it easier to follow along

bchen1116: good call @chukarsten! I'll fix that.

chukarsten (approved):
Looks good!

evalml/automl/automl_search.py
```diff
@@ -403,6 +404,10 @@ def search(self, data_checks="auto", show_iteration_plot=True):
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
         """
+        if self._searched:
+            logger.info("AutoMLSearch has already been run and will not run again on the same instance. Re-initialize AutoMLSearch to search again.")
+            return
```
dsherry, line 409: @bchen1116 hmm after our discussion yesterday I thought we were going with option 3 [from the issue writeup](https://github.com/alteryx/evalml/issues/1424#issuecomment-746543694), not option 2. Did I remember wrong?

evalml/automl/automl_search.py
dsherry: Nice

docs/source/release_notes.rst
```diff
@@ -15,8 +15,10 @@ Release Notes
         * Fixed AutoMLSearch stacktrace when a cutom objective was passed in as a primary objective or additional objective :pr:`1575`
         * Limit ``load_fraud`` dataset loaded into ``automl.ipynb`` :pr:`1646`
         * Fix custom index bug for ``MAPE`` objective :pr:`1641`
+        * ``add_to_rankings`` updates ``AutoMLSearch.best_pipeline`` when necessary :pr:`1647`
         * Fixed bug where time series baseline estimators were not receiving ``gap`` and ``max_delay`` in ``AutoMLSearch`` :pr:`1645`
     * Changes
+        * Rerunning search for ``AutoMLSearch`` results in a message thrown rather than failing the search, and removed ``has_searched`` property :pr:`1647`
```
dsherry, line 21: Please add the removal of `has_searched` to the breaking changes section

evalml/automl/automl_search.py
bchen1116: @dsherry, I voiced for option 2 since the `X_train, y_train` data are now passed through the init of `AutoMLSearch`, and during the discussion yesterday, I believe everyone agreed, which is why I proceeded with option 2. Do you think option 3 would be better? My main issue with option 3 is that rerunning the same AutoMLClass, with the same data and defaults, should lead to the same results, and therefore isn't very useful. Let me know your thoughts, and we can also discuss more after standup if needed!

dsherry: That's entirely fair. Unless we provided a way for people to reconfigure the automl settings on a single automl search instance, rerunning the search from scratch should produce highly similar results. 👍 

Sounds good to me. Just wanted to make sure I wasn't missing something.

evalml/automl/automl_search.py
```diff
@@ -403,6 +404,10 @@ def search(self, data_checks="auto", show_iteration_plot=True):
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
         """
+        if self._searched:
+            logger.info("AutoMLSearch has already been run and will not run again on the same instance. Re-initialize AutoMLSearch to search again.")
```
dsherry, line 408: Nit-pick: "AutoMLSearch.search() has already been run..."

dsherry (approved):
@bchen1116 LGTM!

Issue #1580, opened 2020-12-18
Integrate time series classification pipelines into AutoMLSearch


freddyaboulton:

freddyaboulton:
Setting this as blocked by #1517 because there will be lots of code duplication between `TimeSeriesBaselineRegressionPipeline`, `TimeSeriesBaselineBinaryPipeline`, and `TimeSeriesBaselineMulticlassPipeline` unless we come up with a general pattern for allowing `self.estimator.predict(X=None, y)` in our time series pipelines. We can technically do this issue by reusing the `ModeBaseline` pipelines but I don't think that would count as a complete AutoMLSearch integration.

PR #1651, opened 2021-01-05
Allow empty features in time series pipelines

### LINKED ISSUE #1517: Allow X to be None in Time Series Pipelines

freddyaboulton:
### Pull Request Description
Fixes #1517.

@dsherry and @bchen1116 Can't tag you as reviewers but please take a look when you can!

-----

### DIFF
--- a/evalml/pipelines/components/estimators/estimator.py
+++ b/evalml/pipelines/components/estimators/estimator.py
@@ -21,6 +21,7 @@ class Estimator(ComponentBase):
 
     To see some examples, check out the definitions of any Estimator component.
     """
+    predict_uses_y = False
 
     @property
     @classmethod
--- a/evalml/pipelines/components/estimators/estimator.py
+++ b/evalml/pipelines/components/estimators/estimator.py
@@ -21,6 +21,7 @@ class Estimator(ComponentBase):
 
     To see some examples, check out the definitions of any Estimator component.
     """
+    # We can't use the inspect module to dynamically determine this because of issue 1582
     predict_uses_y = False
 
     @property
--- a/evalml/pipelines/components/estimators/regressors/time_series_baseline_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/time_series_baseline_regressor.py
@@ -53,12 +53,8 @@ class TimeSeriesBaselineRegressor(Estimator):
         return self
 
     def predict(self, X, y=None):
-        if X is None:
-            X = pd.DataFrame()
         if y is None:
             raise ValueError("Cannot predict Time Series Baseline Regressor if y is None")
-        X = _convert_to_woodwork_structure(X)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
         y = _convert_to_woodwork_structure(y)
         y = _convert_woodwork_types_wrapper(y.to_series())
 
--- a/evalml/pipelines/components/estimators/regressors/time_series_baseline_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/time_series_baseline_regressor.py
@@ -19,6 +19,7 @@ class TimeSeriesBaselineRegressor(Estimator):
     hyperparameter_ranges = {}
     model_family = ModelFamily.BASELINE
     supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+    predict_uses_y = True
 
     def __init__(self, gap=1, random_state=0, **kwargs):
         """Baseline time series regressor that predicts using the naive forecasting approach.
--- a/evalml/pipelines/regression/time_series_baseline_regression.py
+++ b/evalml/pipelines/regression/time_series_baseline_regression.py
@@ -1,64 +1,7 @@
-import numpy as np
-import pandas as pd
-
 from evalml.pipelines import TimeSeriesRegressionPipeline
-from evalml.utils.gen_utils import (
-    _convert_to_woodwork_structure,
-    _convert_woodwork_types_wrapper,
-    drop_rows_with_nans,
-    pad_with_nans
-)
 
 
 class TimeSeriesBaselineRegressionPipeline(TimeSeriesRegressionPipeline):
     """Baseline Pipeline for time series regression problems."""
     _name = "Time Series Baseline Regression Pipeline"
     component_graph = ["Time Series Baseline Regressor"]
-
-    def fit(self, X, y):
-        """Fit a time series regression pipeline.
-
-        Arguments:
-            X (ww.DataTable, pd.DataFrame, np.ndarray, or None): The input training data of shape [n_samples, n_features]
-            y (ww.DataColumn, pd.Series, np.ndarray): The target training targets of length [n_samples]
-
-        Returns:
-            self
-        """
-        if X is None:
-            X = pd.DataFrame()
-
-        X = _convert_to_woodwork_structure(X)
-        y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        y = _convert_woodwork_types_wrapper(y.to_series())
-
-        X_t = self._compute_features_during_fit(X, y)
-        if X_t.empty:
-            X_t = pd.DataFrame(np.zeros(len(y)))
-
-        y_shifted = y.shift(-self.gap)
-        X_t, y_shifted = drop_rows_with_nans(X_t, y_shifted)
-        self.estimator.fit(X_t, y_shifted)
-        return self
-
-    def predict(self, X, y, objective=None):
-        """Make predictions using selected features.
-
-        Arguments:
-            X (ww.DataTable, pd.DataFrame, np.ndarray, or None): Data of shape [n_samples, n_features]
-            y (ww.DataColumn, pd.Series, np.ndarray): The target training targets of length [n_samples]. y is required for this pipeline.
-            objective (Object or string): The objective to use to make predictions
-
-        Returns:
-            pd.Series: Predicted values.
-        """
-        if X is None:
-            X = pd.DataFrame()
-        X = _convert_to_woodwork_structure(X)
-        y = _convert_to_woodwork_structure(y)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        y = _convert_woodwork_types_wrapper(y.to_series())
-
-        predictions = self.estimator.predict(None, y)
-        return pad_with_nans(predictions, max(0, X.shape[0] - predictions.shape[0]))
--- a/evalml/pipelines/time_series_classification_pipelines.py
+++ b/evalml/pipelines/time_series_classification_pipelines.py
@@ -84,7 +84,7 @@ class TimeSeriesClassificationPipeline(ClassificationPipeline):
         return self._estimator_predict_helper(features, y, return_probabilities=False)
 
     def _estimator_predict_proba(self, features, y):
-        """Get estimator prediced probabilities.
+        """Get estimator predicted probabilities.
 
         This helper passes y as an argument if needed by the estimator.
         """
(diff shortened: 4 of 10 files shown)


evalml/pipelines/components/estimators/regressors/time_series_baseline_regressor.py
```diff
@@ -52,12 +53,8 @@ def fit(self, X, y=None):
         return self
 
     def predict(self, X, y=None):
-        if X is None:
```
freddyaboulton, line 55: Deleting these because it's not needed

evalml/pipelines/regression/time_series_baseline_regression.py
```diff
@@ -1,64 +1,7 @@
...
-    pad_with_nans
-)
 
 
 class TimeSeriesBaselineRegressionPipeline(TimeSeriesRegressionPipeline):
     """Baseline Pipeline for time series regression problems."""
     _name = "Time Series Baseline Regression Pipeline"
     component_graph = ["Time Series Baseline Regressor"]
-
-    def fit(self, X, y):
```
freddyaboulton, line 18: This code was needed to accept the `X=None` case so we can delete it now! This means adding baselines for time series classification will be easy now since we can just reuse the time series baseline regressor and not worry about implementing pipeline `fit` or `predict` !

evalml/pipelines/components/estimators/estimator.py
```diff
@@ -21,6 +21,8 @@ class Estimator(ComponentBase):
 
     To see some examples, check out the definitions of any Estimator component.
     """
+    # We can't use the inspect module to dynamically determine this because of issue 1582
+    predict_uses_y = False
```
freddyaboulton, line 25: We need a way of knowing whether or not to pass `y` as an argument to `predict` or `predict_proba` since one (more in the future) estimator allows this and others don't. I think this is the lowest-touch way of doing this.

One alternative is to list `y` as an argument for all estimators. However, I think that would be confusing for users to list `y` if it's not used because they might think our estimators are "cheating" by looking at the target value. Maybe I'm just being paranoid though.

chukarsten (approved):
This seems conceptually sound.  Definitely a fan of removing unused code and simplification.

bchen1116 (approved):
LGTM!

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -61,21 +60,41 @@ def fit(self, X, y):
...
+
+    def _estimator_predict(self, features, y):
+        """Get estimator predictions.
+
+        This helper passes y as an argument if needed by the estimator.
+        """
+        return self._estimator_predict_helper(features, y, return_probabilities=False)
+
+    def _estimator_predict_proba(self, features, y):
+        """Get estimator prediced probabilities.
```
angela97lin, line 87: Typo here :d

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -113,14 +132,15 @@ def predict_proba(self, X, y=None):
...
+        proba = self._estimator_predict_proba(features_no_nan, y_encoded)
         proba.columns = self._encoder.classes_
         return pad_with_nans(proba, max(0, features.shape[0] - proba.shape[0]))
 
     def _compute_predictions(self, X, y, objectives):
         """Compute predictions/probabilities based on objectives."""
-        y_predicted = None
-        y_predicted_proba = None
+        y_predicted = pd.Series([])
+        y_predicted_proba = pd.DataFrame([])
```
angela97lin, line 143: I may be confused but why this change? This'll always return a valid prediction value, even if its empty?

evalml/pipelines/time_series_classification_pipelines.py
freddyaboulton: Thank you!

evalml/pipelines/time_series_classification_pipelines.py
freddyaboulton: `drop_rows_with_nans` (which is called in `score`) needs a pandas data type and not `None`. I changed `drop_rows_with_nans` to accept None and so I changed these back to `None`.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -319,6 +322,37 @@ class Pipeline(pipeline_class):
...
+        label_map = {0: "good", 1: "bad", 2: "best"}
+        y = pd.Series(y).map(lambda label: label_map[label])
+        expected_unique_values = {"good", "bad", "best"}
+        objectives = ["MCC Multiclass", "Log Loss Multiclass"]
+
+    if use_none_X:
+        X = None
+
+    pl.fit(X, y)
+    if expected_unique_values:
```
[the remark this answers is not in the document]
freddyaboulton: Yep, removing the if statement!

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -61,21 +60,41 @@ def fit(self, X, y):
...
+        This helper passes y as an argument if needed by the estimator.
+        """
+        return self._estimator_predict_helper(features, y, return_probabilities=False)
+
+    def _estimator_predict_proba(self, features, y):
+        """Get estimator predicted probabilities.
+
+        This helper passes y as an argument if needed by the estimator.
+        """
+        return self._estimator_predict_helper(features, y, return_probabilities=True)
```
dsherry, line 91: @freddyaboulton could we do this
```python
def _estimator_predict(self, features, y):
        """Get estimator predictions.
        This helper passes y as an argument if needed by the estimator.
        """
        y_arg = None
        if self.estimator.predict_uses_y:
            y_arg = y
        return self.estimator.predict(features, y_arg, return_probabilities=False)
```
And similar for predict_proba?

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -172,21 +187,22 @@ def threshold(self, value):
...
             if not objective.is_defined_for_problem_type(self.problem_type):
                 raise ValueError(f"Objective {objective.name} is not defined for time series binary classification.")
 
         if self.threshold is None:
-            predictions = self.estimator.predict(features_no_nan)
+            predictions = self._estimator_predict(features_no_nan, y_encoded)
         else:
-            ypred_proba = self.estimator.predict_proba(features_no_nan).iloc[:, 1]
+            proba = self._estimator_predict_proba(features_no_nan, y_encoded)
+            proba = proba.iloc[:, 1]
```
dsherry, line 201: Why is this necessary? Not covered by `_estimator_predict_proba`?

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -319,6 +322,36 @@ class Pipeline(pipeline_class):
...
+        objectives = ['MCC Binary', "Log Loss Binary"]
+    elif pl.problem_type == ProblemTypes.TIME_SERIES_MULTICLASS:
+        X, y = X_y_multi
+        label_map = {0: "good", 1: "bad", 2: "best"}
+        y = pd.Series(y).map(lambda label: label_map[label])
+        expected_unique_values = {"good", "bad", "best"}
+        objectives = ["MCC Multiclass", "Log Loss Multiclass"]
+
+    if use_none_X:
+        X = None
```
dsherry, line 347: Do we already have coverage elsewhere for the case when `use_none_X` is `True`? I.e. do we need `@pytest.mark.parametrize("use_none_X", [True, False])`?

evalml/utils/gen_utils.py
```diff
@@ -401,19 +401,24 @@ def _not_nan(pd_data):
     return mask
 
 
-def drop_rows_with_nans(pd_data_1, pd_data_2):
+def drop_rows_with_nans(*pd_data):
     """Drop rows that have any NaNs in both pd_data_1 and pd_data_2.
```
dsherry, line 405: Nit-pick: update this docstring too

evalml/pipelines/regression/time_series_baseline_regression.py
dsherry: Ooh that's great!

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -61,21 +60,41 @@ def fit(self, X, y):
         y = self._encode_targets(y)
 
         X_t = self._compute_features_during_fit(X, y)
-        if X_t.empty:
-            raise RuntimeError("Pipeline computed empty features during call to .fit. This means "
-                               "that either 1) you passed in X=None to fit and don't have a DelayFeatureTransformer "
-                               "in your pipeline or 2) you do have a DelayFeatureTransformer but gap=0 and max_delay=0. "
-                               "Please add a DelayFeatureTransformer or change the values of gap and max_delay")
```
dsherry, line 68: Ah got it, great that we can eliminate this!

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -146,13 +166,8 @@ def score(self, X, y, objectives):
         y_encoded = self._encode_targets(y)
         y_shifted = y_encoded.shift(-self.gap)
         y_pred, y_pred_proba = self._compute_predictions(X, y, objectives)
-        non_nan_mask = _get_rows_without_nans(y_shifted, y_pred, y_pred_proba)
-        if y_pred is not None:
-            y_pred = y_pred.iloc[non_nan_mask]
-        if y_pred_proba is not None:
-            y_pred_proba = y_pred_proba.iloc[non_nan_mask]
-        y_labels = y_shifted.iloc[non_nan_mask]
```
dsherry, line 154: Is it necessary to remove the `iloc` stuff from where this code got moved to?

dsherry (approved):
I left a suggestion in the time series classification pipeline, and a couple questions, otherwise 🚢

evalml/pipelines/time_series_classification_pipelines.py
freddyaboulton: Yea, `_estimator_predict_proba` returns the probabilities for all classes. For binary pipelines in `predict`, we threshold the predictions for the "positive" class if the threshold is not None. Hence the need for `.iloc[:, 1]`.

evalml/pipelines/time_series_classification_pipelines.py
freddyaboulton: The calls to `iloc` are now happening in `drop_rows_with_nans`!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
freddyaboulton: Yes I think we do for now. In the other tests with `use_none_X`, we always have a `DelayedFeaturesTransformer` in the pipeline. I think it's important to have coverage when that's not the case. 

That being said, we have coverage for this case for ts regression pipelines in `test_time_series_baseline_regression.py`. In #1580 , I plan on extending the time baseline regressor to classification problems so in that case, I think we can move fold in this coverage into the tests for the baseline.

evalml/utils/gen_utils.py
freddyaboulton: Thank you!

evalml/pipelines/time_series_classification_pipelines.py
freddyaboulton: Yes we can! I originally thought we couldn't because some estimators like `XGBoost` don't have `y` as an argument in their predict signature but the metaclass adds it!

PR #1652, opened 2021-01-05
Remove `text_columns` parameter from LSA and TextFeaturizer components

### LINKED ISSUE #1614: Remove `text_columns` parameter from LSA and TextFeaturizer
Now that we're using Woodwork, we should remove the `text_columns` parameter and automatically infer from natural language columns instead.

angela97lin:
Closes #1614.


evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -273,7 +273,7 @@ def test_make_pipeline_no_column_names(input_type, problem_type):
 def test_make_pipeline_text_columns(input_type, problem_type):
     X = pd.DataFrame({"numerical": [1, 2, 3, 1, 2],
                       "categorical": ["a", "b", "a", "c", "c"],
-                      "text": ["string one", "another", "text for a column", "text string", "hello world"]})
+                      "text": ["string one", "another", "text for a column, this should be a text column!!", "text string", "hello world"]})
```
angela97lin, line 276: Making this longer so that WW automatically detects its a text column. Could also just use `infer_feature_types` heh

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -101,11 +100,11 @@ def transform(self, X, y=None):
         if self._features is None or len(self._features) == 0:
             return X
 
-        text_columns = self._get_text_columns(X)
-        es = self._make_entity_set(X, text_columns)
+        es = self._make_entity_set(X, self._text_columns)
```
freddyaboulton, line 103: In `_make_entity_set`, I think we need to switch from "text" to "natural_language" because "text" is now deprecated:

image

evalml/tests/component_tests/test_lsa.py
```diff
@@ -47,46 +46,15 @@ def test_lsa_no_text():
     assert len(X_t.columns) == 2
 
 
-def test_some_missing_col_names(text_df, caplog):
```
freddyaboulton, line 50: At first, I was confused as to why we were getting rid of the warning but I think it makes sense to do so since the user is no longer provided the text column names as input.

freddyaboulton (approved):
@angela97lin I think this looks good!

chukarsten (approved):
This looks good.

evalml/pipelines/components/transformers/preprocessing/text_transformer.py
```diff
@@ -7,32 +7,21 @@
...
-            if col_name in X.columns:
-                columns.append(col_name)
-            else:
-                missing_columns.append(col_name)
-        if len(columns) == 0:
-            raise AttributeError("None of the provided text column names match the columns in the given DataFrame")
-        if len(columns) < len(self._all_text_columns):
-            logger.warn("Columns {} were not found in the given DataFrame, ignoring".format(missing_columns))
-        return columns
+        text_column_vals = X.select('natural_language')
```
bchen1116, line 25: nice! pretty awesome that we can now use ww's `natural_language` type

evalml/pipelines/utils.py
```diff
@@ -59,7 +58,8 @@ def _get_preprocessing_components(X, y, problem_type, text_columns, estimator_cl
 
     pp_components.append(Imputer)
 
-    if text_columns:
+    text_columns = list(X.select('NaturalLanguage').columns)
```
bchen1116, line 61: Is there a reason we switch from `natural_language` from earlier and `NaturalLanguage` here? It'd be nice to use the same format for consistency

evalml/pipelines/utils.py
```diff
@@ -59,7 +58,8 @@ def _get_preprocessing_components(X, y, problem_type, text_columns, estimator_cl
...
+    if len(text_columns) != 0:
```
bchen1116, line 62: nit-pick, can we say `> 0`?

evalml/tests/component_tests/test_lsa.py
```diff
@@ -97,15 +65,16 @@ def test_invalid_text_column():
             np.nan,
             None,
             'I\'m happy again!!! lalalalalalalalalalala']})
-    lsa = LSA(text_columns=['col_1'])
+    X = infer_feature_types(X, {'col_1': 'NaturalLanguage'})
```
bchen1116, line 68: Does this mean that if a user had this dataframe `X` and didn't run `infer_feature_types`, then ww wouldn't pick up on the column being of type `NaturalLanguage`?

evalml/tests/component_tests/test_lsa.py
```diff
@@ -127,12 +96,12 @@ def test_int_col_names():
...
               'Red, the blood of angry men - black, the dark of ages past']
          })
-    lsa = LSA(text_columns=[4.75, -1])
+    lsa = LSA()
     lsa.fit(X)
     expected_col_names = set(['LSA(4.75)[0]',
                               'LSA(4.75)[1]',
-                              'LSA(-1)[0]',
-                              'LSA(-1)[1]'])
+                              'LSA(-1.0)[0]',
```
bchen1116, line 103: Why did this column label become a float versus the original int?

bchen1116 (commented):
Just left a few questions!

evalml/tests/component_tests/test_lsa.py
angela97lin: Yup, we no longer need to check since we're not relying and checking the user's input :3

evalml/tests/component_tests/test_lsa.py
angela97lin: I believe it does, but in case the heuristic for determining types changes, adding this to assert that we want this to be natural language :d

evalml/tests/component_tests/test_lsa.py
angela97lin: Good question! I believe it should have been a float to begin with:
image

However, because we previously asked users to pass in text col names and then use those col names and we previously passed in -1, we got -1 as an int instead.

bchen1116 (approved):
Nice!

angela97lin:
Blocking on https://github.com/alteryx/evalml/issues/1662 so putting this back as In Progress 😭

dsherry:
@angela97lin BTW since we decided yesterday that this is not ready to merge, I'm converting this back to a draft.

PR #1665, opened 2021-01-07
InvalidDataTargetCheck additions

chukarsten:
- [x] For regression, error if the problem type was not "numeric"

- [ ] For binary, error if the problem type was "categorical" (but first we should confirm that if a categorical column has only two unique values (other than nan) woodwork will label it as binary, not as "categorical" (this was a little unclear - I think this already is the case and is tested.)

- [x] For multiclass, error if the problem type was binary

- [x] For multiclass, warn if the problem type had a high number of unique values--perhaps set a max cap at over 5%.

addresses [Issue #1548 ](https://github.com/alteryx/evalml/issues/1548)

### DIFF
--- a/evalml/data_checks/data_check_message_code.py
+++ b/evalml/data_checks/data_check_message_code.py
@@ -31,7 +31,7 @@ class DataCheckMessageCode(Enum):
     TARGET_MULTICLASS_NOT_ENOUGH_CLASSES = "target_multiclass_not_enough_classes"
     """Message code for target data for a multi classification problem that does not have more than two unique classes."""
 
-    TARGET_MULTICLASS_HIGH_UNIQUE_CLASS_WARNING = "target_multiclass_high_unique_class_warning"
+    TARGET_MULTICLASS_HIGH_UNIQUE_CLASS = "target_multiclass_high_unique_class_warning"
     """Message code for target data for a multi classification problem that has an abnormally large number of unique classes relative to the number of target values."""
 
     HIGH_VARIANCE = "high_variance"
--- a/evalml/data_checks/data_check_message_code.py
+++ b/evalml/data_checks/data_check_message_code.py
@@ -28,6 +28,12 @@ class DataCheckMessageCode(Enum):
     TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS = "target_multiclass_not_two_examples_per_class"
     """Message code for target data for a multi classification problem that does not have two examples per class."""
 
+    TARGET_MULTICLASS_NOT_ENOUGH_CLASSES = "target_multiclass_not_enough_classes"
+    """Message code for target data for a multi classification problem that does not have more than two unique classes."""
+
+    TARGET_MULTICLASS_HIGH_UNIQUE_CLASS_WARNING = "target_multiclass_high_unique_class_warning"
+    """Message code for target data for a multi classification problem that has an abnormally large number of unique classes relative to the number of target values."""
+
     HIGH_VARIANCE = "high_variance"
     """Message code for when high variance is detected for cross-validation."""
 
--- a/evalml/data_checks/invalid_targets_data_check.py
+++ b/evalml/data_checks/invalid_targets_data_check.py
@@ -1,5 +1,5 @@
-import woodwork as ww
 import pandas as pd
+import woodwork as ww
 
 from evalml.data_checks import (
     DataCheck,
@@ -16,6 +16,7 @@ from evalml.utils.gen_utils import (
 )
 
 MULTICLASS_CONTINUOUS_THRESHOLD = .05
+
 
 class InvalidTargetDataCheck(DataCheck):
     """Checks if the target data contains missing or invalid values."""
@@ -118,7 +119,7 @@ class InvalidTargetDataCheck(DataCheck):
                     message_code=DataCheckMessageCode.TARGET_MULTICLASS_NOT_ENOUGH_CLASSES,
                     details=details).to_dict())
 
-            num_class_to_num_value_ratio = len(unique_values)/len(y)
+            num_class_to_num_value_ratio = len(unique_values) / len(y)
             if num_class_to_num_value_ratio >= MULTICLASS_CONTINUOUS_THRESHOLD:
                 details = {"class_to_value_ratio": num_class_to_num_value_ratio}
                 messages["warnings"].append(DataCheckWarning(
@@ -126,7 +127,6 @@ class InvalidTargetDataCheck(DataCheck):
                     data_check_name=self.name,
                     message_code=DataCheckMessageCode.TARGET_MULTICLASS_HIGH_UNIQUE_CLASS_WARNING,
                     details=details).to_dict())
-
 
         if len(value_counts) == 2 and is_supported_type:
             if set(unique_values) != set([0, 1]):
--- a/evalml/data_checks/invalid_targets_data_check.py
+++ b/evalml/data_checks/invalid_targets_data_check.py
@@ -15,11 +15,11 @@ from evalml.utils.gen_utils import (
     numeric_and_boolean_ww
 )
 
-MULTICLASS_CONTINUOUS_THRESHOLD = .05
-
 
 class InvalidTargetDataCheck(DataCheck):
     """Checks if the target data contains missing or invalid values."""
+
+    multiclass_continuous_threshold = .05
 
     def __init__(self, problem_type, objective, n_unique=100):
         """Check if the target is invalid for the specified problem type.
@@ -120,7 +120,7 @@ class InvalidTargetDataCheck(DataCheck):
                     details=details).to_dict())
 
             num_class_to_num_value_ratio = len(unique_values) / len(y)
-            if num_class_to_num_value_ratio >= MULTICLASS_CONTINUOUS_THRESHOLD:
+            if num_class_to_num_value_ratio >= self.multiclass_continuous_threshold:
                 details = {"class_to_value_ratio": num_class_to_num_value_ratio}
                 messages["warnings"].append(DataCheckWarning(
                     message="Target has a large number of unique values, could be regression target.",
--- a/evalml/data_checks/invalid_targets_data_check.py
+++ b/evalml/data_checks/invalid_targets_data_check.py
@@ -98,7 +98,7 @@ class InvalidTargetDataCheck(DataCheck):
                                                      details=details).to_dict())
 
         if self.problem_type == ProblemTypes.REGRESSION and not pd.api.types.is_numeric_dtype(y.to_series()):
-            messages["errors"].append(DataCheckError(message="Target data type should be floating point for regression type problems.",
+            messages["errors"].append(DataCheckError(message="Target data type should be numeric for regression type problems.",
                                                      data_check_name=self.name,
                                                      message_code=DataCheckMessageCode.TARGET_UNSUPPORTED_TYPE,
                                                      details={}).to_dict())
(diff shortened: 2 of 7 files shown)


evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -81,20 +83,20 @@ def test_invalid_target_data_check_multiclass_two_examples_per_class():
     expected_message = "Target does not have at least two instances per class which is required for multiclass classification"
 
     # with 1 class not having min 2 instances
-    assert invalid_targets_check.validate(X, y=pd.Series([0, 1, 1, 2, 2])) == {
+    assert invalid_targets_check.validate(X, y=pd.Series([0] + [1] * 19 + [2] * 80)) == {
```
chukarsten, line 86: This is an edit to make the targets pass the test that looks for targets with large numbers of unique values relative to the total number of target values.  I do this in a few places.

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -87,17 +92,41 @@ def validate(self, X, y):
...
+            if len(unique_values) <= 2:
+                details = {"num_classes": len(unique_values)}
+                messages["errors"].append(DataCheckError(
+                    message="Target does not have more than two classes, which is required for multiclass classification.",
+                    data_check_name=self.name,
+                    message_code=DataCheckMessageCode.TARGET_MULTICLASS_NOT_ENOUGH_CLASSES,
+                    details=details).to_dict())
+
+            num_class_to_num_value_ratio = len(unique_values) / len(y)
+            if num_class_to_num_value_ratio >= self.multiclass_continuous_threshold:
```
bchen1116, line 123: Not sure I'm understanding this. If there's 3 unique values in a list of length 50, this would warn that the problem could be regression? `3/50 == 0.06 > 0.05`. If so, this default seems really high, but definitely open to discuss.

evalml/data_checks/no_variance_data_check.py
```diff
@@ -47,8 +47,8 @@ def _check_for_errors(self, column_name, count_unique, any_nulls):
 
         elif count_unique == 2 and not self._dropnan and any_nulls:
             return DataCheckWarning(message=f"{column_name} has two unique values including nulls. "
-                                            "Consider encoding the nulls for "
-                                            "this column to be useful for machine learning.",
+                                    "Consider encoding the nulls for "
```
bchen1116, line 50: Nice catch!

bchen1116 (commented):
Just left a question

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -18,13 +19,17 @@
 class InvalidTargetDataCheck(DataCheck):
     """Checks if the target data contains missing or invalid values."""
 
+    multiclass_continuous_threshold = .05
+
     def __init__(self, problem_type, objective, n_unique=100):
         """Check if the target is invalid for the specified problem type.
 
         Arguments:
+            problem_type (str or ProblemTypes): The specific problem type to data check for.
```
ParthivNaresh, line 28: Thanks for adding this!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -87,17 +92,41 @@ def validate(self, X, y):
...
+                    message="Target does not have more than two classes, which is required for multiclass classification.",
+                    data_check_name=self.name,
+                    message_code=DataCheckMessageCode.TARGET_MULTICLASS_NOT_ENOUGH_CLASSES,
+                    details=details).to_dict())
+
+            num_class_to_num_value_ratio = len(unique_values) / len(y)
+            if num_class_to_num_value_ratio >= self.multiclass_continuous_threshold:
+                details = {"class_to_value_ratio": num_class_to_num_value_ratio}
+                messages["warnings"].append(DataCheckWarning(
+                    message="Target has a large number of unique values, could be regression target.",
```
ParthivNaresh, line 126: Nitpick `could be a regression type problem` instead of `could be regression target` to match your message above.

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -87,17 +92,41 @@ def validate(self, X, y):
...
                                                      data_check_name=self.name,
                                                      message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_UNIQUE_VALUES,
                                                      details=details).to_dict())
-        if self.problem_type == ProblemTypes.MULTICLASS and value_counts.min() <= 1:
-            least_populated = value_counts[value_counts <= 1]
-            details = {"least_populated_class_labels": least_populated.index.tolist()}
-            messages["errors"].append(DataCheckError(message="Target does not have at least two instances per class which is required for multiclass classification",
+
+        if self.problem_type == ProblemTypes.REGRESSION and not pd.api.types.is_numeric_dtype(y.to_series()):
+            messages["errors"].append(DataCheckError(message="Target data type should be floating point for regression type problems.",
```
ParthivNaresh, line 101: Wouldn't integers be allowed here as well? I believe `is_numeric_dtype` will return True regardless of if the dtype is `Int64` or `float64`. Maybe the message should be `Target data type should be numerical for regression type problems.`

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -314,3 +320,67 @@ def test_invalid_target_data_check_initialize_with_none_objective():
     with pytest.raises(DataCheckInitError, match="Encountered the following error"):
         DataChecks([InvalidTargetDataCheck], {"InvalidTargetDataCheck": {"problem_type": "multiclass",
                                                                          "objective": None}})
+
+
+def test_invalid_target_data_check_regression_problem_nonnumeric_data():
```
ParthivNaresh, line 325: You could parametrize with a non-regression problem type to check the differences if you want to

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -314,3 +320,67 @@ def test_invalid_target_data_check_initialize_with_none_objective():
...
+    y_multiclass = pd.Series([1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3] * 25)
+    y_binary = pd.Series([0, 1, 1, 1, 0, 0] * 25)
+
+    data_check_error = DataCheckError(
+        message=f"Target does not have more than two classes, which is required for multiclass classification.",
+        data_check_name=invalid_targets_data_check_name,
+        message_code=DataCheckMessageCode.TARGET_MULTICLASS_NOT_ENOUGH_CLASSES,
+        details={"num_classes": len(set(y_binary))}).to_dict()
+
+    invalid_targets_check = InvalidTargetDataCheck("multiclass", get_default_primary_search_objective("regression"))
```
ParthivNaresh, line 358: Any reason for `get_default_primary_search_objective("regression")` here?

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -314,3 +320,67 @@ def test_invalid_target_data_check_initialize_with_none_objective():
...
+    assert invalid_targets_check.validate(X, y=y_multiclass) == {"warnings": [], "errors": []}
+    assert invalid_targets_check.validate(X, y=y_binary) == {"warnings": [], "errors": [data_check_error]}
+
+
+def test_invalid_target_data_check_multiclass_problem_almostcontinuous_data():
+    X = pd.DataFrame()
+    invalid_targets_check = InvalidTargetDataCheck("multiclass", get_default_primary_search_objective("multiclass"))
+
+    y_multiclass_high_classes = pd.Series(list(range(0, 100)) * 3)  # 100 classes, 300 samples, .33 class.sample ratio
```
ParthivNaresh, line 367: `class.sample` to `class/sample`

ParthivNaresh (approved):
Looks excellent, just a few wording changes!

evalml/data_checks/invalid_targets_data_check.py
ParthivNaresh: Good point, I think that's why this is a `warning` rather than an `error`, but maybe there could be an additional check taking into account the size of the target data when calculating this ratio?

freddyaboulton (approved):
@chukarsten Thanks for the improvements!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -87,17 +92,41 @@ def validate(self, X, y):
...
+            messages["errors"].append(DataCheckError(message="Binary class targets require exactly two unique values.",
                                                      data_check_name=self.name,
                                                      message_code=DataCheckMessageCode.TARGET_BINARY_NOT_TWO_UNIQUE_VALUES,
                                                      details=details).to_dict())
-        if self.problem_type == ProblemTypes.MULTICLASS and value_counts.min() <= 1:
-            least_populated = value_counts[value_counts <= 1]
-            details = {"least_populated_class_labels": least_populated.index.tolist()}
-            messages["errors"].append(DataCheckError(message="Target does not have at least two instances per class which is required for multiclass classification",
+
+        if self.problem_type == ProblemTypes.REGRESSION and not pd.api.types.is_numeric_dtype(y.to_series()):
```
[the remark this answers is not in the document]
angela97lin: Yes! We should use Woodwork, checking `"numeric" not in y.semantic_tags` should work 😁

evalml/data_checks/invalid_targets_data_check.py
chukarsten: Yup, I think this is a good place for a discussion.  I just implemented what was in the issue as I wasn't present for the discussion - a 5% threshold.  I'm certainly down for a smarter method of doing it.

evalml/data_checks/data_check_message_code.py
```diff
@@ -28,6 +28,12 @@ class DataCheckMessageCode(Enum):
     TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS = "target_multiclass_not_two_examples_per_class"
     """Message code for target data for a multi classification problem that does not have two examples per class."""
 
+    TARGET_MULTICLASS_NOT_ENOUGH_CLASSES = "target_multiclass_not_enough_classes"
+    """Message code for target data for a multi classification problem that does not have more than two unique classes."""
+
+    TARGET_MULTICLASS_HIGH_UNIQUE_CLASS_WARNING = "target_multiclass_high_unique_class_warning"
```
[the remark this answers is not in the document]
chukarsten: You're right!

evalml/data_checks/invalid_targets_data_check.py
chukarsten: Yes!

evalml/data_checks/invalid_targets_data_check.py
chukarsten: Agree.  I definitely meant to put that initially and I got distracted.  Thanks for the catch.

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
chukarsten: So I went ahead and parametrized it...I guess I'm not very familiar with what we ultimately want.  I know in some of my tests I use non-numerics for a categorical, but I don't think we're setup to do that, right?  ["Happy", "Birthday", "Birthday"] isn't considered a multiclass target, but [0, 1, 2, 2, 2, 1] is, right?

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
chukarsten: pasta, thanks

evalml/data_checks/invalid_targets_data_check.py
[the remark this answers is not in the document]
chukarsten: Thanks for the note...I need to get up to date at a higher level on what woodwork is.  Any good reference?

Issue #1662, opened 2021-01-07
Custom woodwork types aren't preserved in components


dsherry:
EDIT: following #1668, updating this issue.

**Problem**
Our components currently do not preserve the input custom woodwork types. After #1668, we updated automl and `ComponentGraph` to keep track of custom Woodwork types, such that if a user overrides a Woodwork type, it will be passed from component to component successfully. However, this is not true on a Component level: if a user passes an overridden Woodwork column to only a component, they will find that the returned Woodwork data structure did not preserve the override. That's odd.

**Symptoms**
User using Component will notice that evalml will reset any overridden Woodwork types.

**Fix**
Update all transformers `transform` to return Woodwork with the original overridden types, if applicable.

Note:
This behavior is relevant only in components where the original column should be preserved. In some cases, such as the standard scaler, type change is expected. In other cases, such as the OHE, the original column is removed and this doesn't apply.

dsherry:
@angela97lin for tracking purposes, did I remember right that you already have a branch for this? Or am I incorrect?

angela97lin:
@dsherry Not yet! I think maybe you were thinking about #1652, which I had a branch up for but is blocking on #1406?

angela97lin:
Not blocking #1652 since the text columns are removed during transformation, and #1668 updates the ComponentGraph to keep track of and update carrying along the text columns within the component graph.

PR #1668, opened 2021-01-08
Update components and pipelines to return Woodwork data structures

### LINKED ISSUE #1406: Update pipeline and components to return Woodwork data structures
#1393 updated pipelines to accept Woodwork data structures, and #1288 addresses updating pipelines and components to accept Woodwork data structures as input. However, the output for methods like `transform` and `predict` are still pandas DataFrames, which is odd. This issue tracks updating our methods to return Woodwork data structures.

angela97lin:
Closes #1406

Not *too* much implementation wise but notes that may be useful:
- Sometimes it doesn't always work to just call `.to_series()`/`.to_dataframe()` on a Woodwork data structure, because the types inferred might not be what we want. Example is: Woodwork could convert to a series w/ Int64. Converting that to numpy would create an array with an object type, which is usually not what we want. 😬 
- Partial dependence method update context: previously, we could pass our pipelines directly to scikit-learn's partial dependence method after a bit of finagling (adding attributes expected). This is no longer true because we're returning Woodwork data structures from predict and scikit-learn doesn't know how to handle that. Thus, I needed to wrap our pipeline in `scikit_learn_wrapped_estimator` (implemented for ensemble), updating the fields now on this wrapped object. IMO kinda cleaner than previous, where we had to delete these ugly attributes we added just for scikit-learn to operate since we don't care for the wrapped obj afterwards :d
- Imputer handling indices and restoring to original was moved to SimpleImputer. This way, we support the same behavior for SimpleImputer and don't need to maintain it in two separate places since the Imputer is just a combination of SimpleImputers.
- I noticed there was a lot of duplicate code in `fit_features` and `compute_final_component_features` so I refactored using a new helper function, `_fit_transform_features_helper`
- Currently, the ComponentGraph object is responsible for keeping track of original logical types passed by a user and transforming the transformed data to the original types as necessary. We will have to update this in #1662 
- Unrelated, but cleaned up docstrings for random state for consistency as I was updating each component.
- A lot of duplicate code between fit/transform and fit_transform for FeatureSelector. Removed fit_transform.
- Updated a lot of tests for consistency. Mock fit should still return self. Mock transforms still need to return WW data structures. 
- Combine `_compute_predictions` for classification and time series classification pipelines. 
- Cleaned up a lot of duplicate impl for components fit/transform/fit_transform and removed if unnecessary.
- Removed `random_state` from OutliersDataCheck. No functional change, but since we're currently not using IsolationForest anymore, this isn't necessary.
- Updated user_guide/components notebook: removed duplicate sections of LinearRegressor code and DropNull code before code generation.

Perf tests graphs (I want to double check on this, but having some issues running the tests):
Time automl takes increased by a bit. In most cases, this was less than a second increase. In the most extreme case, it was a 12 second increase. I think this is expected and not too large.
image

Performance is... odder. I don't know how much of this is due to randomness, and would like to try again but here are the results thus far:
image


docs/source/user_guide/pipelines.ipynb
```diff
@@ -389,8 +392,8 @@
     "\n",
     "    def fit(self, X, y=None):\n",
```
angela97lin, line 393: Updating demo code to be a transformer that handles woodwork :') necessary for fitting of pipeline below!

evalml/model_understanding/graphs.py
```diff
@@ -454,26 +454,27 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
             "positive" class.
 
     """
+    from evalml.pipelines.components.utils import scikit_learn_wrapped_estimator
```
angela97lin, line 457: What's going on here: previously, we could pass our pipelines directly to scikit-learn's partial dependence method after a bit of finagling (adding attributes expected). This is no longer true because we're returning Woodwork data structures from predict and scikit-learn doesn't know how to handle that.

Thus, updating this code to wrap our pipeline in `scikit_learn_wrapped_estimator` (implemented for ensemble), updating the fields now on this wrapped object. 

IMO kinda cleaner than previous, where we had to delete these ugly attributes we added just for scikit-learn to operate since we don't care for the wrapped obj afterwards :d

evalml/model_understanding/graphs.py
```diff
@@ -454,26 +454,28 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
```
angela97lin, line 457: Unfortunately, calling the import in here because otherwise we have circular dependencies. 

Specifically because our objectives calls on graph code which calls on pipelines (to wrap pipelines) which calls on objectives:
```
During handling of the above exception, another exception occurred:
../venv/lib/python3.7/site-packages/_pytest/config/__init__.py:441: in _importconftest
    mod = conftestpath.pyimport()
../venv/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
evalml/__init__.py:8: in <module>
    import evalml.objectives
evalml/objectives/__init__.py:2: in <module>
    from .cost_benefit_matrix import CostBenefitMatrix
evalml/objectives/cost_benefit_matrix.py:6: in <module>
    from evalml.model_understanding.graphs import confusion_matrix
evalml/model_understanding/__init__.py:1: in <module>
    from .graphs import (
evalml/model_understanding/graphs.py:435: in <module>
    from evalml.pipelines.components.utils import scikit_learn_wrapped_estimator
evalml/pipelines/__init__.py:36: in <module>
    from .classification_pipeline import ClassificationPipeline
evalml/pipelines/classification_pipeline.py:5: in <module>
    from evalml.objectives import get_objective
E   ImportError: cannot import name 'get_objective' from 'evalml.objectives' (/Users/angela.lin/Desktop/evalml/evalml/objectives/__init__.py)
```

I think this is okay for this PR... and maybe we can think about restructuring if necessary separately?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -64,19 +63,10 @@ def transform(self, X, y=None):
         self._check_input_for_columns(X)
 
         cols = self.parameters.get("columns") or []
-        return self._modify_columns(cols, X, y)
+        cols = self._modify_columns(cols, X, y)
+        return _convert_to_woodwork_structure(cols)
 
     def fit_transform(self, X, y=None):
-        """Fit transformer to data, then transform data.
```
angela97lin, line 70: No real use in this docstring, might as well have it inherit.

evalml/tests/component_tests/test_target_encoder.py
```diff
@@ -157,20 +157,6 @@ def test_get_feature_names():
     np.testing.assert_array_equal(encoder.get_feature_names(), np.array(['col_1', 'col_2', 'col_3']))
 
 
-def test_custom_indices():
```
angela97lin, line 160: Moved this to `evalml/tests/pipeline_tests/regression_pipeline_tests/test_regression.py` since it was using a pipeline and is beyond just a component test. I'm honestly not too sure what this test is testing for... just that custom indices work don't break pipelines? 🤷

evalml/tests/pipeline_tests/regression_pipeline_tests/test_regression.py
```diff
@@ -19,6 +21,21 @@ def test_invalid_targets_regression_pipeline(target_type, dummy_regression_pipel
...
     X, y = load_diabetes()
-    mock_regression_pipeline = linear_regression_pipeline_class(parameters={'Linear Regressor': {'n_jobs': 1}})
-    mock_regression_pipeline.fit(X, y)
-    assert not pd.isnull(mock_regression_pipeline.predict(X)).any()
+    regression_pipeline = linear_regression_pipeline_class(parameters={'Linear Regressor': {'n_jobs': 1}})
+    regression_pipeline.fit(X, y)
+    assert not pd.isnull(regression_pipeline.predict(X).to_series()).any()
+
+
+def test_custom_indices():
```
angela97lin, line 29: Moved this from target encoder tests. For what the name / test implies, target encoder isn't necessary for testing, so might as well use OHE (since target encoder won't run when core_dependencies=True)

evalml/model_understanding/graphs.py
```diff
@@ -461,19 +461,20 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
         raise ValueError("Pipeline to calculate partial dependence for must be fitted")
     if pipeline.model_family == ModelFamily.BASELINE:
         raise ValueError("Partial dependence plots are not supported for Baseline pipelines")
-    if isinstance(pipeline, evalml.pipelines.ClassificationPipeline):
```
angela97lin, line 464: Previously, we could pass our pipelines directly to scikit-learn's partial dependence method after a bit of finagling (adding attributes expected). This is no longer true because we're returning Woodwork data structures from predict and scikit-learn doesn't know how to handle that. Thus, I needed to wrap our pipeline in scikit_learn_wrapped_estimator (implemented for ensemble), updating the fields now on this wrapped object. IMO kinda cleaner than previous, where we had to delete these ugly attributes we added just for scikit-learn to operate since we don't care for the wrapped obj afterwards :d

evalml/automl/automl_search.py
```diff
@@ -577,10 +577,7 @@ def _tune_binary_threshold(self, pipeline, X_threshold_tuning, y_threshold_tunin
             pipeline.threshold = 0.5
             if X_threshold_tuning:
                 y_predict_proba = pipeline.predict_proba(X_threshold_tuning)
-                if isinstance(y_predict_proba, pd.DataFrame):
```
angela97lin, line 580: Cleanup: we no longer need to worry about whether y_predict_proba is returning np/pd, so no need for this check :D

evalml/pipelines/component_graph.py
```diff
@@ -83,89 +88,95 @@ def fit(self, X, y):
...
-            X (pd.DataFrame): The input training data of shape [n_samples, n_features]
-            y (pd.Series): The target training data of length [n_samples]
+            X (ww.DataTable, pd.DataFrame): The input training data of shape [n_samples, n_features]
+            y (ww.DataColumn, pd.Series): The target training data of length [n_samples]
         """
         self._compute_features(self.compute_order, X, y, fit=True)
         return self
 
     def fit_features(self, X, y):
-        """ Fit all components save the final one, usually an estimator
```
angela97lin, line 93: There was a lot of duplicate code in `fit_features` and` compute_final_component_features` so I refactored using a new helper function, `_fit_transform_features_helper`

evalml/pipelines/component_graph.py
```diff
@@ -356,10 +378,10 @@ def __iter__(self):
         return self
 
     def __next__(self):
-        """Iterator for graphs, prints the components in the graph in order
```
angela97lin, line 359: I don't think it is printing..? So updating docstring, lmk if I missed anything 🤔

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -107,9 +107,10 @@ def fit(self, X, y=None):
     def predict(self, X):
         X_encoded = self._encode_categories(X)
         predictions = super().predict(X_encoded)
-        if self._label_encoder:
-            predictions = pd.Series(self._label_encoder.inverse_transform(predictions.astype(np.int64)))
-        return predictions
+        if not self._label_encoder:
```
angela97lin, line 110: Reorganized this way so that we don't convert to series if not necessary.

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
```diff
@@ -10,61 +14,35 @@ def get_names(self):
...
-        if not isinstance(X_t, pd.DataFrame) and isinstance(X, pd.DataFrame):
-            X_dtypes = X.dtypes.to_dict()
-            selected_col_names = self.get_names()
-            col_types = {key: X_dtypes[key] for key in selected_col_names}
-            return pd.DataFrame(X_t, columns=selected_col_names, index=X.index).astype(col_types)
-        else:
-            return pd.DataFrame(X_t)
+            raise RuntimeError("Feature selector requires a transform method or a component_obj that implements transform")
 
-    def fit_transform(self, X, y=None):
```
angela97lin, line 45: Unnecessary duplicate of fit/transform.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -52,9 +51,6 @@ def fit(self, X, y=None):
         if (X.dtypes == bool).all():
             X = X.astype('category')
 
-        # Convert None to np.nan, since None cannot be properly handled
```
angela97lin, line 55: I think since we do conversion to Woodwork, this isn't necessary anymore as WW handles this :)

evalml/pipelines/components/transformers/preprocessing/__init__.py
```diff
@@ -3,5 +3,5 @@
 from .text_transformer import TextTransformer
 from .lsa import LSA
 from .text_featurizer import TextFeaturizer
-from .time_series import DelayedFeatureTransformer
```
angela97lin, line 6: Renamed to keep consistent w/ rest of our components

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -68,26 +64,26 @@ def transform(self, X, y=None):
...
-            return pd.DataFrame(X_t, columns=X_null_dropped.columns)
+            X_t = pd.DataFrame(X_t, columns=X_null_dropped.columns)
+            return _convert_to_woodwork_structure(X_t)
+
         X_t = pd.DataFrame(X_t, columns=X_null_dropped.columns)
-        if len(category_cols) > 0:
-            X_t[category_cols] = X_t[category_cols].astype('category')
-        return X_t
+        X_t = X_t.infer_objects()
+        X_t.index = X_null_dropped.index
```
angela97lin, line 85: Imputer handling indices and restoring to original was moved to SimpleImputer. This way, we support the same behavior for SimpleImputer and don't need to maintain it in two separate places since the Imputer is just a combination of SimpleImputers.

evalml/pipelines/components/utils.py
```diff
@@ -141,24 +144,25 @@ def predict(self, X):
         """Make predictions using selected features.
 
         Arguments:
-            X (pd.DataFrame): Features
+            X (ww.DataTable, pd.DataFrame): Features
 
         Returns:
-            pd.Series: Predicted values
+            np.ndarray: Predicted values
```
angela97lin, line 150: No change in behavior, just incorrect docstring :p

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -22,11 +23,13 @@ def test_time_series_baseline(mock_decode, pipeline_class, gap, X_none, ts_data)
     clf = pipeline_class(parameters={"pipeline": {"gap": gap, "max_delay": 1},
                                      "Time Series Baseline Estimator": {'gap': gap, 'max_delay': 1}})
     expected_y = y.shift(1) if gap == 0 else y
+    expected_y = expected_y.reset_index(drop=True)
```
angela97lin, line 26: Addresses assert_series_equal checking for index 😦

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -908,9 +909,9 @@ def test_score_multiclass_single(mock_predict, mock_fit, mock_encode, X_y_binary
 @patch('evalml.pipelines.MulticlassClassificationPipeline._encode_targets')
 @patch('evalml.pipelines.MulticlassClassificationPipeline.fit')
 @patch('evalml.pipelines.ComponentGraph.predict')
-def test_score_nonlinear_multiclass(mock_encode, mock_fit, mock_predict, nonlinear_multiclass_pipeline_class, X_y_multi):
+def test_score_nonlinear_multiclass(mock_predict, mock_fit, mock_encode, nonlinear_multiclass_pipeline_class, X_y_multi):
```
angela97lin, line 912: Incorrect order before!

evalml/pipelines/classification_pipeline.py
```diff
@@ -122,17 +123,19 @@ def score(self, X, y, objectives):
...
+        y_predicted, y_predicted_proba = self._compute_predictions(X, y, objectives)
+        if y_predicted is not None:
+            y_predicted = _convert_woodwork_types_wrapper(y_predicted.to_series())
+        if y_predicted_proba is not None:
+            y_predicted_proba = _convert_woodwork_types_wrapper(y_predicted_proba.to_dataframe())
         return self._score_all_objectives(X, y, y_predicted, y_predicted_proba, objectives)
 
-    def _compute_predictions(self, X, objectives):
-        """Scan through the objectives list and precompute"""
+    def _compute_predictions(self, X, y, objectives, time_series=False):
```
angela97lin, line 133: Updated and combined w/ time series classification pipeline implementation!

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
```diff
@@ -73,31 +73,17 @@ def transform(self, X, y=None):
...
+            transformed = imputer.transform(X[[column]]).to_dataframe()
             if transformed.empty:
                 cols_to_drop.append(column)
             else:
-                X_t[column] = transformed
+                X_t[column] = transformed[column]
         X_t = X_t.drop(cols_to_drop, axis=1)
-        return X_t
-
-    def fit_transform(self, X, y=None):
```
angela97lin, line 91: Unnecessary, okay to use the default transformer impl of fit_transform.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -52,42 +51,38 @@ def fit(self, X, y=None):
...
```
angela97lin, line 84: Need this to infer categorical as boolean :)

evalml/pipelines/components/transformers/preprocessing/delayed_feature_transformer.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, max_delay=2, delay_features=True, delay_target=True, gap=1,
 
     def fit(self, X, y=None):
         """Fits the DelayFeatureTransformer."""
+        return self
```
angela97lin, line 47: All fit should return self, even if it doesn't do anything!

evalml/pipelines/components/transformers/scalers/standard_scaler.py
```diff
@@ -16,3 +21,13 @@ def __init__(self, random_state=0, **kwargs):
         super().__init__(parameters=parameters,
                          component_obj=scaler,
                          random_state=random_state)
+
+    def transform(self, X, y=None):
```
angela97lin, line 25: This is necessary because scikit-learn's StandardScaler's transform method doesn't take in a y, so trying to pass it will error.

https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.transform

evalml/tests/automl_tests/test_automl.py
```diff
@@ -415,17 +416,19 @@ def test_automl_bad_data_check_parameter_type():
         automl.search(data_checks=[MockDataCheckErrorAndWarning])
 
 
-@patch('evalml.pipelines.RegressionPipeline.fit')
```
angela97lin, line 418: Hmmmm. I suspect that this isn't actually testing what we want it to. We mock `InvalidTargetDataCheck` to return "R2", and then test that the default objective name is "R2", but the two aren't necessarily related. Let's delete this for now, and I can file an issue to track testing this if I'm understanding this correctly.

evalml/tests/component_tests/test_column_selector_transformers.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import pytest
+from pandas.testing import assert_frame_equal
```
angela97lin, line 4: In general, updated to import rather than calling pandas.testing. assert_frame_equal everywhere. Makes those lines much shorter and readable IMO :D

evalml/tests/component_tests/test_column_selector_transformers.py
```diff
@@ -105,39 +106,22 @@ def test_drop_column_transformer_input_invalid_col_name(class_to_test):
...
 
     transformer = class_to_test(columns=[0, 1, 2, 3])
-    np.testing.assert_allclose(transformer.transform(X).values, answer2)
+    assert_frame_equal(answer2, transformer.transform(X).to_dataframe())
 
     transformer = class_to_test(columns=[])
-    np.testing.assert_allclose(transformer.transform(X).values, answer3)
-
-
-@pytest.mark.parametrize("class_to_test,answers",
```
angela97lin, line 129: I think this test is redundant with the former, since we're using a numpy array, and numpy array will automatically have int col names! So removing and combining.

evalml/tests/component_tests/test_components.py
```diff
@@ -535,85 +535,44 @@ def test_transformer_transform_output_type(X_y_binary):
...
                 assert transform_output.shape[0] == X.shape[0]
                 assert transform_output.shape[1] >= X.shape[1]
-                assert isinstance(transform_output.columns, pd.Index)
+                assert isinstance(transform_output.to_dataframe().columns, pd.Index)
             else:
                 assert transform_output.shape == X.shape
-                assert (transform_output.columns == X_cols_expected).all()
-
-
-def test_estimator_predict_output_type(X_y_binary, helper_functions):
```
angela97lin, line 578: This is specific to estimators, so moving to `test_estimators.py`

evalml/tests/component_tests/test_components.py
```diff
@@ -664,53 +623,6 @@ def __init__(self, parameters=None, component_obj=None, random_state=0):
     est.predict_proba(X)
 
 
-def test_estimator_check_for_fit_with_overrides(X_y_binary):
```
angela97lin, line 667: Similarly, these are specific to estimators, so moving to `test_estimators.py`

chukarsten (commented):
Screen Shot 2021-01-22 at 10 33 06 AM

...Lord give me strength hahaha.

Makefile
```diff
@@ -18,19 +18,19 @@ lint-fix:
 
 .PHONY: test
 test:
-	pytest evalml/ --doctest-modules --doctest-continue-on-failure
```
angela97lin, line 21: Using to test timing. Can remove later 🤷

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
```diff
@@ -82,7 +83,7 @@ def test_stacked_ensemble_n_jobs_negative_one(X_y_binary, logistic_regression_bi
     clf.fit(X, y)
     y_pred = clf.predict(X)
     assert len(y_pred) == len(y)
-    assert not np.isnan(y_pred).all()
+    assert not np.isnan(y_pred.to_series()).all()
```
chukarsten, line 86: I feel like the woodwork structures should have started as a numpy array base class lol

evalml/tests/component_tests/test_stacked_ensemble_regressor.py
```diff
@@ -125,15 +126,15 @@ def test_stacked_fit_predict_regression(X_y_regression, stackable_regressors):
     clf.fit(X, y)
     y_pred = clf.predict(X)
     assert len(y_pred) == len(y)
-    assert isinstance(y_pred, pd.Series)
-    assert not np.isnan(y_pred).all()
+    assert isinstance(y_pred, ww.DataColumn)
+    assert not np.isnan(y_pred.to_series()).all()
```
chukarsten, line 130: I guess this looks like something we should add to our woodwork wishlist.  Shouldn't have to convert this to find out whether it is all Nan.  I forget which, whether numpy or pandas, but one of them has .isnan() defined.

chukarsten (approved):
Angela - wow.  I can't believe you went through all that.  That was a lot of changes and great attention to detail.  My eyes are bleeding a little bit.  Hopefully, next time, we can keep it to a little less of a heroic PR?  I don't see anything in here that will require a re-review though - just some doc strings to maybe touch up on.  Again, great work.

evalml/tests/component_tests/test_stacked_ensemble_regressor.py
angela97lin: Indeed! I think both have some variant: numpy has `isnan` (https://numpy.org/doc/stable/reference/generated/numpy.isnan.html), pandas has `isna` (or also `isnull`?):
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.isna.html

Hoping the new WW API change will help with this though 😀

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
angela97lin: Bahaha I can see why! My guess is that they chose to start with a pandas base since they take advantage of a lot of the new nullable pandas types~

evalml/pipelines/classification_pipeline.py
```diff
@@ -122,17 +123,19 @@ def score(self, X, y, objectives):
         y = _convert_woodwork_types_wrapper(y.to_series())
         objectives = [get_objective(o, return_instance=True) for o in objectives]
         y = self._encode_targets(y)
-        y_predicted, y_predicted_proba = self._compute_predictions(X, objectives)
-
+        y_predicted, y_predicted_proba = self._compute_predictions(X, y, objectives)
+        if y_predicted is not None:
```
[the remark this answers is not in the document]
angela97lin: Nope! I was thinking about this too, but ultimately decided to stick with this. The difference, I think, is what we're expecting from the user / how the user will use it.

If we have _convert_woodwork_types_wrapper(None) return None, we're allowing the user to more freely use this and then having to deal with checking for None later, where None values don't make sense (ex: scoring).

If we disallow _convert_woodwork_types_wrapper(None), then the user needs to more conscious about their input and ensure that its not None. They shouldn't call _convert_woodwork_types_wrapper if their input is None. While updating and keeping this requirement, I caught a few instances where we were accidentally passing in None or forgetting to pass an input, and disallowing None caught it, which is why I ultimately stuck with this but open to more thoughts, or you think this would lead to a better user experience! 😁

evalml/pipelines/components/ensemble/stacked_ensemble_classifier.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, input_pipelines=None, final_estimator=None,
                 None and 1 are equivalent. If set to -1, all CPUs are used. For n_jobs below -1, (n_cpus + 1 + n_jobs) are used.
                 Defaults to None.
                 - Note: there could be some multi-process errors thrown for values of `n_jobs != 1`. If this is the case, please use `n_jobs = 1`.
-            random_state (int, np.random.RandomState): seed for the random number generator
+            random_state (int, np.random.RandomState): Seed for the random number generator. Defaults to 0.
```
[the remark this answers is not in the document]
angela97lin: 🏃‍♀️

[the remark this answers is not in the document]
angela97lin: On a more serious note, feel free to merge @bchen1116! I'll adjust accordingly 😁

evalml/pipelines/classification_pipeline.py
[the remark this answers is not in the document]
chukarsten: Sounds good!

evalml/tests/component_tests/test_estimators.py
```diff
@@ -85,3 +87,91 @@ def test_all_estimators_check_fit_input_type_regression(data_type, X_y_regressio
...
+    y_series_no_name = pd.Series(y_np)
+    y_series_with_name = pd.Series(y_np, name='target')
+    datatype_combos = [(X_np, y_np, range_index, np.unique(y_np)),
+                       (X_np, y_list, range_index, np.unique(y_np)),
+                       (X_df_no_col_names, y_series_no_name, range_index, y_series_no_name.unique()),
+                       (X_df_with_col_names, y_series_with_name, X_df_with_col_names.columns, y_series_with_name.unique())]
+
+    for component_class in _all_estimators_used_in_search():
+        for X, y, X_cols_expected, y_cols_expected in datatype_combos:
+            print('Checking output of predict for estimator "{}" on X type {} cols {}, y type {} name {}'
```
bchen1116, line 109: Do we need this print statement now?

evalml/tests/component_tests/test_lgbm_classifier.py
```diff
@@ -127,12 +128,12 @@ def test_fit_string_features(X_y_binary):
...
 
-    np.testing.assert_almost_equal(y_pred, y_pred_sk, decimal=5)
-    np.testing.assert_almost_equal(y_pred_proba, y_pred_proba_sk, decimal=5)
+    np.testing.assert_almost_equal(y_pred_sk, y_pred.to_series().values, decimal=5)
+    np.testing.assert_almost_equal(y_pred_proba_sk, y_pred_proba.to_dataframe().values, decimal=5)
 
 
-@patch('evalml.pipelines.components.estimators.estimator.Estimator.predict_proba')
-@patch('evalml.pipelines.components.estimators.estimator.Estimator.predict')
+@patch('evalml.pipelines.components.estimators.estimator.Estimator.predict_proba', return_value=ww.DataTable(pd.DataFrame()))
```
bchen1116, line 135: Why did we need to include a `return_value` with the patches?

bchen1116 (commented):
Giant PR! Wow this was a lot of edits. 

There's some issues popping up in AutoMLSearch in the [docs](https://feature-labs-inc-evalml--1668.com.readthedocs.build/en/1668/demos/fraud.html#Search-for-best-pipeline), [here](https://feature-labs-inc-evalml--1668.com.readthedocs.build/en/1668/user_guide/automl.html#AutoML-in-EvalML) as well. I think this was caused through these changes somewhere, since the errors seem to stem from type errors thrown by ww. Would be useful to fix before merging this PR!

Also, do we want to change the [tutorials](https://feature-labs-inc-evalml--1668.com.readthedocs.build/en/1668/user_guide/components.html#Custom-Transformers) in the docs to use and return woodwork instead of pandas? 

Other than that, left a few comments/questions, but looks good! Hopefully I didn't miss any major errors. This PR was a doozy

angela97lin:
@bchen1116 Thank you for reviewing and pointing out the docs issues!! I'll dig into them and see what's up 😁

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -25,7 +25,7 @@ def __init__(self,
             allowed_pipelines (list(class)): A list of PipelineBase subclasses indicating the pipelines allowed in the search. The default of None indicates all pipelines for this problem type are allowed.
             max_iterations (int): The maximum number of iterations to be evaluated.
             tuner_class (class): A subclass of Tuner, to be used to find parameters for each pipeline. The default of None indicates the SKOptTuner will be used.
-            random_state (int, np.random.RandomState): The random seed/state. Defaults to 0.
+            random_state (int, np.random.RandomState): Seed for the random number generator. Defaults to 0.
```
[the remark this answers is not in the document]
angela97lin: o lordy

evalml/tests/component_tests/test_estimators.py
angela97lin: Oo I didn't add this, I just moved this--I believe @dsherry added this a while ago because it helped when testing. I think we should keep it--I also noticed it helped me to see which estimator was erroring out while I was debugging 😁

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -131,7 +131,7 @@ def test_multi_auto(X_y_multi, multiclass_core_objectives):
     best_pipeline = automl.best_pipeline
     assert best_pipeline._is_fitted
     y_pred = best_pipeline.predict(X)
-    assert len(np.unique(y_pred)) == 3
+    assert len(np.unique(y_pred.to_series())) == 3
```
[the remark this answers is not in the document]
gsheni: Yes, this will be addressed with the upcoming Series accessor.

evalml/model_understanding/graphs.py
angela97lin: Yes, thank you for pointing this out! 

I guess the argument against it was, we didn't want to add attributes that scikit-learn had that might not align with what we have, strictly for the sake of calculating partial dependence, for example. Specifically, `_estimator_type` is a bit weird: it makes sense for scikit-learn since they're expecting estimators, but here we're passing in pipelines... so it doesn't make as much sense? (https://scikit-learn.org/stable/modules/generated/sklearn.inspection.partial_dependence.html#sklearn.inspection.partial_dependence)

I think the new implementation helps a bit: it doesn't require us to temporarily set attributes and then delete them. I revisited again and added these attributes to our scikit-learn wrapper instead. This cleans up the partial dependence code as you suggested, but I'd be happy to file an issue though if you think we should still improve upon this / if it still makes sense to add `_estimator_type` 😁

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
angela97lin: Yupperino, `super().predict(X_encoded)` should return a Woodwork object, so we can return it directly. Otherwise, we need to get the series to do some magic, before converting it back to WW (blegh).

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
```diff
@@ -10,61 +14,35 @@ def get_names(self):
...
         except AttributeError:
-            raise RuntimeError("Transformer requires a transform method or a component_obj that implements transform")
-        if not isinstance(X_t, pd.DataFrame) and isinstance(X, pd.DataFrame):
-            X_dtypes = X.dtypes.to_dict()
-            selected_col_names = self.get_names()
-            col_types = {key: X_dtypes[key] for key in selected_col_names}
-            return pd.DataFrame(X_t, columns=selected_col_names, index=X.index).astype(col_types)
-        else:
-            return pd.DataFrame(X_t)
+            raise RuntimeError("Feature selector requires a transform method or a component_obj that implements transform")
```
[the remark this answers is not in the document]
angela97lin: Yup, you're right! Updated this to return `evalml.exceptions.exceptions.MethodPropertyNotFoundError`, since that's more consistent with our other transformers.

evalml/pipelines/components/transformers/preprocessing/delayed_feature_transformer.py
angela97lin: Yup, I believe returning `self` is done in our `ComponentBase` class after some logic! By implementing fit here, I think that fit method was overridden by this one--which originally didn't do anything and didn't return `self` so was therefore inconsistent :)

evalml/tests/component_tests/test_baseline_classifier.py
```diff
@@ -51,10 +54,15 @@ def test_baseline_binary_random(X_y_binary):
     clf = BaselineClassifier(strategy="random", random_state=0)
     clf.fit(X, y)
     assert clf.classes_ == [0, 1]
-    np.testing.assert_allclose(clf.predict(X), get_random_state(0).choice(np.unique(y), len(X)))
+
+    predictions = clf.predict(X)
+    assert_series_equal(pd.Series(get_random_state(0).choice(np.unique(y), len(X)), dtype="Int64"), predictions.to_series())
```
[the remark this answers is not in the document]
angela97lin: I think we're just trying to test that the baseline classifier's `predict` method returns as we'd expect, here for the case where the strategy is "random" (just randomly select from target to use as prediction). I'll break this down into two separate lines, so hopefully it'll be a bit more clear :)

docs/source/user_guide/components.ipynb
```diff
@@ -402,45 +416,6 @@
     "AutoML will perform a search over the allowed ranges for each parameter to select models which produce optimal performance within those ranges. AutoML gets the allowed ranges for each component from the component's `hyperparameter_ranges` class attribute. Any component parameter you add an entry for in `hyperparameter_ranges` will be included in the AutoML search. If parameters are omitted, AutoML will use the default value in all pipelines. "
    ]
   },
-  {
-   "cell_type": "code",
```
angela97lin, line 406: I think this is accidental duplicate code, deleting 😱

docs/source/user_guide/components.ipynb
```diff
@@ -481,60 +455,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# custom component\n",
-    "from evalml.pipelines.components import Transformer\n",
-    "import pandas as pd\n",
+    "# We can also do this for custom components\n",
     "from evalml.pipelines.components.utils import generate_component_code\n",
     "\n",
-    "class MyDropNullColumns(Transformer):\n",
```
angela97lin, line 489: No need to repeat this, I believe the only difference is the name and there's nothing special about this necessary for code gen so deleting!

PR #1676, opened 2021-01-11
Add ability to freeze hyperparameters in AutoMLSearch

### LINKED ISSUE #1625: Make it easy to freeze individual parameters during automl search
**Example**
Suppose I want to run automl, but because of my knowledge about the dataset and the domain in question, I know I want to use the "most_frequent" imputation strategy for filling missing values, rather than letting automl gradually try to find the best imputation strategy. In that case, it would be convenient if automl allowed me to input that parameter setting before the search begins.

**Problem**
This is supported today, but via the following ugly path:
* Run `make_pipelines` by hand to get the pipelines which automl would generate for your problem
* Subclass those pipelines, and for any parameter you'd like to fix to one value, set it to that single value in the `custom_hyperparameter_ranges` for each pipeline
* Run automl with a list of the pipeline classes you created

We can improve on that.

This has come up a few times in the past few months; thanks @bchen1116 for finally pushing me to file it 😆

bchen1116:
fix #1625 

Design discussion [here](https://alteryx.quip.com/9wcHAca7Ioo1/AutoMLSearch-API-Parameters-for-Pipelines)
Updated documentation [here](https://feature-labs-inc-evalml--1676.com.readthedocs.build/en/1676/user_guide/automl.html#Limiting-the-AutoML-Search-Space)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -2088,3 +2089,87 @@ def test_automl_validate_objective(non_core_objective, X_y_regression):
     with pytest.raises(ValueError, match='is not allowed in AutoML!'):
         AutoMLSearch(X_train=X, y_train=y, problem_type=non_core_objective.problem_types[0],
                      additional_objectives=[non_core_objective.name])
+
+
+def test_automl_pipeline_params_simple(X_y_binary):
```
bchen1116, line 2094: Currently, these tests run AutoMLSearch to make sure that the parameters we specified are used. I noticed that it was harder to check for `best_pipeline` parameters, which is what I'm currently using to check, if I mock `.fit()` and `.score()`. If anyone has recommendations for how to potentially speed this up, that would be super helpful! (It currently takes ~9.23 seconds for the three new test functions I added)

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -139,8 +141,20 @@ def _transform_parameters(self, pipeline_class, proposed_parameters):
             if 'number_features' in init_params:
                 component_parameters['number_features'] = self.number_features
             # Pass the pipeline params to the components that need them
-            for param_name, value in self._pipeline_params.items():
-                if param_name in init_params:
-                    component_parameters[param_name] = value
+            if component_class.name in self._pipeline_params:
```
freddyaboulton, line 144: I don't think we should do this because this would overwrite the parameters proposed by the tuner with a random value. I think that defeats our automl tuner because we're no longer using the pipeline scores to inform what the next parameter should be. 

I think the problem is that the default parameters of the pipeline are not necessarily in the new constrained search space. I think we should make sure of that when IterativeAlgorithm [creates the first batch](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L66) but leave all subsequent batches alone.

All subsequent batches will have parameters within the valid ranges because we supplied the user's parameters to make_pipeline (the design document says that if a user passes in custom pipelines, it's assumed they set the right hyperparameter ranges themselves).

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Can you explain the problem with mocking `fit` and `score`? I wouldn't go by `best_pipeline`. I'd mock `fit` and `score` and the analyze the `_results` dict or rankings table to make sure the parameters are within the new ranges.

evalml/automl/automl_algorithm/iterative_algorithm.py
bchen1116: Ah, good call. I'll change it so that it only sets the component parameters if we are in the first batch, and otherwise, it should skip this step.

evalml/tests/automl_tests/test_automl.py
bchen1116: Right, my initial idea was that it would be nice to make sure the final parameters don't include parameters not in our hyperparam ranges, but if we initialize the pipelines correctly, we can assume the tuner will work properly, especially since we've already had tuner tests to cover this. This makes sense! I'll add in the mocking and loop through the results instead of looking at `best_pipeline`

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -138,9 +140,21 @@ def _transform_parameters(self, pipeline_class, proposed_parameters):
...
             if 'number_features' in init_params:
                 component_parameters['number_features'] = self.number_features
-            # Pass the pipeline params to the components that need them
-            for param_name, value in self._pipeline_params.items():
-                if param_name in init_params:
-                    component_parameters[param_name] = value
+            # For first batch, pass the pipeline params to the components that need them
+            if component_class.name in self._pipeline_params and self._batch_number == 0:
+                for param_name, value in self._pipeline_params[component_class.name].items():
+                    if isinstance(value, (Integer, Real)):
```
chukarsten, line 146: That's cool, I didn't know you could do that.

chukarsten (approved):
This looks pretty solid.  I would certainly let someone else take a look at this, because I'm not extremely familiar with the source issue or what's driving this change.  But the way you did it looks good.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -247,3 +246,41 @@ class MockTextClassificationPipeline(BinaryClassificationPipeline):
     assert pipeline.parameters['Text Featurization Component'] == expected_params
     assert isinstance(pipeline[0], TextFeaturizer)
     assert pipeline[0]._all_text_columns == ['text_col_1', 'text_col_2']
+
+
+@pytest.mark.parametrize("parameter", [1, "hello", 1.3, -1.0006, [1, 3, 4], (2, 3, 4)])
+def test_iterative_algorithm_pipeline_params(parameter, dummy_binary_pipeline_classes):
```
freddyaboulton, line 252: Can you run more than one batch in this test and `test_iterative_algorithm_pipeline_params_skopt` ? Can be done in a separate test if you prefer that but I think we should cover that case!

docs/source/user_guide/automl.ipynb
```diff
@@ -358,27 +358,30 @@
...
     "from evalml.model_family import ModelFamily\n",
-    "from evalml.pipelines.utils import make_pipeline\n",
     "import woodwork as ww\n",
     "\n",
     "X, y = load_fraud(n_rows=1000)\n",
     "\n",
     "# example of setting parameter to just one value\n",
-    "custom_hyperparameters = {'Imputer': {\n",
-    "    'numeric_impute_strategy': ['mean']\n",
+    "pipeline_parameters = {'Imputer': {\n",
```
freddyaboulton, line 368: nit-pick: Shouldn't these be called hyperparameters?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2088,3 +2089,60 @@ def test_automl_validate_objective(non_core_objective, X_y_regression):
...
+            assert row['parameters']['Imputer']['numeric_impute_strategy'] == 'most_frequent'
+        if 'Logistic Regression Classifier' in row['parameters']:
+            assert row['parameters']['Logistic Regression Classifier']['C'] == 20
+            assert row['parameters']['Logistic Regression Classifier']['penalty'] == 'none'
+        if 'Elastic Net Classifier' in row['parameters']:
+            assert row['parameters']['Elastic Net Classifier']['alpha'] == 0.75
+            assert row['parameters']['Elastic Net Classifier']['l1_ratio'] == 0.2
+
+
+@patch('evalml.pipelines.RegressionPipeline.fit')
```
freddyaboulton, line 2114: Nice!

freddyaboulton (approved):
@bchen1116 Looks great! I like you this handles all the different ways someone might want to "freeze" some hyperparams.

docs/source/user_guide/automl.ipynb
bchen1116: It is, but it defines both the hyperparameter and parameter ranges. Since I discussed it above this cell with 
```
To limit the search over certain hyperparameter ranges, you can specify a pipeline_parameters argument with your pipeline parameters.
These parameters will also limit the hyperparameter search space. 
```
I figured it would be fine to call it pipeline_parameters, since this was what I would be passing in. Would it make more sense to name it `pipeline_hyperparamters` instead?

bchen1116: Changed!

ParthivNaresh (approved):
Looks great, excellent test coverage

PR #1686, opened 2021-01-12
Update tuner to allow single value parameters

bchen1116:
first step to solve #1625 

We want to allow tuners to accept single values as input; however, we will not add these values to the search space. These values are then treated as fixed values. We use this to help with the implementation for #1625


evalml/tests/tuner_tests/test_skopt_tuner.py
```diff
@@ -129,7 +137,6 @@ def test_skopt_tuner_invalid_parameters_score():
     tuner.add({'Mock Classifier': {'param a': 0, 'param b': 1.0, 'param c': 'option a'}}, np.inf)
     tuner.add({'Mock Classifier': {'param a': 0, 'param b': 1.0, 'param c': 'option a'}}, None)
     tuner.propose()
-    print(random_state)
```
bchen1116, line 132: remove unnecessary print statements

chukarsten:
It seems you have a exception in grid_search_tuner that is uncovered by some of the tests you dropped.  It's only one line but you can find it in the codecov.io coverage changes.

chukarsten (approved):

chukarsten (changes requested):
I think you might just need to address the uncovered exception and this is gucci.

jeremyliweishih (approved):
Looks great to me but could you add coverage making sure that the tuners can `propose` with single values as well?

chukarsten (approved):
Nice, thanks!

PR #1690, opened 2021-01-13
Two-Way Dependence Plots

chukarsten:
… the two way part dep calculated, just trying to integrated it into the plots.

### Pull Request Description

-----

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -470,7 +470,6 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         warnings.warn("There are null values in the features, which will cause NaN values in the partial dependence output. Fill in these values to remove the NaN values.", NullsInColumnWarning)
     try:
         avg_pred, values = sk_partial_dependence(pipeline, X=X, features=features, grid_resolution=grid_resolution)
-        #avg_pred returns 1xgrid_resolution
     finally:
         # Delete scikit-learn attributes that were temporarily set
         del pipeline._estimator_type
@@ -485,16 +484,15 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         data = pd.DataFrame({"feature_values": np.tile(values[0], avg_pred.shape[0]),
                              "partial_dependence": np.concatenate([pred for pred in avg_pred])})
     elif len(features) == 2:
-        # data = pd.DataFrame({"%s values" % features[0]: values[0],
-        #                      "%s values" % features[1]: values[1],
-        #                      "partial_dependence": avg_pred[0]})
         data = pd.DataFrame(avg_pred[0])
-        data.index = features[0]
-        data.columns = features[1]
+        data.index = values[0]
+        data.columns = values[1]
+        data = {"partial_dependence": data,
+                "feature_values": {"x": values[0],
+                                   "y": values[1]}}
 
     if classes is not None:
         data['class_label'] = np.repeat(classes, len(values[0]))
-
     return data
 
 
@@ -533,6 +531,22 @@ def graph_partial_dependence(pipeline, X, features, class_label=None, grid_resol
                         xaxis={'title': f'{feature_name}'},
                         yaxis={'title': 'Partial Dependence'},
                         showlegend=False)
+    if not isinstance(features, (list, tuple)):
+        fig = graph_one_way_part_dep(_go, class_label, feature_name, layout, part_dep, pipeline)
+    else:
+        fig = graph_two_way_part_dep(_go, layout, part_dep)
+
+    return fig
+
+
+def graph_two_way_part_dep(_go, layout, part_dep):
+    trace = _go.Contour(x=part_dep["feature_values"]["x"], y=part_dep["feature_values"]["y"],
+                        z=part_dep["partial_dependence"], name="Partial Dependence")
+    fig = _go.Figure(layout=layout, data=[trace])
+    return fig
+
+
+def graph_one_way_part_dep(_go, class_label, feature_name, layout, part_dep, pipeline):
     if isinstance(pipeline, evalml.pipelines.MulticlassClassificationPipeline):
         class_labels = [class_label] if class_label is not None else pipeline.classes_
         _subplots = import_or_raise("plotly.subplots", error_msg="Cannot find dependency plotly.graph_objects")
@@ -544,7 +558,6 @@ def graph_partial_dependence(pipeline, X, features, class_label=None, grid_resol
         # Don't specify share_xaxis and share_yaxis so that we get tickmarks in each subplot
         fig = _subplots.make_subplots(rows=rows, cols=cols, subplot_titles=class_labels)
         for i, label in enumerate(class_labels):
-
             # Plotly trace indexing begins at 1 so we add 1 to i
             fig.add_trace(_go.Scatter(x=part_dep.loc[part_dep.class_label == label, 'feature_values'],
                                       y=part_dep.loc[part_dep.class_label == label, 'partial_dependence'],
@@ -560,7 +573,6 @@ def graph_partial_dependence(pipeline, X, features, class_label=None, grid_resol
                             name='Partial Dependence',
                             line=dict(width=3))
         fig = _go.Figure(layout=layout, data=[trace])
-
     return fig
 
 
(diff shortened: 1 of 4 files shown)


chukarsten:
Unknown

evalml/model_understanding/graphs.py
```diff
@@ -480,33 +492,46 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
...
-    if classes is not None:
-        data['class_label'] = np.repeat(classes, len(values[0]))
+    if isinstance(features, int) or isinstance(features, str):
+        data = pd.DataFrame({"feature_values": np.tile(values[0], avg_pred.shape[0]),
+                             "partial_dependence": np.concatenate([pred for pred in avg_pred])})
+        if classes is not None:
+            data['class_label'] = np.repeat(classes, len(values[0]))
+    elif isinstance(features, (list, tuple)):
+        if len(features) == 2:
+            data = pd.DataFrame(avg_pred[0])
```
freddyaboulton, line 502: This would only return the partial dependence for the first class for multiclass problems. I think we should either modify this to return the grid for all of the classes or raise an exception that two-way plots are not allowed for multiclass problems.

I don't think modifying this to return the values for all classes would be a heavy lift. I think doing something like this:

```python
df = pd.DataFrame(avg_pred.reshape((-1, avg_pred.shape[-1])))
df.columns = values[1]
df.index = np.tile(values[0], avf_pred.shape[0])
df['classes'] = np.repeat(classes, len(values[0]))
```

should get us most of the way there (though I didn't test 😬 ). The only thing left to modify is how the data is passed to the contour plot.

evalml/model_understanding/graphs.py
```diff
@@ -466,10 +478,10 @@ def partial_dependence(pipeline, X, feature, grid_resolution=100):
     elif isinstance(pipeline, evalml.pipelines.RegressionPipeline):
         pipeline._estimator_type = "regressor"
     pipeline.feature_importances_ = pipeline.feature_importance
-    if ((isinstance(feature, int) and X.iloc[:, feature].isnull().sum()) or (isinstance(feature, str) and X[feature].isnull().sum())):
+    if ((isinstance(features, int) and X.iloc[:, features].isnull().sum()) or (isinstance(features, str) and X[features].isnull().sum())):
         warnings.warn("There are null values in the features, which will cause NaN values in the partial dependence output. Fill in these values to remove the NaN values.", NullsInColumnWarning)
     try:
-        avg_pred, values = sk_partial_dependence(pipeline, X=X, features=[feature], grid_resolution=grid_resolution)
+        avg_pred, values = sk_partial_dependence(pipeline, X=X, features=features, grid_resolution=grid_resolution)
```
freddyaboulton, line 484: Sklearn will raise a value error if features is a list containing both ints and strings. I don't know if we should raise our own to be a little more explicit about what the problem is?

Example:

image

evalml/model_understanding/graphs.py
```diff
@@ -433,25 +433,37 @@ def graph_binary_objective_vs_threshold(pipeline, X, y, objective, steps=100):
...
 
     Returns:
         pd.DataFrame: DataFrame with averaged predictions for all points in the grid averaged
-            over all samples of X and the values used to calculate those predictions. The dataframe will
-            contain two columns: "feature_values" (grid points at which the partial dependence was calculated) and
-            "partial_dependence" (the partial dependence at that feature value). For classification problems, there
-            will be a third column called "class_label" (the class label for which the partial
-            dependence was calculated). For binary classification, the partial dependence is only calculated for the
-            "positive" class.
+            over all samples of X and the values used to calculate those predictions.
```
freddyaboulton, line 455: Solid docstring!

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -892,10 +892,21 @@ def test_partial_dependence_warning(logistic_regression_binary_pipeline_class):
...
+        partial_dependence(pipeline, X, features='a', grid_resolution=20)
+
+
+def test_partial_dependence_errors(logistic_regression_binary_pipeline_class):
+    X = pd.DataFrame({'a': [2, None, 2, 2], 'b': [1, 2, 2, 1], 'c': [0, 0, 0, 0]})
+    y = pd.Series([0, 1, 0, 1])
+    pipeline = logistic_regression_binary_pipeline_class(parameters={"Logistic Regression Classifier": {"n_jobs": 1}})
+    pipeline.fit(X, y)
+
+    with pytest.raises(ValueError) as execinfo:
```
freddyaboulton, line 907: nit-pick: I think pytest.raises(ValueError, match="Too many features") will work too.

freddyaboulton (approved):
@chukarsten Looks great! The only thing I think we should resolve before merge is how we should handle two-way plots for multiclass problems.

evalml/model_understanding/graphs.py
```diff
@@ -516,8 +541,50 @@ def graph_partial_dependence(pipeline, X, feature, class_label=None, grid_resolu
...
+
+    if isinstance(features, (list, tuple)):
+        fig = graph_two_way_part_dep(_go, features, part_dep)
+    else:
+        fig = graph_one_way_part_dep(_go, class_label, features, part_dep, pipeline)
+
+    return fig
+
+
+def graph_two_way_part_dep(_go, features, part_dep):
```
angela97lin, line 554: Are these just meant to be private helper functions for `graph_partial_dependence`? :O If yes, let's prefix with an underscore to indicate that, if no and they're meant to be user-facing, let's add them to the API ref!

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -904,24 +915,43 @@ def test_graph_partial_dependence(test_pipeline):
...
     assert len(fig_dict['data']) == 1
     assert fig_dict['data'][0]['name'] == "Partial Dependence"
 
-    part_dep_data = partial_dependence(clf, X, feature='mean radius', grid_resolution=20)
+    part_dep_data = partial_dependence(clf, X, features='mean radius', grid_resolution=20)
     assert np.array_equal(fig_dict['data'][0]['x'], part_dep_data['feature_values'])
     assert np.array_equal(fig_dict['data'][0]['y'], part_dep_data['partial_dependence'].values)
 
 
+def test_graph_two_way_partial_dependence(test_pipeline):
```
angela97lin, line 930: Amazing!! Yes, RE Freddy's comment, I'd love to see what happens if we pass in a mixed tuple with an int and a string col. Does it work? Let's add a test for it too 😁

angela97lin (approved):
Looking great! Added a few minor comments but I think Freddy's comments are solid.

evalml/model_understanding/graphs.py
chukarsten: I'm legitimately mystified by bumpy's reshape function and even more by people that can so intuitively use it.  Ok, cool, I'll implement this and try and see if I can follow the one-way multi-class example.

bchen1116 (approved):
Agree with Angela and Freddy, but LGTM otherwise!

chukarsten:
Unknown-3

evalml/model_understanding/graphs.py
```diff
@@ -433,30 +433,52 @@ def graph_binary_objective_vs_threshold(pipeline, X, y, objective, steps=100):
...
+
+    Raises:
+        ValueError: if the user provides a tuple of not exactly two features.
+        ValueError: if the provided pipeline isn't fitted.
+        ValueError: if the provided pipeline is a Baseline pipeline.
     """
     X = _convert_to_woodwork_structure(X)
     X = _convert_woodwork_types_wrapper(X.to_dataframe())
 
+    if isinstance(features, (list, tuple)):
```
freddyaboulton, line 476: Awesome

evalml/model_understanding/graphs.py
```diff
@@ -534,21 +580,43 @@ def graph_partial_dependence(pipeline, X, feature, class_label=None, grid_resolu
         # Don't specify share_xaxis and share_yaxis so that we get tickmarks in each subplot
         fig = _subplots.make_subplots(rows=rows, cols=cols, subplot_titles=class_labels)
         for i, label in enumerate(class_labels):
-
             # Plotly trace indexing begins at 1 so we add 1 to i
```
freddyaboulton, line 583: nit-pick: This comment applies to the `col=(i % 2) + 1` expression so I think we move it closer to that or get rid of it

freddyaboulton (approved):
@chukarsten Thanks for adding support for 2-way for multiclass! Tests are solid!

evalml/model_understanding/graphs.py
chukarsten: Just to resolve - this is now gone.  I've merged all back into the original graph_partial_dep().  But that's a good call.

evalml/model_understanding/graphs.py
chukarsten: Thanks!!! Good call.  I must have a bunch of stuff missing in here but I definitely have PR fatigue on this one.

Issue #616, opened 2020-04-10
Improve Fraud Dataset 


gsheni:
- The Fraud Dataset doesn't work well when all the rows and columns are used. The top 2 models generated have only 1 feature. The rest of the models have 66 or 12 features
- I realize this is a synthetic dataset so this may be intended (or not a problem). Feel free to close this issue if that is the case. 

## With 2500 Rows, Some columns (following docs example)
```python
import evalml
from evalml import AutoClassificationSearch
from evalml.objectives import FraudCost

fraud_objective = FraudCost(retry_percentage=.5,
                            interchange_fee=.02,
                            fraud_payout_percentage=.75,
                            amount_col='amount')
X, y = evalml.demos.load_fraud(n_rows=2500)
X = X.drop(['datetime', 'expiration_date', 'country', 'region', 'provider'], axis=1)
X_train, X_holdout, y_train, y_holdout = evalml.preprocessing.split_data(X, y, test_size=0.2, random_state=0)
automl = AutoClassificationSearch(objective=fraud_objective,
                                  additional_objectives=['auc', 'recall', 'precision'],
                                  max_pipelines=5)

automl.search(X_train, y_train)
automl.describe_pipeline(automl.rankings.iloc[0]["id"])
```

## With All Rows, All columns 
```python
X, y = evalml.demos.load_fraud()
X_train, X_holdout, y_train, y_holdout = evalml.preprocessing.split_data(X, y, test_size=0.2, random_state=0)
automl = AutoClassificationSearch(objective=fraud_objective,
                                  additional_objectives=['auc', 'recall', 'precision'],
                                  max_pipelines=5)
automl.search(X_train, y_train)
automl.describe_pipeline(automl.rankings.iloc[0]["id"])
```

PR #1700, opened 2021-01-15
Add infer_feature_types to start.ipynb

### LINKED ISSUE #1667: Add infer_feature_types to start guide
[`infer_feature_types` is the method](https://evalml.alteryx.com/en/stable/generated/evalml.utils.infer_feature_types.html#evalml.utils.infer_feature_types) which allows users to a) infer the type of the target and of each feature in their inputted feature matrix and b) override those types as needed.

Currently, [the AutoML user guide shows an example](https://evalml.alteryx.com/en/stable/user_guide/automl.html#AutoML-in-EvalML) of how to use `infer_feature_types` to override the inferred type with a custom type. We should update [the start guide](https://evalml.alteryx.com/en/stable/start.html) to have similar content.

We currently use the breast cancer dataset in the start guide. I don't think any of the features would be a good example of changing a data type. So, we should either update the start guide to use one of our other demo data sets (like the fraud dataset), or we should create/find a new dataset to use for the start guide.

As discussed with @angela97lin

bchen1116:
fix #1667 

Updated docs [here](https://feature-labs-inc-evalml--1700.com.readthedocs.build/en/1700/start.html)

### DIFF
--- a/docs/source/start.ipynb
+++ b/docs/source/start.ipynb
@@ -45,7 +45,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you’d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
+    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you\u2019d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
+    "\n",
     "- Reformat the `expiration_date` column so it reflects a more familiar date format.\n",
     "- Cast the `lat` and `lng` columns from float to str.\n",
     "- Use `infer_feature_types` to specify what types certain columns should be. For example, to avoid having the `provider` column be inferred as natural language text, we have specified it as a categorical column instead.\n",
--- a/docs/source/start.ipynb
+++ b/docs/source/start.ipynb
@@ -38,14 +38,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "X, y = evalml.demos.load_fraud(n_rows=1000)"
+    "X, y = evalml.demos.load_fraud(n_rows=1000, return_pandas=True)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you\u2019d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
+    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you’d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
     "- Reformat the `expiration_date` column so it reflects a more familiar date format.\n",
     "- Cast the `lat` and `lng` columns from float to str.\n",
     "- Use `infer_feature_types` to specify what types certain columns should be. For example, to avoid having the `provider` column be inferred as natural language text, we have specified it as a categorical column instead.\n",
--- a/docs/source/start.ipynb
+++ b/docs/source/start.ipynb
@@ -45,7 +45,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


docs/source/start.ipynb
```diff
@@ -37,7 +38,32 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# convert X to pandas dataframe in order for us to change the data\n",
+    "X = X.to_dataframe()\n",
+    "X['expiration_date'] = X['expiration_date'].apply(lambda x: '20{}-01-{}'.format(x.split(\"/\")[1], x.split(\"/\")[0]))\n",
+    "X[['lat', 'lng']] = X[['lat', 'lng']].astype('str')\n",
+    "X = infer_feature_types(X, feature_types= {'store_id': 'categorical',\n",
```
angela97lin, line 61: I see lat, lng being manipulated here but not mentioned at all in the markdown. Do we need this change?

docs/source/start.ipynb
```diff
@@ -37,7 +38,32 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "X, y = evalml.demos.load_breast_cancer()"
+    "X, y = evalml.demos.load_fraud(n_rows=1000)"
```
angela97lin, line 41: You can do `evalml.demos.load_fraud(n_rows=1000, return_pandas=True)` to avoid having to call `.to_dataframe()`

angela97lin (approved):
Left comments on how to clean things up a bit, otherwise LGTM! I'm curious about the runtime and the performance though, and haven't looked at docs. Let's confirm that doc builds don't take too much longer (since fraud is more involved this is likely), and that we don't get terrible results (I just have https://github.com/alteryx/evalml/issues/616) in mind.

chukarsten (approved):
+1 for Angela's suggestions

docs/source/start.ipynb
ParthivNaresh: @bchen1116 
You could perhaps provide a brief list just describing what you're doing here:

First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you’d like to change the types inferred by EvalML, you can use the infer_feature_types utility method. Here's what we're going to do:
- Reformat the `expiration_date` column so it reflects a more familiar date format.
- Cast the `lat` and `lng` columns from `float` to `str`.
- Use `infer_feature_types` to specify what types certain columns should be. For example, to avoid having the column `provider` be inferred as one based on text, we have specified it as `categorical` instead.

The infer_feature_types utility method takes a pandas or numpy input and converts it to a Woodwork data structure, providing you with flexibility!

ParthivNaresh (approved):
Everything looks good! Just a little change to the markdown to provide some clarity.

bchen1116:
@angela97lin the doc builds ([here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/535728/), [here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/535726/), and [here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/535592/)) for this PR take than a little longer than other PRs ([here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/535702/) and [here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/535601/)) but not much longer. I think it should be fine to merge.

docs/source/start.ipynb
```diff
@@ -37,7 +38,36 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you\u2019d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
+    "\n",
+    "- Reformat the `expiration_date` column so it reflects a more familiar date format.\n",
+    "- Cast the `lat` and `lng` columns from float to str.\n",
+    "- Use `infer_feature_types` to specify what types certain columns should be. For example, to avoid having the `provider` column be inferred as natural language text, we have specified it as a categorical column instead.\n",
+    "\n",
+    "The `infer_feature_types` utility method takes a p1andas or numpy input and converts it to a [Woodwork](https://woodwork.alteryx.com/en/stable/) data structure, providing you with flexibility to cast the data as necessary. "
```
angela97lin, line 54: Typo!

docs/source/start.ipynb
```diff
@@ -37,7 +38,36 @@
...
    "source": [
-    "X, y = evalml.demos.load_breast_cancer()"
+    "X, y = evalml.demos.load_fraud(n_rows=1000, return_pandas=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, let's clean the data. Since EvalML accepts a pandas input, it can run type inference on this data directly. If you\u2019d like to change the types inferred by EvalML, you can use the `infer_feature_types` utility method. Here's what we're going to do with the following dataset:\n",
```
angela97lin, line 48: Really nit-picky / we don't do a great job at this but for readability, can we stick to a more consistent POV 😝  Either second or third is fine but I think it reads easier than jumping back and forth

angela97lin (commented):
Yup, build time doesn't seem too much longer! Added a few more comments but otherwise 🚢

PR #1701, opened 2021-01-15
Update CatBoost estimators to return self in fit rather than underlying model for consistency

angela97lin:
Updates CatBoost estimators to return self in fit rather than underlying model for consistency. Our other estimators return self, catboost should too.

This fell out of work I've been doing w/ components and pipelines returning Woodwork and general cleanup of EvalML 😛

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -28,6 +28,7 @@ Release Notes
         * Fixed bug where time series baseline estimators were not receiving ``gap`` and ``max_delay`` in ``AutoMLSearch`` :pr:`1645`
         * Fixed jupyter notebooks to help the RTD buildtime :pr:`1654`
         * Added ``positive_only`` objectives to ``non_core_objectives`` :pr:`1661`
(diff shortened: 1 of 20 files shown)


chukarsten (approved):
This looks good.  I was trying to see if we had any tests that locked down what an estimator child class returns for it's methods, and couldn't find any right off the bat.  Shoot me a message on slack if that exists.  I think it's important that we have some tests for these functions, like feature_importance() that I think need to return numpy arrays but we either don't check for it or test for it.

ParthivNaresh (approved):
Looks good!

bchen1116 (approved):
LGTM! Nice catch!

angela97lin:
@chukarsten I think you're right in that our other tests don't explicitly check that, for example, the return type of `fit` is self. Feature importance is a little less obvious but I think in most cases we have `test_feature_importance` which checks using numpy's `np.testing.assert_almost_equal` method whether or not our feature importance method returns the same values as using the external library directly, for example. That check will yell if its not comparing two numpy arrays and therefore helps us check if the return value is the expected numpy array :d

PR #1704, opened 2021-01-19
Add prophet regressor

### LINKED ISSUE #1499: Time series regression: add Prophet estimator
We have done preliminary work in #1766 but did not merge as `pystan` has a `GPL3` license. We can move this issue forward once we solve the licensing issue.

- Add ProphetRegressot component
- Adding ProphetRegressor to AutoML
- Run performance tests to verify behavior
- Add installation instructions to docs (windows and linux)

https://facebook.github.io/prophet/

jeremyliweishih:
Fixes #1499.

- [ ] File PR to add to AutoML and update installation docs.


evalml/utils/gen_utils.py
```diff
@@ -514,3 +514,38 @@ def save_plot(fig, filepath=None, format='png', interactive=False, return_filepa
 
     if return_filepath:
         return filepath
+
+
+class suppress_stdout_stderr(object):
```
jeremyliweishih, line 519: Prophet doesn't have a flag for verbosity so we need this context manager to remove their training output:

https://github.com/facebook/prophet/issues/223

requirements.txt
```diff
@@ -9,3 +9,5 @@ matplotlib>=3.3.3
 graphviz>=0.13
 seaborn>=0.11.1
 category_encoders>=2.0.0
+pystan
```
jeremyliweishih, line 12: I left this open for now until we know what versions work with evalml. I can also just use latest versions for now.

evalml/tests/component_tests/test_components.py
```diff
@@ -798,14 +798,18 @@ def test_all_transformers_check_fit(X_y_binary):
...
-def test_all_estimators_check_fit(X_y_binary, test_estimator_needs_fitting_false, helper_functions):
-    X, y = X_y_binary
+def test_all_estimators_check_fit(X_y_binary, ts_data, test_estimator_needs_fitting_false, helper_functions):
     estimators_to_check = [estimator for estimator in _all_estimators() if estimator not in [StackedEnsembleClassifier, StackedEnsembleRegressor, TimeSeriesBaselineEstimator]] + [test_estimator_needs_fitting_false]
     for component_class in estimators_to_check:
         if not component_class.needs_fitting:
             continue
 
         component = helper_functions.safe_init_component_with_njobs_1(component_class)
+        if component.supported_problem_types == [ProblemTypes.TIME_SERIES_REGRESSION]:
```
jeremyliweishih, line 808: Prophet does not follow sklearn's estimator API (and this is our first estimator that only does time series problem types) and tests in `test_components.py` break so I've added some work arounds here for now. Happy to discuss any of these changes.

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,102 @@
...
+        prophet = import_or_raise("fbprophet", error_msg=p_error_msg)
+
+        prophet_parameters = copy.copy(parameters)
+
+        prophet_regressor = prophet.Prophet(**prophet_parameters)
+        super().__init__(parameters=parameters,
+                         component_obj=prophet_regressor,
+                         random_state=random_state)
+
+    def build_prophet_df(self, X, y=None):
```
jeremyliweishih, line 57: Prophet requires input to be a single dataframe with a `ds` column for dates and `y`for targets.

.circleci/config.yml
```diff
@@ -163,6 +163,15 @@ jobs:
               conda config --add channels conda-forge
               conda activate curr_py
               conda install numba -q -y
+      - run:
```
jeremyliweishih, line 166: Need this for windows installation. In the PR to add prophet to automl we should also include installation instructions in our docs.

jeremyliweishih:
PR should be good to go. Windows py3.6 tests hang on automl tests for some reason though. I'll take a look into it.

evalml/tests/component_tests/test_components.py
freddyaboulton: Nice! I think it's fine to have this code outside a util function since it's only used twice and only in our tests.

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,102 @@
...
+        # check for datetime column
+        if 'ds' in X.columns:
+            date_col = X['ds']
+        elif isinstance(X.index, pd.DatetimeIndex):
+            date_col = X.reset_index()
+            date_col = date_col['index']
+        else:
+            date_col = X.select_dtypes(include='datetime')
```
freddyaboulton, line 65: I think this is good but I have these concerns:

1. I'm don't think we should use `select_dtypes` because there could be multiple datetime columns in the dataframe. 
2. I'm not sure if we should automatically assume the `ds` column is the right date column. It would force users to rename their data prior to search (and users wouldn't know to do that unless they knew how we implemented this). 

I prefer to add a `date_column` parameter on init rather than guessing and being wrong. When it comes to automl, the user can specify the value in the `problem_configuration` and maybe we just need to update `make_pipeline` to not add prophet if there isn't a date column.

The one thing to watch out for is the `X=None` case. I think in that case, we'd have to assume the index of y is dates and if not raise a ValueError.

What do you think?

evalml/tests/component_tests/test_prophet_regressor.py
```diff
@@ -0,0 +1,79 @@
+import pytest
```
freddyaboulton, line 1: I think we need coverage for the `X=None` case, which would require tweaking how `build_prophet_df` works.

freddyaboulton (approved):
@jeremyliweishih This is great! I think the only thing I want to resolve before merge is the behavior of `build_prophet_df`. 

I can't wait to get this into automl. As a sneak peak, I tried it on the temperature data:

```python
from evalml.model_understanding.graphs import graph_prediction_vs_actual_over_time

temperature_data = pd.read_csv("/Users/freddy.boulton/sources/ts_regression_datasets/daily-min-temperatures.csv")
y = temperature_data.pop('Temp')

X = temperature_data.rename(columns={'Date': 'ds'})
X_train, X_test = X.iloc[:(X.shape[0] - 365)], X.iloc[-365:]
y_train, y_test = y.iloc[:(X.shape[0] - 365)], y.iloc[-365:]

class ProphetPipeline(TimeSeriesRegressionPipeline):
    component_graph = [ProphetRegressor]

prophet = ProphetPipeline({'pipeline': {'gap': 0, 'max_delay': 0}})

prophet.fit(X_train, y_train)

graph_prediction_vs_actual_over_time(prophet, X_test, y_test, X_test['ds'])
```

image

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,102 @@
...
+                      "seasonality_prior_scale": seasonality_prior_scale,
+                      "holidays_prior_scale": holidays_prior_scale,
+                      "seasonality_mode": seasonality_mode}
+
+        parameters.update(kwargs)
+
+        p_error_msg = "prophet is not installed. Please install using `pip install pystan` and `pip install fbprophet`."
+        prophet = import_or_raise("fbprophet", error_msg=p_error_msg)
+
+        prophet_parameters = copy.copy(parameters)
```
bchen1116, line 50: Is there a reason we do this extra step?

evalml/tests/component_tests/test_components.py
```diff
@@ -849,8 +857,7 @@ def test_no_fitting_required_components(X_y_binary, test_estimator_needs_fitting
                 component.transform(X, y)
 
 
-def test_serialization(X_y_binary, tmpdir, helper_functions):
-    X, y = X_y_binary
+def test_serialization(X_y_binary, ts_data, tmpdir, helper_functions):
```
bchen1116, line 860: I think we should also add prophet to `test_describe_component()` to make sure the name and params appear as expected

evalml/tests/component_tests/test_prophet_regressor.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_problem_types():
+    assert set(ProphetRegressor.supported_problem_types) == {ProblemTypes.TIME_SERIES_REGRESSION}
+
+
+def test_fit_predict_ts(ts_data):
+    X, y = ts_data
+
+    def build_prophet_df(X, y=None):
```
bchen1116, line 23: Agree with Freddy, could be nice to make the `build_prophet_df` method a util method so you can use in both implmentation and tests

evalml/tests/component_tests/test_prophet_regressor.py
```diff
@@ -0,0 +1,79 @@
...
+
+    assert (y_pred == y_pred_p).all()
+
+
+def test_fit_predict_no_ds(X_y_binary):
+    X, y = X_y_binary
+
+    clf = ProphetRegressor()
+    with pytest.raises(ValueError):
+        clf.fit(X, y)
```
bchen1116, line 79: Might also be nice to add tests where multiple columns are datetime types, but would require the implementation to change slightly to handle this.

bchen1116 (commented):
Looks good! Left a few comments, but agree with Freddy that it might be nice to add a 'column' arg to Prophet's `init` so we don't have to guess which column to use, esp if there are multiple datetime cols

evalml/utils/gen_utils.py
jeremyliweishih: @freddyaboulton  the sktime source grabbed it from the issue so I think linking the issue is fine here

https://github.com/facebook/prophet/issues/223#issuecomment-326455744

.circleci/config.yml
```diff
@@ -163,6 +163,15 @@ jobs:
...
+          name: Install cython and pystan (for prophet)
+          command: |
+              C:\Users\circleci\Miniconda3\shell\condabin\conda-hook.ps1
+              conda config --add channels conda-forge
+              conda activate curr_py
+              conda install libpython m2w64-toolchain -c msys2
+              conda install numpy cython -c conda-forge
+              python -m pip install pystan
```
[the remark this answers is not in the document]
jeremyliweishih: @freddyaboulton [pystan's installation instructions](https://pystan.readthedocs.io/en/latest/windows.html#steps) recommends installing the its dependencies first using conda before installing pystan. I opted to use pip to install pystan but I can shift it to conda as well.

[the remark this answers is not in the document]
freddyaboulton: Darn, I was hoping `conda install pystan` would take care of everything hehe. In that case, happy to leave this as is!

jeremyliweishih:
Still stuck on windows py3.6 tests taking 1hr+. Related to this issue(https://github.com/facebook/prophet/issues/122) but it doesn't seem like their suggestions are working. Will keep working on it.

PR #1708, opened 2021-01-20
Integrate `pandas` version 1.2.0

### LINKED ISSUE #1618: Integrate `pandas` version 1.2.0
When `pandas` version 1.2.0 was released our unit tests across the board broke. This was a rather large update with some API changes and dependency changes. This issue tracks the following:

- investigating what needs to be changed within evalml to integrate pandas 1.2.0 
- fixing evalml unit tests to work with pandas 1.2.0
- uncapping the the pandas version in `core-requirements.txt`

angela97lin:
Closes #1618

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
@@ -3,7 +3,7 @@ from evalml.utils.gen_utils import (
     _convert_to_woodwork_structure,
     _convert_woodwork_types_wrapper
 )
-
+import pandas as pd
 
 def _extract_year(col, encode_as_categories=False):
     return col.dt.year, None
@@ -99,7 +99,7 @@ class DateTimeFeaturizer(Transformer):
             for feature in features_to_extract:
                 name = f"{col_name}_{feature}"
                 features, categories = self._function_mappings[feature](X_t[col_name], self.encode_as_categories)
-                X_t[name] = features
+                X_t[name] = pd.Series(features)
                 if categories:
                     self._categories[name] = categories
         return X_t.drop(self._date_time_col_names, axis=1)
--- a/evalml/pipelines/components/transformers/preprocessing/time_series.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series.py
@@ -92,7 +92,7 @@ class DelayedFeatureTransformer(Transformer):
                 col = X[col_name]
                 if col_name in categorical_columns:
                     col = X_categorical[col_name]
-                X = X.assign(**{f"{col_name}_delay_{t}": col.shift(t) for t in range(1, self.max_delay + 1)})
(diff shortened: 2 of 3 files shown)


evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -215,17 +215,17 @@ def _check_equality(data, expected, check_index_type=True):
         pd.testing.assert_frame_equal(data, expected, check_index_type)
 
 
-@pytest.mark.parametrize("data,num_to_pad,expected",
-                         [(pd.Series([1, 2, 3]), 1, pd.Series([np.nan, 1, 2, 3])),
+@pytest.mark.parametrize("data, num_to_pad, expected",
```
bchen1116, line 218: nit: I didn't realize you can add spaces here, but I'd personally prefer no spaces to keep the formatting consistent with the other tests we have in other files

bchen1116 (commented):
If I'm understanding this correctly, pandas' `value_counts` method now first orders them by frequency (in descending order), then to break ties, it takes the order of occurrence into account? 

Does that mean in your example in the description, `currency_NAD`, `currency_HTG`, and `currency_PAB` occur as frequently as `currency_MOP`, `currency_MUR`, `currency_NIS` in the new encoding, but that the first 3 occur AFTER the second 3 in the original input data? 

Just wanted to check I am understanding this properly.

angela97lin:
@bchen1116 Correct! `value_counts` always sorted by frequency first but now breaks ties based on their occurrence in the original data, so the test cases that required updates are when the frequencies were equivalent.

bchen1116 (approved):
@angela97lin got it! looks good to me then!

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -215,17 +215,17 @@ def _check_equality(data, expected, check_index_type=True):
...
+                         [(pd.Series([1, 2, 3]), 1, pd.Series([np.nan, 1, 2, 3], dtype="Float64")),
```
gsheni, line 219: You can also do `dtype=pd.Float64Dtype()` to make it clear that you are using the new Nullable Float Dtype/.

Note that would be different than `dtype=float64` (regular, pandas float64, not new dtype)

chukarsten (approved):

evalml/tests/utils_tests/test_gen_utils.py
angela97lin: Do you know if it's any different from using the string or is it more stylistic in nature? I was more inclined to use the string version since that's primarily what we've been doing for "Int64" but could file something to update our instances to use an instance of the actual pandas dtype object :)

gsheni: Not sure if it's any different, it might be more stylistic in nature. You can also do `pd.Int64Dtype()`

PR #1711, opened 2021-01-20
Pin LightGBM version to remove bug from docs

### LINKED ISSUE #1659: LightGBM error in `cost_benefit_matrix` demo
When we build our docs we see the following:

```
reading sources... [  0%] demos/cost_benefit_matrix
[LightGBM] [Fatal] Check failed: (best_split_info.right_count) > (0) at /__w/1/s/python-package/compile/src/treelearner/serial_tree_learner.cpp, line 661 .
[LightGBM] [Fatal] Check failed: (best_split_info.right_count) > (0) at /__w/1/s/python-package/compile/src/treelearner/serial_tree_learner.cpp, line 661 .
````

This error is reflecting on our docs [on main](https://evalml.alteryx.com/en/main/demos/cost_benefit_matrix.html#AutoML-Search-with-Cost-Benefit-Matrix ) and it seems like the last version that this didn't show up on was version [0.15.0](https://evalml.alteryx.com/en/v0.15.0/demos/cost_benefit_matrix.html#AutoML-Search-with-Cost-Benefit-Matrix).

There seems to be a relevant issue on LightGBM as well.

### PR DESCRIPTION
**Pull request by bchen1116**
fix #1659 

Versions 3.1.0 and above raise the bug we see in the docs, which isn't reproducible elsewhere. The bug fix, which they filed [here](https://github.com/microsoft/LightGBM/issues/3679) and addressed [here](https://github.com/microsoft/LightGBM/pull/3694), hasn't been released yet.

Using the current [branch](https://github.com/microsoft/LightGBM/issues/3679#issuecomment-751684070) to build the docs in circleci (through ssh), our doc build passes without any errors. This suggests that once LightGBM releases a new version, we can unpin the LGBM version and remove the bug.
image

In comparison, this is what the bug looks like (using LightGBM version 3.1.1)
image

Updated docs [here](https://feature-labs-inc-evalml--1711.com.readthedocs.build/en/1711/demos/cost_benefit_matrix.html#AutoML-Search-with-Log-Loss)

**EDIT**
Here's a stacktrace of the docs during the build:
image
Looks like the booster update for LGBM fails, causing it to raise this error. 

We don't get the same error when running the code locally in Jupyter. We do get the error when building the docs locally, which means it could be an error caused through Sphinx? But I think the easiest solution is to wait for the newest release.

### CODE REVIEW

File: evalml/tests/dependency_update_check/latest_dependency_versions.txt [hunk 1]
```diff
@@ -6,7 +6,7 @@ featuretools==0.23.0
 graphviz==0.16
 ipywidgets==7.6.3
 kaleido==0.1.0
-lightgbm==3.1.1
```
freddyaboulton, line 9: I don't think we need to update this? This just tracks the latest published version, not the latest version allowed in our requirements.

docs/source/release_notes.rst
```diff
@@ -37,6 +37,7 @@ Release Notes
         * Added labeling to ``graph_confusion_matrix`` :pr:`1632`
         * Rerunning search for ``AutoMLSearch`` results in a message thrown rather than failing the search, and removed ``has_searched`` property :pr:`1647`
         * Changed tuner class to allow and ignore single parameter values as input :pr:`1686`
+        * Capped LightGBM version limit to remove bug in docs :pr:`1711`
```
freddyaboulton, line 40: Any idea why this is only reproducible in the docs?

freddyaboulton (approved):
@bchen1116 Looks good! I think the only thing is that we need to remember to uncap this once the latest version is published 😆

evalml/tests/dependency_update_check/latest_dependency_versions.txt
bchen1116: Ohh did not know that. I will change this, thanks!

docs/source/release_notes.rst
bchen1116: Unsure why, but when I ran locally through Jupyter, I didn't run into this bug, even when switching versions for LightGBM. I can look a little more tomorrow and see if there's a reason though.

angela97lin (approved):
Thank you for doing this! 👍 If there isn't already an issue filed for this, can you file an issue to add support for >=3.1.0 so we can keep track? :D

docs/source/release_notes.rst
angela97lin: Maybe looking at the entire stack trace will be helpful?

bchen1116:
Filed new issue to track unpinning the version [here](https://github.com/alteryx/evalml/issues/1713)

docs/source/release_notes.rst
bchen1116: Updated the more complete stacktrace in the description box. I suspect something in nbsphinx might be causing the bug to appear, but I'm unable to repro it in Jupyter.

PR #1719, opened 2021-01-21
Remove random_state arg from get_pipelines in AutoMLSearch

### LINKED ISSUE #1638: Get rid of the `random_state` argument in `get_pipeline`
Currently, the `best_pipeline` property in `AutoMLSearch` will [always return](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L530) a pipeline with `random_state=0` because that's the default value for `random_state` in the `get_pipeline` method.

I think this is confusing because if I create an `AutoMLSearch` object with `random_state=5`, for example, I expect all of the pipelines created and returned by `AutoMLSearch` to have `random_state=5`.

I think we should remove the `random_state` argument from `get_pipeline` and instead make sure that the random state is set to whatever it was when the pipeline was fit.

bchen1116:
fix #1638


evalml/tests/automl_tests/test_automl.py
```diff
@@ -2180,3 +2180,16 @@ def test_automl_pipeline_params_kwargs(mock_fit, mock_score, X_y_multi):
...
+
+@pytest.mark.parametrize("random_state", [0, 1, 9, np.random.RandomState(100)])
+@patch('evalml.pipelines.MulticlassClassificationPipeline.score')
+@patch('evalml.pipelines.MulticlassClassificationPipeline.fit')
+def test_automl_pipeline_random_state(mock_fit, mock_score, random_state, X_y_multi):
+    X, y = X_y_multi
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='multiclass', random_state=random_state, n_jobs=1)
+    automl.search()
+    if isinstance(random_state, int):
+        random_state = automl.random_state
```
freddyaboulton, line 2193: @bchen1116 I thought the plan was to create the pipeline with `random_state=4` if the user passed in 4 as the random state to `AutoMLSearch`. This is creating a pipeline with `np.random.RandomState(4)`, no?

bchen1116: @freddyaboulton the end result here is the same, right? If the user passes in `random_state=4` to AutoMLSearch, we set the `random_state` of AutoMLSearch to be `np.random.RandomState(4)`. If we initialize a pipeline with `random_state=4`, the pipeline's `random_state` will then also be initialized as `np.random.RandomState(4)` since it just uses the int as a seed. 

I used this as a shortcut since the end values would be equivalent. Let me know what you think!

freddyaboulton: @bchen1116 My concern is that the state of `self.random_state` changes throughout the lifetime of `AutoMLSearch`. So the pipeline is not actually being initialized with `np.random.RandomState(4)`. 

image

I think this means we should save the original value of the `random_state` parameter before `random_seed` is created. I think this is another reason why I don't like `np.random.RandomState` hehehe

bchen1116: Got it, this makes sense. Thanks for the clarification!

chukarsten: Good catch @freddyaboulton.  I'll re-review after you adjust, Bryan.

bchen1116:
Will need to resolve [this issue](https://github.com/alteryx/evalml/issues/1725) to remove support for `np.random.RandomState` before coming back to this PR

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2177,3 +2177,16 @@ def test_automl_pipeline_params_kwargs(mock_fit, mock_score, X_y_multi):
         if 'Decision Tree Classifier' in row['parameters']:
             assert 0.1 < row['parameters']['Decision Tree Classifier']['ccp_alpha'] < 0.5
             assert row['parameters']['Decision Tree Classifier']['max_depth'] == 1
+
+
+@pytest.mark.parametrize("random_state", [0, 1, 9])
```
freddyaboulton, line 2182: So fresh and so clean. Awesome.

freddyaboulton (approved):
@bchen1116 Awesome! 🤩

docs/source/release_notes.rst
```diff
@@ -36,6 +36,7 @@ Release Notes
         * Fixed stacking argument ``n_jobs`` for IterativeAlgorithm :pr:`1706`
         * Updated CatBoost estimators to return self in ``.fit()`` rather than the underlying model for consistency :pr:`1701`
         * Added ability to initialize pipeline parameters in ``AutoMLSearch`` constructor :pr:`1676`
+        * Remove ``random_state`` arg from ``get_pipelines`` in ``AutoMLSearch`` :pr:`1719`
```
angela97lin, line 39: Remove --> Removed

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2177,3 +2177,16 @@ def test_automl_pipeline_params_kwargs(mock_fit, mock_score, X_y_multi):
...
+@patch('evalml.pipelines.MulticlassClassificationPipeline.score')
+@patch('evalml.pipelines.MulticlassClassificationPipeline.fit')
+def test_automl_pipeline_random_state(mock_fit, mock_score, random_state, X_y_multi):
+    X, y = X_y_multi
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='multiclass', random_state=random_state, n_jobs=1)
+    automl.search()
+
+    for i, row in automl.rankings.iterrows():
+        if 'Base' not in list(row['parameters'].keys())[0]:
```
angela97lin, line 2191: O same question as other PR LOL, why the check for `Base` here :o

angela97lin (approved):
Noice 👍

evalml/tests/automl_tests/test_automl.py
bchen1116: The baseline pipeline isn't initialized with a `random_state` [arg](https://github.com/alteryx/evalml/blob/4b3bfe4d000da71973d4044ab39623bb478723c5/evalml/automl/automl_search.py#L635)

PR #1721, opened 2021-01-21
Allow pipelines from AutoMLSearch to be pickled 

### LINKED ISSUE #1400: Cannot pickle pipelines from automl
Trying to pickle a pipeline created by automl.search give this error 

```
AttributeError: Can't pickle local object 'make_pipeline.<locals>.GeneratedPipeline'
```

Reproduce with any dataset

```python
import evalml
import pandas as pd
import pickle 

df = pd.read_csv("/Users/james.kanter/Downloads/housing.csv")
automl = evalml.AutoMLSearch(problem_type="regression", max_iterations=2)
X = df.copy()
y = X.pop("SalePrice")
automl.search(X, y)
pickle.dumps(automl.best_pipeline)
```

This came up because I was trying to deploy a model using https://github.com/bentoml/BentoML/ which relied on pickle in the example I was trying to use.

bchen1116:
fix #1400 

The issue arises from `pipeline_class(id)` returning a `make_pipelines.<locals>.GeneratedPipeline` object. For objects, pickle only accepts classes defined on a global scale, which `GeneratedPipeline` did not satisfy since it was defined in the `make_pipeline` method. I introduced the stand-alone classes `GeneratedPipelineBinary, ...` in order to make these pipelines pickle-able.

Saving pipeline:
image

Loading and running pipeline (in a different kernel):
image


evalml/tests/automl_tests/test_automl.py
```diff
@@ -1264,15 +1268,15 @@ class Pipeline2(DummyPipeline):
     if objective.name.lower() == "cost benefit matrix":
         automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type_value, max_iterations=3,
                               allowed_pipelines=[Pipeline1, Pipeline2], objective=objective(0, 0, 0, 0),
-                              additional_objectives=[], n_jobs=1)
+                              additional_objectives=[], train_best_pipeline=False, n_jobs=1)
```
bchen1116, line 1271: Mocking fit for time series regression resulted in the following error when trying to fit the best pipeline during training: `AttributeError: 'NoneType' object has no attribute 'fit'`. The pipeline didn't have a `self._component_obj` value.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1268,7 +1272,7 @@ class Pipeline2(DummyPipeline):
     elif problem_type_value == ProblemTypes.TIME_SERIES_REGRESSION:
         automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type_value, max_iterations=3,
                               allowed_pipelines=[Pipeline1, Pipeline2], objective=objective,
-                              additional_objectives=[], problem_configuration={'gap': 0, 'max_delay': 0}, n_jobs=1)
+                              additional_objectives=[], problem_configuration={'gap': 0, 'max_delay': 0}, train_best_pipeline=False, n_jobs=1)
```
bchen1116, line 1275: Mocking fit for time series regression resulted in the following error when trying to fit the best pipeline during training: `AttributeError: 'NoneType' object has no attribute 'fit'`. The pipeline didn't have a `self._component_obj` value. Would really appreciate input on why it wouldn't work for only time series regression

freddyaboulton: Was the best pipelines the baseline pipeline? The baseline estimator does not have a [component object ](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/estimators/regressors/time_series_baseline_estimator.py#L45)

bchen1116: Interestingly enough, when the best_pipeline is the 'Time Series Baseline Estimator', the tests pass, but when it's `Pipeline1` or `Pipeline2`, with the estimators being the 'MockTimeSeriesRegressor', they fail. That mock also sets `component_obj` to None, which might be why it's failing.

chukarsten (approved):
Looks good to me!

evalml/pipelines/generate_pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+class GeneratedPipelineTimeSeriesMulticlass(TimeSeriesMulticlassClassificationPipeline):
+    """Generated Pipeline class for Time Series Multiclass Classification Pipelines
+    """
+    custom_name = ''
+    component_graph = []
+    custom_hyperparameters = None
+
+
+class GeneratedPipelineTimeSeriesRegression(TimeSeriesRegressionPipeline):
+    """Generated Pipeline class for Time Series Regression Pipelines
```
ParthivNaresh, line 54: Is it necessary to add `custom_name`, `component_graph`, and `custom_hyperparameters` to each of these classes, since those values are defaulted as None in `pipeline_base`?

evalml/pipelines/utils.py
```diff
@@ -140,6 +148,26 @@ class GeneratedPipeline(base_class):
     return GeneratedPipeline
 
 
+def get_generated_pipeline(problem_type):
+    """Returns the class for the generated pipeline based on the problem type
+
+    Arguments:
+        problem_type (Problem_Type): The problem_type that the pipeline is for
```
ParthivNaresh, line 155: Nitpick but I think this should be `problem_type (ProblemTypes)`

evalml/pipelines/utils.py
```diff
@@ -140,6 +148,26 @@ class GeneratedPipeline(base_class):
...
+        return GeneratedPipelineBinary
+    elif problem_type == ProblemTypes.MULTICLASS:
+        return GeneratedPipelineMulticlass
+    elif problem_type == ProblemTypes.REGRESSION:
+        return GeneratedPipelineRegression
+    elif problem_type == ProblemTypes.TIME_SERIES_REGRESSION:
+        return GeneratedPipelineTimeSeriesRegression
+    elif problem_type == ProblemTypes.TIME_SERIES_BINARY:
+        return GeneratedPipelineTimeSeriesBinary
+    else:
```
ParthivNaresh, line 167: Instead of having the catch all be `GeneratedPipelineTimeSeriesMulticlass`, I'd recommend checking for it specifically (`elif problem_type == ProblemTypes.TIME_SERIES_MULTICLASS:`) and then raising an error with `else` specifying that there's an unknown `ProblemTypes`

evalml/pipelines/generate_pipeline.py
```diff
@@ -0,0 +1,58 @@
+from .binary_classification_pipeline import BinaryClassificationPipeline
```
angela97lin, line 1: Nit-picking, but can we call this file `generated_pipelines` per all the classes?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2180,3 +2184,34 @@ def test_automl_pipeline_params_kwargs(mock_fit, mock_score, X_y_multi):
...
+        X, y = X_y_regression
+        pipeline = GeneratedPipelineRegression
+
+    a = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type)
+    a.search()
+
+    for i, row in a.rankings.iterrows():
+        if 'Baseline' not in list(row['parameters'].keys())[0]:
+            assert a.get_pipeline(row['id']).__class__ == pipeline
+            assert pickle.loads(pickle.dumps(a.get_pipeline(row['id'])))
```
ParthivNaresh, line 2217: Looks good!

evalml/pipelines/utils.py
```diff
@@ -140,6 +148,26 @@ class GeneratedPipeline(base_class):
     return GeneratedPipeline
 
 
+def get_generated_pipeline(problem_type):
```
angela97lin, line 151: Can we call this `get_generated_pipeline_class` to make it more clear that we're returning a class, not an instance?

evalml/pipelines/utils.py
```diff
@@ -140,6 +148,26 @@ class GeneratedPipeline(base_class):
...
+    elif problem_type == ProblemTypes.MULTICLASS:
+        return GeneratedPipelineMulticlass
+    elif problem_type == ProblemTypes.REGRESSION:
+        return GeneratedPipelineRegression
+    elif problem_type == ProblemTypes.TIME_SERIES_REGRESSION:
+        return GeneratedPipelineTimeSeriesRegression
+    elif problem_type == ProblemTypes.TIME_SERIES_BINARY:
+        return GeneratedPipelineTimeSeriesBinary
+    else:
+        return GeneratedPipelineTimeSeriesMulticlass
```
angela97lin: Maybe it'd be to define a dictionary mapping of problem type to class and then just return the value, rather than all of these if conditions? :P Also works fine now that else returns `GeneratedPipelineTimeSeriesMulticlass` but we're more likely to run into trouble if/when we add a new pipeline class lol

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -714,3 +717,29 @@ def test_automl_supports_time_series_classification(mock_binary_fit, mock_multi_
 
         assert result['parameters']['Delayed Feature Transformer'] == configuration
         assert result['parameters']['pipeline'] == configuration
+
+
+@pytest.mark.parametrize('problem_type', [ProblemTypes.TIME_SERIES_MULTICLASS, ProblemTypes.TIME_SERIES_BINARY])
```
ParthivNaresh, line 722: Why are the `TimeSeries` checks in `test_automl_search_classification` and `test_automl_search_regression` but the non-TimeSeries in `test_automl`?

evalml/automl/automl_search.py
```diff
@@ -862,7 +862,11 @@ def get_pipeline(self, pipeline_id, random_state=0):
         parameters = pipeline_results.get('parameters')
         if pipeline_class is None or parameters is None:
             raise PipelineNotFoundError("Pipeline class or parameters not found in automl results")
-        return pipeline_class(parameters, random_state=random_state)
+        pipeline = get_generated_pipeline(self.problem_type)
+        pipeline.custom_hyperparameters = pipeline_class.custom_hyperparameters
+        pipeline.custom_name = pipeline_class.name
+        pipeline.component_graph = pipeline_class.component_graph
+        return pipeline(parameters, random_state=random_state)
```
ParthivNaresh, line 869: Nicely done!

ParthivNaresh (approved):
Looks great, just a few comments and questions about organization and docstrings, nothing blocking

docs/source/api_reference.rst
```diff
@@ -133,6 +133,12 @@ Pipeline Base Classes
     TimeSeriesBinaryClassificationPipeline
     TimeSeriesMulticlassClassificationPipeline
     TimeSeriesRegressionPipeline
+    GeneratedPipelineBinary
```
angela97lin, line 136: Curious, do you think it make sense to include these in our API reference? I ask because I'm not really sure they provide too much value on their own lol. Are these classes basically just implementation detail for enabling pickling?

angela97lin: Specifically, I ask this because if a user were to interface with `GeneratedPipelineBinary` as they get different pipelines, they'd notice that all of its attributes are changing. That's a bit odd. I'd argue we'd even want to make this private 🤷

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: Will fix!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2180,3 +2184,34 @@ def test_automl_pipeline_params_kwargs(mock_fit, mock_score, X_y_multi):
...
+
+    elif problem_type == ProblemTypes.REGRESSION:
+        X, y = X_y_regression
+        pipeline = GeneratedPipelineRegression
+
+    a = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type)
+    a.search()
+
+    for i, row in a.rankings.iterrows():
+        if 'Baseline' not in list(row['parameters'].keys())[0]:
```
angela97lin: Does this mean we can't check for Baseline? Or why this conditional here?

evalml/pipelines/generate_pipeline.py
bchen1116: I think it might be necessary, because otherwise, `component_graph` and the other values become interpreted as abstract methods:
`TypeError: Can't instantiate abstract class GeneratedPipelineRegression with abstract methods component_graph`

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -714,3 +717,29 @@ def test_automl_supports_time_series_classification(mock_binary_fit, mock_multi_
...
+@patch('evalml.pipelines.TimeSeriesMulticlassClassificationPipeline.score')
+@patch('evalml.pipelines.TimeSeriesBinaryClassificationPipeline.score')
+@patch('evalml.pipelines.TimeSeriesMulticlassClassificationPipeline.fit')
+@patch('evalml.pipelines.TimeSeriesBinaryClassificationPipeline.fit')
+def test_automl_time_series_classification_pickle_generated_pipeline(mock_binary_fit, mock_multi_fit, mock_binary_score, mock_multiclass_score,
+                                                                     problem_type, X_y_binary, X_y_multi):
+    if problem_type == ProblemTypes.TIME_SERIES_BINARY:
+        X, y = X_y_binary
+        pipeline = GeneratedPipelineTimeSeriesBinary
+        problem_type = 'time series binary'
```
angela97lin, line 758: I think you can remove this and the following `problem_type = 'time series multiclass'` and just pass it in from your fixture, no?

evalml/tests/automl_tests/test_automl.py
angela97lin: Could be useful to augment these tests with checking that the right parameters/component_graph/custom_name/custom_hyperparameters etc are equivalent? 😁

angela97lin (commented):
Looking good! Nothing blocking, but left some cleanup suggestions.

I don't think we should publicly share `GeneratedPipeline*` classes. If a user were to mess around with those classes as they called `get_pipeline`, they'd see the class change, which is confusing, and the classes aren't super useful on their own?

```
pipeline_ = automl.get_pipeline(0)
print (GeneratedPipelineBinary.component_graph)
pipeline_ = automl.get_pipeline(1)
print (GeneratedPipelineBinary.component_graph) # Different!
```

Also, I don't think this can pickle stacking ensemble pipelines (which probably makes sense since we have an object as a parameter). Could be addressable or out of scope for this PR (let's file something). Regardless, we should make that clear.

Food for thought: How does this interact with automl/pipeline's `load`/`save` methods? If a user wants to save their pipelines, what should they be using--those methods, or just pickle? Can we make this clear, whether via docs or by combining the two methods somehow?

angela97lin:
@bchen1116 Oops I see you've updated while I was reviewing so some of the comments might not show up on the files! LMK if anything is confusing :)

docs/source/api_reference.rst
bchen1116: Yeah, it's probably not necessary. I added it just in case, but I can remove it!

evalml/tests/automl_tests/test_automl.py
bchen1116: Ah, not sure why this is here. Removing!

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -305,3 +309,42 @@ def test_automl_supports_time_series_regression(mock_fit, mock_score, X_y_regres
...
+
+    a = AutoMLSearch(X_train=X, y_train=y, problem_type='regression')
+    a.search()
+    a.add_to_rankings(RegressionPipelineCustoms({}))
+    seen_name = False
+    for i, row in a.rankings.iterrows():
+        automl_pipeline = a.get_pipeline(row['id'])
+        assert automl_pipeline.__class__ == pipeline
+        assert pickle.loads(pickle.dumps(automl_pipeline))
+        if automl_pipeline.custom_name == RegressionPipelineCustoms.custom_name:
```
bchen1116, line 333: Testing here to check that `custom_name` and other attributes appear as expected.

bchen1116:
@angela97lin updated the [docs](https://feature-labs-inc-evalml--1721.com.readthedocs.build/en/1721/user_guide/automl.html#Saving-AutoMLSearch-and-pipelines-from-AutoMLSearch) to include info about saving pipelines and about incompatibility with stacked ensembler

PR #1731, opened 2021-01-25
t-SNE Model Understanding

### LINKED ISSUE #1709: Integrate t-SNE into Model Understanding
We'd like to be able to leverage t-SNE as a tool for visualizing high dimensional data into `model_understanding`.

https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html

ParthivNaresh:
Fixes #1709


chukarsten (changes requested):
I think all this looks good to me.  I'm not 100% clear on tsne, but it looks like dimensionality reduction and when I .show()'ed the figs it seemed consistent.  

One thing that other graph tests have done is test that the underlying algorithm's data is present in the fig_dict.  ala `fig_dict_data['data'][0]['values'] == tsne(params)` I think just to be consistent with the others you should probably throw that in here.  Everything else looks excellent.

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,59 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
                         yaxis={'title': 'Target Values and Predictions'})
 
     return _go.Figure(data=data, layout=layout)
+
+
+def t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric='euclidean'):
+    """Get the transformed output after fitting X to the embedded space.
+
+     Arguments:
+        X (ww.DataTable, pd.DataFrame): Features.
```
bchen1116, line 900: If you accept ww.DataTable as X input, you should convert `to_dataframe()` before you call TSNE, otherwise it throws an error. 
image

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,59 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+        local minimum, increasing the learning rate may help.
+        metric (str, optional): The metric to use when calculating distance between instances in a feature array.
+        marker_line_width (int, optional): Determines the line width of the marker boundary.
+        marker_size (int, optional): Determines the size of the marker.
+
+    Returns:
+        np.ndarray (n_samples, n_components)
+
+    """
+    X = X.to_dataframe() if isinstance(X, ww.DataTable) else np.array(X)
```
bchen1116, line 934: So we either pass in a numpy array or a pandas dataframe? What if we just call `_convert_to_woodwork_type` and `_convert_woodwork_types_wrapper`, like we do in components, in order to handle the multiple input types?

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,59 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+
+    Returns:
+        np.ndarray (n_samples, n_components)
+    """
+    t_sne_ = TSNE(n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric)
+    X_new = t_sne_.fit_transform(X)
+    return X_new
+
+
+def graph_t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric='euclidean', marker_line_width=2, marker_size=7):
```
bchen1116, line 916: Not sure if I'm missing something, but I don't see anything graphed in both my Jupyter notebook and jupyter lab, even after I run `make installdeps`. 
image

Are you able to get it to show in both notebook and lab? Can you post a screenshot of the resulting graph?
Also, it takes over a minute to finish this method, whereas `t_sne` was fairly quick. Is this the same for you?

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,59 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
                         yaxis={'title': 'Target Values and Predictions'})
 
     return _go.Figure(data=data, layout=layout)
+
+
+def t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric='euclidean'):
```
bchen1116, line 896: not sure if necessary, but might be useful to either add that this only supports numeric data, or have an error thrown if a user tries to pass in non-numeric data, just so the error message they get is coming form evalml rather than sklearn.

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -1417,3 +1418,18 @@ def test_visualize_decision_trees(fitted_tree_estimators, tmpdir):
     src = visualize_decision_tree(estimator=est_reg, filled=True, max_depth=2)
     assert src.format == 'pdf'
     assert isinstance(src, graphviz.Source)
+
+
+def test_t_sne():
```
bchen1116, line 1423: can you add tests changing the values of the other `graph_t_sne` and `t_sne` parameters? and seeing if those will still work as we expect?

bchen1116 (commented):
Left a few comments. Main issue is that I can't graph and see what the graph looks like. Also, would be nice to add this to the `api_references.rst` doc and `model_understanding.ipynb` so when we release this, users can see the docs and see the usage.

evalml/model_understanding/graphs.py
ParthivNaresh: @bchen1116 Should look something like this, and I'm able to get the results instantly but I'm also using a small dataset. I'm sure the time increases exponentially with the dataset size

Screen Shot 2021-01-26 at 1 22 23 PM

chukarsten (approved):
LGTM

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,73 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+        import_or_raise("ipywidgets", warning=True)
+
+    if not marker_line_width >= 0:
+        raise ValueError("The parameter marker_line_width must be non-negative")
+    if not marker_size >= 0:
+        raise ValueError("The parameter marker_size must be non-negative")
+
+    X_embedded = t_sne(X, n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric, **kwargs)
+
+    xx, yy = np.meshgrid(X, X)
```
freddyaboulton, line 952: This doesn't work as expected if X is a Data Table. This causes a stack trace in line `marker_color=color`. I vote we get rid of this if it's only used for `marker_color` though. 

image

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,73 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+    t_sne_ = TSNE(n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric, **kwargs)
+    X_new = t_sne_.fit_transform(X)
+    return X_new
+
+
+def graph_t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric='euclidean', marker_line_width=2, marker_size=7, **kwargs):
+    """Visualization tool for high dimensional data.
+
+    Arguments:
+        X (np.ndarray, pd.DataFrame, ww.DataTable): Data to be transformed.
```
freddyaboulton, line 927: nit-pick: I think we should say that the data must be all numeric features.

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,73 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+
+    X = _convert_to_woodwork_structure(X)
+    X = _convert_woodwork_types_wrapper(X.to_dataframe())
+    t_sne_ = TSNE(n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric, **kwargs)
+    X_new = t_sne_.fit_transform(X)
+    return X_new
+
+
+def graph_t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric='euclidean', marker_line_width=2, marker_size=7, **kwargs):
+    """Visualization tool for high dimensional data.
```
freddyaboulton, line 924: nit-pick: I think we should make this for specific like, "Plot projection of data onto lower dimensional space with TSNE algorithm" or something like that?

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -1418,3 +1420,76 @@ def test_visualize_decision_trees(fitted_tree_estimators, tmpdir):
...
+def test_graph_t_sne(data_type, perplexity, learning_rate):
+    go = pytest.importorskip('plotly.graph_objects', reason='Skipping plotting test because plotly not installed')
+    if data_type == 'np':
+        X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    elif data_type == 'pd':
+        X = pd.DataFrame([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    elif data_type == 'ww':
+        X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+        X = _convert_to_woodwork_structure(X)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
```
freddyaboulton, line 1486: This is converting back to dataframe so we're not covering the data table case.

freddyaboulton (changes requested):
@ParthivNaresh Looks great! The one thing I want to resolve before merge is the `marker_color=color` line in `graph_t_sne`. I think it's slowing us down. The rest are non-blocking nitpicks on docstrings and a minor test change.

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,73 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+
+    X_embedded = t_sne(X, n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric, **kwargs)
+
+    xx, yy = np.meshgrid(X, X)
+    color = xx.ravel()
+    fig = _go.Figure()
+    fig.add_trace(_go.Scatter(
+        x=X_embedded[:, 0], y=X_embedded[:, 1],
+        mode='markers',
+        marker_color=color
```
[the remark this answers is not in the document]
ParthivNaresh: Good idea, I agree

evalml/tests/model_understanding_tests/test_graphs.py
ParthivNaresh: Good catch

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +891,68 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+
+    """
+    _go = import_or_raise("plotly.graph_objects", error_msg="Cannot find dependency plotly.graph_objects")
+
+    if not marker_line_width >= 0:
+        raise ValueError("The parameter marker_line_width must be non-negative")
+    if not marker_size >= 0:
+        raise ValueError("The parameter marker_size must be non-negative")
+
+    X_embedded = t_sne(X, n_components=n_components, perplexity=perplexity, learning_rate=learning_rate, metric=metric, **kwargs)
```
freddyaboulton, line 948: Maybe we should hard-code `n_components=2` and remove it as a function parameter? I think it would avoid errors in `n_components=1` and I don't know if we should let users pass >=3 components if we're only going to plot the first two anyways?

freddyaboulton (approved):
Looks great @ParthivNaresh !

PR #1732, opened 2021-01-25
Use same CV data splitter for stacked ensembling model created for AutoML

### LINKED ISSUE #1592: Stacked ensembler: use same CV data splitter as the rest of automl

angela97lin:
Closes #1592 

We cannot use `TrainingValidationSplit` for the stacked ensembling component's CV, as it raises:
```TypeError: split() takes from 2 to 3 positional arguments but 4 were given```

This is because the signature for `split()` should be: 
```def split(self, X, y=None, groups=None)```
(https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/model_selection/_split.py)

This PR adds `group` as a parameter for TrainingValidationSplit's `split`, even though it is not used, to maintain consistency with its base class.

Still, the larger issue is that because TrainingValidationSplit only creates one split, we get this error even after making the aforementioned change: `ValueError: cross_val_predict only works for partitions`

Tracking this down, I believe this is caused because the one split means that we may not see some data as part of the training split (only exists in test split). Unrelated yet related SO post: https://stackoverflow.com/questions/43275506/scikit-learn-cross-val-predict-only-works-for-partitions

As a result of this, I am purposely disallowing any data splitters where the number of splits < 2.

Note: opened #1746 to track longer term plans and ideas for improving the stacking ensemble's CV split.

### CODE REVIEW

File: evalml/preprocessing/data_splitters/training_validation_split.py [hunk 1]
```diff
@@ -32,15 +32,17 @@ def get_n_splits():
         """Returns the number of splits of this object"""
         return 1
 
-    def split(self, X, y=None):
+    def split(self, X, y=None, groups=None):
```
angela97lin, line 35: Updated for consistency with BaseCrossValidator, which this subclasses 😬

evalml/pipelines/utils.py
```diff
@@ -253,26 +253,30 @@ def generate_pipeline_code(element):
...
+            If set to None, will use the default stacking ensemble data split. Defaults to None.
+        problem_type (ProblemType): Problem type of pipeline
         n_jobs (int or None): Integer describing level of parallelism used for pipelines.
             None and 1 are equivalent. If set to -1, all CPUs are used. For n_jobs below -1, (n_cpus + 1 + n_jobs) are used.
             Defaults to -1.
 
     Returns:
         Pipeline with appropriate stacked ensemble estimator.
     """
+    if data_splitter is not None and data_splitter.get_n_splits() < 2:
```
freddyaboulton, line 273: This is going to raise `ValueError` in `AutoMLSearch` when the dataset is large or the user manually passes in a `TrainingValidationSplit`.That would stop the search early and show a stacktrace so I think we should do something about it in `AutoMLSearch`. I think there are two things we can do:

1. I think we should set `ensembling=False` under-the-hood in `AutoMLSearch` and log a message if the user passes in a data splitter with not enough splits.
2. Raise a value error in init if the data splitter does not have enough splits.

I think I prefer 1, that's kind of what we do when users don't specify enough batches or pipelines, but what do you think?

freddyaboulton (changes requested):
@angela97lin I think this is great! The one thing I think we should address before merge is that raising a `ValueError` in `_make_stacked_ensemble_pipeline` would stop `AutoMLSearch` early.

evalml/pipelines/utils.py
chukarsten: I guess the deciding point is: either AutoML smartly handle the user's desired data splitting strategy and then does an AutoML search (that's expensive in time?) or stops early the moment it knows the data split was unsuccessful.  Let me know if I've characterized that correctly. 

If we raise a value error in init, is there necessarily anything a user can do about the number of splits?  At that point would they have to increase the size of their dataset?  I kind of feel that if I'm personally going to use an AutoML search algorithm, I'd like to know before I kick it off that I won't benefit from a StackedEnsemble being in the running rather than get to the end to find that my dataset wasn't large enough.

chukarsten (approved):
I'm just participating in the conversation.  I agree that Freddy's comment should be addressed before moving on.  But a solid PR.

angela97lin:
Thank you guys for the comments! This is a really good point raised and I want to dig into it a bit more. 

The problem is that the stacked ensemble doesn't play well with data splits that result in only 1 split. This PR updates automl to pass the automl data splitter to the stacked ensemble, which might cause the ensemble to get a data splitter with one split (TrainingValidationSplit), and error out.

Is it a good idea to either disallow ensembling because of this behavior? Basically, for the default case we're saying that a user with large data is unable to use stacking. That seems unwise / odd. What do you guys think about *not* passing the automl data splitter if it causes an issue and just using the stacking ensemble's default cv method? 

- Pro: we allow users to use ensembling w/ large data.
- Con now we potentially have this mismatch of behavior, as we currently have in main, where the stacking ensemble doesn't use the same CV. I'd argue this might be okay anyways but we might not have the same "apples-to-apples" comparison 🤷 

Thoughts? @dsherry @freddyaboulton @chukarsten

angela97lin:
Closing after further discussion with @chukarsten @freddyaboulton @dsherry; any future work done for the data splitting and stacking in AutoML interaction will be tracked by #1746 :)

PR #1733, opened 2021-01-25
Add support for scikit-learn v0.24.0

### LINKED ISSUE #1593: Support sklearn 0.24.0

angela97lin:
Closes #1593.

There are two items that changed for scikit-learn 0.24.0:
1. scikit-learn now raises a warning in permutation_importance: https://github.com/scikit-learn/scikit-learn/blob/7c2f928cbc8d5261e5b85b5e63d549a794116c3b/sklearn/inspection/_partial_dependence.py#L504
2. scikit-learn throws `Setting a random_state has no effect since shuffle is False. You should leave random_state to its default (None), or set shuffle=True.` when random state is set and shuffle is False.

I've written about my proposed solutions below for both of these issues, but always welcome more discussion 😁 
--- 

Item 1: scikit-learn now raises a warning in permutation_importance: https://github.com/scikit-learn/scikit-learn/blob/7c2f928cbc8d5261e5b85b5e63d549a794116c3b/sklearn/inspection/_partial_dependence.py#L504

In order to support this version, we could:
- Update our `test_jupyter_graph_check` test to check that the number of warnings returned == 1, not 0 as before.
- Update our partial dependency impl to set `kind=legacy`. Then, no warnings will be raised. However, we would have to set our dependencies for scikit-learn to be `scikit-learn>=0.24.0` since `kind` is not a valid parameter in previous versions.- Update our partial dependency impl to set `kind=average`. Update our partial dependency impl to handle the new output (Bunch, not ndarrays). Then, no warnings will be raised. However, we would have to set our dependencies for scikit-learn to be `scikit-learn>=0.24.0` since `kind` is not a valid parameter in previous versions.

For this PR, I'm choosing to just stick with option #1. This is a change that will be enforced in scikit-learn 1.1, and we will have to revisit then, but for now, option #1 makes the most sense and allows us to support older versions. 

---

Item 2: scikit-learn throws `Setting a random_state has no effect since shuffle is False. You should leave random_state to its default (None), or set shuffle=True.` when random state is set and shuffle is False.

In order to support this version, we could:
- Update our codebase to check if random_state is None and shuffle=False and throw. This is nice in that the exception is still coming from evalml
- Leave as is, and add test to check that this error is thrown as expected for scikit-learn data splitters..? I guess since we're requiring data splitters to be a child class of the scikit-learn BaseCrossValidator, this might be a valid thing to do, but it could just be better to make it clear that this is a scikit-learn limitation.

For this PR, I chose to stick with option #2. Reasoning is, this is specific for some scikit-learn splitters, but works for our TimeSeriesSplit (even though its base class is the scikit-learn BaseCrossValidator). Still, its not something that we need to disallow entirely then?


evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -75,30 +75,28 @@ def test_make_data_splitter_default(problem_type, large_data):
         assert data_splitter.max_delay == 7
 
 
-def test_make_data_splitter_parameters():
+@pytest.mark.parametrize("problem_type, expected_data_splitter", [(ProblemTypes.REGRESSION, KFold),
```
angela97lin, line 78: Cleanup by using `parametrize` 😁

evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -115,6 +113,16 @@ def test_make_data_splitter_error():
...
+        make_data_splitter(X, y, 'XYZ')
+
+
+def test_make_data_splitter_error_shuffle_random_state():
+    n = 10
+    X = pd.DataFrame({'col_0': list(range(n)),
+                      'target': list(range(n))})
+    y = X.pop('target')
+
+    with pytest.raises(ValueError, match="Setting a random_state has no effect since shuffle is False."):
```
freddyaboulton, line 127: Does our `TrainingValidationSplit` also raise this exception?

freddyaboulton (approved):
@angela97lin This looks great! I agree with the plan for the two problems you mentioned. Just making sure, but the only unit tests that would fail if someone were to run `0.23.2` are `test_make_data_splitter_error_shuffle_random_state` and `test_jupyter_graph_check`, correct?

evalml/tests/automl_tests/test_automl_utils.py
angela97lin: Nope, I don't think so!

evalml/automl/utils.py
```diff
@@ -53,15 +53,17 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
         if not problem_configuration:
             raise ValueError("problem_configuration is required for time series problem types")
-        data_splitter = TimeSeriesSplit(n_splits=n_splits, gap=problem_configuration.get('gap'),
-                                        max_delay=problem_configuration.get('max_delay'))
-    if X.shape[0] > _LARGE_DATA_ROW_THRESHOLD:
-        data_splitter = TrainingValidationSplit(test_size=_LARGE_DATA_PERCENT_VALIDATION, shuffle=True)
+        return TimeSeriesSplit(n_splits=n_splits, gap=problem_configuration.get('gap'),
+                               max_delay=problem_configuration.get('max_delay'))
+
     return data_splitter
```
bchen1116, line 69: if we're not using `data_splitter`, can't we remove line 55 and just return `None` here?

bchen1116 (approved):
LGTM! Just left one comment, but I think going with the options you chose are good!

docs/source/release_notes.rst
```diff
@@ -17,14 +17,15 @@ Release Notes
         * Added time series support for ``make_pipeline`` :pr:`1566`
         * Added target name for output of pipeline ``predict`` method :pr:`1578`
         * Added multiclass check to ``InvalidTargetDataCheck`` for two examples per class :pr:`1596`
-        * Support graphviz 0.16 :pr:`1657`
+        * Added support for ``graphviz`` ``v0.16`` :pr:`1657`
```
chukarsten, line 20: Good for you.

chukarsten (approved):
This looks good.  I'm not sure I get the changes to EvalML/automl/utils.py make_data_split().  But I think it's just stylistic, so no big deal.

angela97lin:
@chukarsten Partially stylistic! The reason why I shifted and returned rather than what was previously there was because in the previous code, we would set the data_splitter based on the problem type, and then at the very end if the dataset was too large, we'd set it to `TrainingValidationSplit` regardless. However, scikit-learn 0.24.0 breaks if `shuffle=False` and `random_state!=None` for the binary/multiclass/regression case. So if the user has a super large binary dataset, we *want* to give them `TrainingValidationSplit`. Say they want to specify a random state and shuffle=False. Without the refactor, we would run into errors when it tries the initial set to `StratifiedKFold/KFold` before we set it to `TrainingValidationSplit`. Now, we just set to `TrainingValidationSplit` and don't run into that error :P

docs/source/release_notes.rst
```diff
@@ -5,20 +5,28 @@ Release Notes
     * Enhancements
         * Added ``graph_t_sne`` as a visualization tool for high dimensional data :pr:`1731`
         * Added the ability to see the linear coefficients of features in linear models terms :pr:`1738`
+        * Added support for ``scikit-learn`` ``v0.24.0`` :pr:`1733`
     * Fixes
     * Changes
+        * Updated components and pipelines to return ``Woodwork`` data structures :pr:`1668`
```
angela97lin, line 11: Oops, sneaking this in here. This should have been moved :P

PR #1738, opened 2021-01-26
Displaying Linear Model Coefficients

### LINKED ISSUE #1735: Understanding linear model coefficients
We'd like to be able to return a DataFrame detailing the coefficients of a linear model to identify features that have the greatest predictive power over the target variable.

ParthivNaresh:
Fixes #1735


evalml/model_understanding/graphs.py
```diff
@@ -890,3 +890,25 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
                         yaxis={'title': 'Target Values and Predictions'})
 
     return _go.Figure(data=data, layout=layout)
+
+
+def get_linear_coefficients(estimator, features=None):
```
jeremyliweishih, line 895: Does this show up in the API reference? If not we should add it in.

jeremyliweishih (approved):
LGTM. Only one comment about API reference

evalml/model_understanding/graphs.py
ParthivNaresh: Added it!

freddyaboulton (changes requested):
@ParthivNaresh This is great and the tests are solid! I think we should include the estimate for the intercept if the model was fit with one. I think you can just look at the `intercept_` attribute of the `_component_obj` of the estimator.

freddyaboulton (approved):
Looks great @ParthivNaresh !

evalml/model_understanding/graphs.py
```diff
@@ -890,3 +890,26 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, dates):
...
+    """Returns a dataframe showing the features with the greatest predictive power for a linear model.
+
+    Arguments:
+        estimator (Estimator): Fitted linear model family estimator.
+        features (list[str]): List of feature names associated with the underlying data.
+
+    Returns:
+        pd.DataFrame displaying the features by importance.
```
angela97lin, line 903: pd.DataFrame: DataFrame displaying the features by importance.

or something like that :)

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -1418,3 +1422,48 @@ def test_visualize_decision_trees(fitted_tree_estimators, tmpdir):
...
+
+def test_linear_coefficients_errors():
+    X = pd.DataFrame([[1, 2, 3, 5],
+                      [3, 5, 2, 1],
+                      [5, 2, 2, 2],
+                      [3, 2, 3, 3]])
+    y = pd.Series([2, 1, 3, 4])
+
+    dt = DecisionTreeRegressor()
+    dt.fit(X, y)
```
angela97lin, line 1435: Really nit-picky but to make it more obvious, can we mock the DT's fit method to show that we don't need it (and potentially save some time lolol)

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -1418,3 +1422,48 @@ def test_visualize_decision_trees(fitted_tree_estimators, tmpdir):
...
+    with pytest.raises(ValueError, match="Linear coefficients are only available for linear family models"):
+        get_linear_coefficients(dt)
+
+    lin = LinearRegressor()
+
+    with pytest.raises(ValueError, match="This linear estimator is not fitted yet."):
+        get_linear_coefficients(lin)
+
+
+def test_linear_coefficients_output():
```
angela97lin, line 1446: If I'm understanding correctly, this test uses two different linear models and checks that the method works. I think we could parametrize this test to use go through all of the linear model estimators and save duplication of code (no need to write the same asserts multiple times for different estimators)

angela97lin (approved):
Looking good! I added some comments for cleanup but impl LGTM :)

PR #1748, opened 2021-01-27
Non Numeric Bug with Partial Dependence

### LINKED ISSUE #1509: Partial dependence fails on non-numeric columns from featuretool's retail dataset
Just discussed with @gsheni and Colin, who were running into issues using the `partial_dependence` method on the featuretools retail dataset. It looked like it was being passed to scikit-learn, which was unable to do calculations on the object-type data.

Next steps: repro locally.

Resources:
https://github.com/scikit-learn/scikit-learn/issues/14969

Also useful: https://christophm.github.io/interpretable-ml-book/pdp.html

chukarsten:
fixes #1509

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -471,7 +471,17 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         ValueError: if the provided pipeline isn't fitted.
         ValueError: if the provided pipeline is a Baseline pipeline.
     """
+
+
+
     X = _convert_to_woodwork_structure(X)
+    try:
+        max_num_cats = max(X.select("categorical").describe().loc["nunique"])
+        if max_num_cats > grid_resolution:
+            grid_resolution = max_num_cats + 1
+    except ValueError as e:
+        if "max() arg is an empty sequence" in str(e):
+            pass
     X = _convert_woodwork_types_wrapper(X.to_dataframe())
     # if X[feature].dtype.name == "category":
     #     raise ValueError("Partial dependence is not supported for categorical features.")
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -472,8 +472,6 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         ValueError: if the provided pipeline is a Baseline pipeline.
     """
 
-
-
     X = _convert_to_woodwork_structure(X)
     try:
         max_num_cats = max(X.select("categorical").describe().loc["nunique"])
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -477,7 +477,7 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
     X = _convert_to_woodwork_structure(X)
     # Dynamically set the grid resolution to the maximum number of categories
     # in the categorical variables if there are more categories than resolution cells
-    X_cats =  X.select("categorical")
+    X_cats = X.select("categorical")
     if X_cats.shape[1] != 0:
         max_num_cats = max(X_cats.describe().loc["nunique"])
         grid_resolution = max([max_num_cats + 1, grid_resolution])
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -450,7 +450,7 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
             If features is a string, it must be a valid column name in X.
             If features is a tuple of int/strings, it must contain valid column integers/names in X.
         grid_resolution (int): Number of samples of feature(s) for partial dependence plot.  If this value
-            does not exceed the maximum number of categories present in categorical data within X, it will be
+            is less than the maximum number of categories present in categorical data within X, it will be
             set to the max number of categories + 1.
 
     Returns:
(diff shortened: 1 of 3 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -449,7 +449,9 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
             If features is an int, it must be the index of the feature to use.
             If features is a string, it must be a valid column name in X.
             If features is a tuple of int/strings, it must contain valid column integers/names in X.
-        grid_resolution (int): Number of samples of feature(s) for partial dependence plot
+        grid_resolution (int): Number of samples of feature(s) for partial dependence plot.  If this value
+            does not exceed the maximum number of categories present in categorical data within X, it will be
```
ParthivNaresh, line 453: If this number does not exceed or if this number does exceed?

ParthivNaresh (approved):
Everything looks good, just wanted to clarify the docstring wording

evalml/model_understanding/graphs.py
ParthivNaresh: I think the wording could be clarified a bit. If grid_resolution is less than the maximum number of..etc`

evalml/model_understanding/graphs.py
```diff
@@ -471,7 +473,17 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         ValueError: if the provided pipeline isn't fitted.
         ValueError: if the provided pipeline is a Baseline pipeline.
     """
+
     X = _convert_to_woodwork_structure(X)
+    # Dynamically set the grid resolution to the maximum number of categories
+    # in the categorical variables if there are more categories than resolution cells
+    try:
+        max_num_cats = max(X.select("categorical").describe().loc["nunique"])
+        if max_num_cats > grid_resolution:
```
angela97lin, line 482: Curious, what happens when max_num_cats == grid_resolution?

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -922,6 +925,30 @@ def test_partial_dependence_errors(logistic_regression_binary_pipeline_class):
         partial_dependence(pipeline, X, features=(0, 'b'))
 
 
+def test_partial_dependence_more_categories_than_grid_resolution(logistic_regression_binary_pipeline_class):
```
angela97lin, line 928: Just a suggestion but maybe we could somehow make it more clear that the number of categories is greater than the grid resolution, whether that's by setting grid_resolution specifically to a number and then asserting that there are `currency` has more than that number of categories, which is why we trigger this new code? Or alternatively we could have a tiny dataset like the previous test and set grid_resolution to something even tinier?

angela97lin (changes requested):
Looking good 😁 

I think we can clean up the impl and test a bit but thank you for fixing this!!

evalml/model_understanding/graphs.py
chukarsten: That's a good call, I should add in explicit test cases for this.  Let me fiddle with the test I have and add tests for max_num_cats < grid_resolution & max_num_cats == grid_resolution.

evalml/model_understanding/graphs.py
```diff
@@ -471,7 +473,17 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         ValueError: if the provided pipeline isn't fitted.
         ValueError: if the provided pipeline is a Baseline pipeline.
     """
+
     X = _convert_to_woodwork_structure(X)
+    # Dynamically set the grid resolution to the maximum number of categories
+    # in the categorical variables if there are more categories than resolution cells
+    try:
+        max_num_cats = max(X.select("categorical").describe().loc["nunique"])
```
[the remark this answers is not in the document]
chukarsten: I struggled with not setting
`max_num_cats = np.inf`

evalml/model_understanding/graphs.py
```diff
@@ -471,7 +473,17 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
...
+        if max_num_cats > grid_resolution:
+            grid_resolution = max_num_cats + 1
+    except ValueError as e:
```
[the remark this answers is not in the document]
chukarsten: True.  I always struggle with the "pythonic" way of doing these things.  I see a lot of pythonistas using try except in these cases as how to handle this kind of control flow so I figured I'd try it.  But I can certainly accept this change if it's clearer.

evalml/model_understanding/graphs.py
chukarsten: So just to resolve this - what actually happens is the same issue as initially reported.  Only the case where grid resolution > max cats lets us get through and generate partial_dependence.

freddyaboulton (approved):
@chukarsten  Looks good to me! Thanks for the fix 😄

evalml/model_understanding/graphs.py
```diff
@@ -471,7 +473,14 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
         ValueError: if the provided pipeline isn't fitted.
         ValueError: if the provided pipeline is a Baseline pipeline.
     """
+
     X = _convert_to_woodwork_structure(X)
+    # Dynamically set the grid resolution to the maximum number of categories
+    # in the categorical variables if there are more categories than resolution cells
+    X_cats =  X.select("categorical")
+    if X_cats.shape[1] != 0:
```
angela97lin, line 481: 👏

angela97lin (approved):
Looks great! Thank you for updating the impl and tests 😁

PR #1751, opened 2021-01-27
Drop support for py3.6

### LINKED ISSUE #1712: Drop support for python 3.6
Per https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table, the drop support for 3.6 should have happened on Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016).

However, looking at the docs for numpy: https://numpy.org/doc/stable/release/1.19.0-notes.html
```
The Python versions supported for this release are 3.6-3.8. Downstream developers should use Cython >= 0.29.16 for Python 3.8 support and OpenBLAS >= 3.7 to avoid problems on the Skylake architecture.
```

It's important to also note that python 3.9 is not listed. We need to check that numpy supports python 3.9

EDIT: Looks like numpy 1.19.3 supports 3.9, so we would have to update our min version to this: https://github.com/numpy/numpy/releases/tag/v1.19.3

jeremyliweishih:
Fixes #1712.


freddyaboulton (approved):
@jeremyliweishih Looks good to me!

docs/source/install.ipynb
```diff
@@ -16,7 +16,7 @@
     "pip install evalml\n",
     "```\n",
     "\n",
-    "For older Python versions (3.6.0 in particular), you must install `pip >= 18.0` and `setuptools > 40.0.0` in order to install evalml.\n",
+    "For older Python versions you must install `pip >= 18.0` and `setuptools > 40.0.0` in order to install evalml.\n",
```
angela97lin, line 19: Do we still need this line? :o Wondering since it seems like from the deletion that we wrote this for 3.6 in particular.

angela97lin (approved):
Looks good! Added a comment about doc clarification, we should test whether we still need that or not?

chukarsten (approved):

bchen1116 (approved):
🚀

dsherry (approved):
🚢 !

Issue #1347, opened 2020-10-26
Prediction explanations should aggregate contributions across all levels of categorical features


freddyaboulton:
If a user calls any of the explain predictions functions on a pipeline that has a OHE, the resulting table will have a row for each level of a categorical variable. We should aggregate the contributions across all levels of a categorical feature into a single row in the table so that the output is more intuitive.

There are two ways to do this:

1. Store a mapping from the encoded feature names to the original feature name. Then, I guess we would sum the contributions from all of the encoded features and call that the original features' contribution.
2. Pass the entire pipeline into the shap algorithm instead of just the pipeline estimator like we do for permutation_importance/partial dependence. Unfortunately, the shap library checks against a [hardcoded list of estimators ](https://github.com/slundberg/shap/blob/master/shap/explainers/_tree.py#L609)so I don't think it's viable.

I think we're left with 1 - looks like the library author says [summing is reasonable](https://github.com/slundberg/shap/issues/397) for encoded columns so it's not too bad of a plan.

freddyaboulton:
@chukarsten I think this is doable now that #1416 got done

freddyaboulton:
@chukarsten @dsherry This issue as written only mentions the OHE. I think the plan we have in place for that is pretty clear and reasonable. 

But we can also apply the same plan for other components, e.g. the `TextFeaturizer` and `DateTimeFeaturizer`. Is aggregating the shap values for the features created by those components in scope for this issue?

I think aggregating the shap values for features created by the `TextFeaturizer` makes sense because the features are not easily interpretable by humans and the names given to these features are a bit confusing. 

I'm on the fence about the `DateTimeFeaturizer` . I think it's pretty clear what the features created by the DateTimeFeaturizer are, e.g. `date_day_of_week`, `date_month`, and users may want to know how the day-of-week impacts a prediction vs the month. So I'm in favor of not aggregating those features but curious what you guys think.

chukarsten:
@freddyaboulton I think your instincts serve you well and these additional components should be done.  I would certainly err on the side of keeping the PRs smaller and limiting the scope of this one to OHE makes sense and filing discrete issues  for TF and DTF make sense.  Just to keep reviews manageable and to better propagate lessons learned from completion of OHE PR to the others.

freddyaboulton:
Discussed the requirements with @rpeck @dsherry @chukarsten @angela97lin and @bchen1116 :

[...]

PR #1762, opened 2021-01-29
Speed up permutation importance

### LINKED ISSUE #1416: Permutation importance: speed up by precomputing estimator features
**Background**
Our current permutation importance method shuffles the data and then evaluates the pipeline, computing features for the estimator and then training and predicting with the estimator.

Where this breaks down is when a preprocessing step generates multiple columns, like one-hot encoding. The one-hot encoder will create many boolean features from a single categorical input feature. If we have to do that creation for every shuffle of the input data, that's expensive.

**Proposal**
The optimization is to first precompute the preprocessed data to be passed to the estimator, then apply the shuffling to that data and train and evaluate the estimator. In order to do this, we need to know which generated features are "connected" in that they were created from the same input feature, so that the values for those generated features can be shuffled in sync with each other, i.e. they always appear in the same row with each other.

If we can write an algo to do this sort of shuffle, we can avoid many calls to the categorical encoders like OHE, and potentially get a huge permutation importance speedup on datasets with many categorical features.

**Future implications**
Once we support pipelines which aren't linear sequences of transformers followed by one estimator, this optimization may need to become more sophisticated, but that's a problem for the future. (#273 )

freddyaboulton:
### Pull Request Description
Fixes #1416 

Currently this only supports pipelines where each feature is created from at most one other feature. 

Handling pipelines where that is not the case (PCA, multiple estimators) will be more complicated. We'd have to cache the features before each of these transformers/estimators and do the permutation there. 

### 10x speedup if pipeline has text features
On 500 rows of the fraud dataset:

image

### About 2x speedup on a "standard" pipeline
image

-----


evalml/pipelines/components/transformers/transformer.py
```diff
@@ -68,3 +68,6 @@ def fit_transform(self, X, y=None):
             except MethodPropertyNotFoundError as e:
                 raise e
         return _convert_to_woodwork_structure(X_t)
+
+    def _get_feature_provenance(self):
```
freddyaboulton, line 72: Keeping this private for now since it's only used for permutation importance but this mechanism can help with https://github.com/alteryx/evalml/issues/1347

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -523,70 +523,6 @@ def test_graph_confusion_matrix_title_addition(X_y_binary):
     assert fig_dict['layout']['title']['text'] == 'Confusion matrix with added title text, normalized using method "true"'
 
 
-def test_get_permutation_importance_invalid_objective(X_y_regression, linear_regression_pipeline_class):
```
freddyaboulton, line 526: There has been talk about splitting up `test_graphs.py`. I figured I would get a head start by creating `test_permutation_importance.py` and moving all of the permutation tests there.

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -0,0 +1,274 @@
...
+                                                       'Target Encoder': {'cols': ['region', 'country']}}),
+              (LinearPipelineCreateFeatureThenDropIt, {'Select Columns Transformer': {'columns': ['amount']},
+                                                       'DoubleColumns': {'drop_old_columns': False},
+                                                       'Drop Columns Transformer': {'columns': ['amount_doubled']}})
+              ]
+
+
+@pytest.mark.parametrize('pipeline_class, parameters', test_cases)
+@patch('evalml.pipelines.PipelineBase._supports_fast_permutation_importance', new_callable=PropertyMock)
+def test_fast_permutation_importance_matches_sklearn_output(mock_supports_fast_importance, pipeline_class, parameters,
```
freddyaboulton, line 147: I think it's important to test that our optimization calculates the values correctly. Mocking doesn't really make sense here but since I'm only using 100 points and `n_jobs=1` I think it's ok.

evalml/model_understanding/graphs.py
```diff
@@ -308,13 +386,14 @@ def calculate_permutation_importance(pipeline, X, y, objective, n_repeats=5, n_j
     if not objective.is_defined_for_problem_type(pipeline.problem_type):
         raise ValueError(f"Given objective '{objective.name}' cannot be used with '{pipeline.name}'")
 
-    def scorer(pipeline, X, y):
-        scores = pipeline.score(X, y, objectives=[objective])
-        return scores[objective.name] if objective.greater_is_better else -scores[objective.name]
-    perm_importance = sk_permutation_importance(pipeline, X, y, n_repeats=n_repeats, scoring=scorer, n_jobs=n_jobs, random_state=random_state)
+    if pipeline._supports_fast_permutation_importance:
```
bchen1116, line 389: nice!

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -0,0 +1,274 @@
...
+    pipeline = pipeline_class(parameters=parameters)
+    pipeline.fit(X, y)
+    fast_scores = calculate_permutation_importance(pipeline, X, y, objective='Log Loss Binary',
+                                                   random_state=random_seed)
+
+    mock_supports_fast_importance.return_value = False
+    slow_scores = calculate_permutation_importance(pipeline, X, y, objective='Log Loss Binary',
+                                                   random_state=0)
+
+    pd.testing.assert_frame_equal(fast_scores, slow_scores)
```
bchen1116, line 169: Test looks good! Since we're ideally getting a speed boost by doing faster performance, would we be able to do assertions that the time for `fast_scores` is <= time for `slow_scores`?

bchen1116 (approved):
Looks good! The time speedup is pretty awesome!

evalml/pipelines/component_graph.py
```diff
@@ -92,7 +93,12 @@ def fit(self, X, y):
             X (ww.DataTable, pd.DataFrame): The input training data of shape [n_samples, n_features]
             y (ww.DataColumn, pd.Series): The target training data of length [n_samples]
         """
+        if isinstance(X, ww.DataTable):
```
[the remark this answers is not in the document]
freddyaboulton: Good point! I'll delete this in favor of that.

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Good point! I tried a few times locally and the time for fast scores is smaller than slow scores. I'm pushing this up and keeping an eye if the results vary from run to run.

freddyaboulton: It flaked on windows so I'm reverting it lol.

chukarsten (approved):
Like it.  Lots of good work here.  A few nits/suggestions/questions but nothing to stop a merge.

evalml/pipelines/component_graph.py
```diff
@@ -231,6 +235,56 @@ def _compute_features(self, component_list, X, y=None, fit=False):
...
+
+        Arguments:
+            component_list (list(str)): Names of components in topological order.
+            input_feature_names (list(str)): Names of the features in the input dataframe.
+
+        Returns:
+           dictionary: mapping of feature name of set feature names that were created from that feature.
+        """
+        if not component_list:
+            return {}
```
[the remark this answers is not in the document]
freddyaboulton: Pipelines with empty component graphs are valid so I wanted to exit before we get the `final_estimator_features`.

evalml/model_understanding/graphs.py
```diff
@@ -283,6 +284,83 @@ def graph_roc_curve(y_true, y_pred_proba, custom_class_names=None, title_additio
...
+def _fast_permutation_importance(pipeline, X, y, objective, n_repeats=5, n_jobs=None, random_seed=None):
+    """Calculate permutation importance faster by only computing the estimator features once.
+
+    Only used for pipelines that support this optimization.
+
+    Arguments:
+        pipeline (PipelineBase or subclass): Fitted pipeline
+        X (ww.DataTable, pd.DataFrame): The input data used to score and compute permutation importance
+        y (ww.DataColumn, pd.Series): The target data
+        objective (str, ObjectiveBase): Objective to score on
```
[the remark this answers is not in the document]
freddyaboulton: Good point. I think the docstring for `calculate_permutation_importance` is enough. I'll add a type hint for the return type there!

dsherry:
Will take a look shortly. I love the performance numbers! 10x speedup for text is 🔥 
@freddyaboulton is it possible to add data points to that graph? Helpful to see how many samples were taken heh.

> Currently this only supports pipelines where each feature is created from at most one other feature.

That seems like a reasonable limitation for the immediate future. We'll need to revisit this at some point, yes? Transformers which create columns today: OHE, target encoder, datetime, text, timeseries delayed features, and soon seasonality. We currently have LDA/PCA transformers defined but aren't using them in automl--adding those would definitely violate this limitation.

freddyaboulton:
@dsherry Agreed on enhancing this implementation to support for more transformers. When we get there, we also need to consider pipelines with multiple estimators. The estimators act like transformers in that case - they ingest many columns and output many columns.

angela97lin:
@freddyaboulton Taking a look through this right now, but also no need to wait for me hehe

evalml/pipelines/component_graph.py
```diff
@@ -231,6 +235,56 @@ def _compute_features(self, component_list, X, y=None, fit=False):
                 output_cache[component_name] = output
         return output_cache
 
+    def _get_feature_provenance(self, component_list, input_feature_names):
```
angela97lin, line 238: Curious, why `component_list` as an argument? I noticed the only time we use this is in calling 
`self._feature_provenance = self._get_feature_provenance(self.compute_order, X.columns)`; would it make sense to just remove the arguments and internally use the component graph's compute_order? Or is this for flexibility? Are there use cases for this? (Maybe computing not the whole graph but some path in it? 🤔 )

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -0,0 +1,272 @@
...
+        'OHE_2': ['One Hot Encoder', 'SelectCategorical2'],
+        'DT': ['DateTime Featurization Component', "SelectNumeric"],
+        'OHE_3': ['One Hot Encoder', 'DT'],
+        'Estimator': ['Random Forest Classifier', 'OHE_3', 'OHE_1', 'OHE_2'],
+    }
+
+
+test_cases = [(LinearPipelineWithDropCols, {"Drop Columns Transformer": {'columns': ['country']}}),
+              (LinearPipelineWithImputer, {}),
+              (LinearPipelineSameFeatureUsedByTwoComponents, {'DateTime Featurization Component': {'encode_as_categories': True}}),
```
angela97lin, line 117: Amazing. I was wondering what happens when the datetime featurizer creates categorical columns that the OHE then uses, since it'd rely on the output cols of the datetime featurizer as provenance. 🤩

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -0,0 +1,272 @@
...
+def test_fast_permutation_importance_matches_sklearn_output(mock_supports_fast_importance, pipeline_class, parameters,
+                                                            has_minimal_dependencies):
+    if has_minimal_dependencies and pipeline_class == LinearPipelineWithTargetEncoderAndOHE:
+        pytest.skip("Skipping test_fast_permutation_importance_matches_sklearn_output for target encoder cause "
+                    "dependency not installed.")
+    X, y = load_fraud(100)
+
+    # Do this to make sure we use the same int as sklearn under the hood
+    random_state = np.random.RandomState(0)
+    random_seed = random_state.randint(np.iinfo(np.int32).max + 1)
```
angela97lin, line 156: Is this different behavior from what we use in `get_random_seed` / can we just use that? 👀

evalml/tests/model_understanding_tests/test_permutation_importance.py
angela97lin: classic windows lol

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -0,0 +1,272 @@
...
+        'SelectCategorical2': ["Select Columns Transformer", "Imputer"],
+        'OHE_1': ['One Hot Encoder', 'SelectCategorical1'],
+        'OHE_2': ['One Hot Encoder', 'SelectCategorical2'],
+        'DT': ['DateTime Featurization Component', "SelectNumeric"],
+        'OHE_3': ['One Hot Encoder', 'DT'],
+        'Estimator': ['Random Forest Classifier', 'OHE_3', 'OHE_1', 'OHE_2'],
+    }
+
+
+test_cases = [(LinearPipelineWithDropCols, {"Drop Columns Transformer": {'columns': ['country']}}),
```
angela97lin, line 115: Wow, this is parametrization taken to the next level LMAO

angela97lin (approved):
This is some really great work!! The speedups are *chefs kiss*. Implementation looks great, just left a few minor questions and comments (some just so I understand this better hehe).

Also, `_get_feature_provenance` is really interesting to see from a conceptual level, to better understand how our components transform data. (Also useful since I've been working on that describes transformations :P)

evalml/pipelines/component_graph.py
```diff
@@ -231,6 +235,56 @@ def _compute_features(self, component_list, X, y=None, fit=False):
...
+                    provenance[component_input] = provenance[component_input].union(set(component_output))
+
+                # Case 2: The transformer created features from a feature created from an original feature.
+                # Add it to the provenance of the original feature it was created from
+                else:
+                    for in_feature, out_feature in provenance.items():
+                        if component_input in out_feature:
+                            provenance[in_feature] = out_feature.union(set(component_output))
+
+        # Get rid of features that are not in the dataset the final estimator uses
```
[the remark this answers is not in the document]
freddyaboulton: Yes! The other case is when we create features from features but delete the intermediate feature, e.g we create a month column from a datetime feature and then we ohe the month column. Since the ohe drops the month column we want to remove it from the provenance of the datetime feature.

evalml/pipelines/component_graph.py
```diff
@@ -231,6 +235,56 @@ def _compute_features(self, component_list, X, y=None, fit=False):
...
+
+        If a feature is then calculated from feature 'a', e.g. 'a_squared', then the provenance would instead
+        be {'cats': ['a', 'a_squared', 'b']}.
+
+        Arguments:
+            component_list (list(str)): Names of components in topological order.
+            input_feature_names (list(str)): Names of the features in the input dataframe.
+
+        Returns:
+           dictionary: mapping of feature name of set feature names that were created from that feature.
```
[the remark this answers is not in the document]
freddyaboulton: Thank you! 😅

evalml/pipelines/components/transformers/preprocessing/datetime_featurizer.py
```diff
@@ -113,3 +113,11 @@ def get_feature_names(self):
            mapping the unique feature values to their integer encoding.
         """
         return self._categories
+
+    def _get_feature_provenance(self):
+        provenance = {}
+        for col_name in self._date_time_col_names:
+            provenance[col_name] = []
+            for feature in self.parameters['features_to_extract']:
+                provenance[col_name].append(f'{col_name}_{feature}')
```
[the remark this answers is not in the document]
freddyaboulton: Good observation! I like the "from scratch" pattern because it keeps the definition of the provenance separate from the transform logic but you're right it can lead to some duplication. I think in most cases I tried to follow the "from scratch" pattern but I couldn't use it for all because the OHE, for example, has special logic for making names unique so the provenance basically has to happen in `get_feature_names`.

evalml/pipelines/components/transformers/encoders/target_encoder.py
```diff
@@ -68,3 +68,6 @@ def get_feature_names(self):
             np.array: The feature names after encoding
         """
         return self._component_obj.get_feature_names()
+
+    def _get_feature_provenance(self):
```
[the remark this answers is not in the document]
freddyaboulton: There is no uniform logic yet! For the target encoder we get a `ComponentNotYetFittedError` but for the date time featurizer we get a stack trace lol.

I think that's ok for now since we're only using this internally. If we ever decide to make it public we should definitely modify the metaclass to add some guardrails.

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Unfortunately it is! Great suggestion though!

evalml/pipelines/component_graph.py
freddyaboulton: Great point! This is an oversight on my part. I think I was refactoring some code and thought I could make it a static method but that's not really reasonable. 

I'm deleting the component_list parameter now!

evalml/pipelines/component_graph.py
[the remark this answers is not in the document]
angela97lin: Oooo got it, thanks for explaining! sgtm 😀

PR #1770, opened 2021-01-31
Address MultiIndex problem with LightGBM

### LINKED ISSUE #1710: Multiindex Lightgbm problem
I'm not sure whether the problem arises on evalml side or Lightgbm but I have a problem with multiindex X passing to `AutoMLSearch.search()`

```
Batch 1: (4/9) LightGBM Regressor w/ Imputer            Elapsed:00:01
	Starting cross validation
			Fold 0: Encountered an error.
			Fold 0: All scores will be replaced with nan.
			Fold 0: Please check ...\evalml_debug.log for the current hyperparameters and stack trace.
			Fold 0: Exception during automl search: logical_types contains columns that are not present in dataframe: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
			Fold 1: Encountered an error...
```
Sorry, I can't give you a proper example. I still think this is better than nothing.

On my side, I fixed the problem with 
```
X_train.pipe(lambda df: df.set_axis(['_'.join(col).strip() for col in df.columns.values], axis=1))
```

In addition, being still a novice in using the library, but I can't find a proper way to see logs of errors. I see the same information in evalml_debug.log without a clear traceback of the error. This is why I don't know the exact reason for the problem.

angela97lin:
Closes #1710 

Added context on the issue here: https://github.com/alteryx/evalml/issues/1710#issuecomment-770035322. Not sure if this is the best way to solve this issue so would love any thoughts and suggestions :')

### DIFF
--- a/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
@@ -101,6 +101,7 @@ class LightGBMClassifier(Estimator):
     def fit(self, X, y=None):
         X_encoded = self._encode_categories(X, fit=True)
         y_encoded = self._encode_labels(y)
+        X_encoded.columns = list(map('_'.join, X_encoded.columns.values))
         return super().fit(X_encoded, y_encoded)
 
     def predict(self, X):
--- a/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
@@ -76,6 +76,8 @@ class LightGBMClassifier(Estimator):
         """Encodes each categorical feature using ordinal encoding."""
         X_encoded = _convert_to_woodwork_structure(X)
         X_encoded = _rename_column_names_to_numeric(X_encoded)
+        if pd.api.types.is_iterator(X_encoded.columns.values):
+            X_encoded.columns = list(map('_'.join, X_encoded.columns.values))
         cat_cols = list(X_encoded.select('category').columns)
         X_encoded = _convert_woodwork_types_wrapper(X_encoded.to_dataframe())
         if len(cat_cols) == 0:
@@ -101,7 +103,6 @@ class LightGBMClassifier(Estimator):
     def fit(self, X, y=None):
         X_encoded = self._encode_categories(X, fit=True)
         y_encoded = self._encode_labels(y)
-        X_encoded.columns = list(map('_'.join, X_encoded.columns.values))
         return super().fit(X_encoded, y_encoded)
 
     def predict(self, X):
--- a/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
@@ -75,12 +75,13 @@ class LightGBMClassifier(Estimator):
     def _encode_categories(self, X, fit=False):
         """Encodes each categorical feature using ordinal encoding."""
         X_encoded = _convert_to_woodwork_structure(X)
-        X_encoded = _rename_column_names_to_numeric(X_encoded)
         cat_cols = list(X_encoded.select('category').columns)
         X_encoded = _convert_woodwork_types_wrapper(X_encoded.to_dataframe())
-        new_names = list(map('_'.join, X_encoded.columns))
-        X_encoded.columns = new_names
-        cat_cols = [name_mappings[col] for col in cat_cols]
+
+        rename_cols_dict = dict((col, col_num) for col_num, col in enumerate(list(X_encoded.columns)))
+        X_encoded = _rename_column_names_to_numeric(X_encoded)
+        cat_cols = [rename_cols_dict[col] for col in cat_cols]
+
         if len(cat_cols) == 0:
             return X_encoded
         if fit:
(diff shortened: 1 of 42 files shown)


evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
```diff
@@ -70,10 +70,14 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=20, max
 
     def _encode_categories(self, X, fit=False):
         """Encodes each categorical feature using ordinal encoding."""
-        X_encoded = _convert_to_woodwork_structure(X)
-        X_encoded = _rename_column_names_to_numeric(X_encoded)
-        cat_cols = list(X_encoded.select('category').columns)
-        X_encoded = _convert_woodwork_types_wrapper(X_encoded.to_dataframe())
+        X = _convert_to_woodwork_structure(X)
```
angela97lin, line 73: Blargh. I'd like to combine lightgbm classifier and regressor, since they share a lot of the same core logic. Probs better to file than to address here tho 😬

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -74,10 +74,14 @@ def __init__(self, boosting_type="gbdt", learning_rate=0.1, n_estimators=100, ma
...
-        X_encoded = _rename_column_names_to_numeric(X_encoded)
-        cat_cols = list(X_encoded.select('category').columns)
-        X_encoded = _convert_woodwork_types_wrapper(X_encoded.to_dataframe())
+        X = _convert_to_woodwork_structure(X)
+        cat_cols = list(X.select('category').columns)
+        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+
+        X_encoded = _rename_column_names_to_numeric(X)
+        rename_cols_dict = dict(zip(X.columns, X_encoded.columns))
+        cat_cols = [rename_cols_dict[col] for col in cat_cols]
```
angela97lin, line 83: Since we moved this to after conversion to dataframe (need dataframe to set column names easily for tuples), need to find mapping from old to new names and update.

evalml/utils/gen_utils.py
```diff
@@ -208,17 +208,23 @@ def _rename_column_names_to_numeric(X):
...
+
     if isinstance(X, ww.DataTable):
         X_t = X.to_dataframe()
-        logical_types = X.logical_types
-    name_to_col_num = dict((col, col_num) for col_num, col in enumerate(list(X.columns)))
-    X_renamed = X_t.rename(columns=name_to_col_num, inplace=False)
+    else:
+        X_t = X.copy()
+
+    if len(X_t.columns) > 0 and isinstance(X_t.columns[0], tuple):
```
angela97lin, line 219: Handling tuples. Working with WW directly doesn't allow renames unless passed another tuple, and this helps convert tuples to strings to 'flatten' the tuple. Ex: ('a', 1) --> '('a', 1)' to convert tuple multi-index cols to string cols

freddyaboulton: I think the implementation here looks good! I would vote for `isinstance(X_t.columns, pd.MultiIndex)` to make it clearer what's we're guarding against. 

Looks like specifying tuples for column names automatically creates a MultiIndex:

image

freddyaboulton (approved):

I think the fix you have here makes sense. The other thing that comes to mind is converting X to numpy XGBoost and LightGBM `fit` rather than having to make all these conversions to the dataframe.

That would require changing more code but may be better long term if there isn't a reason that we need to pass a dataframe to xgboost or lightgbm.

evalml/utils/gen_utils.py
angela97lin: Yupperino, agreed that it'll make it more clear. Will update 😁

evalml/tests/component_tests/test_lgbm_regressor.py
```diff
@@ -185,3 +185,18 @@ def test_regression_goss(X_y_regression):
     clf.fit(X, y)
     assert clf.parameters['bagging_freq'] == 0
     assert clf.parameters['bagging_fraction'] == 0.9
+
+
+@pytest.mark.parametrize("data_type", ['pd', 'ww'])
+def test_lightgbm_multiindex(data_type, X_y_regression, make_data_type):
```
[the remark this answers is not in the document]
angela97lin: Ah yes! I thought that since xgboost can handle tuples that the tests wouldn't be necessary, but since it uses the _rename_column_names_to_numeric method it was impacted and I had to tweak some stuff. Thank you!!

angela97lin:
@freddyaboulton Ooo, that's a really interesting idea! I wonder what that would mean for the LightGBM estimators specifically. Right now, we take advantage of the `auto` feature to determine categorical columns, which requires a pandas DF (and us making sure that the categorical cols as determined by WW are set to `category` dtype). (https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier) If we converted to numpy, we'd have to manually determine the mapping from col name to integer index for the numpy array and pass that along (assuming also dtypes don't get funky). Could be an interesting alternative to explore though if we see more problems like this crop up / more edge cases hehe.

bchen1116 (approved):
LGTM! I agree with leaving it as a pandas input so that we can preserve categorical columns for lightGBM.

PR #1775, opened 2021-02-01
Class Imbalance Data splitter samplers

### LINKED ISSUE #973: Balanced sampling for classification problems

### LINKED ISSUE #1786: Implement SMOTENC as a data splitting strategy for handling categorical data
Extension of #973, but we want to be able to handle categorical data as well as numeric data.

bchen1116:
fix #1786

Creating a few class imbalance data splitters to determine which strategies will be beneficial for EvalML

Current issues/questions:
- SMOTE cannot handle categorical data, so our current implementations can only handle numeric data. SMOTE also treats all data as continuous for its synthetic generation ability, so this is something we should keep in mind. For categorical and numeric data, we can incorporate [SMOTENC](https://imbalanced-learn.org/stable/generated/imblearn.over_sampling.SMOTENC.html), which I am addressing in a separate PR now (new issue [here](https://github.com/alteryx/evalml/issues/new)).
- The default implementation for the samplers cannot handle small samples of super imbalanced data well (for instance, 200 rows with a [0.02, 0.98] class balance). In this scenario, especially for `KMeansSMOTE`, we would want the user to pass in the data_splitter object initialized with the proper thresholds. However, how should we tell users to do this? Is this a message we should try to catch/throw in AutoMLSearch (might be hard to catch), or can we just put it in the docs?
- These names might be complicated to users. Any suggestions on name simplifications or simplifications for how users can pass these splitters to EvalML? 
- Is it confusing to create and use the `sampler` arg? I figured it would be nice to avoid an import for the user, but should we stick with just having them pass in the class instance for consistency?

Docs for API reference [here](https://feature-labs-inc-evalml--1775.com.readthedocs.build/en/1775/api_reference.html#data-splitter-classes) and for usage [here](https://feature-labs-inc-evalml--1775.com.readthedocs.build/en/1775/user_guide/automl.html#Imbalanced-Data)
Clean Performance doc [here](https://alteryx.quip.com/VVTFAfbT1Nwm/Data-Sampler-Class-Imbalance-tests-Cleaned).


evalml/automl/automl_search.py
```diff
@@ -267,9 +268,9 @@ def __init__(self,
         # make everything ww objects
         self.X_train = _convert_to_woodwork_structure(X_train)
         self.y_train = _convert_to_woodwork_structure(y_train)
-
+        self.sampler = sampler if (sampler is not None and sampler != 'None') else None
```
bchen1116, line 271: 'None' is primarily for LG test usage. I updated a LG branch to take in the ['sampler' arg](https://github.com/alteryx/evalml_looking_glass/compare/bc_imbalance_sampling#diff-96b21daac4fea193abcbda0eed3e0a0f89bf31eef2dc2593dd8b1749369a8acfR24)

evalml/automl/automl_search.py
```diff
@@ -553,6 +554,8 @@ def _find_best_pipeline(self):
                     X_train, X_threshold_tuning, y_train, y_threshold_tuning = split_data(X_train, y_train, self.problem_type,
                                                                                           test_size=0.2,
                                                                                           random_state=self.random_state)
+                if getattr(self.data_splitter, "transform", None):
```
bchen1116, line 557: If we used a sampler splitter, we want to transform the data so that we can train the best pipeline, if needed

evalml/automl/automl_search.py
```diff
@@ -674,8 +677,12 @@ def _compute_cv_scores(self, pipeline):
                 logger.debug(f"Skipping fold {i} because CV for stacked ensembles is not supported.")
                 break
             logger.debug(f"\t\tTraining and scoring on fold {i}")
-            X_train, X_valid = self.X_train.iloc[train], self.X_train.iloc[valid]
-            y_train, y_valid = self.y_train.iloc[train], self.y_train.iloc[valid]
+            if not self.sampler:
+                X_train, X_valid = self.X_train.iloc[train], self.X_train.iloc[valid]
+                y_train, y_valid = self.y_train.iloc[train], self.y_train.iloc[valid]
+            else:
+                X_train, y_train = train[0], train[1]
```
bchen1116, line 684: The sampler splitter passes out the `[(X_train, y_train), (X_test, y_test)]` data as tuples, not indices. We can't pass out indices with `SMOTE`, which generates synthetic data

evalml/tests/utils_tests/test_dependencies.py
```diff
@@ -9,7 +9,7 @@ def test_has_minimal_deps(has_minimal_dependencies):
     lines = open(reqs_path, 'r').readlines()
     lines = [line for line in lines if '-r ' not in line]
     reqs = requirements.parse(''.join(lines))
-    extra_deps = [req.name for req in reqs]
+    extra_deps = [req.name if req.name != 'imbalanced-learn' else 'imblearn' for req in reqs]
```
bchen1116, line 12: The pip install is for `imbalanced-learn` but the import itself is `imblearn`, which is why I added this conditional here

evalml/preprocessing/data_splitters/smote_split.py
```diff
@@ -0,0 +1,124 @@
...
+
+            Returns:
+                tuple(list): A tuple containing the resulting X_train, X_valid, y_train, y_valid data.
+        """
+        X, y = self._to_woodwork(X, y)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=self.test_size, random_state=self.random_state)
+        X_train_resample, y_train_resample = self.kmsmote.fit_resample(X_train, y_train)
+        X_train_resample, y_train_resample = self._to_woodwork(X_train_resample, y_train_resample, to_pandas=False)
+        X_test, y_test = self._to_woodwork(X_test, y_test, to_pandas=False)
+        return iter([((X_train_resample, y_train_resample), (X_test, y_test))])
```
chukarsten: I'm seeing a bit of reused code here (correct me if I'm wrong).  Is it worth a base TVSplit class that has these methods defined so we can cut down on some of the duplicated code?

evalml/preprocessing/data_splitters/smote_split.py
```diff
@@ -0,0 +1,124 @@
...
+            Returns:
+                tuple(list): A tuple containing the resulting X_train, X_valid, y_train, y_valid data.
+        """
+        X, y = self._to_woodwork(X, y)
+        for i, (train_indices, test_indices) in enumerate(super().split(X, y)):
+            X_train, X_test, y_train, y_test = X.iloc[train_indices], X.iloc[test_indices], y.iloc[train_indices], y.iloc[test_indices]
+            X_train_resample, y_train_resample = self.kmsmote.fit_resample(X_train, y_train)
+            X_train_resample, y_train_resample = self._to_woodwork(X_train_resample, y_train_resample, to_pandas=False)
+            X_test, y_test = self._to_woodwork(X_test, y_test, to_pandas=False)
+            yield iter(((X_train_resample, y_train_resample), (X_test, y_test)))
```
chukarsten: I think same thing here, it might be worth having like a BaseCVSplit class for this functionality.

evalml/tests/preprocessing_tests/test_k_means_smote_split.py
```diff
@@ -0,0 +1,101 @@
...
+import pandas as pd
+import pytest
+from sklearn.model_selection import StratifiedKFold, train_test_split
+
+from evalml.preprocessing.data_splitters import (
+    KMeansSMOTECVSplit,
+    KMeansSMOTETVSplit
+)
+
+im = pytest.importorskip('imblearn.over_sampling', reason='Skipping plotting test because imblearn not installed')
```
chukarsten, line 11: lol, is it a plotting test that we're skipping?

evalml/tests/preprocessing_tests/test_k_means_smote_split.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_kmeans_smote_nsplits():
```
chukarsten, line 14: You might need to drop some knowledge on me: what's the purpose of this?  I don't get why a splitter is hardcoded to return n_splits (and I think this is a personal problem of mine).

chukarsten (changes requested):

chukarsten:
Sorry, I submitted my change request early asI think your push gave me an error while I was in review:  just wanted to suggest one more thing.  How do we feel about parametrizing your tests all into a test_splits.py or test_splitter.py?  They all seem to be doing the same thing but with different classes.

But all in all, great work.  This was a big one with lots of functionality.

evalml/tests/preprocessing_tests/test_k_means_smote_split.py
bchen1116: I was just following convention of [TrainingValidationSplit](https://github.com/alteryx/evalml/blob/d92a3cb37f65a47119ac6a29820b404afc09825a/evalml/preprocessing/data_splitters/training_validation_split.py#L31). I think it's just additional information that could be helpful for users that we've provided previously.

evalml/preprocessing/data_splitters/rus_split.py
```diff
@@ -0,0 +1,125 @@
...
+    def get_n_splits():
+        """Returns the number of splits of this object"""
+        return 1
+
+    def split(self, X, y=None):
+        """Divides the data into training and testing sets
+
+            Arguments:
+                X (ww.DataTable): DataTable of points to split
+                y (ww.DataTable): DataColumn of points to split
```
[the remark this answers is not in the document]
bchen1116: Mmm unsure but I removed the default since we do need X and y values in order to split the data properly

evalml/preprocessing/data_splitters/rus_split.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class RandomUnderSamplerCVSplit(BaseCVSplit):
+    """Split the training data into KFold cross validation sets and uses RandomUnderSampler to balance the training data.
+       Keeps the validation data the same. Works only on continuous, numeric data."""
+
+    def __init__(self, sampling_strategy='auto', replacement=False, n_splits=3, shuffle=True, random_state=0):
+        error_msg = "imbalanced-learn is not installed. Please install using 'pip install imbalanced-learn'"
+        im = import_or_raise("imblearn.under_sampling", error_msg=error_msg)
+        self.rus = im.RandomUnderSampler(sampling_strategy=sampling_strategy, replacement=False, random_state=random_state)
```
chukarsten, line 26: Sorry dude, hate to be a jerk face, but I think you need to pass replacement here, too.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2177,3 +2187,44 @@ def test_automl_pipeline_random_state(mock_fit, mock_score, random_state, X_y_mu
...
+
+@pytest.mark.parametrize("sampler",
+                         ["KMeansSMOTETVSplit", "KMeansSMOTECVSplit",
+                          "SMOTETomekCVSplit", "SMOTETomekTVSplit",
+                          "RandomUnderSamplerCVSplit", "RandomUnderSamplerTVSplit",
+                          "None", "Nope"])
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_sampler(mock_fit, mock_score, sampler, X_y_binary):
+    pytest.importorskip('imblearn', reason='Skipping plotting test because imblearn not installed')
```
chukarsten, line 2200: There's a few of these copy-pastas left, I think.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2177,3 +2187,44 @@ def test_automl_pipeline_random_state(mock_fit, mock_score, random_state, X_y_mu
...
+
+@pytest.mark.parametrize("sampler",
+                         [KMeansSMOTETVSplit, KMeansSMOTECVSplit,
+                          SMOTETomekCVSplit, SMOTETomekTVSplit,
+                          RandomUnderSamplerCVSplit, RandomUnderSamplerTVSplit,
+                          SMOTENCCVSplit, SMOTENCTVSplit])
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_sampler_instance(mock_fit, mock_score, sampler, X_y_binary):
+    pytest.importorskip('imblearn', reason='Skipping plotting test because imblearn not installed')
```
chukarsten, line 2222: This one, too.

chukarsten (changes requested):
Sorry, there were just a few more instances of some of the specific cases I had pointed out earlier - specifically the reference to skipping "plotting" tests as well as passing the replacement arg through.  Overall, though, you cut down on a LOT of duplicate lines.  Looking really good.  I'm hoping you and I can take what we've done here and keep a similar eye when looking at different pieces of the code baseline.  If you think there's places where we can cut down on complexity via inheritance, please please submit an issue.

chukarsten (approved):
You're awesome.

evalml/utils/gen_utils.py
```diff
@@ -355,6 +355,26 @@ def _convert_woodwork_types_wrapper(pd_data):
     return pd_data
 
 
+def _to_woodwork(X, y, to_pandas=True):
```
angela97lin, line 358: A bit confused about this method, especially since it's called `_to_woodwork` but can also return pandas? I'm guessing there are two things that are being done here: conversion to WW and back, and the check for all_numeric. But I wonder if these two should be mingled together in a util method for generic purposes like this when it seems pretty specific to the data splitter (allowing only numeric)? I remember wondering if it was worth creating a util method that converts to WW and then back but then decided not realllly since in some cases, we want to do some checks/grab some information (here, the numeric cols and whether all cols are numeric).

Long ramble of thoughts but I guess my suggestion here would be, could we at least rename this to something more descriptive :p

evalml/tests/utils_tests/test_dependencies.py
angela97lin: Why is this not an issue with scikit-learn? Same patten, scikit-learn is the pip install but import is sklearn 🤔

evalml/utils/gen_utils.py
bchen1116: Makes sense. I wanted to make it a util to reduce repetitive code, but now that @chukarsten helped me whittle down the code a lot more, I guess it isn't as necessary for this one. I've left it as a method in `gen_utils` and changed the name to something hopefully more useful, but also open to just having it as a method in the base classes for the data splitters.

evalml/tests/utils_tests/test_dependencies.py
bchen1116: I think it's because this test looks only at the extra requirements in the `requirements.txt` file, not including the modules in `core_requirements.txt`, which is where scikit-learn is.

bchen1116:
Cleaned perf test doc updated [here](https://alteryx.quip.com/VVTFAfbT1Nwm/Data-Sampler-Class-Imbalance-tests-Cleaned)

dsherry:
@bchen1116 I'm excited to review this!

Since this PR doesn't update our default splitting strategy in automl, this won't close #973, correct?

bchen1116:
@dsherry you're right! I'll fix that.

dsherry:
@bchen1116 I'm reviewing now, will get you comments soon!

evalml/preprocessing/data_splitters/base_splitters.py
```diff
@@ -0,0 +1,92 @@
...
+    _convert_numeric_dataset,
+    _convert_to_woodwork_structure
+)
+
+
+class BaseTVSplit(BaseCrossValidator):
+    """Base class for training validation data splitter."""
+
+    def __init__(self, sampler=None, test_size=None, random_state=0):
+        self.sampler = sampler
```
ParthivNaresh, line 14: Should we add docstrings to the init functions?

ParthivNaresh (approved):
Excellent work, the tests are really comprehensive! I just have some documentation/typo related changes but everything else looks awesome

evalml/automl/utils.py
```diff
@@ -51,16 +57,29 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
     Returns:
```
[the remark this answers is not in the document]
bchen1116: What do you mean for arguments?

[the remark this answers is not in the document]
ParthivNaresh: For `make_data_splitter`, the arguments section looks empty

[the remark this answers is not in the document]
bchen1116: @ParthivNaresh should be fixed!

evalml/preprocessing/data_splitters/smotenc_split.py
```diff
@@ -0,0 +1,49 @@
+from evalml.preprocessing.data_splitters.base_splitters import (
+    BaseCVSplit,
+    BaseTVSplit
+)
+from evalml.utils import import_or_raise
+
+
+def _allowed_categorical(categorical_features):
```
bchen1116, line 8: Determines whether the categorical_feature array is a valid input. Since we don't pass the X training data here, we can't do checks on if the length of the input array is longer than the X array itself.

evalml/automl/automl_search.py
dsherry: Could we move the 'None' value into LG instead? Let's keep this logic as simple as we can.

evalml/automl/automl_search.py
```diff
@@ -94,6 +94,8 @@ def __init__(self,
                  problem_configuration=None,
                  train_best_pipeline=True,
                  pipeline_parameters=None,
+                 sampler=None,
```
dsherry, line 97: Are we ready to make this "public"? If not we should change the name to `_sampler`. If we are ready, perhaps this should be `data_splitting_sampler`?

evalml/automl/automl_search.py
```diff
@@ -94,6 +94,8 @@ def __init__(self,
...
+                 categorical_columns=None,
```
dsherry, line 98: @bchen1116 could you please explain why this input is necessary? If I follow right, we should be using woodwork to get this information.

evalml/automl/automl_search.py
dsherry: I suggest you use `hasattr` instead

evalml/automl/automl_search.py
```diff
@@ -581,6 +591,8 @@ def _find_best_pipeline(self):
                     X_train, X_threshold_tuning, y_train, y_threshold_tuning = split_data(X_train, y_train, self.problem_type,
                                                                                           test_size=0.2,
                                                                                           random_seed=self.random_seed)
+                if getattr(self.data_splitter, "transform", None):
+                    X_train, y_train = self.data_splitter.transform(X_train, y_train)
```
dsherry, line 595: Hmm, so we're applying this transformation in `_find_best_pipeline` but not during training?

evalml/automl/engine/engine_base.py
```diff
@@ -74,8 +74,12 @@ def train_and_score_pipeline(pipeline, automl, full_X_train, full_y_train):
...
                 break
             logger.debug(f"\t\tTraining and scoring on fold {i}")
-            X_train, X_valid = full_X_train.iloc[train], full_X_train.iloc[valid]
-            y_train, y_valid = full_y_train.iloc[train], full_y_train.iloc[valid]
+            if getattr(automl.data_splitter, "transform", None):
+                X_train, y_train = train[0], train[1]
+                X_valid, y_valid = valid[0], valid[1]
+            else:
+                X_train, X_valid = full_X_train.iloc[train], full_X_train.iloc[valid]
+                y_train, y_valid = full_y_train.iloc[train], full_y_train.iloc[valid]
```
dsherry, line 82: Oof, yeah. I would categorize this as tech debt. I understand this is necessary because the oversampling methods generate new rows, whereas the existing sklearn-style splitters we support do not.

I'll leave this question here for consideration although I don't have an answer right away: is there anything we can do to our existing data splitter API to make this easier?

dsherry (changes requested):
@bchen1116 sorry it took me so long to finish up here! It is cool that we're almost ready to do a first merge for oversampling code.

I had a few questions/suggestions about the approach. I also don't think we're ready to update `make_data_splitter`, which updates the default used by automl.

docs/source/user_guide/automl.ipynb
```diff
@@ -482,6 +482,88 @@
...
+    "X_all_numeric[['categorical', 'category_2']] = X_all_numeric[['categorical', 'category_2']].apply(lambda col: pd.factorize(col, sort=True)[0])\n",
+    "\n",
+    "# using a mask to specify categorical columns\n",
+    "cat_feat = [False, True, True]\n",
+    "# can also use indices\n",
+    "# cat_feat = [1, 2]\n",
+    "\n",
+    "splitter = SMOTENCTVSplit(categorical_features=cat_feat)\n",
+    "automl = AutoMLSearch(X_train=X_all_numeric, y_train=y, problem_type='binary', data_splitter=splitter)\n",
+    "automl.search()"
```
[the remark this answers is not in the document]
bchen1116: That makes sense! I added it to show readers what to expect, but it's probably pretty self-explanatory.

evalml/automl/automl_search.py
```diff
@@ -267,9 +270,9 @@ def __init__(self,
         # make everything ww objects
         self.X_train = _convert_to_woodwork_structure(X_train)
         self.y_train = _convert_to_woodwork_structure(y_train)
-
+        self.sampler = sampler if (sampler is not None and sampler != 'None') else None
```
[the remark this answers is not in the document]
dsherry: Oops sorry, this comment is out of date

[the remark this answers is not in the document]
bchen1116: @dsherry I added the `sampler` argument so that users can pass in a string if they wanted to use a sampled `data_splitter`. I figured it would be a little easier for them to switch between different sampling methods compared to having to import them, but that was one of the questions I raised in the description of this PR. Do you think it'd make sense to only accept the `data_splitter` arg and remove `sampler`?

evalml/automl/automl_search.py
freddyaboulton: Yea I agree that it'd be nice if we could avoid adding this argument to automl search.

evalml/automl/utils.py
```diff
@@ -51,16 +57,29 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
+            return {
+                "KMeansSMOTETVSplit": KMeansSMOTETVSplit(),
+                "KMeansSMOTECVSplit": KMeansSMOTECVSplit(),
+                "SMOTETomekTVSplit": SMOTETomekTVSplit(),
+                "SMOTETomekCVSplit": SMOTETomekCVSplit(),
+                "RandomUnderSamplerTVSplit": RandomUnderSamplerTVSplit(),
+                "RandomUnderSamplerCVSplit": RandomUnderSamplerCVSplit()
+            }[sampler]
+        except KeyError:
+            raise ValueError("Provided key {} does not exist for samplers".format(sampler))
```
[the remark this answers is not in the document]
bchen1116: @dsherry got it!

evalml/automl/automl_search.py
bchen1116: @dsherry, both the `split` and `transform` method resample the data. `transform` just doesn't also handle the splitting.

evalml/utils/woodwork_utils.py
```diff
@@ -98,3 +100,23 @@ def _retain_custom_types_and_initalize_woodwork(old_datatable, new_dataframe, lt
             except (ValueError, TypeError):
                 pass
     return ww.DataTable(new_dataframe, logical_types=retained_logical_types)
+
+
+def _convert_numeric_dataset_for_data_sampler(X, y, to_pandas=True):
```
dsherry, line 105: @bchen1116 could you please split this into two methods instead of using this `to_pandas` flag? Its a creative solution, but I think the code will be easier to understand and debug if our methods have one possible input type and one possible output type.

evalml/tests/utils_tests/test_dependencies.py
dsherry: 👍 

Once we add full support for oversampling, I hope we can either move `imbalanced-learn` to `core-requirements.txt` or remove it in favor of our own impls. At that point, this won't be an issue, right?

dsherry (approved):
@bchen1116 this is looking pretty good to me!

I had two main comments. Neither blocking merge but we should address them either in this PR or a follow-on PR:

1. Let's clean up the inheritance structure we're using. Before this PR we had `TrainingValidationSplit` which inherits from sklearn `BaseCrossValidator`. Now we'd like to add a few splitters which perform undersampling or oversampling.

I also see there's duplicated code between `BaseTVSplit` and `BaseCVSplit`.

My suggestion:
* Define one class `BaseSamplingSplitter` which inherits from `BaseCrossValidator`.
* That's where you can put all the changes that are currently in your two base classes
* Have all the oversamplers inherit from `BaseSamplingSplitter`
* The TV methods can just ignore the constructor value for `n_splits` and hard-code it to 1, and the CV methods can keep `n_splits` exposed via the constructor

2. I left a comment about `_convert_numeric_dataset_for_data_sampler`. I think we should split that into two methods, one for each direction we need.

evalml/preprocessing/data_splitters/base_splitters.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Arguments:
+                X (ww.DataTable): DataTable of points to split
+                y (ww.DataTable): DataColumn of points to split
+
+        Returns:
+            tuple((pd.DataFrame, pd.Series), (pd.DataFrame, pd.Series)): A tuple containing the resulting ((X_train, y_train), (X_test, y_test)) post-transformation.
+        """
+        X, y = _convert_numeric_dataset(X, y)
+        X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=self.test_size, random_state=self.random_seed)
```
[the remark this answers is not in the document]
bchen1116: It defaulted to returning pandas instead of woodwork. Since we no longer need to return woodwork datatables post-split, I can remove the support for returning woodwork in that util.

PR #1781, opened 2021-02-02
1434 prediction explanations dataframe output

### LINKED ISSUE #1434: Prediction explanations: output pandas dataframe
For easy programmatic access, let's update our prediction explanation feature to output a pandas dataframe.

We can choose to keep the current human-readable print format, as long as we have an intermediate representation in a dataframe.

This comes from feedback from the Oct 2020 blitz.

freddyaboulton:
### Pull Request Description
Fixes #1434 

-----

### DIFF
--- a/docs/source/user_guide/model_understanding.ipynb
+++ b/docs/source/user_guide/model_understanding.ipynb
@@ -454,21 +454,19 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "report.feature_names.value_counts()"
+    "### Changing Output Formats\n",
+    "\n",
+    "Instead of getting the prediction explanations as text, you can get the report as a python dictionary or pandas dataframe. All you have to do is pass `output_format=\"dict\"` or `output_format=\"dataframe\"` to either `explain_prediction`, `explain_predictions`, or `explain_predictions_best_worst`."
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Changing Output Formats\n",
-    "\n",
-    "Instead of getting the prediction explanations as text, you can get the report as a python dictionary. All you have to do is pass `output_format=\"dict\"` to either `explain_prediction`, `explain_predictions`, or `explain_predictions_best_worst`."
+    "### Single prediction as a dictionary"
    ]
   },
   {
@@ -478,6 +476,44 @@
    "outputs": [],
    "source": [
     "import json\n",
+    "single_prediction_report = explain_prediction(pipeline=pipeline, input_features=X.iloc[3:4],\n",
+    "                                              training_data=X, top_k=6, include_shap_values=True,\n",
+    "                                              output_format=\"dict\")\n",
+    "print(json.dumps(single_prediction_report, indent=2))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Single prediction as a dataframe"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "single_prediction_report = explain_prediction(pipeline=pipeline, input_features=X.iloc[3:4],\n",
+    "                                              training_data=X, top_k=6, include_shap_values=True,\n",
+    "                                              output_format=\"dataframe\")\n",
+    "single_prediction_report"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Best and worst predictions as a dictionary"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "report = explain_predictions_best_worst(pipeline=pipeline, input_features=X, y_true=y,\n",
     "                                        num_to_explain=1, top_k_features=6,\n",
     "                                        include_shap_values=True, output_format=\"dict\")\n",
@@ -485,14 +521,10 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "report = explain_predictions_best_worst(pipeline=pipeline, input_features=X, y_true=y,\n",
-    "                                        num_to_explain=1, top_k_features=6,\n",
-    "                                        include_shap_values=True, output_format=\"dataframe\")"
+    "### Best and worst predictions as a dataframe"
    ]
   },
   {
@@ -501,6 +533,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "report = explain_predictions_best_worst(pipeline=pipeline, input_features=X, y_true=y,\n",
+    "                                        num_to_explain=1, top_k_features=6,\n",
+    "                                        include_shap_values=True, output_format=\"dataframe\")\n",
     "report"
    ]
   }
--- a/docs/source/user_guide/model_understanding.ipynb
+++ b/docs/source/user_guide/model_understanding.ipynb
@@ -473,6 +473,100 @@
     "                                        include_shap_values=True, output_format=\"dict\")\n",
     "print(json.dumps(report, indent=2))"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "\n",
+    "def core(explanation):\n",
+    "    return {\"prefix\": explanation['rank']['prefix'],\n",
+    "          \"rank\": explanation['rank']['index'],\n",
+    "          \"predicted_probabilities\": explanation['predicted_values'][\"probabilities\"],\n",
+    "          \"predicted_value\": explanation['predicted_values']['predicted_value'],\n",
+    "          \"target_value\": explanation['predicted_values']['target_value'],\n",
+    "          \"error_metric\": explanation['predicted_values']['error_name'],\n",
+    "          \"error_value\": explanation['predicted_values']['error_value'],\n",
+    "          \"index_in_data\": explanation['predicted_values']['index_id']}\n",
+    "\n",
+    "\n",
+    "def dict_to_series(explanation):\n",
+    "    r = core(explanation)\n",
+    "    r[\"explanations\"] = explanation['explanations']\n",
+    "    return [r]\n",
+    "\n",
+    "\n",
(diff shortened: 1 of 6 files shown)


evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -399,30 +491,51 @@ def test_output_format_checked():
 @pytest.mark.parametrize("problem_type,output_format,answer,explain_predictions_answer,custom_index",
```
chukarsten, line 491: This is an epic parameterization.

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -207,13 +247,13 @@ def test_explain_predictions_value_errors():
...
+    with pytest.raises(ValueError, match="Parameter output_format must be either text, dict, or dataframe. Received bar"):
         explain_predictions(None, input_features, output_format="bar")
-    with pytest.raises(ValueError, match="Parameter output_format must be either text or dict. Received xml"):
+    with pytest.raises(ValueError, match="Parameter output_format must be either text, dict, or dataframe. Received xml"):
         explain_prediction(None, input_features=input_features, training_data=None, output_format="xml")
 
     input_features, y_true = pd.DataFrame(data=range(15)), pd.Series(range(15))
-    with pytest.raises(ValueError, match="Parameter output_format must be either text or dict. Received foo"):
+    with pytest.raises(ValueError, match="Parameter output_format must be either text, dict, or dataframe. Received foo"):
         explain_predictions_best_worst(None, input_features, y_true=y_true, output_format="foo")
```
chukarsten, line 257: I think we can probably get rid of the foo and bar tests, right?  They don't add anything the "xml" one does.

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -399,30 +491,51 @@ def test_output_format_checked():
...
+        mock_make_table.return_value = "table goes here"
+    elif output_format == "dataframe":
+        shap_table = pd.DataFrame({
+            "feature_names": [0],
+            "feature_values": [0],
+            "qualitative_explanation": [0],
+            "quantitative_explanation": [0],
+        })
+        # Use side effect so that we always get a new copy of the dataframe
+        mock_make_table.side_effect = lambda *args, **kwargs: shap_table.copy()
```
chukarsten, line 535: This is fancy and cool.  I'm going to need to go figure out what this does.

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -475,54 +592,24 @@ def _add_custom_index(answer, index_best, index_worst, output_format):
                                                        num_to_explain=1, output_format=output_format)
     if output_format == "text":
         compare_two_tables(best_worst_report.splitlines(), answer.splitlines())
+    elif output_format == "dataframe":
+        # Check dataframes equal without caring about column order
+        assert sorted(best_worst_report.columns.tolist()) == sorted(answer.columns.tolist())
+        pd.testing.assert_frame_equal(best_worst_report, answer[best_worst_report.columns])
```
chukarsten, line 598: Won't assert_frame_equal compare the columns as well?  L597 might be redundant?

chukarsten (approved):
This looks good to me!

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
freddyaboulton: 100%

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
freddyaboulton: Yes but in the event `best_worst_report` has fewer columns than expected, the check will pass because `answer[best_worst_report.columns]` (needed to reorder the columns) will subset to the columns in `best_worst_report`. 

The root cause is that I modify the `index_id` column at the beginning of the test to add the expected custom index and this unfortunately messes up the column order. I don't think there's a way to get assert_frame_equal to ignore the column order so this two-step check verifies the same columns are present in both and that they contain the same values. I think this is ok given how many different scenarios we're testing. The alternative would be to manually specify the answer for each test but that would be hairy.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -423,3 +451,25 @@ def make_dict(self, data):
...
+        for rank, index in enumerate(data.index_list):
+            shap_table = self.table_maker.make_dataframe(index, data.pipeline, data.input_features)
+            shap_table['rank'] = rank
+            if self.make_predicted_values_maker:
+                heading = self.make_predicted_values_maker.make_dataframe(index, data.y_pred, data.y_true, data.errors,
+                                                                          pd.Series(data.input_features.index))
+                for key, value in heading.items():
+                    if key == "probabilities":
+                        for class_name, probability in value.items():
+                            shap_table[f'label_{class_name}_probability'] = probability
```
dsherry, line 466: Can we replace this `for` with
```python
labels = [f'label_{class_name}_probability' for class_name in value.keys()]
shap_table[labels] = value.values()
```
I guess that code would rely on `value.values()` being ordered in the same way as `value.keys()`, and I don't see this set up as an `OrderedDict`, so maybe what you have is fine :)

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -423,3 +451,25 @@ def make_dict(self, data):
...
+                    else:
+                        shap_table[key] = value
+            if self.heading_maker:
+                heading = self.heading_maker.make_dataframe(rank)
+                shap_table['rank'] = heading['index']
+                shap_table['prefix'] = heading['prefix']
+            report.append(shap_table)
+        df = pd.concat(report).reset_index(drop=True)
```
dsherry, line 474: So this code will create one df for each prediction, and then concatenate all the columns together horizontally? How do users know which columns correspond to which prediction?

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -423,3 +451,25 @@ def make_dict(self, data):
                                                                  data.input_features)["explanations"]
             report.append(section)
         return {"explanations": report}
+
+    def make_dataframe(self, data):
```
dsherry, line 455: Is there some way we can have this implementation use `_make_single_prediction_shap_table` to make the tables per-prediction, and then concatenate them together? This would avoid confusion and bugs stemming from current or future differences in the implementations.

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -448,8 +448,18 @@
    "source": [
     "from evalml.model_understanding.prediction_explanations import explain_predictions\n",
     "\n",
-    "report = explain_predictions(pipeline=pipeline, input_features=X.iloc[[0, 4, 9]], include_shap_values=True)\n",
-    "print(report)"
+    "report = explain_predictions(pipeline=pipeline, input_features=X.iloc[[0, 4, 9]], include_shap_values=True,\n",
+    "                            output_format='dataframe')\n",
```
dsherry, line 452: @freddyaboulton could you explain how to read the output here?

Screen Shot 2021-02-03 at 12 26 19 PM

Each grouping of rows corresponds to a different prediction explanation, yes? And that's indicated by the "rank" column. What confuses me is that "rank" ranges from 0 to n-1 whereas the indices you passed to `input_features` are 0, 4 and 9. I'm concerned someone looking at the dataframe output will have trouble matching those row groups up with the original dataframe row that info corresponds to.

If I'm following correctly here, I suggest we either find a way to get the original dataframe index into this dataframe report, or we change our language a bit and call this "prediction number" instead of "rank". The latter option seems more appealing.

What do you think?

dsherry (approved):
@freddyaboulton this is looking great!

I left some comments and questions on the implementation. Nothing blocking, but it would be great to close out those conversations soon.

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -448,8 +448,18 @@
    "source": [
     "from evalml.model_understanding.prediction_explanations import explain_predictions\n",
```
dsherry, line 449: @freddyaboulton I see the updates below are for the "Explaining Multiple Predictions" section. Should we also update the "Explaining Individual Predictions" section? I see that example still uses `print(table)`
Screen Shot 2021-02-03 at 12 32 53 PM

evalml/model_understanding/prediction_explanations/_user_interface.py
freddyaboulton: I think that's what we're doing! `self.table_maker.make_dataframe(index, data.pipeline, data.input_features)` calls `_make_single_prediction_shap_table` to make the shap table per prediction. Then we add metadata to each table (prediction probabilities, etc) and then we concat all the tables at the end.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -246,6 +253,10 @@ def make_text(self, *args, **kwargs):
     def make_dict(self, *args, **kwargs):
         """Makes the report section formatted as a dictionary."""
 
+    @abc.abstractmethod
+    def make_dataframe(self, *args, **kwargs):
+        """Makes the report section formatted as a dictionary."""
```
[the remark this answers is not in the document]
freddyaboulton: Yes! 😅

docs/source/user_guide/model_understanding.ipynb
freddyaboulton: @dsherry You're spot on about the grouping! I agree with the rename of `rank` to `prediction_number`. Adding the original dataframe index is possible but we don't do that in `explain_prediction` for the other output formats.

freddyaboulton: image

docs/source/user_guide/model_understanding.ipynb
freddyaboulton: I added a section on changing the output format for `explain_prediction` in the "Changing output formats" section!

evalml/model_understanding/prediction_explanations/_user_interface.py
freddyaboulton: To resolve: They'll use the `prediction_number` column we talked about above!

evalml/model_understanding/prediction_explanations/_user_interface.py
freddyaboulton: Yea I think we can add another for-comprehension to make sure the order is the same but at that point the main benefit of that implementation over this one is saving an indent. So I think I'll stick to this one!

Issue #1504, opened 2020-12-03
Prediction explanations should support time series problems


freddyaboulton:
As an evalml user, I'd like to be able to use `explain_prediction`, `explain_predictions`, and `explain_prediction_best_worst` to analyze the predictions of my time series pipeline.

I think there are two main things to do:

* Our pipeline predict method now accepts `y` so the api of the explainer functions should reflect that.
* Add a default metric for time series problems in `explain_predictions_best_worst`

freddyaboulton:
Another thing we should consider is that `explain_prediction` and `explain_predictions` don't really work for time series. Right now, the user passes in a subset of independent row(s) as the input data and we generate a prediction for each row. 

That api doesn't really work for time series because the features of the current row depend on previous rows. If the previous rows are not present in the data the user passes in, then our features will have NaNs and we'll error out. 

I think there are a couple of things we can do:

1. Don't let users pass in time series pipelines for `explain_prediction` and `explain_predictions`
2. Keep the api the same but edit the docstring and user guide to let users know the data must be contiguous. I think we'll still have to be careful and check that the user passes in "enough" data to generate features.
3. Change the api of `explain_prediciton` and `explain_predictions`. The user passes in the entire dataset and the indices of the rows they want to explain. This is similar to know `explain_predictions_best_worst` works but instead of computing the indices of the rows to explain with a ranking function, the user passes in the ranking themselves. 

I like option 3 the best because I'm worried users will easily do the wrong thing with option 2. Option 1 doesn't seem like a viable long term solution.

Any additional options I'm overlooking?

chukarsten:
@dsherry @freddyaboulton This has been a doozy.  

1. What do we think the behavior should be in the case where we attempt to call explain_predictions() on a pipeline that's fit with a certain time delay and the requested feature to predict on requires us to delay before the beginning of the time series data?
2. Shouldn't explain_predictions() just be an iterative implementation of explain_prediction()?

PR #1782, opened 2021-02-02
Prediction Explanations for Time Series

chukarsten:
addresses #1504 
-----


evalml/model_understanding/prediction_explanations/explainers.py
```diff
@@ -20,11 +20,11 @@
 )
 
 # Container for all of the pipeline-related data we need to create reports. Helps standardize APIs of report makers.
-_ReportData = namedtuple("ReportData", ["pipeline", "input_features",
+_ReportData = namedtuple("ReportData", ["pipeline", "input_features", "training_data",
                                         "y_true", "y_pred", "y_pred_values", "errors", "index_list", "metric"])
 
 
-def explain_prediction(pipeline, input_features, top_k=3, training_data=None, include_shap_values=False,
+def explain_prediction(pipeline, input_features, y=None, top_k_features=3, training_data=None, include_shap_values=False,
```
freddyaboulton, line 27: I was thinking that we would add an argument called `index_to_explain` here and `indices_to_explain` to `explain_predictions` and have the user pass in the entire training data for `input_features`.

`_make_single_prediction_shap_table` would stay the same but we would subset the pipelines indices to these rows, e.g `pipeline_features = pipeline.compute_estimator_features(input_features).to_dataframe().iloc[indices_to_explain]`

If any of these rows contain NaNs in the features, then we raise a value error. 

I think this would be better than manually checking if the DelayedFeatureTransformer is in the pipeline and modifying the index. (We will likely add new time series transformers that modify the data in the future)

chukarsten:
Started over again.

PR #1784, opened 2021-02-02
Support custom woodwork types in components

### LINKED ISSUE #1662: Custom woodwork types aren't preserved in components

angela97lin:
Closes #1662. Another large-ish PR but not too much logic change--just replacing the final line to convert to WW to new util method, and writing a lot of tests. I initially wanted to iterate through all the components and test like the pattern we already have in `test_components.py` but realized that each component has its own nuanced behavior, so decided against it and added the tests to their individual files. The parameterization is similar in many cases though. 😬 

- Moves Woodwork-related util methods to `woodwork_utils.py` and move tests to `test_woodwork_utils.py`
- Add private utility method to help best reassign logical types after transformation.
- Merged `_convert_to_woodwork_structure` with `infer_feature_types` by combining the two implementations.

Notes on compatible and incompatible types. Mostly for my own understanding, but also helped inform test cases.

**Datetime**
- Datetime --> Double: errors
- Datetime --> Integer: errors
- Datetime --> Categorical: works
- Datetime --> Natural Language: works

**Categorical** 
- Categorical --> Datetime: errors
- Categorical --> Double: works, sometimes. ("1", "2" okay, "a", "b" not okay)
- Categorical --> Natural Language: works
- Categorical --> Integer: errors

**Integer**
- Integer --> Datetime: works... but isn't great lol (literally .000000001 ms)
- Integer --> Double: works
- Integer --> Categorical: works
- Integer --> Natural Language: works

**Double**
- Double --> Datetime: works... but isn't great lol (literally .000000001 ms)
- Double --> Integer: works sometimes (if possible to convert)
- Double --> Categorical: works
- Double --> Natural Language: works

**Natural Language **
- Natural Language --> Datetime: errors
- Natural Language --> Double: works, sometimes
- Natural Language --> Categorical: works
- Natural Language --> Integer: works, sometimes

- There's no good way of confirming when it is safe to convert floats to category or when it is not without trying. For example: can convert ['1', '2'] (categories) to float, but not ['a', 'b']. Hence, current implementation will try to set the type. If it fails, then don't do anything. If it succeeds, add to logical type dictionary to pass along to the initialization of Woodwork DataTable, so that Woodwork does not have to do type inference, saving us some time.
- Look into: featuretools component which changed the columns of the original input DF?


docs/source/api_reference.rst
```diff
@@ -572,3 +572,5 @@ General Utils
     drop_rows_with_nans
     infer_feature_types
     save_plot
+    is_all_numeric
```
angela97lin, line 575: Unrelated change but adding to API ref :)

evalml/utils/gen_utils.py
```diff
@@ -17,8 +17,6 @@
 
 logger = get_logger(__file__)
 
-numeric_and_boolean_ww = [ww.logical_types.Integer, ww.logical_types.Double, ww.logical_types.Boolean]
```
angela97lin, line 20: Moved all Woodwork related utils to `woodwork_utils.py` :)

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -117,7 +117,7 @@ def test_simple_imputer_all_bool_return_original(data_type, make_data_type):
 
 @pytest.mark.parametrize("data_type", ['pd', 'ww'])
 def test_simple_imputer_bool_dtype_object(data_type, make_data_type):
-    X = pd.DataFrame([True, np.nan, False, np.nan, True], dtype=object)
+    X = pd.DataFrame([True, np.nan, False, np.nan, True], dtype='boolean')
```
angela97lin, line 120: This update (and many of the others made in the tests) was made because now we try to convert the columns back to the original types. Before, the original dtype was specified as object, so the column would have been stored as a categorical col and we would have expected X_expected_arr to return a categorical col. We need to explicitly set to boolean for the expected outcome to be boolean. 

I think these conversions make sense over previous behavior--user input is boolean, so output is boolean.

evalml/pipelines/components/transformers/transformer.py
```diff
@@ -34,17 +35,17 @@ def transform(self, X, y=None):
         Returns:
             ww.DataTable: Transformed X
         """
+        X_ww = _convert_to_woodwork_structure(X)
```
angela97lin, line 38: I decided to move these out of the try/except because I think it's more clear what the try/except is for. If it errors out while doing Woodwork conversions (as it did while I was testing stuff), it shouldn't throw a `MethodPropertyNotFoundError` :)

evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -271,153 +268,6 @@ def test_rename_column_names_to_numeric():
     assert X_renamed.logical_types == {0: ww.logical_types.Categorical, 1: ww.logical_types.Categorical}
 
 
-def test_convert_woodwork_types_wrapper_with_nan():
```
angela97lin, line 274: Moved to `test_woodwork_utils.py`

evalml/pipelines/component_graph.py
```diff
@@ -200,17 +200,6 @@ def _compute_features(self, component_list, X, y=None, fit=False):
                         parent_x = pd.Series(_convert_woodwork_types_wrapper(parent_x.to_series()), name=parent_input)
                     x_inputs.append(parent_x)
             input_x, input_y = self._consolidate_inputs(x_inputs, y_input, X, y)
-            col_intersection = set(X.columns.keys()).intersection(set(input_x.columns.keys()))
```
angela97lin, line 203: We don't need this logic now, since components handle this 🥳

angela97lin:
@thehomebrewnerd @tamargrey Added you guys because I thought it'd be great if you guys could take a look at `_retain_custom_types_and_initalize_woodwork` in `woodwork_utils.py` if you get the chance! 😁

evalml/utils/woodwork_utils.py
```diff
@@ -0,0 +1,112 @@
+
+import numpy as np
+import pandas as pd
+import woodwork as ww
+
+numeric_and_boolean_ww = [ww.logical_types.Integer, ww.logical_types.Double, ww.logical_types.Boolean]
```
thehomebrewnerd, line 6: Maybe not too important for now, but note that this could fail to capture everything if a user defines a custom numeric or boolean type in the TypeSystem...could be something like WholeNumber or Currency.

evalml/utils/woodwork_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Returns:
+        A Woodwork data structure where the data type of each column was either specified or inferred.
+    """
+    ww_data = _convert_to_woodwork_structure(data)
+    if feature_types is not None:
+        if len(ww_data.shape) == 1:
+            ww_data = ww_data.set_logical_type(feature_types)
+        else:
+            ww_data = ww_data.set_types(logical_types=feature_types)
```
thehomebrewnerd, line 27: This function would probably be more efficient if you could pass in any specified logical types to the woodwork initialization call in `_convert_to_woodwork_structure`. With the current approach, you end up inferring the types for all columns and then possibly changing them. If you could pass any known logical types to the woodwork init call, you could avoid the inference cost for those columns.

evalml/utils/woodwork_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    retained_logical_types = {}
+    if ltypes_to_ignore is None:
+        ltypes_to_ignore = []
+    col_intersection = set(old_datatable.columns).intersection(set(new_dataframe.columns))
+    logical_types = old_datatable.logical_types
+    for col in col_intersection:
+        if logical_types[col] in ltypes_to_ignore:
+            continue
+        try:
+            new_dataframe[col].astype(logical_types[col].pandas_dtype)
```
tamargrey, line 108: Would you be able to skip this check if `new_dataframe[col].dtype == logical_types[col].pandas_dtype`?

evalml/utils/woodwork_utils.py
```diff
@@ -0,0 +1,112 @@
...
+            retained_logical_types[col] = old_datatable[col].logical_type
+        except (ValueError, TypeError):
+            pass
```
thehomebrewnerd, line 111: I think the `astype` check here could be computationally expensive for large datasets. I think you should only need to do it if `logical_types[col].pandas_dtype != new_dataframe[col].dtype`.

thehomebrewnerd (commented):
One other general comment here. I see a lot of calls that look like this in the code:
```python
        X_ww = _convert_to_woodwork_structure(X)
        X_t = _convert_woodwork_types_wrapper(X_ww.to_dataframe())
```
If I'm understanding this correctly this is because Woodwork is using dtypes that numpy doesn't like. With this it seems we are doing a lot of back-and-forth dtype conversions. It might be worth looking into whether you should override the Woodwork default type system and use `pandas_dtype` and `backup_dtype` values that would not require these conversions, which I think might be expensive.

I'm not completely sure this would work, but might be worth investigating to improve performance at some point.

angela97lin:
@tamargrey @thehomebrewnerd thank you for your comments, they're super helpful!! Nate, I shelved the suggestion to override Woodwork's default type system in https://github.com/alteryx/evalml/issues/1796, but I'm really excited to see if we can implement that and manage to get better performance numbers from it 😁

evalml/utils/woodwork_utils.py
angela97lin: Yes, you're totally right and this falls within the discussion we had about updating custom types. Going to remove `_convert_to_woodwork_structure` and move the impl to `infer_feature_types` since they're basically the same.

evalml/tests/component_tests/test_simple_imputer.py
freddyaboulton: Agreed! Glad we don't have to do this anymore.

evalml/tests/component_tests/test_standard_scaler.py
```diff
@@ -0,0 +1,28 @@
...
+            X = ww.DataTable(X_df, logical_types={0: logical_type})
+        except TypeError:
+            continue
+
+        std_scaler = StandardScaler()
+        std_scaler.fit(X, y)
+        transformed = std_scaler.transform(X, y)
+        assert isinstance(transformed, ww.DataTable)
+        if logical_type == Categorical:
+            assert transformed.logical_types == {0: Categorical}
```
freddyaboulton, line 26: I agree this is the intended behavior of this pr but I think it's also confusing! If I input a categorical column with values [1., 2., 3.], at the end I have a categorical column with values [-1.2247, 0, 1.2247]. 

I think we should change the scaler to ignore categorical columns.

evalml/tests/component_tests/test_imputer.py
```diff
@@ -327,3 +335,31 @@ def test_imputer_multitype_with_one_bool(data_type, make_data_type):
...
+        except TypeError:
+            continue
+
+        imputer = Imputer()
+        imputer.fit(X, y)
+        transformed = imputer.transform(X, y)
+        assert isinstance(transformed, ww.DataTable)
+
+        if logical_type == Integer:
+            assert transformed.logical_types == {0: Double}
```
freddyaboulton, line 363: How come integers are never carried over? I understand that if  if the numeric_impute_strategy is "mean" then you can get doubles at the end. But what if the impute strategy is mode?

evalml/tests/component_tests/test_delayed_features_transformer.py
```diff
@@ -272,3 +280,32 @@ def test_delay_feature_transformer_y_is_none(delayed_features_data):
...
+        try:
+            X = ww.DataTable(X_df, logical_types={0: logical_type})
+        except TypeError:
+            continue
+        dft = DelayedFeatureTransformer(max_delay=1, gap=11)
+        dft.fit(X, y)
+        transformed = dft.transform(X, y)
+        assert isinstance(transformed, ww.DataTable)
+        if logical_type == Integer or logical_type == Double or logical_type == Categorical:
+            assert transformed.logical_types == {0: logical_type,
```
freddyaboulton, line 303: I think we should modify the delayed feature transformer to carry over the ints. I'd be happy to file an issue if you are on board!

evalml/tests/component_tests/test_datetime_featurizer.py
```diff
@@ -139,3 +147,33 @@ def test_datetime_featurizer_numpy_array_input():
...
+
+
+@pytest.mark.parametrize("X_df", [pd.DataFrame(pd.to_datetime(['20190902', '20200519', '20190607'], format='%Y%m%d')),
+                                  pd.DataFrame(pd.Series([1, 2, 3], dtype="Int64")),
+                                  pd.DataFrame(pd.Series([1., 2., 3.], dtype="float")),
+                                  pd.DataFrame(pd.Series(['a', 'b', 'a'], dtype="category")),
+                                  pd.DataFrame(pd.Series([True, False, True], dtype="boolean")),
+                                  pd.DataFrame(pd.Series(['this will be a natural language column because length', 'yay', 'hay'], dtype="string"))])
+@pytest.mark.parametrize("with_datetime_col", [True, False])
+def test_datetime_featurizer_woodwork_custom_overrides_returned_by_components(X_df, with_datetime_col):
```
freddyaboulton, line 159: Can you add coverage for `encode_as_categories=True`?

freddyaboulton (approved):
@angela97lin Looks good to me! I have some questions on the unit tests but nothing blocking.

evalml/utils/woodwork_utils.py
angela97lin: Yes, you're right! I filed https://github.com/alteryx/evalml/issues/1810 to track this, thank you!

evalml/tests/component_tests/test_delayed_features_transformer.py
angela97lin: Sure, that sounds good! I'm guessing `shift` just returns floats which is why we end up with Doubles right now?

evalml/tests/component_tests/test_datetime_featurizer.py
angela97lin: Yes, thank you for calling this out!

evalml/tests/component_tests/test_imputer.py
angela97lin: Hehe, you're right, this is dependent on impute strategy. Going to update `test_simple_imputer_woodwork_custom_overrides_returned_by_components` and `test_imputer_woodwork_custom_overrides_returned_by_components` to handle testing each of the different strategies. 😁

evalml/tests/component_tests/test_delayed_features_transformer.py
freddyaboulton: Yes!

evalml/tests/component_tests/test_imputer.py
```diff
@@ -327,3 +335,33 @@ def test_imputer_multitype_with_one_bool(data_type, make_data_type):
...
+def test_imputer_woodwork_custom_overrides_returned_by_components(X_df, has_nan, numeric_impute_strategy):
+    y = pd.Series([1, 2, 1])
+    if has_nan:
+        X_df.iloc[len(X_df) - 1, 0] = np.nan
+    override_types = [Integer, Double, Categorical, NaturalLanguage, Boolean]
+    for logical_type in override_types:
+        try:
+            X = ww.DataTable(X_df, logical_types={0: logical_type})
+        except TypeError:
+            continue
```
chukarsten, line 356: I think I'm tired, but what is the intent behind this chunk here?

chukarsten (approved):
Oh man.  Great work getting through all this and testing it all...what a feat of endurance.  Just a few questions I had here and there.  If you want to get back to me on them, just ping me on Slack so I can go check it out, but nothing to really stop merge.

evalml/tests/component_tests/test_imputer.py
angela97lin: Ah so basically, I'm trying every possible logical type conversion that's allowed (see desc of PR for allowed vs error), but rather than manually keeping track of all possible conversions, I'm just trying. Ex: Int --> Nat Lang is valid conversion and therefore a valid override a user might provide to EvalML for us to test. Nat Lang --> Int might not always be (and will raise a TypeError). If its not, don't try to pass to EvalML to test.

evalml/data_checks/data_checks.py
```diff
@@ -87,9 +87,9 @@ def validate(self, X, y=None):
             "warnings": [],
             "errors": []
         }
-        X = _convert_to_woodwork_structure(X)
+        X = infer_feature_types(X)
         if y is not None:
-            y = _convert_to_woodwork_structure(y)
+            y = infer_feature_types(y)
```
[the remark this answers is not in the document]
angela97lin: Honestly, I've seen this so often I swear I do too 😆

evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -63,8 +64,8 @@ def fit(self, X, y=None):
         Returns:
             self
         """
-        X = _convert_to_woodwork_structure(X)
-        cat_cols = list(X.select('category').columns)
+        X = infer_feature_types(X)
+        cat_cols = list(X.select(['category', 'boolean']).columns)
```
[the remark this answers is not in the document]
angela97lin: I can double check we're testing this but IIRC, it's so that we treat boolean columns like categorical cols. We don't want to pass it to our numeric_imputer which might try to use "mean" as the impute strategy and end up with nonsense for the imputed values.

evalml/pipelines/components/transformers/preprocessing/drop_null_columns.py
```diff
@@ -50,5 +47,7 @@ def transform(self, X, y=None):
         Returns:
             ww.DataTable: Transformed X
         """
-        X_t = _convert_to_woodwork_structure(X)
+        X_t = infer_feature_types(X)
```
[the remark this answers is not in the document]
angela97lin: WW does expose a similar `drop` method, but not with an error handling parameter 😢 : https://woodwork.alteryx.com/en/stable/generated/woodwork.datatable.DataTable.drop.html#woodwork.datatable.DataTable.drop

PR #1794, opened 2021-02-08
Update estimators to keep track of input feature names

### LINKED ISSUE #1757: Update estimators to keep track of input feature names
Right now, we don't have a good way to get the original feature names passed into an estimator. Pipelines have this information via `input_feature_names`. This could be a good addition, and would make #1718 much easier to tackle.

angela97lin:
Closes #1757


evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -103,9 +104,11 @@ def _encode_labels(self, y):
         return y_encoded
 
     def fit(self, X, y=None):
+        X_encoded = _convert_to_woodwork_structure(X)
         X_encoded = self._encode_categories(X, fit=True)
         y_encoded = self._encode_labels(y)
-        return super().fit(X_encoded, y_encoded)
+        self._component_obj.fit(X_encoded, y_encoded)
```
angela97lin, line 110: Updating to just call here, because Estimator will call `self.input_feature_names = list(X.columns)` but since we've changed the feature names in `_encode_categories` (since LGBM isn't compatible with weird names), this would override the original names.

evalml/pipelines/components/estimators/classifiers/xgboost_classifier.py
```diff
@@ -42,8 +46,15 @@ def __init__(self, eta=0.1, max_depth=6, min_child_weight=1, n_estimators=100, r
                          random_state=random_state)
 
     def fit(self, X, y=None):
+        X = _convert_to_woodwork_structure(X)
```
angela97lin, line 49: Similar to LGBM, updating to just call implementation, because Estimator will call `self.input_feature_names = list(X.columns)` but since we've changed the feature names in `_encode_categories` (since XGBoost isn't compatible with weird names), this would override the original names.

evalml/tests/component_tests/test_estimators.py
```diff
@@ -21,22 +27,24 @@ def test_estimators_feature_name_with_random_ascii(X_y_binary, X_y_multi, X_y_re
...
             col_names = ['column_{}'.format(ascii_char) for ascii_char in string.printable]
             X = pd.DataFrame(X, columns=col_names)
+            assert clf.input_feature_names is None
             clf.fit(X, y)
             assert len(clf.feature_importance) == len(X.columns)
             assert not np.isnan(clf.feature_importance).all().all()
             predictions = clf.predict(X).to_series()
             assert len(predictions) == len(y)
             assert not np.isnan(predictions).all()
+            assert (clf.input_feature_names == col_names)
```
angela97lin, line 47: Rather than create a whole new test, tacking it on here, where we test passing weird names to estimators :d

evalml/tests/component_tests/test_lgbm_classifier.py
```diff
@@ -168,8 +162,7 @@ def test_correct_args(mock_fit, mock_predict, mock_predict_proba, X_y_binary):
 
 @patch('evalml.pipelines.components.estimators.estimator.Estimator.predict_proba')
 @patch('evalml.pipelines.components.estimators.estimator.Estimator.predict')
-@patch('evalml.pipelines.components.component_base.ComponentBase.fit')
-def test_categorical_data_subset(mock_fit, mock_predict, mock_predict_proba, X_y_binary):
+def test_categorical_data_subset(mock_predict, mock_predict_proba, X_y_binary):
```
angela97lin, line 165: I had a hard time with these tests; I updated to remove the ComponentBase fit mock since it no longer calls that but wasn't sure if I was ripping out what the test was trying to test for in the first place.

My understanding is that it was trying to test if our encoding was passed to the super class' fit/predict/predict_proba implementations correctly. Since LGBM no longer calls the super class' impl, this isn't necessary?

chukarsten (approved):
I like this PR.  I like moving the functionality up into the parent class.  I *think* the convert X/y to woodwork can be moved out and reused a little bit more.  I submitted a PR to your PR, check it out and let me know if that works.  If not, no big deal.

evalml/pipelines/components/estimators/estimator.py
```diff
@@ -12,7 +12,7 @@
 class Estimator(ComponentBase):
     """A component that fits and predicts given data.
 
-    To implement a new Transformer, define your own class which is a subclass of Transformer, including
+    To implement a new Estimator, define your own class which is a subclass of Estimator, including
```
bchen1116, line 15: Good catch!

evalml/tests/component_tests/test_lgbm_classifier.py
bchen1116: Yep, sounds good to me!

bchen1116 (approved):
LGTM!

PR #1797, opened 2021-02-08
Data Health: Sparsity

chukarsten:
Implements #1797


evalml/data_checks/__init__.py
```diff
@@ -15,3 +15,4 @@
 from .class_imbalance_data_check import ClassImbalanceDataCheck
 from .high_variance_cv_data_check import HighVarianceCVDataCheck
 from .multicollinearity_data_check import MulticollinearityDataCheck
+from .sparsity_data_check import SparsityDataCheck
```
angela97lin, line 18: Nice!! We should add this to the API reference under Data Checks too 😁

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+                below which, a column exhibits sparsity.  Should be between 0 and 1.
+            unique_count_threshold (int): The minimum number of times a unique
+                value has to be present in a column to not be considered "sparse."
+        """
+        self.problem_type = handle_problem_types(problem_type)
+        if self.problem_type not in [ProblemTypes.MULTICLASS, ProblemTypes.TIME_SERIES_MULTICLASS]:
+            raise ValueError("Sparsity is only defined for multiclass problem types.")
+        self.threshold = threshold
+        if threshold < 0 or threshold > 1:
+            raise ValueError("Threshold must be a float between 0 and 1, inclusive.")
```
angela97lin, line 33: Error message is different from check--we're checking between 0 and 1 exclusively :o?

evalml/data_checks/data_check_message_code.py
```diff
@@ -57,3 +57,6 @@ class DataCheckMessageCode(Enum):
 
     IS_MULTICOLLINEAR = "is_multicollinear"
     """Message code for when data is potentially multicollinear."""
+
+    TOO_SPARSE = "too sparse"
+    """Message code for when binary or multiclass data has values that are too sparsely populated."""
```
angela97lin, line 62: Does this work for binary problems? Curious since the docstring / check is for multi-class/time-series multiclass

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+    """Checks if there are any columns with sparsely populated values in the input."""
+
+    def __init__(self, problem_type, threshold, unique_count_threshold=10):
+        """Checks each column in the input to determine the sparsity of the values in those columns.
+        Arguments:
+            problem_type (str or ProblemTypes): The specific problem type to data check for.
+                'multiclass' or 'time series multiclass' is the only accepted problem type.
+            threshold (float): The threshold value, or percentage of each column's unique values,
+                below which, a column exhibits sparsity.  Should be between 0 and 1.
+            unique_count_threshold (int): The minimum number of times a unique
```
bchen1116, line 25: Can we add a note of the default value here

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+                value has to be present in a column to not be considered "sparse."
+        """
+        self.problem_type = handle_problem_types(problem_type)
+        if self.problem_type not in [ProblemTypes.MULTICLASS, ProblemTypes.TIME_SERIES_MULTICLASS]:
```
angela97lin, line 29: I think `is_multiclass` was a new util method added by @dsherry (thanks Dylan!) a while ago that could help with this c: 

It's hard to go through and update everything to use that but might as well try for new additions?

angela97lin (approved):
Loooooks great! Added a few comments mostly for cleanup but otherwise, 🚢 !!

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @staticmethod
+    def sparsity_score(col, count_threshold=10):
+        """This function calculates a sparsity score for the given value counts by calculating the percentage of
+        values that fail to exceed the count_threshold.
+        Arguments:
+            col (pd.Series): Feature values.
+            count_threshold (int): The number of instances below which a value is considered sparse.
+        Returns:
+            (float): Sparsity score, or the percentage of the unique values that fail to exceed count_threshold.
```
bchen1116, line 85: note the default

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+                                                                                        self.problem_type),
+                                                      data_check_name=self.name,
+                                                      message_code=DataCheckMessageCode.TOO_SPARSE,
+                                                      details={"column": col_name, "sparsity_score": res.loc[col_name]}).to_dict()
+                                     for col_name in too_sparse_cols])
+        return messages
+
+    @staticmethod
+    def sparsity_score(col, count_threshold=10):
+        """This function calculates a sparsity score for the given value counts by calculating the percentage of
```
bchen1116, line 79: Description doesn't seem to match what it does. In the test case where the values are all the same, ie `[1, 1, 1, 1, 1, 1, 1, 1]`, we get a `sparsity_score` of 1, but that doesn't mean that 100% of values failed to exceed the count threshold, right? I think it's reversed.

evalml/data_checks/sparsity_data_check.py
```diff
@@ -0,0 +1,90 @@
...
+    _convert_woodwork_types_wrapper
+)
+
+warning_too_unique = "Input columns ({}) for {} problem type are too sparse."
+
+
+class SparsityDataCheck(DataCheck):
+    """Checks if there are any columns with sparsely populated values in the input."""
+
+    def __init__(self, problem_type, threshold, unique_count_threshold=10):
```
bchen1116, line 18: Just a thought, but would it be useful to have an arg that could take in a float to represent the minimum percent of the total data for sparsity? Ie, if we pass it 0.1, it could mean that all values have to appear 10% or more in the column for it to be considered non-sparse. Thinking in terms of really large versus smaller datasets, having a percentage might be more useful than just an int. But this seems fine for now!

Some thoughts that I had.

bchen1116 (approved):
LGTM! Left a few comments on things to fix

evalml/data_checks/sparsity_data_check.py
chukarsten: You're write, the doc string is just wrong.  Sorry, I feel like I've been going back and forth on this "score."  Let me fix it.

evalml/data_checks/sparsity_data_check.py
chukarsten: I agree, and I think this is a good idea.  I will file an issue for it as I think it makes way more sense than defining a discrete value count threshold.  But for this PR, I just wanted to directly mimic PS' implementation.

chukarsten: Issue #1801 filed so that we can come back to this.

PR #1798, opened 2021-02-08
Rename random state to random seed

### LINKED ISSUE #1737: Rename `random_state` to `random_seed` in EvalML
PR [1727](https://github.com/alteryx/evalml/pull/1727) removed support for `np.random.RandomState` and set everything to use an int seed instead. As stated [here](https://github.com/alteryx/evalml/pull/1727#pullrequestreview-576502964), it might be confusing if we continue to internally call the seed the `random_state`, so changing this to instead be called `random_seed` and changing the args to reflect this would maintain transparency with users.

freddyaboulton:
### Pull Request Description
Fixes #1737  by adding `random_seed` as an argument and keeping `random_state` but setting it to None. A warning will show if `random_state` is used.

-----


evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -19,7 +19,7 @@ def __init__(self,
                  max_iterations=None,
                  tuner_class=None,
                  text_columns=None,
-                 random_state=0,
+                 random_seed=0,
```
freddyaboulton, line 22: I'm replacing random_state with random_seed here, rather than adding it as an argument, because this is not a user-facing class.

evalml/pipelines/components/transformers/dimensionality_reduction/pca.py
```diff
@@ -28,13 +29,14 @@ def __init__(self, variance=0.95, n_components=None, random_state=0, **kwargs):
         parameters = {"variance": variance,
                       "n_components": n_components}
         parameters.update(kwargs)
+        random_seed = deprecate_arg("random_state", "random_seed", random_state, random_seed)
         if n_components:
-            pca = SkPCA(n_components=n_components, **kwargs)
+            pca = SkPCA(n_components=n_components, random_state=random_seed, **kwargs)
```
freddyaboulton, line 34: Not sure why we were not passing random_state to PCA before.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1800,6 +1800,7 @@ def total_pipelines(automl, num_batches, batch_size):
     automl.search()
     assert automl._pipelines_per_batch == 2
     assert automl._automl_algorithm.pipelines_per_batch == 2
+    assert automl._automl_algorithm.pipelines_per_batch == 2
```
bchen1116, line 1803: repeated line?

evalml/utils/gen_utils.py
```diff
@@ -511,3 +511,23 @@ def save_plot(fig, filepath=None, format='png', interactive=False, return_filepa
 
     if return_filepath:
         return filepath
+
+
+def deprecate_arg(old_arg, new_arg, old_value, new_value):
```
bchen1116, line 516: Should we add a test for this helper since it's public?

docs/source/release_notes.rst
```diff
@@ -8,6 +8,7 @@ Release Notes
         * Sped up permutation importance for some pipelines :pr:`1762`
     * Fixes
     * Changes
+        * Added ``random_seed`` as an argument to our automl/pipeline/component api. Using ``random_state`` will raise a warning :pr:`1798`
```
bchen1116, line 11: We probably can't add this to `breaking_changes`, but is there another category similar that we can put it in to warn users? Just wondering since it'll soon be a breaking change once we officially deprecate.

bchen1116 (approved):
Nice job fixing everything and raising the warnings! LGTM

evalml/utils/gen_utils.py
angela97lin, line 516: Nice, this will be useful for any future dep patterns we follow 🥳

angela97lin (approved):
Left a few comments but nice and thorough. Thank you, LGTM!! 🚢

evalml/pipelines/components/estimators/classifiers/catboost_classifier.py
```diff
@@ -36,8 +36,8 @@ class CatBoostClassifier(Estimator):
     SEED_MAX = SEED_BOUNDS.max_bound
```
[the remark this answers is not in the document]
freddyaboulton: I think we can get rid of them actually!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -65,7 +65,7 @@ def next_batch(self):
 
         next_batch = []
         if self._batch_number == 0:
-            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_state=self.random_state)
+            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_seed=self.random_seed)
```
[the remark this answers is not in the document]
freddyaboulton: `self.random_seed == self._random_state`. We get `self.random_seed` from the parent class. `self._random_state` is used to sample params in the first batch if the user provides pipelines with frozen hyperparams. I'll get rid of `self._random_state` since we don't need it anymore!

evalml/utils/gen_utils.py
freddyaboulton: Done!

docs/source/release_notes.rst
freddyaboulton: I think this is fine for now? Here we're telling users we'll raise a warning and the warning itself will tell them about the impending deprecation.

PR #1802, opened 2021-02-09
What do you think?

chukarsten:
Wanted to try to see if I could move that code chunk into the parent class.  What do you think?

### DIFF
--- a/evalml/pipelines/components/estimators/classifiers/catboost_classifier.py
+++ b/evalml/pipelines/components/estimators/classifiers/catboost_classifier.py
@@ -62,7 +62,7 @@ class CatBoostClassifier(Estimator):
     def fit(self, X, y=None):
         X = _convert_to_woodwork_structure(X)
         cat_cols = list(X.select('category').columns)
-        X, y = super()._foo(X, y)
+        X, y = super()._manage_woodwork(X, y)
 
         self.input_feature_names = list(X.columns)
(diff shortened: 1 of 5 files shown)


evalml/pipelines/components/estimators/classifiers/catboost_classifier.py
```diff
@@ -62,11 +62,9 @@ def __init__(self, n_estimators=10, eta=0.03, max_depth=6, bootstrap_type=None,
     def fit(self, X, y=None):
         X = _convert_to_woodwork_structure(X)
         cat_cols = list(X.select('category').columns)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
-        self.input_feature_names = list(X.columns)
-        y = _convert_to_woodwork_structure(y)
-        y = _convert_woodwork_types_wrapper(y.to_series())
+        X, y = super()._manage_woodwork(X, y)
```
angela97lin, line 65: I had a thought after the post-standup discussion from @rpeck's comment:

How about in the catboost estimators we define `_manage_woodwork()` which does the call to get `cat_cols` and then we can just call X, y = self._manage_woodwork(X, y) without the above two lines? Saves on the duplication and we still have this pattern of abstracting away the WW code :o

angela97lin (approved):
Thanks again for this, I think this is great 😁  Feel free to merge whenever!

PR #1809, opened 2021-02-09
Updating percent better than baseline computation

### LINKED ISSUE #1449: For some objectives where baseline was 0, "pct better than baseline" is nan
```javascript
{'F1': nan,
 'MCC Binary': nan,
 'Log Loss Binary': 93.29789549298991,
 'AUC': 58.36492736629537,
 'Precision': nan,
 'Balanced Accuracy Binary': 63.46659876071641,
 'Accuracy Binary': 12.876088314169193}
```

I've created a Jupyter notebook that reproduces this problem in `evalml`, and attached it and the associated datafile to a thread in Slack.

freddyaboulton:
### Pull Request Description
Fixes #1449

* For objectives bounded strictly in the range [0, 1], the percent difference is the absolute difference (not normalized by the baseline)
* For other objectives, division by 0 is now inf rather than nan

-----

### DIFF
--- a/evalml/objectives/cost_benefit_matrix.py
+++ b/evalml/objectives/cost_benefit_matrix.py
@@ -14,6 +14,8 @@ class CostBenefitMatrix(BinaryClassificationObjective):
     greater_is_better = True
     score_needs_proba = False
     perfect_score = np.inf
+    # Range (-Inf, Inf)
+    is_bounded_like_percentage = False
 
     def __init__(self, true_positive, true_negative, false_positive, false_negative):
         """Create instance of CostBenefitMatrix.
--- a/evalml/objectives/cost_benefit_matrix.py
+++ b/evalml/objectives/cost_benefit_matrix.py
@@ -14,8 +14,7 @@ class CostBenefitMatrix(BinaryClassificationObjective):
     greater_is_better = True
     score_needs_proba = False
     perfect_score = np.inf
-    # Range (-Inf, Inf)
-    is_bounded_like_percentage = False
+    is_bounded_like_percentage = False  # Range (-Inf, Inf)
 
     def __init__(self, true_positive, true_negative, false_positive, false_negative):
         """Create instance of CostBenefitMatrix.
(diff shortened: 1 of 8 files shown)


evalml/objectives/cost_benefit_matrix.py
```diff
@@ -14,6 +14,8 @@ class CostBenefitMatrix(BinaryClassificationObjective):
     greater_is_better = True
     score_needs_proba = False
     perfect_score = np.inf
+    # Range (-Inf, Inf)
```
bchen1116, line 17: nit-pick but I thought this was an accidental comment rather than on purpose. Maybe we can move it to be next to `is_bounded_like_percentage` rather than being above?

evalml/objectives/lead_scoring.py
```diff
@@ -9,6 +9,8 @@ class LeadScoring(BinaryClassificationObjective):
     greater_is_better = True
     score_needs_proba = False
     perfect_score = math.inf
+    # Range (-Inf, Inf)
```
bchen1116, line 12: same comment

bchen1116 (approved):
LGTM!

evalml/objectives/standard_metrics.py
```diff
@@ -243,6 +263,8 @@ class LogLossBinary(BinaryClassificationObjective):
     greater_is_better = False
     score_needs_proba = True
     perfect_score = 0.0
+    # Range [0, Inf)
```
angela97lin, line 266: Thank you for including these range comments. Super helpful 😁

angela97lin (approved):
Nice work, LGTM! 🥳 

Just out of curiosity, what did we decide for 'edge cases' such as the correlation matrix? I guess it's not really what we currently consider as an objective... but was wondering if we decided on anything :3

evalml/objectives/cost_benefit_matrix.py
chukarsten: Wouldn't this go next to `perfect_score`?

chukarsten (approved):
Looks good to me.

freddyaboulton:
@angela97lin Correlation is just like MCC in the sense that it's bounded [-1, 1]. I think it's best to not treat like a percentage in that case. If baseline gets a negative score and our pipeline gets a positive score, then I'd want the percent better to be above 100% which isn't always the case if we treat it like a percentage.

PR #1813, opened 2021-02-09
Update visualize_decision_tree to include feature names

### LINKED ISSUE #1718: Update visualize_decision_tree to track feature names
Currently `visualize_decision_tree` in `model_understanding` sets column names in its cells as the format `X[14]` or `X[4]`. ~~We should be able to pass in `feature_names` as a parameter to `visualize_decision_tree` and then to `export_graphviz` so that the visualization reflects the feature names.~~

Hijacking this issue a bit, but I think we shouldn't need to pass in another parameter for this, since #1757 addressed estimators keeping track of feature names they were trained on. We should just be able to grab that information now, without the help of the user :)

VisualizeDT

angela97lin:
Closes #1718. Also removes `feature_names` as a parameter from `decision_tree_data_from_estimator`, since we can extract the feature names used to fit from the estimator now.

Not sure what the best way to add testing to our codebase to ensure the output png has the right name, but here's a code snippet and the resulting image 😁 

```
from sklearn import datasets
from evalml.pipelines.components import DecisionTreeRegressor, DecisionTreeClassifier
from evalml.model_understanding.graphs import (
    visualize_decision_tree
)

def X_y():
    X, y = datasets.make_classification(n_samples=100, n_features=20,
                                        n_informative=2, n_redundant=2, random_state=0)

    return X, y


X_b, y_b = X_y()
X_b = pd.DataFrame(X_b, columns=[f'Testing_{col}' for col in range(len(X_b[0]))])
dt = DecisionTreeClassifier()
dt.fit(X_b, y_b)
visualize_decision_tree(dt, filepath='test.png')
```

test


evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -1213,21 +1213,10 @@ def test_decision_tree_data_from_estimator_wrong_type(logit_estimator):
...
-    features_array = tuple([f'Testing_{col_}' for col_ in range(est_class._component_obj.n_features_)])
-    formatted_ = decision_tree_data_from_estimator(est_class, feature_names=features_array)
-    assert isinstance(formatted_, OrderedDict)
-
-
 def test_decision_tree_data_from_estimator(fitted_tree_estimators):
     est_class, est_reg = fitted_tree_estimators
 
-    formatted_ = decision_tree_data_from_estimator(est_reg, feature_names=[f'Testing_{col_}' for col_ in range(est_reg._component_obj.n_features_)])
+    formatted_ = decision_tree_data_from_estimator(est_reg)
```
angela97lin, line 1219: Removing feature names parameter. Since this test already checks that the output has the feature names but now we're just grabbing them from the estimator rather than setting explicitly, no other code changes needed.

jeremyliweishih (approved):
LGTM. Only comment I have is maybe we can make the feature names stand out a bit (bold or italicize it?).

angela97lin:
@jeremyliweishih That's a great suggestion! With our current implementation though, I'm not sure there's an API available to easily do this though so we might need to punt on that: https://scikit-learn.org/stable/modules/generated/sklearn.tree.export_graphviz.html#sklearn.tree.export_graphviz

bchen1116 (approved):
Looks good!

Issue #1729, opened 2021-01-22
Make it easy to score automl search pipelines on holdout


dsherry:
I'd like us to consider the following use-cases:
* Train all pipelines on training data and return trained pipelines
* Train all pipelines on training data, score on holdout and return the scores
* Both of the above, with the ability to specify a subset instead of training on all pipelines

Let's explore some APIs for the above.

To close this issue: design doc.

freddyaboulton:
@dsherry @chukarsten I've prepared a design doc [here](https://alteryx.quip.com/ySavAbaTT3zg/Updating-AutoML-to-make-it-easy-to-train-pipelines-on-entire-data)!

PR #1814, opened 2021-02-10
Separate CV for Stacked Ensembler in AutoMLSearch

### LINKED ISSUE #1746: AutoML: use separate CV split for ensembling
~~#1732 updates our stacked ensembling pipeline to use the same data splitter that is used in `AutoML`. However, @rpeck noted that we may not want to do this. We continued with #1732 because we believed that it was still an improvement on our current approach (scikit-learn default).~~

This issue tracks long-term updates we may want to make to our data splitter for stacking in AutoML. 

---
Update: while continuing to update #1732, we ran into a conundrum with the interaction between stacking and AutoML that made us revisit whether it really was a good idea to use the same data split for the stacking ensemble as we use in AutoML. We decided no, and that as Raymond had pointed out, we probably want to use a separate CV for our ensemble. (@dsherry  also mentioned a good nugget of info that using ensembling allows the model to put less excessive importance on the more complex models, so CV probably helps with that--please correct me if I've paraphrased incorrectly 😂 ). 

Rather than continuing that work then, we should use this issue to discuss updating AutoML for stacking: **Specifically, we should create a separate split for CV for the stacked ensembling.** This would be similar to what we currently have in place for binary-threshold tuning.

bchen1116:
fix #1746 

Perf tests results are [here](https://alteryx.quip.com/RcJxA0VXrfc2/Stacked-Ensembling-CV-Folds)

As @chukarsten brings up [here](https://github.com/alteryx/evalml/pull/1814#discussion_r584178786), should we add an `ensembling_split` arg to AutoMLSearch so that users can define what % of the data they want for the ensembling split? Should we test other split sizes?


evalml/automl/engine/sequential_engine.py
```diff
@@ -20,7 +20,8 @@ def evaluate_batch(self, pipelines):
         while self._should_continue_callback() and index < len(pipelines):
             pipeline = pipelines[index]
             self._pre_evaluation_callback(pipeline)
-            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
+            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train,
```
chukarsten, line 23: I'm not convinced that ensembling related logic falls within the Engine's area of responsibility.  I feel like we should keep the engine as responsible for training, fitting, scoring pipelines that are fed to it via train_and_score_pipeline() and freddy's upcoming fit_batch() and evaluate_batch() functionality.  I'm not sure - I'm super tired.  it just seems weird passing X_ensemble and y_ensemble to train_and_score. @dsherry, what do you think?

chukarsten (changes requested):
Hey Bryan, putting changes requested just so we can have a better discussion when I'm less tired, haha.  This looks good, I'd just like to think together about what the engine should be doing vs. automl.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2251,6 +2257,65 @@ def test_automl_pipeline_random_seed(mock_fit, mock_score, random_seed, X_y_mult
...
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', random_state=0, n_jobs=1, max_batches=19, ensembling=ensembling, train_best_pipeline=False, optimize_thresholds=False)
+    automl.search()
+    if ensembling:
+        assert automl.ensembling
+        # check that the X_train data is all used for the length
+        assert len(X_train) == (len(mock_fit.call_args_list[-2][0][0]) + len(mock_score.call_args_list[-2][0][0]))
+        # last call will be the stacking ensembler
+        assert len(X_ensemble) == (len(mock_fit.call_args_list[-1][0][0]) + len(mock_score.call_args_list[-1][0][0]))
+    else:
+        # verify that there is use nor creation of ensembling CV data
```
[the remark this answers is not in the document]
bchen1116: typo 😅

evalml/automl/automl_search.py
```diff
@@ -318,6 +321,10 @@ def __init__(self,
                                        num_ensemble_batches)
             else:
                 self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self.max_batches - 1))
+        if run_ensembling:
+            self.X_train, self.X_ensemble, self.y_train, self.y_ensemble = split_data(self.X_train, self.y_train, problem_type=self.problem_type, test_size=0.20)
```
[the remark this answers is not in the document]
bchen1116: @chukarsten right, I see your point. Through my discussions with @dsherry, we decided that a 0.8/0.2 split seemed fine for separating the ensembling, but I do see your concern about making it more adjustable/customizable. I don't see a problem with adding it as a (private for now) arg to AutoMLSearch, in case a user wants to make it finer? 

What do you think, @dsherry?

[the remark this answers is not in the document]
chukarsten: If you guys chatted about it, let's just file an issue to handle it later, then!

evalml/automl/engine/sequential_engine.py
bchen1116: @chukarsten, do you mean you'd prefer a check like:

```python
# in sequential_engine.py
# while loop over pipeines
if pipeline.model_family == ModelFamily.Ensemble:
     evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_ensemble, self.y_ensemble)
else:
     evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
```

versus what I currently have? Or do something different altogether?

evalml/automl/automl_search.py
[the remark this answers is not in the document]
bchen1116: I'll leave this up for discussion for now, in case other people have inputs.

evalml/automl/engine/sequential_engine.py
chukarsten: Yea, I think I like that. I think at least keeping it out of train and score pipelines will be helpful.

chukarsten (approved):
image

evalml/automl/automl_search.py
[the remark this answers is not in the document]
dsherry: Ah yes, let's please make this configurable! I think its fine to default it to 20%.

evalml/automl/engine/sequential_engine.py
freddyaboulton: I like that too @bchen1116 !

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2251,6 +2257,65 @@ def test_automl_pipeline_random_seed(mock_fit, mock_score, random_seed, X_y_mult
...
+    X_train, X_ensemble, y_train, y_ensemble = split_data(X, y, problem_type='binary', test_size=0.2)
+    mock_split_data.return_value = (X_train, X_ensemble, y_train, y_ensemble)
+    # don't train the best pipeline since we check usage of the ensembling CV through the .fit mock
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', random_state=0, n_jobs=1, max_batches=19, ensembling=ensembling, train_best_pipeline=False, optimize_thresholds=False)
+    automl.search()
+    if ensembling:
+        assert automl.ensembling
+        # check that the X_train data is all used for the length
+        assert len(X_train) == (len(mock_fit.call_args_list[-2][0][0]) + len(mock_score.call_args_list[-2][0][0]))
+        # last call will be the stacking ensembler
```
freddyaboulton, line 2275: I don't think this will be true when we add another estimator to AutoMLSearch, e.g. KnnClassifier.

Maybe we should make max_batches depend on the number of allowed estimators for binary problems.

freddyaboulton (changes requested):
@bchen1116 I think this is great! I think the only thing blocking merge is that the concatenation of X_train and X_ensemble should preserve logical types passed in by the user.

evalml/automl/engine/sequential_engine.py
```diff
@@ -20,7 +21,10 @@ def evaluate_batch(self, pipelines):
         while self._should_continue_callback() and index < len(pipelines):
             pipeline = pipelines[index]
             self._pre_evaluation_callback(pipeline)
-            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
+            if pipeline.model_family == ModelFamily.ENSEMBLE:
```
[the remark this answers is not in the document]
chukarsten: Sure, that works for me.  Would have to see what that looks like with the call to super().  Do you mean putting this whole if/else branch in super and swapping the X/y?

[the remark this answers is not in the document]
freddyaboulton: It's trickier than I thought since the logic depends on a single pipeline but the function takes in a list of pipelines. Best to do it later but it would be nice if we didn't have to replicate this along all engines.

evalml/automl/automl_search.py
```diff
@@ -575,7 +582,15 @@ def _find_best_pipeline(self):
             if self._train_best_pipeline:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
-                X_train, y_train = self.X_train, self.y_train
+                if self._best_pipeline.model_family == ModelFamily.ENSEMBLE:
+                    X_train, y_train = self.X_ensemble, self.y_ensemble
+                else:
+                    if self.X_ensemble:
+                        X_train = self.X_train.to_dataframe().append(self.X_ensemble.to_dataframe(), ignore_index=True)
```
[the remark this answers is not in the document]
bchen1116: oh good catch! I can fix this

evalml/automl/engine/sequential_engine.py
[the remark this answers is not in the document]
bchen1116: @chukarsten @freddyaboulton what if we did something like
```python
# in engine_base.py
# in engine class
def handle_train_and_score_pipeline(pipeline):
    if pipeline.model_family == ModelFamily.ENSEMBLE:
         evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_ensemble, self.y_ensemble)
    else:
         evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
    return evaluation_result
```

Then in `SequentialEngine`, you can simply call `super().handle_train_and_score_pipeline(pipeline)`? 

This way, it'll be in the base Engine class, although we wouldn't be adding it to `EngineBase.evaluate_batch`, and it might just be an extra method call that we don't need.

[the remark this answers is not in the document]
freddyaboulton: @bchen1116 I like that! I think we'll need to do re-use the logic for #1729 so it may be best to return the data to use (`X_ensemble` vs `X_train`) rather than the evaluation result.

I think it's best for me to handle this in the PR for #1729. I'm convinced this is scope creep now cause the best way to do this depends on requirements not relating to this original issue. 

Sorry for the misdirection @bchen1116!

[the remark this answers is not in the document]
bchen1116: Ah, yep that makes sense! Thanks for pointing that out!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2251,6 +2257,79 @@ def test_automl_pipeline_random_seed(mock_fit, mock_score, random_seed, X_y_mult
...
+    mock_split_data.return_value = (X_train, X_ensemble, y_train, y_ensemble)
+    # don't train the best pipeline since we check usage of the ensembling CV through the .fit mock
+    ensemble_pipelines = len(get_estimators("binary")) + 2
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', random_state=0, n_jobs=1, max_batches=ensemble_pipelines, ensembling=ensembling,
+                          train_best_pipeline=False, optimize_thresholds=False, _ensembling_split_size=ensemble_split_size)
+    automl.search()
+    if ensembling:
+        assert automl.ensembling
+        # check that the X_train data is all used for the length
+        assert len(X_train) == (len(mock_fit.call_args_list[-2][0][0]) + len(mock_score.call_args_list[-2][0][0]))
```
angela97lin, line 2277: Trying to follow along, but why are we checking this against the length of the mock_fit and length of the mock_score args? 🤔

angela97lin (approved):
This looks good! Just had a comment for my own understanding hehe

evalml/tests/automl_tests/test_automl.py
bchen1116: This checks that the second-to-last pipeline (which isn't ensembling) uses the full X_train dataset between the fit and score calls. The `mock_fit.call_args_list[-2][0][0]` gets the X input for the fit call, and the same for `mock_score`. I just did this to make sure that we're using the training data we expect as opposed to the ensembling data. 

Hopefully that clarifies!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2251,6 +2257,79 @@ def test_automl_pipeline_random_seed(mock_fit, mock_score, random_seed, X_y_mult
...
+@pytest.mark.parametrize("ensemble_split_size", [0.1, 0.2])
+@patch('evalml.automl.automl_search.split_data')
+@patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.3})
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_automl_ensembling_training(mock_fit, mock_score, mock_split_data, ensemble_split_size, ensembling, X_y_binary):
+    X, y = X_y_binary
+    X_train, X_ensemble, y_train, y_ensemble = split_data(X, y, problem_type='binary', test_size=ensemble_split_size)
+    mock_split_data.return_value = (X_train, X_ensemble, y_train, y_ensemble)
+    # don't train the best pipeline since we check usage of the ensembling CV through the .fit mock
+    ensemble_pipelines = len(get_estimators("binary")) + 2
```
freddyaboulton, line 2270: Nice 👍

evalml/automl/automl_search.py
```diff
@@ -575,7 +588,15 @@ def _find_best_pipeline(self):
             if self._train_best_pipeline:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
-                X_train, y_train = self.X_train, self.y_train
+                if self._best_pipeline.model_family == ModelFamily.ENSEMBLE:
+                    X_train, y_train = self.X_ensemble, self.y_ensemble
+                else:
+                    if self.X_ensemble:
+                        X_train = _convert_woodwork_types_wrapper(self.X_train.to_dataframe()).append(_convert_woodwork_types_wrapper(self.X_ensemble.to_dataframe()), ignore_index=True)
```
freddyaboulton, line 595: @bchen1116 I think we need something like this:

In AutoMLSearch init
```python
self.X_train = infer_feature_types(X_train)
# This part I'm not so sure about
self.X_feature_types = self.X_train.logical_types
```

In _find_best_pipeline:
```python
X_train = infer_feature_types(self.X_train.to_dataframe().append(self.X_ensemble.to_dataframe()), self.X_feature_types)
```

The issue is that we want to preserve the types provided by the user in case they override the default types. 

Maybe @angela97lin knows the best way/better way to concatenate two datatables while preserving types?

angela97lin: I think what you've proposed looks good here @freddyaboulton! Unfortunately, I'm not sure there's a good way to concatenate two DataTables without going to the underlying DataFrames :'(

bchen1116: Got it! I'll put that in, thanks for the info!

freddyaboulton (approved):
@bchen1116 This looks good to me!!! 🎉  🥳

evalml/automl/automl_search.py
```diff
@@ -575,7 +589,15 @@ def _find_best_pipeline(self):
             if self._train_best_pipeline:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
-                X_train, y_train = self.X_train, self.y_train
+                if self._best_pipeline.model_family == ModelFamily.ENSEMBLE:
+                    X_train, y_train = self.X_ensemble, self.y_ensemble
+                else:
+                    if self.X_ensemble:
+                        X_train = infer_feature_types(self.X_train.to_dataframe().append(self.X_ensemble.to_dataframe(), ignore_index=True), self.X_feature_types)
+                        y_train = infer_feature_types(self.y_train.to_series().append(self.y_ensemble.to_series(), ignore_index=True))
```
dsherry, line 597: Hmm. Converting to pandas and then calling `infer_feature_types` again means we lose of the custom user-defined types. @angela97lin had to do a bunch of plumbing to avoid this during pipeline evaluation!

An example of this (which would be a great test to add to this PR): user overrides woodwork to set a text feature as categorical. We mock out the ensemble fit method to ensure that the categorical type makes its way to the metalearner instead of the default inferred text type.

It also feels strange to split up the data in the `AutoMLSearch` constructor, and then merge it again here.

The most compelling option I could think of right now is: in `AutoMLSearch` at least, pass around indices instead of dataframes/datatables, until the moment we actually need a sub-datatable in order to proceed; then at that point, slice as needed.

We could also keep references to both a) the complete data and b) the splitted data, and use whichever we need when we need it. The difficulty there is that effectively doubles our memory consumption.

Let's discuss.

dsherry (changes requested):
@bchen1116 this is close! The only blocking comment I have is about the way we're doing the data splitting. Let's find a way to preserve user-overriden woodwork types. I left an example and a couple options in the comments, but we can discuss.

evalml/automl/automl_search.py
bchen1116: @dsherry I see what you mean. For the custom user-defined types, wouldn't passing in `self.X_train.logical_types` to `infer_feature_types` solve that problem? 

As for the issue of splitting and merging unnecessarily, I think the best idea would then be to keep a reference to the indices of the ensembling data rather than the full split ensembling data. This would also help with memory consumption but should allow us to keep the same functionality as everything before. What do you think?

dsherry: @bchen1116 oh maybe so! @angela97lin does that line up with your understanding? I suggest you check with her on this.

> I think the best idea would then be to keep a reference to the indices of the ensembling data rather than the full split ensembling data. 

If I'm following you correctly here, yep, I'm suggesting we look into doing the splitting on indices first instead of on datatables.

If using indices is not required to fix the type problem, we could file it as a separate issue, but let's be careful about the code we use to do the concatenation here. In fact I'd suggest we use `pd.concat` instead of `append` here.

angela97lin: @bchen1116 @dsherry 
RE using `infer_feature_types`: If the *only* thing that's happening is we're taking a subset of rows, then passing `self.X_train.logical_types` to `infer_feature_types` should suffice since nothing is happening to the `logical_types`.

For the transformers, we converted to pandas, and then potentially transformed the data before we converted back to Woodwork, so extra plumbing was necessary (and stored in `_retain_custom_types_and_initalize_woodwork`).

Of course, I agree that testing would be important here to make sure we're not missing anything :)


I do agree that it feels pretty strange to split up the data in the AutoMLSearch constructor, and then merge it again here though...

bchen1116: Just updated the implementation to track the indices instead, so we no longer need to do any `concat/append` logic when training the best pipeline.

evalml/automl/automl_search.py
```diff
@@ -318,6 +317,19 @@ def __init__(self,
                                        num_ensemble_batches)
             else:
                 self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self.max_batches - 1))
+        if run_ensembling:
+            if not (0 < _ensembling_split_size < 1):
+                raise ValueError(f"Ensembling split size must be between 0 and 1 exclusive, received {_ensembling_split_size}")
+            _, self.ensembling_indices = next(TrainingValidationSplit(test_size=_ensembling_split_size, shuffle=True, stratify=self.y_train.to_series(), random_state=self.random_seed).split(self.X_train.to_dataframe(), self.y_train.to_series()))
```
dsherry, line 323: @bchen1116 yep, computing the indices here and passing them to the engine looks good to me!

Could you use our `split_data` helper here instead? That will do a standard TV split for regression data, a stratified split for classification data and an ordered TV split for timeseries data. We should use that any time we want to do a two-part split. Otherwise we could be giving the ensembler a biased or unfair split!

Also just as a stylistic aside, its definitely slick to do this in one line, but IMO its preferable to split it into multiple lines, so others can quickly and easily understand what's up. We don't have a max line length enforced in our linter right now, but, yeah! :)

evalml/automl/automl_search.py
```diff
@@ -170,6 +171,9 @@ def __init__(self,
 
             train_best_pipeline (boolean): Whether or not to train the best pipeline before returning it. Defaults to True
 
+            _ensembling_split_size (float): The amount of the training data we'll set aside for training ensemble metalearners. Only used when ensembling is True.
```
dsherry, line 174: Thanks

evalml/automl/automl_search.py
```diff
@@ -575,7 +587,11 @@ def _find_best_pipeline(self):
             if self._train_best_pipeline:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
-                X_train, y_train = self.X_train, self.y_train
+                if self._best_pipeline.model_family == ModelFamily.ENSEMBLE:
+                    X_train, y_train = self.X_train.iloc[self.ensembling_indices], self.y_train.iloc[self.ensembling_indices]
+                else:
+                    X_train = self.X_train
+                    y_train = self.y_train
```
dsherry, line 594: This looks great

evalml/automl/engine/sequential_engine.py
```diff
@@ -20,7 +21,15 @@ def evaluate_batch(self, pipelines):
...
-            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
+            X, y = self.X_train, self.y_train
+            if pipeline.model_family == ModelFamily.ENSEMBLE:
+                X, y = self.X_train.iloc[self.ensembling_indices], self.y_train.iloc[self.ensembling_indices]
+            else:
+                if self.ensembling_indices is not None:
+                    training_indices = [i for i in range(len(self.X_train)) if i not in self.ensembling_indices]
+                    X = self.X_train.iloc[training_indices]
+                    y = self.y_train.iloc[training_indices]
+            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, X, y)
```
dsherry, line 32: 👏

evalml/automl/engine/sequential_engine.py
```diff
@@ -20,7 +21,15 @@ def evaluate_batch(self, pipelines):
         while self._should_continue_callback() and index < len(pipelines):
             pipeline = pipelines[index]
             self._pre_evaluation_callback(pipeline)
-            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
+            X, y = self.X_train, self.y_train
+            if pipeline.model_family == ModelFamily.ENSEMBLE:
+                X, y = self.X_train.iloc[self.ensembling_indices], self.y_train.iloc[self.ensembling_indices]
+            else:
+                if self.ensembling_indices is not None:
```
dsherry, line 28: Total nit-pick: flatten by using `elif self.ensembling_indices is not None`

evalml/tests/automl_tests/test_sequential_engine.py
```diff
@@ -13,6 +18,53 @@ def test_evaluate_no_data():
         engine.evaluate_batch([])
 
 
+def test_add_ensemble_data():
```
dsherry, line 21: @bchen1116 I like your division of test coverage here! The first test checks the accessors, and the second checks the functionality using mocking. Very nice.

evalml/tests/automl_tests/test_sequential_engine.py
```diff
@@ -13,6 +18,53 @@ def test_evaluate_no_data():
...
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+def test_ensemble_data(mock_fit, mock_score, dummy_binary_pipeline_class, stackable_classifiers):
+    X = pd.DataFrame({"a": [i for i in range(100)]})
+    y = pd.Series([i % 2 for i in range(100)])
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', max_batches=19, ensembling=True, _ensembling_split_size=0.25)
+    mock_should_continue_callback = MagicMock(return_value=True)
+    mock_pre_evaluation_callback = MagicMock()
+    mock_post_evaluation_callback = MagicMock()
+    for training_indices, ensembling_indices in TrainingValidationSplit(test_size=0.25, shuffle=True, stratify=y, random_state=0).split(X, y):
```
dsherry, line 48: If I follow right we'll need to update this to use `split_data` too

dsherry (approved):
@bchen1116 great!!! I left one blocking comment: we should use `split_data` to compute the ensembling split in the `AutoMLSearch` constructor. Otherwise we won't be using the correct split strategy for classification data, timeseries data etc.

PR #1816, opened 2021-02-10
Augment InvalidTargetDataCheck to check for mismatched indices in target and features

### LINKED ISSUE #1726: Augment InvalidTargetDataCheck to check for mismatched indices in target and training data
This fell out of https://github.com/alteryx/evalml/issues/1723. My guess is that this would fit best in `InvalidTargetDataCheck` but open to suggestions.

We've seen a few cases where indices cause bugs (imputer, encoder, etc). It could be a good idea to warn users if their data has mismatched indices, since that could leave to unexpected behavior. For example, pandas could try to backfill some indices and introduce NaN and cause errors while modelling.

@freddyaboulton FYI 😀

angela97lin:
Closes #1726.

I chose to check if the values of the indices match, so we don't run into weird behavior with RangeIndex vs Index, but would love to discuss if that's too flexible or if we should check for explicit matches!

After discussion w/ @freddyaboulton @chukarsten and @dsherry, separated out case where different lengths cause index mismatch vs different values. Also separated out the case where the indices have the same values (set equivalent) but are of a different order. I thought this could be a good idea because if we just printed the set difference, it would be an empty set and confuse users as to why there are mismatched indices.

Choosing not to change this behavior as pointed out by @dsherry [here](https://github.com/alteryx/evalml/issues/1726#issuecomment-772647856) and filed https://github.com/alteryx/evalml/issues/1820 to track that update instead.

TODO:

- [x] Make sure tests are passing for mismatched indices, not-in-order-but-same indices, different length indices
- [x] Separate out different lengths case from different indices case
- [x] Only list X number of mismatched indices, rather than everything


bchen1116 (approved):
LGTM! I think checking for list equality is fine for this.

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -142,4 +142,13 @@ def validate(self, X, y):
...
 
+        if X is not None:
+            X = _convert_to_woodwork_structure(X)
+            X_index = list(X.to_dataframe().index)
+            y_index = list(y_df.index)
+            if X_index != y_index:
+                messages["warnings"].append(DataCheckWarning(message="Input target and features have mismatched indices",
+                                                             data_check_name=self.name,
+                                                             message_code=DataCheckMessageCode.MISMATCHED_INDICES,
+                                                             details={"feature_index": X_index, "target_index": y_index}).to_dict())
```
freddyaboulton, line 153: I don't think it's a good idea to return the entire indices of X and y because they can be really big. Maybe we just return a sample of the diff?

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -142,4 +142,13 @@ def validate(self, X, y):
                                                      message_code=DataCheckMessageCode.TARGET_INCOMPATIBLE_OBJECTIVE,
                                                      details=details).to_dict())
 
+        if X is not None:
+            X = _convert_to_woodwork_structure(X)
+            X_index = list(X.to_dataframe().index)
+            y_index = list(y_df.index)
+            if X_index != y_index:
```
freddyaboulton, line 149: I think this makes sense! pandas treats `RangeIndex` and `Int64Index` the same as long as they have the same values when it comes to doing operations on different series/dataframes so no NaNs will be imputed. So I think list comparison is a good idea.

freddyaboulton (approved):
@angela97lin Looks good! My only comment is that outputting the entire indices may overload users if their data is big.

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -385,4 +411,27 @@ def test_invalid_target_data_check_multiclass_problem_almostcontinuous_data():
                                                                              "errors": []}
 
     y_multiclass_low_classes = pd.Series(list(range(0, 3)) * 100)  # 2 classes, 300 samples, .01 class/sample ratio
+    X = pd.DataFrame({"col": range(len(y_multiclass_low_classes))})
     assert invalid_targets_check.validate(X, y=y_multiclass_low_classes) == {"warnings": [], "errors": []}
+
+
+def test_invalid_target_data_check_mismatched_indices():
```
chukarsten, line 418: I think these are really good tests and well written.  I like that I can read them and understand exactly what the check is validating for.  I think this is fine as is, but since I don't understand entirely how this propagates to the user...I'll ask : is it useful ultimately for there to be a distinction between the length of the index being the same and the values being different and vice versa?  If not, then let's continue mission.  If so, an issue to further augment works.

chukarsten (approved):
Another Angela classic.  Well played.

evalml/data_checks/invalid_targets_data_check.py
angela97lin: @freddyaboulton Ooo yes, I love this suggestion!! Do you think returning the entire diff is okay, or should we do something like use a parameter `n_unique` to determine how many we should display / keep like we have for unique target values?

Pro: more flexibility
Con: another parameter

Or I could also combine it with n_unique / make that handle both cases 🤔

angela97lin:
@freddyaboulton @chukarsten @dsherry I believe I addressed everything that we discussed post-standup. Also separated out the case of equivalent indices sets, but not order. Please take another look if you get the chance! 😁

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -385,4 +411,64 @@ def test_invalid_target_data_check_multiclass_problem_almostcontinuous_data():
...
+    X_index_missing = list(set(y_more_than_ten_diff_indices.index) - set(X.index))
+    y_index_missing = list(set(X_large.index) - set(y_more_than_ten_diff_indices.index))
+    assert invalid_targets_check.validate(X_large, y_more_than_ten_diff_indices) == {
+        "warnings": [DataCheckWarning(message="Input target and features have mismatched indices",
+                                      data_check_name=invalid_targets_data_check_name,
+                                      message_code=DataCheckMessageCode.MISMATCHED_INDICES,
+                                      details={"indices_not_in_features": X_index_missing[:10],
+                                               "indices_not_in_target": y_index_missing[:10]}).to_dict()],
+        "errors": []
+    }
```
chukarsten, line 458: The legend of n_unique <3

chukarsten (approved):

ParthivNaresh (approved):
Excellent test case coverage, looks ready to go!

Issue #1778, opened 2021-02-02
No stack trace visible in debug log


dsherry:
RE #1710 : when there's an error during automl search, our logger says the full stack trace is visible in `evalml_debug.log`, but its not there!

We should find a way to include the stack trace

#1174 tracks overhauling our logging strategy altogether. But this issue can be closed out first, independent of #1174.

@jeremyliweishih FYI since this would simplify error debugging in LG

PR #1838, opened 2021-02-12
Engine API

### LINKED ISSUE #1296: Add Search Engine API

dsherry:
Fix #1296 

Changes
* Add `EngineBase` base class to define engine interface, and `SequentialEngine` which preserves existing behavior, evaluates pipelines in series
* Define `_pre_evaluation_callback` for logging and `_post_evaluation_callback` to add evaluation results to `AutoMLSearch`
* Move `AutoMLSearch. _compute_cv_scores` to `EngineBase.train_and_score_pipeline` as a `staticmethod`
* Refactored the `AutoMLSearch` binary threshold tuning logic into `tune_binary_threshold` so it could be reused. Fixed bug where all bin class pipelines were getting a threshold of 0.5 -- [now will get `None`](https://github.com/alteryx/evalml/pull/1838/files#diff-068c8096e3b254dc70a1399ca4a70f74aea2f8e17e03ae111460078891bf4c7eR2107) as was the original intent.
* Tweak criterion for doing bin threshold tuning to apply to timeseries binary classification as well.
* Now that `AutoMLSearch` constructor expects data, move much of the `search` setup logic into the constructor to simplify the behavior.
* Have `AutoMLAlgorithm.add_result` raise error if pipeline is unrecognized
* Update and simplify keyboard interrupt code; no intended change in user behavior
* Rename `_check_stopping_condition` to `_should_continue`
* Add traceback to the end of `log_error_callback`, which I believe partially handles #1778 (@bchen1116 heads up). This was helpful to me when debugging so I just went ahead and did it.
* Fixed some test bugs
    * In `test_automl.py`, `CustomClassificationObjective` wasn't subclassed from `BinaryClassificationObjective`, same for regression
    * Many tests were mocking pipeline `score` but not providing a return value, causing scores to be `nan`

Future:
* Pass a config object to engine classes' constructor instead of `AutoMLSearch` instance. I couldn't easily do this today because of the logging callback -- it expects a reference to `AutoMLSearch` to add the error to the `results` structure
* When we implement parallel evaluation, we'll have to decide what to do with the early stopping parameters, as those won't map over easily to the parallel batched evaluation format.
* **The changes in this PR are not threadsafe**. The calls to the should_continue/pre/post callbacks access and modify `AutoMLSearch` state without any sort of locking or atomicity. This is fine for this PR, but when we implement parallel evaluation, we'll need to address this.
* Add ability to accept an `EngineBase` subclass as input to `AutoMLSearch`, rather than creating internally. This would require thought about how to pass parameters and create the instance. We'd probably wanna switch to using a config object first for `AutoMLSearch`

Big credit to @christopherbunn for providing the initial implementation this PR is based on! 🙏

### DIFF
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 
+from evalml.exceptions import PipelineNotFoundError
 from evalml.tuners import SKOptTuner
 
 
@@ -51,6 +52,8 @@ class AutoMLAlgorithm(ABC):
             score_to_minimize (float): The score obtained by this pipeline on the primary objective, converted so that lower values indicate better pipelines.
             pipeline (PipelineBase): The trained pipeline object which was used to compute the score.
         """
+        if pipeline.name not in self._tuners:
+            raise PipelineNotFoundError(f"No such pipeline allowed in this AutoML search: {pipeline.name}")
         self._tuners[pipeline.name].add(pipeline.parameters, score_to_minimize)
 
     @property
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,8 +1,6 @@
 import copy
-import sys
 import time
-import traceback
-from collections import OrderedDict, defaultdict
+from collections import defaultdict
 
 import cloudpickle
 import numpy as np
@@ -13,9 +11,11 @@ from .pipeline_search_plots import PipelineSearchPlots
 
 from evalml.automl.automl_algorithm import IterativeAlgorithm
 from evalml.automl.callbacks import log_error_callback
+from evalml.automl.engine import SequentialEngine
 from evalml.automl.utils import (
     get_default_primary_search_objective,
-    make_data_splitter
+    make_data_splitter,
+    tune_binary_threshold
 )
 from evalml.data_checks import (
     AutoMLDataChecks,
@@ -24,11 +24,7 @@ from evalml.data_checks import (
     EmptyDataChecks,
     HighVarianceCVDataCheck
 )
-from evalml.exceptions import (
-    AutoMLSearchException,
-    PipelineNotFoundError,
-    PipelineScoreError
-)
+from evalml.exceptions import AutoMLSearchException, PipelineNotFoundError
 from evalml.model_family import ModelFamily
 from evalml.objectives import (
     get_core_objectives,
@@ -36,11 +32,9 @@ from evalml.objectives import (
     get_objective
 )
 from evalml.pipelines import (
-    BinaryClassificationPipeline,
     MeanBaselineRegressionPipeline,
     ModeBaselineBinaryPipeline,
     ModeBaselineMulticlassPipeline,
-    PipelineBase,
     TimeSeriesBaselineBinaryPipeline,
     TimeSeriesBaselineMulticlassPipeline,
     TimeSeriesBaselineRegressionPipeline
@@ -239,6 +233,7 @@ class AutoMLSearch:
 
         self.patience = patience
         self.tolerance = tolerance or 0.0
+
         self._results = {
             'pipeline_results': {},
             'search_order': [],
@@ -258,7 +253,7 @@ class AutoMLSearch:
         self.allowed_pipelines = allowed_pipelines
         self.allowed_model_families = allowed_model_families
         self._automl_algorithm = None
-        self._start = None
+        self._start = 0.0
         self._baseline_cv_scores = {}
         self.show_batch_output = False
 
@@ -275,6 +270,92 @@ class AutoMLSearch:
                                                    n_splits=3, shuffle=True, random_seed=self.random_seed)
         self.data_splitter = self.data_splitter or default_data_splitter
         self.pipeline_parameters = pipeline_parameters if pipeline_parameters is not None else {}
+        self.search_iteration_plot = None
+        self._interrupted = False
+
+        self._engine = SequentialEngine(self.X_train,
+                                        self.y_train,
+                                        self,
+                                        should_continue_callback=self._should_continue,
+                                        pre_evaluation_callback=self._pre_evaluation_callback,
+                                        post_evaluation_callback=self._post_evaluation_callback)
+
+        if self.allowed_pipelines is None:
+            logger.info("Generating pipelines to search over...")
+            allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
+            logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
+            self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
+
+        if self.allowed_pipelines == []:
+            raise ValueError("No allowed pipelines to search")
+
+        run_ensembling = self.ensembling
+        if run_ensembling and len(self.allowed_pipelines) == 1:
+            logger.warning("Ensembling is set to True, but the number of unique pipelines is one, so ensembling will not run.")
+            run_ensembling = False
+
+        if run_ensembling and self.max_iterations is not None:
+            # Baseline + first batch + each pipeline iteration + 1
+            first_ensembling_iteration = (1 + len(self.allowed_pipelines) + len(self.allowed_pipelines) * self._pipelines_per_batch + 1)
+            if self.max_iterations < first_ensembling_iteration:
+                run_ensembling = False
+                logger.warning(f"Ensembling is set to True, but max_iterations is too small, so ensembling will not run. Set max_iterations >= {first_ensembling_iteration} to run ensembling.")
+            else:
+                logger.info(f"Ensembling will run at the {first_ensembling_iteration} iteration and every {len(self.allowed_pipelines) * self._pipelines_per_batch} iterations after that.")
+
+        if self.max_batches and self.max_iterations is None:
+            self.show_batch_output = True
+            if run_ensembling:
+                ensemble_nth_batch = len(self.allowed_pipelines) + 1
+                num_ensemble_batches = (self.max_batches - 1) // ensemble_nth_batch
+                if num_ensemble_batches == 0:
+                    logger.warning(f"Ensembling is set to True, but max_batches is too small, so ensembling will not run. Set max_batches >= {ensemble_nth_batch + 1} to run ensembling.")
+                else:
+                    logger.info(f"Ensembling will run every {ensemble_nth_batch} batches.")
+
+                self.max_iterations = (1 + len(self.allowed_pipelines) +
+                                       self._pipelines_per_batch * (self.max_batches - 1 - num_ensemble_batches) +
+                                       num_ensemble_batches)
+            else:
+                self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self.max_batches - 1))
+        self.allowed_model_families = list(set([p.model_family for p in (self.allowed_pipelines)]))
+
+        logger.debug(f"allowed_pipelines set to {[pipeline.name for pipeline in self.allowed_pipelines]}")
+        logger.debug(f"allowed_model_families set to {self.allowed_model_families}")
+        if len(self.problem_configuration):
+            pipeline_params = {**{'pipeline': self.problem_configuration}, **self.pipeline_parameters}
+        else:
+            pipeline_params = self.pipeline_parameters
+
+        self._automl_algorithm = IterativeAlgorithm(
+            max_iterations=self.max_iterations,
+            allowed_pipelines=self.allowed_pipelines,
+            tuner_class=self.tuner_class,
+            random_seed=self.random_seed,
+            n_jobs=self.n_jobs,
+            number_features=self.X_train.shape[1],
+            pipelines_per_batch=self._pipelines_per_batch,
+            ensembling=run_ensembling,
+            pipeline_params=pipeline_params
+        )
+
+    def _pre_evaluation_callback(self, pipeline):
+        if self.start_iteration_callback:
+            self.start_iteration_callback(pipeline.__class__, pipeline.parameters, self)
+        desc = f"{pipeline.name}"
+        if len(desc) > AutoMLSearch._MAX_NAME_LEN:
+            desc = desc[:AutoMLSearch._MAX_NAME_LEN - 3] + "..."
+        desc = desc.ljust(AutoMLSearch._MAX_NAME_LEN)
+        batch_number = 1
+        if self._automl_algorithm is not None and self._automl_algorithm.batch_number > 0:
+            batch_number = self._automl_algorithm.batch_number
+        update_pipeline(logger,
+                        desc,
+                        len(self._results['pipeline_results']) + 1,
+                        self.max_iterations,
+                        self._start,
+                        batch_number,
+                        self.show_batch_output)
 
     def _validate_objective(self, objective):
         non_core_objectives = get_non_core_objectives()
@@ -364,16 +445,11 @@ class AutoMLSearch:
         else:
             return DataChecks(data_checks)
 
-    def _handle_keyboard_interrupt(self, pipeline, current_batch_pipelines):
+    def _handle_keyboard_interrupt(self):
         """Presents a prompt to the user asking if they want to stop the search.
 
-        Arguments:
-            pipeline (PipelineBase): Current pipeline in the search.
-            current_batch_pipelines (list): Other pipelines in the batch.
-
         Returns:
-            list: Next pipelines to search in the batch. If the user decides to stop the search,
-                an empty list will be returned.
+            bool: If True, search should terminate early
         """
         leading_char = "\n"
         start_of_loop = time.time()
@@ -381,12 +457,12 @@ class AutoMLSearch:
             choice = input(leading_char + "Do you really want to exit search (y/n)? ").strip().lower()
             if choice == "y":
                 logger.info("Exiting AutoMLSearch.")
-                return []
+                return True
             elif choice == "n":
                 # So that the time in this loop does not count towards the time budget (if set)
                 time_in_loop = time.time() - start_of_loop
                 self._start += time_in_loop
-                return [pipeline] + current_batch_pipelines
+                return False
             else:
                 leading_char = ""
 
@@ -425,68 +501,11 @@ class AutoMLSearch:
             logger.error(result["message"])
         if self._data_check_results["errors"]:
             raise ValueError("Data checks raised some warnings and/or errors. Please see `self.data_check_results` for more information or pass data_checks='disabled' to search() to disable data checking.")
-        if self.allowed_pipelines is None:
-            logger.info("Generating pipelines to search over...")
-            allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
-            logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
-            self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
-
-        if self.allowed_pipelines == []:
-            raise ValueError("No allowed pipelines to search")
-
-        run_ensembling = self.ensembling
-        if run_ensembling and len(self.allowed_pipelines) == 1:
-            logger.warning("Ensembling is set to True, but the number of unique pipelines is one, so ensembling will not run.")
-            run_ensembling = False
-
-        if run_ensembling and self.max_iterations is not None:
-            # Baseline + first batch + each pipeline iteration + 1
-            first_ensembling_iteration = (1 + len(self.allowed_pipelines) + len(self.allowed_pipelines) * self._pipelines_per_batch + 1)
-            if self.max_iterations < first_ensembling_iteration:
-                run_ensembling = False
-                logger.warning(f"Ensembling is set to True, but max_iterations is too small, so ensembling will not run. Set max_iterations >= {first_ensembling_iteration} to run ensembling.")
-            else:
-                logger.info(f"Ensembling will run at the {first_ensembling_iteration} iteration and every {len(self.allowed_pipelines) * self._pipelines_per_batch} iterations after that.")
-
-        if self.max_batches and self.max_iterations is None:
-            self.show_batch_output = True
-            if run_ensembling:
-                ensemble_nth_batch = len(self.allowed_pipelines) + 1
-                num_ensemble_batches = (self.max_batches - 1) // ensemble_nth_batch
-                if num_ensemble_batches == 0:
-                    logger.warning(f"Ensembling is set to True, but max_batches is too small, so ensembling will not run. Set max_batches >= {ensemble_nth_batch + 1} to run ensembling.")
-                else:
-                    logger.info(f"Ensembling will run every {ensemble_nth_batch} batches.")
-
-                self.max_iterations = (1 + len(self.allowed_pipelines) +
-                                       self._pipelines_per_batch * (self.max_batches - 1 - num_ensemble_batches) +
-                                       num_ensemble_batches)
-            else:
-                self.max_iterations = 1 + len(self.allowed_pipelines) + (self._pipelines_per_batch * (self.max_batches - 1))
-        self.allowed_model_families = list(set([p.model_family for p in (self.allowed_pipelines)]))
-
-        logger.debug(f"allowed_pipelines set to {[pipeline.name for pipeline in self.allowed_pipelines]}")
-        logger.debug(f"allowed_model_families set to {self.allowed_model_families}")
-        if len(self.problem_configuration):
-            pipeline_params = {**{'pipeline': self.problem_configuration}, **self.pipeline_parameters}
-        else:
-            pipeline_params = self.pipeline_parameters
-
-        self._automl_algorithm = IterativeAlgorithm(
-            max_iterations=self.max_iterations,
-            allowed_pipelines=self.allowed_pipelines,
-            tuner_class=self.tuner_class,
-            random_seed=self.random_seed,
-            n_jobs=self.n_jobs,
-            number_features=self.X_train.shape[1],
-            pipelines_per_batch=self._pipelines_per_batch,
-            ensembling=run_ensembling,
-            pipeline_params=pipeline_params
-        )
 
         log_title(logger, "Beginning pipeline search")
         logger.info("Optimizing for %s. " % self.objective.name)
         logger.info("{} score is better.\n".format('Greater' if self.objective.greater_is_better else 'Lower'))
+        logger.info(f"Using {self._engine.__class__.__name__} to train and score pipelines.")
 
         if self.max_batches is not None:
             logger.info(f"Searching up to {self.max_batches} batches for a total of {self.max_iterations} pipelines. ")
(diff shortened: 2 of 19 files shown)


evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -51,6 +52,8 @@ def add_result(self, score_to_minimize, pipeline):
             score_to_minimize (float): The score obtained by this pipeline on the primary objective, converted so that lower values indicate better pipelines.
             pipeline (PipelineBase): The trained pipeline object which was used to compute the score.
         """
+        if pipeline.name not in self._tuners:
+            raise PipelineNotFoundError(f"No such pipeline allowed in this AutoML search: {pipeline.name}")
```
dsherry, line 56: I noticed this edge case was untested, added

evalml/automl/automl_search.py
```diff
@@ -258,7 +253,7 @@ def __init__(self,
         self.allowed_pipelines = allowed_pipelines
         self.allowed_model_families = allowed_model_families
         self._automl_algorithm = None
-        self._start = None
+        self._start = 0.0
```
dsherry, line 256: Its weird but if this is initialized as `None`, it'd cause `add_to_rankings` to error out during logging if the search hasn't started yet.

evalml/automl/automl_search.py
```diff
@@ -275,6 +270,91 @@ def __init__(self,
...
         self.data_splitter = self.data_splitter or default_data_splitter
         self.pipeline_parameters = pipeline_parameters if pipeline_parameters is not None else {}
+        self.search_iteration_plot = None
+        self._interrupted = False
+
+        self._engine = SequentialEngine(self,
+                                        should_continue_callback=self._should_continue,
+                                        pre_evaluation_callback=self._pre_evaluation_callback,
+                                        post_evaluation_callback=self._post_evaluation_callback)
+        self._engine.set_data(self.X_train, self.y_train)
```
dsherry, line 280: Setting the data as a step after the constructor is unnecessary right now, but I think it'll be a useful API for the parallel impl, so leaving in.

evalml/automl/automl_search.py
```diff
@@ -275,6 +270,91 @@ def __init__(self,
...
+            tuner_class=self.tuner_class,
+            random_seed=self.random_seed,
+            n_jobs=self.n_jobs,
+            number_features=self.X_train.shape[1],
+            pipelines_per_batch=self._pipelines_per_batch,
+            ensembling=run_ensembling,
+            pipeline_params=pipeline_params
+        )
+
+    def _pre_evaluation_callback(self, pipeline):
```
dsherry, line 341: This is called before each pipeline is evaluated. It calls `start_iteration_callback` and then logs the pipeline.

evalml/automl/automl_search.py
```diff
@@ -364,32 +444,31 @@ def _validate_data_checks(self, data_checks):
...
                 time_in_loop = time.time() - start_of_loop
                 self._start += time_in_loop
-                return [pipeline] + current_batch_pipelines
+                return False
             else:
                 leading_char = ""
 
+    @staticmethod
+    def _read_input(self, leading_char=''):
+        return input(leading_char + "Do you really want to exit search (y/n)? ").strip().lower()
```
dsherry, line 470: I split this into a separate method to make it easy to mock. Codecov is complaining about it... I'm inclined to ignore haha

evalml/automl/automl_search.py
```diff
@@ -425,68 +504,11 @@ def search(self, data_checks="auto", show_iteration_plot=True):
             logger.error(result["message"])
         if self._data_check_results["errors"]:
             raise ValueError("Data checks raised some warnings and/or errors. Please see `self.data_check_results` for more information or pass data_checks='disabled' to search() to disable data checking.")
-        if self.allowed_pipelines is None:
```
dsherry, line 428: All this code got moved into the init

evalml/automl/automl_search.py
```diff
@@ -495,96 +517,111 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
+                loop_interrupted = False
+            except KeyboardInterrupt:
+                loop_interrupted = True
+                if self._handle_keyboard_interrupt():
+                    break
+            full_rankings = self.full_rankings
+            current_batch_idx = full_rankings['id'].isin(new_pipeline_ids)
+            current_batch_pipeline_scores = full_rankings[current_batch_idx]['score']
+            if len(current_batch_pipeline_scores) and current_batch_pipeline_scores.isna().all():
+                raise AutoMLSearchException(f"All pipelines in the current AutoML batch produced a score of np.nan on the primary objective {self.objective}.")
```
dsherry, line 554: I moved the keyboard interrupt handling up to `AutoMLSearch.search`. I think `AutoMLSearch` should be solely responsible for handling keyboard interrupt, and the engines should be responsible for only adding completed pipelines to the results.

There's a bigger project hiding in here... if we really wanted atomicity of pipeline evaluations, we'd want to build a queue structure which can track when jobs error out and allow `AutoMLSearch` to either reschedule them or ignore them.

evalml/automl/automl_search.py
```diff
@@ -495,96 +517,111 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
         best_pipeline = self.rankings.iloc[0]
         if not (self._best_pipeline and self._best_pipeline == self.get_pipeline(best_pipeline['id'])):
             self._best_pipeline = self.get_pipeline(best_pipeline['id'])
             if self._train_best_pipeline:
                 X_threshold_tuning = None
                 y_threshold_tuning = None
                 X_train, y_train = self.X_train, self.y_train
-                if self.optimize_thresholds and self.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and self.objective.can_optimize_threshold and is_binary(self.problem_type):
+                if is_binary(self.problem_type) and self.objective.is_defined_for_problem_type(self.problem_type) \
+                   and self.optimize_thresholds and self.objective.can_optimize_threshold:
```
dsherry, line 583: I tweaked the set of criterion we use here. This should apply for timeseries binary classification as well as plain binary classification. @freddyaboulton FYI

evalml/automl/automl_search.py
```diff
@@ -495,96 +517,111 @@ def search(self, data_checks="auto", show_iteration_plot=True):
...
-        Arguments:
-            pipeline (Pipeline): Pipeline instance to threshold
-            X_threshold_tuning (ww.DataTable): X data to tune pipeline to
-            y_threshold_tuning (ww.DataColumn): Target data to tune pipeline to
+        Returns:
+            int: the number of pipeline evaluations made in the search
+        """
+        return len(self._results['pipeline_results'])
+
+    def _should_continue(self):
```
dsherry, line 598: I kept most of the code from `_check_stopping_condition` here, no intended change in behavior, the diff just looks kinda ugly.

evalml/automl/automl_search.py
```diff
@@ -657,83 +684,17 @@ def _get_mean_cv_scores_for_all_objectives(cv_data, objective_name_to_class):
...
-                evaluation_entry['binary_classification_threshold'] = cv_pipeline.threshold
-            cv_data.append(evaluation_entry)
-        training_time = time.time() - start
-        cv_scores = pd.Series([fold['score'] for fold in cv_data])
-        cv_score_mean = cv_scores.mean()
-        logger.info(f"\tFinished cross validation - mean {self.objective.name}: {cv_score_mean:.3f}")
-        return {'cv_data': cv_data, 'training_time': training_time, 'cv_scores': cv_scores, 'cv_score_mean': cv_score_mean}
-
-    def _add_result(self, trained_pipeline, parameters, training_time, cv_data, cv_scores):
+    def _post_evaluation_callback(self, pipeline, evaluation_results):
```
dsherry, line 687: This is called after each pipeline is evaluated. It computes the results and adds them to the `AutoMLSearch` results structure.

evalml/automl/engine/engine_base.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, automl=None, should_continue_callback=None, pre_evaluation_callback=None, post_evaluation_callback=None):
+        """Base class for the engine API which handles the fitting and evaluation of pipelines during AutoML.
+
+        Arguments:
+            automl (AutoMLSearch): a reference to the AutoML search. Used to access configuration and by the error callback.
+            should_continue_callback (function): returns true if another pipeline from the list should be evaluated, false otherwise.
+            pre_evaluation_callback (function): optional callback invoked before pipeline evaluation.
+            post_evaluation_callback (function): optional callback invoked after pipeline evaluation, with args pipeline and evaluation results. Expected to return a list of pipeline IDs corresponding to each pipeline evaluation.
+        """
+        self.automl = automl
```
dsherry, line 33: This code still asks for a direct reference to `AutoMLSearch`, even though the callbacks mostly don't depend on that direct access. Two reasons for that:
* The `AutoMLSearch` error callback currently expects a reference to the `AutoMLSearch` so it can optionally add the error to `self._results`. We should look into changing this but I didn't wanna do it in this PR.
* We don't have a config object yet, so it was the easiest way to pass things like `problem_type` and `objective` to the static method `train_and_score_pipeline` below

evalml/automl/engine/engine_base.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Arguments:
+            pipeline_batch (list(PipelineBase)): A batch of pipelines to be fitted and evaluated
+
+        Returns:
+            list (int): a list of the new pipeline IDs which were created by the AutoML search.
+        """
+
+    @staticmethod
+    def train_and_score_pipeline(pipeline, automl, full_X_train, full_y_train):
```
dsherry, line 62: Code here mostly moved from `AutoMLSearch._compute_cv_scores`

evalml/automl/engine/sequential_engine.py
```diff
@@ -0,0 +1,28 @@
...
+            raise ValueError("Dataset has not been loaded into the engine.")
+        new_pipeline_ids = []
+        pipelines = copy(pipelines)
+        while self._should_continue_callback() and len(pipelines) > 0:
+            pipeline = pipelines[0]
+            self._pre_evaluation_callback(pipeline)
+            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, self.X_train, self.y_train)
+            new_pipeline_ids.append(self._post_evaluation_callback(pipeline, evaluation_result))
+            pipelines.pop(0)
+        return new_pipeline_ids
```
dsherry, line 28: A weakness of this code: if `train_and_score_pipeline` or one of the callbacks errors out, the whole `while` loop would halt.

We use the error callback inside `train_and_score_pipeline` to handle errors during pipeline evaluation, so this isn't a big concern. But if the data splitter errors, it would move up the stack. I think that behavior is acceptable for now.

evalml/automl/utils.py
```diff
@@ -68,3 +69,19 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
+
+def tune_binary_threshold(pipeline, objective, problem_type, X_threshold_tuning, y_threshold_tuning):
+    """Tunes the threshold of a binary pipeline to the X and y thresholding data
+
+    Arguments:
+        pipeline (Pipeline): Pipeline instance to threshold
+        X_threshold_tuning (ww.DataTable): X data to tune pipeline to
+        y_threshold_tuning (ww.DataColumn): Target data to tune pipeline to
+    """
+    if is_binary(problem_type) and objective.is_defined_for_problem_type(problem_type) and objective.can_optimize_threshold:
```
dsherry, line 82: Code moved from `AutoMLSearch`, no change in behavior except the specific conditions on this line.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -277,30 +276,10 @@ def test_non_optimizable_threshold(mock_fit, mock_score, X_y_binary):
...
     mock_score.assert_called()
-    assert automl.best_pipeline.threshold is not None
-    assert automl.results['pipeline_results'][0]['cv_data'][0].get('binary_classification_threshold') == 0.5
-    assert automl.results['pipeline_results'][0]['cv_data'][1].get('binary_classification_threshold') == 0.5
-    assert automl.results['pipeline_results'][0]['cv_data'][2].get('binary_classification_threshold') == 0.5
-
-
-@patch('evalml.pipelines.MulticlassClassificationPipeline.score')
-@patch('evalml.pipelines.MulticlassClassificationPipeline.fit')
-def test_non_optimizable_threshold_multi(mock_fit, mock_score, X_y_multi):
```
dsherry, line 288: This test doesn't belong in the `AutoMLSearch` tests! No thresholds are defined for multiclass pipelines.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -340,34 +319,6 @@ def test_max_time_units(X_y_binary):
         AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective='F1', max_time=(30, 'minutes'))
 
 
-def test_early_stopping(caplog, logistic_regression_binary_pipeline_class, X_y_binary):
```
dsherry, line 343: Moved to `test_automl.py`

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -91,31 +90,6 @@ def add_result_callback(results, trained_pipeline, automl_obj, counts=counts):
     assert counts["add_result_callback"] == max_iterations
 
 
-def test_early_stopping(caplog, linear_regression_pipeline_class, X_y_regression):
```
dsherry, line 94: Moved to `test_automl.py`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -91,8 +98,8 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type):
         assert isinstance(results['cv_data'], list)
         for cv_result in results['cv_data']:
             assert cv_result.keys() == expected_cv_data_keys
-            if automl_type == ProblemTypes.BINARY:
-                assert isinstance(cv_result['binary_classification_threshold'], float)
+            if objective == 'F1':
+                assert cv_result['binary_classification_threshold'] == 0.5
```
dsherry, line 102: Hmm, I am forgetting why this should be `0.5` and not `None`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -824,25 +829,42 @@ def test_add_to_rankings(mock_fit, mock_score, dummy_binary_pipeline_class, X_y_
...
+    mock_score.return_value = {'Log Loss Binary': 0.5678}
+    test_pipeline_2 = dummy_binary_pipeline_class(parameters={'Mock Classifier': {'a': 1.234}})
+    automl.add_to_rankings(test_pipeline_2)
+    assert automl.best_pipeline.name == test_pipeline.name
+    assert automl.best_pipeline.parameters == test_pipeline.parameters
+    assert automl.best_pipeline.component_graph == test_pipeline.component_graph
+    assert len(automl.rankings) == 2
+    assert len(automl.full_rankings) == 3
+    assert 0.5678 not in automl.rankings['score'].values
+    assert 0.5678 in automl.full_rankings['score'].values
```
bchen1116, line 859: Why would this new score for `test_pipeline_2` not show up in full rankings? Seems like the full rankings got updated since the length did increase.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1461,6 +1518,33 @@ def test_max_batches_works(mock_pipeline_fit, mock_score, mock_regression_fit, m
...
+    }
+
+    scores = [0.95, 0.84, 0.96]  # 0.96 is only 1% greater so it doesn't trigger patience due to tolerance
+    for id in mock_results['search_order']:
+        mock_results['pipeline_results'][id] = {}
+        mock_results['pipeline_results'][id]['score'] = scores[id]
+        mock_results['pipeline_results'][id]['pipeline_class'] = logistic_regression_binary_pipeline_class
+    automl._results = mock_results
+
+    automl._should_continue()
```
bchen1116, line 1543: can we `assert not automl._should_continue()` here? Just to make sure it returned false?

bchen1116 (commented):
Big PR! Tests and implementation looks good, but had a few cleanup suggestions. Had a question about `add_to_rankings`. It seems like the performance of it changed compared to main, because when I run
```python

class NewPipeline(BinaryClassificationPipeline):
    component_graph = ['Imputer', 'LightGBM Classifier']

X, y = load_breast_cancer()
automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective='f1')
automl.search()    
automl.add_to_rankings(NewPipeline({}))
automl.full_rankings
```
I get 
image
whereas running on main doesn't give me this same error.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -84,7 +84,9 @@ def next_batch(self):
             num_pipeline_classes = (len(self._first_batch_results) + 1) if self.ensembling else len(self._first_batch_results)
             idx = (self._batch_number - 1) % num_pipeline_classes
             pipeline_class = self._first_batch_results[idx][1]
-            for i in range(self.pipelines_per_batch):
+
+            n_pipelines_to_add = self.pipelines_per_batch
+            for i in range(n_pipelines_to_add):
```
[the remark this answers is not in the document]
dsherry: Oops, nah this was left over from changes I backed out. Will delete

evalml/automl/engine/engine_base.py
```diff
@@ -0,0 +1,142 @@
...
+            post_evaluation_callback (function): optional callback invoked after pipeline evaluation, with args pipeline and evaluation results. Expected to return a list of pipeline IDs corresponding to each pipeline evaluation.
+        """
+        self.automl = automl
+        self._should_continue_callback = should_continue_callback
+        self._pre_evaluation_callback = pre_evaluation_callback
+        self._post_evaluation_callback = post_evaluation_callback
+        self.X_train = None
+        self.y_train = None
+
+    def set_data(self, X_train, y_train=None):
```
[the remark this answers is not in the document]
dsherry: Ok, the fact that both you and @bchen1116 mentioned this is a clear sign it should get deleted 😆

evalml/automl/engine/engine_base.py
```diff
@@ -0,0 +1,142 @@
...
+                diff_string = f"Missing target values in the training set after data split: {diff_train}. " if diff_train else ""
+                diff_string += f"Missing target values in the validation set after data split: {diff_valid}." if diff_valid else ""
+                if diff_string:
+                    raise Exception(diff_string)
+            objectives_to_score = [automl.objective] + automl.additional_objectives
+            cv_pipeline = None
+            try:
+                X_threshold_tuning = None
+                y_threshold_tuning = None
+                if automl.optimize_thresholds and automl.objective.is_defined_for_problem_type(ProblemTypes.BINARY) and automl.objective.can_optimize_threshold and is_binary(automl.problem_type):
```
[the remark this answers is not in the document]
dsherry: Yes, thank you!

evalml/automl/automl_search.py
dsherry: @chukarsten that's a good point. At the moment I think its more of a nebulous long-term concern; I'd like to see how the parallel project shakes out before we file anything for it.

evalml/automl/engine/engine_base.py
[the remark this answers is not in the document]
dsherry: Will delete! Thank you for raising it, good to know this was confusing for both you and @chukarsten

evalml/automl/engine/sequential_engine.py
```diff
@@ -0,0 +1,28 @@
...
+        Arguments:
+            pipelines (list(PipelineBase)): A batch of pipelines to be fitted and evaluated.
+
+        Returns:
+            list (int): a list of the new pipeline IDs which were created by the AutoML search.
+        """
+        if self.X_train is None or self.y_train is None:
+            raise ValueError("Dataset has not been loaded into the engine.")
+        new_pipeline_ids = []
+        pipelines = copy(pipelines)
```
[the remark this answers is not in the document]
dsherry: You are right! We clone the pipeline in `train_and_score_pipeline`, so that we can train one on each CV fold.

The reason I am copying this list is that `pipelines.pop(0)` modifies the list. Because python passes by reference this means its the list from up the stack which is modified. That's not great behavior IMO, too confusing.

So the question is, how can we have this code iterate through a list of pipelines without altering the parent pipeline list from up the stack?

Your comment is making me think we should use an index instead of copying/popping. I'll give that a shot.

evalml/automl/engine/sequential_engine.py
dsherry: That's not dumb! You are correct, adding proper error handling now is the responsible thing to do. I was being lazy. Will do.

evalml/automl/utils.py
dsherry: That is a great point. I'll add that.

evalml/automl/engine/sequential_engine.py
dsherry: I took a close look at this. I still agree that we should update the way we do error handling. But I'd like us to do it in a separate PR. There's a decent amount to consider. This PR doesn't change the existing error handling pattern; we will have to in order to complete the parallel work.

evalml/tests/automl_tests/test_automl.py
dsherry: @bchen1116 so in this case, we add a new pipeline with a worse log loss than the current best for the given model family. Under that scenario, I would expect `full_rankings` to have a row for the new pipeline, but `rankings` should remain unchanged, since the new pipeline doesn't beat the old for that family. Take a peek at the docstrings for `rankings` and `full_rankings` to see the intended behavior of those accessors.

evalml/tests/automl_tests/test_automl.py
dsherry: Oh yeah, definitely. I didn't modify the code for this test but I will do so now. Thanks!

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -311,35 +274,41 @@ def test_automl_supports_time_series_regression(mock_fit, mock_score, X_y_regres
         assert result['parameters']['pipeline'] == configuration
 
 
-@patch('evalml.pipelines.RegressionPipeline.fit')
 @patch('evalml.pipelines.RegressionPipeline.score')
-def test_automl_pickle_generated_pipeline(mock_regression_score, mock_regression_fit, X_y_regression):
-    class RegressionPipelineCustoms(RegressionPipeline):
+@patch('evalml.pipelines.RegressionPipeline.fit')
+def test_automl_pickle_generated_pipeline(mock_regression_fit, mock_regression_score, X_y_regression):
```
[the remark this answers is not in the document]
dsherry: That is a good point :) I think when we started using mocking we didn't realize that you could set `return_value` and `side_effect` in the call to `patch`.

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -311,35 +274,41 @@ def test_automl_supports_time_series_regression(mock_fit, mock_score, X_y_regres
...
     pipeline = GeneratedPipelineRegression
 
-    a = AutoMLSearch(X_train=X, y_train=y, problem_type='regression')
+    allowed_estimators = get_estimators('regression')
+    allowed_pipelines = [make_pipeline(X, y, estimator, problem_type='regression') for estimator in allowed_estimators]
+    allowed_pipelines.append(RegressionPipelineCustom)
+    a = AutoMLSearch(X_train=X, y_train=y, problem_type='regression', allowed_pipelines=allowed_pipelines)
     a.search()
-    a.add_to_rankings(RegressionPipelineCustoms({}))
+    a.add_to_rankings(RegressionPipelineCustom({}))
```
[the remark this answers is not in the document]
dsherry: If you don't add `RegressionPipelineCustom` to the `allowed_pipelines` when creating the search, there won't be a tuner created for that pipeline! This PR actually adds an exception to `AutoMLAlgorithm` for that precise case. (I will go make sure we have a negative test for that)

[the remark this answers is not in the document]
dsherry: Yep, search this PR for the string "No such pipeline allowed in this AutoML search" and you'll see it

evalml/tests/automl_tests/test_engine_base.py
```diff
@@ -0,0 +1,63 @@
...
+def test_train_and_score_pipelines_error(mock_fit, mock_score, dummy_binary_pipeline_class, X_y_binary, caplog):
+    X, y = X_y_binary
+    mock_score.side_effect = Exception('yeet')
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', max_time=1, max_batches=1,
+                          allowed_pipelines=[dummy_binary_pipeline_class])
+    pipeline = dummy_binary_pipeline_class({})
+    evaluation_result = EngineBase.train_and_score_pipeline(pipeline, automl, automl.X_train, automl.y_train)
+    assert mock_fit.call_count == automl.data_splitter.get_n_splits()
+    assert mock_score.call_count == automl.data_splitter.get_n_splits()
+    assert evaluation_result.get('training_time') is not None
```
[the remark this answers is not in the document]
dsherry: I wanted to make sure that the top-level fields expected in the return from `EngineBase.train_and_score_pipeline` are all populated

evalml/tests/automl_tests/test_sequential_engine.py
```diff
@@ -0,0 +1,97 @@
...
+    mock_should_continue_callback = MagicMock(side_effect=[True, False])
+    mock_pre_evaluation_callback = MagicMock()
+    mock_post_evaluation_callback = MagicMock(side_effect=[123, 456])
+
+    engine = SequentialEngine(automl=automl,
+                              should_continue_callback=mock_should_continue_callback,
+                              pre_evaluation_callback=mock_pre_evaluation_callback,
+                              post_evaluation_callback=mock_post_evaluation_callback)
+    engine.set_data(automl.X_train, automl.y_train)
+    new_pipeline_ids = engine.evaluate_batch(pipelines)
```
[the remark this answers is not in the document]
dsherry: I'm gonna say no for today, but yeah this may be the way to go in subsequent PRs for the parallel work

dsherry:
@bchen1116 thanks for your comments!

That example you included is fantastic. You are correct. The way this code is laid out in this PR currently, it would prevent users from adding pipelines to the rankings leaderboard if they were not initially included in `allowed_pipelines`. After reflecting on your comments here and in the PR, I agree, there's no reason to add this limitation. I think I got confused, because its kinda weird that the existing code always tries to add pipelines to the automl algorithm tuners--it doesn't make sense to do that if no tuner exists!

I just updated the code to preserve the existing behavior for `add_to_rankings` (no exception thrown), and also tried to clean up the implementation. `AutoMLAlgorithm.add_result` will still raise an exception, because that seems like a helpful thing for that API to do, but the `AutoMLSearch` results callback swallows the exception and continues.

evalml/tests/automl_tests/test_automl.py
bchen1116: got it, somehow I read both assert lines to be `assert 0.5678 not in ...`. Thanks for clarifying!

evalml/automl/engine/sequential_engine.py
```diff
@@ -0,0 +1,26 @@
...
+    def evaluate_batch(self, pipelines):
+        """Evaluate a batch of pipelines using the current dataset and AutoML state.
+
+        Arguments:
+            pipelines (list(PipelineBase)): A batch of pipelines to be fitted and evaluated.
+
+        Returns:
+            list (int): a list of the new pipeline IDs which were created by the AutoML search.
+        """
+        if self.X_train is None or self.y_train is None:
```
bchen1116, line 16: If we're checking that `self.X_train, self.y_train` aren't None, should we also do this for the AutoMLSearch object? If that's None, this call should also fail
image

bchen1116 (approved):
Thanks for addressing the comments! The implementation looks good to me, just left a few nit-picky comments but nothing too pressing here. 

Do you think it'd be useful to run perf tests on this to ensure that the new search algo performances don't differ too much from what we currently have on main?

evalml/automl/automl_search.py
```diff
@@ -275,6 +270,92 @@ def __init__(self,
...
+            else:
+                logger.info(f"Ensembling will run at the {first_ensembling_iteration} iteration and every {len(self.allowed_pipelines) * self._pipelines_per_batch} iterations after that.")
+
+        if self.max_batches and self.max_iterations is None:
+            self.show_batch_output = True
+            if run_ensembling:
+                ensemble_nth_batch = len(self.allowed_pipelines) + 1
+                num_ensemble_batches = (self.max_batches - 1) // ensemble_nth_batch
+                if num_ensemble_batches == 0:
+                    logger.warning(f"Ensembling is set to True, but max_batches is too small, so ensembling will not run. Set max_batches >= {ensemble_nth_batch + 1} to run ensembling.")
```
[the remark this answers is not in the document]
dsherry: It doesn't matter! If we don't have enough evaluations to reach the ensembling, enabling/disabling ensembling will have no effect.

evalml/automl/engine/engine_base.py
```diff
@@ -0,0 +1,135 @@
...
+
+        Arguments:
+            pipeline_batch (list(PipelineBase)): A batch of pipelines to be fitted and evaluated
+
+        Returns:
+            list (int): a list of the new pipeline IDs which were created by the AutoML search.
+        """
+
+    @staticmethod
+    def train_and_score_pipeline(pipeline, automl, full_X_train, full_y_train):
```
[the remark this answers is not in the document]
dsherry: We are passing `X_train` from the automl obj -- see `SequentialEngine.evaluate_batch`. The variable name here is not importance since this is a static method.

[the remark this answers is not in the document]
dsherry: Ah if you're asking why not use `automl.X_train` here, yeah, I would like us to eventually not have a direct reference to the automl search in this method. Why: because this code will eventually be running on a remote worker.

evalml/automl/engine/sequential_engine.py
dsherry: That's true. I suppose it can't hurt to add that. Since there's only one call site at the moment, and if we didn't set automl search there all the tests will fail, I will hold off on adding that right now.

evalml/automl/engine/engine_base.py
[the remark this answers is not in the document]
bchen1116: Mmm, got it, that makes sense. Thanks for clarifying!

evalml/automl/utils.py
```diff
@@ -68,3 +69,19 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
         return TimeSeriesSplit(n_splits=n_splits, gap=problem_configuration.get('gap'),
                                max_delay=problem_configuration.get('max_delay'))
+
+
+def tune_binary_threshold(pipeline, objective, problem_type, X_threshold_tuning, y_threshold_tuning):
+    """Tunes the threshold of a binary pipeline to the X and y thresholding data
+
+    Arguments:
+        pipeline (Pipeline): Pipeline instance to threshold
+        X_threshold_tuning (ww.DataTable): X data to tune pipeline to
```
angela97lin, line 79: Nit-picking at its finest but maybe change this to "Features to tune pipeline to" to match docstring for `y_threshold_tuning`? :)

evalml/automl/utils.py
```diff
@@ -68,3 +69,19 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
             raise ValueError("problem_configuration is required for time series problem types")
         return TimeSeriesSplit(n_splits=n_splits, gap=problem_configuration.get('gap'),
                                max_delay=problem_configuration.get('max_delay'))
+
+
+def tune_binary_threshold(pipeline, objective, problem_type, X_threshold_tuning, y_threshold_tuning):
```
angela97lin, line 74: Just curious, is `y_threshold_tuning` a preoptimization? We currently don't have anything that does this right?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1461,6 +1517,33 @@ def test_max_batches_works(mock_pipeline_fit, mock_score, mock_regression_fit, m
     assert automl.full_rankings.shape[0] == n_results
 
 
+def test_early_stopping(caplog, logistic_regression_binary_pipeline_class, X_y_binary):
+    X, y = X_y_binary
+    with pytest.raises(ValueError, match='patience value must be a positive integer.'):
+        AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective='AUC', max_iterations=5, allowed_model_families=['linear_model'], patience=-1, random_seed=0)
+    with pytest.raises(ValueError, match='tolerance value must be'):
+        AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective='AUC', max_iterations=5, allowed_model_families=['linear_model'], patience=1, tolerance=1.5, random_seed=0)
```
angela97lin, line 1525: Nit-picking and you probably just moved these but maybe we can separate these first two tests into its own tests, since its more related to invalid values passed in for patience/tolerance?

evalml/automl/utils.py
```diff
@@ -68,3 +69,19 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
+    """Tunes the threshold of a binary pipeline to the X and y thresholding data
+
+    Arguments:
+        pipeline (Pipeline): Pipeline instance to threshold
+        X_threshold_tuning (ww.DataTable): X data to tune pipeline to
+        y_threshold_tuning (ww.DataColumn): Target data to tune pipeline to
+    """
+    if is_binary(problem_type) and objective.is_defined_for_problem_type(problem_type) and objective.can_optimize_threshold:
+        pipeline.threshold = 0.5
```
angela97lin, line 83: RE other comment, pipeline threshold set here.

evalml/tests/automl_tests/test_automl.py
angela97lin: The way that `tune_binary_threshold` is currently defined, if `is_binary(problem_type) and objective.is_defined_for_problem_type(problem_type) and objective.can_optimize_threshold` (true for F1 objective), the pipeline will automatically set its threshold to 0.5.

angela97lin (commented):
Wow, this is an incredible PR with a lot of great refactoring 👁️ 👁️

evalml/tests/automl_tests/test_automl.py
dsherry: Sure!

evalml/automl/utils.py
dsherry: Hm, not sure I follow. We need both the features and the target labels in order to compute the "optimal" threshold value.

dsherry:
Perf test results [here](https://alteryx.quip.com/QZPyAwGyOi62/engine-API-perf-testsipynb-Engine-API-perf-test-results). Flat as expected!

evalml/tests/automl_tests/test_automl.py
dsherry: Ah thanks, so according to our current definition, this expected value is correct. 👍

evalml/automl/utils.py
angela97lin: Oops, I didn't see the line using y_threshold_tuning and was confused about whether or not it was used. My bad!

PR #1840, opened 2021-02-12
Add Traceback to the AutoMLSearch Logs

### LINKED ISSUE #1778: No stack trace visible in debug log

bchen1116:
fix #1778 

Screenshot of the `evalml_debug.log` file:
image

### DIFF
--- a/evalml/automl/callbacks.py
+++ b/evalml/automl/callbacks.py
@@ -44,4 +44,6 @@ def log_error_callback(exception, traceback, automl, **kwargs):
         logger.info(f"\t\t\tFold {fold_num}: All scores will be replaced with nan.")
     logger.info(f"\t\t\tFold {fold_num}: Please check {logger.handlers[1].baseFilename} for the current hyperparameters and stack trace.")
     logger.debug(f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.hyperparameters}")
+    trace = '\n'.join(traceback)
+    logger.debug(f"\t\t\tFold {fold_num}: Traceback:\n{trace}")
     logger.info(f"\t\t\tFold {fold_num}: Exception during automl search: {str(exception)}")
--- a/evalml/automl/callbacks.py
+++ b/evalml/automl/callbacks.py
@@ -36,7 +36,7 @@ def log_error_callback(exception, traceback, automl, **kwargs):
     """Logs the exception thrown as an error. Will not throw. This is the default behavior for AutoMLSearch."""
     fold_num = kwargs.get('fold_num')
     pipeline = kwargs.get('pipeline')
-    trace = '\n'.join(traceback)
+    trace = "\n".join(traceback) if traceback else ""
     if isinstance(exception, PipelineScoreError):
         logger.info(f"\t\t\tFold {fold_num}: Encountered an error scoring the following objectives: {', '.join(exception.exceptions)}.")
         logger.info(f"\t\t\tFold {fold_num}: The scores for these objectives will be replaced with nan.")
--- a/evalml/automl/callbacks.py
+++ b/evalml/automl/callbacks.py
@@ -36,14 +36,15 @@ def log_error_callback(exception, traceback, automl, **kwargs):
     """Logs the exception thrown as an error. Will not throw. This is the default behavior for AutoMLSearch."""
     fold_num = kwargs.get('fold_num')
     pipeline = kwargs.get('pipeline')
+    trace = '\n'.join(traceback)
     if isinstance(exception, PipelineScoreError):
         logger.info(f"\t\t\tFold {fold_num}: Encountered an error scoring the following objectives: {', '.join(exception.exceptions)}.")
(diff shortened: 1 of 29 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -1848,6 +1848,11 @@ def test_automl_error_callback_none(mock_fit, mock_score, X_y_binary, caplog):
     automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", error_callback=None, train_best_pipeline=False, n_jobs=1)
     automl.search()
     assert msg in caplog.text
+    traceback_string1 = "return self._execute_mock_call(*args, **kwargs)"
```
bchen1116, line 1851: These are messages that exist in the tracebacks only

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1858,7 +1863,11 @@ def test_automl_error_callback_silent(mock_fit, mock_score, X_y_binary, caplog):
     caplog.clear()
     automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", error_callback=silent_error_callback, train_best_pipeline=False, n_jobs=1)
     automl.search()
-    assert msg not in caplog.text
+    traceback_string1 = "return self._execute_mock_call(*args, **kwargs)"
+    traceback_string2 = "return self._mock_call(*args, **kwargs)"
+    traceback_string3 = "cv_pipeline.fit(X_train, y_train)"
+    for messages in [traceback_string1, traceback_string2, traceback_string3, msg]:
+        assert messages not in caplog.text
```
bchen1116, line 1870: Only `silent_error_callback` doesn't log the traceback

evalml/automl/callbacks.py
```diff
@@ -44,4 +44,6 @@ def log_error_callback(exception, traceback, automl, **kwargs):
         logger.info(f"\t\t\tFold {fold_num}: All scores will be replaced with nan.")
     logger.info(f"\t\t\tFold {fold_num}: Please check {logger.handlers[1].baseFilename} for the current hyperparameters and stack trace.")
     logger.debug(f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.hyperparameters}")
+    trace = '\n'.join(traceback)
```
freddyaboulton, line 47: If the exception a `PipelineScoreError`, then `e.message` contains information about how each objective failed. I think we should add that to the traceback. 

Maybe we can do something like this:

```python
    trace = '\n'.join(traceback)
    if isinstance(exception, PipelineScoreError):
        logger.info(f"\t\t\tFold {fold_num}: Encountered an error scoring the following objectives: {', '.join(exception.exceptions)}.")
        logger.info(f"\t\t\tFold {fold_num}: The scores for these objectives will be replaced with nan.")
        trace += e.message
    else:
        logger.info(f"\t\t\tFold {fold_num}: Encountered an error.")
        logger.info(f"\t\t\tFold {fold_num}: All scores will be replaced with nan.")

    logger.info(f"\t\t\tFold {fold_num}: Please check {logger.handlers[1].baseFilename} for the current hyperparameters and stack trace.")
    logger.info(f"\t\t\tFold {fold_num}: Exception during automl search: {str(exception.__name__)}")
    logger.debug(f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.hyperparameters}")
    logger.debug(f"\t\t\tFold {fold_num}: Traceback:\n{trace}")
```

evalml/tests/automl_tests/test_automl.py
freddyaboulton: It's confusing though cause there is no traceback since nothing errors. I think we should update the test. Could be done in a separate PR but it'd be a small change.

freddyaboulton (approved):
@bchen1116 Looks good! Thanks for updating all of our existing tests!

I have a comment on adding some info to the traceback in the event of a PipelineScoreError that I'd like to resolve before merge.

evalml/tests/automl_tests/test_automl.py
chukarsten: @bchen1116 @freddyaboulton if an issue is coming out of this, can we also maybe parameterize these tests into one?  Ping me if you do it in a separate PR, I'd like to make sure it doesn't get lost in the backlog.

chukarsten (approved):

bchen1116:
@freddyaboulton @chukarsten simplified the tests, so I don't think we'll need to file that in a new issue.

evalml/tests/automl_tests/test_automl.py
bchen1116: @freddyaboulton @chukarsten simplified the tests, so I don't think we'll need to file that in a new issue.

freddyaboulton (approved):
Looks great @bchen1116 !

PR #1842, opened 2021-02-12
Fixing coverage in TextFeaturizer

freddyaboulton:
### Pull Request Description

With the removal of `text_columns` in #1652, the `_get_feature_provenance` method of `TextFeaturizer` and `LSA` was [no longer being covered](https://codecov.io/gh/alteryx/evalml/src/main/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py) because the test dataset doesn't have any text features and the text features were specified with `text_columns`. 

-----

### DIFF
--- a/evalml/tests/model_understanding_tests/test_permutation_importance.py
+++ b/evalml/tests/model_understanding_tests/test_permutation_importance.py
@@ -71,6 +71,9 @@ class LinearPipelineWithTextFeatures(BinaryClassificationPipeline):
     component_graph = ['Imputer', 'Drop Columns Transformer', TextFeaturizer, OneHotEncoder, 'Random Forest Classifier']
 
 
+class LinearPipelineWithTextFeaturizerNoTextFeatures(LinearPipelineWithTextFeatures):
+    """Testing a pipeline with TextFeaturizer but no text features."""
+
 class LinearPipelineWithDoubling(BinaryClassificationPipeline):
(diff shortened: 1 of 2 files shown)


evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -149,6 +152,9 @@ def test_fast_permutation_importance_matches_sklearn_output(mock_supports_fast_i
                     "dependency not installed.")
     X, y = load_fraud(100)
 
+    if pipeline_class == LinearPipelineWithTextFeatures:
+        X = X.set_types(logical_types={'provider': 'NaturalLanguage'})
+
```
chukarsten, line 157: Just for my edification, is the reason this test covers those lines because the change in logical types forces a derivation of the provenance?

chukarsten (approved):
Cool, took me a second to understand what was going on, but cool.

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Yea the provenance is only computed if the text featurizer creates features. If you run it on a dataset without text features, the provenance will always be an empty dict. Since the text featurizer will (now) only compute features on columns of logical type `NaturalLanguage`, we have to create one for the component to be able to do its thing.

PR #1844, opened 2021-02-14
Workflow - Windows Unit Tests

### LINKED ISSUE #1827: GitHub Actions - Workflow - windows python X.X unit tests
GitHub Actions workflow for windows python X.X unit tests.

Quip Doc: https://alteryx.quip.com/uEJnAwT2Bwih/Workflow-windowstestallpythonversions

ParthivNaresh:
Fixes #1827


.github/workflows/windows_unit_tests.yml
```diff
@@ -0,0 +1,74 @@
...
+          . $env:USERPROFILE\Miniconda3\shell\condabin\conda-hook.ps1
+          conda activate curr_py
+          python -m pip install --upgrade pip
+          python -m pip install .
+          python -m pip install -r test-requirements.txt
+      - name: Run unit tests
+        run: |
+          . $env:USERPROFILE\Miniconda3\shell\condabin\conda-hook.ps1
+          conda activate curr_py
+          pytest evalml/ -n 4 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
```
freddyaboulton, line 74: Why not make this the `win-git-test` command?

freddyaboulton (approved):
@ParthivNaresh Looks good to me! Looks like the runtime for the windows tests is about 30 minutes which is about the same as circleci so I think it's a good idea to use 4 workers!

.github/workflows/windows_unit_tests.yml
ParthivNaresh: I had it like that previously and for some reason it had failed, changed it back now, thanks!

bchen1116 (approved):
nice!

Makefile
```diff
@@ -20,17 +20,17 @@ lint-fix:
...
+.PHONY: git-test-minimal-deps
+git-test-minimal-deps:
+	pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies
 
-.PHONY: win-circleci-test
-win-circleci-test:
-	pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure -v
+.PHONY: win-git-test
+win-git-test:
+	pytest evalml/ -n 4 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
```
dsherry, line 33: @ParthivNaresh so this was enough to avoid the memory issues on the GH actions box?

I see the runtime of the two GH actions on this PR was about 33min. I also see the runtimes of the windows job on circleci on `main` has ranged from 28min-33min in the recent past. This makes me think by cutting the number of workers we're incurring a slight runtime cost, but its not as bad as I would've guessed. Seems acceptable. Sweet!

dsherry (approved):
Looks great, nice going!

PR #1846, opened 2021-02-15
Workflow - Linux Unit Tests

### LINKED ISSUE #1825: GitHub Actions - Workflow - linux python X.X unit tests
GitHub Actions workflow for linux python X.X unit tests.

Quip Doc: https://alteryx.quip.com/ZdlbAixZ9If3/Workflow-testallpythonversions

ParthivNaresh:
Fixes #1825


chukarsten (approved):

gsheni:
codecov can be finicky and it might be worth disabling as a CI check until it gets merged into main (then re-enabling later on). 
For this branch, the codecov reports are being generated:
https://app.codecov.io/gh/alteryx/evalml/branch/1825-Linux-Python-Unit-Tests

gsheni:
To clarify, I would suggest the following steps.

1. Codecov CI check for `main` has to be disabled by @dsherry. As an example see per below (look at the last 2 uncheck boxes):
Screen Shot 2021-03-01 at 11 40 41 AM

2. Merge this MR into `main`
3. Make sure **codecov** report generates for main at https://app.codecov.io/gh/alteryx/evalml/branch/main
4. Re-enable CodeCov CI check for main.

.github/workflows/linux_unit_tests.yml
```diff
@@ -0,0 +1,79 @@
...
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - python_version: "3.7"
+            core_dependencies: false
+            codecov: false
+          - python_version: "3.8"
+            core_dependencies: false
+            codecov: false
```
dsherry, line 23: @ParthivNaresh this should be `true`, right? Our old CI collects `coverage` report data on python 3.8 both with and without `core_dependencies`. That way, we cover `if` statements which are only checked in one of the two cases, and then `codecov.io` merges the two coverage reports. Does that make sense?

.coveragerc
```diff
@@ -1,5 +1,14 @@
...
     pragma: no cover
-
+    def __repr__
+    raise AssertionError
+    raise NotImplementedError
+    if __name__ == .__main__.:
+    if self._verbose:
+    if verbose:
+    if profile:
+    pytest.skip
```
dsherry, line 14: @ParthivNaresh would you mind explaining why its necessary to add these lines? If they're not necessary, perhaps we can delete this change?

test-requirements.txt
```diff
@@ -3,5 +3,5 @@ pytest-xdist==1.26.1
 pytest-cov==2.6.1
 nbval==0.9.3
 IPython>=5.0.0
-codecov==2.1.0
+codecov==2.1.8
```
dsherry, line 6: 👍

dsherry (changes requested):
@ParthivNaresh cool!! Looking close.

I had one blocking comment: We need to enable `coverage` report for both runs of python 3.8, not just for `core_dependencies=true`. Otherwise a few of our `core_dependencies=false` unit tests won't be included in the combined coverage report.

And one other comment about whether the changes to `.coveragerc` are actually necessary

.github/workflows/linux_unit_tests.yml
ParthivNaresh: @gsheni does the fact that codecov.io merges these reports make it better? Just want to make sure we're all on the same page haha

gsheni: Oh I wasn't aware that CodeCov merges report. yes, that is fine, and @dsherry is correct
- https://docs.codecov.io/docs/merging-reports

.coveragerc
ParthivNaresh: I can absolutely get rid of it, I was trying to emulate the featuretools approach to the .coveragerc file as best as I could but I can delete them if they seem unnecessary in our case

.github/workflows/linux_unit_tests.yml
dsherry: Yeah our CI doesn't make it super obvious that that's what's going on under the hood, lol. Because we just ship both the `core_deps=false` and `core_deps=true` runs over to `codecov.io` and then they handle merging (summing) the reports. But yep, we need both at the moment!

.coveragerc
```diff
@@ -1,5 +1,6 @@
+[run]
+source=evalml
```
dsherry, line 2: @ParthivNaresh what does this do?

dsherry (approved):
@ParthivNaresh thanks!! 🚢 😁

.coveragerc
ParthivNaresh: @dsherry This is [specifying](https://coverage.readthedocs.io/en/coverage-5.0/config.html#run) the source directory that the coverage is going to measure against

Makefile
```diff
@@ -22,11 +22,11 @@ test:
 
 .PHONY: git-test
 git-test:
-	pytest evalml/ -n 8 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
+	pytest evalml/ -n 4 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
```
freddyaboulton, line 25: Did we need to reduce the number of parallel processes because the github vms have less memory?

ParthivNaresh: It seems that having 8 processes causes one of them to crash during execution, and I do think it has something to do with there being less memory. It requires deeper digging though because I don't think it should require all the space it's been given

Issue #1730, opened 2021-01-25
Data checks: recommended actions


dsherry:
TODO: after all issues have been addressed, loop Raymond in.

PR #1855, opened 2021-02-17
Update outliers data check implementation

### LINKED ISSUE #1745: Data Health/Checks: Probability of Outliers
Idea: calculate the probability that no outliers exist in the non-null entries in the data.

```
def no_outlier_prob(num_records: int, pct_outliers: float) -> float:
    """
    This functions calculates the probability that there are no true
    outliers in a numeric (integer or float) field. It is based on creating
    samples 100,000 samples consisting of a given number of records, and
    then repeating this over a grid of sample sizes. Each value in a sample
    is drawn from a log normal distribution, and then the number of
    potential outliers in the data is determined using the skew adjusted box
    plot approach based on the medcouple statistic. It was observed that the
    distribution of the percentage of outliers could be described by a gamma
    distribution, with the shape and scale parameters changing with the
    sample size. For each sample size, the shape and scale parameters of the
    gamma distriubtion were estimated using maximum likelihood methods. The
    set of estimate shape and scale parameters for different sample size were
    then used to fit equations that relate these two parameters to the sample
    size. These equations use a transendental logrithmic functional form that
    provides a seventh order Taylor series approximation to the two true
    functional relationships, and was estimated using least squares
    regression.

    :param num_records: the integer number of non-missing values in a field
    :param pct_outliers: the percentage of potential outliers in a field
    :return: the probability that no outliers are present in the field
    """

    # calculate the shape and scale parameters of the approximate
    # gamma distribution given the number of records in the data.
    # For both measures, the values are are from a least squares regression
    # model
    log_n = np.log(num_records)
    log_shape = (
        25.8218734380722
        + -29.2320460088643 * log_n
        + 14.8228030299864 * log_n ** 2
        + -4.08052512660036 * log_n ** 3
        + 0.641429075842177 * log_n ** 4
        + -0.0571252717322226 * log_n ** 5
        + 0.00268694343911156 * log_n ** 6
        + -5.19415149920567e-05 * log_n ** 7
    )
    shape_param = np.exp(log_shape)
    log_scale = (
        -19.8196822259052
        + 18.5359212447622 * log_n
        + -8.80487628113388 * log_n ** 2
        + 2.27711870991327 * log_n ** 3
        + -0.344443407676357 * log_n ** 4
        + 0.029820831994345 * log_n ** 5
        + -0.00136611527293756 * log_n ** 6
        + 2.56727158170901e-05 * log_n ** 7
    )
    scale_param = np.exp(log_scale)

    # calculate and return the probability of no true outliers for a gamma
    # cumulative density function
    prob_val = 1.0 - gamma.cdf(pct_outliers, shape_param, scale=scale_param)

    return prob_val
```

What we'll probably want to do is implement a data check that checks the feature to ensure that calculated probability of no outliers is above a certain threshold.  To accomplish this we'll need to do the following.

1. Set a threshold, below which, we throw a warning that there might be outliers present in the data.
2. Set a threshold, below which, we throw warnings that there are probably outliers present in the data.
3. Implement the DataCheck class and the .validate(X, y=None) function to check the inputs and return the proper DataCheckWarnings/DataCheckErrors.

angela97lin:
Closes #1745

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -12,7 +12,7 @@ Release Notes
         * Added ``is_bounded_like_percentage`` property for objectives. If true, the ``calculate_percent_difference`` method will return the absolute difference rather than relative difference :pr:`1809`
         * Changed ``TargetEncoder`` to preserve custom indices in the data :pr:`1836`
         * Refactored ``explain_predictions`` and ``explain_predictions_best_worst`` to only compute features once for all rows that need to be explained :pr:`1843`
-        * Updated ``OutliersDataCheck`` implementation to calculate the probability of having no outliers :pr:`1745`
+        * Updated ``OutliersDataCheck`` implementation to calculate the probability of having no outliers :pr:`1855`
     * Fixes
     * Changes
         * Modified ``calculate_percent_difference`` so that division by 0 is now inf rather than nan :pr:`1809`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -19,6 +19,7 @@ Release Notes
     * Documentation Changes
     * Testing Changes
         * Added back coverage for ``_get_feature_provenance`` in ``TextFeaturizer`` after ``text_columns`` was removed :pr:`1842`
+        * Pin graphviz version for windows builds :pr:`1847`
 
 
 **v0.18.2 Feb. 10, 2021**
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -12,6 +12,7 @@ Release Notes
         * Added ``is_bounded_like_percentage`` property for objectives. If true, the ``calculate_percent_difference`` method will return the absolute difference rather than relative difference :pr:`1809`
         * Changed ``TargetEncoder`` to preserve custom indices in the data :pr:`1836`
         * Refactored ``explain_predictions`` and ``explain_predictions_best_worst`` to only compute features once for all rows that need to be explained :pr:`1843`
+        * Updated ``OutliersDataCheck`` implementation to calculate the probability of having no outliers :pr:`1745`
     * Fixes
     * Changes
         * Modified ``calculate_percent_difference`` so that division by 0 is now inf rather than nan :pr:`1809`
--- a/evalml/data_checks/outliers_data_check.py
+++ b/evalml/data_checks/outliers_data_check.py
@@ -6,13 +6,142 @@ from evalml.data_checks import (
     DataCheckWarning
 )
 from evalml.utils import _convert_woodwork_types_wrapper, infer_feature_types
+import numpy as np
+from scipy.stats import gamma
+import sys
+from typing import List, Union
 
+import numpy as np
 
 class OutliersDataCheck(DataCheck):
     """Checks if there are any outliers in input data by using IQR to determine score anomalies. Columns with score anomalies are considered to contain outliers."""
 
     def __init__(self):
         """Checks if there are any outliers in the input data."""
+
+    def no_outlier_prob(num_records: int, pct_outliers: float) -> float:
+        """
+        This functions calculates the probability that there are no true
+        outliers in a numeric (integer or float) field. It is based on creating
+        samples 100,000 samples consisting of a given number of records, and
+        then repeating this over a grid of sample sizes. Each value in a sample
+        is drawn from a log normal distribution, and then the number of
+        potential outliers in the data is determined using the skew adjusted box
+        plot approach based on the medcouple statistic. It was observed that the
+        distribution of the percentage of outliers could be described by a gamma
+        distribution, with the shape and scale parameters changing with the
+        sample size. For each sample size, the shape and scale parameters of the
+        gamma distriubtion were estimated using maximum likelihood methods. The
+        set of estimate shape and scale parameters for different sample size were
+        then used to fit equations that relate these two parameters to the sample
+        size. These equations use a transendental logrithmic functional form that
+        provides a seventh order Taylor series approximation to the two true
+        functional relationships, and was estimated using least squares
+        regression.
+
+        :param num_records: the integer number of non-missing values in a field
+        :param pct_outliers: the percentage of potential outliers in a field
+        :return: the probability that no outliers are present in the field
+        """
+
+        # calculate the shape and scale parameters of the approximate
+        # gamma distribution given the number of records in the data.
+        # For both measures, the values are are from a least squares regression
+        # model
+        log_n = np.log(num_records)
+        log_shape = (
+            25.8218734380722
+            + -29.2320460088643 * log_n
+            + 14.8228030299864 * log_n ** 2
+            + -4.08052512660036 * log_n ** 3
+            + 0.641429075842177 * log_n ** 4
+            + -0.0571252717322226 * log_n ** 5
+            + 0.00268694343911156 * log_n ** 6
+            + -5.19415149920567e-05 * log_n ** 7
+        )
+        shape_param = np.exp(log_shape)
+        log_scale = (
+            -19.8196822259052
+            + 18.5359212447622 * log_n
+            + -8.80487628113388 * log_n ** 2
+            + 2.27711870991327 * log_n ** 3
+            + -0.344443407676357 * log_n ** 4
+            + 0.029820831994345 * log_n ** 5
+            + -0.00136611527293756 * log_n ** 6
+            + 2.56727158170901e-05 * log_n ** 7
+        )
+        scale_param = np.exp(log_scale)
+
+        # calculate and return the probability of no true outliers for a gamma
+        # cumulative density function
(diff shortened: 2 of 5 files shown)


docs/source/release_notes.rst
```diff
@@ -12,6 +12,7 @@ Release Notes
         * Added ``is_bounded_like_percentage`` property for objectives. If true, the ``calculate_percent_difference`` method will return the absolute difference rather than relative difference :pr:`1809`
         * Changed ``TargetEncoder`` to preserve custom indices in the data :pr:`1836`
         * Refactored ``explain_predictions`` and ``explain_predictions_best_worst`` to only compute features once for all rows that need to be explained :pr:`1843`
+        * Updated ``OutliersDataCheck`` implementation to calculate the probability of having no outliers :pr:`1745`
```
angela97lin, line 15: Not sure what we are allowed to say given internals and where this impl came from 👀

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -86,3 +86,39 @@ def test_outliers_data_check_string_cols():
                                       details={"columns": ["d"]}).to_dict()],
         "errors": []
     }
+
+
+def test_outlier_score():
```
angela97lin, line 91: Private but added some tests anyways for sanity sake / code coverage :)

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
+        )
+        scale_param = np.exp(log_scale)
+
+        # calculate and return the probability of no true outliers for a gamma
+        # cumulative density function
+        prob_val = 1.0 - gamma.cdf(pct_outliers, shape_param, scale=scale_param)
+        return prob_val
+
+    @staticmethod
+    def _outlier_score(column, convert_column=False):
```
freddyaboulton, line 132: Why do we need the `convert_column` arg? Looks like we always set it to False

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
+            high_values = column.filter(high_indices).tolist()
+
+            # calculate outlier probability
+            pct_outliers = (len(low_values) + len(high_values)) / len(column_nonan)
+
+            num_records = len(column_nonan)
+            score = OutliersDataCheck._no_outlier_prob(num_records, pct_outliers)
+            result = {
+                "score": score,
+                "values": {
```
freddyaboulton, line 174: I think we should get rid of most of the stuff in `values` since the `validate` method only checks the score. 

That being said, returning at least a sample of `low_indices` and `high_indices` could be useful for #1730 and users to decide what to do about the outliers.

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
-            return pd.concat([lower_bound, upper_bound], axis=1)
-
-        iqr = get_IQR(X, k=2.0)
-        has_outliers = ((X < iqr['lower_bound']) | (X > iqr['upper_bound'])).any()
-        cols = list(has_outliers.index[has_outliers])
-        warning_msg = "Column(s) {} are likely to have outlier data.".format(", ".join([f"'{col}'" for col in cols]))
+        has_outliers = []
+        for col in X.columns:
+            results = OutliersDataCheck._outlier_score(X[col], False)
+            if results is not None and results["score"] <= 0.9:
```
freddyaboulton, line 58: I think we should add a comment justifying the 0.9 threshold.

freddyaboulton (approved):
@angela97lin Looks great! I have some small comments about cleaning up the apis of the private static methods.

evalml/data_checks/outliers_data_check.py
angela97lin: Hehe yah in this data check we set it to false since we only grab numeric cols, this is mostly to maintain the same functionality as the data health tools!

evalml/data_checks/outliers_data_check.py
angela97lin: RE comment above, this is mostly to align with what's being used elsewhere, so the same functionality is available in EvalML when we converge :)

evalml/data_checks/outliers_data_check.py
angela97lin: Will do! Right now, this is a threshold for data 'needing improvement' that was decided, I can annotate this c:

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
+        sample size. For each sample size, the shape and scale parameters of the
+        gamma distriubtion were estimated using maximum likelihood methods. The
+        set of estimate shape and scale parameters for different sample size were
+        then used to fit equations that relate these two parameters to the sample
+        size. These equations use a transendental logrithmic functional form that
+        provides a seventh order Taylor series approximation to the two true
+        functional relationships, and was estimated using least squares
+        regression.
+
+        Original credit goes to Jad Raad and Dan Putler of Alteryx.
```
chukarsten, line 88: Lol I can feel a great weight being lifted from your conscience

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
+
+        Arguments:
+            column (pd.Series): A column of data to check for outliers in
+            convert_column (bool): If True, convert column to np.int64, if possible.
+
+        Returns:
+            dict: Dictionary containing outlier information
+        """
+        column_nonan = column[~pd.isna(column)]
```
chukarsten, line 144: I know this isn't your code, but...
nit: `column_nonan = column.dropna()`

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,136 @@ def validate(self, X, y=None):
...
+
+        if column_nonan.shape[0] == 0:
+            return None
```
chukarsten, line 147: Is this still necessary?  This seems like an implicit redundancy of highly null data check.  If not, it would be nice to snipe it and dedent the rest.

evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -86,3 +86,39 @@ def test_outliers_data_check_string_cols():
...
+    results = OutliersDataCheck._outlier_score(has_outlier, convert_column=True)
+    assert results['score'] != 1.0
+    len(results['values']['high_values']) != 0 or len(results['values']['low_values']) != 0
+    no_outlier = pd.Series(np.arange(10)).astype(object)
+    results = OutliersDataCheck._outlier_score(no_outlier, convert_column=True)
+    assert results['score'] == 1.0
+    len(results['values']['high_values']) == 0 and len(results['values']['low_values']) == 0
+
+
+def test_outlier_score_all_nan():
```
chukarsten, line 122: Yea, I think this might be a bit repetitious as we already have a data check for this.  Do we check for all nans in other DataChecks?

chukarsten (approved):
Nice, way to knock out this initially poorly defined PR ;). I think we might be able to remove the nan check stuff, but I'll leave that up to you.

evalml/data_checks/outliers_data_check.py
angela97lin: I understand what you mean by this and the following comment, but I think this check is still nice to have here. In this method, we're checking if the length of the series is 0 because if it is, all of the other math that continues will error out (ex: np.percentile call) so we want to exit early. It is true that we have the highly null data check that will warn users if their data is all null, but our outlier data check still shouldn't error out if the user passes in data with all nan!

evalml/data_checks/outliers_data_check.py
angela97lin: Indeeeeed :')

evalml/data_checks/outliers_data_check.py
```diff
@@ -50,20 +52,135 @@ def validate(self, X, y=None):
...
+
+            low_values = column.filter(low_indices).tolist()
+            high_values = column.filter(high_indices).tolist()
+
+            # calculate outlier probability
+            pct_outliers = (len(low_values) + len(high_values)) / len(column_nonan)
+
+            num_records = len(column_nonan)
+            score = OutliersDataCheck._no_outlier_prob(num_records, pct_outliers)
+            result = {
```
bchen1116, line 171: Is there a reason we return all of this additional info in the result dic when we don't use anything else besides `result['score']` for the resulting message? Are we trying to use these values in the future for something else?

bchen1116 (approved):
LGTM! Left one question, but nothing blocking

evalml/data_checks/outliers_data_check.py
angela97lin: Yes, keeping these values so when we have everything in one place internally/externally, we can just use the evalml impl :)

PR #1857, opened 2021-02-18
Add Custom Downsampler

### LINKED ISSUE #973: Balanced sampling for classification problems

bchen1116:
part 1 of 2 for #973. 

This PR tracks adding in the custom sampler to EvalML, and we will turn this into a data splitter once merged with the [class imbalance PR](https://github.com/alteryx/evalml/pull/1775/). This does not include the actual data splitter itself.

Functionality of this sampler (and future data splitter) [here](https://alteryx.quip.com/Z1JbAwHFpoU3/Data-Splitting-Notes#aTMACAHdHbG)


evalml/preprocessing/data_splitters/balanced_classification_splitter.py
```diff
@@ -0,0 +1,91 @@
+import numpy as np
+
+from evalml.utils.woodwork_utils import (
+    _convert_woodwork_types_wrapper,
+    infer_feature_types
+)
+
+
+class BalancedClassificationSampler():
```
dsherry, line 9: Missing base class

evalml/preprocessing/data_splitters/balanced_classification_splitter.py
```diff
@@ -0,0 +1,91 @@
...
+        normalized_counts = y.value_counts(normalize=True)
+        class_ratios = normalized_counts.copy().values
+        class_ratios /= class_ratios[-1]
+        # if no class ratios are greater than what we consider balanced, then the target is balanced
+        if all(class_ratios <= self.balanced_ratio):
+            return {}
+        # if any classes have than min_samples counts and are less than min_percentage of the total data,
+        # then it's severely imbalanced
+        if any(counts < self.min_samples) and any(normalized_counts < self.min_percentage):
+            return {}
```
freddyaboulton, line 61: Playing devil's advocate here, should severely imbalanced and balanced data produce the same behavior?

The user's data can fall into four cases:
1. Balanced
2. Severely Imbalanced
3. Imbalanced and can downsample
4. Imbalanced but can't downsample because of `min_samples`

Right now users can't tell whether they are in cases 1, 2, or 4 and I think there is value in communicating that. Maybe it belongs in our class imbalance data check though.

evalml/tests/preprocessing_tests/test_balanced_classification_splitter.py
```diff
@@ -0,0 +1,342 @@
...
+    pd.testing.assert_series_equal(y1, y2)
+
+
+@pytest.mark.parametrize("index", [[f'hello_{i}' for i in range(1000)],
+                                   random.shuffle([i + 0.5 for i in range(1000)]),
+                                   pd.MultiIndex.from_arrays([
+                                       [f"index_{i}" for i in range(1000)],
+                                       [i for i in range(1000)]
+                                   ])])
+def test_classification_imbalanced_custom_indices(index):
```
freddyaboulton, line 240: 👍

evalml/preprocessing/data_splitters/balanced_classification_splitter.py
```diff
@@ -0,0 +1,91 @@
...
+                If we don't need to resample, returns empty dictionary.
+        """
+        counts = y.value_counts()
+        normalized_counts = y.value_counts(normalize=True)
+        class_ratios = normalized_counts.copy().values
+        class_ratios /= class_ratios[-1]
+        # if no class ratios are greater than what we consider balanced, then the target is balanced
+        if all(class_ratios <= self.balanced_ratio):
+            return {}
+        # if any classes have than min_samples counts and are less than min_percentage of the total data,
```
freddyaboulton, line 58: typo: "have less than min_samples"?

freddyaboulton (approved):
@bchen1116 I think this is great! The tests are very solid.

Nothing blocking from me.

evalml/preprocessing/data_splitters/balanced_classification_splitter.py
bchen1116: @freddyaboulton Yeah, I think in discussion with @dsherry, we needed to figure out how to get the class imbalance data check to work with this. Definitely agree that the behavior could be confusing, but I think as long as the data check can help us catch the severely-imbalanced case and warn the user, it clarifies the case.

bchen1116: @freddyaboulton just filed a new [issue](https://github.com/alteryx/evalml/issues/1864) to track updating the class imbalance data check

chukarsten (approved):
Very nice job - super thorough testing.  Love it.  I had a comment that I don't *think* is merge blocking that we should probably resolve (or it is resolved already) concerning what to do when a class cannot be downsampled to the desired ratio.

evalml/tests/preprocessing_tests/test_balanced_classification_splitter.py
```diff
@@ -0,0 +1,343 @@
...
+    assert y2.value_counts()[1] == 100
+
+
+def test_classification_data_drop():
+    # tests for whether or not the `max(0, counts[k] - goal_value)` code works as expected
+    X = pd.DataFrame({"a": [i for i in range(420)]})
+    y = pd.Series([0] * 90 + [1] * 100 + [2] * 120 + [3] * 40 + [4] * 70)
+    # will downsample the [2] target
+    # will try to downsample [0] and [4], but max(0, x) will prevent that
+    bcs = BalancedClassificationSampler(balanced_ratio=1, min_percentage=0.01)
```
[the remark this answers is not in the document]
bchen1116: Mmm, yeah that's fair. My thought was that we cannot sample if we end up with less than `min_samples` values of a class. For instance, if we have 40 instances of class 1 and 100 instances of class 2, we might not want to sample the 100 instances down even if we had wanted a 1:1 ratio, whereas in a case of 40,000 vs 100,000 samples, it would be fine to sample the 100,000 to 40,000 cases since we have enough samples. As for warning or telling the user, I wasn't sure if this was something I should warn the user about or do silently. Thoughts @chukarsten @dsherry?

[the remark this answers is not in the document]
dsherry: I think the understanding should be that if the number of instances falls below the threshold for certain classes, as it does for "0" and "4" here, that the method will simply not downsample those classes. No error in that case.

PR #1870, opened 2021-02-22
Workflow - Build Conda Package

### LINKED ISSUE #1824: GitHub Actions - Workflow - build_conda_pkg
GitHub Actions workflow for build_conda_pkg.

Quip Doc: https://alteryx.quip.com/YbZzAp4EURNj/Workflow-buildcondapkg

ParthivNaresh:
Fixes #1824


evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -23,7 +23,7 @@ scikit-optimize==0.8.1
 scipy==1.6.2
 seaborn==0.11.1
 shap==0.39.0
-sktime==0.5.3
+sktime==0.6.0
```
jeremyliweishih, line 26: is this dependency change for this PR?

jeremyliweishih: oo this was for the dependency test right? I just merged in the PR so this isn't needed anymore 😄

ParthivNaresh: No this was unrelated but needed to pass the dependency check, sktime had a release a few hours ago and it doesn't break any of our code but I thought I should include it for bug fixes etc.

ParthivNaresh: ^^Ignore this haha I didn't see your next comment

.github/workflows/build_conda_pkg.yml
```diff
@@ -0,0 +1,66 @@
+name: Build Conda Package
+
+on:
+  pull_request:
```
freddyaboulton, line 4: I think we only want this to run on pushes to main

freddyaboulton (approved):
@ParthivNaresh I think this looks good!

.github/workflows/build_conda_pkg.yml
ParthivNaresh: @freddyaboulton Currently `build_conda_pkg` runs in CircleCI as part of the other workflows, so on every commit, primarily for the `shellcheck`. Are we changing this behaviour @dsherry ?

freddyaboulton: Ah, I think I missed `if: github.ref == 'refs/head/main'` line below. If that means that the package is only built on commits to main, then I think we're good to go!

But I think we can make it simpler by using `on: push: branches: -main`? I don't think we need the shellcheck to run on PR branches but let me know if I'm misunderstanding!

ParthivNaresh: Got it, I was under the impression we were keeping the same behaviour as the CircleCI implementation but if we're good to not do a shellcheck on PR branches then I can definitely make the change to only run on `main`. I agree that I don't think it's necessary.

dsherry: Sounds good to me!

I also wonder if shellcheck would belong better in our `lint` jobs.

.github/workflows/build_conda_pkg.yml
```diff
@@ -0,0 +1,63 @@
+name: Build Conda Package
+
+on:
+  push:
+    branches:
+      - main
```
dsherry, line 6: @ParthivNaresh could you please add
```
  pull_request:
    types: [opened, synchronize]
```
and show that this action runs green on this PR before merge?

We've used that syntax in other actions, and I think it means the job will run in this PR but not in other PRs against `main`.

dsherry (changes requested):
Looks great! One minor change requested, to prove that the action is green

angela97lin (approved):
Goooodbye codecov!! LGTM 🥂

.github/workflows/build_conda_pkg.yml
dsherry: Hey, my bad, I clearly do not know enough of the GH actions syntax yet 😆  I was hoping there was a way to run the action in this PR just to prove that it works before merging. If this is too difficult to do (which I think is the case after 5min of googling), no worries, let's merge it and revert if there are problems.

dsherry (approved):
🚢 😁

PR #1874, opened 2021-02-22
Fix Binary Classification Pipeline predict method does not use objective if it is passed as a named argument

### LINKED ISSUE #1819: Binary Classification Pipeline  predict method does not use objective if it is passed as a named argument
Repro:

```python
from evalml.objectives import BinaryClassificationObjective
import pandas as pd
from unittest.mock import MagicMock


class Pipeline(BinaryClassificationPipeline):


X, y = load_breast_cancer()


class CustomObjective(BinaryClassificationObjective):
    greater_is_better = False
    name = "Foo Objective"
    perfect_score = 0
    score_needs_proba = False
    
    def objective_function(self, y_true, y_predicted, X=None):
        pass
    
    def decision_function(self, ypred_proba, threshold, X):
        # The decoder will map 0 to benign so we should expect the
        # predictions to only contain "benign"
        return pd.Series([0] * len(ypred_proba))


pipeline = Pipeline({})
pipeline.fit(X, y)
pipeline._threshold = 0.3
expected_preds = ['benign'] * len(y)

assert not pipeline.predict(X, objective=CustomObjective()).to_series().tolist() == expected_preds
```

The strange thing is that it works as expected if the objective is not used as a kwarg:

```python
assert pipeline.predict(X, CustomObjective()).to_series().tolist() == expected_preds
```

The problem is in the pipeline meta class. If the objective is passed as a kwarg, then `y=None` and we [don't pass the objective to the method](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/pipeline_base_meta.py#L25).

angela97lin:
Closes #1819 

Also fixes issue with time series pipelines and feature_importances. In `main`, `input_feature_names` does not have the estimator, so calling `feature_importances` errors out. This is because time series pipelines call `_fit_transform_features_helper` with all components but the last.


evalml/pipelines/component_graph.py
```diff
@@ -146,6 +146,8 @@ def _fit_transform_features_helper(self, needs_fitting, X, y=None):
                 parent_output = infer_feature_types(parent_output)
             final_component_inputs.append(parent_output)
         concatted = pd.concat([component_input.to_dataframe() for component_input in final_component_inputs], axis=1)
+        if needs_fitting:
```
angela97lin, line 149: since we call `self._compute_features(self.compute_order[:-1]...`, we miss out on the last element (usually estimator) in `self.input_feature_names`. Manually adding here.

evalml/pipelines/classification_pipeline.py
```diff
@@ -88,7 +88,7 @@ def predict(self, X, objective=None):
         Returns:
             ww.DataColumn: Estimated labels
         """
-        predictions = self._predict(X, objective).to_series()
+        predictions = self._predict(X, objective=objective).to_series()
```
angela97lin, line 91: No difference in behavior, just allows us to use in kwargs for testing c:

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -123,7 +124,7 @@ def test_make_pipeline_all_nan_no_categoricals(input_type, problem_type):
         y = pd.Series([0, 2, 1, 2])
 
     for estimator_class in estimators:
-        for problem_type in estimator_class.supported_problem_types:
```
angela97lin, line 126: This is no longer valid because if the fixture `problem_type` was time series then `pipeline_class = _get_pipeline_base_class(problem_type)` will now have a type of `TimeSeriesPipelineBaseMeta `. If the `problem_type` in this for loop (duplicate names, yikes) is binary, then pipeline will be of type PipelineBaseMeta and the `assert isinstance(pipeline, type(pipeline_class))` line will fail.

evalml/pipelines/pipeline_meta.py
```diff
@@ -25,9 +46,8 @@ def _check_for_fit(self, X=None, y=None, objective=None):
             elif y is None:
                 return method(self, X)
             # For time series classification pipelines, predict will take X, y, objective
-            elif len(inspect.getfullargspec(method).args) == 4:
+            elif method.__name__ == 'predict':
```
jeremyliweishih, line 49: If adding this case solves the problem for time series problems could we just add this case to `PipelineBaseMeta`? I think its better organizationally to use inheritance but since you can't inherit the inner function, the current implementation has some code reuse.

```
@wraps(method)
        def _check_for_fit(self, X=None, y=None, objective=None):
            klass = type(self).__name__
            if not self._is_fitted:
                raise PipelineNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
           elif klass == '...` and method.__name__ == 'predict':
                return method(self, X, y, objective)
            elif X is None and y is None:
                return method(self)
            elif y is None:
                return method(self, X)
            else:
                return method(self, X, y)
```

evalml/pipelines/component_graph.py
jeremyliweishih: Can you help me understand how this ties into the issue with the metaclasses?

evalml/tests/pipeline_tests/test_pipelines.py
jeremyliweishih: These tests still iterate through all problem types and their estimators right? If so this should be good.

jeremyliweishih (approved):
looks good! Just provided a possible alternative to implementation but tests look great.

evalml/pipelines/component_graph.py
angela97lin: It doesn't--it was a bug I noticed while writing the tests and trying to call `feature_importance` on time series pipelines 😅

evalml/pipelines/pipeline_meta.py
angela97lin: I'm down for either! You're right, maybe since we only have one method right now its not worth inheritance... Took the suggestion to subclass via @freddyaboulton and @dsherry 's discussion on the issue. Do you guys have thoughts about this?

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Si, si, fixture goes through all of problem types and then we get all the estimators for the problem type!

evalml/pipelines/pipeline_meta.py
freddyaboulton: @angela97lin I think the apis for time series vs non-time series pipelines are different enough that trying to handle them both in one metaclass makes the implementation pretty tricky to understand/maintain.

I think having the separate metaclasses makes the distinction between time-series vs non-time-series clear and should (hopefully) make the meta class implementations easier to understand.

To that end, this is what I was thinking we'd do when we wrote up the issue:

1. Delete `y` from the `PipelineBaseMeta` `_check_for_fit` helper since `y` is never allowed for non time series pipelines.
2. Rather than switching based on whether `X` and/or `y` are None, switch on the method name like you did for predict `if method.__name__ == 'predict':``elif method.__name__ == 'feature_importance`

I don't have a strong opinion about whether we should inherit from `PipelineBaseMeta` or `BaseMeta` since the work is being done by the inter `_check_for_fit` function which can't be shared.

Let me know what you think!

angela97lin: @freddyaboulton Thank you for explaining your thoughts! I like your suggestions a lot.

- I agree with the idea of having them as two separate metaclasses. I agree that it makes it a bit more clear in terms of logic and implementation.
- I deleted `y` from the `PipelineBaseMeta`
- RE using names: I decided to update to check via name (though `method.__name__ == 'feature_importance'` ends up being `method.__name__ == '__get__'` since feature importance is an attribute, so I threw it in the `else` case). I do think for now, this makes things clear because we're only checking a few methods/attributes (predict_proba, predict, feature_importance). I do wonder if this will be maintainable once we add more methods to check, but we might never get there so I'm all for readability! 🥳

freddyaboulton: @angela97lin Awesome! I didn't know that the "method" passed to the meta class would be `__get__` for properties. I think it's ok to catch all properties in the `else` clause because `method(self)` should work for all properties because they don't take arguments!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1634,22 +1634,81 @@ def test_pipeline_not_fitted_error(mock_fit, problem_type, X_y_binary, X_y_multi
...
+        with patch(to_patch) as mock_predict:
             clf.predict(X)
             mock_predict.assert_called()
-        with patch('evalml.pipelines.ClassificationPipeline.predict_proba') as mock_predict_proba:
-            clf.predict_proba(X)
-            mock_predict_proba.assert_called()
+            _, kwargs = mock_predict.call_args
+            assert kwargs['objective'] is None
+
+            clf.predict(X, 'Log Loss Binary')
```
freddyaboulton, line 1653: Can you do a case where the objective is passed as a kwarg?

freddyaboulton (approved):
@angela97lin Looks great!

evalml/pipelines/pipeline_meta.py
```diff
@@ -16,18 +15,34 @@ def check_for_fit(cls, method):
...
+        """
+        @wraps(method)
+        def _check_for_fit(self, X=None, y=None, objective=None):
+            klass = type(self).__name__
+            if not self._is_fitted:
+                raise PipelineNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
+            if method.__name__ == 'predict_proba':
                 return method(self, X, y)
+            elif method.__name__ == 'predict':
+                return method(self, X, y, objective)
```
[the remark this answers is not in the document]
angela97lin: Ah, interestingly enough, the feature importance case is covered by `PipelineBaseMeta`, not `TimeSeriesPipelineBaseMeta`, so the `else` case is never reached if I have it (codecov caught this). I think this is because the implementation for feature importance lies in `PipelineBase` and isn't overridden by time series pipelines (confirmed via overriding time series pipelines' feature importance).

evalml/pipelines/utils.py
```diff
@@ -138,7 +138,7 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
     hyperparameters = custom_hyperparameters
     base_class = _get_pipeline_base_class(problem_type)
 
-    class GeneratedPipeline(base_class):
+    class GeneratedPipeline(base_class, metaclass=type(base_class)):
```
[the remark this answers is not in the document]
angela97lin: Oops, this wasn't necessary after all hehe reverting

Issue #1822, opened 2021-02-11
Deprecate the explain_prediction function


freddyaboulton:
Once #1818 is merged, the following two function calls will be the same:

```python
explain_prediction(pipeline, input_features, y, index_to_explain=3)
```

```python
explain_predictions(pipeline, input_features, y, indices_to_explain=[3])
```

I think this makes `explain_prediciton` redundant.

dsherry:
Yes!

dsherry:
I think we should follow our pattern for deprecation:
* This issue tracks adding a deprecation warning now, and filing an issue to delete `explain_prediction` altogether
* We'll schedule that second issue for the following release

I suspect it would be ok to simply delete this method, but I think its good for us to stay in the habit of giving advance notice for deprecations.

PR #1885, opened 2021-02-23
Pre-v0.19.0 Doc Update

chukarsten:
With the completion of #1822 , which deprecated the explain_prediction() function, it introduced a ton of ugly warnings into the docs (https://feature-labs-inc-evalml--1874.com.readthedocs.build/en/1874/demos/fraud.html). To get around this, I threw some warning filters into the jupyter notebooks, specifically three of the demos and the model understanding one.  I don't foresee this being a long term solution....but given that this cropped up mid release, I'm applying a bandaid.


jeremyliweishih (changes requested):
This looks good but in the docs you posted I see a lot of `Argument 'random_state' has been deprecated in favor of 'random_seed'. Passing 'random_state' in future versions will result in an error.` as well. Could we remove all mentions of `random_state` from the notebooks as well?

freddyaboulton (changes requested):
Good catch @chukarsten !

To get rid of the warnings that come up during search, I think we should use `random_seed` instead of `random_state` in the `EngineBase` [train_and_score_pipeline](https://github.com/alteryx/evalml/blob/main/evalml/automl/engine/engine_base.py#L93) method. 

I think the other warnings can be removed with using `random_seed` instead of `random_state` in the docs in the calls to `split_data`. 

The only warnings we can't get rid of are the deprecation warnings regarding `explain_prediction`. I think those would be fine to keep to warn about the coming deprecation but maybe we can pass `once` instead of `ignore` to the warnings filter so they only come up once.

dsherry (approved):
👍

chukarsten:
Closing this in favor of a new PR based off main. #1889

PR #1894, opened 2021-02-25
Add ARIMARegressor

### LINKED ISSUE #1498: Time series regression: Add ARIMA estimator
[Statsmodels](https://www.statsmodels.org/stable/generated/statsmodels.tsa.arima_model.ARIMA.html) is probably the right implementation to use.

jeremyliweishih:
Fixes #1498.


ParthivNaresh:
Updating docs, API reference, etc as part of #1946. Let me know if it's necessary to include it here instead.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -0,0 +1,151 @@
...
+    a_clf = arima.ARIMA(endog=y, exog=X, order=(1, 0, 0), dates=X.index)
+    a_clf.fit(solver='nm')
+    y_pred_a = a_clf.predict(params=(1, 0, 0))
+
+    X = X.reset_index()
+    clf = ARIMARegressor(p=1, d=0, q=0, date_column='index')
+    clf.fit(X=X, y=y)
+    y_pred = clf.predict(X=X, y=y)
+
+    assert (y_pred == y_pred_a).all()
```
bchen1116, line 151: If you decide to leave `get_dates_fit` and `get_dates_predict` as public methods, can you add some tests for that?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+                 random_seed=0, random_state=None, **kwargs):
+        random_seed = deprecate_arg("random_state", "random_seed", random_state, random_seed)
+        order = (p, d, q)
+        parameters = {'order': order}
+        parameters.update(kwargs)
+        self.date_column = date_column
+
+        p_error_msg = "ARIMA is not installed. Please install using `pip install statsmodels`."
+        arima = import_or_raise("statsmodels.tsa.arima_model", error_msg=p_error_msg)
+        arima.ARIMA(endog=np.zeros(p + d + q + 1), **parameters)  # to validate parameters
```
bchen1116, line 37: What happens if this passes or fails? Do we have tests for this?

bchen1116 (changes requested):
Just left some comments and suggestions on the code! I think `get_dates_fit` and `get_dates_predict` should be private methods, but if we choose to leave them public, it'd be nice to have some tests for those methods.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+        return self
+
+    def predict(self, X, y=None):
+        X, y = self._manage_woodwork(X, y)
+        dates = self.get_dates_predict(X, y)
+        start = dates.min()
+        end = dates.max()
+        params = self.parameters['order']
+        if X is not None:
+            y_pred = self._component_obj.predict(params=params, start=start, end=end, exog=X)
```
freddyaboulton, line 104: In the statsmodel docs they use the ARIMA object slightly differently. Basically they call `fit` to get a "results object" and then call `predict` from the "results object" and not the original ARIMA object. [Example](https://www.statsmodels.org/devel/examples/notebooks/generated/statespace_forecasting.html#:~:text=They%20are%20predict%20and%20get_prediction,results%20(similar%20to%20get_forecast%20).).

I think we should follow that approach unless there's a good reason not too!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+        self.date_column = date_column
+
+        p_error_msg = "ARIMA is not installed. Please install using `pip install statsmodels`."
+        arima = import_or_raise("statsmodels.tsa.arima_model", error_msg=p_error_msg)
+        arima.ARIMA(endog=np.zeros(p + d + q + 1), **parameters)  # to validate parameters
+        super().__init__(parameters=parameters,
+                         component_obj=None,
+                         random_seed=random_seed)
+
+    def get_dates_fit(self, X, y):
```
freddyaboulton, line 42: `get_dates_fit` is doing two things: getting the dates from the column or index and then changing the index of `X` to match that of `y`. Can we you break up those two things into different functions? I think it'll be easier to follow the logic at that point. Also, once we're there I think we'll be able to reuse the method for getting the dates in `predict` and get rid of `get_dates_predict`?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+    name = "ARIMA Regressor"
+    hyperparameter_ranges = {
+        "p": Integer(0, 10),
+        "d": Integer(0, 10),
+        "q": Integer(0, 10),
+    }
+    model_family = ModelFamily.ARIMA
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, date_column=None, p=1, d=0, q=0,
```
freddyaboulton, line 27: Can you please add `trend` as a parameter/hyperparameter? [Source](https://www.statsmodels.org/v0.12.0/generated/statsmodels.tsa.arima.model.ARIMA.html#statsmodels.tsa.arima.model.ARIMA) I think supporting the string values of n, c, t, and ct is enough for now. I think it would be great if this component was as self-contained as possible!

freddyaboulton (changes requested):
@ParthivNaresh This is great! I have some comments that I think should be addressed before merge but the overall this is solid!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+                 random_seed=0, random_state=None, **kwargs):
+        random_seed = deprecate_arg("random_state", "random_seed", random_state, random_seed)
+        order = (p, d, q)
+        parameters = {'order': order}
+        parameters.update(kwargs)
+        self.date_column = date_column
+
+        p_error_msg = "ARIMA is not installed. Please install using `pip install statsmodels`."
+        arima = import_or_raise("statsmodels.tsa.arima_model", error_msg=p_error_msg)
```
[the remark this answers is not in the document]
ParthivNaresh: Great catch!

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -0,0 +1,151 @@
...
+
+    a_clf = arima.ARIMA(endog=y, exog=X, order=(1, 0, 0), dates=X.index)
+    a_clf.fit(solver='nm')
+    y_pred_a = a_clf.predict(params=(1, 0, 0))
+
+    clf = ARIMARegressor(p=1, d=0, q=0)
+    clf.fit(X=X, y=y)
+    y_pred = clf.predict(X=X, y=y)
+
+    assert (y_pred == y_pred_a).all()
```
chukarsten, line 41: Do you think this is parameterizable?  (If yes, feel free to come to my #1990 PR and tell me my tests are parameterizable too ;)

evalml/tests/component_tests/test_arima_regressor.py
chukarsten: Isn't this functionality (looking at a datatable/frame of features and determining date time columns) used somewhere else?  If you go the route of public methods, would definitely like an issue to refactor those other estimators or whatnot that use that logic.  @freddyaboulton do you use this in time series?

chukarsten (approved):
I agree with the comments left by @freddyaboulton and @bchen1116 , but think once those issues are addressed this is good to go.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,114 @@
...
+    }
+    model_family = ModelFamily.ARIMA
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, date_column=None, p=1, d=0, q=0,
+                 random_seed=0, random_state=None, **kwargs):
+        random_seed = deprecate_arg("random_state", "random_seed", random_state, random_seed)
+        order = (p, d, q)
+        parameters = {'order': order}
+        parameters.update(kwargs)
```
jeremyliweishih, line 32: From my understanding the tuner/algorithm will send in `p`, `d`, and `q` in as keys and values in the `parameters` dict. So if we pop `p`, `d`, and `q` here we can avoid passing it into the `component_obj` and only pass in `order` which it takes.

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -663,7 +663,7 @@ def test_jupyter_graph_check(import_check, jupyter_check, X_y_binary, X_y_regres
     jupyter_check.return_value = True
     with pytest.warns(None) as graph_valid:
         graph_partial_dependence(clf, X, features=0, grid_resolution=20)
-        assert len(graph_valid) == 1  # scikit-learn partial_dependence warning
+        assert len(graph_valid) == 1
```
angela97lin, line 666: Indifferent about keeping this comment here so we know where this warning is coming from (aka not us) /shrug

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -71,7 +71,7 @@ def test_all_estimators(has_minimal_dependencies):
     if has_minimal_dependencies:
         assert len((_all_estimators_used_in_search())) == 10
     else:
-        assert len(_all_estimators_used_in_search()) == 16
+        assert len(_all_estimators_used_in_search()) == 17
```
angela97lin, line 74: My understanding is that this PR should only be adding the estimator, *not* to search which is handled by #1946? If that's true, this shouldn't be updated, correct?

evalml/tests/component_tests/test_estimators.py
```diff
@@ -114,6 +116,8 @@ def test_estimator_predict_output_type(X_y_binary, helper_functions):
                        (X_df_with_col_names, y_series_with_name, X_df_with_col_names.columns, y_series_with_name.unique())]
 
     for component_class in _all_estimators_used_in_search():
+        if component_class.__name__ == 'ARIMARegressor':
```
angela97lin, line 119: Same RE adding ARIMA to `_all_estimators_used_in_search`; we should make sure it's *not* added as an estimator for search, rather than having our tests skip over it

evalml/tests/component_tests/test_components.py
```diff
@@ -1067,15 +1075,22 @@ def test_estimator_fit_respects_custom_indices(use_custom_index, estimator_class
...
+        X, y = ts_data
+        ts_problem = True
     else:
         X, y = X_y_binary
 
     X = pd.DataFrame(X)
     y = pd.Series(y)
 
-    if use_custom_index:
+    if use_custom_index and ts_problem:
```
angela97lin, line 1090: Rather than needing an extra variable, we could just check if `ProblemTypes.TIME_SERIES_REGRESSION in estimator.supported_problem_types` (don't remember syntax)?

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def test_feature_importance(ts_data):
+    X, y = ts_data
+    clf = ARIMARegressor()
+    clf.fit(X, y)
+    clf.feature_importance == np.zeros(1)
+
+
+def test_fit_predict_ts_with_Xandy_index(ts_data):
```
angela97lin, line 28: nit pick, can we call this test_fit_predict_ts_with_X_and_y_index lol

evalml/tests/component_tests/test_components.py
```diff
@@ -568,7 +569,7 @@ def test_default_parameters(cls):
     assert cls.default_parameters == cls().parameters, f"{cls.__name__}'s default parameters don't match __init__."
 
 
-@pytest.mark.parametrize("cls", [cls for cls in all_components() if cls not in [StackedEnsembleRegressor, StackedEnsembleClassifier]])
+@pytest.mark.parametrize("cls", [cls for cls in all_components() if cls not in [StackedEnsembleRegressor, StackedEnsembleClassifier, ARIMARegressor]])
```
angela97lin, line 572: Just wondering, why does ARIMA get the same treatment here?

evalml/tests/component_tests/test_arima_regressor.py
angela97lin: same with the rest of the tests, I think it helps with readability a bit 😂

angela97lin (changes requested):
Added some nit-picky comments, but the biggest thing before merge is, I don't think we should be adding this to `_all_estimators_used_in_search`?

docs/source/release_notes.rst
```diff
@@ -33,6 +33,7 @@ Release Notes
         * Added a GitHub Action for Detecting dependency changes :pr:`1933`
         * Create a separate CV split to train stacked ensembler on for AutoMLSearch :pr:`1814`
         * Added a GitHub Action for Linux unit tests :pr:`1846`
+        * Added ARIMARegressor :pr:`1894`
```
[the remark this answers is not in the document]
angela97lin: (this is more about the `` code `` heh)

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+    name = "ARIMA Regressor"
+    hyperparameter_ranges = {
+        "p": Integer(0, 10),
+        "d": Integer(0, 10),
+        "q": Integer(0, 10),
+    }
+    model_family = ModelFamily.ARIMA
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, date_column=None, trend='n', p=1, d=0, q=0,
```
angela97lin, line 27: Can we add docstrings for these parameters?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+            date_col = y.index
+        if X is not None:
+            if self.date_column in X.columns:
+                date_col = X.pop(self.date_column)
+            elif isinstance(X.index, pd.DatetimeIndex):
+                date_col = X.index
+
+        if date_col is None:
+            msg = "ARIMA regressor requires input data X to have a datetime column specified by the 'date_column' parameter. " \
+                  "If not it will look for the datetime column in the index of X or y."
```
angela97lin, line 64: To confirm this error message vs the one below, this will look for the datetime in "the index of X or y", but predict will only look at X? :)

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+        if y is not None:
+            y.index = date_col
+        return X, y
+
+    def fit(self, X, y=None):
+        if y is None:
+            raise ValueError('ARIMA Regressor requires y as input.')
+
+        p_error_msg = "ARIMA is not installed. Please install using `pip install statsmodels`."
+        arima = import_or_raise("statsmodels.tsa.arima.model", error_msg=p_error_msg)
```
angela97lin, line 98: Do we need this in `fit` if we have it in `__init__`?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+        new_params = {}
+        for key, val in self.parameters.items():
+            if key not in ['p', 'd', 'q']:
+                new_params[key] = val
+        if X is not None:
+            arima_with_data = arima.ARIMA(endog=y, exog=X, dates=dates, **new_params)
+        else:
+            arima_with_data = arima.ARIMA(endog=y, dates=dates, **new_params)
+
+        component_obj = arima_with_data
```
angela97lin, line 112: Do we need this line? I think this is fine?
```
self._component_obj = arima_with_data.fit()
return self
```

evalml/tests/model_understanding_tests/test_graphs.py
freddyaboulton, line 666: Nice!

evalml/tests/component_tests/test_estimators.py
```diff
@@ -114,6 +116,8 @@ def test_estimator_predict_output_type(X_y_binary, helper_functions):
                        (X_df_with_col_names, y_series_with_name, X_df_with_col_names.columns, y_series_with_name.unique())]
 
     for component_class in _all_estimators_used_in_search():
```
freddyaboulton, line 118: I think we should add Arima to the list of components not allowed in AutoMLSearch for the time being!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
freddyaboulton, line 27: I've been thinking about whether expecting dates to be specified in the constructor is a good idea when it comes to `AutoMLSearch` integration. I think it's good! The one downside of this approach is that we can't use a DateTimeFeaturizer in the same pipeline as Arima because the DateTimeFeaturizer will drop the original date column. But I think that's ok since we can pass the date info as the index as well.

evalml/tests/component_tests/test_components.py
freddyaboulton, line 572: I don't think ARIMARegressor needs to be in the list of excluded component classes, i.e. the arima default parameters should not raise a warning.

freddyaboulton (approved):
@ParthivNaresh Thank you for making the changes! I think we need to make sure Arima is not used in automl search for the time being but other than that this is good to go!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
ParthivNaresh: Well technically because the `predict` function wasn't going to use `y` I changed that, but since it does check `y` for a datetime index, I'll change it back again, thanks!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
ParthivNaresh: Complete oversight, thanks!

ParthivNaresh: We end up using it a few lines below when we initialize the ARIMA model in different ways depending on if X is passed or not :)

evalml/tests/model_understanding_tests/test_graphs.py
ParthivNaresh: There was initially a warning being displayed I believe, by changing the `ARIMA` implementation to the non-deprecated one, it removed another warning. So it went from `len(graph_valid) == 2` to `len(graph_valid) == 1`

bchen1116 (approved):
Great work! I like the changes made!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -0,0 +1,145 @@
...
+        "d": Integer(0, 10),
+        "q": Integer(0, 10),
+    }
+    model_family = ModelFamily.ARIMA
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+
+    def __init__(self, date_column=None, trend='n', p=1, d=0, q=0,
+                 random_seed=0, **kwargs):
+        """
+        Arguments:
```
angela97lin, line 30: 👏 👏 👏

evalml/pipelines/components/estimators/regressors/arima_regressor.py
angela97lin: Ah what I meant was, is there a case where we catch this in `fit` when we didn't catch it in `__init__`? Can a user use `fit` without init? I'm assuming the use case is:
```
arima = ARIMARegressor(...) # will catch import error
arima.fit(...) # will we ever catch import error here?
```

evalml/utils/gen_utils.py
```diff
@@ -159,7 +159,7 @@ def _get_subclasses(base_class):
                        'BaselineRegressionPipeline', 'ModeBaselineMulticlassPipeline', 'BaselineMulticlassPipeline',
                        'TimeSeriesBaselineRegressionPipeline', 'TimeSeriesBaselineBinaryPipeline',
                        'TimeSeriesBaselineMulticlassPipeline', 'KNeighborsClassifier',
-                       'SVMClassifier', 'SVMRegressor'}
+                       'SVMClassifier', 'SVMRegressor', 'ARIMARegressor'}
```
angela97lin, line 162: 👍 👍 👍

angela97lin (approved):
Wanted to follow up about needing the import_or_raise statement in `fit` but otherwise, looks good to go! Thanks for making the changes, this looks really good 🥳

PR #1896, opened 2021-02-25
Add DataCheckAction class and DataCheckActionCode enum

### LINKED ISSUE #1878: Add `DataCheckAction` class

angela97lin:
Closes #1878


evalml/data_checks/data_check_action.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __eq__(self, other):
+        """Checks for equality. Two DataCheckAction objs are considered equivalent if all of their attributes are equivalent."""
+        return (self.action_code == other.action_code and
+                self.details == other.details)
+
+    def to_dict(self):
+        action_dict = {"code": self.action_code.name}
+        if self.details is not None:
+            action_dict.update({"details": self.details})
```
freddyaboulton, line 23: I think it's better if `details` is always present and just set to an empty dict if there are no details.

Just so that users don't have to check if there are details before trying to access them.

evalml/data_checks/data_check_action.py
```diff
@@ -0,0 +1,24 @@
+class DataCheckAction:
+    """Base class for all DataCheckActions."""
+
+    def __init__(self, action_code, details=None):
+        """
+        A recommended action returned by a DataCheck.
+
+        Arguments:
+            action_code (DataCheckActionCode): Action code associated with the action.
```
freddyaboulton, line 9: Since the action recommendation process will be interactive, I think it will be easier for users to modify the actions if they can specify the action code via string rather than having to import the `DataCheckActionCode`, similar to what we do for problem types. 

```python
data_check = DefaultDataChecks()
actions = data_check.validate(X, y)["actions"]
actions[0].action_code = "do_nothing"
```

as opposed to

```python
from evalml.datachecks import DataCheckActionCode
data_check = DefaultDataChecks()
actions = data_check.validate(X, y)["actions"]
actions[0].action_code = DataCheckActionCode.DO_NOTHING
```

Let me know what you think. This is out of scope for this PR but I think we should make the actions easy to modify as we build them.

freddyaboulton (approved):
@angela97lin I think this is great!

bchen1116 (approved):
Looks good to me! I agree with Freddy's suggestions though, it would be nice to default to the empty dict and return the string for the user.

evalml/data_checks/data_check_action.py
angela97lin: Filed #1904 :)

PR #1900, opened 2021-02-25
Update Woodwork to 0.0.10

### LINKED ISSUE #1899: Update Woodwork to 0.0.10
Rather than updating to 0.0.9, update WW to 0.0.10 where issue with DataTables resetting index has been resolved.

angela97lin:
Closes #1899 

Changes made in `0.0.10` to `mutual_information` (https://github.com/alteryx/woodwork/pull/563) s.t. columns which have completely unique values are no longer considered. Hence, had to tweak some of our tests so that the return values were as expected.

### DIFF
--- a/evalml/data_checks/target_leakage_data_check.py
+++ b/evalml/data_checks/target_leakage_data_check.py
@@ -69,11 +69,11 @@ class TargetLeakageDataCheck(DataCheck):
         Example:
             >>> import pandas as pd
             >>> X = pd.DataFrame({
-            ...    'leak': [10, 42, 31, 51, 61],
+            ...    'leak': [10, 44, 31, 51, 44],
             ...    'x': [42, 54, 12, 64, 12],
             ...    'y': [13, 5, 13, 74, 24],
             ... })
-            >>> y = pd.Series([10, 42, 31, 51, 40])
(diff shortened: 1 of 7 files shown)


bchen1116 (approved):
LGTM

evalml/data_checks/target_leakage_data_check.py
```diff
@@ -69,11 +69,11 @@ def validate(self, X, y):
         Example:
             >>> import pandas as pd
             >>> X = pd.DataFrame({
-            ...    'leak': [10, 42, 31, 51, 61],
+            ...    'leak': [10, 44, 31, 51, 44],
             ...    'x': [42, 54, 12, 64, 12],
             ...    'y': [13, 5, 13, 74, 24],
             ... })
-            >>> y = pd.Series([10, 42, 31, 51, 40])
+            >>> y = pd.Series([10, 42, 31, 51, 42])
```
freddyaboulton, line 76: I saw your comment about how woodwork will not calculate mutual information if one of the columns has all unique values. I guess that means that the mutual information is 0 in that case.

Does this mean we should update the `DefaultDataChecks` to use `method='pearson'` for regression? Columns can still be 95% correlated with the target if the target is all unique.

Issue #1883, opened 2021-02-23
Update AutoMLSearch with `actions` parameter


angela97lin:
Design doc here: https://alteryx.quip.com/3yqBAaGDDjWk/Data-Checks-Recommended-Actions

PR #1907, opened 2021-03-01
Introduce pipeline / component creation from DataCheckActions

### LINKED ISSUE #1879: Introduce pipeline / component creation from DataCheckActions

angela97lin:
Closes #1879

Currently, the only action we have is `DropCol` 😅


evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -0,0 +1,478 @@
...
+    assert not np.isnan(pipeline.predict(X).to_series()).values.any()
+
+    pipeline_score = pipeline.score(X, y, [objective])[objective]
+    comparison_pipeline_score = comparison_pipeline.score(X, y, [objective])[objective]
+
+    if problem_type == ProblemTypes.BINARY or problem_type == ProblemTypes.MULTICLASS:
+        assert not np.isnan(pipeline.predict_proba(X).to_dataframe()).values.any()
+        assert (pipeline_score <= comparison_pipeline_score)
+    else:
+        assert (pipeline_score >= comparison_pipeline_score)
```
angela97lin, line 466: Moved from `test_pipelines.py` since these are stored in `evalml/pipelines/utils.py` and `test_pipelines.py` is huge

jeremyliweishih (approved):
cool! this looks good to me. When the data check actions start adding up we can revisit and think about if we want to stick with if statements or [another implementation](https://stackoverflow.com/questions/60208/replacements-for-switch-statement-in-python).

bchen1116 (approved):
LGTM! Yep, agree with @jeremyliweishih, when there are more actions, it'd be useful to look into the keyed dictionary return rather than if/else statements, just to make it a little cleaner. No changes needed now!

freddyaboulton (approved):
@angela97lin This looks good!

The design doc mentions adding a `preprended_components` argument to `make_pipeline`. Is there an issue for that? Or are you planning on doing it in #1883 ?

angela97lin:
@freddyaboulton Yupperino, I was thinking about doing it as part of #1883, though I guess no reason why I couldn't do it here/separately, heh. Gonna stick with that plan rather than ask for reapprovals on more impl!

PR #1909, opened 2021-03-01
Update `describe_pipeline` to return more information about pipeline, including id for ensemble

### LINKED ISSUE #1872: Update `describe_pipeline` to return more information about pipeline, including id for ensemble
Fallout of https://alteryx.quip.com/VyJ6Ac1sAc7c/AutoML-Ensembling-Indicate-which-pipelines-were-selected-for-ensembling.

We should update `describe_pipeline` s.t. when we want to describe an ensemble pipeline, we return information about the ids of the pipelines that were used to build the ensemble pipeline.

angela97lin:
Closes #1872
- Fix bug where ensemble was being added to `IterativeAlgorithm._best_pipeline_info`
- Fix `pipeline.describe()` not having `return_dict` parameter


evalml/pipelines/pipeline_base.py
```diff
@@ -154,11 +154,11 @@ def get_component(self, name):
         """
         return self._component_graph.get_component(name)
 
-    def describe(self):
+    def describe(self, return_dict=False):
```
angela97lin, line 157: Was missing `return_dict` parameter, even though it was in docstrings :p

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -548,32 +548,23 @@ def test_indexing(X_y_binary, logistic_regression_binary_pipeline_class):
...
-    assert "Number of features: " not in out
-
-    for component in lrp:
-        if component.hyperparameter_ranges:
-            for parameter in component.hyperparameter_ranges:
-                assert parameter in out
-        assert component.name in out
-
+@pytest.mark.parametrize("is_linear", [True, False])
+def test_describe(is_linear, caplog, logistic_regression_binary_pipeline_class, nonlinear_binary_pipeline_class):
```
angela97lin, line 552: Combined `test_describe` and `test_describe_nonlinear`

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -115,7 +115,7 @@ def add_result(self, score_to_minimize, pipeline, trained_pipeline_results):
             self._first_batch_results.append((score_to_minimize, pipeline.__class__))
 
         current_best_score = self._best_pipeline_info.get(pipeline.model_family, {}).get('score', np.inf)
-        if score_to_minimize is not None and score_to_minimize < current_best_score:
+        if score_to_minimize is not None and score_to_minimize < current_best_score and pipeline.model_family != ModelFamily.ENSEMBLE:
```
angela97lin, line 118: We don't want to add the ensemble model, which uses `self._best_pipeline_info` for input pipelines, to _best_pipeline_info.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -548,32 +548,23 @@ def test_indexing(X_y_binary, logistic_regression_binary_pipeline_class):
         clf[:1]
 
 
-def test_describe(caplog, logistic_regression_binary_pipeline_class):
```
angela97lin, line 551: Combined `test_describe` and `test_describe_nonlinear` and `test_describe_fitted` and `test_describe_nonlinear_fitted`

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -408,7 +408,6 @@ class DummyEstimator(Estimator):
     assert pipeline.parameters == new_pipeline.parameters
     for component, new_component in zip(pipeline._component_graph, new_pipeline._component_graph):
         assert isinstance(new_component, type(component))
-    assert pipeline.describe() == new_pipeline.describe()
```
angela97lin, line 411: This actually just asserts None == None lol... also doesn't make sense to test equality via describe though (name is different), so no need for this test.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -154,6 +154,7 @@ def test_iterative_algorithm_results(mock_stack, ensembling_value, dummy_binary_
             random_seeds_the_same = [(estimator.pipeline.random_seed == algo.random_seed)
                                      for estimator in estimators_used_in_ensemble]
             assert all(random_seeds_the_same)
+            assert ModelFamily.ENSEMBLE not in algo._best_pipeline_info
```
angela97lin, line 157: Confirms we don't add ensemble  to `_best_pipeline_info`

evalml/automl/automl_search.py
```diff
@@ -778,10 +778,19 @@ def describe_pipeline(self, pipeline_id, return_dict=False):
...
+        automl_dict = dict()
+
         pipeline = self.get_pipeline(pipeline_id)
         pipeline_results = self._results['pipeline_results'][pipeline_id]
+        automl_dict.update(pipeline_results)
 
         pipeline.describe()
+
+        if pipeline.model_family == ModelFamily.ENSEMBLE:
+            input_pipeline_ids = [self._automl_algorithm._best_pipeline_info[model_family]["id"] for model_family in self._automl_algorithm._best_pipeline_info]
```
freddyaboulton, line 790: If I'm understanding correctly, `_best_pipeline_info` will only contain one key per model family. So after the first ensemble pipeline is created, the data in `_best_pipeline_info` (may) be overwritten. This means that the `input_pipeline_ids` would always correspond to the latest ensemble and not necessarily the ensemble that was asked for.

Repro:

```python

X, y = load_breast_cancer()

# max_batches=19 means we ensemble twice
automl = AutoMLSearch(X, y, problem_type="binary", ensembling=True, max_batches=19)
automl.search()
automl.describe_pipeline(49)
```

We know the ids > 49 are wrong because the ensemble could have only been created with ids 1-48.

```
Input for ensembler are pipelines with IDs: [85, 32, 65, 53, 78, 16, 64]
```

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: Nice!

freddyaboulton (changes requested):
@angela97lin I think this is great! I think the only thing blocking merge is that this implementation always returns the ids for the latest ensemble and not the ensemble that was asked for.

evalml/automl/automl_algorithm/iterative_algorithm.py
angela97lin: I believe it would have "worked" but the second ensemble would try to add an ensemble as an input pipeline, which is probably not what we want :d

evalml/automl/automl_search.py
angela97lin: Amazing, thank you for catching this! 😅 

I updated the logic so that `self._results['pipeline_results']` stores the ids (using `_best_pipeline_info`) whenever an ensemble is built. I think this is fine, since when adding the ensemble, `_best_pipeline_info` has the right info for that pipeline (and as you pointed out, can later get overridden).

Test wise, updated `test_describe_pipeline_with_ensembling` to test for two ensembles and tested what you pointed out--that the ensemble should be created from ids less than the ensemble. Added a mock score that always decreases (latter pipelines do better) and then test that the second ensemble uses pipeline ids that come after the first ensemble pipeline.

evalml/automl/automl_search.py
```diff
@@ -811,7 +823,7 @@ def describe_pipeline(self, pipeline_id, return_dict=False):
             logger.info(all_objective_scores)
 
         if return_dict:
-            return pipeline_results
+            return automl_dict
```
bchen1116, line 826: Why do we create `automl_dict` rather than just return `pipeline_results` like we did before? Is `pipeline_results` not a dictionary already?

bchen1116 (approved):
LGTM! Just left a question for my own understanding, but nothing blocking.

freddyaboulton (approved):
@angela97lin Thanks for fixing! I think this is great!!

evalml/automl/automl_search.py
angela97lin: No reason! In previous impl I needed to add something new but I ended up adding to `pipeline_results` instead so the two are the same again. I'll remove :)

Issue #1835, opened 2021-02-11
Demo: Two-way partial dependence plots


chukarsten:
Two-way partial dependence was implemented, but the model_understanding.ipynb notebook was not updated to demonstrate it.  (Thanks @chukarsten).  A section should be added to demonstrate some simple two-way partial dependence underneath the current one-way partial dependence plot and any other supporting cells in the notebook updated to include the functionality.

dsherry:
@chukarsten this was closed last week, correct? I think the PR didn't close it. Marking as close, feel free to correct if I'm wrong.

chukarsten:
You're good!  You see, I told you people (me) weren't properly maintaining the board!

PR #1919, opened 2021-03-03
Demo of Two-way Part Dep Plots

chukarsten:
Addresses #1835 

Added some updates to the model_understanding.ipynb!

image

image


docs/source/user_guide/model_understanding.ipynb
```diff
@@ -179,6 +179,31 @@
...
+    "Two-way partial dependence plots are also possible and invoke the same API."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "partial_dependence(pipeline, X, features=('worst perimeter', 'worst radius'))"
```
chukarsten, line 195: I'm not sure why, but the two-way partial dependence calculation takes longer than the graph_partial_dependence function, which, iirc, calls partial_dependence.  It takes a fair amount of time...5 minutes? We would be accepting an increase to the build_docs runtime.

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -649,353 +644,6 @@ def test_graph_binary_objective_vs_threshold(mock_cb_thresholds, data_type, X_y_
     assert np.array_equal(data['y'], mock_cb_thresholds.return_value['score'])
 
 
-def check_partial_dependence_dataframe(pipeline, part_dep, grid_size=20):
```
chukarsten, line 652: Moved the partial_dependence tests over to their own file for organizational purposes.

docs/source/user_guide/model_understanding.ipynb
freddyaboulton: I think if we set `grid_resolution=20` it should take a lot less time (it's what the `graph_partial_dependence` call below is doing.)

I believe this scales quadratically with the grid size in the two-way case and if so we should make a note of it here and/or the docstring.

freddyaboulton (approved):
@chukarsten Looks good! I left a comment about using a smaller grid size in the notebook to speed it up.

PR #1928, opened 2021-03-04
Make HighVarianceCVDataCheck a standalone object/method instead of a data check

### LINKED ISSUE #1861: Make HighVarianceCVDataCheck a standalone object/method instead of a data check
This fell out of the discussion for the data checks recommended actions epic (#1730). We added `HighVarianceCVDataCheck` because we wanted it to have the same API as a data check. We didn't want to sprinkle in these extra checks in between search randomly. However, it's worth revisiting whether or not `HighVarianceCVDataCheck` should really be a data check, as it doesn't fit the paradigm that data checks check the input data and checks are done before search.

angela97lin:
Closes #1861


docs/source/release_notes.rst
```diff
@@ -18,6 +18,7 @@ Release Notes
         * Updated ``add_results`` in ``AutoMLAlgorithm`` to take in entire pipeline results dictionary from ``AutoMLSearch`` :pr:`1891`
         * Updated ``ClassImbalanceDataCheck`` to look for severe class imbalance scenarios :pr:`1905`
         * Deleted the ``explain_prediction`` function :pr:`1915`
+        * Removed ``HighVarianceCVDataCheck`` and convered it to a standalone method instead :pr:`1928`
```
freddyaboulton, line 21: I think this should also be marked as a breaking change

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2430,3 +2430,20 @@ def test_automl_raises_deprecated_random_state_warning(X_y_multi):
         automl = AutoMLSearch(X_train=X, y_train=y, problem_type='multiclass', random_state=10)
         assert automl.random_seed == 10
         assert str(warn[0].message).startswith("Argument 'random_state' has been deprecated in favor of 'random_seed'")
+
+
+def test_automl_check_for_high_variance(X_y_binary, dummy_binary_pipeline_class, caplog):
```
freddyaboulton, line 2435: Not entirely in the scope of this issue but I just noticed we don't test if the high variance warning runs during search. I think it would be worth doing that eventually but we can tackle that in a separate issue. What do you think?

evalml/automl/automl_search.py
```diff
@@ -764,6 +757,18 @@ def _post_evaluation_callback(self, pipeline, evaluation_results):
             self.add_result_callback(self._results['pipeline_results'][pipeline_id], pipeline, self)
         return pipeline_id
 
+    def _check_for_high_variance(self, pipeline, cv_scores, threshold=0.2):
+        """Checks cross-validation scores and logs a warning if variance is higher than specified threshhold."""
+        pipeline_name = pipeline.name
+        high_variance_cv = False
+
+        if cv_scores.mean() != 0:
+            high_variance_cv = bool(abs(cv_scores.std() / cv_scores.mean()) > threshold)
```
freddyaboulton, line 766: Isn't `abs(cv_scores.std() / cv_scores.mean()) > threshold` already a bool?

freddyaboulton (approved):

evalml/automl/automl_search.py
angela97lin: @freddyaboulton Unfortunately it resolves to a `numpy.bool_` which is different 😭

jeremyliweishih (approved):
very nice

evalml/automl/automl_search.py
freddyaboulton: Damn. But then how does the old code work? 🤔

angela97lin: Old code (and this code) check `if high_variance_cv` which will evaluate to True regardless of python or numpy bool, but now we want to take `high_variance_cv` and set it in `pipeline_results`. (Previously, just checked if # of warnings greater than 0, then set `high_variance_cv` to True so not using it directly.)

I wasn't aware of the difference myself until `test_search_results` broke at `assert isinstance(results['high_variance_cv'], bool)` without this type conversion 😬

evalml/tests/automl_tests/test_automl.py
angela97lin: Yup, you're right--we test that the output is bool but not whether we actually log. I'll add a test 🥳

evalml/automl/automl_search.py
```diff
@@ -764,6 +757,18 @@ def _post_evaluation_callback(self, pipeline, evaluation_results):
             self.add_result_callback(self._results['pipeline_results'][pipeline_id], pipeline, self)
         return pipeline_id
 
+    def _check_for_high_variance(self, pipeline, cv_scores, threshold=0.2):
+        """Checks cross-validation scores and logs a warning if variance is higher than specified threshhold."""
+        pipeline_name = pipeline.name
+        high_variance_cv = False
+
+        if cv_scores.mean() != 0:
```
chukarsten, line 765: What case is this capturing?

chukarsten (approved):
Dig it.  Nice work.

docs/source/release_notes.rst
```diff
@@ -18,13 +18,16 @@ Release Notes
         * Updated ``add_results`` in ``AutoMLAlgorithm`` to take in entire pipeline results dictionary from ``AutoMLSearch`` :pr:`1891`
         * Updated ``ClassImbalanceDataCheck`` to look for severe class imbalance scenarios :pr:`1905`
         * Deleted the ``explain_prediction`` function :pr:`1915`
+        * Removed ``HighVarianceCVDataCheck`` and convered it to an ``AutoMLSearch`` method instead :pr:`1928`
```
bchen1116, line 21: Typo: convered?

evalml/automl/automl_search.py
angela97lin: Hm, this was more a rework of what was already there and the fear of `cv_scores.mean() == 0` meaning you divide by 0 in the lines that follow... but it seems like numpy just handles division by 0 as nan, and nan won't ever be greater than the threshold so maybe we don't need this after all. I'll remove!

evalml/automl/automl_search.py
```diff
@@ -764,6 +757,14 @@ def _post_evaluation_callback(self, pipeline, evaluation_results):
             self.add_result_callback(self._results['pipeline_results'][pipeline_id], pipeline, self)
         return pipeline_id
 
+    def _check_for_high_variance(self, pipeline, cv_scores, threshold=0.2):
```
bchen1116, line 760: nitpick, but should we enforce that users can't pass in `threshold < 0`, even though this is a private method?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2430,3 +2430,46 @@ def test_automl_raises_deprecated_random_state_warning(X_y_multi):
...
+
+def test_automl_check_for_high_variance(X_y_binary, dummy_binary_pipeline_class):
+    X, y = X_y_binary
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary')
+    cv_scores = pd.Series([1, 1, 1])
+    pipeline = dummy_binary_pipeline_class(parameters={})
+    assert not automl._check_for_high_variance(pipeline, cv_scores)
+
+    cv_scores = pd.Series([0, 0, 0])
+    assert not automl._check_for_high_variance(pipeline, cv_scores)
```
bchen1116, line 2443: Might be nice to test different thresholds just to be super comprehensive

bchen1116 (approved):
Looks good! I left a few comments on testing/thresholding, but nothing blocking.

PR #1934, opened 2021-03-05
Update AutoML to use objective decision function during scoring for custom objectives

### LINKED ISSUE #1868: Automl does not use objective decision function during scoring
Our classification pipeline uses the `_compute_predictions` method in `score` to get the predicted labels and probabilities. 

However, `_compute_predictions` [does not pass the objective](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/classification_pipeline.py#L138) to the pipeline `predict` method.  

Is this intended?

angela97lin:
Closes #1868

- Added `BinaryClassificationPipelineMixin` to share logic between binary classification / time series binary classification pipelines regarding the pipeline threshold.
- Moved `optimize_threshold` from ClassificationPipeline to BinaryClassificationPipelineMixin. I believe this makes sense since only binary classification pipelines, not multiclass classification pipelines, know about thresholds.


evalml/pipelines/binary_classification_pipeline.py
```diff
@@ -37,10 +42,14 @@ def _predict(self, X, objective=None):
         if self.threshold is None:
             return self._component_graph.predict(X)
         ypred_proba = self.predict_proba(X).to_dataframe()
+        predictions = self._predict_with_objective(X, ypred_proba, objective)
+        return infer_feature_types(predictions)
+
+    def _predict_with_objective(self, X, ypred_proba, objective):
```
angela97lin, line 48: Helper function that takes in `predict_proba`, so we don't have to recalculate it for each objective.

evalml/pipelines/binary_classification_pipeline.py
```diff
@@ -60,3 +69,34 @@ def _score(X, y, predictions, objective):
...
+                    raise ValueError(f'Invalid objective {objective.name} specified for problem type {self.problem_type}')
+                y_pred_to_use = y_pred
+                if self.threshold is not None and not objective.score_needs_proba:
+                    y_pred_to_use = self._predict_with_objective(X, y_pred_proba, objective)
+                score = self._score(X, y, y_pred_proba if objective.score_needs_proba else y_pred_to_use, objective)
+                scored_successfully.update({objective.name: score})
+            except Exception as e:
+                tb = traceback.format_tb(sys.exc_info()[2])
+                exceptions[objective.name] = (e, tb)
+        if exceptions:
```
chukarsten, line 98: Hmmm, the naming of the variables  and the try/except kind of implied to me that you were going to be forgiving of one or any of the objectives failing.  But I guess what you're trying to do is just get a catalog of what worked and what didn't without raising at the *first* instance of a failed objective.  Am I following that?  So either you get the scores only if everything scores correctly or you get and error detailing what failed and what was successful.

evalml/pipelines/binary_classification_pipeline.py
```diff
@@ -60,3 +69,34 @@ def _score(X, y, predictions, objective):
...
         return ClassificationPipeline._score(X, y, predictions, objective)
+
+    def _compute_predictions(self, X, y, objectives, time_series=False):
+        """Compute predictions/probabilities based on objectives."""
+        y_predicted = None
+        y_predicted_proba = None
+        if any(o.score_needs_proba for o in objectives) or (any(not o.score_needs_proba for o in objectives) and self.threshold is not None):
+            y_predicted_proba = self.predict_proba(X, y) if time_series else self.predict_proba(X)
+        if any(not o.score_needs_proba for o in objectives) and self.threshold is None:
+            y_predicted = self._predict(X, y, pad=True) if time_series else self._predict(X)
```
chukarsten, line 80: Is this pad something that we always want on or that the user would always expect to be True?

chukarsten (changes requested):
I think there's something funky with the double anys on L77 of binary_classification_pipeline.py.  Besides that, this looks pretty solid.

evalml/pipelines/binary_classification_pipeline.py
```diff
@@ -60,3 +69,34 @@ def _score(X, y, predictions, objective):
         if predictions.ndim > 1:
             predictions = predictions.iloc[:, 1]
         return ClassificationPipeline._score(X, y, predictions, objective)
+
+    def _compute_predictions(self, X, y, objectives, time_series=False):
+        """Compute predictions/probabilities based on objectives."""
+        y_predicted = None
+        y_predicted_proba = None
+        if any(o.score_needs_proba for o in objectives) or (any(not o.score_needs_proba for o in objectives) and self.threshold is not None):
```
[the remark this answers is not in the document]
angela97lin: Heh, yeah this is confusing stuff, good stuff to clarify. Hopefully this helps (and points out any errors I might have missed). Here's the parred-down version of the code in main for `_predict()`:

```
    def _predict(self, X, objective=None):
        if self.threshold is None:
            return self._component_graph.predict(X)
        ypred_proba = self.predict_proba(X).to_dataframe()
        ypred_proba = ypred_proba.iloc[:, 1]
        if objective is None:
            return infer_feature_types(ypred_proba > self.threshold)
        return infer_feature_types(objective.decision_function(ypred_proba, threshold=self.threshold, X=X))

```

Basically, if we have a `threshold` set for our pipeline, we use `predict_proba` and then either call `ypred_proba > self.threshold` or the decision function if passed an objective.

This logic here for computing the predictions means is trying to say:
1. Calculate predict_proba if any objective needs it for scoring (old behavior)
2. If objective can optimize threshold (aka score doesn't need probability--see `can_optimize_threshold` property) and the threshold has been set, then we will still want the predict probability to calculate the last two lines.

To your confusion @chukarsten:
1. If we only have an objective w/ score_needs_proba: we calculate (conditional is True)
2. If we only have an objective that has an optimizable threshold and threshold has been set: we calculate
3. If we only have an objective that has an optimizable threshold and threshold has not been set: we don't need to calculate (will call `self._component_graph.predict(X)` line above)

Writing this up though, maybe I can simplify this to `if any(o.score_needs_proba for o in objectives) or self.threshold is not None)` since `any(not o.score_needs_proba for o in objectives)` is captured in setting the threshold 🤔

evalml/pipelines/binary_classification_pipeline.py
angela97lin: Correct! This is actually mostly copied from `PipelineBase` and our current behavior, with lines 91-93 being the difference. But yes, you're spot on. We don't want to error out on the first objective scoring, and want to calculate what we can. We decided this because for AutoML, maybe some objectives just happen to error out but you don't really care about them. In that case, it doesn't make sense to halt search just for this objective, so best to just call that error out and then move on. (When *all* objectives fail though in AutoML though, we scream)

evalml/pipelines/binary_classification_pipeline.py
angela97lin: I think for time-series, yes! @freddyaboulton correct me if I'm wrong 😅

evalml/pipelines/binary_classification_pipeline.py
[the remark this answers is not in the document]
angela97lin: @chukarsten 
Okay, trying to wrap my head around this some more, and I guess the benefit of having `if any(o.score_needs_proba for o in objectives) or self.threshold is not None)` is that you only calculate if the threshold is set. In the case where score_needs_proba is False and the threshold is None, you don't need to calculate.


```
score needs proba | threshold is None | calculate predict_proba?
------------------------------------------------------------------------
           T               T                   YES
------------------------------------------------------------------------
           T               F                   YES
------------------------------------------------------------------------
           F               T                   YES
------------------------------------------------------------------------
           F               F                   NO

```

evalml/pipelines/binary_classification_pipeline.py
freddyaboulton: Yes! It's an internal detail of time series pipelines but we always want the predicted probabilities and predictions to be padded. @angela97lin Can you add coverage for time series pipelines to `test_binary_predict_pipeline_use_objective`?

evalml/pipelines/binary_classification_pipeline.py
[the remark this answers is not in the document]
freddyaboulton: @angela97lin Maybe I'm confused now too 😂  but shouldn't the last two 'YES' and 'NO' be flipped? If the threshold is None and the score does not need probability, you don't need to calculate predict_proba?

Overall, I'm in favor of rewritting in terms of `if any(o.score_needs_proba for o in objectives) or self.threshold is not None)`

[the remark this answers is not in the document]
angela97lin: @freddyaboulton I think I'm the confused one, haha--fixed! But sounds good, I'll simplify this 😁

evalml/pipelines/binary_classification_pipeline.py
freddyaboulton: I think it'd be best if there was only one `_score_all_objectives` implementation because I want the `PipelineScoreError` logic to only live in one place if we decide to change it in the future.

This works and I don't want to hold up the merge longer but I think we should at least file an issue for refactoring in the future. What do you think @angela97lin @chukarsten ?

freddyaboulton (approved):
@angela97lin This looks good to me! I left some scattered comments throughout. One thing that would be nice to address before merge is adding a time series test to `test_binary_predict_pipeline_use_objective`

evalml/pipelines/binary_classification_pipeline.py
```diff
@@ -1,23 +1,17 @@
...
 from evalml.utils import infer_feature_types
 
 
-class BinaryClassificationPipeline(ClassificationPipeline):
+class BinaryClassificationPipeline(BinaryClassificationPipelineMixin, ClassificationPipeline):
     """Pipeline subclass for all binary classification pipelines."""
-    _threshold = None
     problem_type = ProblemTypes.BINARY
 
-    @property
```
angela97lin, line 12: Moved all of this to `BinaryClassificationPipelineMixin`

evalml/pipelines/binary_classification_pipeline.py
angela97lin: Did some more refactoring magic, so this is all in one place again 🎈

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2009,37 +2009,3 @@ def test_binary_pipeline_string_target_thresholding(make_data_type, logistic_reg
...
     assert pipeline.threshold is not None
-
-
-@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
-@patch('evalml.pipelines.BinaryClassificationPipeline.score')
-@patch('evalml.pipelines.BinaryClassificationPipeline.predict_proba')
-@patch('evalml.pipelines.MulticlassClassificationPipeline.fit')
-@patch('evalml.pipelines.MulticlassClassificationPipeline.score')
-@patch('evalml.pipelines.MulticlassClassificationPipeline.predict')
-def test_pipeline_thresholding_errors(mock_multi_predict, mock_multi_score, mock_multi_fit,
```
angela97lin, line 2020: Removed multiclass tests and moved to test_binary_classification_pipelines.py

PR #1935, opened 2021-03-05
Delete data check code from AutoML search

### LINKED ISSUE #1882: Delete data check code from AutoML

After discussion with @dsherry, this issue will now track deleting the data check code from AutoMLSearch.

AutoMLSearch is an object that currently has two major stages. It initializes and sets up state, and then searches for models. Data checking is extra functionality that could require more user input and manual handling, as `infer_feature_types` is. Hence, it makes sense to remove it from the AutoMLSearch process.

angela97lin:
Closes #1882


docs/source/release_notes.rst
```diff
@@ -23,6 +23,7 @@ Release Notes
         * Updated ``add_results`` in ``AutoMLAlgorithm`` to take in entire pipeline results dictionary from ``AutoMLSearch`` :pr:`1891`
         * Updated ``ClassImbalanceDataCheck`` to look for severe class imbalance scenarios :pr:`1905`
         * Deleted the ``explain_prediction`` function :pr:`1915`
+        * Deleted data check code in ``AutoMLSearch`` :pr:`1935`
```
freddyaboulton, line 26: I think this is also a breaking change. We should also mention this means we're deleting the `data_check_results` property and that `data_checks` is not a valid `search` argument anymore.

evalml/automl/automl_search.py
```diff
@@ -171,6 +160,8 @@ def __init__(self,
 
             train_best_pipeline (boolean): Whether or not to train the best pipeline before returning it. Defaults to True
 
+            pipeline_parameters
```
freddyaboulton, line 163: Typo?

docs/source/user_guide/automl.ipynb
```diff
@@ -142,17 +142,6 @@
     "print(regression_objective.name)"
    ]
   },
-  {
```
freddyaboulton, line 145: I still think we should mention data checks in the automl user guide.

If I'm following the intention of the data check actions epic, we're moving from "AutoML will check your data for you" to "Evalml has functionality to check your data and recommend fixes before you start AutoML". 

I think it's important to document the relationship between data checks and automl and that it makes sense to document this relationship here. 

So I think we should:
1. Add back the data checks section (changing it a bit to make it clear that `AutoMLSearch` is not running the checks)
2. Import `DefaultDataChecks` and run on the data. 
3. Say that the results of the data checks don't reveal anything that would cause problems so we can run search. 
4. Link to the data checks user guide page for more information

Let me know what you think!

freddyaboulton (changes requested):
@angela97lin The code changes look great! I think we should synch about whether or not we should delete the data checks section from the automl user guide before merging!

docs/source/user_guide/automl.ipynb
angela97lin: Hmmm, I like how this sounds... my only qualm about this is it would make sense for me that the AutoML user guide currently focuses on the functionality within automl (we currently list out error callbacks, passing in custom pipelines, etc.). This seems more like a flow of our the user should use EvalML.
I guess TLDR is: for what our docs currently look like, I like this change. Maybe this also means we should separate out our docs and not shove everything into the AutoML guide; we could have a intro user guide that steps through the steps a user would take, and then a dedicated AutoML guide that deep-dives into AutoMLSearch specifics.

Will update this, thank you for the great feedback @freddyaboulton!

bchen1116 (approved):
LGTM! Agree this should be a breaking change and should discuss what happens with `automl.ipynb`

docs/source/release_notes.rst
dsherry: Agreed! Let's add a sentence or two in the breaking changes about how to handle this, too. Something like "Please run the data checks before AutoML like so: `results = DefaultDataChecks().validate(X_train, y_train)`"

docs/source/user_guide/automl.ipynb
```diff
@@ -62,6 +62,36 @@
     "X = infer_feature_types(X, feature_types={'provider': 'categorical'})"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Data Checks\n",
+    "\n",
+    "Before calling `AutoMLSearch.search`, we should run some sanity checks on our data to ensure that the input data being passed will not run into some common issues before running a potentially time-consuming search. EvalML has various data checks that makes this easy. Each data check will return a collection of warnings and errors if it detects potential issues with the input data. This allows users to inspect their data to avoid confusing errors that may arise during the search process. You can learn about each of the data checks available through our [data checks guide](data_checks.ipynb) \n",
```
freddyaboulton, line 71: Beautiful

freddyaboulton (approved):
@angela97lin Looks great! This change add [two uncovered lines ](https://app.codecov.io/gh/alteryx/evalml/compare/1935/changes)to our repo. 

Can you please add coverage in `test_data_checks.py` It should be quick but we can also handle that in a separate issue.

In my opinion, only the first uncovered line is important: `raise ValueError(f"Parameter data_checks must be a list. Received {type(data_check_classes).__name__}.")`. We're probably going to get rid of the `AutoMLDataChecks` class since AutoMLSearch no longer knows about data checks so the second uncovered line is not high priority.

chukarsten (approved):

angela97lin:
@freddyaboulton Thanks for catching this! Sucks that codecov let me slip this by but I added a test and just decided to remove `AutoMLDataChecks` in this PR since it's not really useful anymore / was only used in automl.

docs/source/release_notes.rst
```diff
@@ -6,9 +6,16 @@ Release Notes
...
     * Fixes
     * Changes
+        * Removed ``data_checks`` parameter, ``data_check_results`` and data checks logic from ``AutoMLSearch`` :pr:`1935`
     * Documentation Changes
     * Testing Changes
 
+.. warning::
+
+    **Breaking Changes**
+        * Removed ``data_checks`` parameter, ``data_check_results`` and data checks logic from ``AutoMLSearch`` To run data checks, please call DataCheck.validate(X_train, y_train) or take a look at our documentation for more examples. :pr:`1935`
```
dsherry, line 16: Could we say "To run the data checks which were previously run by default in `AutoMLSearch`, please call `DefaultDataChecks().validate(X_train, y_train) or" etc?

docs/source/user_guide/automl.ipynb
```diff
@@ -62,6 +62,36 @@
...
+    "\n",
+    "data_checks = DefaultDataChecks(\"binary\", \"log loss binary\")\n",
+    "data_checks.validate(X, y)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Since there were no warnings or errors returned, we can safely continue with the search process."
```
dsherry, line 92: This is great! Gonna be really cool when we can replace this with the data check actions :D

evalml/data_checks/data_checks.py
```diff
@@ -99,11 +99,3 @@ def validate(self, X, y=None):
             messages["actions"].extend(messages_new["actions"])
 
         return messages
-
-
-class AutoMLDataChecks(DataChecks):
```
dsherry, line 104: 👍

evalml/automl/automl_search.py
```diff
@@ -433,33 +417,6 @@ def _validate_problem_configuration(self, problem_configuration=None):
...
-                return DefaultDataChecks(problem_type=self.problem_type, objective=self.objective, n_splits=self.data_splitter.get_n_splits())
-            elif data_checks == "disabled":
-                return EmptyDataChecks()
-            else:
-                raise ValueError("If data_checks is a string, it must be either 'auto' or 'disabled'. "
-                                 f"Received '{data_checks}'.")
-        elif data_checks is None:
-            return EmptyDataChecks()
-        else:
-            return DataChecks(data_checks)
```
dsherry, line 461: I think its fine to delete this today. I could see us lifting/parroting this code in the future for the data check actions helper.

dsherry (approved):
Excellent!

docs/source/user_guide/automl.ipynb
angela97lin: Agreeeeed 🥳

PR #1943, opened 2021-03-08
Default optimize_thresholds to True in AutoMLSearch

### LINKED ISSUE #648: Re-enable binary classification threshold tuning by default
We added this feature on the #346 feature branch, then backed it out in #606 because it was recomputing `predict` and slowing down automl.

We should re-enable this by default. In order to do so, we'd have to cache the prediction output, which is currently computed in score. The long-term solution is to memoize predictions with a cache (#466), but short-term we should be able to do something.

This also relates to #579, which tracks cleaning up the duplicate code between the pipeline classes' `score` methods.

bchen1116:
fix #648 

Perf test [here](https://alteryx.quip.com/EABPAuX7922k/Default-Optimizethreshold-to-True-Perf-Test)

In-depth dive added [here](https://alteryx.quip.com/EABPAuX7922k/Default-Optimizethreshold-to-True-Perf-Test#VWeACAvcD0b)


docs/source/demos/cost_benefit_matrix.ipynb
```diff
@@ -73,8 +73,11 @@
    "source": [
     "from evalml.demos.churn import load_churn\n",
     "from evalml.preprocessing import split_data\n",
+    "import pandas as pd\n",
+    "\n",
     "X, y = load_churn()\n",
     "X = X.set_types({'PaymentMethod':'Categorical', 'Contract': 'Categorical'}) # Update data types Woodwork did not correctly infer\n",
+    "y = pd.Series([0 if i == 'No' else 1 for i in y.to_series().values])\n",
```
bchen1116, line 80: Since we're training the best pipeline and thresholding it, it requires the binary targets to be [0, 1]

evalml/tests/automl_tests/test_automl.py
```diff
@@ -83,7 +83,7 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type, obje
         expected_pipeline_class = MulticlassClassificationPipeline
         X, y = X_y_multi
 
-    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=automl_type, objective=objective, max_iterations=2, n_jobs=1)
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=automl_type, objective=objective, optimize_thresholds=False, max_iterations=2, n_jobs=1)
```
bchen1116, line 86: needed to set `optimize_thresholds` as False in order to avoid issues during training `best_pipeline` with `predict_proba`. Otherwise, I could have mocked `predict_proba` for all the tests but this was a faster approach with similar results.

docs/source/demos/cost_benefit_matrix.ipynb
freddyaboulton: This feels like a pretty big (unintended) breaking change. If I'm following correctly, we can't use string-valued targets in `AutoMLSearch` if `train_best_pipeline=True` and `optimize_thresholds=True`?

evalml/automl/utils.py
```diff
@@ -72,14 +72,19 @@ def make_data_splitter(X, y, problem_type, problem_configuration=None, n_splits=
...
-def tune_binary_threshold(pipeline, objective, problem_type, X_threshold_tuning, y_threshold_tuning):
+def tune_binary_threshold(pipeline, objective, problem_type, X_threshold_tuning, y_threshold_tuning, best_pipeline=False):
     """Tunes the threshold of a binary pipeline to the X and y thresholding data
 
     Arguments:
-        pipeline (Pipeline): Pipeline instance to threshold
-        X_threshold_tuning (ww.DataTable): Features to tune pipeline to
-        y_threshold_tuning (ww.DataColumn): Target data to tune pipeline to
+        pipeline (Pipeline): Pipeline instance to threshold.
+        objective (ObjectiveBase): The objective we want to tune with. If not tuneable and best_pipeline is True, will use to F1.
```
freddyaboulton, line 80: @bchen1116 Can you explain this design choice? It feels arbitrary that we optimize the threshold against F1 when it's possible the user deliberately did not specify F1 as the primary objective or include it in `additional_objectives`.

freddyaboulton (commented):
@bchen1116 I have a couple design questions before taking a closer look!

docs/source/demos/cost_benefit_matrix.ipynb
bchen1116: @freddyaboulton, this was really weird. Running it locally works as expected, like this
image

But it fails when I try to make the docs. Not sure why it fails in both the docs and looking glass runs, but it throws an error in `objective.optimize_threshold` [here](https://github.com/alteryx/evalml/pull/1943/files#diff-d33b07734e238a4b877a6333e876204ecc236bd00d3a4549668e88bf6b3d802dR93)

evalml/automl/utils.py
bchen1116: This was something @dsherry and I decided on. It seemed to be the best default objective to use since it gives a fair interpretation of the data for both balanced and imbalanced datasets. 

This is definitely worth looking into and discussing, if you have ideas.

freddyaboulton: Why does the threshold tuning always need to happen even when the primary objective does not support it? 

My preference would be that the threshold would only be tuned for the best pipeline when the primary objective supports it for two reasons:

1. This is what's happening during search and I think it's important to keep the "training steps" the same between the best pipeline training and search
2. As a user, it feels weird that my pipeline is being tuned against something I did not specify as a search parameter.

evalml/automl/automl_search.py
```diff
@@ -593,8 +595,18 @@ def _find_best_pipeline(self):
                 else:
                     X_train = self.X_train
                     y_train = self.y_train
-                best_pipeline = self._engine.train_pipeline(best_pipeline, X_train, y_train,
-                                                            self.optimize_thresholds, self.objective)
+                if is_binary(self.problem_type) and self.objective.is_defined_for_problem_type(self.problem_type) and self.optimize_thresholds:
```
chukarsten: Thoughts?  Just added a line to kill the if-else
```
objective = self.objective
if is_binary(self.problem_type) and self.objective.is_defined_for_problem_type(self.problem_type) and self.optimize_thresholds:
    X_train, X_threshold_tuning, y_train, y_threshold_tuning = split_data(X_train, y_train, self.problem_type,
                                                                                          test_size=0.2,
                                                                                          random_seed=self.random_seed)
     if not self.objective.can_optimize_threshold:
         objective = get_objective("F1", return_instance=True)

     best_pipeline = self._engine.train_pipeline(best_pipeline, X_train, y_train, self.optimize_thresholds, objective)
```

evalml/automl/engine/engine_base.py
```diff
@@ -96,7 +96,7 @@ def train_pipeline(pipeline, X, y, optimize_thresholds, objective):
 
         Returns:
             pipeline (PipelineBase) - trained pipeline.
-        """
+        """
```
chukarsten: Am I reading this right?  Is white space being added here?

evalml/automl/utils.py
chukarsten: This is a good conversation.  The only thing I have to add is "True, will use F1".  Typo.  

With respect to the F1 point, I agree that its strange that objective has a kind of defacto default while it's required to be specified.  It seems that if you're trying to tune a threshold and the objective is not tunable...well, shoot, that sounds like an _exceptional_ situation, no? ;)

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -758,3 +762,21 @@ def test_automl_time_series_classification_threshold(mock_binary_fit, mock_binar
...
+@patch('evalml.pipelines.BinaryClassificationPipeline.score')
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+@patch('evalml.pipelines.BinaryClassificationPipeline.predict_proba')
+def test_tuning_threshold_objective(mock_predict, mock_fit, mock_score, mock_get_objective, objective, X_y_binary):
+    X, y = X_y_binary
+    mock_score.return_value = {objective: 0.5}
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective=objective)
+    automl.search()
+
+    if objective != "F1":
```
chukarsten, line 778: I feel kind of dense here.  What are you testing?  If objective == "F1" that the objective is called more than once?  And if it's "F1" that's only called one time?

evalml/automl/engine/engine_base.py
bchen1116: Yeah, unsure why this was here. Fixed!

docs/source/demos/cost_benefit_matrix.ipynb
freddyaboulton: @bchen1116 and I dug into this and figured out it's because we need to encode the targets before `objective.optimize_threshold`. 

We discussed two implementations:
1. Calling `pipeline._encode_targets` in the call to `optimize_thresholds` like this:
```python
            pipeline.threshold = objective.optimize_threshold(y_predict_proba,
                                                              pipeline._encode_targets(y_threshold_tuning.to_series()),
                                                              X=X_threshold_tuning)
```
2. Making the threshold tuning a binary classfication pipeline method
```python
pipeline.tune_threshold(X, y_true, y_pred_proba, objective)
```

I prefer the second option because it's the pipeline's responsibility to encode targets before fit/score so I think the consistent thing to do would be to let the pipeline encode targets before threshold tuning. Calling `_encode_targets` outside a pipeline method doesn't feel right to me.

evalml/automl/utils.py
bchen1116: Yep, great points. We can discuss this after standup today and figure out the best solution to go with!

evalml/tests/automl_tests/test_automl_search_classification.py
chukarsten: Ahhh, ok.

evalml/automl/utils.py
bchen1116: After discussion with @dsherry @freddyaboulton @angela97lin @chukarsten, we decided to remove this change to default to F1. We will leave `tune_threshold=True`, but we will not tune if the objective passed in isn't tunable.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -227,20 +227,21 @@ def test_additional_objectives(X_y_binary):
 
 
 @patch('evalml.objectives.BinaryClassificationObjective.optimize_threshold')
+@patch('evalml.pipelines.BinaryClassificationPipeline._encode_targets', side_effect=lambda y: y)
```
bchen1116, line 230: need to patch `encode_targets` in order to optimize the threshold

docs/source/user_guide/automl.ipynb
```diff
@@ -454,6 +454,7 @@
     "automl_with_ensembling = AutoMLSearch(X_train=X, y_train=y,\n",
     "                                      problem_type=\"binary\",\n",
     "                                      allowed_model_families=[ModelFamily.RANDOM_FOREST, ModelFamily.LINEAR_MODEL],\n",
+    "                                      train_best_pipeline=False,\n",
```
freddyaboulton, line 457: To speed up the build?

evalml/automl/engine/engine_base.py
```diff
@@ -147,7 +147,7 @@ def train_and_score_pipeline(pipeline, automl, full_X_train, full_y_train):
                 logger.debug(f"\t\t\tFold {i}: starting training")
                 cv_pipeline = EngineBase.train_pipeline(pipeline, X_train, y_train, automl.optimize_thresholds, automl.objective)
                 logger.debug(f"\t\t\tFold {i}: finished training")
-                if automl.optimize_thresholds and pipeline.can_tune_threshold_with_objective(automl.objective):
+                if automl.optimize_thresholds and pipeline.can_tune_threshold_with_objective(automl.objective) and automl.objective.can_optimize_threshold:
```
freddyaboulton, line 150: Nice!

evalml/pipelines/classification_pipeline.py
```diff
@@ -136,3 +137,18 @@ def _compute_predictions(self, X, y, objectives, time_series=False):
...
+    def optimize_threshold(self, X, y, y_pred_proba, objective):
+        """Optimize the pipeline threshold given the objective to use. Only used for binary problems with objectives that can be thresholded.
+
+        Arguments:
+            X (ww.DataTable): Input features
+            y (ww.DataColumn): Input target values
+            y_pred_proba (ww.DataColumn, np.nparray): The predicted probabilities of the target outputted by the pipeline
+            objective (ObjectiveBase): The objective to threshold with. Must be thresholdable.
+        """
+        if is_binary(self.problem_type) and objective.can_optimize_threshold:
```
freddyaboulton, line 150: I think we can replace this with `self.can_tune_threshold_with_objective(objective)`

evalml/tests/automl_tests/test_automl_search_classification.py
freddyaboulton: Yep, otherwise sklearn will throw an exception. The encoder is fit during `pipeline.fit` but since we're mocking fit, that doesn't happen.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
+
+def test_binary_pipeline_string_target_thresholding(make_data_type, X_y_binary):
+    class BinaryClass(BinaryClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    X, y = X_y_binary
+    X = make_data_type('ww', (X))
+    y = make_data_type('ww', pd.Series([f"String value {i}" for i in y]))
+    objective = get_objective("F1", return_instance=True)
+    pipeline = BinaryClass({})
```
freddyaboulton, line 2009: Can you please pass `n_jobs=1`?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    class MulticlassClass(MulticlassClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    X, y = X_y_multi
+    X = make_data_type('ww', (X))
+    y = make_data_type('ww', pd.Series([f"String value {i}" for i in y]))
+    objective = get_objective("F1 Macro", return_instance=True)
+    pipeline = MulticlassClass({})
```
freddyaboulton, line 2028: Same here about `n_jobs`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1255,7 +1256,7 @@ def test_catch_keyboard_interrupt(mock_fit, mock_score, mock_input,
     mock_input.side_effect = user_input
     X, y = X_y_binary
     callback = KeyboardInterruptOnKthPipeline(k=when_to_interrupt)
-    automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", max_iterations=5, start_iteration_callback=callback, objective="f1")
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", max_iterations=5, start_iteration_callback=callback, optimize_thresholds=False, objective="f1")
```
freddyaboulton, line 1259: Nice!

freddyaboulton (approved):
@bchen1116 This looks great!

docs/source/user_guide/automl.ipynb
bchen1116: Ah, was left over from when we got an error with label encoder. Removing!

evalml/pipelines/classification_pipeline.py
```diff
@@ -136,3 +136,18 @@ def _compute_predictions(self, X, y, objectives, time_series=False):
         if any(not o.score_needs_proba for o in objectives):
             y_predicted = self._predict(X, y, pad=True) if time_series else self._predict(X)
         return y_predicted, y_predicted_proba
+
+    def optimize_threshold(self, X, y, y_pred_proba, objective):
+        """Optimize the pipeline threshold given the objective to use. Only used for binary problems with objectives that can be thresholded.
```
angela97lin, line 141: LOL super nit-picky but is `thresholded` a word 😂  Can we say "with objectives whose thresholds can be tuned" or something?

evalml/pipelines/classification_pipeline.py
```diff
@@ -136,3 +136,18 @@ def _compute_predictions(self, X, y, objectives, time_series=False):
...
+
+        Arguments:
+            X (ww.DataTable): Input features
+            y (ww.DataColumn): Input target values
+            y_pred_proba (ww.DataColumn, np.nparray): The predicted probabilities of the target outputted by the pipeline
```
angela97lin, line 146: Just curious, why is this expected to be `y_pred_proba (ww.DataColumn, np.nparray)` as opposed to the other WW only inputs?

evalml/pipelines/classification_pipeline.py
```diff
@@ -136,3 +136,18 @@ def _compute_predictions(self, X, y, objectives, time_series=False):
...
+            objective (ObjectiveBase): The objective to threshold with. Must be thresholdable.
```
angela97lin, line 147: Same nit-picking as before, is thresholdable a word? lolol

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
+    pipeline = BinaryClass(parameters={"Logistic Regression Classifier": {"n_jobs": 1}})
+    pipeline.fit(X, y)
+    assert pipeline.threshold is None
+    pred_proba = pipeline.predict_proba(X, y).iloc[:, 1]
+    pipeline.optimize_threshold(X, y, pred_proba, objective)
+    assert pipeline.threshold is not None
+
+
+def test_pipeline_thresholding_errors(make_data_type, X_y_multi, X_y_binary):
+    class BinaryClass(BinaryClassificationPipeline):
```
angela97lin, line 2018: Just curious / nit-picky but why not use logistic_regression_multiclass_pipeline_class and logistic_regression_binary_pipeline_class fixtures?

angela97lin (approved):
Looks good! Mostly left some nit picky comments heh

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
     if is_binary(problem_type):
         pipeline.score(X, y, objectives=[CostBenefitMatrix(1, 1, -1, -1), "F1"])
+
+
+def test_binary_pipeline_string_target_thresholding(make_data_type, X_y_binary):
+    class BinaryClass(BinaryClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    X, y = X_y_binary
+    X = make_data_type('ww', (X))
```
angela97lin, line 2006: Extra parantheses?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
+
+def test_pipeline_thresholding_errors(make_data_type, X_y_multi, X_y_binary):
+    class BinaryClass(BinaryClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    class MulticlassClass(MulticlassClassificationPipeline):
+        component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']
+
+    X, y = X_y_multi
+    X = make_data_type('ww', (X))
```
angela97lin, line 2025: Same here, extra parentheses?

evalml/pipelines/classification_pipeline.py
```diff
@@ -136,3 +136,18 @@ def _compute_predictions(self, X, y, objectives, time_series=False):
         if any(not o.score_needs_proba for o in objectives):
             y_predicted = self._predict(X, y, pad=True) if time_series else self._predict(X)
         return y_predicted, y_predicted_proba
+
+    def optimize_threshold(self, X, y, y_pred_proba, objective):
```
angela97lin, line 140: Maybe this should actually go in `binary_classification_pipeline.py` if it's only used for binary problems?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1991,3 +1996,50 @@ def test_score_error_when_custom_objective_not_instantiated(problem_type, logist
...
+    objective = get_objective("F1", return_instance=True)
+    pipeline = BinaryClass(parameters={"Logistic Regression Classifier": {"n_jobs": 1}})
+    pipeline.fit(X, y)
+    assert pipeline.threshold is None
+    pred_proba = pipeline.predict_proba(X, y).iloc[:, 1]
+    pipeline.optimize_threshold(X, y, pred_proba, objective)
+    assert pipeline.threshold is not None
+
+
+def test_pipeline_thresholding_errors(make_data_type, X_y_multi, X_y_binary):
```
angela97lin, line 2017: Major nitpick but maybe we can mock fit/predict here since they're not necessary? :)

evalml/pipelines/classification_pipeline.py
bchen1116: Ah, was an accident. Thanks!

evalml/tests/pipeline_tests/test_pipelines.py
bchen1116: Oh I didn't even know these were a thing. Fixed!

evalml/pipelines/classification_pipeline.py
bchen1116: I put it in the classification pipeline class since it also should apply for TimeSeriesBinaryClassificationPipelines, which builds on ClassificationPipeline. Adding tests for that!

Issue #1954, opened 2021-03-10
Classification: data splitting errors if target is float


freddyaboulton:
### Binary repro

```python
from evalml.demos import load_breast_cancer, load_wine
from evalml.automl import AutoMLSearch

X, y = load_breast_cancer()
y = y.to_series().map(lambda target: -3.14 if target == "benign" else 4.2)

automl = AutoMLSearch(X, y, problem_type="binary")
```

```python
--> 73         for i, (train, valid) in enumerate(automl.data_splitter.split(X_pd, y_pd)):
     74             if pipeline.model_family == ModelFamily.ENSEMBLE and i > 0:
     75                 # Stacked ensembles do CV internally, so we do not run CV here for performance reasons.

~/miniconda3/envs/evalml/lib/python3.8/site-packages/sklearn/model_selection/_split.py in split(self, X, y, groups)
    330                 .format(self.n_splits, n_samples))
    331 
--> 332         for train, test in super().split(X, y, groups):
    333             yield train, test
    334 

~/miniconda3/envs/evalml/lib/python3.8/site-packages/sklearn/model_selection/_split.py in split(self, X, y, groups)
     78         X, y, groups = indexable(X, y, groups)
     79         indices = np.arange(_num_samples(X))
---> 80         for test_index in self._iter_test_masks(X, y, groups):
     81             train_index = indices[np.logical_not(test_index)]
     82             test_index = indices[test_index]

~/miniconda3/envs/evalml/lib/python3.8/site-packages/sklearn/model_selection/_split.py in _iter_test_masks(self, X, y, groups)
    691 
    692     def _iter_test_masks(self, X, y=None, groups=None):
--> 693         test_folds = self._make_test_folds(X, y)
    694         for i in range(self.n_splits):
    695             yield test_folds == i

~/miniconda3/envs/evalml/lib/python3.8/site-packages/sklearn/model_selection/_split.py in _make_test_folds(self, X, y)
    643         allowed_target_types = ('binary', 'multiclass')
    644         if type_of_target_y not in allowed_target_types:
--> 645             raise ValueError(
    646                 'Supported target types are: {}. Got {!r} instead.'.format(
    647                     allowed_target_types, type_of_target_y))

ValueError: Supported target types are: ('binary', 'multiclass'). Got 'continuous' instead.
```

### Multiclass repro

```python
X, y = load_wine()
mapping = {"class_0": 3.2, "class_1": 4.53, "class_2": 6.28}
[...]

PR #1959, opened 2021-03-10
Remove warning raised when numeric binary classification targets are not (0, 1)

### LINKED ISSUE #1927: Numeric binary classification target warning raised when target is not numeric
Repro:

```python

X, y = load_breast_cancer()

automl = AutoMLSearch(X, y, problem_type="binary", max_batches=1)
automl.search()
```

Output:

image

The purpose of the warning is to warn users when they have a numeric binary classification target whose unique values aren't 0,1 but the warning is always raised when the target is a string. I don't think that's intended.

angela97lin:
Closes #1927

Confirmed that it's okay to remove this warning:
```python

X, y = load_breast_cancer()
y = y.to_series()
y = y.map({"benign":50, "malignant":-100})
automl = AutoMLSearch(X, y, problem_type="binary", max_batches=10, ensembling=True) #test ensembler too
automl.search()
```

image


evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -48,47 +48,6 @@ def test_invalid_target_data_check_numeric_binary_classification_valid_float():
     assert invalid_targets_check.validate(X, y) == {"warnings": [], "errors": [], "actions": []}
 
 
-def test_invalid_target_data_check_numeric_binary_classification_error():
```
bchen1116, line 51: nitpick, but Is it possible to leave this in, but rather than checking for a warning, check that there isn't a warning thrown/the resulting warning list is empty? Just to ensure we have coverage that this case is supported

bchen1116 (approved):
LGTM! Left a testing suggestion, but nothing blocking.

chukarsten (approved):

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -134,13 +134,6 @@ def validate(self, X, y):
                     message_code=DataCheckMessageCode.TARGET_MULTICLASS_HIGH_UNIQUE_CLASS,
                     details=details).to_dict())
 
-        if len(value_counts) == 2 and is_supported_type:
```
freddyaboulton, line 137: Looks good! I agree we should keep the `TARGET_BINARY_INVALID_VALUES` code since we'll probably need to use it in #1954

freddyaboulton (approved):
Looks good @bchen1116 !

PR #1966, opened 2021-03-11
Add params to the data splitter for visibility

bchen1116:
With this change, when we call `automl.data_splitter`, we will see

```python
BalancedClassificationDataCVSplit(balanced_ratio=4, min_percentage=0.1,
                 min_samples=100, n_splits=3, random_seed=0, shuffle=True)
```
rather than 
```python
BalancedClassificationDataCVSplit(balanced_ratio=None, min_percentage=None,
                 min_samples=None, n_splits=3, random_seed=0, shuffle=True)
```

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,7 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
         * Added multiple oversampling and undersampling sampling methods as data splitters for imbalanced classification :pr:`1775`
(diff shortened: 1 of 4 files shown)


angela97lin (approved):
If I'm understanding correctly, the test in `evalml/tests/automl_tests/test_automl.py` tests this updated change, and then the test you added in `evalml/tests/preprocessing_tests/test_balanced_classification_data_splitter.py` is just to check for defaults / set attributes? LGTM!

PR #1968, opened 2021-03-11
Include recommended actions for each data check's set of actions

### LINKED ISSUE #1881: Include recommended actions for each data check's set of actions

angela97lin:
First half of #1881.

Addresses:
- `DataChecks`
- `UniquenessDataCheck`
- `SparsityDataCheck`
- `TargetLeakageDataCheck`
- `HighlyNullDataCheck`
- `IDColumnsDataCheck`
- `NoVarianceDataCheck`

These data checks are relatively simple as their recommended action is just to drop the problem column.

Currently, different data checks could result in the same data check action (drop a specific col). Should the `DataChecks` class address this and remove duplicates? 

In this PR, I address this question with "yes", because I imagine this will be useful when creating components--we don't want the second DropColumns component to error out when it can't find the column, but open to thoughts / concerns! Note that this doesn't affect different transformations on the same col (impute then drop).


evalml/tests/data_checks_tests/test_highly_null_data_check.py
```diff
@@ -96,41 +99,16 @@ def test_highly_null_data_check_input_formats():
...
+        "actions": [DataCheckAction(DataCheckActionCode.DROP_COL, details={"column": 0}).to_dict(),
+                    DataCheckAction(DataCheckActionCode.DROP_COL, details={"column": 1}).to_dict(),
+                    DataCheckAction(DataCheckActionCode.DROP_COL, details={"column": 2}).to_dict()]
     }
+    #  test Woodwork
+    ww_input = ww.DataTable(pd.DataFrame([[None, None, None, None, 0], [None, None, None, "hi", 5]]))
+    assert highly_null_check.validate(ww_input) == expected
 
     #  test 2D list
-    assert highly_null_check.validate([[None, None, None, None, 0], [None, None, None, "hi", 5]]) == {
```
angela97lin, line 103: Just cleaning up. We were using the same `expected` value for each of these cases.

evalml/data_checks/data_checks.py
```diff
@@ -96,6 +96,9 @@ def validate(self, X, y=None):
             messages_new = data_check.validate(X, y)
             messages["warnings"].extend(messages_new["warnings"])
             messages["errors"].extend(messages_new["errors"])
-            messages["actions"].extend(messages_new["actions"])
 
+            new_actions = messages_new["actions"]
+            for new_action in new_actions:
+                if new_action not in messages["actions"]:
```
bchen1116, line 102: Good catch!

bchen1116 (approved):
Looks good to me! I like that the data checks class handles removing duplicate columns names that we want to drop.

chukarsten (approved):
This looks solid as is.  If you want to add the explicit test to show the removal of anticipated redundant columns, that would be cool.  But you probably have better things to do lol

angela97lin:
@chukarsten There's no better way to spend time than to write more unit tests 😂  Thanks for the suggestion, added!

PR #1974, opened 2021-03-12
Remove build_docs CI job in favor of RTD GH builder

dsherry:
We were waiting on resolution of an RTD builder bug (timeout after ~60sec). Not resolved yet, but rerunning seems to fix that typically, so I think its time!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,27 +138,10 @@ jobs:
             - run: |
                 source test_python/bin/activate
                 codecov --required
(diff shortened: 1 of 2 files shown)


angela97lin (approved):

bchen1116 (approved):
✌️

ParthivNaresh (approved):
You can delete `install_dependencies_docs` under the commands section on line 35 if you want, no big deal otherwise!

PR #1975, opened 2021-03-12
Proof of Concept: Separating the Engine from AutoMLSearch

### LINKED ISSUE #1914: Spike: New Engine Design
The Engine api/implementation has been changing rapidly (particularly in #1814 and #1913) since #1838 was merged a little under two weeks ago!

Once the dust settles on the EngineBase and SequentialEngine implementations/api we should take our latest learnings from the ParallelEngine work (#1869) and see if we can improve upon our initial design.

Some questions to consider:

1. Engines are created within AutoML but need a reference to AutoML. Can we make the engines not need a reference to AutoML?
2. [train_and_score_pipeline](https://github.com/alteryx/evalml/blob/main/evalml/automl/engine/engine_base.py#L54) is pretty big. Can we break it up into manageable/reusable pieces?
3. Currently, each engine needs to implement its own "loop" for training/scoring a batch of pipelines. This makes it hard to reuse logic needed to train and score an individual pipeline across our engines, e.g. when to use the ensembling dataset, how to handle errors. It could be worth it to "invert" the design - Engine subclasses only need to implement how to fit and/or score a single pipeline and we have a single "loop" for how to fit and/or score a batch in the base class. 
4. How do we handle early stopping/keyboard interrupt the same way in sequential and parallel engines?
5. How do we handle max_iterations?  Currently the SequentialEngine receives a callback from the AutoMLSearch instance and the callback updates the state of the AutoMLSearch to determine whether the engine should continue processing pipelines.  This doesn't *seem* possible with the pipelines evaluated in parallel.
6. How do we test the parallel engine?  Running class instances that have had functions patched via unittest.mock have those patches "stripped" after they go through Dask.Client.map().

Acceptance Criteria:
Document that proposes changes to the engine design or justifies the current design for the medium-term.

### LINKED ISSUE #1972: Dask Logging
It seems likely that the logger needs to be refactored to accomodate a parallel, asynchronous engine.  If the current implementation writes to an actual file, the Dask or any other asynchronous implementation will struggle with writing to the same file asynchronously.

freddyaboulton:
### Pull Request Description

Fixes #1914. Fixes #1972 

I think there are good reasons to go this route:
1. Moves us away from using callbacks that have a reference to AutoMLSearch. The engine is solely responsible for submitting "work". AutoML is in charge of updating its progress and results.
2. Less code duplication. Our old interface required developers to implement logic for processing a whole batch of pipelines. There's a lot of core logic that goes into that implementation: when to use the ensemble split, when to terminate mid-batch, etc, that's hard to share across all engine implementations. With this approach, the engine only does "work". AutoML handles all the surrounding ml logic. 
3. Lets users "bring their own engine" by allowing the engine to be passed as an AutoMLSearch argument. There is also less code to write to implement a new engine. Developers/users only need to worry about how to train/score/evaluate a single pipeline. AutoMLSearch handles the batch-wide logic.

### Demo with dask

image

image

Cons (strategically placed at the bottom 😂):
1. Currently breaks how our console output/log works for the sequential case. I think we're going to have to do this for any approach we take though. See #1972. Edit: Pushed up a change to how the logging works that works well for general engines. The messages are written to the log once the evaluation job completes.
2. We need the engine to return the stacktrace if a computation fails. I think that's possible for the sequential case and dask but not sure how easy that is in the general case. 
3. Still thinking about others. I imagine I'll have a better idea as I get the tests to pass.

### Performance comparison: About 70% faster on AutoML

image

-----


evalml/automl/engine/ray_engine.py
```diff
@@ -0,0 +1,60 @@
+from evalml.automl.engine.engine_base import (
```
freddyaboulton, line 1: Ray basically works out-of-the-box! I'm including this mainly as a selling point for the proposed design. I realize ray integration is out-of-scope and could be done later.

image

evalml/automl/automl_search.py
```diff
@@ -562,14 +577,7 @@ def _should_continue(self):
         if self._interrupted:
             return False
 
-        # for add_to_rankings
```
freddyaboulton, line 565: These lines were added for `add_to_rankings` and baseline evaluation because the engine used to call the `_should_continue_callback`. The engine doesn't do that anymore so we can delete these lines!

freddyaboulton:
Performance tests are [here](https://alteryx.quip.com/ceF8AKJF3cYH/DaskEngine-Performance-Test-Results)!

Dask is 3.79 times faster on average!

https://alteryx.quip.com/-/blob/caQAAAyeftb/NUlMoP95tyleoZVGYMPG8A

evalml/automl/engine/dask_engine.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class DaskComputation(EngineComputation):
+    """A Future-like wrapper around jobs created by the DaskEngine."""
+
+    def __init__(self, dask_future):
+        self.work = dask_future
+        self.meta_data = {}
+
+    def done(self) -> bool:
```
bchen1116, line 20: Feels a little random to do type-hinting on certain functions, but not others. Not a big deal, but just my thoughts

evalml/automl/engine/sequential_engine.py
```diff
@@ -1,105 +1,57 @@
...
-                training_indices = [i for i in range(len(self.X_train)) if i not in self.ensembling_indices]
-                X = self.X_train.iloc[training_indices]
-                y = self.y_train.iloc[training_indices]
-            evaluation_result = EngineBase.train_and_score_pipeline(pipeline, self.automl, X, y)
-            new_pipeline_ids.append(self._post_evaluation_callback(pipeline, evaluation_result))
-            index += 1
-        return new_pipeline_ids
+    def done(self) -> bool:
+        """Is the computation done?"""
+        return True
```
bchen1116, line 20: Why do we auto-return just true here?

evalml/automl/engine/sequential_engine.py
```diff
@@ -1,105 +1,57 @@
...
-        fitted_pipelines = {}
-        for pipeline in pipelines:
-            try:
-                fitted_pipeline = EngineBase.train_pipeline(
-                    pipeline, self.X_train, self.y_train,
-                    self.automl.optimize_thresholds,
-                    self.automl.objective
-                )
+    def cancel(self) -> None:
+        """Cancel the current computation."""
```
bchen1116, line 31: Was this purposely left empty?

bchen1116 (commented):
Skimmed through the implementation code and left some initial questions for clarifying purposes. Will revisit on Monday. This PR is a doozy 😅  but great work so far!

evalml/automl/engine/dask_engine.py
freddyaboulton: I'll get rid of them! I added them at first when I was sketching stuff out and forgot every place hehe

evalml/automl/engine/dask_engine.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get_result(self):
+        """Get the computation result.
+        Will block until the computation is finished.
+
+        Raises Exception: If computation fails. Returns traceback.
+        """
+        return self.work.result()
+
+    def cancel(self) -> None:
```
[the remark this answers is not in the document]
freddyaboulton: Right now we always send out a full batch of pipelines even if we know that the search will finish before the full batch (because of `max_iterations`). In the situations where the search finishes sooner than expected, I think we should cancel the jobs that are still running. This functionality isn't in the `AutoMLSearch` loop yet though

angela97lin (commented):
Took a look over the impl, and this is some really awesome stuff! Left a few nit-picks and general comments for now :')

evalml/automl/callbacks.py
```diff
@@ -15,23 +15,6 @@ def raise_error_callback(exception, traceback, automl, **kwargs):
     raise exception
 
 
-def log_and_save_error_callback(exception, traceback, automl, **kwargs):
```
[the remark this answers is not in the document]
freddyaboulton: Yea I don't think we should support the "saving" callbacks anymore because the expectation is that they're able to modify the global automl search state from within the evaluation function which I don't think is a good idea during parallel evaluation. 

I think we could support something like this in the future if we return the exception/traceback from the remote worker but I don't think it's a high priority. The exception and traceback are displayed in the log already except in the "silent" callback so I don't see the value-add of storing the exception itself.

evalml/automl/engine/dask_engine.py
```diff
@@ -0,0 +1,83 @@
...
+        """Is the computation done?"""
+        return self.work.done()
+
+    def get_result(self):
+        """Get the computation result.
+        Will block until the computation is finished.
+
+        Raises Exception: If computation fails. Returns traceback.
+        """
+        return self.work.result()
```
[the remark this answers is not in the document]
freddyaboulton: In dask, when you call `result` it returns the result of the function you sent to the worker. So the value of our `get_result` depends on the job. For pipeline evaluation it's the scores, pipeline class, and log messages. For pipeline scoring it's the scores and for pipeline training it is just the trained pipeline. 

Since the computation itself is could be anything I think the docsting is adequate ("Get the computation result.").

evalml/automl/engine/sequential_engine.py
freddyaboulton: Yea we don't need to cancel the sequential jobs in the same way we would for a dask/other parallel library job so I'm just leaving it empty because codecov doesn't like `pass` statements.

evalml/automl/callbacks.py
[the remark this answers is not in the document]
freddyaboulton: Yes!

evalml/automl/automl_search.py
```diff
@@ -707,7 +720,7 @@ def _check_for_high_variance(self, pipeline, cv_scores, threshold=0.2):
         if cv_scores_std != 0 and cv_scores_mean != 0:
             high_variance_cv = bool(abs(cv_scores_std / cv_scores_mean) > threshold)
         if high_variance_cv:
-            logger.warning(f"High coefficient of variation (cv >= {threshold}) within cross validation scores. {pipeline_name} may not perform as estimated on unseen data.")
+            logger.warning(f"\t\tHigh coefficient of variation (cv >= {threshold}) within cross validation scores.\n\t\t{pipeline_name} may not perform as estimated on unseen data.")
```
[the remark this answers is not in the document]
freddyaboulton: Good point. I added the extra tab cause I thought it would be nice to "nest" the pipeline log underneath the batch number during automl search but it's probably equally good to remove the extra tab.

evalml/automl/automl_search.py
```diff
@@ -880,7 +895,26 @@ def train_pipelines(self, pipelines):
             Note that the any pipelines that error out during training will not be included in the dictionary
             but the exception and stacktrace will be displayed in the log.
         """
-        return self._engine.train_batch(pipelines)
+        fitted_pipelines = {}
```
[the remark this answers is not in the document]
freddyaboulton: Yea we don't support duplicate names! Good catch, in `main` we check if the pipeline names are unique before doing work and raise an exception if so. I'll add that here too!

evalml/automl/engine/dask_engine.py
```diff
@@ -0,0 +1,83 @@
...
+                                         optimize_thresholds=automl_data.optimize_thresholds,
+                                         objective=automl_data.objective)
+        return DaskComputation(dask_future)
+
+    def submit_scoring_job(self, automl_data, pipeline, X, y, objectives):
+        X, y = self.send_data_to_cluster(X, y)
+        dask_future = self.client.submit(score_pipeline, pipeline=pipeline,
+                                         X=X, y=y, objectives=objectives)
+        computation = DaskComputation(dask_future)
+        computation.meta_data["pipeline_name"] = pipeline.name
```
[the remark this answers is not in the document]
freddyaboulton: Not on purpose! It is an oversight. At first, I wrote the dask engine by accessing custom_name but our regression baseline pipelines dont have `custom_name` set (they have `_name` set instead) so stuff broke. I changed the dask engine but since there aren't any ray unit tests I didn't change the ray engine hehe

evalml/automl/automl_search.py
```diff
@@ -896,4 +930,31 @@ def score_pipelines(self, pipelines, X_holdout, y_holdout, objectives):
...
+        computations = []
+        for pipeline in pipelines:
+            computations.append(self._engine.submit_scoring_job(self.automl_data, pipeline, X_holdout, y_holdout, objectives))
+
+        while computations:
+            computation = computations.pop(0)
+            if computation.done():
+                pipeline_name = computation.meta_data["pipeline_name"]
+                try:
+                    scores[pipeline_name] = computation.get_result()
```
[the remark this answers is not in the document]
freddyaboulton: It's not exactly the same because the results are not the same between the training job and the scoring job. So when we call `get_result()` in the scoring job we expect to just get back a dictionary of scores as opposed to a pipeline class. 

I guess we could modify the scoring job to return the pipeline that was scored as well that would side-step the need for the metadata.

evalml/automl/engine/engine_base.py
```diff
@@ -7,177 +7,204 @@
...
+        """Cancel the computation."""
 
-        Returns:
-            list (int): A list of the new pipeline IDs which were created by the AutoML search.
-        """
 
-    @abstractmethod
-    def train_batch(self, pipelines):
-        """Train a batch of pipelines using the current dataset.
+class JobLogger:
```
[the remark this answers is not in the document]
freddyaboulton: I think it can go anywhere! But since we only need it for one pipeline job (that's also defined in engine base) I figured I'd put it here as opposed to its own file.

evalml/automl/engine/engine_base.py
```diff
@@ -7,177 +7,204 @@
...
+                          X_threshold_tuning, y_threshold_tuning)
+    return cv_pipeline
+
+
+def train_and_score_pipeline(pipeline, automl_data, full_X_train, full_y_train, logger):
+    """Given a pipeline, config and data, train and score the pipeline and return the CV or TV scores
+
+    Arguments:
+        pipeline (PipelineBase): The pipeline to score
+        automl_data (AutoMLSearch): The AutoML search, used to access config and for the error callback
```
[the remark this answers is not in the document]
freddyaboulton: Thank you! hehe

evalml/automl/engine/engine_base.py
```diff
@@ -7,177 +7,204 @@
...
 
-        Returns:
-            list (int): A list of the new pipeline IDs which were created by the AutoML search.
-        """
 
-    @abstractmethod
-    def train_batch(self, pipelines):
-        """Train a batch of pipelines using the current dataset.
+class JobLogger:
+    """Mimics the behavior of a RootLogger but stores all messages rather than actually logging them.
```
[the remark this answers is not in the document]
freddyaboulton: I'm gonna go with `logging.Logger`. Good call!

evalml/automl/engine/engine_base.py
```diff
@@ -7,177 +7,204 @@
...
+    """Given a pipeline, config and data, train and score the pipeline and return the CV or TV scores
+
+    Arguments:
+        pipeline (PipelineBase): The pipeline to score
+        automl_data (AutoMLSearch): The AutoML search, used to access config and for the error callback
+        full_X_train (ww.DataTable): Training features
+        full_y_train (ww.DataColumn): Training target
+
+    Returns:
+        dict: A dict containing cv_score_mean, cv_scores, training_time and a cv_data structure with details.
```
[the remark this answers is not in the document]
freddyaboulton: Good catch! Updating!

evalml/automl/callbacks.py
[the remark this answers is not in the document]
bchen1116: Makes sense!

evalml/automl/engine/dask_engine.py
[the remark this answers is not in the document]
bchen1116: Ahhh got it got it. that makes sense, thanks!

evalml/automl/engine/ray_engine.py
freddyaboulton: I'll get rid of it before merge!

evalml/automl/automl_search.py
```diff
@@ -313,13 +323,15 @@ def __init__(self,
...
-                                        self,
-                                        should_continue_callback=self._should_continue,
-                                        pre_evaluation_callback=self._pre_evaluation_callback,
-                                        post_evaluation_callback=self._post_evaluation_callback)
+        if not engine:
+            self._engine = SequentialEngine()
+        else:
+            self._engine = engine
+
+        self.automl_data = AutoMLData(self.ensembling_indices,
```
kmax12, line 331: take this with a grain of salt since i only am doing a quick pass, but i wonder if `AutoMLConfig` or something like that might be a better name for this. When I see "data" I assume it is referring to `X` and `y`

evalml/automl/automl_search.py
```diff
@@ -538,8 +556,8 @@ def _find_best_pipeline(self):
                     train_indices = self.data_splitter.transform_sample(X_train, y_train)
                     X_train = X_train.iloc[train_indices]
                     y_train = y_train.iloc[train_indices]
-                best_pipeline = self._engine.train_pipeline(best_pipeline, X_train, y_train,
-                                                            self.optimize_thresholds, self.objective)
+                best_pipeline = train_pipeline(best_pipeline, X_train, y_train,
```
kmax12, line 559: i wonder if we want this to still be called through the engine?

a use case to imagine is that you are running evalml on your laptop and you've connected it to a remote dask cluster that has plenty of compute (perhaps the `X` and `y` are reference to dask dataframes where the data is remote too). you want all the training to happen on the remote dask cluster. 

if i understand the code correctly, right now the search phase will happen on the remote cluster, but the final training of the pipeline would happen locally.

evalml/automl/automl_search.py
```diff
@@ -494,12 +498,26 @@ def search(self, show_iteration_plot=True):
...
             try:
-                new_pipeline_ids = self._engine.evaluate_batch(current_batch_pipelines)
+                computations = []
+                new_pipeline_ids = []
+                log_title(logger, f"Evaluating Batch Number {self._get_batch_number()}")
+                for pipeline in current_batch_pipelines:
+                    self._pre_evaluation_callback(pipeline)
+                    computation = self._engine.submit_evaluation_job(self.automl_data, pipeline, self.X_train, self.y_train)
+                    computations.append(computation)
+                while self._should_continue() and len(computations) > 0:
```
kmax12, line 508: what is the behavior of this when using the dask engine and all pipelines in a batch are currently computing? does it just iterate as fast as it can over and over trying to find whatever computation has completed? 

I wonder if this code can be simplified using a method like: https://docs.dask.org/en/latest/futures.html#distributed.as_completed

kmax12 (commented):
interesting stuff. left some comments based on a quick review

evalml/automl/automl_search.py
freddyaboulton: Good call! That was my intention but I think I forgot to actually make the change hehe.

evalml/automl/automl_search.py
freddyaboulton: You're right about the current behavior! I thought about using/writing our own `as_completed` but the one tricky thing is that we want to stop the search as soon as one of our stopping criteria is met but `as_completed` will only pull off futures if they're done. I think this will make it hard to respect time based stopping rules. 

For example, let's say the user has a 2 hour stop limit and fires off four jobs. The first three finish in 90 minutes but the fourth one won't finish until 6 hours.

If we do this:

```python
for computation in as_completed(computations):
    if not self._should_continue():
         break
```

We actually won't reach the break until hour 6 because that's when the final future finished.

I think we can definitely explore ways to consolidate/simplify the logic for pulling futures off of our queue but I'd rather leave that improvement for a separate PR.

evalml/automl/automl_search.py
```diff
@@ -880,7 +898,27 @@ def train_pipelines(self, pipelines):
...
+                try:
+                    fitted_pipeline = computation.get_result()
+                    fitted_pipelines[fitted_pipeline.name] = fitted_pipeline
+                except Exception as e:
+                    logger.error(f'Train error for {pipeline.name}: {str(e)}')
+                    tb = traceback.format_tb(sys.exc_info()[2])
+                    logger.error("Traceback:")
+                    logger.error("\n".join(tb))
+            else:
+                computations.append(computation)
```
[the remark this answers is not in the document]
freddyaboulton: It's not exactly the same because the thing you get back from `get_result` is not the same between training and scoring and the logic for handling exceptions is not the same as well.

This might be a good spot for @kmax12 's suggestion to use `as_completed` or something like it.

If we do that, I think this will look like:

```python
    for computation in as_completed(computations):
                try:
                    fitted_pipeline = computation.get_result()
                    fitted_pipelines[fitted_pipeline.name] = fitted_pipeline
                except Exception as e:
                    logger.error(f'Train error for {pipeline.name}: {str(e)}')
                    tb = traceback.format_tb(sys.exc_info()[2])
                    logger.error("Traceback:")
                    logger.error("\n".join(tb))
```

I think at that point we're not reusing the while-loop logic so the differences between the two methods might be clearer.

evalml/tests/automl_tests/dask_testing.py
```diff
@@ -0,0 +1,102 @@
...
+
+from evalml.automl.utils import AutoMLData
+from evalml.exceptions import PipelineScoreError
+from evalml.objectives.utils import get_objective
+from evalml.pipelines import BinaryClassificationPipeline
+from evalml.preprocessing.data_splitters import TrainingValidationSplit
+
+
+# Top-level replacement for AutoML object to supply data for testing purposes.
+def err_call(*args, **kwargs):
```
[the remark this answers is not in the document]
freddyaboulton: I think providing just a docstring will make codecov happy so I'll push that up now.

evalml/automl/engine/sequential_engine.py
freddyaboulton: I should have added some comments, this is confusing 🙈 

The issue is that in order to separate the engine from the AutoMLSearch loop, we need the sequential computations to behave the same way as concurrent computations from AutoMLSearch's point-of-view. One way to do this is by delaying the computation in the sequential engine until `get_result` is called. Since AutoMLSearch will call `get_result` only when the computation is "done", by always returning `True` we make sure that `get_result` is called in the order that the jobs are submitted. So the computations happen sequentially!

bchen1116 (approved):
Looks good to me! Left two comments about renaming and adding doc string, but the tests are very solid! Love the comparisons between the performance between parallel and sequential.

evalml/automl/automl_search.py
kmax12: i wonder if we need some sort of sleep call in the loop then. if you have a bunch of pipelines training for a while, i dont think you want this loop iterating as fast as it can.

evalml/automl/engine/engine_base.py
```diff
@@ -7,177 +7,205 @@
...
-
-            evaluation_entry = {"all_objective_scores": ordered_scores, "score": score, 'binary_classification_threshold': None}
-            if is_binary(automl.problem_type) and cv_pipeline is not None and cv_pipeline.threshold is not None:
-                evaluation_entry['binary_classification_threshold'] = cv_pipeline.threshold
-            cv_data.append(evaluation_entry)
-        training_time = time.time() - start
-        cv_scores = pd.Series([fold['score'] for fold in cv_data])
-        cv_score_mean = cv_scores.mean()
-        logger.info(f"\tFinished cross validation - mean {automl.objective.name}: {cv_score_mean:.3f}")
-        return {'cv_data': cv_data, 'training_time': training_time, 'cv_scores': cv_scores, 'cv_score_mean': cv_score_mean}
```
dsherry, line 183: I'm assuming the impl here didn't change much?

evalml/automl/utils.py
```diff
@@ -110,3 +112,8 @@ def check_all_pipeline_names_unique(pipelines):
         plural, tense = ("s", "were") if len(duplicate_names) > 1 else ("", "was")
         duplicates = ", ".join([f"'{name}'" for name in sorted(duplicate_names)])
         raise ValueError(f"All pipeline names must be unique. The name{plural} {duplicates} {tense} repeated.")
+
+
+AutoMLData = namedtuple("AutoMLData", ["ensembling_indices", "data_splitter", "problem_type",
```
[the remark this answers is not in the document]
dsherry: Yep I also prefer `AutoMLConfig`

evalml/tests/automl_tests/dask_testing.py
```diff
@@ -0,0 +1,102 @@
+import time
```
dsherry, line 1: Style nit-pick: call this file `dask_test_utils.py` instead of `dask_testing.py`

evalml/tests/automl_tests/dask_testing.py
```diff
@@ -0,0 +1,102 @@
...
+error_callback = err_call
+random_seed = 0
+automl_data = AutoMLData(ensembling_indices=ensembling_indices,
+                         data_splitter=data_splitter,
+                         problem_type=problem_type,
+                         objective=objective,
+                         additional_objectives=additional_objectives,
+                         optimize_thresholds=optimize_thresholds,
+                         error_callback=error_callback,
+                         random_seed=random_seed)
```
dsherry, line 30: To avoid strange errors arising from accidentally mutating global state: could you please create this stuff in a test fixture, which runs once per test, instead of once at import-time as things stand currently?

You could even move it to [evalml/tests/conftest.py](https://github.com/alteryx/evalml/blob/main/evalml/tests/conftest.py).

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -82,7 +82,7 @@ def add_result_callback(results, trained_pipeline, automl_obj, counts=counts):
                           add_result_callback=add_result_callback, n_jobs=1)
     automl.search()
 
-    assert counts["start_iteration_callback"] == max_iterations
+    assert counts["start_iteration_callback"] == len(get_estimators("regression")) + 1
```
dsherry, line 85: Interesting, why did this have to change? Were we not calling this for the baseline pipeline before? That seems like a positive change to me.

evalml/tests/automl_tests/test_dask_engine.py
```diff
@@ -0,0 +1,215 @@
...
+    TestSVMPipeline,
+    automl_data
+)
+
+
+@pytest.mark.usefixtures("X_y_binary_cls")
+class TestDaskEngine(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls) -> None:
```
dsherry, line 28: Remove the ` -> None` ?

evalml/tests/automl_tests/test_dask_engine.py
```diff
@@ -0,0 +1,215 @@
...
+        cls.client = Client()
```
dsherry, line 29: Nice. This starts a local dask cluster, once for all of the tests in this class, correct? Any idea how many cores it'll take, how many workers it'll have?

evalml/tests/automl_tests/test_dask_engine.py
```diff
@@ -0,0 +1,215 @@
...
+        seq_eval_results = score_pipelines(pipelines, SequentialEngine())
+        seq_scores = [s["Log Loss Binary"] for s in seq_eval_results]
+
+        # Check there are the proper number of pipelines and all their scores are same.
+        assert len(par_eval_results) == len(pipelines)
+        assert set(par_scores) == set(seq_scores)
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.client.close()
```
dsherry, line 215: Very nice

I suggest moving this to be at the top near setup so its obvious to new readers what's going on.

dsherry (approved):
@freddyaboulton this is amazing! So many good changes :D the dask engine code looks 💯 . The `EngineComputation` abstraction is super helpful. Well done to you and to @chukarsten 😁 

I don't have many blocking comments. I agree with what Max said about adding a `sleep` in the main loop. I left a couple comments about the keyboard interrupt stuff in the main loop(s). I do think we should call that config object `AutoMLConfig` as suggested. And I left a couple comments on the test utils.

Since this is a significant change, please run the performance tests once more before merging as a protective measure.

evalml/automl/automl_search.py
```diff
@@ -313,13 +323,15 @@ def __init__(self,
...
 
-        self._engine = SequentialEngine(self.X_train,
-                                        self.y_train,
-                                        self.ensembling_indices,
-                                        self,
-                                        should_continue_callback=self._should_continue,
-                                        pre_evaluation_callback=self._pre_evaluation_callback,
-                                        post_evaluation_callback=self._post_evaluation_callback)
+        if not engine:
+            self._engine = SequentialEngine()
```
dsherry, line 327: @freddyaboulton @chukarsten : 

I am thinking that sooner or later we should switch this to default to `DaskEngine` with a local dask cluster. What would we need to do in order to make that happen? By my accounting: make sure the team feels good about this move, list dask in our requirements, update our documentation to explain what's going on and how to configure a remote cluster, and confirm that the perf testing we've done is adequate to make this change.

evalml/tests/automl_tests/test_dask_engine.py
chukarsten: @dsherry instantiating with no arguments created a Client() based off a LocalCluster() with number of workers related to the number of available cores on the machine running it.

evalml/automl/automl_search.py
chukarsten: I was kind of thinking the same thing working through all this.  It seems that, in the grand scheme of things, a DaskEngine with 1 worker eliminates the need for a SequentialEngine.  And yeah, that sounds about right, but the proper documentation for how to limit the DaskEngine to a certain number of workers and perhaps an expectation of how many default workers there should be are required. 

So currently the AutoMLSearch is designed to take in an `engine` argument.  The `EngineBase` object expects a `client` argument, in the case of a `DaskEngine`.  And the client object, when it's not run as a default instance, takes a `cluster` argument.  The `cluster` is where you lay out all the details behind the resources and amount of parallelism you want to use.  As such, I think we need to answer the questions 1.) what are the default settings for workers we want for the DaskEngine? and 2.) what are the default settings for workers we want for the AutoMLSearch?  We could either go the way of trying to make it as fast as possible for the user by default or protect their resource by default.  I think whichever of the two philosophies (faster vs. resource conservative) we choose, we should probably be consistent between DaskEngine and AutoMLSearch.

evalml/automl/automl_search.py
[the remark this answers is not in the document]
freddyaboulton: Filed https://github.com/alteryx/evalml/issues/2096 to track consolidating these two while loops!

evalml/automl/automl_search.py
```diff
@@ -880,7 +898,27 @@ def train_pipelines(self, pipelines):
...
         """
-        return self._engine.train_batch(pipelines)
+        check_all_pipeline_names_unique(pipelines)
+        fitted_pipelines = {}
+        computations = []
+        for pipeline in pipelines:
+            computations.append(self._engine.submit_training_job(self.automl_data, pipeline, self.X_train, self.y_train))
+
+        while computations:
+            computation = computations.pop(0)
```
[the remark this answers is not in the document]
freddyaboulton: Filed #2096!

evalml/tests/automl_tests/test_automl_search_regression.py
freddyaboulton: Correct!

evalml/automl/engine/engine_base.py
freddyaboulton: Correct!

evalml/automl/automl_search.py
```diff
@@ -624,7 +635,11 @@ def _add_baseline_pipelines(self):
             max_delay = self.problem_configuration['max_delay']
             baseline = pipeline_class(parameters={"pipeline": {"gap": gap, "max_delay": max_delay},
                                                   "Time Series Baseline Estimator": {"gap": gap, "max_delay": max_delay}})
-        self._engine.evaluate_batch([baseline])
+        self._pre_evaluation_callback(baseline)
+        logger.info(f"Evaluating Baseline Pipeline: {baseline.name}")
+        computation = self._engine.submit_evaluation_job(self.automl_data, baseline, self.X_train, self.y_train)
+        data, pipeline, job_log = computation.get_result()
```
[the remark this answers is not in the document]
freddyaboulton: Done!

evalml/automl/automl_search.py
```diff
@@ -494,12 +498,26 @@ def search(self, show_iteration_plot=True):
...
+                    computation = self._engine.submit_evaluation_job(self.automl_data, pipeline, self.X_train, self.y_train)
+                    computations.append(computation)
+                while self._should_continue() and len(computations) > 0:
+                    computation = computations.pop(0)
+                    if computation.done():
+                        data, pipeline, job_log = computation.get_result()
+                        pipeline_id = self._post_evaluation_callback(pipeline, data, job_log)
+                        new_pipeline_ids.append(pipeline_id)
+                    else:
+                        computations.append(computation)
```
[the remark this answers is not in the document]
freddyaboulton: Great suggestion! Pushing up this refactor now.

freddyaboulton:
Updated perf tests results are [here](https://alteryx.quip.com/ceF8AKJF3cYH/DaskEngine-Performance-Test-Results#caQACAxNn5E). Same performance improvements after making the requested changes so I think this is good to merge!

PR #1981, opened 2021-03-15
Update AutoMLSearch with `actions` parameter

### LINKED ISSUE #1883: Update AutoMLSearch with `actions` parameter

angela97lin:
Closes #1883

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -34,7 +34,10 @@ from evalml.pipelines import (
     TimeSeriesBaselineRegressionPipeline
 )
 from evalml.pipelines.components.utils import get_estimators
-from evalml.pipelines.utils import make_pipeline
+from evalml.pipelines.utils import (
+    _make_component_list_from_actions,
+    make_pipeline
+)
 from evalml.preprocessing import split_data
 from evalml.problem_types import ProblemTypes, handle_problem_types
 from evalml.tuners import SKOptTuner
@@ -83,6 +86,7 @@ class AutoMLSearch:
                  problem_configuration=None,
                  train_best_pipeline=True,
                  pipeline_parameters=None,
+                 data_check_actions=None,
                  _ensembling_split_size=0.2,
                  _pipelines_per_batch=5):
         """Automated pipeline search
@@ -163,6 +167,8 @@ class AutoMLSearch:
             train_best_pipeline (boolean): Whether or not to train the best pipeline before returning it. Defaults to True.
 
             pipeline_parameters (dict): A dict of the parameters used to initalize a pipeline with.
+
+            data_check_actions (list(DataCheckActions)): A list of DataCheckAction objects which specify actions to take to further configure the AutoML object or pipelines created. Defaults to None.
 
             _ensembling_split_size (float): The amount of the training data we'll set aside for training ensemble metalearners. Only used when ensembling is True.
                 Must be between 0 and 1, exclusive. Defaults to 0.2
@@ -269,11 +275,17 @@ class AutoMLSearch:
         self.search_iteration_plot = None
         self._interrupted = False
 
+        if data_check_actions is not None:
+            self.data_check_actions = data_check_actions
+            self._prepended_components = _make_component_list_from_actions(self.data_check_actions)
+
         if self.allowed_pipelines is None:
             logger.info("Generating pipelines to search over...")
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
-            self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
+            self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type,
+                                                    custom_hyperparameters=self.pipeline_parameters,
+                                                    components_to_prepend=self._prepended_components) for estimator in allowed_estimators]
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -104,7 +104,7 @@ def _get_pipeline_base_class(problem_type):
         return TimeSeriesMulticlassClassificationPipeline
 
 
-def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
+def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None, components_to_prepend=None):
     """Given input data, target data, an estimator class and the problem type,
         generates a pipeline class with a preprocessing chain which was recommended based on the inputs.
         The pipeline will be a subclass of the appropriate pipeline base class for the specified problem_type.
@@ -116,6 +116,7 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
         problem_type (ProblemTypes or str): Problem type for pipeline to generate
         custom_hyperparameters (dictionary): Dictionary of custom hyperparameters,
             with component name as key and dictionary of parameters as the value
+        components_to_prepend (list(ComponentBase)): List of components to prepend to the pipeline. Defaults to None.
 
     Returns:
         class: PipelineBase subclass with dynamically generated preprocessing components and specified estimator
@@ -129,6 +130,8 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
         raise ValueError(f"{estimator.name} is not a valid estimator for problem type")
     preprocessing_components = _get_preprocessing_components(X, y, problem_type, estimator)
     complete_component_graph = preprocessing_components + [estimator]
+    if complete_component_graph is not None:
+        complete_component_graph = components_to_prepend + complete_component_graph
 
     if custom_hyperparameters and not isinstance(custom_hyperparameters, dict):
         raise ValueError(f"if custom_hyperparameters provided, must be dictionary. Received {type(custom_hyperparameters)}")


evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -32,8 +32,8 @@ def __init__(self,
         self.max_iterations = max_iterations
         self._tuner_class = tuner_class or SKOptTuner
         self._tuners = {}
-        for p in self.allowed_pipelines:
-            self._tuners[p.name] = self._tuner_class(p.hyperparameters, random_seed=self.random_seed)
+        for pipeline in self.allowed_pipelines:
```
angela97lin, line 35: No functional change, but I noticed it's hard to use `pdb` in this loop because `p` gets mistaken for a command. Can revert if confusing 😂

freddyaboulton: Good call!

evalml/pipelines/utils.py
```diff
@@ -267,5 +270,5 @@ def _make_component_list_from_actions(actions):
     components = []
     for action in actions:
         if action.action_code == DataCheckActionCode.DROP_COL:
-            components.append(DropColumns(columns=action.details["columns"]))
+            components.append((DropColumns, {"columns": action.details["columns"]}))
```
freddyaboulton, line 273: @angela97lin I think there's a mismatch. This code is expecting `action.details["columns"]` but our data checks populate `details={"column": col_name}`, e.g. the id column data check, no variance data check. I think we should make them all be "columns" cause that's what the drop columns expects.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -106,7 +106,8 @@ def test_make_pipeline_all_nan_no_categoricals(input_type, problem_type):
 
 @pytest.mark.parametrize("input_type", ["pd", "ww"])
 @pytest.mark.parametrize("problem_type", ProblemTypes.all_problem_types)
-def test_make_pipeline(input_type, problem_type):
+@pytest.mark.parametrize("prepended_components", [None, [], [DropColumns], [DropColumns, DropColumns]])
```
freddyaboulton, line 109: Our linear component graph structure can't handle duplicate components with the same name. This means we can't have two "drop columns" actions in AutoMLSearch: 

```python
from evalml.data_checks import DataCheckAction, DataCheckActionCode

X, y = load_breast_cancer()

actions = [DataCheckAction(DataCheckActionCode.DROP_COL, details={"columns": ["mean radius"]}),
           DataCheckAction(DataCheckActionCode.DROP_COL, details={"columns": ["mean texture"]})]
automl = AutoMLSearch(X, y, problem_type="binary", data_check_actions=actions)
automl.search()
```

```python
Using default limit of max_batches=1.

Generating pipelines to search over...
*****************************
* Beginning pipeline search *
*****************************

Optimizing for Log Loss Binary. 
Lower score is better.

Using SequentialEngine to train and score pipelines.
Searching up to 1 batches for a total of 9 pipelines. 
Allowed model families: random_forest, extra_trees, decision_tree, catboost, lightgbm, linear_model, xgboost

Batch 1: (1/9) Mode Baseline Binary Classification P... Elapsed:00:00
	Starting cross validation
	Finished cross validation - mean Log Loss Binary: 12.868
~/sources/evalml/evalml/pipelines/component_graph.py in instantiate(self, parameters)
     77             try:
---> 78                 new_component = component_class(**component_parameters, random_seed=self.random_seed)
     79             except (ValueError, TypeError) as e:

~/sources/evalml/evalml/pipelines/components/transformers/column_selectors.py in __init__(self, columns, random_seed, **kwargs)
     15         if columns and not isinstance(columns, list):
---> 16             raise ValueError(f"Parameter columns must be a list. Received {type(columns)}.")
     17 

ValueError: Parameter columns must be a list. Received <class 'str'>.

The above exception was the direct cause of the following exception:

<ipython-input-50-39598b64bdb5> in <module>
     13            DataCheckAction(DataCheckActionCode.DROP_COL, details={"columns": ["mean texture"]})]
     14 automl = AutoMLSearch(X, y, problem_type="binary", data_check_actions=actions)
---> 15 automl.search()

~/sources/evalml/evalml/automl/automl_search.py in search(self, show_iteration_plot)
    508             try:
    509                 if not loop_interrupted:
--> 510                     current_batch_pipelines = self._automl_algorithm.next_batch()
    511             except StopIteration:
    512                 logger.info('AutoML Algorithm out of recommendations, ending')

~/sources/evalml/evalml/automl/automl_algorithm/iterative_algorithm.py in next_batch(self)
     63         next_batch = []
     64         if self._batch_number == 0:
---> 65             next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_seed=self.random_seed)
     66                           for pipeline_class in self.allowed_pipelines]
     67 

~/sources/evalml/evalml/automl/automl_algorithm/iterative_algorithm.py in <listcomp>(.0)
     63         next_batch = []
     64         if self._batch_number == 0:
---> 65             next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_seed=self.random_seed)
     66                           for pipeline_class in self.allowed_pipelines]
     67 

~/sources/evalml/evalml/pipelines/classification_pipeline.py in __init__(self, parameters, random_seed)
     23         """
     24         self._encoder = LabelEncoder()
---> 25         super().__init__(parameters, random_seed=random_seed)
     26 
     27     def fit(self, X, y):

~/sources/evalml/evalml/pipelines/pipeline_base.py in __init__(self, parameters, random_seed)
     77         else:
     78             self._component_graph = ComponentGraph(component_dict=self.component_graph, random_seed=self.random_seed)
---> 79         self._component_graph.instantiate(parameters)
     80 
     81         self.input_feature_names = {}

~/sources/evalml/evalml/pipelines/component_graph.py in instantiate(self, parameters)
     80                 self._is_instantiated = False
     81                 err = "Error received when instantiating component {} with the following arguments {}".format(component_name, component_parameters)
---> 82                 raise ValueError(err) from e
     83 
     84             component_instances[component_name] = new_component

ValueError: Error received when instantiating component Drop Columns Transformer with the following arguments {'columns': 'mean texture'}
```

I think this should be resolved before merge. Our data checks are currently expected to output multiple actions (id, highly null, target leakage) but if AutoML can't support that after this PR then adding the actions param to AutoMLSearch seems premature imo.

Let me know what you think!

freddyaboulton (changes requested):
@angela97lin I think this looks good! I think we need to figure out how to let AutoMLSearch handle multiple drop columns actions. I think we'll have to consolidate into a single drop columns transformer. The other thing I noticed is that `_make_component_list_from_actions` expects a `columns` field but the data checks populate a field called `column`

evalml/pipelines/utils.py
bchen1116: Agree here! Also, would it be possible to also condense these same actions into a singular component? For instance, if `_make_component_list_from_actions` returns a list of 5 `DropColumns` tuples, each referring to a different column, I'd prefer to condense the columns we need to drop into a list and pass that to a single `DropColumns` component. This would keep the component graph cleaner for the user as well.

bchen1116 (commented):
Left a comment, but agree with Freddy's points that there are things we need to address before moving this forward.

evalml/pipelines/utils.py
angela97lin: Oopsies, thank you for catching this!

@bchen1116 and @freddyaboulton, thank you both for commenting RE: consolidating columns into one component. I think the initial idea was that it would be easier to see if there were multiple components, as opposed to combining a bunch of seemingly random columns together (ex: one ID col, one highly null col), and to map an action to a component. But I hear both of ya, it could be clearer if we just combined everything... Would love to discuss a bit post-standup before making the change heh

evalml/tests/pipeline_tests/test_pipeline_utils.py
angela97lin: Thank you for pointing this out! I was scratching my head as to why this test seemed to pass, but it's `search` that breaks. If we don't choose to combine components, or even if we do, a possible solution is to add custom names like 'Drop Columns Selector from DataCheckAction' or something 🤔

evalml/pipelines/utils.py
freddyaboulton: @angela97lin Sounds good! My main concern with having one component for each action is that our linear pipeline graph doesn't support it. To your point, I don't think combining all actions into a single component is the only solution. I think making sure they each have unique ways is also a possible solution. We could also consider a hybrid approach where we group all the drop columns per data check, e.g. the IdColumnDatacheck produces one component for dropping all id columns called "Drop Id Columns."

evalml/tests/pipeline_tests/test_pipeline_utils.py
freddyaboulton: We should probably file an issue for this too hehe I think we should handle the duplicate component name case a bit better hehe

freddyaboulton: Just filed https://github.com/alteryx/evalml/issues/2015 to track the IterativeAlgorithm bug. I think that issue is blocking this PR but we can talk about it in stand-up hehe

evalml/automl/automl_search.py
```diff
@@ -266,11 +272,21 @@ def __init__(self,
         self.search_iteration_plot = None
         self._interrupted = False
 
+        self._prepended_components = []
+        self._prepended_components_parameters = {}
+        if data_check_actions is not None:
+            self.data_check_actions = data_check_actions
+            _prepended_components_and_parameters = _make_component_list_from_actions(self.data_check_actions)
+            for component, parameters in _prepended_components_and_parameters:
+                self._prepended_components_parameters.update({component.name: parameters})
```
angela97lin, line 281: This will be problematic if multiple of the same components, but for now, combining components since AutoML isn't great at handling multiple of the same components either 😬  (https://github.com/alteryx/evalml/issues/2049)

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -152,4 +155,5 @@ def _transform_parameters(self, pipeline_class, proposed_parameters):
                     if param_name in init_params:
                         component_parameters[param_name] = value
             parameters[component_class.name] = component_parameters
+        parameters.update(self._prepended_params)
```
dsherry, line 158: 👍

evalml/data_checks/highly_null_data_check.py
```diff
@@ -47,7 +47,7 @@ def validate(self, X, y=None):
                                                                  "code": "HIGHLY_NULL",\
                                                                  "details": {"column": "lots_of_null"}}],\
                                                     "actions": [{"code": "DROP_COL",\
-                                                                 "metadata": {"column": "lots_of_null"}}]}
+                                                                 "metadata": {"columns": "lots_of_null"}}]}
```
dsherry, line 50: @angela97lin this is a breaking change, yes? Could you please mention it in the release notes?

evalml/pipelines/utils.py
```diff
@@ -129,6 +130,10 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
         raise ValueError(f"{estimator.name} is not a valid estimator for problem type")
     preprocessing_components = _get_preprocessing_components(X, y, problem_type, estimator)
     complete_component_graph = preprocessing_components + [estimator]
+    if components_to_prepend is not None:
+        if len(set(components_to_prepend)) != len(components_to_prepend):
+            raise ValueError("`make_pipelines` does not support duplicate components")
```
dsherry, line 135: Nice

evalml/pipelines/utils.py
```diff
@@ -265,7 +270,9 @@ def _make_component_list_from_actions(actions):
         List of components used to address the input actions
     """
     components = []
-    for action in actions:
-        if action.action_code == DataCheckActionCode.DROP_COL:
-            components.append(DropColumns(columns=action.metadata["columns"]))
+    cols_to_drop = []
+    for drop_action in [action for action in actions if action.action_code == DataCheckActionCode.DROP_COL]:
+        cols_to_drop.extend(drop_action.metadata["columns"])
```
dsherry, line 275: Looks good. What happens if `drop_action.metadata` has no key "columns"? Do we have test coverage for that case? I suggest you use `drop_action.metadata.get("columns", [])`

evalml/tests/pipeline_tests/test_pipeline_utils.py
dsherry: This thread totally rocks. Great hunting and thanks for getting that issue fixed @angela97lin !

evalml/pipelines/utils.py
```diff
@@ -265,7 +270,9 @@ def _make_component_list_from_actions(actions):
         List of components used to address the input actions
```
dsherry, line 270: I think this needs to be updated from `list(ComponentBase)` to `list(tuple(ComponentBase, dict))`, correct?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -120,21 +121,26 @@ def test_make_pipeline(input_type, problem_type):
...
-            delayed_features = []
-            if is_time_series(problem_type):
-                delayed_features = [DelayedFeatureTransformer]
-            if estimator_class.model_family == ModelFamily.LINEAR_MODEL:
-                estimator_components = [OneHotEncoder, StandardScaler, estimator_class]
-            elif estimator_class.model_family == ModelFamily.CATBOOST:
-                estimator_components = [estimator_class]
+            if prepended_components == [DropColumns, DropColumns]:
+                with pytest.raises(ValueError, match="does not support duplicate components"):
+                    make_pipeline(X, y, estimator_class, problem_type, components_to_prepend=prepended_components)
```
dsherry, line 129: 👍

dsherry (approved):
@angela97lin great work! I left a couple comments but nothing major! 🚢  😁

chukarsten (changes requested):
A few very small things: one suggested type checking the components_to_prepend and two, the comment about the duplicate column names.  Looks good, though.  Keep it up!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -23,7 +23,8 @@ def __init__(self,
                  n_jobs=-1,  # TODO remove
                  number_features=None,  # TODO remove
                  ensembling=False,
-                 pipeline_params=None):
+                 pipeline_params=None,
+                 _prepended_params=None):
```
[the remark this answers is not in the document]
dsherry: To date, that's been our pattern for "beta" parameters which we're not ready to make public. It makes it easier for us to make breaking changes / renames to new parameters without potentially disrupting users. We can certainly chat more about this pattern.

Should we treat this as a "beta" parameter? I am fine with that. It is easy to remove the underscore when we want to promote public use.

PR #1986, opened 2021-03-16
Support for Multiclass Classification in Class Imbalance data checks

### LINKED ISSUE #1864: Update Class Imbalance Data Check
With the introduction of a custom [downsampler](https://github.com/alteryx/evalml/pull/1857) data splitter, we want to update the class imbalance data check to perform better on samples that our data splitter calls 'severely imbalanced'. Additionally, we will need to update the threshold to be smarter for multiclass classification problems, since our [current threshold of 0.10](https://github.com/alteryx/evalml/blob/b1b3961628717d45ae9d60c48d2e0fceec7671f4/evalml/data_checks/class_imbalance_data_check.py#L14) won't be robust to multiclassification cases with 10 or more target classes.

bchen1116:
fix #1864 

Summary:
* Compute the number of samples in each class of the training data
* Identify the majority class which has the largest number of samples
* For each class, compute the ratio of majority class samples to that classes sample count
* If that ratio exceeds a threshold (default 9:1), raise a class imbalance warning
* If that ratio exceeds a threshold AND the number of samples for that class < 100, raise a severe class imbalance warning

Did not incorporate steps for number of classes in order to keep this data check a little simpler and keep the logic in line with what users would expect. (ie at larger numbers of classes, we don't automatically lower the threshold, even though it might be imbalanced for the model to learn).

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
         * Added multiple oversampling and undersampling sampling methods as data splitters for imbalanced classification :pr:`1775`
         * Added params to balanced classification data splitters for visibility :pr:`1966`
         * Updated ``make_pipeline`` to not add ``Imputer`` if input data does not have numeric or categorical columns :pr:`1967`
+        * Updated ``class_imbalance_data_check`` to better handle multiclass imbalances :pr:``
     * Fixes
     * Changes
         * Removed ``data_checks`` parameter, ``data_check_results`` and data checks logic from ``AutoMLSearch`` :pr:`1935`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,7 +6,7 @@ Release Notes
         * Added multiple oversampling and undersampling sampling methods as data splitters for imbalanced classification :pr:`1775`
         * Added params to balanced classification data splitters for visibility :pr:`1966`
         * Updated ``make_pipeline`` to not add ``Imputer`` if input data does not have numeric or categorical columns :pr:`1967`
-        * Updated ``class_imbalance_data_check`` to better handle multiclass imbalances :pr:``
+        * Updated ``class_imbalance_data_check`` to better handle multiclass imbalances :pr:`1986`
         * Added recommended actions for the output of data check's ``validate`` method :pr:`1968`
     * Fixes
     * Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,7 +6,7 @@ Release Notes
         * Added multiple oversampling and undersampling sampling methods as data splitters for imbalanced classification :pr:`1775`
         * Added params to balanced classification data splitters for visibility :pr:`1966`
         * Updated ``make_pipeline`` to not add ``Imputer`` if input data does not have numeric or categorical columns :pr:`1967`
-        * Updated ``class_imbalance_data_check`` to better handle multiclass imbalances :pr:`1986`
+        * Updated ``ClassImbalanceDataCheck`` to better handle multiclass imbalances :pr:`1986`
         * Added recommended actions for the output of data check's ``validate`` method :pr:`1968`
     * Fixes
         * Updated binary classification pipelines to use objective decision function during scoring of custom objectives :pr:`1934`
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -77,7 +77,7 @@ class ClassImbalanceDataCheck(DataCheck):
         y = infer_feature_types(y)
         y = _convert_woodwork_types_wrapper(y.to_series())
 
-        fold_counts = y.value_counts(normalize=False)
+        fold_counts = y.value_counts(normalize=False, sort=True)
         if len(fold_counts) == 0:
             return results
         # search for targets that occur less than twice the number of cv folds first
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -17,8 +17,9 @@ class ClassImbalanceDataCheck(DataCheck):
 
         Arguments:
             threshold (float): The minimum threshold allowed for class imbalance before a warning is raised.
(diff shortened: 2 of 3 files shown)


evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
```diff
@@ -236,7 +236,7 @@ def test_class_imbalance_nonnumeric(input_type):
     X = pd.DataFrame()
     y_bools = pd.Series([True, False, False, False, False])
     y_binary = pd.Series(["yes", "no", "yes", "yes", "yes"])
-    y_multiclass = pd.Series(["red", "green", "red", "red", "blue", "green", "red", "blue", "green", "red"])
+    y_multiclass = pd.Series(["red", "green", "red", "red", "blue", "green", "red", "blue", "green", "red", "red", "red"])
```
bchen1116, line 239: increase `red` class to maintain the class imbalance data checks

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -77,6 +78,8 @@ def validate(self, X, y):
         y = _convert_woodwork_types_wrapper(y.to_series())
 
         fold_counts = y.value_counts(normalize=False)
+        if len(fold_counts) == 0:
+            return results
```
dsherry, line 82: @bchen1116 in what case does this situation happen? Should this result in an exception, or a data check error, instead of silently continuing?

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -86,8 +89,7 @@ def validate(self, X, y):
                                                   data_check_name=self.name,
                                                   message_code=DataCheckMessageCode.CLASS_IMBALANCE_BELOW_FOLDS,
                                                   details={"target_values": below_threshold_values}), results)
-
-        counts = fold_counts / fold_counts.sum()
+        counts = fold_counts / (fold_counts + fold_counts.values[0])
```
dsherry, line 92: Oh wow, this was the only code change needed?! Very cool 😎 

This logic assumes that `fold_counts` is sorted from the highest to lowest count. For clarity/redundancy, I suggest you set `sort=True` in the call to `value_counts` on line 80. [`True` is the default for `sort`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.value_counts.html), but for code comprehension I think its helpful to have.

What if the target has `nan`s -- do we have test coverage for this case? I think the impl is all set because `dropna` defaults to `False`, but it would be good coverage to add.

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
```diff
@@ -98,9 +98,9 @@ def test_class_imbalance_data_check_multiclass(input_type):
     X = pd.DataFrame()
     y = pd.Series([0, 2, 1, 1])
     y_imbalanced_default_threshold = pd.Series([0, 2, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
-    y_imbalanced_set_threshold = pd.Series([0, 2, 2, 2, 3, 3, 1, 1, 1, 1])
-    y_imbalanced_cv = pd.Series([0, 0, 1, 2, 2, 1, 1, 1])
-    y_long = pd.Series([0, 0, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4])
+    y_imbalanced_set_threshold = pd.Series([0, 2, 2, 2, 2, 3, 3, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
+    y_imbalanced_cv = pd.Series([0, 1, 2, 2, 1, 1, 1])
+    y_long = pd.Series([0, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4])
```
dsherry, line 103: Why were these changes necessary? I only see one change in this unit test and I'm not sure how its connected, appears to just be a message formatting thing,

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
dsherry: Oh I see, so because we are computing "majority vs current" ratio for each class, instead of a hard 10% threshold as before, we need more "red" in order for "green" and "blue" to trigger the error below. Yes?

dsherry (approved):
Looks great! I left a couple impl comments

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
bchen1116: Yep, your next statement explains why this was changed!

evalml/data_checks/class_imbalance_data_check.py
bchen1116: @dsherry This primarily occurs if `y` is empty or if `y` is all null/nan values. One of our tests passes in an empty `y`, so this was added to handle that before we try to index into it.

docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
         * Added multiple oversampling and undersampling sampling methods as data splitters for imbalanced classification :pr:`1775`
         * Added params to balanced classification data splitters for visibility :pr:`1966`
         * Updated ``make_pipeline`` to not add ``Imputer`` if input data does not have numeric or categorical columns :pr:`1967`
+        * Updated ``class_imbalance_data_check`` to better handle multiclass imbalances :pr:`1986`
```
angela97lin, line 9: Really nit-picking now LOL but can we update this to ``ClassImbalanceDataCheck``

angela97lin (approved):
Neat, I'm a big fan of this improvement! LGTM 🥳

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
```diff
@@ -359,3 +359,50 @@ def test_class_imbalance_severe(min_samples, input_type):
         "errors": [],
         "actions": []
     }
+
+
+def test_class_imbalance_large_multiclass():
```
ParthivNaresh, line 364: Excellent!

ParthivNaresh (approved):
Very cleanly done, I love it

PR #1989, opened 2021-03-16
Adds recommended actions for InvalidTargetDataCheck and update _make_component_list_from_actions to address this action

### LINKED ISSUE #1881: Include recommended actions for each data check's set of actions

angela97lin:
Closes #1881

This suddenly became a much bigger PR, so happy to split this up or explain more if it's too confusing 😅 

- Add actions to `InvalidTargetDataCheck` to impute target with missing values.
- Add `TargetImputer` component that can impute target with missing values
- Update `_make_component_list_from_actions` to handle new code, `IMPUTE_COL`
- Update `_retain_custom_types_and_initalize_woodwork` to handle DataColumns
- Update `InvalidTargetDataCheck` to separate out when target is fully null vs has nulls with two different DataCheckMessageCodes (TARGET_HAS_NULL vs TARGET_IS_EMPTY_OR_FULLY_NULL). Only add an action when TARGET_HAS_NULL
     - For the fully null or empty case, we return, rather than letting the other checks run. I think this makes sense than having other warnings (ex: not having two unique values) also be returned, as it is what the immediate issue is.
- Cleanup: updated `InvalidTargetDataCheck to return `TARGET_BINARY_NOT_TWO_UNIQUE_VALUES` for time series binary problems as well
- Cleanup: updated `InvalidTargetDataCheck` to return `TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS` for time series multiclass as well

ANGE TODO / to check:
- TargetImputer in pipelines, then fit and score. Make sure no errors.


evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -133,11 +135,8 @@ def test_invalid_target_data_input_formats():
                                   details={"target_values": []}).to_dict()],
         "actions": []
     }
-    #  test Woodwork
-    y = pd.Series([None, None, None, 0])
-    X = pd.DataFrame({"col": range(len(y))})
-    messages = invalid_targets_check.validate(X, y)
-    assert messages == {
+
+    expected = {
```
angela97lin, line 139: Just cleaning up duplicate expected values :)

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -57,7 +59,7 @@ def validate(self, X, y):
                                                                    "code": "TARGET_HAS_NULL",\
                                                                    "details": {"num_null_rows": 2, "pct_null_rows": 50}}],\
                                                        "warnings": [],\
-                                                       "actions": []}
+                                                       "actions": [{'code': 'IMPUTE_COL', 'details': {'column': None, 'impute_strategy': 'most_frequent', 'is_target': True}}]}
```
angela97lin, line 62: Wanted a way to specify that we want to impute the target without relying on the name of the column

evalml/utils/woodwork_utils.py
```diff
@@ -69,37 +69,45 @@ def _convert_woodwork_types_wrapper(pd_data):
     return pd_data
 
 
-def _retain_custom_types_and_initalize_woodwork(old_datatable, new_dataframe, ltypes_to_ignore=None):
+def _retain_custom_types_and_initalize_woodwork(old_woodwork_data, new_pandas_data, ltypes_to_ignore=None):
```
angela97lin, line 72: Updated to handle DataColumns/Series

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -0,0 +1,95 @@
...
+
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils import (
+    _convert_woodwork_types_wrapper,
+    _retain_custom_types_and_initalize_woodwork,
+    infer_feature_types
+)
+
+
+class TargetImputer(Transformer):
```
freddyaboulton, line 13: @angela97lin Did you try subclassing the evalml `SimpleImputer` or using composition? Seems like we're duplicating the implementation, except that this component only looks at y.

freddyaboulton (changes requested):
@angela97lin This is great! The test coverage is fantastic.  I have some questions on the implementation of the target encoder I want to discuss before merge! Otherwise, this is good to go in my opinion.

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -89,6 +91,8 @@ def validate(self, X, y):
                                                     data_check_name=self.name,
                                                     message_code=DataCheckMessageCode.TARGET_HAS_NULL,
                                                     details={"num_null_rows": num_null_rows, "pct_null_rows": pct_null_rows}).to_dict())
+            results["actions"].append(DataCheckAction(DataCheckActionCode.IMPUTE_COL,
+                                                      details={"column": None, "is_target": True, "impute_strategy": "most_frequent"}).to_dict())
```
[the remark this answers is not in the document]
angela97lin: Ahaha yeah I had chosen `most_frequent` as the default to handle both classification and regression but agree that better things can be done for regression--likely the mean. Since `InvalidTargetDataCheck` already takes in the problem_type, it's easy enough to use that to determine which impute_strategy to use.

Thanks for bringing this up, I think it's simple enough that maybe we might as well make this enhancement here :d

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -0,0 +1,95 @@
...
+        y = _convert_woodwork_types_wrapper(y_ww.to_series())
+        y_df = y.to_frame()
+
+        # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
+        if (y_df.dtypes == bool).all():
+            return _retain_custom_types_and_initalize_woodwork(y_ww, y)
+
+        transformed = self._component_obj.transform(y_df)
+        if transformed.shape[1] == 0:
+            return ww.DataColumn(pd.Series([]))
```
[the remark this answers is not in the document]
angela97lin: Hmm, this is a good point. Confusing, since the data check is supposed to prevent an exception like this from blowing up things downstream, so frustrating that adding a component will still blow things up. I'm okay with raising an exception here though, as it'll make the stack trace less ugly and hopefully provide a better error message.

bchen1116 (commented):
Great test coverage! Agree with Freddy's comments on subclassing `SimpleImputer` and raising exceptions when necessary, but looks good to me otherwise!

evalml/pipelines/components/transformers/imputers/target_imputer.py
angela97lin: I did briefly, but there wasn't too much codeshare (since different checks for 1d vs 2d) and I liked having the methods in this class explicitly for the docstrings too so I ended deciding against it, but I'm down to whip it back up--maybe I overlooked some places that could be cleaned up 😅

angela97lin: Heh, looks like `_get_subclasses` breaks because it only expects "leaf nodes" and by subclassing SimpleImputer, TargetImputer becomes a leaf node and SimpleImputer is no longer an acceptable subclass. I think I'm going to punt on subclassing `SimpleImputer` for now and file an issue for this :d

freddyaboulton: @angela97lin Sounds good! Thanks for looking into this!

angela97lin: Filed #2021 (Year 2021, hey!), thank you for suggesting this @freddyaboulton!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -82,18 +90,27 @@ def validate(self, X, y):
                                                     details={"unsupported_type": y.logical_type.type_string}).to_dict())
         y_df = _convert_woodwork_types_wrapper(y.to_series())
         null_rows = y_df.isnull()
-        if null_rows.any():
+        if null_rows.all():
+            results["errors"].append(DataCheckError(message="Target values are either empty or fully null.",
+                                                    data_check_name=self.name,
+                                                    message_code=DataCheckMessageCode.TARGET_IS_EMPTY_OR_FULLY_NULL,
+                                                    details={}).to_dict())
+            return results
```
angela97lin, line 98: Return here because I don't think it's beneficial to add other errors in this case (such as not having enough samples, etc) when this is the prime error!

evalml/tests/component_tests/test_target_imputer.py
```diff
@@ -0,0 +1,185 @@
...
+    np.testing.assert_almost_equal(y, np.array([np.nan, 0, 2]))
+
+
+def test_target_imputer_does_not_reset_index():
+    y = pd.Series(np.arange(10))
+    y[5] = np.nan
+    assert y.index.tolist() == list(range(10))
+
+    y.drop(0, inplace=True)
+    pd.testing.assert_series_equal(pd.Series([1.0, 2, 3, 4, np.nan, 6, 7, 8, 9], dtype=float, index=list(range(1, 10))), y)
```
bchen1116, line 130: Is there a reason you typed `1.0` here instead of just `1`? hyper-nit, but it seems to disrupt the flow

bchen1116 (approved):
Nice! These tests are super thorough, and big fan of the cleanup! I left a few nitpicks and documentation fix comments, but nothing blocking!

evalml/tests/component_tests/test_target_imputer.py
```diff
@@ -0,0 +1,185 @@
...
+    imputer = TargetImputer(impute_strategy='mean')
+    y_expected = pd.Series([1, 2, 0])
+    _, y_t = imputer.fit_transform(None, y)
+    assert_series_equal(y_expected, y_t.to_series(), check_dtype=False)
+
+
+@pytest.mark.parametrize("fill_value, y, y_expected", [(None, pd.Series([np.nan, 0, 5]), pd.Series([0, 0, 5])),
+                                                       (None, pd.Series([np.nan, "a", "b"]), pd.Series(["missing_value", "a", "b"]).astype("category")),
+                                                       (3, pd.Series([np.nan, 0, 5]), pd.Series([3, 0, 5])),
+                                                       (3, pd.Series([np.nan, "a", "b"]), pd.Series([3, "a", "b"]).astype("category"))])
```
chukarsten, line 45: This last parametrized test case is a very interesting one.  Do we want to match types?  Like if the integer 3 is put in, do we want it filling with the integer 3?  Or the string 3?  Do we want to allow cross-type imputation?  Or perhaps raise a value error.

chukarsten (approved):
Indeed, this was a doozy.  I think the only thing that I'd be a little iffy on is what we do when the user specifies a constant imputation of one type but the column is full of data of the other type.  I don't really see anything blocking, but would like to address that!  Great job!

evalml/data_checks/invalid_targets_data_check.py
angela97lin: Right, if the user picks a fully null target, we tell them to go impute them. Once they have, they'll need to run the data checks again and could potentially run into the error of not having enough samples still, but I don't think it's super useful to tell them on their fully null target that they also might be missing enough samples of a specific class :')

evalml/tests/component_tests/test_target_imputer.py
angela97lin: Hehe nope, just copied from `test_simple_imputer_does_not_reset_index`, will update both!

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -82,18 +90,27 @@ def validate(self, X, y):
                                                     details={"unsupported_type": y.logical_type.type_string}).to_dict())
         y_df = _convert_woodwork_types_wrapper(y.to_series())
         null_rows = y_df.isnull()
-        if null_rows.any():
+        if null_rows.all():
+            results["errors"].append(DataCheckError(message="Target values are either empty or fully null.",
```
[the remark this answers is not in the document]
angela97lin: Yeah, they're different in that empty refers to len(y) == 0, and fully null is len(y) != 0 but all nan values 😢

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -0,0 +1,88 @@
...
+        """Transforms input target data by imputing missing values. 'None' and np.nan values are treated as the same.
+        Arguments:
+            X (ww.DataTable, pd.DataFrame): Features. Ignored.
+            y (ww.DataColumn, pd.Series): Target data to impute.
+        Returns:
+            ww.DataColumn: Transformed y
+        """
+        y_ww = infer_feature_types(y)
+        y = _convert_woodwork_types_wrapper(y_ww.to_series())
+        y_df = y.to_frame()
```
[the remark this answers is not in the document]
angela97lin: 🤮

evalml/tests/component_tests/test_components.py
```diff
@@ -534,7 +535,10 @@ def test_transformer_transform_output_type(X_y_binary):
 
             component.fit(X, y=y)
             transform_output = component.transform(X, y=y)
-            assert isinstance(transform_output, ww.DataTable)
+            if isinstance(component, TargetImputer):
+                assert isinstance(transform_output[1], ww.DataColumn)
```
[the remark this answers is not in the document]
angela97lin: Yes, for sure! Good call 😁

evalml/tests/component_tests/test_target_imputer.py
angela97lin: Yeah, that's a good question! This follows the behavior in SimpleImputer / Imputer right now, but I think it's okay because the type of the series is `category`, which allows for mixed-type categories:

image

Happy to file an issue if you think this is worth a greater discussion though!

PR #1991, opened 2021-03-17
Suppress warning messages in documentation

### LINKED ISSUE #1887: Handle deprecation warnings properly in the documentation
Per the previous release and the deprecation warnings applied to explain_prediction() in #1822 , deprecation errors crop up in the user facing docs.  https://feature-labs-inc-evalml--1874.com.readthedocs.build/en/1874/demos/fraud.html. These will naturally disappear with the completion of #1832 , but it will be best to tackle this now as part of our preparation to adopt the process of deprecating functions in this way.

No bueno.  So we need to come up with a solution that handles this for all the user-facing docs.  It's good that a user sees a warning when working in their console or IDE, but we don't want this in the docs.  Curiously, when I try to reproduce locally, I do not get the warnings on my local jupyter notebook.  We should explore whether there's some global warning suppression that we can perform on the documentation.

More than likely, this issue won't be completed before the explain_prediction() is removed.  If a good test needs to be made, checkout v. 0.19.0 and see how the explain_prediction() function was deprecated and replicate that as a starting point.

image

bchen1116:
fix #1887 

I decided to add hidden cells that caught the warning messages in each notebook. However, it might confuse users when the first code cell execution starts at [2] instead of [1]. 

Adding a warning into `explain_predictions()`:
image

Having the warning in `explain_predictions()`, but also catching warnings in notebooks:
image

### DIFF
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -11,7 +11,7 @@ psutil>=5.6.3
 requirements-parser>=0.2.0
 shap>=0.35.0
 texttable>=1.6.2
-woodwork==0.0.10
+woodwork==0.0.11
 featuretools>=0.20.0
 nlp-primitives>=1.1.0
 networkx>=2.5
--- a/docs/source/.ipython/profile_default/ipython_config.py
+++ b/docs/source/.ipython/profile_default/ipython_config.py
@@ -1,0 +1,6 @@
+c = get_config()
+
+c.InteractiveShellApp.exec_lines = [
+    'import warnings',
+    'warnings.filterwarnings('ignore')'
+]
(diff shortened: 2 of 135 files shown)


dsherry:
> However, it might confuse users when the first code cell execution starts at [2] instead of [1].

Ooh good point. I agree. I didn't think of that when we were considering the options. @bchen1116 is there any way around that?

bchen1116:
Moving forward with the current hidden cell method, but filed an issue [here](https://github.com/alteryx/evalml/issues/2011) to track simplifying this

chukarsten (approved):
LGTM

ParthivNaresh (approved):
Looks good man, what was the outcome of the configuration file we were looking at before?

bchen1116:
@ParthivNaresh I couldn't get the config file to work out. I think ReadTheDocs ends up looking at a different file on their side, so I sent them an email to follow up on how we can get our config in/update the path!

PR #1992, opened 2021-03-17
Polynomial Detrending Component

### LINKED ISSUE #1944: Polynomial detrending transformer
As an evalml user, I would like to use polynomial detrending in time series regression pipelines.

To close out this issue:
1. Implementation of `PolynomialDetrender` transformer that implements `fit`, `transform`, and `inverse_transform` methods.
2. General solution for applying `inverse_transform` at the pipeline level while predicting/scoring.

Out of scope:

1. Automatic additon to `AutoMLSearch`, i.e. modifying `_get_preprocessing_component`. Reason for this is that I think we can go about it both ways: Always create pipelines with and without detrenders or only add detrenders when we detect a trend. The latter relates to #1730

freddyaboulton:
### Pull Request Description
Fix #1944

-----

### DIFF
--- a/docs/source/install.ipynb
+++ b/docs/source/install.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Install\n",
     "\n",
-    "EvalML is available for Python 3.7+. It can be installed with pip or conda.\n",
+    "EvalML is available for Python 3.7 and 3.8 with experimental support 3.9. It can be installed with pip or conda.\n",
     "\n",
     "## Pip with all dependencies\n",
     "\n",
@@ -65,7 +65,11 @@
     "\n",
     "```bash\n",
     "brew install graphviz\n",
-    "```"
+    "```\n",
+    "\n",
+    "## Python 3.9 support\n",
+    "\n",
+    "Evalml can still be installed with pip in python 3.9 but note that `sktime`, one of our dependencies, will not be installed because that library does not yet support python 3.9. This means the ``PolynomialDetrending`` component will not be usable in python 3.9. You can try to install `sktime` [from source](https://www.sktime.org/en/latest/installation.html#building-from-source) in python 3.9 to use the ``PolynomialDetrending`` component but be warned that we only test it in python 3.7 and 3.8."
    ]
   }
  ],
--- a/evalml/pipelines/components/component_base_meta.py
+++ b/evalml/pipelines/components/component_base_meta.py
@@ -19,6 +19,8 @@ class ComponentBaseMeta(BaseMeta):
             klass = type(self).__name__
             if not self._is_fitted and self.needs_fitting:
                 raise ComponentNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
+            elif method.__name__ == 'inverse_transform':
+                return method(self, X, y)
             elif X is None and y is None:
                 return method(self)
             elif y is None:
--- a/evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
+++ b/evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
@@ -1,0 +1,107 @@
+import pandas as pd
+import woodwork as ww
+from skopt.space import Integer
+
+from evalml.pipelines.components.transformers.transformer import Transformer
+from evalml.utils import (
+    _convert_woodwork_types_wrapper,
+    import_or_raise,
+    infer_feature_types
+)
+
+
+class PolynomialDetrender(Transformer):
+    """Removes trends from time series by fitting a polynomial to the data."""
+    name = "Polynomial Detrender"
+
+    hyperparameter_ranges = {"degree": Integer(1, 3)}
+
+    def __init__(self, degree=1, random_seed=0, **kwargs):
+        """Initialize the PolynomialDetrender.
+
+        Arguments:
+            degree (int): Degree for the polynomial. If 1, linear model is fit to the data.
+                If 2, quadratic model is fit, etc.
+            random_seed (int): Seed for the random number generator. Defaults to 0.
+        """
+
+        params = {"degree": degree}
+        params.update(kwargs)
+        error_msg = "sktime is not installed. Please install using 'pip install sktime'"
+
+        trend = import_or_raise("sktime.forecasting.trend", error_msg=error_msg)
+        detrend = import_or_raise("sktime.transformations.series.detrend", error_msg=error_msg)
+
+        detrender = detrend.Detrender(trend.PolynomialTrendForecaster(degree=degree))
+
+        super().__init__(parameters=params,
+                         component_obj=detrender,
+                         random_seed=random_seed)
+
+    def fit(self, X, y=None):
+        """Fits the PolynomialDetrender.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, optional): Ignored.
+            y (ww.DataColumn, pd.Series): Target variable to detrend. Ignored.
+
+        Returns:
+            self
+        """
+        if y is None:
+            raise ValueError("y cannot be None for PolynomialDetrender!")
+        y_dt = infer_feature_types(y)
+        y_series = _convert_woodwork_types_wrapper(y_dt.to_series())
+        self._component_obj.fit(y_series)
+        return self
+
+    def transform(self, X, y=None):
+        """Removes fitted trend from target variable.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, optional): Ignored.
+            y (ww.DataColumn, pd.Series): Target variable to detrend.
+
+        Returns:
+            tuple of ww.DataTable, ww.DataColumn: The input features are returned without modification. The target
+                variable y is detrended
+        """
+        if y is None:
+            return X, y
+        y_dt = infer_feature_types(y)
+        y_series = _convert_woodwork_types_wrapper(y_dt.to_series())
+        y_t = self._component_obj.transform(y_series)
+        y_t = ww.DataColumn(pd.Series(y_t, index=y_series.index))
+        return X, y_t
+
+    def fit_transform(self, X, y=None):
+        """Removes fitted trend from target variable.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, optional): Ignored.
+            y (ww.DataColumn, pd.Series): Target variable to detrend.
+
+        Returns:
+            tuple of ww.DataTable, ww.DataColumn: The first element are the input features returned without modification.
+                The second element is the target variable y with the fitted trend removed.
+        """
+        return self.fit(X, y).transform(X, y)
+
+    def inverse_transform(self, X, y):
+        """Adds back fitted trend to target variable.
+
+        Arguments:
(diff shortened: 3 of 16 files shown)


evalml/pipelines/components/component_base_meta.py
```diff
@@ -19,6 +19,8 @@ def _check_for_fit(self, X=None, y=None):
             klass = type(self).__name__
             if not self._is_fitted and self.needs_fitting:
                 raise ComponentNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
+            elif method.__name__ == 'inverse_transform':
```
freddyaboulton, line 22: This is a short-term fix to make the component metaclass aware of the `inverse_transform` method. The current implementation will not work because if `y` is `None`, the method will run with only the `X` argument which is not what we want.

I think we can change this implementation to something like:
```python
def _check_for_fit(self, *args, **kwargs):
    klass = type(self).__name__
    if not self._is_fitted and self.needs_fitting:
        raise ComponentNotYetFittedError(f'This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}.')
    return method(self, *args, **kwargs)
```

I think this would be more maintainable in the long-term.

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -0,0 +1,107 @@
...
+            tuple of ww.DataTable, ww.DataColumn: The input features are returned without modification. The target
+                variable y is detrended
+        """
+        if y is None:
+            return X, y
+        y_dt = infer_feature_types(y)
+        y_series = _convert_woodwork_types_wrapper(y_dt.to_series())
+        y_t = self._component_obj.transform(y_series)
+        y_t = ww.DataColumn(pd.Series(y_t, index=y_series.index))
+        return X, y_t
```
freddyaboulton, line 75: I'm returning `X` and `y` because our `ComponentGraph` assumes targets are returned as the second value in a tuple

freddyaboulton:
FYI it's impossible to get 100% coverage on the patch:

```python
        # We don't expect to install sktime in python 3.9. Let's verify it's not present:
        if 'sktime' in module and is_running_py_39_or_above:
            with pytest.raises(ModuleNotFoundError):
                import_module(module)
            continue
```

Those lines only run when python 3.9 is installed and we don't measure coverage for python 3.9

evalml/tests/component_tests/test_polynomial_detrender.py
```diff
@@ -0,0 +1,94 @@
...
+        PolynomialDetrender(degree=3).fit_transform(X, None)
+
+    pdt = PolynomialDetrender(degree=3).fit(X, y)
+    with pytest.raises(ValueError, match="y cannot be None for PolynomialDetrender!"):
+        pdt.inverse_transform(X, None)
+
+
+@pytest.mark.parametrize("input_type", ['np', 'pd', 'ww'])
+@pytest.mark.parametrize("use_int_index", [True, False])
+@pytest.mark.parametrize("degree", [1, 2, 3])
```
ParthivNaresh, line 31: Very nice!

ParthivNaresh (approved):
Everything looks great, nicely done!

bchen1116 (approved):
Looks great to me! The test coverage is solid.

docs/source/install.ipynb
```diff
@@ -65,7 +65,11 @@
     "\n",
     "```bash\n",
     "brew install graphviz\n",
-    "```"
+    "```\n",
+    "\n",
+    "## Python 3.9 support\n",
+    "\n",
+    "Evalml can still be installed with pip in python 3.9 but note that `sktime`, one of our dependencies, will not be installed because that library does not yet support python 3.9. This means the ``PolynomialDetrending`` component will not be usable in python 3.9. You can try to install `sktime` [from source](https://www.sktime.org/en/latest/installation.html#building-from-source) in python 3.9 to use the ``PolynomialDetrending`` component but be warned that we only test it in python 3.7 and 3.8."
```
angela97lin, line 72: 👍👍

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -0,0 +1,107 @@
...
+    """Removes trends from time series by fitting a polynomial to the data."""
+    name = "Polynomial Detrender"
+
+    hyperparameter_ranges = {"degree": Integer(1, 3)}
+
+    def __init__(self, degree=1, random_seed=0, **kwargs):
+        """Initialize the PolynomialDetrender.
+
+        Arguments:
+            degree (int): Degree for the polynomial. If 1, linear model is fit to the data.
```
angela97lin, line 23: Nitpick: mention default of 1? :)

evalml/tests/component_tests/test_polynomial_detrender.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_polynomial_detrender_init():
+    delayed_features = PolynomialDetrender(degree=3)
+    assert delayed_features.parameters == {"degree": 3}
+
+
+def test_polynomial_detrender_raises_value_error_target_is_none(ts_data):
+    X, y = ts_data
+
+    with pytest.raises(ValueError, match="y cannot be None for PolynomialDetrender!"):
```
angela97lin, line 21: Realllly nit-picking now but do we want to add tests for fit/transform too for completeness :')

evalml/tests/component_tests/test_polynomial_detrender.py
```diff
@@ -0,0 +1,94 @@
...
+    expected_answer = pd.Series(detrended_values, index=expected_index)
+
+    X, y = X_input, y_input
+
+    if input_type == 'np':
+        X = X_input.values
+        y = y_input.values
+    elif input_type == 'ww':
+        X = ww.DataTable(X_input)
+        y = ww.DataColumn(y_input)
```
angela97lin, line 55: We can use the `make_data_type` fixture for this I believe? hehe

angela97lin (approved):
LGTM, pretty darn cool stuff--the tests were great in understanding what's going on. Left some nitpicky comments per usual but 👍👍👍

evalml/tests/component_tests/test_polynomial_detrender.py
freddyaboulton: Thanks for the tip! I didn't know about `make_data_type`! I'm not sure if it's worth it in this case because there is no pandas to numpy conversion and the woodwork conversion is straight-forward enough in this case. But I will keep it in mind in the future.

evalml/pipelines/components/component_base_meta.py
freddyaboulton: @ParthivNaresh I think we're currently raising an exception in `inverse_transform` if `y` is `None`.  The problem is that the meta class doesn't pass the `y` argument if it's None to `inverse_transform` in the current implementation. This causes an error because we're expecting a `y` argument, even if it's None.

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -0,0 +1,107 @@
...
+                If 2, quadratic model is fit, etc.
+            random_seed (int): Seed for the random number generator. Defaults to 0.
+        """
+
```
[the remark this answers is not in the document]
freddyaboulton: Great idea!

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -0,0 +1,107 @@
...
+        super().__init__(parameters=params,
+                         component_obj=detrender,
+                         random_seed=random_seed)
+
+    def fit(self, X, y=None):
+        """Fits the PolynomialDetrender.
+
+        Arguments:
+            X (ww.DataTable, pd.DataFrame, optional): Ignored.
+            y (ww.DataColumn, pd.Series): Target variable to detrend. Ignored.
```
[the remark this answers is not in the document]
freddyaboulton: Typo 😂  Good catch!

evalml/tests/component_tests/test_components.py
```diff
@@ -490,7 +491,8 @@ def all_init(self, *args, **kwargs):
             component = component_class(test_arg="test")
             component_with_different_kwargs = component_class(diff_test_arg="test")
             assert component.parameters['test_arg'] == "test"
-            assert component._component_obj.test_arg == "test"
+            if not isinstance(component, PolynomialDetrender):
```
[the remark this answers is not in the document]
freddyaboulton: Raises an sklearn exception!

requirements.txt
```diff
@@ -11,3 +11,4 @@ seaborn>=0.11.1
 category_encoders>=2.0.0
 statsmodels >= 0.12.2
 imbalanced-learn>=0.7.0
+sktime>=0.5.3;python_version<"3.9"
```
dsherry, line 14: 👍  yep good call putting this in requirements and note core-requirements, lgtm

evalml/tests/conftest.py
```diff
@@ -99,6 +100,11 @@ def has_minimal_dependencies(pytestconfig):
     return pytestconfig.getoption("--has-minimal-dependencies")
 
 
+@pytest.fixture
+def is_running_py_39_or_above():
+    return sys.version_info >= (3, 9)
```
dsherry, line 105: Nifty!

evalml/tests/utils_tests/test_dependencies.py
```diff
@@ -1,17 +1,24 @@
...
     lines = [line for line in lines if '-r ' not in line]
     reqs = requirements.parse(''.join(lines))
     extra_deps = [req.name if req.name != 'imbalanced-learn' else 'imblearn' for req in reqs]
     extra_deps += ['plotly.graph_objects']
     for module in extra_deps:
+        # We don't expect to install sktime in python 3.9. Let's verify it's not present:
+        if module == 'sktime' and is_running_py_39_or_above:
+            with pytest.raises(ModuleNotFoundError):
+                import_module(module)
+            continue
```
dsherry, line 20: Got it. So the reason this doesn't appear in our coverage report is because we're currently only generating coverage on python 3.8. 👍

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -0,0 +1,112 @@
...
+        params.update(kwargs)
+        error_msg = "sktime is not installed. Please install using 'pip install sktime'"
+
+        trend = import_or_raise("sktime.forecasting.trend", error_msg=error_msg)
+        detrend = import_or_raise("sktime.transformations.series.detrend", error_msg=error_msg)
+
+        detrender = detrend.Detrender(trend.PolynomialTrendForecaster(degree=degree))
+
+        super().__init__(parameters=params,
+                         component_obj=detrender,
```
dsherry, line 43: Is there a reason we need to set `component_obj` for this component? It appears you're doing custom logic for `fit`/`transform`/`fit_transform`, and that was the main utility of passing `component_obj` down to the base class

dsherry (approved):
🚢 !

PR #1994, opened 2021-03-18
Add error message to partial_dependence

### LINKED ISSUE #1923: Partial dependence: better error handling for single-valued features
**Background**
Sklearn’s method by default will only look at values from the 5% to 95% marks, excluding potential outliers. In order to compute the partial dependence, a grid is computed ranging from the feature value at the 5% mark to the feature value at the 95% mark. The exception in question is raised by sklearn’s partial dependence method when all the feature values from 5% to 95% are identical, therefore meaning the grid between those points cannot be computed.

For most normal data, this is fine. The issue arises when the feature(s) in question contain approx >90% of a single unique value. If a feature is almost all one value, it is not capable of being used in the partial dependence computation.

**Repro**
Repro notebook [here](https://alteryx0-my.sharepoint.com/:u:/g/personal/dylan_sherry_alteryx_com/EcSMjKIDMG1HrQkos8XMB9AB022LWjYpVf9mAyHmz2W9kg?e=Xd4NQr), dataset "santander_kaggle".

Check out this plot of the feature named “var3” from that dataset, which has index “0” in my reproducer notebook. The test set in this cases has 38k samples; you can see over 37k of the values in “var3” are equal to 2. Because of that, there’s no way to define a grid between the 5% and 95% marks.

In this case, sklearn's method raises the following exception:
```
ValueError: percentiles are too close to each other, unable to build the grid. Please choose percentiles that are further apart.
```

**Proposed Change**
1. Let's detect this case in our code before running partial dependence and raise a more clear error message. Like "Feature {feature_name} is mostly one value and cannot be used to compute partial dependence." This'll make it easy for downstream users.
2. Let's also expose the `probabilities` quantiles input. We can default it to (0.05, 0.95) like sklearn does.

bchen1116:
fix #1923

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -568,8 +568,9 @@ def partial_dependence(pipeline, X, features, percentiles=(0.05, 0.95), grid_res
     if pipeline.model_family == ModelFamily.BASELINE:
         raise ValueError("Partial dependence plots are not supported for Baseline pipelines")
 
-    if ((isinstance(features, int) and X.iloc[:, features].value_counts(normalize=True).values[0] > 0.95) or (isinstance(features, str) and X[features].value_counts(normalize=True).values[0] > 0.95)):
-        raise ValueError(f"Feature {features} is mostly one value and cannot be used to compute partial dependence.")
+    if ((isinstance(features, int) and X.iloc[:, features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1]) or
+       (isinstance(features, str) and X[features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1])):
+        raise ValueError(f"Feature {features} is mostly one value and cannot be used to compute partial dependence. Try raising the upper percentage value.")
 
     if ((isinstance(features, int) and X.iloc[:, features].isnull().sum()) or (isinstance(features, str) and X[features].isnull().sum())):
         warnings.warn("There are null values in the features, which will cause NaN values in the partial dependence output. Fill in these values to remove the NaN values.", NullsInColumnWarning)
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -563,12 +563,18 @@ def partial_dependence(pipeline, X, features, percentiles=(0.05, 0.95), grid_res
     if pipeline.model_family == ModelFamily.BASELINE:
         raise ValueError("Partial dependence plots are not supported for Baseline pipelines")
 
-    if ((isinstance(features, int) and X.iloc[:, features].isnull().sum()) or (isinstance(features, str) and X[features].isnull().sum())):
+    feature_list = []
+    if isinstance(features, int):
+        feature_list = X.iloc[:, features]
+    elif isinstance(features, str):
+        feature_list = X[features]
+
+    if len(feature_list) and feature_list.isnull().sum():
         warnings.warn("There are null values in the features, which will cause NaN values in the partial dependence output. Fill in these values to remove the NaN values.", NullsInColumnWarning)
 
-    if ((isinstance(features, int) and X.iloc[:, features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1]) or
-       (isinstance(features, str) and X[features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1])):
-        raise ValueError(f"Feature {features} is mostly one value and cannot be used to compute partial dependence. Try raising the upper percentage value.")
+    if len(feature_list) and feature_list.value_counts(normalize=True).values[0] + 0.01 > percentiles[1]:
+        val = feature_list.value_counts(normalize=True).index[0]
+        raise ValueError(f"Feature '{features}' is mostly one value, {val}, and cannot be used to compute partial dependence. Try raising the upper percentage value.")
 
     wrapped = evalml.pipelines.components.utils.scikit_learn_wrapped_estimator(pipeline)
     avg_pred, values = sk_partial_dependence(wrapped, X=X, features=features, percentiles=percentiles, grid_resolution=grid_resolution)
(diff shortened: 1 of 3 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -561,11 +563,15 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
     if pipeline.model_family == ModelFamily.BASELINE:
         raise ValueError("Partial dependence plots are not supported for Baseline pipelines")
 
+    if ((isinstance(features, int) and X.iloc[:, features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1]) or
```
bchen1116, line 566: we add 0.01 (1%) since SKLearn doesn't allow the percentage of each value to be too close to the percentile (for instance, if a feature contains 95% of a value, SKLearn will still throw the error even if we pass the upper percentile as 0.955). Otherwise, 1% was chosen arbitrarily (compared to 0.0075 or another value)

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -233,8 +233,8 @@ def test_partial_dependence_not_fitted(X_y_binary, logistic_regression_binary_pi
 
 
 def test_partial_dependence_warning(logistic_regression_binary_pipeline_class):
-    X = pd.DataFrame({'a': [2, None, 2, 2], 'b': [1, 2, 2, 1]})
-    y = pd.Series([0, 1, 0, 1])
+    X = pd.DataFrame({'a': [1, 2, None, 2, 2], 'b': [1, 1, 2, 2, 1]})
```
bchen1116, line 236: Added extra value so we don't raise the new `ValueError`

evalml/model_understanding/graphs.py
```diff
@@ -564,8 +566,12 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
     if ((isinstance(features, int) and X.iloc[:, features].isnull().sum()) or (isinstance(features, str) and X[features].isnull().sum())):
         warnings.warn("There are null values in the features, which will cause NaN values in the partial dependence output. Fill in these values to remove the NaN values.", NullsInColumnWarning)
 
+    if ((isinstance(features, int) and X.iloc[:, features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1]) or
+       (isinstance(features, str) and X[features].value_counts(normalize=True).values[0] + 0.01 > percentiles[1])):
+        raise ValueError(f"Feature {features} is mostly one value and cannot be used to compute partial dependence. Try raising the upper percentage value.")
```
ParthivNaresh, line 571: Two questions:
- `Features {features}` will return a number if `features` is an int right? Should we return the name of the feature in this case for easier debugging?
- Should we include some information about the value causing the imbalance in the error message like `The value {X.iloc[:, features].value_counts(normalize=True).values[0]} is present in feature {features} above the upper limit for partial dependence..etc`?

ParthivNaresh (approved):
Looks good to me!

evalml/model_understanding/graphs.py
bchen1116: Ah good catch in that first bullet. Since the `features` variable is passed by the user, I assumed it would be fine to output that same value, but I'll add some quotations around it to make it clearer to the user.

I can add in a small clarification on the first value that's causing this

bchen1116: @ParthivNaresh I added a fix, let me know if it looks good to you!

evalml/model_understanding/graphs.py
```diff
@@ -516,6 +516,8 @@ def partial_dependence(pipeline, X, features, grid_resolution=100):
             If features is an int, it must be the index of the feature to use.
             If features is a string, it must be a valid column name in X.
             If features is a tuple of int/strings, it must contain valid column integers/names in X.
+        percentiles (tuple[float]): The lower and upper percentile used to create the extreme values for the grid.
+            Must be in [0, 1]. Defaults to (0.05, 0.95).
         grid_resolution (int): Number of samples of feature(s) for partial dependence plot.  If this value
             is less than the maximum number of categories present in categorical data within X, it will be
             set to the max number of categories + 1.
```
chukarsten, line 523: Could you add the default value for grid_resolution in for me?  I think I failed to do so ages ago.  My bad.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -390,3 +390,23 @@ def test_graph_partial_dependence_multiclass(logistic_regression_multiclass_pipe
...
+    X = pd.DataFrame({"A": [i % 3 for i in range(1000)], "B": [(j + 3) % 5 for j in range(1000)], "random_col": [0 if i < 50 else 1 for i in range(1000)]})
+    y = pd.Series([i % 2 for i in range(1000)])
+    pipeline = logistic_regression_binary_pipeline_class(parameters={"Logistic Regression Classifier": {"n_jobs": 1}})
+    pipeline.fit(X, y)
+    with pytest.raises(ValueError, match="Feature 'random_col' is mostly one value, 1, and cannot be"):
+        partial_dependence(pipeline, X, features="random_col", grid_resolution=20)
+    with pytest.raises(ValueError, match="Feature 'random_col' is mostly one value, 1, and cannot be"):
+        partial_dependence(pipeline, X, features="random_col", percentiles=(0.01, 0.955), grid_resolution=20)
+    with pytest.raises(ValueError, match="Feature '2' is mostly one value, 1, and cannot be"):
+        partial_dependence(pipeline, X, features=2, percentiles=(0.01, 0.955), grid_resolution=20)
```
chukarsten, line 406: Hmmm, that'll be a yike from me.  I guess it's possible for X to not have a good feature name in that column, eh?  Is it worth it trying to return the string feature name?  If not, no biggie.  We're all responsible adults I guess, lol

chukarsten (approved):
Nice work, Bryan!

evalml/tests/model_understanding_tests/test_partial_dependence.py
bchen1116: Yep, I'll return the name then!

PR #1996, opened 2021-03-18
Update Woodwork requirement to 0.0.11

### LINKED ISSUE #1980: Update Woodwork to 0.0.11
We should update EvalML to use Woodwork 0.0.11, which includes updates to the `mutual_information` method. From standup, it sounds like there are some thoughts on how the parameters (`nrows`?) should be used in EvalML. 

@dsherry @freddyaboulton FYI

angela97lin:
Closes #1980


freddyaboulton (approved):
@angela97lin Looks great! Thanks for fixing 😄

PR #1997, opened 2021-03-19
One Hot Encoder: Drop one redundant feature by default for features with two categories

### LINKED ISSUE #1936: One Hot Encoder: Drop one redundant feature by default for features with two categories
Our one hot encoder creates a feature for every level of the original categorical feature:

```python
from evalml.pipelines import OneHotEncoder
import pandas as pd
df = pd.DataFrame({"category": ["a", "b"], "number": [4,5 ]})
OneHotEncoder().fit_transform(df).to_dataframe()
```

image

The `category_a` and `category_b` columns are completely collinear, which makes one redundant. This could have adverse effects on estimator fitting. I think we should drop one by default.

FYI @rpeck

angela97lin:
Closes #1936

Logic for our own custom dropping for binary features (features with only two categories):
1. During fit, if `is_binary`, determine which features have two categories and which is the majority feature which should be dropped. Store this information. Pass `None` to scikit-learn OHE so that scikit-learn does not drop.
2. Transform. Get output array from scikit-learn. During `get_feature_names`, find the name of the column that we should drop. Ex: col "original" with majority category "a" might have a transformed name of "original_a"; store "original_a" so we know to drop this column. Before returning, drop all columns that have been specified.

----

Impl notes:

- separates out `_get_feature_names` from `get_feature_names`, where `_get_feature_names` is pre-dropping binary features and more for our private implementation, while `get_feature_names` is user-facing and has the expected cols (without the binary feature that is dropped)

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -40,7 +40,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
             categories (list): A two dimensional list of categories, where `categories[i]` is a list of the categories
                 for the column at index `i`. This can also be `None`, or `"auto"` if `top_n` is not None. Defaults to None.
             drop (string, list): Method ("first" or "if_binary") to use to drop one category per feature. Can also be
-                a list specifying which method to use for each feature. Defaults to None.
+                a list specifying which method to use for each feature. Defaults to 'if_binary'.
             handle_unknown (string): Whether to ignore or error for unknown categories for a feature encountered
                 during `fit` or `transform`. If either `top_n` or `categories` is used to limit the number of categories
                 per column, this must be "ignore". Defaults to "ignore".
@@ -94,24 +94,21 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
             raise ValueError("Could not find and encode {} in input data.".format(', '.join(invalid_features)))
 
         X_t = self._handle_parameter_handle_missing(X_t)
-        self.binary_values_to_drop = []
+        self._binary_values_to_drop = []
 
         if len(self.features_to_encode) == 0:
             categories = 'auto'
-
         elif self.parameters['categories'] is not None:
             categories = self.parameters['categories']
             if len(categories) != len(self.features_to_encode) or not isinstance(categories[0], list):
                 raise ValueError('Categories argument must contain a list of categories for each categorical feature')
-
         else:
             categories = []
             for col in X_t[self.features_to_encode]:
                 value_counts = X_t[col].value_counts(dropna=False).to_frame()
-                value_to_drop = None
-                if len(value_counts) == 2 and self.parameters['drop'] == "if_binary":  # a column with only two categories. We want to drop the majority class.
-                    value_to_drop = value_counts.index.tolist()[0]  # the minority class
-                    self.binary_values_to_drop.append((col, value_to_drop))
+                if self.parameters['drop'] == "if_binary" and len(value_counts) == 2:
+                    value_to_drop = value_counts.index.tolist()[0]  # the majority class
+                    self._binary_values_to_drop.append((col, value_to_drop))
                 if top_n is None or len(value_counts) <= top_n:
                     unique_values = value_counts.index.tolist()
                 else:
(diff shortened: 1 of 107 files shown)


evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -196,7 +206,8 @@ def _make_name_unique(name, seen_before):
         return name
 
     def get_feature_names(self):
-        """Return feature names for the categorical features after fitting.
+        """TODO might need to make this private
```
angela97lin, line 209: ANGE TODO: make this private because otherwise confusing for users, make public fxn that drops these

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -223,13 +233,36 @@ def get_feature_names(self):
...
                 # then add an int to make it unique
                 proposed_name = self._make_name_unique(f"{col}_{category}", seen_before)
+                if (col, category) in self._binary_values_to_drop:
+                    encoded_features_to_drop.append(proposed_name)
 
                 unique_names.append(proposed_name)
                 unique_encoded_columns.append(proposed_name)
                 seen_before.add(proposed_name)
-            provenance[col] = unique_encoded_columns
+            self._features_to_drop.extend(encoded_features_to_drop)
```
freddyaboulton, line 242: nit-pick: `encoded_features_to_drop` will only ever have at most one element right? I wonder if we can append directly to `_features_to_drop` as opposed to creating a new list and then calling extend of `_features_to_drop`?

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -161,6 +161,20 @@ def test_drop_parameter_is_array():
     assert col_names == expected_col_names
 
 
+def test_drop_binary_and_top_n_2():
```
freddyaboulton, line 164: Good idea to cover this!

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -494,12 +514,19 @@ def test_ohe_column_names_unique():
     df = pd.DataFrame({"A": ["x_y"], "A_x": ["y"]})
     df_transformed = OneHotEncoder().fit_transform(df)
     assert set(df_transformed.columns) == {"A_x_y", "A_x_y_1"}
-    df = pd.DataFrame({"A": ["x_y", "z"], "A_x": ["y_1", "y"], "A_x_y": ["1", "y"]})
+
+    df = pd.DataFrame({"A": ["x_y", "z", "z"], "A_x": ["y", "a", "a", ], "A_x_y": ["1", "y", "y"]})
```
freddyaboulton, line 518: Why did we add a third row here? Was the sorting of the majority class not deterministic?

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -494,12 +514,19 @@ def test_ohe_column_names_unique():
...
+    df_transformed = OneHotEncoder().fit_transform(df)
+    # category y in A_x gets mapped to A_x_y_1 because A_x_y already exists
+    # category y_1 in A_x gets mapped to A_x_y_1_1 because A_x_y_1 already exists
```
freddyaboulton, line 521: typo: "category 1 in A_x_y" ?

freddyaboulton (approved):
@angela97lin Nice! This was a surprisingly tricky one. I think this looks good. I have some minor non-blocking comments and a question on `test_ohe_column_names_unique`

evalml/tests/component_tests/test_one_hot_encoder.py
angela97lin: I believe it is, but it's dependent on the impl of `value_counts`--I wanted to *not* rely on that :d

freddyaboulton: Great! Just wanted to verify.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
bchen1116: I think `encoded_features_to_drop` technically can have multiple values as long as multiple columns have binary categorical values that need to be encoded, but agree still that we can just append to `self._features_to_drop` rather than creating a new list.

bchen1116 (approved):
Nice! LGTM

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
angela97lin: Bahaha I agree with both of you! The reason why I have it this way is because of `unique_encoded_columns_without_dropped` and `provenance`, lol. `provenance` is calculated for each column. We don't want to include the encoded features that we will drop. If I just directly append to `_features_to_drop`, I will still need to keep track of which categories for this column need to be dropped so I don't add to the unique_encoded_columns. If you guys have a suggestion to clean this up then I'd love to hear it, but that's why it's structured like this right now!

Issue #2023, opened 2021-03-23
Calculating for high variance when the mean of cv_scores is zero


angela97lin:
Right now, our calculation for high variance is:
```
        cv_scores_std = cv_scores.std()
        cv_scores_mean = cv_scores.mean()
        if cv_scores_std != 0 and cv_scores_mean != 0:
            high_variance_cv = bool(abs(cv_scores_std / cv_scores_mean) > threshold)```
```

This was put in place to prevent a divide-by-zero warning / error. However, this impl is not ideal: if the mean is close to zero (but not zero), high variance will likely be true (dividing by small number). 

In addition, it is possible for there to be high variance if the mean is zero but the standard deviation is very high. Right now, we default to False but we may want to look for a impl that is smarter in this situation.

chukarsten:
@angela97lin can you clarify what led to this issue coming up?  This doesn't seem like a common occurence.

First approach is to maybe add epsilon to prevent a divide by zero.

angela97lin:
@chukarsten This came up in #2024, where I covered the cases to not emit a runtime warning, but in an impl that might not be ideal. I wonder if there's a better approach than dividing by the mean; the approach of adding an epsilon will avoid the `RuntimeWarning` but also mean that high variance will likely always be triggered, since dividing by a tiny-close-to-zero value will be ~inf.

freddyaboulton:
@angela97lin Is this still a problem? I manually checked your [0.0, -1.0, 1.0] example and I don't get a divide by zero error:

```python
import evalml
from evalml import AutoMLSearch
from evalml.utils import infer_feature_types
import numpy as np


X, y = evalml.demos.load_fraud(n_rows=1000)


automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', objective='f1', max_batches=1)

automl.search()

assert not automl._check_for_high_variance(automl.get_pipeline(0), np.array([0, -1, 1]))
```

I think we only compute the coefficient of variation when both the [std and mean are non-zero](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L726) so we're protected from an exception.

We can debate if returning False for the high variance check is useful when the mean is zero but I think the original concern is the division by 0 which I don't think can happen.

angela97lin:
[...]

PR #2024, opened 2021-03-23
Update `AutoMLSearch._check_for_high_variance` to not emit `RuntimeWarning`

### LINKED ISSUE #1964: F1 objective causes a Runtime warning in AutoMLSearch
Repro:
```python

X, y = load_fraud(n_rows=50000)
automl = AutoMLSearch(X_train=X, y_train=y, objective='f1', problem_type='binary')
automl.search()
```
Results in:
image

I don't observe this issue for `log loss binary` or `auc` metrics, so it might be something that's changed with how F1 is calculated in SKlearn.

angela97lin:
Closes #1964 

The original reproducer resulted in `RuntimeWarning` because we were dividing 0 by 0 (using arrays, I believe using floats themselves produce a slightly different warning):

Ex:
```
a = np.array([0.0])
0.0 / a.mean()
```

We will also run into a similar RuntimeWarning if we have non-0 / 0
```
a = np.array([0.0])
1 / a.mean()
```
image

This PR only calculates high variance (`abs(cv_scores.std() / cv_scores.mean()`) if `cv_scores.std()` is not 0. If `cv_scores.std()` is 0, we don't care what the mean is.

To resolve the latter case (mean is 0), this PR just states that high variance is False. Not ideal, but the thought process was that if mean is 0 and std is not 0 (the std == 0 case is covered above), then high variance will be triggered regardless of how little the variance is (divide by zero). It seems like a better approach to avoid false positives, and opt for false negatives here. I've filed https://github.com/alteryx/evalml/issues/2023 to track improving our impl to be more robust against this case.

For fun, on the original reproducer:
image

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -249,6 +249,7 @@ Transformers are components that take in data as input and output transformed da
     TextFeaturizer
     DelayedFeatureTransformer
     DFSTransformer
+    PolynomialDetrender
 
 Estimators
 ~~~~~~~~~~
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -286,6 +286,7 @@ Regressors are components that output a predicted target value.
     :template: estimator_class.rst
     :nosignatures:
 
+    ARIMARegressor
     CatBoostRegressor
     ElasticNetRegressor
     LinearRegressor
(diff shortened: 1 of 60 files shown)


ParthivNaresh (approved):
Great description of the problem and an accompanying solution! Thanks for filing an issue to track `cv_scores.mean()` being 0.

bchen1116 (approved):
Nice! The writeup is 👌

freddyaboulton (approved):
Looks good @angela97lin !

PR #2030, opened 2021-03-24
Add Undersampler Component

### LINKED ISSUE #2020: Create an undersampler component
Convert our custom under sampler into a component that we can add to the pipeline.

bchen1116:
fix #2020


evalml/pipelines/component_graph.py
```diff
@@ -54,7 +54,10 @@ def from_list(cls, component_list, random_seed=0):
 
             component_dict[component_name] = [component_class]
             if previous_component is not None:
-                component_dict[component_name].append(f"{previous_component}.x")
+                if "sampler" in previous_component:
+                    component_dict[component_name].extend([f"{previous_component}.x", f"{previous_component}.y"])
```
bchen1116, line 58: Need to track both X and y outputs from the undersampler (using 'sampler' to prepare for potential future oversampler transformers)

evalml/tests/component_tests/test_components.py
```diff
@@ -534,6 +536,8 @@ def test_transformer_transform_output_type(X_y_binary):
 
             component.fit(X, y=y)
             transform_output = component.transform(X, y=y)
+            if 'sampler' in component.name:
+                transform_output = transform_output[0]
```
bchen1116, line 540: Undersampler outputs both X and y in transform, unlike other transformers that output just X

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -736,3 +737,44 @@ def test_component_graph_dataset_with_different_types():
...
+             'Elastic Net': [ElasticNetClassifier, 'Undersampler.x', 'Undersampler.y'],
+             'Logistic Regression': [LogisticRegressionClassifier, 'Random Forest', 'Elastic Net']}
+
+    component_graph = ComponentGraph(graph)
+    component_graph.instantiate({})
+    assert component_graph.get_parents('Imputer') == []
+    assert component_graph.get_parents('OneHot') == ['Imputer.x']
+    assert component_graph.get_parents('Undersampler') == ['OneHot.x']
+    assert component_graph.get_parents('Random Forest') == ['Undersampler.x', 'Undersampler.y']
+    assert component_graph.get_parents('Elastic Net') == ['Undersampler.x', 'Undersampler.y']
```
chukarsten, line 756: Interesting, I think I'm a little out of the loop with respect to the component graphs.  Is it expected that there's a branch from the undersampler to elastic net and rf? i assume because theyre both estimators that makes both estimators children of the last component?

That kind of leads me to another question : is this a test specific to the undersampler?  Or would the test look like this without the undersampler?

```
assert component_graph.get_parents('Imputer') == []
assert component_graph.get_parents('OneHot') == ['Imputer.x']
assert component_graph.get_parents('Random Forest') == ['OneHot.x', 'OneHot.y']
assert component_graph.get_parents('Elastic Net') == ['OneHot.x', 'OneHot.y']
```

If it would look like the above with OneHot taking the place of Undersampler, I think we really just need a more generalized test that tests the behavior with respect to the transformer/estimator boundary in the component graph without necessarily using the Undersampler.

chukarsten (changes requested):
Bryan, solid work.  It's nice to see the base class of the Undersampler get put to work.  A few things I'd like to resolve before merge would be:

1. What to do about transform()?  Can we just relegate the transform() task to the base class?
2. The change in the transform() API with the return values.  I think we should do our best to expect all component's transform() to return the same number of values and for each return value to have the same type.  But I'd like to know other people's thoughts, too.
3. The test added in test_component_graph, should it be more generalized?  Or should it specifically reference the undersampler.  

I think a lot of these can be addressed in a post-standup discussion!  Still, overall great work.  Definitely appreciate the clarity of the tests.

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(self, balanced_ratio=4, min_samples=100, min_percentage=0.1, random_seed=0, **kwargs):
+        """Initializes an undersampling transformer to downsample the majority classes in the dataset.
+
+        Arguments:
+            balanced_ratio (float): The largest majority:minority ratio that is accepted as 'balanced'. For instance, a 4:1 ratio would be
+                represented as 4, while a 6:5 ratio is 1.2. Must be greater than or equal to 1 (or 1:1). Defaults to 4.
+            min_samples (int): The minimum number of samples that we must have for any class, pre or post sampling. If a class must be downsampled, it will not be downsampled past this value.
+                To determine severe imbalance, the minority class must occur less often than this and must have a class ratio below min_percentage.
+                Must be greater than 0. Defaults to 100.
```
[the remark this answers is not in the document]
bchen1116: @chukarsten Yep, if used in AutoMLSearch, it is fairly redundant with the `invalidtargetdatacheck` when the data is imbalanced. I think it's nice to leave in if users use these pipelines outside of AutoMLSearch, such as if they pull out pipelines after searching and train it on their local machine. Because of that, I think it's fine to leave this in, as it won't throw errors during AutoMLSearch, but what do you think?

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+            raise ValueError("y cannot be none")
+        y = infer_feature_types(y)
+        X_pd = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y_pd = _convert_woodwork_types_wrapper(y.to_series())
+        index_df = pd.Series(y_pd.index)
+        indices = self._component_obj.fit_resample(X_pd, y_pd)
+        train_indices = index_df[index_df.isin(indices)].dropna().index.values.tolist()
+        return X.iloc[train_indices], y.iloc[train_indices]
+
+    def transform(self, X, y):
```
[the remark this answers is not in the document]
bchen1116: In our base `Transformer` class, we end up trying to [transform with self._component_obj](https://github.com/alteryx/evalml/blob/cdb769950586768ee8b012f1d7fb4254edb6f379/evalml/pipelines/components/transformers/transformer.py#L44), which would raise an error since we don't have a `transform` method with the undersampler. I overrode it to avoid this issue.

evalml/tests/pipeline_tests/test_component_graph.py
bchen1116: @chukarsten, currently, our component parents for all previous transformers only stores the `x` value in the `component_graph` parents' rather than storing both X and y since our previous transformers didn't alter `y`. I added in the support for `y` since our sampling methods (both over and under) will all be transforming both `X` and `y`, which is why the tests I added were specific to the undersampler. For your example, it would look like:

```python
assert component_graph.get_parents('Imputer') == []
assert component_graph.get_parents('OneHot') == ['Imputer.x']
assert component_graph.get_parents('Random Forest') == ['OneHot.x']
assert component_graph.get_parents('Elastic Net') == ['OneHot.x']
```

evalml/tests/component_tests/test_components.py
bchen1116: Agree, I think it would be nice to have the `transformer.transform()` method return the same number of outputs in all cases.

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
chukarsten: Agreed, I think I should have better considered usage outside of automl.  I think from previous reviews that you alert the user to triggering these imbalances so I think this is good to go.

evalml/tests/pipeline_tests/test_component_graph.py
chukarsten: So is this test a test that it's handling the passing forward of the y value?

bchen1116: yep, this is making sure that when we add in an undersampler component to the pipeline, we keep track of both the X and y output of that component, while for other components we still only track the X output. This is definitely something that we can discuss in conjunction with your point about having the output of `transform()` be normalized among all transformers.

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
chukarsten: Hmmm, this is curious.  Does it make more sense to raise a NotImplementedError?

evalml/tests/component_tests/test_undersampler.py
```diff
@@ -56,7 +56,7 @@ def test_undersample_imbalanced(data_type, make_data_type):
 
     assert len(new_X) == 750
     assert len(new_y) == 750
-    assert all(new_y.to_series().value_counts().values == [600, 150])
+    pd.testing.assert_series_equal(new_y.to_series().value_counts(), pd.Series([600, 150], index=[1, 0]), check_dtype=False)
```
chukarsten, line 59: I think I meant here something similar to 
```
class_imbalance = <some calculation>
assert class_imbalance == 4
```

bchen1116: Ah, I get it. I can add that in

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Arguments:
+            balanced_ratio (float): The largest majority:minority ratio that is accepted as 'balanced'. For instance, a 4:1 ratio would be
+                represented as 4, while a 6:5 ratio is 1.2. Must be greater than or equal to 1 (or 1:1). Defaults to 4.
+            min_samples (int): The minimum number of samples that we must have for any class, pre or post sampling. If a class must be downsampled, it will not be downsampled past this value.
+                To determine severe imbalance, the minority class must occur less often than this and must have a class ratio below min_percentage.
+                Must be greater than 0. Defaults to 100.
+            min_percentage (float): The minimum percentage of the minimum class to total dataset that we tolerate, as long as it is above min_samples.
+                If min_percentage and min_samples are not met, treat this as severely imbalanced, and we will not resample the data.
+                Must be between 0 and 0.5, inclusive. Defaults to 0.1.
```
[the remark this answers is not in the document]
bchen1116: I believe this is resolved based on the brief conversation above? Lmk @chukarsten

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
bchen1116: Hmm, I don't think so. When we try to score/predict on a pipeline, we call `transform()` on all of the transformer components rather than `fit_transform`, which makes sense since we don't want to refit the component if we have already trained the pipeline and are predicting. In this scenario, our sampler shouldn't do anything (since we only sample during training), and so we should allow users to both call `transform()` and get a result from it.

evalml/pipelines/component_graph.py
freddyaboulton: Sounds good!

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
chukarsten: Good to go

evalml/pipelines/component_graph.py
```diff
@@ -54,7 +54,10 @@ def from_list(cls, component_list, random_seed=0):
 
             component_dict[component_name] = [component_class]
             if previous_component is not None:
-                component_dict[component_name].append(f"{previous_component}.x")
+                if "sampler" in previous_component:
```
freddyaboulton, line 57: @bchen1116 This assumes the undersampler is placed right before the final estimator. It makes sense for the sampling to happen there, and I know the pipelines that automl will create will follow that convention but I wonder what should happen when a user defines a pipeline where that isn't the case. 

I don't think it's fruitful for us to catch all possible "bad states" the user could find themselves in but I do think it's easier to find yourself in a bad state with the list representation. 

I don't think any action is required here but I wonder what other people think! Maybe it merits a broader discussion: @chukarsten @dsherry @angela97lin @jeremyliweishih @ParthivNaresh

evalml/pipelines/component_graph.py
```diff
@@ -144,7 +147,8 @@ def _fit_transform_features_helper(self, needs_fitting, X, y=None):
                 parent_output = parent_output.to_series()
                 parent_output = pd.DataFrame(parent_output, columns=[parent])
                 parent_output = infer_feature_types(parent_output)
-            final_component_inputs.append(parent_output)
+            if parent_output is not None:
```
freddyaboulton, line 150: I don't think we need to make this change?

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+         Returns:
+            ww.DataTable, ww.DataColumn: Undersampled X and y data
+        """
+        X = infer_feature_types(X)
+        if y is None:
+            raise ValueError("y cannot be none")
+        y = infer_feature_types(y)
+        X_pd = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y_pd = _convert_woodwork_types_wrapper(y.to_series())
+        index_df = pd.Series(y_pd.index)
```
freddyaboulton, line 76: How come we need to create `index_df`? Does `index_df[index_df.isin(y_pd.index)]` not work?

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+        indices = self._component_obj.fit_resample(X_pd, y_pd)
+        train_indices = index_df[index_df.isin(indices)].dropna().index.values.tolist()
```
freddyaboulton, line 78: I vote we don't drop na here. I don't think it's the sampler's responsibility to do so. What are your thoughts?

evalml/tests/component_tests/test_components.py
chukarsten: Thank you for handling this while I was sweating and delusional!  Sounds good.

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
freddyaboulton: I think this is a cool way of specifying "train only" transformers! I was thinking we'd have to pass info from the pipeline down to the component but I think it's great we can do something simpler.

The one confusing thing is that `fit().transform() != fit_transform()` which I think is what someone coming fresh off sklearn would expect. Maybe we should just be super clear about which transformers are "train only" and what users should expect with them. I think "train only" transformers are a cool concept we can use to differentiate ourselves.

chukarsten (approved):
I think all my concerns are resolved through discussion!  Thanks for taking the time!

freddyaboulton (approved):
@bchen1116 This is awesome! I have a couple of questions on the implementation but the tests (and the entire feature) are very solid! 👏

evalml/pipelines/component_graph.py
bchen1116: Ahh, this is a good point. I believe this code technically allows a pipeline to be initialized with a component graph similar to `['Undersampler', 'Imputer', 'Logistic Regression Classifier']`, although since we only track `y` for the undersampler transform output, we would run into issues during the `fit` call. 

This works:
image

This fails at the estimator:
image
image

My goal was to make the sampler precede the final estimator, but it could be smart to ensure the sampler(s) come directly before the final estimator. It's certainly something we can validate through the `ComponentGraph` during instantiation. What do ya'll think?

evalml/pipelines/components/transformers/samplers/undersampler.py
bchen1116: @freddyaboulton I don't understand the question?

evalml/pipelines/components/transformers/samplers/undersampler.py
bchen1116: Ah, this drops the indices that we don't want to sample. So `index_df[index_df.isin(indices)]` produces a series of either True or nan values, where True means that we do want to keep that index. We drop nans in order to remove all indices we are dropping

evalml/pipelines/components/transformers/samplers/undersampler.py
freddyaboulton: Nevermind, misread the code!

evalml/pipelines/components/transformers/samplers/undersampler.py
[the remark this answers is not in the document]
bchen1116: Yeah, that's a good point! Would it be best to state that this is a 'train only' transformer in the documentation/description of the component?

evalml/pipelines/components/transformers/samplers/undersampler.py
freddyaboulton: Is that true? I just deleted the `dropna()` and all the tests passed on pandas `1.2.3` and `1.1.5`.

bchen1116: WAUUU it worked you're right. Unsure why I did it before now.... I'll delete it, thanks!

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -0,0 +1,56 @@
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils.woodwork_utils import (
+    _convert_woodwork_types_wrapper,
+    infer_feature_types
+)
+
+
+class BaseSampler(Transformer):
```
bchen1116, line 8: Added this base class to expand all of our future under/oversamples with.

chukarsten: I think this is a solid add and we'll reap the rewards of it later.

evalml/pipelines/component_graph.py
bchen1116: After stand-up, we decided to proceed with this approach and allow this to error, although we file an [issue](https://github.com/alteryx/evalml/issues/2053) to track adding support for tracking the `y` output of components when we initialize a `componentGraph` through a list.

ParthivNaresh (approved):
Looks excellent!

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Arguments:
+            X (ww.DataFrame): Training features
+            y (ww.DataColumn): Target features
+
+        Returns:
+            self
+        """
+        if y is None:
+            raise ValueError("y cannot be none")
```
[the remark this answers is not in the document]
bchen1116: Mmm, the check was mainly to make sure `fit` doesn't get called without a `y` value, even though `fit` does nothing to the data or the model

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -0,0 +1,93 @@
...
+    infer_feature_types
+)
+
+
+class Undersampler(Transformer):
+    """Random undersampler component."""
+    name = "Undersampler"
+    hyperparameter_ranges = {
+        'balanced_ratio': Real(1, 10),
+        'min_samples': Integer(50, 1000),
```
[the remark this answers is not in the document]
bchen1116: I didn't put `min_percentage` since that's only used in conjunction with determining if the class is severely imbalanced, in which case we don't sample. `min_samples` controls both the severely imbalanced case as well as the imbalanced case by limiting the amount we can downsample a certain class to. I figured `min_samples` would be more useful to tune than `min_percentage` for this reason. 
Becausse of this, I'm not sure how useful it would be to expose `min_percentage` as a hyperparam for `AutoMLSearch` to iterate though. What do you think @ParthivNaresh?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -736,3 +737,44 @@ def test_component_graph_dataset_with_different_types():
     assert input_feature_names['Elastic Net'] == ['column_3', 'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
                                                   'column_2_1', 'column_2_2', 'column_2_3', 'column_2_4', 'column_2_5', 'column_2_6']
     assert input_feature_names['Logistic Regression'] == ['Random Forest', 'Elastic Net']
+
+
+def test_component_graph_sampler():
```
angela97lin, line 742: Was thinking about this RE target imputer, but can you try adding a test for fitting and predicting with a component graph that has an undersampler and confirm that it works?

bchen1116: Added a test in the pipeline tests!

PR #2038, opened 2021-03-25
Fix AutoMLSearch.train_pipelines not applying undersampling

### LINKED ISSUE #1987: AutoMLSearch.train_pipelines does not apply undersampling
If users run `AutoMLSearch` with an imbalanced dataset, `AutoMLSearch` will use undersampling by default. 

Our current behavior is to train all pipelines with the undersampling. This is taken care of by the sampler's `split` method during search and in the `_find_best_pipeline` method when `AutoMLSearch` goes to train the best pipeline.

The undersampling is [missing](https://github.com/alteryx/evalml/blob/main/evalml/automl/engine/sequential_engine.py#L59) from the `train_batch` logic of the engine and should be added.

angela97lin:
Closes #1987


chukarsten (approved):
Looks good, nice and succinct ;)

bchen1116 (approved):
LGTM!

PR #2057, opened 2021-03-30
Integrate KNN to AutoML

### LINKED ISSUE #1696: Integrate KNN Classifier into AutoML
Remove KNeighborsClassifier from gen_utils._not_used_in_automl and then update the tests to account for the new estimator being counted.

ParthivNaresh:
Fixes #1696 

Performance tests are [here](https://alteryx.quip.com/xp00AOZ75ljr/KNN-Classifier-Performance-Testing)


chukarsten (changes requested):
The PR looks great.  I think I'd just like to get some resolution on that credit card dataset taking twice as long.  I kicked off three more runs (job-ids 613-615) overnight that might be able to answer whether the runners are just being unstable.

bchen1116 (approved):
Agree with Karsten, once we can figure out the weird creditcard run in the perf tests, this looks good to me!

ParthivNaresh:
@chukarsten @bchen1116 I downloaded the creditcard.csv dataset and ran 4 different model families over 20 iterations.
Here are the total search times:
`KNN : 5:28`
`Random Forest : 6:01`
`LightGBM : 5:09`
`XGBoost : 4:13`
I don't see KNN as being a massive outlier here, but across 4 different performance testing runs, KNN was consistently the outlier for `time_automl_fit` taking 3 times as long.

chukarsten:
I'm not sure how we want to proceed with this.  Do we think we should run the larger version of the datasets to see if it's just credit card that's behaving pathologically?  I think this is still a little concerning unless there's something else to mitigate this concern. @ParthivNaresh

ParthivNaresh:
@dsherry @chukarsten @bchen1116 
Now that we've verified that the creditcard dataset (consisting of 284K rows and 30 features) is significantly larger than expected, I reran the tests locally. Across 10 iterations, here are the results for `AutoMLSearch` times:
`KNN` : 1:32
`Random Forest` : 0:48
`LightGBM` : 0:22
`XGBoost` : 0:23
This supports what we're seeing with the perf tests.
My assumption is that as a distance based estimator, it's expected for KNN to take longer to fit on much larger data, and with over 8 million values, this dataset would qualify as such. Do we feel comfortable with integrating this?

bchen1116:
@ParthivNaresh, it looks good to me! The performance difference makes sense with the dataset

dsherry:
@ParthivNaresh thanks, I left a couple comments on the perf test results doc, let's continue discussion there before making a decision about whether or not to merge this.

ParthivNaresh:
Closing this as we've decided to table KNN for the time being.

PR #2064, opened 2021-03-31
Add method to convert actions to a preprocessing pipeline

### LINKED ISSUE #2058: Add method to convert actions to a preprocessing pipeline
Following the close of https://github.com/alteryx/evalml/pull/1981, we decided that rather than adding an `actions` parameter to allow AutoMLSearch to address actions, we will have the user address them beforehand. We will help with this process by creating a pipeline of components that can be used to address actions. 

This design choice will result in two separate pipelines which are required to take the original data to the model result: the preprocessing transformations from actions, and the pipeline generated from AutoMLSearch. Users who want to use their original data must piece together the two pipelines. We will provide a utility method to help with this.

This issue tracks creating a method that will generate a pipeline from actions, as well as a utility method to combine multiple pipelines.

angela97lin:
Closes #2058

To enable this, the following were also changed:
- Updates `PipelineBase` to accept empty component graphs (len(ComponentGraph) == 0), previously had ugly stack trace
- Updates `make_pipeline_from_components` to allow creating pipelines where an estimator is not the last component. EDIT: will be blocked on #2103

### DIFF
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -80,9 +80,10 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
 
         self.input_feature_names = {}
         self.input_target_name = None
-
-        final_component = self._component_graph.get_last_component()
-        self.estimator = final_component if isinstance(final_component, Estimator) else None
+        self.estimator = None
+        if len(self._component_graph.compute_order) > 0:
+            final_component = self._component_graph.get_last_component()
+            self.estimator = final_component if isinstance(final_component, Estimator) else None
         self._estimator_name = self._component_graph.compute_order[-1] if self.estimator is not None else None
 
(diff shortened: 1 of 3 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -80,9 +80,10 @@ def __init__(self, parameters, random_seed=0):
 
         self.input_feature_names = {}
         self.input_target_name = None
-
-        final_component = self._component_graph.get_last_component()
-        self.estimator = final_component if isinstance(final_component, Estimator) else None
+        self.estimator = None
+        if len(self._component_graph.compute_order) > 0:
```
angela97lin, line 84: Removing limitation that a pipeline has to have a non-empty list of components

angela97lin:
Just going to close this for now, can revive when we work more on data check actions :)

PR #2077, opened 2021-04-02
Change undersampler to use minority:majority ratio

### LINKED ISSUE #2076: Change undersampler to use minority:majority ratio
In order to smoothly accomodate for our Oversampler, we want to change the Undersampler's `balanced_ratio` arg to represent minority:majority rather than the current majority:minority ratio.

bchen1116:
fix #2076

### DIFF
--- a/evalml/pipelines/components/transformers/samplers/undersampler.py
+++ b/evalml/pipelines/components/transformers/samplers/undersampler.py
@@ -13,16 +13,16 @@ class Undersampler(BaseSampler):
     """Random undersampler component. This component is only run during training and not during predict."""
     name = "Undersampler"
     hyperparameter_ranges = {
-        'balanced_ratio': Real(1, 10),
+        'balanced_ratio': Real(0.000001, 1),
         'min_samples': Integer(50, 1000),
     }
 
-    def __init__(self, balanced_ratio=4, min_samples=100, min_percentage=0.1, random_seed=0, **kwargs):
+    def __init__(self, balanced_ratio=0.25, min_samples=100, min_percentage=0.1, random_seed=0, **kwargs):
         """Initializes an undersampling transformer to downsample the majority classes in the dataset.
 
         Arguments:
-            balanced_ratio (float): The largest majority:minority ratio that is accepted as 'balanced'. For instance, a 4:1 ratio would be
-                represented as 4, while a 6:5 ratio is 1.2. Must be greater than or equal to 1 (or 1:1). Defaults to 4.
+            balanced_ratio (float): The smallest minority:majority ratio that is accepted as 'balanced'. For instance, a 1:4 ratio would be
+                represented as 0.25, while a 1:1 ratio is 1.0. Must be between 0 and 1, inclusive. Defaults to 0.25.
             min_samples (int): The minimum number of samples that we must have for any class, pre or post sampling. If a class must be downsampled, it will not be downsampled past this value.
                 To determine severe imbalance, the minority class must occur less often than this and must have a class ratio below min_percentage.
                 Must be greater than 0. Defaults to 100.
--- a/evalml/pipelines/components/transformers/samplers/undersampler.py
+++ b/evalml/pipelines/components/transformers/samplers/undersampler.py
@@ -13,7 +13,7 @@ class Undersampler(BaseSampler):
     """Random undersampler component. This component is only run during training and not during predict."""
     name = "Undersampler"
     hyperparameter_ranges = {
-        'balanced_ratio': Real(0.000001, 1),
+        'balanced_ratio': Real(1e-10, 1),
         'min_samples': Integer(50, 1000),
     }
 
--- a/evalml/preprocessing/data_splitters/balanced_classification_sampler.py
+++ b/evalml/preprocessing/data_splitters/balanced_classification_sampler.py
@@ -50,9 +50,10 @@ class BalancedClassificationSampler(SamplerBase):
         """
         counts = y.value_counts()
         normalized_counts = y.value_counts(normalize=True)
-        class_ratios = min(normalized_counts) / normalized_counts
+        minority_class_count = min(normalized_counts)
+        class_ratios = minority_class_count / normalized_counts
         # if no class ratios are larger than what we consider balanced, then the target is balanced
-        if all(class_ratios < self.balanced_ratio):
+        if all(class_ratios >= self.balanced_ratio):
             return {}
         # if any classes have less than min_samples counts and are less than min_percentage of the total data,
         # then it's severely imbalanced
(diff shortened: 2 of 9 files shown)


chukarsten (approved):
Looks good!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -922,6 +922,7 @@ def test_describe_pipeline(mock_fit, mock_score, return_dict, caplog, X_y_binary
     caplog.clear()
     automl_dict = automl.describe_pipeline(0, return_dict=return_dict)
     out = caplog.text
+    print(out)
```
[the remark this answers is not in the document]
bchen1116: Good call, I'll remove!

evalml/preprocessing/data_splitters/balanced_classification_sampler.py
```diff
@@ -50,19 +50,19 @@ def _find_ideal_samples(self, y):
...
         # then it's severely imbalanced
         if any(counts < self.min_samples) and any(normalized_counts < self.min_percentage):
             return {}
         # otherwise, we are imbalanced enough to perform on this
-        undersample_classes = counts[class_ratios > self.balanced_ratio].index.values
+        undersample_classes = counts[class_ratios <= self.balanced_ratio].index.values
         # find goal size, round it down if it's a float
         minority_class = min(counts.values)
-        goal_value = max(int((self.balanced_ratio * minority_class) // 1), self.min_samples)
+        goal_value = max(int((minority_class / self.balanced_ratio) // 1), self.min_samples)
```
dsherry, line 65: Makes sense! Since this PR has made it so that `balanced_ratio_new = 1 / balanced_ratio_old`, we just flip `self.balanced_ratio` over the dividing line. Nice.

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
```diff
@@ -82,15 +85,15 @@ def test_classification_imbalanced_balanced_ratio(num_classes, balanced_ratio):
     indices = bcs.fit_resample(X, y)
     X2 = X.loc[indices]
     y2 = y.loc[indices]
-    if balanced_ratio >= 3:
+    if balanced_ratio <= .3:
```
dsherry, line 88: This is because the default `balanced_ratio` is 0.25? Could you say `if balanced_ratio <= 0.25` for clarity?

(Its probably more trouble than its worth but you could also use introspection to retrieve the default `balanced_ratio` value from the method signature 😆 )

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
```diff
@@ -82,15 +85,15 @@ def test_classification_imbalanced_balanced_ratio(num_classes, balanced_ratio):
...
         # the classes are considered balanced, do nothing
         pd.testing.assert_frame_equal(X, X2)
         pd.testing.assert_series_equal(y, y2)
     else:
         # remove some samples
-        assert len(X2) == {2: (250 * (balanced_ratio + 1)), 3: (200 * (balanced_ratio + 2))}[num_classes]
+        assert len(X2) == {2: (250 + int(250 / balanced_ratio)), 3: (400 + int(200 / balanced_ratio))}[num_classes]
```
dsherry, line 94: What happened to the "+ 1" and "+ 2" here?

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
```diff
@@ -167,13 +170,13 @@ def test_classification_imbalanced_normal_imbalance_binary(min_samples, balanced
     indices = bcs.fit_resample(X, y)
     X2 = X.loc[indices]
     y2 = y.loc[indices]
-    if balanced_ratio >= 6:
+    if balanced_ratio < .2:
```
dsherry, line 173: Same as before, please change to 0.25 for clarity, and I see one more similar usage below.

dsherry (changes requested):
Looking close! I think I found one minor bug, and I had a couple test suggestions.

evalml/preprocessing/data_splitters/balanced_classification_sampler.py
```diff
@@ -50,19 +50,19 @@ def _find_ideal_samples(self, y):
         """
         counts = y.value_counts()
         normalized_counts = y.value_counts(normalize=True)
-        class_ratios = normalized_counts / min(normalized_counts)
-        # if no class ratios are greater than what we consider balanced, then the target is balanced
-        if all(class_ratios <= self.balanced_ratio):
+        class_ratios = min(normalized_counts) / normalized_counts
+        # if no class ratios are larger than what we consider balanced, then the target is balanced
+        if all(class_ratios < self.balanced_ratio):
```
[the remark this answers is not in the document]
bchen1116: Good catch!

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
bchen1116: @dsherry We choose 0.3 since the targets have a class ratio of 1/3 for the minority/majority balance. In scenarios where we consider anything less than 1/3 as balanced (for the sampler), we expect the data to not be undersampled. For instance, if we consider class ratios down to 1/5 as balanced, then this dataset shouldn't be sampled.

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
bchen1116: Similar reasoning as before!

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
bchen1116: @dsherry The `+1` and `+2` were used to handle the addition of the minority class(es), while the (250 * balanced_ratio) handled the number of majority class samples that we wanted. I opted to simply add `(minority class + (calculation for majority class))` instead, which allowed me to pull out the `+1` and `+2` from the equation. I felt this would be easier to follow.

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -13,16 +13,16 @@ class Undersampler(BaseSampler):
     """Random undersampler component. This component is only run during training and not during predict."""
     name = "Undersampler"
     hyperparameter_ranges = {
-        'balanced_ratio': Real(1, 10),
+        'balanced_ratio': Real(1e-10, 1),
```
bchen1116, line 16: Should `balanced_ratio` be a tunable hyperparameter? If a user passes in `balanced_ratio=0.2` as the argument, shouldn't we keep that ratio for all pipelines? 

Additionally, this equivalent parameter in the oversampler components won't be able to be tuned (due to the additional complexities it requires for multiclass samples). Intuitively, I want to keep the performances of these two component classes similar as well.

bchen1116: Issue tracking removal of hyperparameters [here](https://github.com/alteryx/evalml/issues/2104)

evalml/tests/preprocessing_tests/test_balanced_classification_sampler.py
dsherry: Understood, so the test setup changed. 👍

dsherry (approved):
🚢 🙏 😁

PR #2078, opened 2021-04-02
Clean Notebook

### LINKED ISSUE #2067: Mode Baseline Binary Classification Pipeline
Hi,

I was following through this tutorial: https://evalml.alteryx.com/en/stable/start.html

I noticed the following rankings of `automl.search`

image

Notice that the score of `Mode Baseline Binary Classification Pipeline` is 0.0000. The dataset used in this tutorial has a target distribution of ~ 85/15. So even if the base classifier is predicting the majority class as a prediction for all observations, the score will still be >0. 

Accordingly, `percent_better_than_baseline` values will also change for all models, given that this turns out to be a bug.

bchen1116:
fix #2067 

Small change to fix spacing for clarity:
image

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -250,6 +250,7 @@ Transformers are components that take in data as input and output transformed da
     DelayedFeatureTransformer
     DFSTransformer
     PolynomialDetrender
+    Undersampler
 
 Estimators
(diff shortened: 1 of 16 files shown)


freddyaboulton (approved):
Looks good @bchen1116 !!

PR #2079, opened 2021-04-02
Add Oversampler components to EvalML

### LINKED ISSUE #2014: Oversampling Support: Add Oversampler component to EvalML
Part 1 of EvalML's oversampling support. According to our early trial [perf tests](https://alteryx.quip.com/EMqnAle4NI9B/Oversampler-Component-initial-results), we found that creating an oversampler component in the pipeline is both possible and can be beneficial for imbalanced classification datasets. Design doc [here](https://alteryx.quip.com/l7WcAeaxkB82/Oversampling-after-Feature-Engineering-REVISED)

This issue tracks creating and adding an oversampler component that users can add to our pipelines.

bchen1116:
fix #2014 

Design doc [here](https://alteryx.quip.com/TW2nANznZcpN/Oversampling-in-AutoML-Search)

Example code to show the oversampler works in pipelines:
image


evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,133 @@
...
+)
+from evalml.utils import import_or_raise
+from evalml.utils.woodwork_utils import infer_feature_types
+
+
+class SMOTESampler(BaseSampler):
+    """SMOTE Oversampler component. Works on numerical datasets only. This component is only run during training and not during predict."""
+    name = "SMOTE Oversampler"
+    hyperparameter_ranges = {
+        'k_neighbors': Integer(2, 10)
```
dsherry, line 14: @bchen1116 oops we didn't chat about hyperparameter ranges

For v1, let's not expose any automl hyperparameter ranges for our undersampler and oversampler just yet. I think we need to punt thinking about this over to #1717 .

bchen1116: Sounds good!

evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    def __init__(self, sampling_strategy='auto', k_neighbors=5, n_jobs=-1, random_seed=0, **kwargs):
```
dsherry, line 17: For other readers: @bchen1116 and I caught up and we're thinking of using the following API:
* `sampling_ratio`: a float in the range `(0, 1]` indicating the minority:majority ratio which the sampler should try to meet for every class (except the majority class). Default 0.25, aka a 1:4 ratio. This is the "simple" option to use oversampling.
* `sampling_ratio_dict`: a dictionary mapping each class value (which could be string, float, other) to a sampling ratio for that class. Default `None`. This is the "advanced" option to use oversampling.

@bchen1116 a "nice to have" is if the parameters we use for our undersampler match those used by the oversampler. Could we rename the undersampler ratio param to match, and also add a `sampling_ratio_dict` param to the undersampler? If you agree, I suggest you a) get a small separate PR up to rename the undersampler param and b) file an issue to track `sampling_ratio_dict` to undersampler.

evalml/pipelines/components/transformers/samplers/oversamplers.py
dsherry: I just filed an issue to track removing hyperparam ranges from the undersampler component: #2104

dsherry (changes requested):
@bchen1116 and I just chatted about this PR and I left some comments summarizing next steps.

evalml/pipelines/components/utils.py
```diff
@@ -6,7 +6,9 @@
 
 from evalml.exceptions import MissingComponentError
 from evalml.model_family.utils import handle_model_family
-from evalml.pipelines.components import ComponentBase, Estimator, Transformer
+from evalml.pipelines.components.component_base import ComponentBase
+from evalml.pipelines.components.estimators.estimator import Estimator
+from evalml.pipelines.components.transformers.transformer import Transformer
```
bchen1116, line 11: Split these to avoid circular imports when we import `make_balanced_dictionary` to our oversampler file

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -0,0 +1,241 @@
...
+
+@pytest.mark.parametrize("sampler", [SMOTESampler, SMOTENCSampler, SMOTENSampler])
+def test_sampler_dic_overrides_ratio(sampler):
+    X = np.array([[i for i in range(1000)],
+                  [i % 7 for i in range(1000)],
+                  [0.3 * (i % 3) for i in range(1000)]]).T
+    y = np.array([0] * 800 + [1] * 100 + [2] * 100)
+
+    goal_value_dic = {0: 800, 1: 800, 2: 600}
+    oversampler = sampler(sampling_ratio=1, sampling_ratio_dict=goal_value_dic)
```
chukarsten, line 200: Do we want some kind of feedback here to let the user of the sampler know that this behavior is occuring?

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -0,0 +1,241 @@
...
+        oversampler.fit_transform(X, None)
+    oversampler.fit(X, pd.Series([0] * 4 + [1]))
+    oversampler.transform(X, None)
+
+
+@pytest.mark.parametrize("sampler", [SMOTESampler(sampling_ratio=1),
+                                     SMOTENCSampler(sampling_ratio=1, categorical_features=[0]),
+                                     SMOTENSampler(sampling_ratio=1)])
+@pytest.mark.parametrize("data_type", ["np", "pd", "ww"])
+def test_no_oversample(data_type, sampler, make_data_type, X_y_binary):
```
chukarsten, line 66: What is this one testing?

chukarsten (changes requested):
hey Bryan, a few things you might want to consider - is it worth elevating the __init__() and fit_transform() to the BaseSampler?  There's a few copy-pastas.  A duplicate test too!  Overall, I like how it's coming along.

evalml/tests/component_tests/test_utils.py
```diff
@@ -76,3 +81,32 @@ def test_scikit_learn_wrapper(X_y_binary, X_y_multi, X_y_regression):
...
+                          (binary, 0.25, {0: 800, 1: 200}),
+                          (binary, 0.1, {0: 800, 1: 200}),
+                          (multiclass, 1, {0: 800, 1: 800, 2: 800}),
+                          (multiclass, 0.5, {0: 800, 1: 400, 2: 400}),
+                          (multiclass, 0.25, {0: 800, 1: 200, 2: 200}),
+                          (multiclass, 0.1, {0: 800, 1: 150, 2: 80}),
+                          (multiclass, 0.01, {0: 800, 1: 150, 2: 50})])
+def test_make_balancing_dictionary(y, sampling_ratio, result):
+    dic = make_balancing_dictionary(y, sampling_ratio)
+    assert dic == result
```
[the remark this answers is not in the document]
bchen1116: Accidentally left it in the other file. Will remove!😅

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -22,6 +25,33 @@ def fit(self, X, y):
...
         return self
 
+    def _initialize_oversampler(self, X, y, sampler_class):
+        """Initializes the oversampler with the given sampler_ratio or sampler_ratio_dict.
+
+        Arguments:
+            Arguments:
+            X (ww.DataFrame): Training features
+            y (ww.DataColumn): Target features
+            sampler_class (Sampler): The sampler we want to initialize
```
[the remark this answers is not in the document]
bchen1116: Changing to `imblearn.BaseSampler`

evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,165 @@
...
+        error_msg = "imbalanced-learn is not installed. Please install using 'pip install imbalanced-learn'"
+        self.im = import_or_raise("imblearn.over_sampling", error_msg=error_msg)
+        parameters = {"sampling_ratio": sampling_ratio,
+                      "sampling_ratio_dict": sampling_ratio_dict,
+                      "k_neighbors": k_neighbors,
+                      "n_jobs": n_jobs}
+        parameters.update(kwargs)
+        super().__init__(parameters=parameters,
+                         component_obj=None,
+                         random_seed=random_seed)
```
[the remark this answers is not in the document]
bchen1116: The parent class I use is for both the undersampler and oversampler, so I'm not sure I can move this to the parent class. If I split the base classes into an undersampling base and oversampling base, that would work. @chukarsten do you think it'd make sense to do that?

evalml/tests/component_tests/test_oversamplers.py
bchen1116: This is just making sure we don't oversample when we don't need to

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -22,6 +25,33 @@ def fit(self, X, y):
...
+            if len(y_pd.value_counts()) == 2:
+                # imblearn has strange behavior when we set sampling_ratio = 1, so we choose 'auto' instead in that scenario
+                sampler_params['sampling_strategy'] = sampling_ratio if sampling_ratio != 1 else 'auto'
+            else:
+                # otherwise, we make the dictionary using the sampling_ratio
+                dic = make_balancing_dictionary(y_pd, sampling_ratio)
+                sampler_params['sampling_strategy'] = dic
+        sampler = sampler_class(**sampler_params, random_state=self.random_seed)
+        self._component_obj = sampler
+
```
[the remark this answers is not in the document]
bchen1116: @chukarsten, I simplified it to default to using dictionary in all cases. Hopefully that makes it easier to understand.

evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class SMOTENCSampler(BaseOverSampler):
+    """SMOTENC Oversampler component. Uses SMOTENC to generate synthetic samples. Works on a mix of nomerical and categorical columns.
+       This component is only run during training and not during predict."""
+    name = "SMOTENC Oversampler"
+    hyperparameter_ranges = {}
+
+    def __init__(self, categorical_features=[], sampling_ratio=0.25, sampling_ratio_dict=None, k_neighbors=5, n_jobs=-1, random_seed=0, **kwargs):
```
freddyaboulton, line 39: I think we should let woodwork determine which features are categorical features. Is the concern that the features created by the OHE are not marked as categorical?

In the meantime we should set `None` as the default value for `categorical_features`. Setting mutable default arguments can lead to bugs.

evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,90 @@
...
+    name = "SMOTE Oversampler"
+    hyperparameter_ranges = {}
+
+    def __init__(self, sampling_ratio=0.25, sampling_ratio_dict=None, k_neighbors=5, n_jobs=-1, random_seed=0, **kwargs):
+        """Initialize the SMOTE Oversampler component.
+
+        Arguments:
+            sampling_ratio (float): This is the goal ratio of the minority to majority class, with range (0, 1]. A value of 0.25 means we want a 1:4 ratio
+                of the minority to majority class after oversampling. If the targets are multiclass, will create a dictionary using this ratio. Defaults to 0.25.
+            sampling_ratio_dict (dict): Dictionary which has keys corresponding to each class, and the values are the number of samples we want to oversample to for each class key.
```
freddyaboulton, line 17: Wouldn't it be easier for users to specify the `sampling_ratio_dict` if they passed in the desired `sampling_ratio` rather than the desired counts?

My thought is that to use this properly in AutoMLSearch, users would need to know the size of the folds created by the data splitter, which isn't known ahead of time by most users and are also not equally sized. 

Even outside of search though, I think it's just easier to specify the desired "amount of sampling" when it's a percentage rather than an absolute count.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -54,3 +58,67 @@ def transform(self, X, y=None):
...
+            y (ww.DataColumn): Target features
+            sampler_class (imblearn.BaseSampler): The sampler we want to initialize
+        """
+        _, _, _, y_pd = self._prepare_data(X, y)
+        sampler_params = {k: v for k, v in copy.copy(self.parameters).items() if k not in ['sampling_ratio', 'sampling_ratio_dict']}
+        if self.parameters['sampling_ratio_dict'] is not None and len(self.parameters['sampling_ratio_dict']):
+            # dictionary provided and takes priority
+            sampler_params['sampling_strategy'] = self.parameters['sampling_ratio_dict']
+        else:
+            # no dictionary provided. We pass the float if we have a binary situation
```
freddyaboulton, line 104: Can you explain this comment? Looks like `make_balancing_dictionary` is making a dictionary regardless of whether it's a binary or multiclass problem.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -54,3 +58,67 @@ def transform(self, X, y=None):
...
+        Otherwise, we use will create the sampler_ratio_dict dictionary.
+
+        Arguments:
+            X (ww.DataFrame): Training features
+            y (ww.DataColumn): Target features
+            sampler_class (imblearn.BaseSampler): The sampler we want to initialize
+        """
+        _, _, _, y_pd = self._prepare_data(X, y)
+        sampler_params = {k: v for k, v in copy.copy(self.parameters).items() if k not in ['sampling_ratio', 'sampling_ratio_dict']}
+        if self.parameters['sampling_ratio_dict'] is not None and len(self.parameters['sampling_ratio_dict']):
```
freddyaboulton, line 100: What happens when a user passes in a `sampling_ratio_dict` but the keys don't match the actual labels in the target?

My concern is that by the time the target makes it to the sampler in AutoMLSearch, the target has been encoded. But users would need to know about the encoding ahead of time to specify this correctly. I don't think that's ideal. Users are much more likely to specify the dictionary using the unencoded labels because that's the data they see prior to running search.

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -0,0 +1,222 @@
...
+
+    np.testing.assert_equal(X, new_X.to_dataframe().values)
+    np.testing.assert_equal(y, new_y.to_series().values)
+
+
+@pytest.mark.parametrize("sampler", [SMOTESampler(sampling_ratio=1),
+                                     SMOTENCSampler(sampling_ratio=1, categorical_features=[1]),
+                                     SMOTENSampler(sampling_ratio=1)])
+@pytest.mark.parametrize("data_type", ["np", "pd", "ww"])
+def test_oversample_imbalanced_binary(data_type, sampler, make_data_type):
```
freddyaboulton, line 70: I think its worth adding a test case for `sampling_ratio_dict` given that it takes precedence over `sampling_ratio`

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -0,0 +1,222 @@
...
+    elif data_type == "pd":
+        X = X.values
+        y = y.values
+
+    np.testing.assert_equal(X, transform_X.to_dataframe().values)
+    np.testing.assert_equal(y, transform_y.to_series().values)
+
+
+@pytest.mark.parametrize("sampler", [SMOTESampler, SMOTENCSampler, SMOTENSampler])
+def test_oversample_seed_same_outputs(sampler, X_y_binary):
```
freddyaboulton, line 145: Nice!

requirements.txt
```diff
@@ -10,5 +10,5 @@ graphviz>=0.13
 seaborn>=0.11.1
 category_encoders>=2.0.0
 statsmodels >= 0.12.2
-imbalanced-learn>=0.7.0
+imbalanced-learn>=0.8.0
```
freddyaboulton, line 13: Why do we need `0.8.0` now?

freddyaboulton (changes requested):
@bchen1116 This looks awesome! I have some questions about passing `categorical_features` on init and how the `sampling_ratio_dict` should be specified. 

Everything else looks pretty good. The tests are great.

evalml/pipelines/components/transformers/samplers/oversamplers.py
bchen1116: @freddyaboulton I was going to have AutoMLSearch automatically detect the categorical features, but do you think it'd be better to have the component do that instead?

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: Yep, forgot to change the comment after editing.

evalml/tests/component_tests/test_oversamplers.py
bchen1116: Yep, that test is in `test_sampler_dic_overrides_ratio`

requirements.txt
bchen1116: `0.8.0` includes `SMOTEN`, which we use when the dataset is entirely categorical

evalml/pipelines/components/transformers/samplers/oversamplers.py
freddyaboulton: I think that should be the desired end-goal. It reminds me of what we did for the `TextFeaturizer` in automl. At first, we had automl search detect the natural language columns and we had to pass that down all the way to the `IterativeAlgorithm` in order for the pipelines to use it. Eventually we let the component identify the natural language columns and I think the whole implementation is simpler because of it.

If for whatever reason we can't get to that desired end-goal right now, like `SmoteNC` needs to know that the features created by a OHE are also categorical (I'm just brainstorming here), I think it's fine to leave as-is and file issues to improve our components so that we can keep track of what we need to do in order to get rid of the `categorical_features` argument.

requirements.txt
freddyaboulton: Sweet, thanks for explaining!

evalml/tests/component_tests/test_oversamplers.py
freddyaboulton: Thanks for pointing that out!

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: Ah yeah, that's a good point. I didn't think about this.

I think the best way to solve this would be to remove support for passing in a dictionary. This means that sampling for every target will be purely based on the float value passed for `sampling_ratio`, which will allow us to avoid issues of the dictionary keys differing from the targets we have currently

The con with this is that this removes functionality to finely alter the number of samples per each class. I can't think of a way that would allow us to avoid the issue and allow the dictionary input, since the target encoding is done in the pipeline class. What do you think @freddyaboulton? Do you have ideas for this?

evalml/pipelines/components/transformers/samplers/oversamplers.py
bchen1116: mm I see, so you're thinking of passing 

```python
sampling_ratio_dict = {
    0: 1,
    1: 0.5,
    2: 0.4
}
```
versus 
```python
sampling_ratio_dict = {
    0: 1000,
    1: 500,
    2: 400
}
```
where `0` is the majority class?

My initial reasoning was that it'd be easier for users to know how many samples of each class exist rather than the percentage ratios between these classes and the majority class. Definitely see the value of using ratios, so maybe I can file a follow-up issue for this? @freddyaboulton

evalml/tests/component_tests/test_oversamplers.py
bchen1116: @chukarsten I defaulted to updating the docstring to be clearer about this, but I don't think we need to do any logging or raise warnings for this?

evalml/pipelines/components/transformers/samplers/base_sampler.py
freddyaboulton: Thanks for following up @bchen1116 ! 

I think there's a third option, which is to leave as-is and not allow the use of the dictionary in `AutoMLSearch`. I think I prefer that for now? We still give users fine-tuned control of how much oversampling is done per class but they just have to do it prior to search. I think that's ok.

I think the solution you lay out in 2 is technically feasible but I don't like that the pipeline has to modify the parameters of one of its components mid-way during evaluation.

bchen1116: Mmm I see. That'd mean that if users define pipelines outside of AutoMLSearch, we should allow them to run into potentially confusing errors? ie when they have string target values and they pass a dictionary to the pipeline, we would then say those keys don't exist since we do encode them.

I think if we go with this approach, it'd be nice to file an issue to figure out how to best resolve this issue

evalml/pipelines/components/transformers/samplers/oversamplers.py
freddyaboulton: @bchen1116 Sounds good! I think we should tackle that before adding the oversamplers to AutoMLSearch though. Just cause I don't think we can expect users to specify the counts correctly since they (most likely) don't know the split sizes ahead of time.

evalml/pipelines/components/transformers/samplers/base_sampler.py
freddyaboulton: Ah, good point! Now that I think about it, I don't know if my proposal made a ton of sense at all. I was thinking that if users really wanted fine-grained control, they would create an oversampling component with a dict, oversample the data, and then pass the oversampled data to AutoMLSearch but that's not ideal because the sampling is not part of the pipeline.

Ok, so taking a step back, the problem is that users will likely misconfigure the keys of the sampling dictionary because they don't know about how the label encoding happens. 

With that in mind, I agree that our two options are those that you laid out before! Maybe a third one is that we only allow integer-valued keys ranging from 0 to # classes - 1. In essence we require users to pass in the encoded target values as keys. This may not be too harsh, the encoder encodes in ascending order so I think its straightforward for users to adapt the keys.

Given all this, I think we should go with option 1 (not support the dict) or option 3. I'm still hesitant on option 2 but maybe my opinion will change once I see the implementation. One thing to keep in mind is that we're probably going to represent the label encoder as a component soon. Maybe this is a point in favor of passing state around like @dsherry proposed in #2110! (but I still vote against writing separate encoder/decoder components 😂  )

bchen1116:
Filed an issue to track incorporating dictionary to oversamplers [here](https://github.com/alteryx/evalml/issues/2142). I will remove support for accepting dictionary input for now!

evalml/pipelines/components/transformers/samplers/oversamplers.py
chukarsten: That's a good observation, Freddy

chukarsten (approved):
Good changes! Good work!

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -0,0 +1,244 @@
...
+
+    X_com, y_com = component.fit_transform(X2, y)
+    X_im, y_im = imb_sampler.fit_resample(X, y)
+
+    np.testing.assert_equal(X_com.to_dataframe().values, X_im)
+    np.testing.assert_equal(y_com.to_series().values, y_im)
+    np.testing.assert_equal(sorted(y_im), expected_y)
+
+
+def test_smotenc_no_categorical_features_passed(X_y_binary):
```
freddyaboulton, line 220: nit-pick: This test name is misleading cause we're passing categorical features?

bchen1116: Good call!

evalml/pipelines/components/transformers/samplers/oversamplers.py
```diff
@@ -0,0 +1,59 @@
...
+        super().__init__("SMOTENC",
+                         sampling_ratio=sampling_ratio,
+                         k_neighbors=k_neighbors,
+                         n_jobs=n_jobs,
+                         random_seed=random_seed,
+                         **kwargs)
+
+    def _get_categorical(self, X):
+        X = infer_feature_types(X)
+        self.categorical_features = [i for i, val in enumerate(X.types['Logical Type'].items()) if str(val[1]) == 'Categorical']
```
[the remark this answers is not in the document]
bchen1116: For this, we need either a list of indices or a mask of booleans. I opted for the list of indices, and this was the easiest way to get the indices rather than the column names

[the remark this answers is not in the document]
freddyaboulton: Thanks for explaning!

PR #2091, opened 2021-04-06
Update pipeline API to accept component graph and other class attributes as parameters

### LINKED ISSUE #652: API ref: Expose docstring on class attributes
Currently we showcase the value of class attributes when available (see API reference for specific pipelines). However this doesn't show the docstring of these attributes and may be confusing since these attributes may rely other fields (such as `custom_name` and `name`).

We should also include another table similar to instance attributes but for class attributes.

### LINKED ISSUE #1956: Ability to save pipelines using pickle
*Background*
The pipeline `save` and `load` methods currently [use `cloudpickle`](https://github.com/cloudpipe/cloudpickle) instead of [python `pickle`](https://docs.python.org/3/library/pickle.html#:~:text=%E2%80%9CPickling%E2%80%9D%20is%20the%20process%20whereby,back%20into%20an%20object%20hierarchy.).

We closed #1400 a couple months ago. But this led to a bug (#1912) where our code was setting class attributes which apply across all instances of our template classes.

*Plan*
The short-term resolution for #1912 is to revert the change for #1400, so that `get_pipelines` returns a pipeline instance which is *not* `pickle`-able using python `pickle` but also does not have the buggy side-effect behavior when it comes to the class attributes like `name` and `component_graph`

Once that's done, this issue tracks figuring out how we can modify our approach to support python `pickle` of our pipelines, and filing issues to execute that plan.

### LINKED ISSUE #1984: Don't enforce pipeline class names to be in camel case
Currently, our pipelines will raise an `IllFormattedClassNameError` when you access the name of a pipeline class whose name isn't in camel case and that doesn't have a `custom_name` field set.

```python

class Mypipeline(BinaryClassificationPipeline):
    component_graph = ["Logistic Regression Classifier"]

pl = Mypipeline({})

pl.name

-------------------------------------------------------------------------
IllFormattedClassNameError              Traceback (most recent call last)
<ipython-input-10-0d7c0728e2b7> in <module>
----> 1 pl.name

~/sources/evalml/evalml/utils/gen_utils.py in __get__(self, _, klass)
    126 
    127     def __get__(self, _, klass):
--> 128         return self.func(klass)
    129 
    130 

~/sources/evalml/evalml/pipelines/pipeline_base.py in name(cls)
    102             if name == cls.__name__:
    103                 raise IllFormattedClassNameError("Pipeline Class {} needs to follow Pascal Case standards or `custom_name` must be defined.".format(cls.__name__))
--> 104         return name
    105 
    106     @classproperty

IllFormattedClassNameError: Pipeline Class Mypipeline needs to follow Pascal Case standards or `custom_name` must be defined.
```

Why do we need to enforce the class name to be camel case? I don't think any of our data structures or logic *needs* this constraint and I think it'd be nice to let users define their classes how they want.

angela97lin:
Closes #1956

Would also close #1984 and close #652 :)
**Pros**:
- Documentation for pipelines will no longer be an issue since we don't have to use classproperty (https://github.com/alteryx/evalml/issues/652)
- Can still use our representation of pipelines via classes if we want! Mostly tests that are breaking, but otherwise not too much has changed, really.
- Pickling works! This would close out #1956.

**Cons**:
- To use in AutoML, if __init__ is overriden then we need to override new/clone too.
- Conceptually a bit odd: parameters should only be used for first pipeline, then hyperparameter tuning happens. Maybe this is okay, since we do this internally anyways? What should the parameters of a pipeline *mean*?
- For time series, `gap` and `max_delay` need to be specified for input pipelines.

**General notes:**

- `allowed_pipeline` now takes in pipeline instances.
- For pipeline instances to be accepted by AutoMLSearch, must implement their own clone method? Maybe kwargs? Override init --> need to override new, clone
- Keeping track of pipeline instances via `_pipelines_searched` on the AutoMLSearch object. Can't add to `_results` because we deepcopy results often, and deepcopying LightGBM models causes a lot of printout 😬 
- Added concept of `frozen_parameters` to IterativeAlgorithm which represents parameters that should not be changed throughout AutoMLSearch.
 
Defining a pipeline class:
```
class LogisticRegressionBinaryPipeline(BinaryClassificationPipeline):
    component_graph = ['Imputer', 'One Hot Encoder', 'Standard Scaler', 'Logistic Regression Classifier']
    def __init__(self, parameters, random_seed=0):
        return super().__init__(self.component_graph, None, {}, None, random_seed)

    def new(self, parameters, random_seed):
        """Constructs a new pipeline with the same components, parameters, and random state.

        Returns:
            A new instance of this pipeline with identical components, parameters, and random state.
        """
        return self.__class__(parameters, random_seed)
    def clone(self):
        """Constructs a new pipeline with the same components, parameters, and random state.

        Returns:
            A new instance of this pipeline with identical components, parameters, and random state.
        """
        # return self.__class__(self.parameters, random_seed=self.random_seed)
        return self.__class__(self.parameters, self.random_seed)

LogisticRegressionBinaryPipeline({})


automl = AutoMLSearch(X, y, problem_type="binary", error_callback=raise_error_callback, max_batches=20, 
                     allowed_pipelines=[LogisticRegressionBinaryPipeline({})])

```

Without defining a new pipeline class:
```
pipeline = BinaryClassificationPipeline(["One Hot Encoder", "Random Forest Classifier"],
                                        "some custom name",
                                       parameters={},
                                       custom_hyperparameters={},
                                       random_seed=0)


pipeline.fit(X, y)
pipeline._component_graph
pipeline.estimator
# X , y = load_breast_cancer(return_pandas=True)

pipeline.score(X, y, ['log loss binary'])

pipeline.predict(X).to_series()
pipeline.predict_proba(X).to_dataframe()
```

**Pickling:**

Saving in one notebook:
image
```import pickle

with open("best.pkl", 'wb') as f:
    pickle.dump(automl.best_pipeline, f)
```

Loading in another notebook:
image
```
import pickle

best_pipeline = None
with open('best.pkl', 'rb') as f:
    best_pipeline = pickle.load(f)
best_pipeline.fit(X, y)
```

**To discuss:** 
- Should pipeline instances be aware of hyperparameters? Conceptually, there's no reason for them to be, AutoML should be, but each pipeline instance should only care about the parameters to use, not the hyperparameter range.
- A reason why it should be on the pipeline could be so that the same components in different pipelines can have different hyperparameter ranges. Ex: imputer on one pipeline has impute strategy of [mean, most_frequent], imputer on another has just [mean]. However, we currently use names as a way of identifying pipelines (no duplicate names allowed) so if we could store this on automl via pipeline, this could suffice? For now, working on cleaning things up. We could keep hyperparameters on pipelines in this PR if it requires too much change, and then clean up in separate PR.
- Pipeline has random seed, and is passed into AutoMLSearch as an allowed_pipeline. Do we use AutoML random seed? OR pipeline random seed?
- Right now, `pipeline_params` passed in as `custom_hyperparameters` but also used to instantiate what parameters the first pipeline should use.
- `generate_pipeline_code` won't work well with overridden `__init__`; this is true even in main now, but overriding init makes a lot more sense for custom pipeline classes now. But do we still need something like `generate_pipeline_code`?
- There are quite a few doc changes that would need to happen, too. Here, I've just updated the docs to pass but I think it's a better idea to file a separate issue since this PR is getting quite out of hand 😅 
- Pipeline utils `make_pipeline_from_components` is really no longer necessary, but will keep for now, can file something else to remove.

**Post-discussion with @freddyaboulton @chukarsten @dsherry:**
- There are certainly some flaws or gaps to a more ideal API here: We agreed that it's odd to pass in pipeline instances as `allowed_pipelines` when we don't use the parameters. We only use `component_graph`, `name`/`custom_name`, and `custom_hyperparameters`. Perhaps a better API would be to pass in component graphs to AutoMLSearch instead: 

```
automl = AutoMLSearch(X, y, problem_type="binary", allowed_component_graphs = {"Some name for component graph": ComponentGraph(...),})```
```
This would make it more clear that we are not using the parameters from the pipeline instance passed in
- We currently have two ways of specifying `custom_hyperparameters` for pipelines in AutoML: via the pipeline class' `custom_hyperparameters` and via `pipeline_params`. We will need to revisit whether it makes sense for the custom hyperparameter ranges to be stored on the pipeline. It makes sense for each component to be aware of the hyperparameter ranges.
- Don't use parameters on pipeline instance (ex: random_seed), use what is specified by AutoMLSearch. Pipeline instances are just the wrapper to grab component graph and name :d


evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -32,8 +32,8 @@ def __init__(self,
         self.max_iterations = max_iterations
         self._tuner_class = tuner_class or SKOptTuner
         self._tuners = {}
-        for p in self.allowed_pipelines:
-            self._tuners[p.name] = self._tuner_class(p.hyperparameters, random_seed=self.random_seed)
+        for pipeline in self.allowed_pipelines:
```
angela97lin, line 35: No functional change, just renaming because `p` makes it difficult to debug with via `pdb`

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -62,8 +62,11 @@ def next_batch(self):
 
         next_batch = []
         if self._batch_number == 0:
-            next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_seed=self.random_seed)
-                          for pipeline_class in self.allowed_pipelines]
+
+            # next_batch = [pipeline_class(parameters=self._transform_parameters(pipeline_class, {}), random_seed=self.random_seed)
+            #               for pipeline_class in self.allowed_pipelines]
+            next_batch = [pipeline.new(parameters=self._transform_parameters(pipeline, self._pipeline_params), random_seed=self.random_seed)
```
angela97lin, line 68: TODO: how to pass pipeline params to first pipelines... Probs need pipelines to just have..?

evalml/automl/automl_search.py
```diff
@@ -693,8 +695,8 @@ def _post_evaluation_callback(self, pipeline, evaluation_results, job_log):
         self._results['pipeline_results'][pipeline_id] = {
             "id": pipeline_id,
             "pipeline_name": pipeline.name,
-            "pipeline_class": type(pipeline),
-            "pipeline_summary": pipeline.summary,
+            'pipeline': pipeline,
```
angela97lin, line 698: Renaming to `pipeline` to indicate that it's an instance, not the class, that we're dealing with.

evalml/tests/pipeline_tests/test_graphs.py
```diff
@@ -14,14 +14,8 @@ def test_pipeline():
     class TestPipeline(BinaryClassificationPipeline):
```
angela97lin, line 14: Kept this class since it was helpful to have around, but deleted `hyperparameters` because it wasn't necessary.

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -173,6 +174,9 @@ def test_fast_permutation_importance_matches_sklearn_output(mock_supports_fast_i
 class PipelineWithDimReduction(BinaryClassificationPipeline):
```
angela97lin, line 174: Could simplify this to not use classes, but I like that the name makes it clear without overstuffing the parameterize field so I left as is :)

evalml/automl/automl_search.py
```diff
@@ -251,6 +251,7 @@ def __init__(self,
             'pipeline_results': {},
             'search_order': [],
         }
+        self._pipelines_searched = dict()
```
angela97lin, line 254: Using this to store pipeline instances rather than in `results` because we don't want to deepcopy (lightgbm prints lots of output when deepcopied)

docs/source/user_guide/pipelines.ipynb
```diff
@@ -32,7 +32,10 @@
     "from evalml.pipelines import MulticlassClassificationPipeline\n",
     "\n",
     "class CustomMulticlassClassificationPipeline(MulticlassClassificationPipeline):\n",
-    "    component_graph = ['Imputer', 'Random Forest Classifier']"
+    "    component_graph = ['Imputer', 'Random Forest Classifier']\n",
```
angela97lin, line 35: Minimal code to get this notebook to pass. However, it's worth discussing if any of this is even necessary anymore--we spend a majority of this notebook talking about different attributes that can be overridden in the pipeline class, but this can be done in the constructor in one fell swoop now. Still, I would prefer to file a separate issue for this if we decide to merge this spike.

evalml/pipelines/utils.py
```diff
@@ -210,7 +205,8 @@ def generate_pipeline_code(element):
         pipeline_list += ["{} = '{}'".format(k, v)] if isinstance(v, str) else ["{} = {}".format(k, v)]
 
     pipeline_string = "\t" + "\n\t".join(pipeline_list) + "\n" if len(pipeline_list) else ""
-
+    pipeline_string += "\n\tdef __init__(self, parameters, random_seed=0):"
```
angela97lin, line 208: Need to add init because right now we generate `pipeline(parameters)`, but the first parameter is now `component_graph`. I can file something to track this but I think a discussion that'll be useful to have is, is this still useful as is?

evalml/pipelines/utils.py
```diff
@@ -134,15 +134,12 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
...
-    class GeneratedPipeline(base_class):
-        custom_name = f"{estimator.name} w/ {' + '.join([component.name for component in preprocessing_components])}"
-        component_graph = complete_component_graph
-        custom_hyperparameters = hyperparameters
-
-    return GeneratedPipeline
+    return base_class(complete_component_graph, custom_name=name, parameters=parameters, custom_hyperparameters=custom_hyperparameters)
 
 
 def make_pipeline_from_components(component_instances, problem_type, custom_name=None, random_seed=0):
```
angela97lin, line 145: Also honestly no longer really useful, but I think it's okay to leave for now and delete in separate PR.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -594,27 +594,13 @@ def test_automl_allowed_pipelines_search(mock_fit, mock_score, dummy_binary_pipe
     assert start_iteration_callback.call_args_list[1][0][0] == dummy_binary_pipeline_class
 
 
-def test_categorical_hyperparam(X_y_multi):
```
angela97lin, line 597: Weird test, Simple Imputer isn't used and the hyperparameter ranges addressed that.

evalml/automl/automl_search.py
```diff
@@ -284,12 +285,19 @@ def __init__(self,
...
         if self.allowed_pipelines is None:
             logger.info("Generating pipelines to search over...")
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
-            self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
-
+            if is_time_series(self.problem_type):
+                self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, pipeline_params, custom_hyperparameters=pipeline_params) for estimator in allowed_estimators]
+            else:
+                self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, None, custom_hyperparameters=pipeline_params) for estimator in allowed_estimators]
```
angela97lin, line 300: FWIW, trying to separate out `pipeline_params` from custom_hyperparameters (though still passing to custom_hyperparameters since this is what's expected), but can't say I've throughly tested this 😓  just used this to pass parameters to make_pipeline for time series pipelines

@jeremyliweishih @freddyaboulton @dsherry

evalml/tests/conftest.py
```diff
@@ -439,6 +539,15 @@ class NonLinearRegressionPipeline(RegressionPipeline):
...
+
+        def __init__(self, parameters, random_seed=0):
+            return super().__init__(self.component_graph, None, parameters)
+
+        def new(self, parameters, random_seed=0):
+            return self.__class__(parameters, random_seed)
+
+        def clone(self):
+            return self.__class__(self.parameters, random_seed=self.random_seed)
     return NonLinearRegressionPipeline
```
chukarsten, line 551: I know I'm guilty of this, but do you think we need some concerted effort to move the testing pipelines that are generated in other test files etc here to conftest?  It's not at all in the scope of this PR, I would just wonder what your recommendations are, having gone through all this, to make the next sweeping change easier?

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -25,6 +25,7 @@ class DoubleColumns(Transformer):
     that in the future.
     """
     name = "DoubleColumns"
+    hyperparameter_ranges = {}
```
chukarsten, line 28: Do you remember why this is needed?

evalml/tests/pipeline_tests/classification_pipeline_tests/test_baseline_classification.py
```diff
@@ -74,6 +96,50 @@ def test_baseline_binary_mode():
...
+    assert cloned_clf.parameters == parameters
+
+    new_parameters = {
+        "Baseline Classifier": {
+            "strategy": "mode"
+        }
+    }
+    new_clf = clf.new(parameters=new_parameters)
+    assert new_clf.name == expected_name
+    assert new_clf.parameters == new_parameters
```
chukarsten, line 140: So is this test specific to the baseline pipeline classes?  Or is it the expected behavior of all of the pipelines now?

evalml/tests/pipeline_tests/regression_pipeline_tests/test_baseline_regression.py
```diff
@@ -2,7 +2,47 @@
...
 from pandas.testing import assert_series_equal
 
-from evalml.pipelines import BaselineRegressionPipeline
+from evalml.pipelines import (
+    BaselineRegressionPipeline,
+    MeanBaselineRegressionPipeline
+)
+
+
+def test_baseline_regression_init(X_y_binary):
```
chukarsten, line 11: Do you think these tests can move up the inheritance hierarchy?  Like maybe to PipelineBase?  I feel like I'm missing some context on special behaviors of some of the different pipeline child classes as to which exhibit what special behavior, but this seems to be universal to pipelines.  Let me know what you think and I can file an issue if you agree.

- [ ] @chukarsten file an issue if this testing behavior needs to move up

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -366,8 +416,13 @@ def test_make_pipeline_datetime_no_categorical(input_type, problem_type):
 
     for estimator_class in estimators:
         if problem_type in estimator_class.supported_problem_types:
-            pipeline = make_pipeline(X, y, estimator_class, problem_type)
-            assert isinstance(pipeline, type(pipeline_class))
+            parameters = {}
```
chukarsten, line 419: Is it possible to make a helper function to one line the parameters choices?

chukarsten (approved):
Angela, you're a mythic hero.  Really amazing work powering through all of this.  I've got the feeling that this is a good change to implement.  From a relative newbie to the code base, the change makes pipeline utilization a bit more intuitive to me, with the plus side that pickling should work.  I don't think I can foresee any issues with this going forward.  I'd personally accept the weirdness now of the allowed_pipelines accepting instances.  I think we can massage that to make that more natural.

No blocking changes requested, a few nits on docstrings having the default value where the other arguments state it. Again, great work.

evalml/pipelines/pipeline_base.py
```diff
@@ -342,15 +320,14 @@ def parameters(self):
             component_parameters['pipeline'] = self._pipeline_params
         return component_parameters
 
-    @classproperty
-    def default_parameters(cls):
+    def default_parameters(self):
```
freddyaboulton, line 323: I think you're planning on a separate PR to update the docs. In that PR I think we can make this a`property` along with `summary`, `problem_type`, `model_family`, `custom_name` etc so that they're documented!

evalml/pipelines/regression/baseline_regression.py
```diff
@@ -3,12 +3,25 @@
...
+    def new(self, parameters, random_seed=0):
+        return self.__class__(parameters, random_seed=random_seed)
+
+    def clone(self):
+        return self.__class__(self.parameters, random_seed=self.random_seed)
+
+
+class MeanBaselineRegressionPipeline(BaselineRegressionPipeline):
     """Baseline Pipeline for regression problems."""
-    _name = "Mean Baseline Regression Pipeline"
```
freddyaboulton, line 12: Very welcome unrelated change lol

evalml/pipelines/pipeline_base.py
```diff
@@ -46,69 +40,63 @@
...
-        """
-
-    custom_hyperparameters = None
-    custom_name = None
     problem_type = None
 
-    def __init__(self, parameters, random_seed=0):
+    def __init__(self, component_graph,
+                 custom_name=None,
+                 parameters=None,
```
freddyaboulton, line 47: Is there anything about the new design that made you set a default value for parameters? 

My intuition says "No" but just curious why the change was made.

evalml/pipelines/utils.py
freddyaboulton: If we merge this in, then I think the code generation should return something like `pipeline = BinaryClassifcationPipeline(component_graph=...., )` as opposed to creating subclasses in the style of the old api!

freddyaboulton (approved):
@angela97lin This looks great! It must have been frustrating to update all of the tests but I think the new api looks fantastic. I left a couple of small comments. 

This is a big PR and I think we've identified some areas of further improvement. I think it would be a good idea to create an epic and file some of the follow-up issues before merge.  That way we have an idea of what the finished product is supposed to look like before we merge this big change in. I think the list (as of my review) is the following:

1. Update AutoMLSearch to accept `allowed_component_graphs` instead of `allowed_pipelines`
2. Change the pipelines user guide to not create pipeline subclasses
3. Update code generation to not create pipeline subclasses
4. Make `get_*` methods of `PipelineBase` into `properties`
5. What to do with pipelines defined as classes left in the codebase, e.g. the baseline pipelines and text fixtures?

bchen1116 (approved):
Great job getting the SPIKE in so fast and fixing all of the tests! I left a few comments, but I think it'd be important to file issues for updating the docs as well as updating `generate_pipeline_code`.

evalml/pipelines/component_graph.py
```diff
@@ -67,16 +67,16 @@ def instantiate(self, parameters):
 
         Arguments:
             parameters (dict): Dictionary with component names as keys and dictionary of that component's parameters as values.
-                               An empty dictionary {} implies using all default values for component parameters.
+                               An empty dictionary {} or None implies using all default values for component parameters.
         """
         if self._is_instantiated:
             raise ValueError(f"Cannot reinstantiate a component graph that was previously instantiated")
 
+        parameters = parameters or {}
```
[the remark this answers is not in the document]
angela97lin: Hehe yeah, check this out: https://stackoverflow.com/questions/13710631/is-there-shorthand-for-returning-a-default-value-if-none-in-python

So if `parameters` is None, it'll return `{}`; I can make this more clear though by using `parameters if parameters is not None else {}`!

evalml/pipelines/utils.py
```diff
@@ -210,7 +205,8 @@ def generate_pipeline_code(element):
         pipeline_list += ["{} = '{}'".format(k, v)] if isinstance(v, str) else ["{} = {}".format(k, v)]
 
     pipeline_string = "\t" + "\n\t".join(pipeline_list) + "\n" if len(pipeline_list) else ""
-
+    pipeline_string += "\n\tdef __init__(self, parameters, random_seed=0):"
+    pipeline_string += "\n\t\treturn super().__init__(self.component_graph, self.name, parameters, custom_hyperparameters, random_seed=random_seed)\n"
```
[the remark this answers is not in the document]
angela97lin: Yup, filed #2160!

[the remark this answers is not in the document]
angela97lin: Though @bchen1116 I'm curious: what does the output of generate_pipeline_code look like now? Surprised that tests pass f we're missing functionality hehe

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2382,37 +2447,20 @@ def test_automl_check_high_variance_logs_warning(mock_fit_binary, X_y_binary, ca
...
-        custom_name = "My Pipeline 3"
-        component_graph = ["Random Forest Classifier"]
-
-    class OtherPipeline(BinaryClassificationPipeline):
-        custom_name = "Other Pipeline"
-        component_graph = ["Extra Trees Classifier"]
+    pipeline_1 = BinaryClassificationPipeline(component_graph=["Imputer", "Random Forest Classifier"], custom_name="Custom Pipeline")
+    pipeline_2 = BinaryClassificationPipeline(component_graph=["Imputer", "Logistic Regression Classifier"], custom_name="Custom Pipeline")
+    pipeline_3 = BinaryClassificationPipeline(component_graph=["Logistic Regression Classifier"], custom_name="My Pipeline 3")
+    pipeline_4 = BinaryClassificationPipeline(component_graph=["Random Forest Classifier"], custom_name="My Pipeline 3")
```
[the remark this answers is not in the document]
angela97lin: That's wonderful feedback! I certainly think this lowers the barrier to entry for users wanting to define their own pipelines for use too :)

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1915,13 +1973,20 @@ class DummyPipeline(BinaryClassificationPipeline):
...
+            super().__init__(self.component_graph, parameters=parameters, random_seed=random_seed)
 
-    automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", allowed_pipelines=[DummyPipeline],
+        def new(self, parameters, random_seed):
+            return self.__class__(parameters, random_seed=random_seed)
+
+        def clone(self):
+            return self.__class__(self.parameters, random_seed=self.random_seed)
+    pipelines = [DummyPipeline({})]
+    DummyPipeline.num_pipelines_different_seed = 0
```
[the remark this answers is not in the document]
angela97lin: This test is interesting, lol. It tries to check whether or not pipelines instantiated in AutoMLSearch use automl's random seed or not. We need to instantiate a pipeline instance to pass to AutoMLSearch, causing `num_pipelines_different_seed` to increase to 1. I'm setting here back to 0, so that we know the calls are strictly from automl. Alternatively, could have checked for 1 non-42-pipeline but I think this is slightly clearer.

evalml/tests/conftest.py
angela97lin: I think for most cases, pipelines that are shared amongst files are stored here, but we also create a lot of classes within tests to test for specific things. Some of these can be cleaned up, others are specific (ex: testing Catboost). Hopefully, this redesign means that we don't need to follow this pattern of creating classes for tests. I've left the ones in conftest alone because I figured in cases of reusability, maybe having a class is actually the way to go. In this PR I've also tried to rip out those one-off cases where we define a class for a specific test :')

evalml/tests/model_understanding_tests/test_permutation_importance.py
angela97lin: PipelineBase's `get_hyperparameters` has the line `component_hyperparameters = copy.copy(component_class.hyperparameter_ranges)` which fails if `DoubleColumns` does not have a `hyperparameter_ranges` attribute.

I think this was okay previously because we never tried to access the hyperparameters attribute--but now that it's done in init, we do and catch this!

evalml/pipelines/classification/baseline_binary.py
```diff
@@ -3,12 +3,25 @@
 
 class BaselineBinaryPipeline(BinaryClassificationPipeline):
     """Baseline Pipeline for binary classification."""
-    custom_name = "Baseline Classification Pipeline"
+    custom_name = "Baseline Binary Classification Pipeline"
     component_graph = ["Baseline Classifier"]
+    custom_hyperparameters = None
 
+    def __init__(self, parameters, random_seed=0):
+        return super().__init__(self.component_graph,
```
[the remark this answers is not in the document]
angela97lin: Long-term, I think it's okay to keep the classes as is since we use them in AutoMLSearch, and having the class makes it clear what the pipeline is for! :)

docs/source/user_guide/pipelines.ipynb
```diff
@@ -445,6 +437,9 @@
     "            \"numeric_impute_strategy\": ['mean', 'median']\n",
     "        }\n",
     "    }\n",
+    "    def __init__(self, parameters, random_seed=0):\n",
+    "        return super().__init__(self.component_graph, self.name, parameters, custom_hyperparameters, random_seed=random_seed)\n",
+    "\n",
     "\n",
     "pipeline_instance = CustomPipeline(parameters={\"Imputer\": {\"numeric_impute_strategy\": \"median\"}})\n",
     "code = generate_pipeline_code(pipeline_instance)\n",
```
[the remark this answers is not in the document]
angela97lin: Yup, tracking via #2158

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1300,15 +1321,15 @@ class Pipeline2(DummyPipeline):
...
     if objective.name.lower() == "cost benefit matrix":
         automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type_value, max_iterations=3,
-                              allowed_pipelines=[Pipeline1, Pipeline2], objective=objective(0, 0, 0, 0),
+                              allowed_pipelines=[Pipeline1({}), Pipeline2({})], objective=objective(0, 0, 0, 0),
                               additional_objectives=[], optimize_thresholds=False, n_jobs=1)
     elif problem_type_value == ProblemTypes.TIME_SERIES_REGRESSION:
         automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type_value, max_iterations=3,
-                              allowed_pipelines=[Pipeline1, Pipeline2], objective=objective,
+                              allowed_pipelines=[Pipeline1({'pipeline': {'gap': 0, 'max_delay': 0}}), Pipeline2({'pipeline': {'gap': 0, 'max_delay': 0}})], objective=objective,
                               additional_objectives=[], problem_configuration={'gap': 0, 'max_delay': 0}, train_best_pipeline=False, n_jobs=1)
```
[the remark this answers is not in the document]
angela97lin: Agreed! I bet we can remove it from `problem_configuration`, since the issue before was that we passed around pipeline classes and needed to keep track of the pipeline parameters to use when we initialize pipelines. Will file something to track this though :D

evalml/pipelines/pipeline_base.py
```diff
@@ -297,40 +284,31 @@ def _score_all_objectives(self, X, y, y_pred, y_pred_proba, objectives):
...
+            component_class = handle_component_class(component_info[0])
+            component_hyperparameters = copy.copy(component_class.hyperparameter_ranges)
+            if self.custom_hyperparameters and component_name in self.custom_hyperparameters:
+                component_hyperparameters.update(self.custom_hyperparameters.get(component_name, {}))
+            hyperparameter_ranges[component_name] = component_hyperparameters
         return hyperparameter_ranges
 
-    @property
-    def parameters(self):
+    def get_parameters(self):
```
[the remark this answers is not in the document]
angela97lin: Yup, you're right--by making `parameters` a property we fix both problems, so just updated to do that instead 😁

evalml/tests/pipeline_tests/classification_pipeline_tests/test_baseline_classification.py
angela97lin: Expected behavior, but just testing for baseline pipelines because we've defined classes and making sure the classes propagate expected values to the pipeline instance.

evalml/tests/pipeline_tests/regression_pipeline_tests/test_baseline_regression.py
angela97lin: Added some tests for `BinaryClassificationPipeline`, `MulticlassClassificationPipeline`, `RegressionPipeline` but kept these since they test for our specific baseline classes.

evalml/pipelines/pipeline_base.py
angela97lin: Decided to just make these attributes properties now, docs look like this: https://feature-labs-inc-evalml--2091.com.readthedocs.build/en/2091/generated/evalml.pipelines.PipelineBase.html#evalml.pipelines.PipelineBase 😁

freddyaboulton: Very cool!

angela97lin: Woot, thanks for the great suggestion! 🥳

evalml/pipelines/utils.py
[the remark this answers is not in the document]
bchen1116: @angela97lin Ah I forgot I added `json.loads` 🤦🏻‍♂️ This is what it looks like now:

In one file:
image

In a separate file:
image

Seems like the main issue is the `predict_proba` function that gets added in this new change. Thanks for filing!

evalml/automl/automl_search.py
```diff
@@ -720,6 +724,7 @@ def _post_evaluation_callback(self, pipeline, evaluation_results, job_log):
             "percent_better_than_baseline": percent_better_than_baseline[self.objective.name],
             "validation_score": cv_scores[0]
         }
+        self._pipelines_searched.update({pipeline_id: pipeline})
```
dsherry, line 727: @angela97lin @chukarsten @freddyaboulton @bchen1116 : tagging you all here since I see there was a discussion earlier in this file.

Trained pipelines can grow quite large, particularly tree-based models ([one example here](https://stackoverflow.com/questions/61591998/do-xgboost-models-tend-to-be-large-in-size-mb-on-disk)). If we keep references to trained pipelines, our memory consumption will grow linearly over the course of automl search, and could grow to be quite large if enough pipelines are evaluated.

We shouldn't save trained pipelines here.

However, all we really need is the ability to construct an untrained copy of the pipeline. So I have two thoughts on how we can achieve that here:
1. Save an untrained copy of the pipeline here instead of a trained copy
```python
self._pipelines_searched.update({pipeline_id: pipeline.clone()})
```
The code in `get_pipeline` below should still work fine because it calls `PipelineBase.new` to get a new instance, so that's unaffected.

2. Avoid the need for the `self._pipelines_searched` dict entirely by saving an index into `allowed_pipelines` in the pipeline results, like so:
```python
self._results['pipeline_results'][pipeline_id] = {
   ...,
   "pipeline_instance_index": allowed_pipelines.index(pipeline)
}
```
Both of those approaches have pros and cons, but IMO they are both preferable to saving trained pipelines.

What do you think @angela97lin ? I think option 1 is fine, and is not far from what you already have in place.

evalml/automl/automl_search.py
dsherry: Instead of if/else here based on `is_time_series` can we just do this
```python
self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, parameters=parameters, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
```
?

(RE another comment I left above, I renamed your variable `pipeline_params` to `parameters` in my example here)

I don't think the tuner needs to have access to the pipeline-level parameters, right? In other words, saying `custom_hyperparameters=self.pipeline_parameters` here like I did is fine.

evalml/automl/automl_search.py
```diff
@@ -290,19 +290,27 @@ def __init__(self,
         self.search_iteration_plot = None
         self._interrupted = False
 
+        if len(self.problem_configuration):
+            pipeline_params = {**{'pipeline': self.problem_configuration}, **self.pipeline_parameters}
```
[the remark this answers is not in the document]
dsherry: I just noticed this code was simply moved from below. Lol! Sorry I missed that. I stand by my suggestions though :)

evalml/pipelines/classification_pipeline.py
```diff
@@ -9,20 +9,13 @@
...
-        Required Class Variables:
-            component_graph (list): List of components in order. Accepts strings or ComponentBase subclasses in the list
-
-        Arguments:
-            parameters (dict): Dictionary with component names as keys and dictionary of that component's parameters as values.
-                 An empty dictionary {} implies using all default values for component parameters.
-            random_state (int): Deprecated - use random_seed instead.
-            random_seed (int): Seed for the random number generator. Defaults to 0.
-        """
+    def __init__(self, component_graph, custom_name=None, parameters=None, custom_hyperparameters=None, random_seed=0):
```
dsherry, line 12: Looks good!

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -18,28 +18,36 @@
 
 
 class TimeSeriesClassificationPipeline(ClassificationPipeline, metaclass=TimeSeriesPipelineBaseMeta):
-    """Pipeline base class for time series classifcation problems."""
+    """Pipeline base class for time series classification problems."""
 
-    def __init__(self, parameters, random_seed=0):
+    def __init__(self, component_graph, custom_name=None, parameters=None, custom_hyperparameters=None, random_seed=0):
```
dsherry, line 23: Looks good!

dsherry (changes requested):
@angela97lin , I will echo what @chukarsten said: this is truly heroic!!!! 🤩   This makes our pipeline datastructure a lot easier to use and to understand, pays down some pretty intricate tech debt, enables pickles 🥒  and is just generally epic. Please give yourself a huge round of applause for this! 👏👏🙏

I did have a few requests and suggestions for changes. But there's not a ton which is blocking! I filed a few more issues for some specific things and put them into the #2161 follow-on epic. I've read everything except your test changes. Will review those on the next pass.

Summary of key changes requested:
* To avoid memory issues, please don't save trained pipeline instances during automl search. I left suggestions.
* Let's delete the baseline pipeline classes! We don't need them anymore.
* For future-proofing and to prevent weird bugs with parameter ordering, please [use keywords](https://treyhunner.com/2018/04/keyword-arguments-in-python/) to name arguments when using the new pipeline constructor in our code. This came up in several places. I also think in the pipeline init we should make `parameters` the first keyword argument, after the new `component_graph` positional arg.
* I left a couple comments related to computing `allowed_pipelines`

cat_code
😆

evalml/pipelines/utils.py
```diff
@@ -134,15 +136,11 @@ def make_pipeline(X, y, estimator, problem_type, custom_hyperparameters=None):
     if custom_hyperparameters and not isinstance(custom_hyperparameters, dict):
         raise ValueError(f"if custom_hyperparameters provided, must be dictionary. Received {type(custom_hyperparameters)}")
 
-    hyperparameters = custom_hyperparameters
     base_class = _get_pipeline_base_class(problem_type)
+    name = f"{estimator.name} w/ {' + '.join([component.name for component in preprocessing_components])}"
+    parameters = parameters if parameters is not None else {}
```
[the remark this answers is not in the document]
angela97lin: Bahaha, initially had it as this, but decided to add this in for more clarity after @chukarsten's comment here: https://github.com/alteryx/evalml/pull/2091#discussion_r616360237

Okay with doing either though :')

evalml/pipelines/pipeline_base.py
```diff
@@ -46,73 +40,82 @@
...
-        if isinstance(self.component_graph, list):  # Backwards compatibility
-            self._component_graph = ComponentGraph().from_list(self.component_graph, random_seed=self.random_seed)
+
+        self.component_graph = component_graph
+
+        if isinstance(component_graph, list):  # Backwards compatibility
+            self._component_graph = ComponentGraph().from_list(component_graph, random_seed=self.random_seed)
         else:
-            self._component_graph = ComponentGraph(component_dict=self.component_graph, random_seed=self.random_seed)
+            self._component_graph = ComponentGraph(component_dict=component_graph, random_seed=self.random_seed)
```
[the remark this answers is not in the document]
angela97lin: Yes, agreed! Thank you for filing 🥂

docs/source/user_guide/pipelines.ipynb
```diff
@@ -80,7 +85,10 @@
     "                         random_seed=random_seed)\n",
     "    \n",
     "class CustomComponentMulticlassClassificationPipeline(MulticlassClassificationPipeline):\n",
-    "    component_graph = [NewTransformer, 'Random Forest Classifier']"
+    "    component_graph = [NewTransformer, 'Random Forest Classifier']\n",
+    "    \n",
+    "    def __init__(self, parameters, random_seed=0):\n",
+    "        super().__init__(self.component_graph, self.custom_name, parameters, random_seed=random_seed)\n"
```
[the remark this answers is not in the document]
angela97lin: Good eye, updated!

evalml/pipelines/utils.py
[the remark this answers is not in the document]
angela97lin: @bchen1116 Hmm, odd. I tried running on `main` and got this when trying to generate code for a pipeline from AutoMLSearch:

image

evalml/pipelines/utils.py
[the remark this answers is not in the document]
angela97lin: Edit: don't think I needed this at all, was leftover :)

evalml/pipelines/classification/baseline_binary.py
```diff
@@ -3,12 +3,25 @@
 
 class BaselineBinaryPipeline(BinaryClassificationPipeline):
```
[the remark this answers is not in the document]
angela97lin: Filed #2175 to track this! (@freddyaboulton tagged since you also mentioned this 😁 )

evalml/pipelines/pipeline_base.py
```diff
@@ -485,7 +488,15 @@ def clone(self):
         Returns:
             A new instance of this pipeline with identical components, parameters, and random state.
         """
-        return self.__class__(self.parameters, random_seed=self.random_seed)
+        return self.__class__(self.component_graph, self.name, self.parameters, self.custom_hyperparameters, self.random_seed)
+
+    def new(self, parameters, random_seed):
```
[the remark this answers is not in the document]
angela97lin: Yeah, that's a really good question, and I agree--I think we only call `clone()` in EngineBase. I filed #2176!

evalml/automl/automl_search.py
angela97lin: This is a really good call, thanks @dsherry! I agree that this is preferable to saving trained pipelines.

I opted for option 1, which is closer to what I was initially thinking of (storing a dictionary or something of the info a pipeline needs). I think option 2 is a little bit tricker--`allowed_pipelines` holds our original batch of pipelines, not all of our pipelines.

evalml/pipelines/pipeline_base.py
```diff
@@ -46,73 +40,82 @@
...
-            if name == cls.__name__:
-                raise IllFormattedClassNameError("Pipeline Class {} needs to follow Pascal Case standards or `custom_name` must be defined.".format(cls.__name__))
-        return name
-
-    @classproperty
-    def summary(cls):
-        """Returns a short summary of the pipeline structure, describing the list of components used.
+        self._pipeline_params = None
+        if parameters is not None:
+            self._pipeline_params = parameters.get("pipeline", {})
```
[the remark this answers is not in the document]
angela97lin: The `if` statement here is to catch if `parameters` is None, in which case we can't call `get` on `None`!

evalml/automl/automl_search.py
dsherry: Sounds good! Thanks :)

> allowed_pipelines holds our original batch of pipelines, not all of our pipelines.
Ah right. Its confusing... the `allowed_pipelines` hold the complete set of component graphs which automl will search. But yes as you said, those particular pipelines have an initial set of parameters, which happens to correspond to what our first batch will use. We will definitely have a chance to revisit this organization when we're ready to make some automl algo improvements.

evalml/pipelines/pipeline_base.py
```diff
@@ -298,31 +301,31 @@ def _score_all_objectives(self, X, y, y_pred, y_pred_proba, objectives):
...
-        component_graph = copy.copy(cls.component_graph)
+        component_graph = copy.copy(self._component_graph)
         if isinstance(component_graph, list):
             return handle_component_class(component_graph[-1]).model_family
         else:
-            order = ComponentGraph.generate_order(component_graph)
+            order = ComponentGraph.generate_order(component_graph.component_dict)
             final_component = order[-1]
-            return handle_component_class(component_graph[final_component][0]).model_family
+            return handle_component_class(component_graph[final_component].__class__).model_family
```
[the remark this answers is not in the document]
angela97lin: I'm using `self._component_graph` which is the standardized ComponentGraph object, rather than `self.component_graph` which is the set one by the user so different way of accessing. I think it's quite confusing that there are these two different but very similar attributes; my understanding is that we keep it to determine if the original was a linear graph or not but for now, just decided to keep functionality as is. 😅

evalml/automl/automl_search.py
angela97lin: Yes, agreed! Perhaps when we move towards holding just graph representations, we don't need to store the pipelines at all, and just need to keep track of the parameters as we search (#2159)

dsherry (commented):
@angela97lin I left a couple questions. Will review tests shortly and then approve!

docs/source/user_guide/pipelines.ipynb
```diff
@@ -134,8 +142,11 @@
     "class CustomPipeline(MulticlassClassificationPipeline):\n",
     "    component_graph = ['Imputer', 'One Hot Encoder', 'Logistic Regression Classifier']\n",
     "    custom_name = 'A custom pipeline name'\n",
-    "    \n",
-    "print(CustomPipeline.name)"
+    "        \n",
+    "    def __init__(self, parameters, random_seed=0):\n",
+    "        super().__init__(self.component_graph, parameters=parameters, custom_name=self.custom_name, random_seed=random_seed)\n",
```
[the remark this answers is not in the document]
angela97lin: Yup, I already filed https://github.com/alteryx/evalml/issues/2158 since I think this whole guide will need more rewriting 😂

evalml/automl/automl_search.py
```diff
@@ -289,24 +289,30 @@ def __init__(self,
         self.pipeline_parameters = pipeline_parameters if pipeline_parameters is not None else {}
         self.search_iteration_plot = None
         self._interrupted = False
+        self._frozen_pipeline_parameters = {}
+
+        parameters = copy.copy(self.pipeline_parameters)
+        if self.problem_configuration:
+            parameters.update({'pipeline': self.problem_configuration})
+            self._frozen_pipeline_parameters.update({'pipeline': self.problem_configuration})
```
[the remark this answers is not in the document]
angela97lin: Here's the logic right now in `main`: 

We pass the pipeline parameters, noticeably *without* the drop-columns parameters, to `make_pipelines`. This means that there are no custom hyperparameters set for the drop-columns component. Then, we use IterativeAlgo to handle all the wrangling of this specific case. In main, in we move the creation of `allowed_pipelines` from L297 to L302, we would error out:
```
ValueError: Default parameters for components in pipeline Decision Tree Classifier w/ Imputer + Drop Columns Transformer not in the hyperparameter ranges: Point (['most_frequent', 'mean', ['index_col'], 'gini', 'auto', 6]) is not within the bounds of the space ([('most_frequent',), ('mean', 'median', 'most_frequent'), ('index_col',), ('gini', 'entropy'), ('auto', 'sqrt', 'log2'), (4, 10)]).
```
(https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py).

Here, I had passed in `parameters` (rather than a copy) to the custom hyperparameters, which caused the same error. It's still a little odd to me that this logic is dependent on us *not* passing extra information for the drop columns transformer here and then passing that information to the IterativeAlgorithm to handle later, and I was hoping to ease some of that, because if we have more components that take in lists as parameters, we'll have to hardcode more code... but also okay with not doing this for now if it convolutes the logic even further!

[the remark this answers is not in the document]
angela97lin: EDIT: Actually, running some more tests, I think this might still be a good idea. Before, `allowed_pipelines` created classes. Now, we create instances. For time-series problems, we need to pass in the parameters. These parameters are found via `pipeline_parameters`. That's why this block of code was moved up, so that we can pass in the appropriate parameters to the time series pipeline instances. However, this logic that we currently have, `pipeline_parameters` also potentially contains smaller custom hyperparameter spaces, which will error out as parameters.

I noticed this when `test_automl_pipeline_params_kwargs` failed 😓 Open to filing an issue to somehow clean this up, or maybe this is the cleanup we need to be aware of as we restructure and rethink `IterativeAlgorithm`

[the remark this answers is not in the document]
dsherry: Ok, understood. If this looks good to you, I am on board. Yep, I hope we can revisit this after this PR and find a good way to clarify how we pass parameters to the automl algo.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1632,218 +1587,6 @@ class MockPipeline(pipeline_class):
...
-
-    mockAllCustom = MockAllCustom({})
-    expected_code = 'import json\n' \
-                    'from evalml.pipelines.binary_classification_pipeline import BinaryClassificationPipeline' \
-                    '\n\nclass MockAllCustom(BinaryClassificationPipeline):' \
-                    '\n\tcomponent_graph = [\n\t\tCustomTransformer,\n\t\tCustomEstimator\n\t]' \
-                    '\n\tname = \'Mock All Custom Pipeline\'\n\nparameters = json.loads("""{\n\t"My Custom Estimator": {\n\t\t"random_arg": false\n\t}\n}""")' \
-                    '\npipeline = MockAllCustom(parameters)'
-    pipeline = generate_pipeline_code(mockAllCustom)
-    assert pipeline == expected_code
```
dsherry, line 1844: @angela97lin after these tests are removed, do we still have test coverage of our code generation feature for pipelines? I know we've filed #2160 to track updating our `generate_pipeline_code` implementation--I will make a note there that these tests were deleted here but that we should add something similar back as part of that work.

dsherry (approved):
@angela97lin ok this looks ready to go! 👏👏👏

I left some comments but nothing blocking merge.

cat_laser_pickle

evalml/tests/conftest.py
```diff
@@ -388,14 +467,16 @@ def time_series_multiclass_classification_pipeline_class():
     class TSMultiPipeline(TimeSeriesMulticlassClassificationPipeline):
         """Logistic Regression Pipeline for time series multiclass classification problems."""
         component_graph = ['Delayed Feature Transformer', 'Logistic Regression Classifier']
+
+        def __init__(self, parameters, random_seed=0):
+            super().__init__(self.component_graph, parameters=parameters, random_seed=random_seed)
```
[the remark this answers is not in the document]
angela97lin: Nope, doesn't matter for some of these test classes because they're never used in AutoMLSearch, but that sounds good! Thank you!! 👍

evalml/automl/automl_search.py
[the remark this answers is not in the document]
angela97lin: Yep, I'll take another pass at this before merging, and filed #2187 to revisit this!

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Oh gosh, sorry for the confusion--I had just moved them to `evalml/tests/pipeline_tests/test_pipeline_utils.py` to match the structure (since they're located in evalml.pipeline.utils) 😅

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1733,7 +1774,7 @@ def test_iterative_algorithm_pipeline_hyperparameters_make_pipeline_other_errors
     }
     estimators = get_estimators('multiclass', [ModelFamily.EXTRA_TREES])
 
-    pipelines = [make_pipeline(X, y, estimator, 'multiclass', custom_hyperparameters) for estimator in estimators]
+    pipelines = [make_pipeline(X, y, estimator, 'multiclass', None, custom_hyperparameters) for estimator in estimators]
```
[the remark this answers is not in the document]
angela97lin: Yessss, thank you for catching these!!

PR #2097, opened 2021-04-06
Remove warnings in documentation 

### LINKED ISSUE #2011: Find simpler implementation to address warnings in documentation
PR [1991](https://github.com/alteryx/evalml/pull/1991) addresses suppressing warnings in the jupyter files in our documentation. However, there are drawbacks to this PR, the main ones being 1) that the cell run count now starts at [2] instead of [1], and 2) all future notebooks need to incorporate this hidden cell. 

This issue tracks adding a simpler implementation to address the issue. It will likely be solved by using an `.ipython/` config folder with a config file that suppresses the warnings.

bchen1116:
fix #2011 
testing to clean up documentation code
Before:
image

After:
image


jeremyliweishih (approved):
nice and clean 😄

docs/source/disable-warnings.py
```diff
@@ -0,0 +1,2 @@
+import warnings
+warnings.filterwarnings('ignore')
```
angela97lin, line 2: Just thinking out loud... do we want to disable *all* warnings? Is it possible to disable just `DeprecationWarning`s? I know warnings in the past have been useful in showing where some things are not as we expected, could be out of scope for this PR though.

angela97lin (approved):
LGTM! Just left one comment about whether we should filter all warnings, or just `DeprecationWarning`s. Okay to leave out of scope though, but could be good to file.

docs/source/disable-warnings.py
freddyaboulton: Agree on the idea but I also think it's best suited for a different PR. We'll have to modify [deprecate_arg](https://github.com/alteryx/evalml/blob/main/evalml/utils/gen_utils.py#L440) to issue a `DeprecationWarning` for starters which, although is a small change, seems out of scope for this PR. I think the intent was a refactor without changing behavior.

docs/source/conf.py
```diff
@@ -261,6 +263,8 @@ class AccessorMethodDocumenter(AccessorLevelDocumenter, MethodDocumenter):
 
 
 def setup(app):
+    os.makedirs("/home/docs/.ipython/profile_default/startup")
+    shutil.copy("disable-warnings.py", "/home/docs/.ipython/profile_default/startup/")
```
freddyaboulton, line 267: They let you run arbitrary code on start-up? 😨

freddyaboulton (approved):
Cool solution @bchen1116! Looks good to me.

docs/source/disable-warnings.py
bchen1116: Great idea! Filed a new issue [here](https://github.com/alteryx/evalml/issues/2107)

docs/source/conf.py
bchen1116: Yep, apparently! This is what the support team suggested

angela97lin: Let the evil commence

PR #2112, opened 2021-04-07
Fix conf.py file

bchen1116:
Seems like [our documentation is failing](8https://readthedocs.com/projects/feature-labs-inc-evalml/builds/607384/) to build due to a recent change in the conf.py file. This should fix the check and allow it to proceed safely.

Warning suppression still works
Before:
image

After:
image


angela97lin (approved):
Change looks good, but since this issue happened on main and not your branch, does that mean there's not a good way to test that this actually fixes the error until we merge it to main? 🤔

bchen1116:
@angela97lin Unfortunately, I think so. I assumed that the branch builds would be the same as the PR builds, but I think that isn't true. Potentially, I could try making this PR a branch build and see if it throws errors?

angela97lin:
@bchen1116 I see, that sounds like a good thing to try! By branch build you mean going into RtD and manually building right? If so, doesn't seem like too much of an extra cost to try :d

ParthivNaresh (approved):
Sweet

bchen1116:
@angela97lin the [doc builds](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/609038/) successfully and the [warnings do catch](https://evalml.alteryx.com/en/bc_fix_conf/user_guide/model_understanding.html) still! I believe this means that it shouldn't fail when we add this to main.

PR #2113, opened 2021-04-07
Remove Hyperparameter_ranges

### LINKED ISSUE #2104: Undersampler: don't expose hyperparameters to automl yet
I think we need to hold off on having automl search over sampler parameters until #1717 . So I think we need to remove `balanced_ratio` and `min_samples` from the undersampler `hyperparameter_ranges`.

I left a similar comment on #2079 RE the oversampler.

bchen1116:
fix #2104

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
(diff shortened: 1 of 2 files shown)


jeremyliweishih (approved):
nice!

ParthivNaresh (approved):
Looks solid!

Issue #2115, opened 2021-04-07
Partial Dependence Contour Plots Showing NaNs in our model understanding demo


freddyaboulton:
In our model understanding demo, our two-way partial dependence plot shows NaN in the right-most region of the plot:

![image](https://user-images.githubusercontent.com/41651716/113935611-bc17e980-97c4-11eb-9e25-4780106a9a74.png)

Is this expected? I think we may be extending the x axis range farther than it should?

chukarsten:
@freddyaboulton No, this is not expected, some sort of cleaning needs to be done to take care of this.

dsherry:
@freddyaboulton in #2117 you said:
> Fixes #2056, #2115

FYI for future reference, unfortunately [the GitHub "fixes" keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) doesn't support that syntax! You have to say
> Fixes #2056, fixes #2115

Closing this issue as it was fixed by #2117

PR #2117, opened 2021-04-07
Use a bar plot for one-way partial dependence plots of categorical features. Fix two-way plots for categoricals

### LINKED ISSUE #2056: Compute partial dependence for categorical variables

freddyaboulton:
### Pull Request Description
Fixes #2056, #2115

Our function for calculating the partial dependence plot works for categorical features, but our plot function for one-way plots always displayed the data with a scatter rather than the desired bar plot that the issue mentions. Our two-way plots were not working reliably as well. This PR addresses those issues.

Docs Example:

image

### Two way plot with one categorical
image

### Two way plot two categoricals
image

### Works for multiclass as well
image

-----


evalml/model_understanding/graphs.py
```diff
@@ -599,6 +616,33 @@ def partial_dependence(pipeline, X, features, percentiles=(0.05, 0.95), grid_res
     return data
 
 
+def _update_fig_with_two_way_partial_dependence(_go, fig, label_df, part_dep, features, is_categorical,
+                                                label=None, row=None, col=None):
+    y = label_df.index
```
freddyaboulton, line 621: Before we were setting y as the columns and x as the index but that's backwards. The row index is the "vertical" direction so it corresponds to the y axis. I think this is what was causing the NaNs in our docs mentioned in #2115. After the switch, they are gone:

image

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -374,7 +375,7 @@ def test_graph_partial_dependence_multiclass(logistic_regression_multiclass_pipe
     assert [fig_dict["data"][i]['name'] for i in range(3)] == ["class_0", "class_1", "class_2"]
 
     # Check that all the subplots axes have the same range
-    for suplot_1_axis, suplot_2_axis in [('axis2', 'axis3'), ('axis2', 'axis4'), ('axis3', 'axis4')]:
+    for suplot_1_axis, suplot_2_axis in [('axis', 'axis2'), ('axis', 'axis3'), ('axis2', 'axis3')]:
```
freddyaboulton, line 378: We technically make four subplots but only fill in 3 of them. Since before we made a single `update_xaxes` call, all four suplots would get updated. Now, we only do it for the first three suplots (the ones that have data) so we should check `axes`, `axes2`, and `axes3`.

evalml/model_understanding/graphs.py
chukarsten: lol i git blame myself..I remember feeling some qualms with this and should have done a better job checking.  Thanks for fixing it for me.

chukarsten (approved):
This looks great and the output graphs look great.  I thought I had originally ran into the NaN/missing columns bug when I first went through this and thought I fixed it.  I thought wrong.  Thanks for tackling this : looks really good.

evalml/model_understanding/graphs.py
```diff
@@ -665,44 +732,50 @@ def graph_partial_dependence(pipeline, X, features, class_label=None, grid_resol
...
             elif mode == "one-way":
                 x = label_df['feature_values']
                 y = label_df['partial_dependence']
-                fig.add_trace(_go.Scatter(x=x, y=y, line=dict(width=3), name=label),
-                              row=(i + 2) // 2, col=(i % 2) + 1)
+                if is_categorical:
+                    trace = _go.Bar(x=x, y=y, name=label)
+                else:
+                    trace = _go.Scatter(x=x, y=y, line=dict(width=3), name=label)
+                fig.add_trace(trace, row=(i + 2) // 2, col=(i % 2) + 1)
```
bchen1116, line 748: we can simplify this to `row=row, col=col`, since you defined them inlines 735-736

bchen1116 (approved):
Awesome! Left a comment on one quick fix, but LGTM!

evalml/model_understanding/graphs.py
freddyaboulton: Good call! I missed this when I was moving everything around.

PR #2118, opened 2021-04-08
Adding beta warning for time series problems in AutoML

### LINKED ISSUE #2094: Add "beta" warning to time series in automl
Since time series support is not yet ready, we should add a warning when any of the time series types are used in automl. And we should add a note to our docs as well, in the automl user guide.

I would update the "Detecting Problem Type" section to be titled "Problem Types in AutoML", mention the types we currently support, that time series types are in beta, and then add a sub-heading for "Detecting Problem Types" with the current content of that section.

### LINKED ISSUE #2096: Use as_completed in train_pipelines/score_pipelines
This issue tracks writing our own implementation of dask's `as_completed` function to pull off jobs from the queue when they are done. This will be used to consolidate the "while-loop" logic in `score_pipelines` and `train_pipelines`.

It's not exactly the same because the thing you get back from `get_result` is not the same between training and scoring and the logic for handling exceptions is not the same as well.

This might be a good spot for Max 's suggestion to use `as_completed` or something like it.

If we do that, I think this will look like:

```python
    for computation in as_completed(computations):
            try:
                fitted_pipeline = computation.get_result()
                fitted_pipelines[fitted_pipeline.name] = fitted_pipeline
            except Exception as e:
                logger.error(f'Train error for {pipeline.name}: {str(e)}')
                tb = traceback.format_tb(sys.exc_info()[2])
                logger.error("Traceback:")
                logger.error("\n".join(tb))
```

I think at that point we're not reusing the while-loop logic so the differences between the two methods might be clearer. 

If you're on board I'll file an issue after merge!

_Originally posted by @freddyaboulton in https://github.com/alteryx/evalml/pull/1975#discussion_r605169137_

freddyaboulton:
### Pull Request Description
Fixes #2094 

-----


docs/source/user_guide/automl.ipynb
```diff
@@ -16,7 +16,9 @@
...
+    "EvalML supports three common supervised ML problem types. The first is regression, where the target value to model is a continuous numeric value. Next are binary and multiclass classification, where the target value to model consists of two or more discrete values or categories. The choice of which supervised ML problem type is most appropriate depends on domain expertise and on how the model will be evaluated and used. \n",
+    "\n",
+    "EvalML is currently building support for supervised time series problems: time series regression, time series binary classification, and time series multiclass classification. While we've added some features to tackle these kinds of problems, our functionality is still being actively developed so please be mindful of that before using it. \n",
```
freddyaboulton, line 21: The issue originally mentions adding this blurb to the part where we detect problem types but I thought it made more sense here because we're already talking about the problem types we support.

angela97lin (approved):
LGTM! 👍

bchen1116 (approved):
🔜           . 🕰.   !!!

Issue #2119, opened 2021-04-09
Partial dependence null warnings not being issued for two-way partial dependence


freddyaboulton:
Repro

```python
from evalml.pipelines import BinaryClassificationPipeline
from evalml.model_understanding import partial_dependence
import pandas as pd
import numpy as np

class Pipeline(BinaryClassificationPipeline):
    component_graph = ["Imputer", "Random Forest Classifier"]
    
pl = Pipeline({})

X = pd.DataFrame({"a": [1, 2, 3],
                  "b": [4, 5, 6]})
y = pd.Series([0, 1, 0])
pl.fit(X, y)

pred_df = pd.DataFrame({"a":  np.random.random(99).tolist() + [None], "b": [1] + [None]*99})
partial_dependence(pl, pred_df, features=("a", "b"), grid_resolution=10)
```

I would expect the `There are null values in the features, which will cause NaN values in the partial dependence output.` warning to show up along with the `Feature 'b' is mostly one value, 1.0, and cannot be used to compute partial dependence.` value error.

dsherry:
@freddyaboulton in #2120  you said:
> Fixes #2073, #2119

FYI for future reference, unfortunately [the GitHub "fixes" keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) doesn't support that syntax! You have to say
> Fixes #2073, fixes #2119

Closing this issue as it was fixed by #2120

PR #2120, opened 2021-04-09
Adding error if columns are all Nan in partial dependence

### LINKED ISSUE #2073: Handle exception for fully null columns in partial dependence
Currently the below code errors out with `IndexError: index 0 is out of bounds for axis 0 with size 0` on ` if len(feature_list) and feature_list.value_counts(normalize=True).values[0] + 0.01 > percentiles[1]:`. Let's handle this error natively in EvalML.

```
import evalml
import numpy as np
import woodwork as ww

class DTBinaryClassificationPipeline(evalml.pipelines.BinaryClassificationPipeline):
    component_graph = ['Simple Imputer', 'Decision Tree Classifier']


X, y = evalml.demos.load_breast_cancer()
X = X.to_dataframe()
X['NaN'] = np.NaN

pipeline_dt = DTBinaryClassificationPipeline({})
pipeline_dt.fit(X, y)

from evalml.model_understanding.graphs import partial_dependence
partial_dependence(pipeline_dt, X, features='NaN')
```

freddyaboulton:
### Pull Request Description
Fixes #2073, #2119

-----

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -719,7 +719,7 @@ def graph_partial_dependence(pipeline, X, features, class_label=None, grid_resol
         pipeline (PipelineBase or subclass): Fitted pipeline
         X (ww.DataTable, pd.DataFrame, np.ndarray): The input data used to generate a grid of values
             for feature where partial dependence will be calculated at
-        features (int, string, tuple[int or string]): The target feature for which to create the partial dependence plot'fo for.
+        features (int, string, tuple[int or string]): The target feature for which to create the partial dependence plot for.
             If features is an int, it must be the index of the feature to use.
             If features is a string, it must be a valid column name in X.
             If features is a tuple of strings, it must contain valid column int/names in X.
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -558,7 +558,7 @@ def _raise_value_error_if_mostly_one_value(df, percentile):
             values.append(str(normalized_counts.index[0]))
 
     if one_value:
-        raise ValueError(f"Features {', '.join(one_value)} are mostly one value, ({', '.join(values)}), "
+        raise ValueError(f"Features ({', '.join(one_value)}) are mostly one value, ({', '.join(values)}), "
                          f"and cannot be used to compute partial dependence. Try raising the upper percentage value.")
 
 
@@ -602,6 +602,9 @@ def partial_dependence(pipeline, X, features, percentiles=(0.05, 0.95), grid_res
         ValueError: if the user provides a tuple of not exactly two features.
         ValueError: if the provided pipeline isn't fitted.
         ValueError: if the provided pipeline is a Baseline pipeline.
+        ValueError: if any of the features passed in are completely NaN
+        ValueError: if any of the features are low-variance. Defined as having one value occurring more than the upper
+            percentile passed by the user. By default 95%.
     """
 
     X = infer_feature_types(X)
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -522,12 +522,13 @@ def _put_categorical_feature_first(features, first_feature_categorical):
     new_features = features if first_feature_categorical else (features[1], features[0])
     return new_features
 
+
 def _get_feature_names_from_str_or_col_index(X, names_or_col_indices):
     """Helper function to map the user-input features param to column names."""
     feature_list = []
     for name_or_index in names_or_col_indices:
         if isinstance(name_or_index, int):
-            feature_list.append(X.columns[name_or_index])
+            feature_list.append(X.to_dataframe().columns[name_or_index])
         else:
             feature_list.append(name_or_index)
     return feature_list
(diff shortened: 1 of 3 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -558,25 +598,23 @@ def partial_dependence(pipeline, X, features, percentiles=(0.05, 0.95), grid_res
...
+    else:
+        feature_names = _get_feature_names_from_str_or_col_index(X, [features])
+
     if not pipeline._is_fitted:
         raise ValueError("Pipeline to calculate partial dependence for must be fitted")
     if pipeline.model_family == ModelFamily.BASELINE:
         raise ValueError("Partial dependence plots are not supported for Baseline pipelines")
 
-    feature_list = []
-    if isinstance(features, int):
```
freddyaboulton, line 567: Since we only check for `int` or `str` we miss the two-way case, which caused #2119

evalml/model_understanding/graphs.py
```diff
@@ -501,6 +501,46 @@ def graph_binary_objective_vs_threshold(pipeline, X, y, objective, steps=100):
...
+    nan_pct = df.isna().mean()
+    all_nan = nan_pct[nan_pct == 1].index.tolist()
+    all_nan = [f"'{name}'" for name in all_nan]
+
+    if all_nan:
+        raise ValueError("The following features have all NaN values and so the "
+                         f"partial dependence cannot be computed: {', '.join(all_nan)}")
+
+
+def _raise_value_error_if_mostly_one_value(df, percentile):
```
freddyaboulton, line 527: This is to fix #2119. May be scope-creep but it felt weird to have the all-NaN check work for two-way without fixing this warning. The implementation is pretty simple, we basically use the existing code for the one-way check on every column the user passed in, so I thought it was appropriate.

evalml/model_understanding/graphs.py
```diff
@@ -501,6 +501,46 @@ def graph_binary_objective_vs_threshold(pipeline, X, y, objective, steps=100):
...
+
+    for col in df.columns:
+        normalized_counts = df[col].value_counts(normalize=True) + 0.01
+        normalized_counts = normalized_counts[normalized_counts > percentile]
+        if not normalized_counts.empty:
+            one_value.append(f"'{col}'")
+            values.append(str(normalized_counts.index[0]))
+
+    if one_value:
+        raise ValueError(f"Features {', '.join(one_value)} are mostly one value, ({', '.join(values)}), "
```
bchen1116, line 540: nit-pick, but can we put parentheses around the feature names too?
```python
f"Features ({', '.join(one_value)}) are mostly
```
I think it's cleaner when we have multiple features

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -394,19 +395,49 @@ def test_graph_partial_dependence_multiclass(logistic_regression_multiclass_pipe
...
         partial_dependence(pipeline, X, features="random_col", grid_resolution=20)
-    with pytest.raises(ValueError, match="Feature 'random_col' is mostly one value, 1, and cannot be"):
+    with pytest.raises(ValueError, match="Features 'random_col' are mostly one value, \\(1\\), and cannot be"):
         partial_dependence(pipeline, X, features="random_col", percentiles=(0.01, 0.955), grid_resolution=20)
-    with pytest.raises(ValueError, match="Feature 'random_col' is mostly one value, 1, and cannot be"):
+    with pytest.raises(ValueError, match="Features 'random_col' are mostly one value, \\(1\\), and cannot be"):
         partial_dependence(pipeline, X, features=2, percentiles=(0.01, 0.955), grid_resolution=20)
+    with pytest.raises(ValueError, match="Features 'random_col' are mostly one value, \\(1\\), and cannot be"):
+        partial_dependence(pipeline, X, features=('A', "random_col"), percentiles=(0.01, 0.955), grid_resolution=20)
+    with pytest.raises(ValueError, match="Features 'random_col', 'random_col_2' are mostly one value, \\(1, 1\\), and cannot be"):
```
bchen1116, line 412: 👏

bchen1116 (approved):
Nice! LGTM

evalml/model_understanding/graphs.py
chukarsten: Is it weird not using our own datachecks to accomplish this type of thing?  This seems very much like a uniqueness and a no variance datacheck.

evalml/model_understanding/graphs.py
chukarsten: I think this is also another good time to add the raises column to the docstring of the partial dependence function so people can see the linkage of this ValueError to the upper percentile.  It took me a second, so I'm sure it'll cause a bit of head scratching.

chukarsten (approved):
Nice work.  Just a suggesting about adding "raises" to the docstring of the part dep function to suggest the relationship of the upper percentile limit to the "mostly a value" ValueError.  I also made a comment about using our datachecks to do these seeming datachecks, but I don't know how clean or reasonable that is.

ParthivNaresh (approved):
Overall looks great, excellent addition of helper functions

evalml/model_understanding/graphs.py
freddyaboulton: Good point @chukarsten . I'll add that now!

evalml/model_understanding/graphs.py
freddyaboulton: Good suggestion! There's definitely overlap in the logic. I'll file a spike issue.

PR #2122, opened 2021-04-09
Add natural language nan data check

### LINKED ISSUE #2000: LSA component errors due to `nan`s in natural language feature
Related to #1999 

**Repro**
```python
import evalml
import pandas as pd
import woodwork as ww
df = pd.read_csv('/Users/dylan.sherry/Downloads/bos_311_balanced.csv')
dt = ww.DataTable(df, logical_types={'reason': 'Categorical'})
dt = dt.drop('closed_dt') # ignore datetime feature because nans in it produce another bug
automl = evalml.automl.AutoMLSearch(X_train=dt, y_train=y, problem_type='multiclass')
automl.search()
```

Produces:
```
  File "/Users/dylan.sherry/development/evalml/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py", line 121, in transform
    X_lsa = self._lsa.transform(X[self._text_columns]).to_dataframe()
...
  File "/Users/dylan.sherry/development/evalml/evalml/pipelines/components/transformers/preprocessing/lsa.py", line 63, in transform
    transformed = self._lsa_pipeline.transform(X[col])
...
  File "/Users/dylan.sherry/.pyenv/versions/evalml/lib/python3.8/site-packages/sklearn/feature_extraction/text.py", line 219, in decode
    raise ValueError("np.nan is an invalid document, expected byte or "
...
Fold 0: Exception during automl search: np.nan is an invalid document, expected byte or unicode string.
```

jeremyliweishih:
Fixes #2000

- [x] Add to default_data_checks


evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
```diff
@@ -0,0 +1,112 @@
...
+    nl_nan_check = NaturalLanguageNaNDataCheck()
+    assert nl_nan_check.validate(pd.DataFrame({'natural_language': ["string_that_is_long_enough_for_natural_language", "string_that_is_long_enough_for_natural_language"]})) == {
+        "warnings": [],
+        "actions": [],
+        "errors": []
+    }
+
+
+def test_nl_nan_data_check_error_other_cols_with_nan():
+    data = pd.DataFrame(np.random.randint(0, 10, size=(10, 4)))
```
bchen1116, line 35: Is the purpose of this test to show that it doesn't throw any errors if there are `NaN` values in a dataset without `Natural Language` columns? It seems like it'd be more useful if you included a NL column with this (without nulls) to make sure there are no errors raised when other columns have `NaN`s.

evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
```diff
@@ -0,0 +1,112 @@
...
+                                  message_code=DataCheckMessageCode.NATURAL_LANGUAGE_HAS_NAN,
+                                  details={'columns': '0'}).to_dict()]
+    }
+
+    #  test 2D list
+    nl_col_without_nan = ["string_that_is_long_enough_for_natural_language", "string_that_is_long_enough_for_natural_language", "string_that_is_long_enough_for_natural_language"]
+    assert nl_nan_check.validate([nl_col, nl_col_without_nan]) == expected
+
+    # test np.array
+    assert nl_nan_check.validate(np.array([nl_col, nl_col_without_nan])) == expected
```
bchen1116, line 112: Can we also add a test on this behavior on an empty string in a natural language column? 

```python
cols = ["", "string_that_is_long_enough_for_natural_language", "string_that_is_long_enough_for_natural_language"]
```

bchen1116 (approved):
Awesome work! Left a couple comments on testing, but LGTM.

evalml/data_checks/natural_language_nan_data_check.py
```diff
@@ -0,0 +1,61 @@
...
+            "errors": [],
+            "actions": []
+        }
+
+        X = infer_feature_types(X)
+        nl_cols = _convert_woodwork_types_wrapper(X.select("natural_language").to_dataframe())
+        nan_columns = nl_cols.columns[nl_cols.isna().any()].tolist()
+        if len(nan_columns) > 0:
+            nan_columns = [str(col) for col in nan_columns]
+            cols_str = ', '.join(nan_columns) if len(nan_columns) > 1 else nan_columns[0]
```
freddyaboulton, line 56: I don't think we need the `if else` here? ', '.join(nan_columns) should be nan_columns[0] if nan_columns only has one element.

evalml/data_checks/natural_language_nan_data_check.py
```diff
@@ -0,0 +1,61 @@
...
+            ...    }
+        """
+        results = {
+            "warnings": [],
+            "errors": [],
+            "actions": []
+        }
+
+        X = infer_feature_types(X)
+        nl_cols = _convert_woodwork_types_wrapper(X.select("natural_language").to_dataframe())
```
freddyaboulton, line 52: I think it's possible to do this with `X.describe()['nan_count']` to save us the conversion.

freddyaboulton (approved):
Looks good @jeremyliweishih ! I agree with @bchen1116 's test suggestions.

evalml/data_checks/natural_language_nan_data_check.py
```diff
@@ -0,0 +1,61 @@
...
+        Returns:
+            dict: dict with a DataCheckError if NaN values are present in natural language columns.
+
+        Example:
+            >>> import pandas as pd
+            >>> import woodwork as ww
+            >>> import numpy as np
+            >>> data = pd.DataFrame()
+            >>> data['A'] = [None, "string_that_is_long_enough_for_natural_language"]
+            >>> data['B'] = ['string_that_is_long_enough_for_natural_language', 'string_that_is_long_enough_for_natural_language']
```
angela97lin, line 33: Hehe nit-pick / I don't think we did a good job of doing this before but maybe just explicitly set some cols as nat lang cols? Just thinking if WW ever changes their type inference, this might break and no longer be considered nat lang.

evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
```diff
@@ -0,0 +1,112 @@
+import numpy as np
+import pandas as pd
+import woodwork as ww
+
+from evalml.data_checks import (
+    DataCheckError,
+    DataCheckMessageCode,
+    NaturalLanguageNaNDataCheck
+)
+
```
angela97lin, line 10: General comment, we check using None here. While it shouldn't make a difference in behavior, can we also check with `np.nan`?

angela97lin (approved):
LGTM! Just left some nitpicks hehe

evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
jeremyliweishih: Can't use `np.nan` here since their will be conflicting types but I will go ahead and try with empty string.

evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
jeremyliweishih: @bchen1116 I believe an empty string should pass this check since the LSA component was only buging out out on `NaN`

`np.nan is an invalid document, expected byte or unicode string.` from #2000

evalml/data_checks/natural_language_nan_data_check.py
jeremyliweishih: nice! good catch 😄

evalml/data_checks/natural_language_nan_data_check.py
jeremyliweishih: cool will add! I'm going to use `describe_dict()` since its more straightforward for me to use a dict instead of a DT.

evalml/data_checks/natural_language_nan_data_check.py
jeremyliweishih: good call - I'll add it to the docstring and the explicit WW tests!

PR #2126, opened 2021-04-12
Rename `TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS` to `TARGET_MULTICLASS_NOT_TWO_EXAMPLES_PER_CLASS`

### LINKED ISSUE #2052: TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS
The DataCheckMessageCode "TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS" should be renamed and the documentation adjusted.

Per a conversation with Tom McGrath about what the error code meant led us into looking at the message.  It says this 👍 
```
    TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS = "target_multiclass_not_two_examples_per_class"
    """Message code for target data for a multi classification problem that does not have two examples per class."""
```

I think it just needs to be renamed, but a thorough examination of where it is used (which also confirms that it applies to the multiclass scenario) should be conducted.

angela97lin:
Closes #2052. This PR confirms that `TARGET_BINARY_NOT_TWO_EXAMPLES_PER_CLASS` actually refers to the multiclass case where there are not at least two examples per class, and thus it makes sense to rename the code to `TARGET_MULTICLASS_NOT_TWO_EXAMPLES_PER_CLASS`.


freddyaboulton (approved):

PR #2128, opened 2021-04-12
Add Undersampler Component to AutoMLSearch

### LINKED ISSUE #2053: Fix and clean component graph list support
In the [Undersampler component PR](https://github.com/alteryx/evalml/pull/2030), we note that there are limitations in the list support when defining a component graph [here](https://github.com/alteryx/evalml/pull/2030#discussion_r602536241). This isn't an issue when using the newer dictionary definition of a component graph, since we can define both `X` and `y` parents for each component. This issue tracks updating and fixing our list support in our `ComponentGraph` object and allow it to better define the parents by primarily tracking `y` of the parent transformers. 

Min repro:
```python
from evalml.pipelines.components import Transformer

class ChangeXY(Transformer):
    """Drops half of X and y"""
    name = "Drop X Y Transformer"
    hyperparameter_ranges = {}

    def __init__(self, random_seed=0, **kwargs):
        parameters = {}
        parameters.update(kwargs)

                         component_obj=None,
                         random_seed=random_seed)

    def fit(self, X, y=None):
        # do nothing
        return self

    def fit_transform(self, X, y):
        # alter both X and y
        indices = [i for i in range(len(X) // 2)]
        X = X.iloc[indices]
        y = y.iloc[indices]
        return X, y

        # do nothing here
        return X, y

X, y = load_breast_cancer()


# this works since we set the x and y parents
comp_graph = ComponentGraph({"imputer": ['Imputer'],
                             "one hot": ['One Hot Encoder', 'imputer.x'],
                             "changeXY": [ChangeXY, 'one hot.x'],
                             "logreg": ["Logistic Regression Classifier", 'changeXY.x', 'changeXY.y']})

comp_graph.instantiate({})
comp_graph.fit(X, y)


# this fails
class BinPipe(BinaryClassificationPipeline):
    component_graph = ['Imputer', 'One Hot Encoder', ChangeXY, 'Logistic Regression Classifier']

bp = BinPipe({})
bp.fit(X, y)
```

image

### LINKED ISSUE #2098: Use oversampling in automl
#2014 tracks creating an oversampler component. This issue tracks updating automl to use this component automatically.

Requirements before merge:
- Thorough testing of the components in AutoMLSearch to ensure performance is reasonable
- Testing to determine which samplers to default to for different dataset specs
Design doc [here](https://alteryx.quip.com/TW2nANznZcpN/Oversampling-in-AutoML-Search)

bchen1116:
fix #2053 
Part 1 support of #2098 (Only add undersampler for now, adding oversampler components will be a separate PR)
Initial implementation and testing. 2053 was fixed in order to support adding samplers to pipelines.

Quip doc [here](https://alteryx.quip.com/TW2nANznZcpN/Oversampling-in-AutoML-Search)

Perf tests [here](https://alteryx.quip.com/H6sHAv4Xb6iY/Undersampler-Component-Perf-Test)


evalml/pipelines/component_graph.py
```diff
@@ -228,16 +229,16 @@ def _compute_features(self, component_list, X, y=None, fit=False):
...
             self.input_feature_names.update({component_name: list(input_x.columns)})
-
             if isinstance(component_instance, Transformer):
                 if fit:
                     output = component_instance.fit_transform(input_x, input_y)
                 else:
                     output = component_instance.transform(input_x, input_y)
                 if isinstance(output, tuple):
                     output_x, output_y = output[0], output[1]
+                    most_recent_y = output_y
```
bchen1116, line 241: This keeps track of all the changes done to `y`, since not all transformers pass `y`. For example, when we pass a sampling output to the `StandardScaler`, we want the `y` output from the sampler to be passed to the estimator.

evalml/pipelines/utils.py
```diff
@@ -88,8 +94,23 @@ def _get_preprocessing_components(X, y, problem_type, estimator_class):
     if len(categorical_cols.columns) > 0 and estimator_class not in {CatBoostClassifier, CatBoostRegressor}:
         pp_components.append(OneHotEncoder)
 
+    if sampler_name is not None:
```
bchen1116, line 97: We add this before the scaler since a) The scaler turns boolean inputs (like those we expect from the OHE) to numeric values (-1, 1), which means we would be unable to choose the proper columns for SMOTENC, and b) previous literature suggests that it makes sense to sample before normalizing/standardizing the data for the model.

evalml/pipelines/utils.py
```diff
@@ -134,7 +157,9 @@ def make_pipeline(X, y, estimator, problem_type, parameters=None, custom_hyperpa
     problem_type = handle_problem_types(problem_type)
     if estimator not in get_estimators(problem_type):
         raise ValueError(f"{estimator.name} is not a valid estimator for problem type")
-    preprocessing_components = _get_preprocessing_components(X, y, problem_type, estimator)
+    if not is_classification(problem_type):
+        sampler_name = None
```
dsherry, line 161: Could we change this to
```python
if not is_classification(problem_type) and sampler_name is not None:
    raise ValueError(f"Sampling is unsupported for problem_type {str(problem_type)}")
```
?

evalml/automl/utils.py
```diff
@@ -117,3 +112,29 @@ def check_all_pipeline_names_unique(pipelines):
...
+        y (ww.DataColumn): The input target data
+        sampler_type (str): The sampler_type argument passed to AutoMLSearch
+        sampler_balanced_ratio (float): The ratio of min:majority targets that we would consider balanced,
+            or should balance the classes to.
+
+    Returns:
+        str: The string name of the sampling component to use
+    """
+    if sampler_type != 'auto':
+        return sampler_type
```
dsherry, line 131: To keep things simple, let's delete this, delete the `sampler_type` arg to this method and instead add this to automl search init where this method gets called:
```python
if self.sampler_method == 'auto':
    self._sampler_name = get_best_sampler_for_data(self.X_train, self.y_train, self.sampler_balanced_ratio)
```
This way, `get_best_sampler_for_data` stays simple, easier to test, and we can have separate automl search mocked testing to cover the sampler "auto" mode.

Update: ah yes I see your excellent test `test_automl_search_sampler_method` provides coverage for this, so that's great!

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -715,3 +715,61 @@ def test_tuning_threshold_objective(mock_predict, mock_fit, mock_score, mock_enc
...
+def test_automl_search_sampler_ratio(sampling_ratio, size, categorical_features, problem_type, mock_imbalanced_data_X_y, has_minimal_dependencies):
+    X, y = mock_imbalanced_data_X_y(problem_type, categorical_features, size)
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type, sampler_method='auto', sampler_balanced_ratio=sampling_ratio)
+    pipelines = automl.allowed_pipelines
+    if sampling_ratio <= 0.2:
+        # we consider this balanced, so we expect no samplers
+        assert not any(any("sampler" in comp.name for comp in pipeline.component_graph) for pipeline in pipelines)
+    else:
+
+        assert all(any("Undersampler" in comp.name for comp in pipeline.component_graph) for pipeline in pipelines)
```
dsherry, line 733: This is great

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -715,3 +715,61 @@ def test_tuning_threshold_objective(mock_predict, mock_fit, mock_score, mock_enc
...
+    else:
+        assert all(any(sampler_method in comp.name for comp in pipeline.component_graph) for pipeline in pipelines)
+
+
+@pytest.mark.parametrize("sampler_methods", [None, 'Undersampler'])
+@pytest.mark.parametrize("problem_type", ['binary', 'multiclass'])
+@patch("evalml.pipelines.components.estimators.Estimator.fit")
+@patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.5})
+@patch('evalml.pipelines.MulticlassClassificationPipeline.score', return_value={"Log Loss Multiclass": 0.5})
+def test_oversampling_in_pipelines(multiclass_score, binary_score, mock_fit, problem_type, sampler_methods, has_minimal_dependencies):
```
dsherry, line 756: Rename to `test_sampling_in_pipelines` yes?

dsherry: Actually, I think we should delete this test.

Reasoning: `test_automl_search_sampler_ratio` and `test_automl_search_sampler_method` verify that the sampler is or is not included in the pipelines. And from the perspective of test coverage for `AutoMLSearch`, that is sufficient. We can then write separate pipeline or component tests to verify that the undersampler component behaves as we expect it to. But I don't think we should mix the two.

evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -275,3 +275,10 @@ def test_automl_supports_time_series_regression(mock_fit, mock_score, X_y_regres
         for param_key, param_val in configuration.items():
             assert result['parameters']['Delayed Feature Transformer'][param_key] == configuration[param_key]
             assert result['parameters']['pipeline'][param_key] == configuration[param_key]
+
+
+def test_automl_regression_no_sampler(X_y_regression):
+    X, y = X_y_regression
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type='regression')
+    for pipeline in automl.allowed_pipelines:
+        assert not any("sampler" in c.name for c in pipeline.component_graph)
```
dsherry, line 284: 👍

evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -176,3 +171,24 @@ def test_tune_binary_threshold(mock_fit, mock_score, mock_predict_proba, mock_en
...
+        assert name_output is None
+    else:
+        assert name_output == 'Undersampler'
+
+
+@pytest.mark.parametrize("name", [None, 'Undersampler'])
+def test_get_best_sampler_for_data_named(name, mock_imbalanced_data_X_y):
+    X, y = mock_imbalanced_data_X_y("binary", "all", "large")
+    name_output = get_best_sampler_for_data(X, y, name, 0.25)
+    assert name_output == name
```
dsherry, line 194: RE my comment in the impl, you can delete this test.

evalml/tests/conftest.py
```diff
@@ -693,3 +693,36 @@ def _make_data_type(data_type, data):
 @pytest.fixture
 def fraud_100():
     return load_fraud(n_rows=100)
+
+
+@pytest.fixture
+def mock_imbalanced_data_X_y():
+    """Helper function to return an imbalanced binary or multiclass dataset"""
+    def _imbalanced_data_X_y(problem_type, categorical_columns, size):
+        """Take in args to determine the generated dataset to output. For our targets, we maintain a 1:5, or 0.2, class ratio of minority:majority.
```
dsherry, line 702: Nit-pick: "Take in args to determine the generated dataset to output." --> "Generates a dummy classification dataset with particular amounts of class imbalance and categorical input columns." Right?

evalml/tests/conftest.py
```diff
@@ -693,3 +693,36 @@ def _make_data_type(data_type, data):
...
+        We only generate minimum amount for X to set the logical_types, so the length of X and y will be different.
+
+        Arguments:
+            problem_type (str): Either 'binary' or 'multiclass'
+            categorical_columns (str): Determines how many categorical cols to use. Either 'all', 'some', or 'none'.
+            size (str): Either 'large' or 'small'. 'large' returns a dataset of size 420,000, while 'small' returns a size of 4200
```
dsherry, line 708: mind blown

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -801,3 +801,38 @@ def test_component_graph_dataset_with_target_imputer():
...
+    component_graph = ComponentGraph.from_list(component_list)
+    component_graph.instantiate({})
+    component_graph.fit(X, y)
+
+    component_list2 = ['Imputer', 'SMOTE Oversampler', 'Decision Tree Classifier']
+    component_graph2 = ComponentGraph.from_list(component_list2)
+    component_graph2.instantiate({})
+    component_graph2.fit(X, y)
+    assert mock_dt_fit.call_args[0][0] == mock_rf_fit.call_args[0][0]
+    assert mock_dt_fit.call_args[0][1] == mock_rf_fit.call_args[0][1]
```
dsherry, line 838: Niiiice, this is excellent.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -579,6 +579,33 @@ def test_make_component_list_from_actions():
...
+    estimators = get_estimators(problem_type=problem_type)
+
+    for estimator in estimators:
+        pipeline = make_pipeline(X, y, estimator, problem_type, sampler_name=samplers)
+        if has_minimal_dependencies and samplers is not None:
+            samplers = 'Undersampler'
+        # check that we do add the sampler properly
+        component_to_check = pipeline.component_graph
+        if samplers is not None and problem_type != 'regression':
+            # we add the sampler before the scaler if it exists
```
dsherry, line 600: Oh, I think I missed this detail. Why do this?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -579,6 +579,33 @@ def test_make_component_list_from_actions():
...
+            if component_to_check[-2].name == 'Standard Scaler':
+                assert component_to_check[-3].name == samplers
+            else:
+                assert component_to_check[-2].name == samplers
```
dsherry, line 604: Nit-pick: just reference `pipeline.component_graph` directly here instead of declaring variable `component_to_check`, fewer mental hops to make.

dsherry (changes requested):
@bchen1116 excellent!!!

The perf test results look great to me.

I had a few thoughts on little simplifications to the way the pieces connect, and what test coverage is necessary. Otherwise, I think this is almost ready to ship.

docs/source/demos/lead_scoring.ipynb
```diff
@@ -249,7 +250,8 @@
     "                                 data_splitter=data_splitter,\n",
     "                                 additional_objectives=[],\n",
     "                                 max_batches=1,\n",
-    "                                 optimize_thresholds=True)\n",
+    "                                 optimize_thresholds=True,\n",
+    "                                 sampler_method=None)\n",
```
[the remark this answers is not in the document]
bchen1116: I found that if we do over/undersampling, the performance on this model becomes very poor, with the AUC at 0.5. I removed the sampling to improve the performance of the model here.

image

evalml/pipelines/utils.py
bchen1116: @dsherry the comment was here since the next component that is added to the pipeline would be the standard scaler. I added the oversampling component after the OHE but before the standard scaler, since we should do the scaling after any sampling

evalml/tests/pipeline_tests/test_pipeline_utils.py
bchen1116: We want the sampler to come before the scaler since the scaler turns boolean values to int values ([0, 1] -> [-1, 1]), and since literature suggests scaling data should come after sampling

evalml/pipelines/component_graph.py
bchen1116: yep, that makes sense. I'll keep that in mind for the future!

evalml/automl/automl_search.py
```diff
@@ -296,6 +306,14 @@ def __init__(self,
             parameters.update({'pipeline': self.problem_configuration})
             self._frozen_pipeline_parameters.update({'pipeline': self.problem_configuration})
 
+        self.sampler_method = sampler_method
+        self.sampler_balanced_ratio = sampler_balanced_ratio
+        self._sampler_name = None
+        if is_classification(self.problem_type):
+            self._sampler_name = self.sampler_method
+            if self.sampler_method == 'auto':
+                self._sampler_name = get_best_sampler_for_data(self.X_train, self.y_train, self.sampler_method, self.sampler_balanced_ratio)
```
freddyaboulton, line 315: If `_sampler_name` is only used for making pipelines, maybe we should let `make_pipeline` handle the 'auto' to sampler name conversion? Or not store `_sampler_name` in the AutoMLSearch state?

docs/source/demos/lead_scoring.ipynb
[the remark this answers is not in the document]
dsherry: Understood. Yeah agreed. Would you mind filing that?

evalml/automl/automl_search.py
dsherry, line 315: Thanks! Looks great

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -801,3 +801,38 @@ def test_component_graph_dataset_with_target_imputer():
...
+
+
+@patch('evalml.pipelines.components.estimators.RandomForestClassifier.fit')
+@patch('evalml.pipelines.components.estimators.DecisionTreeClassifier.fit')
+def test_component_graph_sampler_same_given_components(mock_dt_fit, mock_rf_fit):
+    pytest.importorskip("imblearn.over_sampling", reason="Cannot import imblearn, skipping tests")
+    X = pd.DataFrame({"a": [i for i in range(100)],
+                      "b": [i % 3 for i in range(100)]})
+    y = pd.Series([0] * 90 + [1] * 10)
+    component_list = ['Imputer', 'SMOTE Oversampler', 'Random Forest Classifier']
```
freddyaboulton, line 828: How come we need both `component_graph.fit` and `component_graph2.fit` if we're only checking the output of the mocks after `component_graph2.fit`? I think we need a `reset_mock` in between if we want to keep both.

docs/source/demos/lead_scoring.ipynb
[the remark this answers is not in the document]
bchen1116: Filed one [here](https://github.com/alteryx/evalml/issues/2221)!

dsherry (approved):
@bchen1116 thanks, looking excellent 😁

evalml/automl/automl_search.py
```diff
@@ -174,6 +178,12 @@ def __init__(self,
 
             pipeline_parameters (dict): A dict of the parameters used to initalize a pipeline with.
 
+            sampler_method (str): The data sampling component to use in the pipelines if the problem type is classification and the target balance is smaller than the sampler_balanced_ratio.
+                Either 'auto', which will use our preferred sampler for the data, the name of the sampling component to use, or None. Defaults to 'auto'.
+
+            sampler_balanced_ratio (float): The minority:majority class ratio that we consider balanced, so a 1:4 ratio would be equal to 0.25. If the class balance is larger than this provided value,
```
freddyaboulton, line 184: @bchen1116 I see this doesn't get passed down to `IterativeAlgorithm`. What's the intended behavior of this parameter? I would assume that setting a value here controls how much sampling is done during search but as things are now, I believe each undersampler would get init with the default value regardless of the value of this parameter.

freddyaboulton (approved):
@bchen1116 This looks great! I have a question about the behavior of `sampler_balanced_ratio`. I think that should get resolved before merging!

evalml/automl/automl_search.py
bchen1116: Ah good call! I thought I had this but I must've deleted it at some point...  will fix!

evalml/automl/automl_search.py
bchen1116: Fixing, but I will leave this in order to set the balanced ratio for the component

evalml/tests/pipeline_tests/test_component_graph.py
bchen1116: These are two different mocks. One is mocking the RFC pipeline, and the other mocks the DTC pipeline

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -717,3 +717,36 @@ def test_tuning_threshold_objective(mock_predict, mock_fit, mock_score, mock_enc
...
         assert automl.best_pipeline.threshold == 0.6
+
+
+@pytest.mark.parametrize("problem_type", ['binary', 'multiclass'])
+@pytest.mark.parametrize("categorical_features", ['none', 'some', 'all'])
+@pytest.mark.parametrize("size", ['small', 'large'])
+@pytest.mark.parametrize("sampling_ratio", [0.8, 0.5, 0.25, 0.2, 0.1, 0.05])
+def test_automl_search_sampler_ratio(sampling_ratio, size, categorical_features, problem_type, mock_imbalanced_data_X_y, has_minimal_dependencies):
+    X, y = mock_imbalanced_data_X_y(problem_type, categorical_features, size)
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type, sampler_method='auto', sampler_balanced_ratio=sampling_ratio)
```
freddyaboulton, line 728: For paranoia's sake, can you search with `max_batches=2` and make sure each pipeline gets the right value during the search? Checking the rankings table should be sufficient. 

I see we use the `_frozen_pipelines_parameter` argument in `IterativeAlgorithm` and I don't think we have coverage for that (at least in `test_iterative_algorithm.py`)

Feel free to merge after that!

dsherry: Ah not a bad idea. So you want to run two batches, and assert that all the pipelines were given the right undersampler parameters? That makes sense.

I wonder if this sort of test would fit better in `test_iterative_algorithm.py`, since its testing `_frozen_pipelines_parameter` functionality and isn't specific to the undersampler component, right?

@bchen1116 if you run `search` please don't forget to mock `fit` and `score` :)

bchen1116: Yep, adding the test to `test_iterative_algorithm.py`!

PR #2129, opened 2021-04-12
Fix `test_describe_pipeline` for pandas 1.2.4

angela97lin:
Fixes test for latest dependency bot. Apparently each time the PR is created, it wipes any other changes so I needed to open my own PR 😬 

Pandas' update to 1.2.4 broke test_describe_pipeline tests. I believe this is because of the way that pandas treats object columns with float precision via https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.2.4.html. We call logger.info on the pandas DataFrame, which calls to_string. I've updated this PR to update our test to conform to pandas 1.2.4 since this is simply a difference in printing/logging, but it's worth updating test_describe_pipeline to be tolerant to such changes with pandas behavior in the future (could look into setting some pandas options hardcoded on our end). Unfortunately, this doesn't work here since the float config has also changed here :d

### DIFF
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -930,9 +930,9 @@ def test_describe_pipeline(mock_fit, mock_score, return_dict, caplog, X_y_binary
     assert "* strategy : mode" in out
     assert "Total training time (including CV): " in out
     assert "Log Loss Binary # Training # Validation" in out
-    assert "0                      1.000       66.0         34.0" in out
-    assert "1                      1.000       67.0         33.0" in out
-    assert "2                      1.000       67.0         33.0" in out
+    assert "0                      1.000     66.000       34.000" in out
+    assert "1                      1.000     67.000       33.000" in out
+    assert "2                      1.000     67.000       33.000" in out
     assert "mean                   1.000          -            -" in out
     assert "std                    0.000          -            -" in out
     assert "coef of var            0.000          -            -" in out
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
Thanks for the fix @angela97lin ! 

I think this is the smallest change we can make to fix our tests but I think its weird we're displaying the dataset size to three decimal places since it should always be an int. In "old pandas" we would display to one decimal place which is better but I still think its weird.

I wonder if the better thing to do is either display the fold size outside of the dataframe or round to ints?

angela97lin:
@freddyaboulton Yeah, I think the logic here is frustrating because we're converting to `object` to handle the NaN values (since we can't convert to int with a series that has nan without pandas throwing).

One option is to use the nullable Int64--that can be okay for this internal structure that isn't expected to be modified by the user. 

Also posting this: https://stackoverflow.com/questions/39690742/convert-float-to-int-and-leave-nulls, will try to see if any of these solutions float our boat :)

evalml/tests/automl_tests/test_automl.py
```diff
@@ -930,9 +930,9 @@ def test_describe_pipeline(mock_fit, mock_score, return_dict, caplog, X_y_binary
     assert "* strategy : mode" in out
     assert "Total training time (including CV): " in out
     assert "Log Loss Binary # Training # Validation" in out
-    assert "0                      1.000     66.000       34.000" in out
-    assert "1                      1.000     67.000       33.000" in out
-    assert "2                      1.000     67.000       33.000" in out
+    assert "0                      1.000         66           34" in out
```
freddyaboulton, line 933: 👏  🥳  💪

Issue #2132, opened 2021-04-12
AutoMLSearch does not apply pipeline_parameters to custom pipelines 


freddyaboulton:
Repro

```python
from evalml.demos import load_breast_cancer
from evalml.pipelines import BinaryClassificationPipeline
from skopt.space import Categorical
from evalml.automl import AutoMLSearch
import pytest


class PipeLine(BinaryClassificationPipeline):
    component_graph = ["One Hot Encoder", "Random Forest Classifier"]
    
X, y = load_breast_cancer()
automl = AutoMLSearch(X, y, problem_type="binary", allowed_pipelines=[PipeLine],
                      pipeline_parameters={"Random Forest Classifier": {"n_estimators": Categorical([50])}},
                      max_batches=2)
automl.search()

rf_params = automl.full_rankings.iloc[:-1].parameters.map(lambda d: d["Random Forest Classifier"]['n_estimators'])

assert set(rf_params) != {50}
```

I think the issue is that `pipeline_parameters` [is only used](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L291) in `allowed_pipelines` is None.

dsherry:
@freddyaboulton in #2135 you said:
> Fixes #2049, #2132, #2125

FYI for future reference, unfortunately [the GitHub "fixes" keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) doesn't support that syntax! You have to say
> Fixes #2049, fixes #2132, fixes #2125

Closing this issue as it was fixed by #2135

Issue #2125, opened 2021-04-12
Pipeline custom hyperparameters don't work for linear pipelines with duplicate components


freddyaboulton:
Repro

```python
from evalml.pipelines import BinaryClassificationPipeline
from skopt.space import Integer
class PipeLineLinear(BinaryClassificationPipeline):
    custom_hyperparameters = {"One Hot Encoder_1": {"top_n": Integer(10, 50)}}
    component_graph = ["One Hot Encoder", "One Hot Encoder", "Random Forest Classifier"]

assert "One Hot Encoder_1" not in PipeLineLinear.hyperparameters
assert PipeLineLinear.hyperparameters == {'One Hot Encoder': {},
                                         'Random Forest Classifier': {'n_estimators': Integer(10, 1000),
                                                                      'max_depth': Integer(1, 10)}}
```

There is no problem when the component graph is specified as a dictionary:

```python
class PipeLineDict(BinaryClassificationPipeline):
    custom_hyperparameters = {"OHE_1": {"top_n": Integer(10, 50)}}
    component_graph = {"OHE": ["One Hot Encoder"],
                       "OHE_1": ["One Hot Encoder", "OHE"],
                       "Random Forest Classifier": ["Random Forest Classifier", "OHE_1"]}

assert PipeLineDict.hyperparameters == {"OHE": {},
                                        "OHE_1": {"top_n": Integer(10, 50)},
                                        'Random Forest Classifier': {'n_estimators': Integer(10, 1000),
                                                                      'max_depth': Integer(1, 10)}}
```

This is similar to #2049

dsherry:
@freddyaboulton in #2133  you said:
> Fixes #2049, #2132, #2125

FYI for future reference, unfortunately [the GitHub "fixes" keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) doesn't support that syntax! You have to say
> Fixes #2049, fixes #2132, fixes #2125

Closing this issue as it was fixed by #2133

Issue #2131, opened 2021-04-12
AutoMLSearch can't tune pipeline with custom hyperparameters if default param not in custom range


freddyaboulton:
Repro

```python
from evalml.demos import load_breast_cancer
from evalml.pipelines import BinaryClassificationPipeline
from skopt.space import Categorical
from evalml.automl import AutoMLSearch
import pytest


class PipelineDict(BinaryClassificationPipeline):
    custom_hyperparameters = {"One Hot Encoder": {"top_n": Categorical([20])}}
    component_graph = {"One Hot Encoder": ["One Hot Encoder"],
                       "Random Forest Classifier": ["Random Forest Classifier", "One Hot Encoder"]}
    
X, y = load_breast_cancer()
automl = AutoMLSearch(X, y, problem_type="binary", allowed_pipelines=[PipelineDict],
                      max_batches=2)
with pytest.raises(ValueError,
                   match="Default parameters for components in pipeline Pipeline Dict not in the hyperparameter ranges"):
    automl.search()
```

I think the problem is that the first batch uses the default parameters. When we call `add_result` with a set of hyperparameters that are not in the range that we use to create the tuner, skopt will throw an error.

We have a [mechanism to deal with this](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L141) when the user specifies custom parameters with `pipeline_params` AutoMLSearch argument but we don't use it for custom_hyperparameter ranges.

I think this is a point in favor of deprecating the custom hyperparameters for pipelines like we discussed doing for #2091. In my opinion, the root cause is that we have two apis for overriding the hyperparameters used in search and we're struggling to maintain both of them.

christopherbunn:
@freddyaboulton I'm having trouble reproducing this on our current main branch. I adjusted the example to work with our current API below. Am I missing something?

```
from evalml.demos import load_breast_cancer
from evalml.pipelines import BinaryClassificationPipeline, MulticlassClassificationPipeline
from skopt.space import Categorical
from evalml.automl import AutoMLSearch
import pytest

X, y = load_breast_cancer()
automl = AutoMLSearch(X, y, problem_type="binary", 
                      allowed_component_graphs={"Pl": ["One Hot Encoder", "Random Forest Classifier"]}, 
                      custom_hyperparameters={"One Hot Encoder": {"top_n": Categorical([20])}}, 
                      max_iterations=2)
automl.search()
```

[...]

Issue #2130, opened 2021-04-12
AutoMLSearch: clean up API for freezing parameters and constraining hyperparam ranges


freddyaboulton:
Repro

```python
from evalml.demos import load_breast_cancer
from evalml.pipelines import BinaryClassificationPipeline
from skopt.space import Categorical
from evalml.automl import AutoMLSearch
import pytest

X, y = load_breast_cancer()    

automl = AutoMLSearch(X, y, problem_type="binary",
                      max_batches=2,
                      pipeline_parameters={"Imputer": {"numeric_impute_strategy": "most_frequent"}})
automl.search()
all_imputer_params = automl.full_rankings.iloc[:-1].parameters.map(lambda d: d['Imputer']['numeric_impute_strategy'])
assert set(all_imputer_params) == {'most_frequent', 'mean'}
```

However using `skopt.space.Categorical` works.

```python
automl = AutoMLSearch(X, y, problem_type="binary",
                      max_batches=2,
                      pipeline_parameters={"Imputer": {"numeric_impute_strategy": Categorical(["most_frequent"])}})
automl.search()
all_imputer_params = automl.full_rankings.iloc[:-1].parameters.map(lambda d: d['Imputer']['numeric_impute_strategy'])
assert set(all_imputer_params) == {'most_frequent'}
```

I think the problem is that skopt does not recognize `most_frequent` as a valid hyperparameter space so it will silently not include it in the `proposed_parameters` dict. Note that we can't specify `["most_frequent"]` because of #2028. 

I think the right thing to do is to not let users pass in "scalar-valued" hyperparameters.

dsherry:
Thanks for this. I would categorize this as saying the `pipeline_parameters` input is not behaving the way we want it to. Not sure I follow the scalar-valued suggestion yet but I think we can update our tuner code to avoid this issue.

jeremyliweishih:
[Here](https://github.com/alteryx/evalml/pull/2138#issuecomment-822669609) I had a very quick spike on separating custom hyperparameter and pipeline parameter functionality. Ideally, I still think we should try to separate out custom hyperparameters and parameters passed into pipelines. Don't have a concrete plan on how to achieve that but will revist after #2091 and #2130.

dsherry:
This issue lists at least a couple problems with our current approach to handling freezing parameter values and accepting user-specified hyperparameter ranges. Let's use it to track simplifying and cleaning up the API for doing this.

angela97lin:
[...]

PR #2133, opened 2021-04-13
Fix AutoMLSearch so users can specify custom hyperparameters to pipelines with duplicate components

### LINKED ISSUE #2049: AutoML cannot handle updating pipelines parameters with duplicate components
If we have two OHEs, for example, we cannot set the parameters for the second OHE.

```

X, y = load_breast_cancer() 
    component_graph = ["One Hot Encoder", "One Hot Encoder", "Random Forest Classifier"]

automl = AutoMLSearch(X, y, problem_type="binary", allowed_pipelines=[PipeLine], pipeline_parameters={"One Hot Encoder": {"top_n":100},
                                                                                                      "One Hot Encoder_1": {"top_n":1}}) 
automl.search()
```

This search runs fine, but looking at a pipeline:
```
automl.get_pipeline(1)
```

Gives...
```
PipeLine(parameters={'One Hot Encoder':{'top_n': 100, 'features_to_encode': None, 'categories': None, 'drop': 'if_binary', 'handle_unknown': 'ignore', 'handle_missing': 'error'}, 'One Hot Encoder_1':{'top_n': 10, 'features_to_encode': None, 'categories': None, 'drop': 'if_binary', 'handle_unknown': 'ignore', 'handle_missing': 'error'}, 'Random Forest Classifier':{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1},})
```

Notice that the second OHE (One Hot Encoder_1) holds the default parameters, and not the specified ones.

freddyaboulton:
### Pull Request Description
Fixes #2049, #2132, #2125

When I started, I wasn't planning on a three-in-one PR but we can't fix #2049 without fixing #2132.

Without fixing #2125 , users can't apply custom hyperparameters by attaching them to the pipeline definition so automl support for pipelines with duplicate components would still be broken. 

-----

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -298,7 +298,11 @@ class AutoMLSearch:
         else:
             for pipeline_class in self.allowed_pipelines:
                 if self.pipeline_parameters:
-                    pipeline_class.custom_hyperparameters = self.pipeline_parameters
+                    if pipeline_class.custom_hyperparameters:
+                        for component_name, params in self.pipeline_parameters.items():
+                            pipeline_class.custom_hyperparameters[component_name] = params
+                    else:
+                        pipeline_class.custom_hyperparameters = self.pipeline_parameters
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -291,7 +291,8 @@ class AutoMLSearch:
             self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
         else:
             for pipeline_class in self.allowed_pipelines:
-                pipeline_class.custom_hyperparameters = self.pipeline_parameters
+                if self.pipeline_parameters:
+                    pipeline_class.custom_hyperparameters = self.pipeline_parameters
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,4 +1,5 @@
 import copy
+import inspect
 import sys
 import time
 import traceback
@@ -37,6 +38,7 @@ from evalml.pipelines import (
     MeanBaselineRegressionPipeline,
     ModeBaselineBinaryPipeline,
     ModeBaselineMulticlassPipeline,
+    PipelineBase,
     TimeSeriesBaselineBinaryPipeline,
     TimeSeriesBaselineMulticlassPipeline,
     TimeSeriesBaselineRegressionPipeline
@@ -260,6 +262,10 @@ class AutoMLSearch:
         except ImportError:
             logger.warning("Unable to import plotly; skipping pipeline search plotting\n")
 
+        if allowed_pipelines is not None and not isinstance(allowed_pipelines, list):
+            raise ValueError("Parameter allowed_pipelines must be either None or a list!")
+        if allowed_pipelines is not None and not all(inspect.isclass(p) and issubclass(p, PipelineBase) for p in allowed_pipelines):
+            raise ValueError("Every element of allowed_pipelines must be a subclass of PipelineBase!")
         self.allowed_pipelines = allowed_pipelines
         self.allowed_model_families = allowed_model_families
         self._automl_algorithm = None
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -289,6 +289,9 @@ class AutoMLSearch:
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
             self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
+        else:
+            for pipeline_class in self.allowed_pipelines:
+                pipeline_class.custom_hyperparameters = self.pipeline_parameters
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -36,6 +36,26 @@ class ComponentGraph:
         self._i = 0
 
     @classmethod
+    def linearized_component_graph(cls, components):
+        """Return a list of (component name, component class) tuples from a pre-initialized component graph defined
+        as either a list or a dictionary. The component names are guaranteed to be unique."""
+        names = []
+        if isinstance(components, list):
+            seen = set()
+            for idx, component in enumerate(components):
+                component_class = handle_component_class(component)
+                component_name = component_class.name
+
+                if component_name in seen:
+                    component_name = f'{component_name}_{idx}'
+                seen.add(component_name)
+                names.append((component_name, component_class))
+        else:
+            for k, v in components.items():
+                names.append((k, handle_component_class(v[0])))
+        return names
+
+    @classmethod
     def from_list(cls, component_list, random_seed=0):
         """Constructs a linear ComponentGraph from a given list, where each component in the list feeds its X transformed output to the next component
 
@@ -45,12 +65,7 @@ class ComponentGraph:
         """
         component_dict = {}
         previous_component = None
-        for idx, component in enumerate(component_list):
-            component_class = handle_component_class(component)
-            component_name = component_class.name
-
-            if component_name in component_dict.keys():
-                component_name = f'{component_name}_{idx}'
+        for component_name, component_class in cls.linearized_component_graph(component_list):
 
             component_dict[component_name] = [component_class]
             if previous_component is not None:
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -38,7 +38,16 @@ class ComponentGraph:
     @classmethod
     def linearized_component_graph(cls, components):
         """Return a list of (component name, component class) tuples from a pre-initialized component graph defined
-        as either a list or a dictionary. The component names are guaranteed to be unique."""
+        as either a list or a dictionary. The component names are guaranteed to be unique.
+
+        Args:
+            components (list(ComponentBase) or Dict[str, ComponentBase]): Components in the pipeline.
+
+        Returns:
+            list((component name, ComponentBase)) - tuples with the unique component name as the first element and the
+                component class as the second element. When the input is a list, the components will be returned in
+                the order they appear in the input.
+        """
         names = []
         if isinstance(components, list):
             seen = set()
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -53,7 +53,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         """Returns list or dictionary of components representing pipeline graph structure
 
         Returns:
-            list(str / ComponentBase subclass): List of ComponentBase subclasses or strings denotes graph structure of this pipeline
+            list(str / ComponentBase subclass): List of ComponentBase subclasses or strings denotes graph structure of this pipeline.
         """
 
     custom_hyperparameters = None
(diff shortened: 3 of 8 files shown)


evalml/automl/automl_search.py
```diff
@@ -289,6 +289,10 @@ def __init__(self,
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
             self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
+        else:
```
freddyaboulton, line 292: This is to fix #2132. Without this change the hyperparameters passed in to ``pipeline_parameters`` would only get used in the first batch.

evalml/pipelines/pipeline_base.py
```diff
@@ -313,20 +310,11 @@ def hyperparameters(cls):
...
-                    component_hyperparameters.update(cls.custom_hyperparameters.get(component_class.name, {}))
-                hyperparameter_ranges[component_class.name] = component_hyperparameters
-        else:
-            for component_name, component_info in component_graph.items():
-                component_class = handle_component_class(component_info[0])
-                component_hyperparameters = copy.copy(component_class.hyperparameter_ranges)
-                if cls.custom_hyperparameters and component_name in cls.custom_hyperparameters:
-                    component_hyperparameters.update(cls.custom_hyperparameters.get(component_name, {}))
-                hyperparameter_ranges[component_name] = component_hyperparameters
+        for component_name, component_class in ComponentGraph.linearized_component_graph(component_graph):
```
freddyaboulton, line 313: This is to fix #2125.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2205,6 +2205,87 @@ def test_automl_pipeline_params_multiple(mock_score, mock_fit, X_y_regression):
...
+        component_graph = {"Imputer": ["Imputer"],
+                           "Imputer_1": ["Imputer", "Imputer"],
+                           "Random Forest Classifier": ["Random Forest Classifier", "Imputer_1"]}
+
+    class PipeLineLinear(BinaryClassificationPipeline):
+        component_graph = ["Imputer", "Imputer", "Random Forest Classifier"]
+
+    automl = AutoMLSearch(X, y, problem_type="binary", allowed_pipelines=[PipelineDict, PipeLineLinear],
+                          pipeline_parameters={"Imputer": {"numeric_impute_strategy": Categorical(["most_frequent"])},
+                                               "Imputer_1": {"numeric_impute_strategy": Categorical(["median"])}},
```
freddyaboulton, line 2224: Need to use `Categorical` because of #2130.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2205,6 +2205,87 @@ def test_automl_pipeline_params_multiple(mock_score, mock_fit, X_y_regression):
...
+
+
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+@patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.02})
+def test_automl_respects_pipeline_custom_hyperparameters_with_duplicate_components(mock_score, mock_fit, X_y_binary):
+
+    X, y = X_y_binary
+
+    class PipelineDict(BinaryClassificationPipeline):
+        custom_hyperparameters = {"Imputer": {"numeric_impute_strategy": Categorical(["most_frequent", 'mean'])},
```
freddyaboulton, line 2260: Need to specify `mean` for the imputers and `50` for the rf classifier because of #2131

evalml/automl/automl_search.py
```diff
@@ -260,6 +262,10 @@ def __init__(self,
         except ImportError:
             logger.warning("Unable to import plotly; skipping pipeline search plotting\n")
 
+        if allowed_pipelines is not None and not isinstance(allowed_pipelines, list):
```
freddyaboulton, line 265: Adding this data validation so that custom_hyperparameters can be overloaded with `pipeline_parameters` without weird bugs.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2701,3 +2780,17 @@ def test_automl_issues_beta_warning_for_time_series(problem_type, X_y_binary):
...
+
+
+def test_automl_validates_data_passed_in_to_allowed_pipelines(X_y_binary, dummy_binary_pipeline_class):
+
+    X, y = X_y_binary
+
+    with pytest.raises(ValueError, match="Parameter allowed_pipelines must be either None or a list!"):
+        AutoMLSearch(X, y, problem_type="binary", allowed_pipelines=dummy_binary_pipeline_class)
+
+    with pytest.raises(ValueError, match="Every element of allowed_pipelines must be a subclass of PipelineBase!"):
```
bchen1116, line 2792: This seems like it could be confusing for the user since the pipeline is technically a subclass of `PipelineBase`. I think it'd be much more helpful to output an error stating the class must not be instantiated, since that's the actual issue here

bchen1116 (approved):
Looks solid! 3 issues in a PR, you're going off.

Left a comment on potentially changing the warning for pipelines that are passed in that are instantiated, but otherwise, test coverage looks good and code changes look clean!

evalml/automl/automl_search.py
```diff
@@ -289,6 +295,10 @@ def __init__(self,
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
             self.allowed_pipelines = [make_pipeline(self.X_train, self.y_train, estimator, self.problem_type, custom_hyperparameters=self.pipeline_parameters) for estimator in allowed_estimators]
+        else:
+            for pipeline_class in self.allowed_pipelines:
+                if self.pipeline_parameters:
+                    pipeline_class.custom_hyperparameters = self.pipeline_parameters
```
angela97lin, line 301: Interesting. Does this mean that if the pipeline classes passed in already had custom_hyperparameters set, we're choosing to ignore them? I guess this touches on the two-ways-to-specify-hyperparameter-ranges we talked about in #2091 😅

evalml/pipelines/pipeline_base.py
angela97lin, line 313: JW, why not call for component_name, component_class in PipelineBase.linearized_component_graph?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2205,6 +2203,87 @@ def test_automl_pipeline_params_multiple(mock_score, mock_fit, X_y_regression):
...
 
+@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
+@patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.02})
+def test_automl_respects_pipeline_parameters_with_duplicate_components(mock_score, mock_fit, X_y_binary):
+
+    X, y = X_y_binary
+
+    class PipelineDict(BinaryClassificationPipeline):
+        component_graph = {"Imputer": ["Imputer"],
+                           "Imputer_1": ["Imputer", "Imputer"],
```
angela97lin, line 2214: Sorry if my brain's a little fried, but what is this doing here? We're passing the imputer's return val to Imputer_1?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2205,6 +2203,87 @@ def test_automl_pipeline_params_multiple(mock_score, mock_fit, X_y_regression):
...
+                           "Random Forest Classifier": ["Random Forest Classifier", "Imputer_1"]}
+
+    class PipeLineLinear(BinaryClassificationPipeline):
```
angela97lin, line 2217: I'm glad this PR fixes this case, though I have to say its a bit weird for the user to specify `Imputer_1` in `pipeline_parameters`; I guess this is dependent on them knowing how the naming would work for duplicate cases. Looks fine though!

evalml/tests/automl_tests/test_automl.py
angela97lin: Counterpoint: Ut just happened to be that @freddyaboulton  tested using an instantiated object but this error would be raised if we just threw in something random ex: AutoMLSearch class so I think this is fine

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -964,6 +964,26 @@ class MockPipelineNone(BinaryClassificationPipeline):
...
+        component_graph = ["One Hot Encoder", "One Hot Encoder", "Random Forest Classifier"]
+
+    assert PipeLineLinear.hyperparameters == {'One Hot Encoder': {},
+                                              "One Hot Encoder_1": {},
+                                              'Random Forest Classifier': {'n_estimators': Integer(10, 1000),
+                                                                           'max_depth': Integer(1, 10)}}
+
+    class PipeLineLinear(BinaryClassificationPipeline):
+        custom_hyperparameters = {"One Hot Encoder_1": {"top_n": Integer(10, 50)}}
+        component_graph = ["One Hot Encoder", "One Hot Encoder", "Random Forest Classifier"]
```
angela97lin, line 979: Same comment above about weirdness w/ user knowing to specify "One Hot Encoder_1"... maybe the right thing to do is just to have some documentation about the naming convention for duplicate components?

angela97lin (approved):
Looks good, can you always resolve 3 issues in one PR? LOL

I had minor comments about the UX of creating duplicate components and knowing how to specify the hyperparameters of each--could just be some documentation somewhere :d

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Hehe yes. I decided to keep the names between the dict and list component graph the same so that we don't have to check for different names but I recognize it's confusing. I'll add a comment.

evalml/tests/automl_tests/test_automl.py
freddyaboulton: I agree it's weird. One thing I thought about doing was be able to specify custom component names for the list api by using tuples. Something like

```python
class MyPipeline(BinaryClassificationPipeline):
     component_graph = [("My Imputer", "Imputer"), ("Cool Classifier", "Logistic Regression Classifier")]
```

but I thought that too big a change for what the original issue asked for. Plus, not sure how long the list api will be around 😅  I will file an issue so we can decide if this is worth doing and prioritize if so.

freddyaboulton: FYI #2146

evalml/pipelines/pipeline_base.py
freddyaboulton: Just cause I saw the existing implementation used a copy of the component graph lol. I don't think that's necessary though? In that case, I can just use the pipeline class property 👍

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2701,3 +2780,17 @@ def test_automl_issues_beta_warning_for_time_series(problem_type, X_y_binary):
         assert len(warn) == 1
         message = "Time series support in evalml is still in beta, which means we are still actively building its core features"
         assert str(warn[0].message).startswith(message)
+
+
+def test_automl_validates_data_passed_in_to_allowed_pipelines(X_y_binary, dummy_binary_pipeline_class):
```
chukarsten, line 2785: I love your tests and I love that your exceptions are screaming at the user.  Absolutely love that.  If it were me, all the exceptions would be in caps ala 1994 AOL messages from your parents.

evalml/tests/pipeline_tests/test_pipelines.py
chukarsten: Yea, I anticipate that's going to be a big gotcha going forward...

evalml/tests/automl_tests/test_automl.py
chukarsten: I put #2146 into the next sprint, per the Elder Council's approval.

chukarsten (changes requested):
I think I just have a docstring change request!  Other than that, way to be super efficient!  Good work!

evalml/automl/automl_search.py
freddyaboulton: That's a good point! I thought this is what the intended behavior should be. My reasoning was that users should use one api over the other. I thought it'd weird for a user to go through the trouble of defining pipeline classes and specifying custom hyperparameters and then decide to specify additional hyperparameters through `AutoMLSearch`. 

I filed #2147 to continue the discussion but in the mean time I think the least aggressive thing to do is to just update the custom hyperparameters with the values in `self.pipeline_parameters` !

evalml/pipelines/component_graph.py
```diff
@@ -35,6 +35,26 @@ def __init__(self, component_dict=None, random_seed=0):
         self._feature_provenance = {}
         self._i = 0
 
+    @classmethod
+    def linearized_component_graph(cls, components):
+        """Return a list of (component name, component class) tuples from a pre-initialized component graph defined
+        as either a list or a dictionary. The component names are guaranteed to be unique."""
```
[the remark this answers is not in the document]
freddyaboulton: You got it!

evalml/pipelines/component_graph.py
```diff
@@ -35,6 +35,26 @@ def __init__(self, component_dict=None, random_seed=0):
...
+        if isinstance(components, list):
+            seen = set()
+            for idx, component in enumerate(components):
+                component_class = handle_component_class(component)
+                component_name = component_class.name
+
+                if component_name in seen:
+                    component_name = f'{component_name}_{idx}'
+                seen.add(component_name)
+                names.append((component_name, component_class))
```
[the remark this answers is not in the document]
freddyaboulton: If you're asking why I'm adding to the set as I go its because I need to keep track of which components have been duplicated. This logic used to exist in `from_list` method of `ComponentGraph` but now I moved it here so that it could be used by different pipeline methods and the iterative algorithm more easily.

Let me know if I'm misunderstanding!

chukarsten (approved):
image

PR #2028, opened 2021-03-24
Remove lists as acceptable hyperparameter ranges in ``AutoMLSearch``

angela97lin:
Closes #2015 

To support lists as parameters for our components, we are removing lists as acceptable hyperparameter ranges in the `pipeline_parameters` parameter to `AutoMLSearch`. This is okay, because users can still use `skopt.space.Categorical` to achieve the same functionality.

PR #2138, opened 2021-04-14
Handle index columns in AutoMLSearch and DataChecks

### LINKED ISSUE #1862: Automatically drop features tagged as "primary key"
There are plans to add a "primary key" semantic tag to woodwork, to indicate a column which can be used for joining multiple tables, in featuretools or elsewhere. Once that is added, we should update evalml to add a DropColumn component to remove column(s) tagged with "primary key".

jeremyliweishih:
Fixes #1862.


chukarsten (approved):
Looks good to me - feel free to take the nit or leave it!

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -352,3 +354,25 @@ def validate(self, X, y):
...
+    X = pd.DataFrame(X)
+    X['index_col'] = pd.Series(range(len(X)))
+    X = ww.DataTable(X)
+    X = X.set_index('index_col')
+
+    class MockDataCheck(DataCheck):
+        def validate(self, X, y):
+            return {"warnings": [], "errors": [], "actions": []}
+
+    assert MockDataCheck().validate(X, y)
```
bchen1116, line 370: is this assert necessary? Can we just assert that it equals the dict of empty warnings/errors/actions?

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -352,3 +354,25 @@ def validate(self, X, y):
...
+
+    MockDataCheck.validate = MagicMock()
+    checks = DataChecks([MockDataCheck, MockDataCheck, MockDataCheck])
+    checks.validate(X, y)
+
+    validate_args = MockDataCheck.validate.call_args_list
+    for arg in validate_args:
+        assert 'index_col' not in arg[0][0].columns
```
bchen1116, line 378: nit, but can we assert that col `0` is in the arg cols just to check that we are checking what we expect to be X?

bchen1116 (approved):
Looks good! I left a few comments on testing just to be nit-picky heh

evalml/tests/data_checks_tests/test_data_checks.py
jeremyliweishih: I needed it for codecov to pass or else it would fail on `return {"warnings": [], "errors": [], "actions": []}` 😄

freddyaboulton (dismissed):

evalml/automl/automl_search.py
```diff
@@ -288,6 +288,15 @@ def __init__(self,
             logger.info("Generating pipelines to search over...")
             allowed_estimators = get_estimators(self.problem_type, self.allowed_model_families)
             logger.debug(f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}")
+
+            index_columns = list(self.X_train.select('index').columns)
```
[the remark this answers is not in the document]
jeremyliweishih: In the current impl the index column will be passed down to the estimator. My thinking was that if a user specifies their own pipelines via `allowed_pipelines` they don't want AutoML to come up with preprocessing components/pipelines and will use their own preprocessing components in their own pipelines. This should be consistent (from my understanding) with behavior like if a user has null values but don't specify an imputer in `alloed_pipelines`.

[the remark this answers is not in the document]
freddyaboulton: I think that makes sense! I agree its consistent with our previous design choices.  My one concern is that if the index column won't be picked up by the data checks, users may not know that there's anything "wrong" with not explicitly dropping it before `AutoMLSearch`. But maybe that's too small a set of users to worry about. If someone explicitly sets a woodwork index, they probably know that means.

jeremyliweishih:
@freddyaboulton @angela97lin @dsherry @chukarsten 

I made the following changes (an easy diff view [here](https://github.com/alteryx/evalml/pull/2138/files/f402749a2ac3301fe865c2589ab660f5492b80f8..818451bb8a5fcc230da0138e15de86cf4d5a8d97)):

- Added `custom_hyperparameters` parameter to `AutoMLSearch` and `IterativeAlgorithm` - this behaves mostly like how `pipeline_parameters` behaved previously
- Changed `pipeline_parameters` parameter to only initialize pipelines with a singular value 

The fundamental issue I wanted to address with these changes is that user defined pipeline/component initialization values and user defined hyperparameter were bundled together in `pipeline_parameters`. In the above changes I changed `pipeline_parameters` to represent values we want pipelines and components to be initialized with and `custom_hyperparameters` to represent user defined ranges. However, this change doesn't support `pipeline_parameters` that fall out of a components hyperparameter ranges (which may be why these two features were bundled into one previously). What do you guys think of this change? I wanted to make sure we're on board with this before ironing out the details (for example: I think `if component_class.name in self._pipeline_params and self._batch_number == 0:` needs to be changed to just `if component_class.name in self._pipeline_params` so that we use those values for every batch.).

freddyaboulton:
@jeremyliweishih Nice! One concern I have with adding the `pipeline_parameters` is that there are two ways to do the same thing. Users can do `pipeline_parameters={"Logistic Regression Classifier": {"C": 0.5}}` and `custom_hyperparameters={"Logistic Regression Classifier": {"C": Categorical([0.5])}` (assuming we fix the bug with freezing hyperparameters  to not the default value lol, e.g. #2131). 

 I'm not sure if that's good for our tuner? Doing `pipeline_parameters={"Logistic Regression Classifier": {"C": 0.5}}` means that the tuner will propose values for `C` (because the hyperparameter ranges have not been updated) but we'll set the value to `0.5` without the tuner knowing. I think the tuner should know the space has been restricted?

I'm wondering if it is better to unblock this PR without introducing an argument to `AutoMLSearch`? I think we can do this by keeping the `custom_hyperparameters` and `pipeline_params` argument to `IterativeAlgorithm` and setting the arguments to the `Drop Columns Transformer` in `pipeline_params` before creating the `IterativeAlgorithm`. We do something similar with the `problem_configuration` now.

I guess, in my mind, `pipeline_parameters` is just a special case of `custom_hyperparameters` (in the former, the search space only has only element). Rather than exposing that distinction to users, I think it's better to handle it internally. 

I'm hopeful we'll be able to properly treat the concept of `pipeline_parameters` as a search space with one element in the future. But until we get there, I think adding `pipeline_parameters` to `IterativeAlgorithm` will suffice (and unblock @angela97lin 's drop columns data check actions work. I'm fairly certain we would have run into this same problem there!).

jeremyliweishih:
@freddyaboulton I removed the additional changes and put this PR back in scope of just adding the ability to handle index columns. However, I also removed the handling of additional parameters passed in to `pipeline_parameters` for `Drop Columns Transformer` due to #2028 disallowing lists as input for `pipeline_parameters` and I believe we can add this functioanlity when we come up with a solution for #2130 and #2131.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2824,6 +2824,30 @@ def test_automl_issues_beta_warning_for_time_series(problem_type, X_y_binary):
         assert str(warn[0].message).startswith(message)
 
 
+@patch('evalml.pipelines.BinaryClassificationPipeline.score', return_value={"Log Loss Binary": 0.3})
+@patch('evalml.automl.engine.sequential_engine.train_pipeline')
+def test_automl_drop_index_columns(mock_train, mock_binary_score, X_y_binary):
```
freddyaboulton, line 2829: Nice test!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -145,6 +145,8 @@ def _transform_parameters(self, pipeline_class, proposed_parameters):
                         component_parameters[param_name] = value.rvs(random_state=self.random_seed)
                     else:
                         component_parameters[param_name] = value
+            if name in self._pipeline_params and name == 'Drop Columns Transformer' and self._batch_number > 0:
```
freddyaboulton, line 148: Nice. I think we'll hopefully get rid of this line soon in #2130 but I agree this is the way to get it done for now!

freddyaboulton (approved):
@jeremyliweishih Looks great!

PR #2148, opened 2021-04-16
Build_conda_pkg: fix shell syntax for dockerhub login

dsherry:
Building off #1870 

I updated our dockerhub password yesterday. The updated password happened to contain special characters. The current syntax used to pass the password into the docker login command breaks if the password contains special characters!

I was able to confirm the above by changing the password to not contain special chars, and then observing that the job ran successfully. However, I get different behavior from the command in question locally. To avoid future problems, and to make this work when we next change our dockerhub password, I'd like to change the syntax of the command. The double-curly braces are a github actions syntax.

On bash locally:
```bash
bash-5.1$ ENV_VAR=abcd1234
bash-5.1$ echo ${{ENV_VAR}}
bash: ${{ENV_VAR}}: bad substitution
bash-5.1$ echo $ENV_VAR
abcd1234
bash-5.1$ echo ${ENV_VAR}
LYuApz5H7c27C2
bash-5.1$ echo "${ENV_VAR}"
LYuApz5H7c27C2
```

I found [this](https://unix.stackexchange.com/questions/4899/var-vs-var-and-to-quote-or-not-to-quote) to be a useful reference.

Update: to prove this change works, I changed config to run the job in my PR, waited until [the action ran green](https://github.com/alteryx/evalml/runs/2365029639?check_suite_focus=true), then reverted the change.

### DIFF
--- a/.github/workflows/build_conda_pkg.yml
+++ b/.github/workflows/build_conda_pkg.yml
@@ -45,7 +45,7 @@ jobs:
           python .github/conda_config.py "$(python setup.py --version)"
           cd evalml-core-feedstock
           echo "Pre docker login"
-          echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login -u ${{ secrets.DOCKERHUB_USER }} --password-stdin
+          echo "${secrets.DOCKERHUB_PASSWORD}" | docker login -u "${secrets.DOCKERHUB_USER}" --password-stdin
           echo "Post docker login"
           export DOCKER_CONTAINERID="$(docker run -td condaforge/linux-anvil-cos7-x86_64)"
           echo "Created container ${DOCKER_CONTAINERID}"
--- a/.github/workflows/build_conda_pkg.yml
+++ b/.github/workflows/build_conda_pkg.yml
@@ -35,6 +35,9 @@ jobs:
           source test_python/bin/activate
           make installdeps-test
       - name: Clone Feedstock, Copy to Container, and Run Update
+        env:
+          DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
+          DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_conda_pkg.yml
```diff
@@ -35,6 +35,9 @@ jobs:
           source test_python/bin/activate
           make installdeps-test
       - name: Clone Feedstock, Copy to Container, and Run Update
+        env:
+          DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
+          DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
```
dsherry, line 40: I don't think this is necessary per se, but I think it helps disambiguate the fact that the double-curlies are for the GH secret, and then the "${}" below are for the shell scripting 

https://docs.github.com/en/actions/reference/encrypted-secrets#example-using-bash

.github/workflows/build_conda_pkg.yml
```diff
@@ -45,16 +48,16 @@ jobs:
           python .github/conda_config.py "$(python setup.py --version)"
           cd evalml-core-feedstock
           echo "Pre docker login"
-          echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login -u ${{ secrets.DOCKERHUB_USER }} --password-stdin
+          echo "${DOCKERHUB_PASSWORD}" | docker login -u "${DOCKERHUB_USER}" --password-stdin
```
dsherry, line 51: The ${} is shell syntax for referencing a variable
The quotes are shell scripting best-practice, just in case our username or password contain special chars (which is what was happening this morning!)

bchen1116 (approved):
Changes make sense to me!

freddyaboulton (approved):
Thanks so much @dsherry !!

jeremyliweishih (approved):
looks good when tests pass!

PR #2152, opened 2021-04-16
Add top-level search method

dsherry:
First draft of a top-level search method. This method is our recommendation for how to use woodwork, data checks and automl search together. It defines an interface which sets up the woodwork schema (if the input didn't already set woodwork types), runs our default data checks and then if there are no errors runs automl search.

Once data check actions have been added, we can add support for that here as well.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -87,7 +87,7 @@ def search(X_train=None, y_train=None, problem_type=None, objective='auto', **kw
     Other keyword arguments which are provided will be passed to AutoMLSearch.
 
     Returns:
-        (AutoMLSearch, dict): the automl search object containing pipelines and rankings, and the results from running the data checks. If the data check results contain errors, the automl
+        (AutoMLSearch, dict): the automl search object containing pipelines and rankings, and the results from running the data checks. If the data check results contain errors, automl search will not be run and an automl search object will not be returned.
     """
     X_train = infer_feature_types(X_train)
     y_train = infer_feature_types(y_train)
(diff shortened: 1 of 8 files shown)


evalml/automl/automl_search.py
```diff
@@ -62,6 +63,53 @@
...
+            When set to 'auto', chooses:
+
+            - LogLossBinary for binary classification problems,
+            - LogLossMulticlass for multiclass classification problems, and
+            - R2 for regression problems.
+
+    Other keyword arguments which are provided will be passed to AutoMLSearch.
+
+    Returns:
+        (AutoMLSearch, dict): the automl search object containing pipelines and rankings, and the results from running the data checks. If the data check results contain errors, the automl
```
jeremyliweishih, line 90: "the automl" ----> "the automl search object will not be returned"

evalml/automl/automl_search.py
```diff
@@ -62,6 +63,53 @@
...
+    """
+    X_train = infer_feature_types(X_train)
+    y_train = infer_feature_types(y_train)
+    problem_type = handle_problem_types(problem_type)
+    if objective == 'auto':
+        objective = get_default_primary_search_objective(problem_type)
+    objective = get_objective(objective, return_instance=False)
+
+    automl_config = kwargs
```
jeremyliweishih, line 99: Maybe we can add more in the docstring about additional kwargs for the automl config? I don't think its very clear we support additional kwargs.

jeremyliweishih (approved):
Looks great! Just had some comments on docstring updates.

evalml/automl/automl_search.py
dsherry: Good point will do

Issue #1970, opened 2021-03-11
Prediction Explanation Force Plots


freddyaboulton:
Currently, we display our prediction explanations in a table. 

Some users may prefer a graphical output, like the [force plot](https://shap.readthedocs.io/en/latest/example_notebooks/tabular_examples/tree_based_models/Force%20Plot%20Colors.html#Force-Plot-Colors)

This issue tracks adding functionality to display prediction explanations graphically.

dsherry:
Yes!

https://shap.readthedocs.io/en/latest/example_notebooks/tabular_examples/tree_based_models/Force%20Plot%20Colors.html

dsherry:
Let's build both
a) a data method, which returns a JSON-serializable python dict containing all the info one would need to show this stuff in the UI
b) a visualization method using shap's default behavior. Ideally this uses the data method, but if they have to be separate, 🤷‍♂️

PR #2157, opened 2021-04-20
Prediction Explanations: Force Plots

chukarsten:
Addresses #1970 

Single row explanations force plots:

Screen Shot 2021-04-21 at 1 19 18 PM


evalml/model_understanding/graphs.py
```diff
@@ -1169,3 +1173,74 @@ def graph_t_sne(X, n_components=2, perplexity=30.0, learning_rate=200.0, metric=
     fig.update_layout(title='t-SNE', yaxis_zeroline=False, xaxis_zeroline=False)
 
     return fig
+
+
+def force_plot(pipeline, rows_to_explain, training_data, matplotlib=False):
```
freddyaboulton, line 1178: The issue mentions adding a "data method" to return the data needed to create the plot in json plot. Is that planned for a separate PR?

chukarsten: I need to chat with @dsherry about what exactly that entails...I think I'll need some specifics but I planned to do that here.

dsherry: Our model understanding tools should be broken into two chunks:
* Data generation: given necessary inputs, return all the information required to visualize the results
* Visualization: call the data generation method, and use pyplot or similar to display the results

Separating these two steps is important because it means that the data generation method could be used to display the same information in a different UI.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -166,86 +169,100 @@ def make_drill_down_dict(provenance, shap_values, normalized_values, pipeline_fe
...
+                              pipeline_features=pipeline_features, original_features=original_features,
+                              expected_value=expected_value)['explanations']
 
         # Not including the drill down dict for dataframes
         # 'drill_down' is always included in the dict output so we can delete it
         for d in data:
             del d['drill_down']
 
+        if not isinstance(data, list):
+            data = [data]
```
chukarsten, line 193: I came across an instance of the data being passed through as a list, like it seems to expect.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -166,86 +169,100 @@ def make_drill_down_dict(provenance, shap_values, normalized_values, pipeline_fe
...
+        for d in data:
+            if "expected_value" in d.keys():
+                del(d["expected_value"])
```
chukarsten, line 196: In the similar style as drill down.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -12,7 +12,7 @@
 
 
 def _make_rows(shap_values, normalized_values, pipeline_features, original_features, top_k, include_shap_values=False,
```
freddyaboulton, line 14: `include_expected_values` is not used. I think we can get rid of it?

evalml/model_understanding/prediction_explanations/explainers.py
```diff
@@ -63,12 +64,13 @@ def explain_predictions(pipeline, input_features, y, indices_to_explain, top_k_f
...
                                              output_format=output_format, top_k_features=top_k_features,
-                                             include_shap_values=include_shap_values)
+                                             include_shap_values=include_shap_values,
+                                             include_expected_value=include_expected_value)
     return report_creator(data)
 
 
 def explain_predictions_best_worst(pipeline, input_features, y_true, num_to_explain=5, top_k_features=3,
-                                   include_shap_values=False, metric=None, output_format="text"):
+                                   include_shap_values=False, include_expected_value=False, metric=None, output_format="text"):
```
freddyaboulton, line 73: I think there is a bug here because the expected value is included regardless of the value of this parameter

```python
exp = explain_predictions(pipeline=pipeline, input_features=X, y=None, indices_to_explain=[3],
                           top_k_features=6, include_shap_values=True, output_format="dict",
                   include_expected_value=False)
for explanation in exp["explanations"]:
    for class_explanation in explanation["explanations"]:
        assert class_explanation["expected_value"] != 0
```

I think in this case it's fine to not surface this parameter to the top-level and instead always include the expected value in the dict output. The expected value really only makes sense for the dict output format so I worry that including this parameter will cause users who use the dataframe or text output to be confused.

freddyaboulton (approved):
@chukarsten Thank you! I think it's great we can use `explain_predictions` as our data function.

I left a couple of questions/suggestions I'd like to resolve before merging.

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,100 @@
...
+        y (pandas.Series): The target data.
+
+    Returns:
+        list(dict()): list of dictionaries where each dict contains force plot data.  Each dictionary
+            entry represents the explanations for a single row.
+            e.x. For single row binary force plots:
+                    [{'malignant': {'expected_value': 0.37,
+                                    'feature_names': ['worst concave points', 'worst perimeter', 'worst radius'],
+                                    'shap_values': [0.09, 0.09, 0.08],
+                                    'plot': AdditiveForceVisualizer}]
```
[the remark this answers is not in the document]
chukarsten: Yup!

docs/source/release_notes.rst
```diff
@@ -80,6 +80,7 @@ Release Notes
         * Added a warning about how time series support is still in beta when a user passes in a time series problem to ``AutoMLSearch`` :pr:`2118`
         * Added ``NaturalLanguageNaNDataCheck`` data check :pr:`2122`
         * Added ValueError to ``partial_dependence`` to prevent users from computing partial dependence on columns with all NaNs :pr:`2120`
+        * Adds force plots :pr:`2157`
```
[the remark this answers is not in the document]
dsherry: Please also mention "prediction explanations"
> Added force plots for prediction explanations

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -565,6 +565,84 @@
...
+    "\n",
+    "from evalml.demos import load_breast_cancer\n",
+    "from evalml.model_understanding.force_plots import graph_force_plot\n",
+    "from evalml.pipelines import BinaryClassificationPipeline\n",
+    "\n",
+    "X, y = load_breast_cancer()\n",
+    "\n",
+    "\n",
+    "pipeline = BinaryClassificationPipeline(component_graph = ['Simple Imputer', 'Random Forest Classifier'])\n",
+    "pipeline.fit(X, y)\n",
```
dsherry, line 593: You don't need to fit a pipeline here. We fit one at the top of the model understanding guide. Please delete.

https://evalml.alteryx.com/en/v0.24.1/user_guide/model_understanding.html

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -565,6 +565,84 @@
...
+    "\n",
+    "shap.initjs() # Allows plots to be visualized in Jupyter notebook\n",
```
dsherry, line 595: Is there a reason why our code shouldn't automate this?

We have a method called `jupyter_check` which should tell you whether or not code is being run in an ipython environment. If true, could `graph_force_plot` call `shap.initjs()` internally?

Relatedly, so far we've only used `jupyter_check` to conditionally import `ipywidgets`, which I believe we do for similar reasons to this call to `shap.initjs()`... would doing that in `graph_force_plot` make this simpler?

Here's an example: https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/graphs.py#L909

Also, I took a look at [what `initjs` is doing](https://github.com/slundberg/shap/blob/master/shap/plots/_force.py#L244). It looks like it checks that ipython is set up, imports the shap logo and displays the logo in an HTML div alongside a javascript bundle.

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -565,6 +565,84 @@
...
+   "outputs": [],
+   "source": [
+    "from evalml.demos import load_wine\n",
+    "from evalml.model_understanding.force_plots import graph_force_plot\n",
+    "from evalml.pipelines import MulticlassClassificationPipeline\n",
+    "\n",
+    "X, y = load_wine()\n",
+    "\n",
+    "pipeline = MulticlassClassificationPipeline(component_graph = ['Simple Imputer', 'Random Forest Classifier'])\n",
+    "pipeline.fit(X, y)\n",
```
dsherry, line 626: Style nit-pick: if its important to show a multiclass example for force plots, I suggest you:
* Rename the variable name of the binary classification pipeline which all the rest of the model understanding tools use to `pipeline_binary`
* At the top of the notebook where we define the binary pipeline, move this code up there so we also define `pipeline_multiclass`
* Then below you can use `pipeline_multiclass`

The reason I like this is that it keeps the notebook's narrative super simple. First train models, then visualize.

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+import shap
+
+from evalml.model_understanding.prediction_explanations import (
+    explain_predictions
+)
+
+
+def graph_force_plot(pipeline, rows_to_explain, training_data, y, matplotlib=False):
```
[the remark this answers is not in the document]
dsherry: +1, we should follow the convention our other graph methods follow unless there's a reason to do otherwise.

I think returning a plot is super useful for jupyter notebooks actually -- they'll auto-render the last returned value.

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,100 @@
...
+        list(dict(shap.AdditiveForceVisualizer)): The same as force_plot(), but with an additional
+            key in each dictionary for the plot.
+    """
+
+    def gen_force_plot(shap_values, training_data, expected_value, matplotlib):
+        """ Helper function to generate a single force plot. """
+        # Ensure there are as many features as shap values.
+        assert training_data.shape[1] == len(shap_values)
+
+        # TODO: Update this to make the force plot display multi-row array visualizer.
```
dsherry, line 31: Do we have an issue filed for this TODO?

evalml/model_understanding/force_plots.py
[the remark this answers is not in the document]
dsherry: Ah, I read a bit further. I see this is because both `graph_force_plot` and `force_plot` accept a list of rows to explain rather than a single row.

I suggest we define `force_plot` and `graph_force_plot` to accept a single row. Then define `force_plots` and `graph_force_plots` to accept multiple and to call the single-row methods.

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,100 @@
...
+                                    'plot': AdditiveForceVisualizer}]
+                 For two row binary force plots:
+                    [{'malignant': {'expected_value': 0.37,
+                                    'feature_names': ['worst concave points', 'worst perimeter', 'worst radius'],
+                                    'shap_values': [0.09, 0.09, 0.08],
+                                    'plot': AdditiveForceVisualizer},
+                    {'malignant': {'expected_value': 0.29,
+                                    'feature_names': ['worst concave points', 'worst perimeter', 'worst radius'],
+                                    'shap_values': [0.05, 0.03, 0.02],
+                                    'plot': AdditiveForceVisualizer}]
```
dsherry, line 72: Is the value of this `'plot'` key a class or a string? I ask because if its a class, we can't easily JSON-serialize this output, so we'd likely have to convert to string somewhere in order for a web UI to consume this.

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,100 @@
...
+    explanations = []
+    prediction_explanations = explain_predictions(pipeline, training_data, y, rows_to_explain,
+                                                  top_k_features=len(training_data.columns), include_shap_values=True,
+                                                  output_format="dict")
+    row_explanations = prediction_explanations["explanations"]
+    for row_explanation in row_explanations:
+        row_exp = row_explanation["explanations"]
+        row_exp_dict = {}
+        for cls_exp in row_exp:
+            cls = cls_exp["class_name"] if cls_exp["class_name"] is not None else "regression"
```
dsherry, line 93: Simplification: use the [dict `get` method](https://docs.python.org/3/library/stdtypes.html#dict.get)
```python
cls = cls_exp.get("class_name", "regression")
```

dsherry, line 93: So, according to this logic, if the `"class_name"` comes back as `None` here, and we set `cls` to `"regression"`, doesn't that mean the assignment to `row_exp_dict[cls]` will be overwriting the same key for every loop in this `for` loop?

evalml/model_understanding/prediction_explanations/_algorithms.py
```diff
@@ -97,10 +99,12 @@ def _compute_shap_values(pipeline, features, training_data=None):
         mappings = []
         for class_shap_values in shap_values:
             mappings.append(_create_dictionary(class_shap_values, feature_names))
-        return mappings
+        return (mappings, explainer) if return_explainer else mappings
     # regression problem
     elif isinstance(shap_values, np.ndarray):
-        return _create_dictionary(shap_values, feature_names)
+        dic = _create_dictionary(shap_values, feature_names)
+        return (dic, explainer) if return_explainer else dic
```
dsherry, line 106: Returning variable length arguments can make consuming output a bit uglier up the stack. I suggest we always return a tuple of two arguments. If `explainer` here happens to return as `None`, that's fine.

Rolling with @freddyaboulton 's suggestion above though, I wonder if we can just delete `return_explainer` and simply always return both the dictionary and the explainer here.

evalml/model_understanding/prediction_explanations/_algorithms.py
```diff
@@ -31,18 +31,20 @@ def _create_dictionary(shap_values, feature_names):
     return mapping
 
 
-def _compute_shap_values(pipeline, features, training_data=None):
+def _compute_shap_values(pipeline, features, training_data=None, return_explainer=False):
```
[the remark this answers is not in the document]
dsherry: +1

Why add this flag? If it costs us extra CPU cycles to compute the explainer to be returned, I think that's a good argument for having a flag. But at first glance through this code, I don't think that's the case.

dsherry (changes requested):
Looking solid, left a few comments

evalml/model_understanding/prediction_explanations/_algorithms.py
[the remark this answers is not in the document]
chukarsten: I don't think so, it gets called in a lot of other tests when the expected value is not desired (non-force plots).

docs/source/release_notes.rst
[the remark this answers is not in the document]
chukarsten: I'm legit mystified how I put this in the wrong section.

docs/source/user_guide/model_understanding.ipynb
chukarsten: gucci

docs/source/user_guide/model_understanding.ipynb
chukarsten: Done and done!

evalml/model_understanding/prediction_explanations/_algorithms.py
chukarsten: Done, this is now returning the expected_value instead of the explainer.  Always.

evalml/model_understanding/force_plots.py
chukarsten: @dsherry yup, we traiged this : https://github.com/alteryx/evalml/issues/2287

evalml/model_understanding/force_plots.py
chukarsten: I didn't think the Web UI was going to consume this.  My understanding was that the graphing function was going to be for the purpose of EvalML users via jupyter notebook, etc. and that the data function, which didn't add the plot to the dict was going to be the one driving the web UI (which was going to render the force plots itself to get all the nice javascript stuff that shap does)

evalml/model_understanding/force_plots.py
chukarsten: The possible values for the key "class_name" are None, the name of the positive class ("malignant" in these tests), or the names of the different classes ["class_0", "class_1", "class_2"].  the row_exp variable is a list that contains either 1 explanation for regression type problems, 1 explanation for binary class problems (the positive class) or multiple explanations for multi-class problems.  If the row for the regression type problems had more than one "class" to explain then yes, it would continuously overwrite that key, but in this case it should be good.

docs/source/user_guide/model_understanding.ipynb
chukarsten: I made the changes to the pipeline names, but some of the visualization, particularly with the binary pipelines, is spread out all over the notebook.  I renamed them, but kept them in the same spot for now.

evalml/model_understanding/force_plots.py
chukarsten: Haha you got me on this one!  "class_name" always exists, but it's the *value* of the "class_name" key that is None.  The .get() function is for if the "class_name" key doesn't exist.  You taught me something here :)

evalml/model_understanding/force_plots.py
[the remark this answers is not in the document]
chukarsten: @dsherry I think I'm going to let #2287 handle the multi row case.  shap._force gives us back a different style of visualizer for multi-row explanations than it does for single row explanations.  How I handle that now is to provide the single row visualization for each row rather than the combined multi-row explanation.  Originally, before I used explain_predictions, `graph_force_plots` properly returned the visualization for both cases, making separate graphing functions unnecessary.  I hope after #2287 is complete, it will again! 

However, I will change the return to just be the plots rather than all the other junk.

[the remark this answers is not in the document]
chukarsten: @freddyaboulton I am not entirely sure what this AdditiveForceVisualizer is.  It seems like it's a python wrapper for some javascript visualization.  Which makes me unsure if I can wrap it into a plotly figure.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -687,9 +724,12 @@ def make_dataframe(self, index, y_pred, y_true, scores, dataframe_index):
 
 
 class _SHAPTable(_SectionMaker):
-    def __init__(self, top_k_features, include_shap_values):
+    def __init__(
+        self, top_k_features, include_shap_values, include_expected_value=False
+    ):
         self.top_k_features = top_k_features
         self.include_shap_values = include_shap_values
+        self.include_expected_value = include_expected_value
```
freddyaboulton, line 732: I don't think the value of this parameter is actually used in the implementation. I vote we remove it from the function signature.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -248,12 +254,15 @@ def make_dataframe(
...
             pipeline_features=pipeline_features,
             original_features=original_features,
+            expected_value=expected_value,
         )["explanations"]
 
         # Not including the drill down dict for dataframes
         # 'drill_down' is always included in the dict output so we can delete it
         for d in data:
             del d["drill_down"]
+            if "expected_value" in d.keys():
```
freddyaboulton, line 264: nit-pick: As things are implemented now, the `expected_value` is always a key so we can get rid of the check.

evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -125,6 +125,7 @@ def _make_text_table(
     original_features,
     top_k,
     include_shap_values=False,
+    include_expected_value=False,
```
freddyaboulton, line 128: I think we can get rid of this parameter since it's not used.

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_force_plots.py
```diff
@@ -0,0 +1,203 @@
...
+        TypeError,
+        match="rows_to_explain should be provided as a list of row index integers!",
+    ):
+        force_plot(None, None, None, None)
+    with pytest.raises(
+        TypeError, match="rows_to_explain should only contain integers!"
+    ):
+        force_plot(
+            pipeline=None,
+            rows_to_explain=["this", "party's", "over"],
```
freddyaboulton, line 28: 😂

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_force_plots.py
```diff
@@ -0,0 +1,203 @@
...
+            y=None,
+        )
+
+
+@patch("evalml.model_understanding.force_plots.jupyter_check")
+@patch("evalml.model_understanding.force_plots.initjs")
+@pytest.mark.parametrize(
+    "rows_to_explain, just_data", product([[0], [0, 1, 2, 3, 4]], [False, True])
+)
+def test_force_plot_binary(
```
freddyaboulton, line 39: Do you think it's worth adding a test to make sure the force plot displays the aggregate contribution of features that have been OHE/text features? I vote yes just because I think it's important to make sure we maintain that behavior since the reports also aggregate the OHE/text features.

freddyaboulton (approved):
@chukarsten Thank you for making the changes! I think one clean-up I'd like to do is remove the `_include_expected_value` parameter since it's not used by the forceplot code and it doesn't seem to change the output of the explainers.

I left a comment about adding some tests for pipelines that use OHE/Text Featurizers so that we check the features are aggregated. Let me know what you think about this change.

I also think there may be a bug with the jupyter display? The following code raises `TypeError: Object of type Timestamp is not JSON serializable`:

```python
from evalml.model_understanding import explain_predictions
from evalml.model_understanding.force_plots import graph_force_plot
import pytest


X, y = load_fraud(1000)

pipeline = BinaryClassificationPipeline(["Text Featurization Component", "DateTime Featurization Component",
                                         "One Hot Encoder", "Random Forest Classifier"])
pipeline.fit(X, y)

results = graph_force_plot(pipeline, rows_to_explain=[1],
                           training_data=X, y=y)

for result in results:
    for cls in result:
        print("Class:", cls)
            display(result[cls]["plot"])
```

image

I think the issue is that the input features are timestamps: 

image


I say we don't block merge on that bug though cause it's coming from the `display` call. My hypothesis is that it might work if users use their own code to display the data from `force_plot`. I think we should file a spike issue to investigate further. Let me know what you think.

docs/source/release_notes.rst
```diff
@@ -37,6 +37,7 @@ Release Notes
 
 **v0.24.2 May. 24, 2021**
     * Enhancements
+        * Adds force plots for prediction explanations :pr:`2157`
```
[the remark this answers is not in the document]
chukarsten: this has lingered so long

evalml/model_understanding/prediction_explanations/_user_interface.py
chukarsten: not sure how that got past....

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_force_plots.py
chukarsten: I do....I just need to figure out what that looks like, haha.

evalml/model_understanding/force_plots.py
dsherry: Agreed. But isn't `force_plot` (aka this code) the data function and `graph_force_plot` the graphing function?

evalml/model_understanding/prediction_explanations/_algorithms.py
```diff
@@ -106,15 +107,26 @@ def _compute_shap_values(pipeline, features, training_data=None):
...
 
+    if is_multiclass(pipeline.problem_type) or is_regression(pipeline.problem_type):
+        expected_value = explainer.expected_value
+    elif is_binary(pipeline.problem_type):
+        try:
+            # Accounts for CatBoost and XGBoost returning expected_value as float
+            # for positive class
+            expected_value = explainer.expected_value[1]
+        except IndexError:
+            expected_value = explainer.expected_value
```
dsherry, line 118: Do we have test coverage for this logic?

dsherry (approved):
@chukarsten looking good to me :) left a couple questions

evalml/model_understanding/force_plots.py
```diff
@@ -0,0 +1,127 @@
...
+                    [{'malignant': {'expected_value': 0.37,
+                                    'feature_names': ['worst concave points', 'worst perimeter', 'worst radius'],
+                                    'shap_values': [0.09, 0.09, 0.08],
+                                    'plot': AdditiveForceVisualizer},
+                    {'malignant': {'expected_value': 0.29,
+                                    'feature_names': ['worst concave points', 'worst perimeter', 'worst radius'],
+                                    'shap_values': [0.05, 0.03, 0.02],
+                                    'plot': AdditiveForceVisualizer}]
+
+    Raises:
```
bchen1116, line 86: Nice! I like this

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_force_plots.py
```diff
@@ -0,0 +1,279 @@
...
+        jupyter_check.return_value = False
+        with pytest.warns(None) as graph_valid:
+            results = graph_force_plot(
+                pipeline,
+                rows_to_explain=rows_to_explain,
+                training_data=X,
+                y=y,
+                matplotlib=False,
+            )
+            assert not initjs.called
```
bchen1116, line 70: What's the difference between this and `initjs.assert_called()`?

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_force_plots.py
```diff
@@ -0,0 +1,279 @@
...
+            y=y,
+            matplotlib=False,
+        )
+
+    # Should have a result per row to explain.
+    assert len(results) == len(rows_to_explain)
+
+    for result in results:
+        class_labels = result.keys()
+        assert len(class_labels) == 1
```
bchen1116, line 196: why do we assert `len(class_labels)==1` here? Does this imply that there's only 1 class we're doing this for on regression pipelines?

bchen1116 (commented):
Looking good! Left a few questions and clean-up comments.

PR #2170, opened 2021-04-21
Have OHE return Boolean columns

### LINKED ISSUE #2169: Have OHE return Boolean columns
Change OHE to return Boolean columns rather than numeric columns for the new columns that are encoded.

bchen1116:
fix #2169

### DIFF
--- a/evalml/tests/component_tests/test_one_hot_encoder.py
+++ b/evalml/tests/component_tests/test_one_hot_encoder.py
@@ -554,7 +554,7 @@ def test_ohe_woodwork_custom_overrides_returned_by_components(X_df):
 
 def test_ohe_output_bools():
     X = ww.DataTable(pd.DataFrame({"bool": [bool(i % 2) for i in range(100)],
-                                   "categorical": ["dog"] * 20 + ["cat"]* 40 + ["fish"] * 40,
+                                   "categorical": ["dog"] * 20 + ["cat"] * 40 + ["fish"] * 40,
                                    "integers": [i for i in range(100)]}))
     y = ww.DataColumn(pd.Series([i % 2 for i in range(100)]))
     ohe = OneHotEncoder()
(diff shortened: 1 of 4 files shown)


evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2050,6 +2050,26 @@ class BinaryPipeline(BinaryClassificationPipeline):
     assert len(preds) == 1000
 
 
+@patch("evalml.pipelines.components.XGBoostClassifier.fit")
+def test_xgboost_ohe(mock_fit):
```
jeremyliweishih, line 2054: is the interaction between OHE and xgboost a special case we need to be careful of?

jeremyliweishih (approved):
I think this looks great Bryan!

evalml/tests/pipeline_tests/test_pipelines.py
freddyaboulton: I'm wondering if we need the test as well. If the purpose is to check that Xgboost works well with bools, then we'd need to not mock fit?

freddyaboulton (approved):
Looks good @bchen1116 !!

evalml/tests/pipeline_tests/test_pipelines.py
bchen1116: Ah fair, I can remove it!

evalml/tests/component_tests/test_one_hot_encoder.py
```diff
@@ -550,3 +550,17 @@ def test_ohe_woodwork_custom_overrides_returned_by_components(X_df):
...
+                                   "categorical": ["dog"] * 20 + ["cat"] * 40 + ["fish"] * 40,
+                                   "integers": [i for i in range(100)]}))
+    y = ww.DataColumn(pd.Series([i % 2 for i in range(100)]))
+    ohe = OneHotEncoder()
+    output = ohe.fit_transform(X, y)
+    for name, types in output.types["Logical Type"].items():
+        if name == 'integers':
+            assert str(types) == "Integer"
+        else:
+            assert str(types) == "Boolean"
```
dsherry, line 566: Looks great! I love that you made this general and not specific to xgboost.

One check I'd add: assert that the right number of features has been generated. That way, if for some reason the OHE isn't generating any features at all for the bool/category inputs, the test will fail.

dsherry (approved):

PR #2182, opened 2021-04-22
WW 0.2.0: Update Data Checks

freddyaboulton:
### Pull Request Description
Update data checks to use the new woodwork accessor

-----

### DIFF
--- a/evalml/data_checks/id_columns_data_check.py
+++ b/evalml/data_checks/id_columns_data_check.py
@@ -6,7 +6,7 @@ from evalml.data_checks import (
     DataCheckMessageCode,
     DataCheckWarning
 )
-from evalml.utils import _convert_woodwork_types_wrapper, infer_feature_types
+from evalml.utils import infer_feature_types
 
 
 class IDColumnsDataCheck(DataCheck):
@@ -30,7 +30,7 @@ class IDColumnsDataCheck(DataCheck):
             - column contains all unique values (and is categorical / integer type)
 
         Arguments:
-            X (ww.DataTable, pd.DataFrame, np.ndarray): The input features to check
+            X (pd.DataFrame, np.ndarray): The input features to check
 
         Returns:
             dict: A dictionary of features with column name or index and their probability of being ID columns
@@ -64,8 +64,7 @@ class IDColumnsDataCheck(DataCheck):
         cols_named_id = [col for col in col_names if (str(col).lower() == "id")]  # columns whose name is "id"
         id_cols = {col: 0.95 for col in cols_named_id}
 
-        X = X.select(include=['Integer', 'Categorical'])
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        X = X.ww.select(include=['Integer', 'Categorical'])
 
         check_all_unique = (X.nunique() == len(X))
         cols_with_all_unique = check_all_unique[check_all_unique].index.tolist()  # columns whose values are all unique
--- a/evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
+++ b/evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
@@ -46,10 +46,10 @@ def test_class_imbalance_data_check_binary(input_type):
         y_balanced = y_balanced.to_numpy()
 
     elif input_type == "ww":
-        X = ww.DataTable(X)
-        y = ww.DataColumn(y)
-        y_long = ww.DataColumn(y_long)
-        y_balanced = ww.DataColumn(y_balanced)
+        X.ww.init()
+        y = ww.init_series(y)
+        y_long = ww.init_series(y_long)
+        y_balanced = ww.init_series(y_balanced)
 
     class_imbalance_check = ClassImbalanceDataCheck(min_samples=1, num_cv_folds=0)
     assert class_imbalance_check.validate(X, y) == {"warnings": [], "errors": [], "actions": []}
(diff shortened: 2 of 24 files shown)


evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
```diff
@@ -46,10 +46,10 @@ def test_class_imbalance_data_check_binary(input_type):
         y_balanced = y_balanced.to_numpy()
 
     elif input_type == "ww":
-        X = ww.DataTable(X)
-        y = ww.DataColumn(y)
-        y_long = ww.DataColumn(y_long)
-        y_balanced = ww.DataColumn(y_balanced)
+        X.ww.init()
```
angela97lin, line 49: O interesting, do we need to set `X = X.ww.init()` here?

angela97lin: (Just wondering between the difference between this and line 114)

angela97lin (approved):
Amazing, looks like this update was nice and clean! 🥂

evalml/data_checks/id_columns_data_check.py
```diff
@@ -64,8 +64,7 @@ def validate(self, X, y=None):
         cols_named_id = [col for col in col_names if (str(col).lower() == "id")]  # columns whose name is "id"
         id_cols = {col: 0.95 for col in cols_named_id}
 
-        X = X.select(include=['Integer', 'Categorical'])
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
```
dsherry, line 68: 👏 👏 👏  yessss

dsherry (approved):
🚢 ✨

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
dsherry: +1, my guess is that the assignment in `X = X.ww.init()` is redundant but harmless

freddyaboulton: Great question! We can't do assign the result of `ww.init` to anything because `ww.init` returns `None`

bchen1116: interesting that `ww.init()` modifies in place while the equivalent for series returns a new series

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
```diff
@@ -111,12 +111,12 @@ def test_class_imbalance_data_check_multiclass(input_type):
         y_long = y_long.to_numpy()
 
     elif input_type == "ww":
-        X = ww.DataTable(X)
-        y = ww.DataColumn(y)
-        y_imbalanced_default_threshold = ww.DataColumn(y_imbalanced_default_threshold)
-        y_imbalanced_set_threshold = ww.DataColumn(y_imbalanced_set_threshold)
-        y_imbalanced_cv = ww.DataColumn(y_imbalanced_cv)
-        y_long = ww.DataColumn(y_long)
+        X = X.ww.init()
```
bchen1116: We shouldn't assign `X =` here then, right?

evalml/tests/data_checks_tests/test_datetime_nan_data_check.py
```diff
@@ -1,6 +1,5 @@
 import numpy as np
 import pandas as pd
-import woodwork as ww
```
bchen1116, line 3: Question, but why do we no longer need to import woodwork, even if we're using `ww.init()` and other ww methods?

bchen1116 (approved):
Looks good! Left a couple of questions, but great job cranking this out so quickly!

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
freddyaboulton: I get what you mean! There is `series.ww.init()` the problem is that if the logical type doesn't match the physical type of the series, it will error out. `ww.init_series` will actually do the type conversion if needed.

image

bchen1116: Got it, thanks for the explanation!

evalml/tests/data_checks_tests/test_datetime_nan_data_check.py
freddyaboulton: Great observation! This is something I hadn't noticed actually. 

In order to use the accessor, woodwork needs to be present in the namespace but it doesn't need to be imported in the same file that you use it. Since our  `evaml/__init__.py` file imports `ww` by way of importing our submodules, `ww` is imported when our tests run so we can use the accessor. 

I suspect I originally kept the import but then deleted it because of lint since it's not "needed".

bchen1116: Oh that's interesting. Thanks for the clarification!

evalml/tests/data_checks_tests/test_class_imbalance_data_check.py
thehomebrewnerd: > Great question! We can't assign the result of `ww.init` to anything because `ww.init` returns `None`. Just fixed a typo about this in this file that I missed.

@freddyaboulton  So, I'm a little late to the party here, but I'm wondering if we should actually return `self` from our Woodwork init method. If I'm thinking about this correctly, doing so would allow for chaining operations like `df.ww.init().ww.describe()` and also make the re-assignment situation valid. Thoughts?

PR #2189, opened 2021-04-26
Use machineFL user token for dependency update bot. 

### LINKED ISSUE #2045: Dependency Bot: Trigger CI Tests
Github action workflows have a limitation where workflows cannot trigger other workflows. In this case our dependency bot putting up a PR cannot trigger our CI tests. To get around this we need to provide a personal access token instead of using the repo token.  Here is an example: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

dsherry:
Fix #2045 

Not sure if this will work since `github-actions` is the PR author but the token was generated with `machineFL` user. But we'll see.

Also adds more reviewers to deps update PRs

### DIFF
--- a/.github/workflows/dependency_check.yml
+++ b/.github/workflows/dependency_check.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Create Pull Request
         uses: FeatureLabs/create-pull-request@v3
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.MACHINEFL_DEPENDENCY_CHECKER_TOKEN }}
           commit-message: Update latest dependencies
           title: Automated Latest Dependency Updates
           author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
@@ -29,4 +29,4 @@ jobs:
           branch: dep-update
           branch-suffix: short-commit-hash
           base: main
-          reviewers: angela97lin, dsherry, jeremyliweishih
+          reviewers: angela97lin, dsherry, jeremyliweishih, freddyaboulton, bchen1116, chukarsten, ParthivNaresh
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -11,6 +11,7 @@ Release Notes
     * Documentation Changes
         * Rename dataset to clarify that its gzipped but not a tarball :pr:`2183`
     * Testing Changes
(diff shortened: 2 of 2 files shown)


.github/workflows/dependency_check.yml
```diff
@@ -29,4 +29,4 @@ jobs:
           branch: dep-update
           branch-suffix: short-commit-hash
           base: main
-          reviewers: angela97lin, dsherry, jeremyliweishih
+          reviewers: angela97lin, dsherry, jeremyliweishih, freddyaboulton, bchen1116, chukarsten, ParthivNaresh
```
dsherry, line 32: Welcome all to the wonderful world of dependency updates 😁

.github/workflows/dependency_check.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Create Pull Request
         uses: FeatureLabs/create-pull-request@v3
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.MACHINEFL_DEPENDENCY_CHECKER_TOKEN }}
```
dsherry, line 23: I think I set this up right...

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

> **Triggering new workflows using a personal access token**
> 
> When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur. For more information, see "Authenticating with the GITHUB_TOKEN."
> 
> If you would like to trigger a workflow from a workflow run, you can trigger the event using a personal access token. You'll need to create a personal access token and store it as a secret. To minimize your GitHub Actions usage costs, ensure that you don't create recursive or unintended workflow runs. For more information on storing a personal access token as a secret, see "Creating and storing encrypted secrets."

jeremyliweishih (approved):
can't check how the secret was added but everything else looks great!

dsherry:
@jeremyliweishih thanks! Yeah let's see what happens the next time the bot runs.

PR #2210, opened 2021-04-30
Spike: have ComponentGraph skip samplers in transform/predict, and refactor sampler code

### LINKED ISSUE #1982: Ensure all pipelines in AutoMLSearch receive the same data splits
Repro

```python
import joblib
from evalml.preprocessing.data_splitters import BalancedClassificationDataCVSplit

splitter = BalancedClassificationDataCVSplit(n_splits=3, random_seed=0, shuffle=True)

X, y = load_fraud(5000)
X = X.to_dataframe()
y = y.to_series().astype("int")

for train, test in splitter.split(X, y):
    print((joblib.hash(train), joblib.hash(test)))

# Output
('75f1b95d7ce307ac6c793055330969aa', '8c89fe1a592c50a700b6d5cbb02dba8b')
('f8c849bbfbed37c13f66c5c742e237cb', '9c4879fb550fded8be9ac03e95a1bf95')
('cdc21f0d6bbf45459c9695258f7f04dc', '5b575765bbe176e732b8eb4dc1bf2822')

for train, test in splitter.split(X, y):
    print((joblib.hash(train), joblib.hash(test)))

# Output
('bf462b82af243c552ac48acad2dfd748', '8c89fe1a592c50a700b6d5cbb02dba8b')
('b8341b536c63c7957c099b05e315f49c', '9c4879fb550fded8be9ac03e95a1bf95')
('780e74673b601790037fc0b17dde56fe', '5b575765bbe176e732b8eb4dc1bf2822')

for train, test in splitter.split(X, y):
    print((joblib.hash(train), joblib.hash(test)

# Output
('385f6c538568ad3a33cf84f61d94144c', '8c89fe1a592c50a700b6d5cbb02dba8b')
('8db65d0a3bdf87ae0f135b9766a260dd', '9c4879fb550fded8be9ac03e95a1bf95')
('2a7293fc1308b8a572091d7c76d20205', '5b575765bbe176e732b8eb4dc1bf2822')
```

This is different from the behavior of the sklearn splitter:

```python
from sklearn.model_selection import StratifiedKFold

kfold = StratifiedKFold(n_splits=3, random_state=0, shuffle=True)

for train, test in kfold.split(X, y):
    print((joblib.hash(train), joblib.hash(test)))

#Output
('6c30ee6a11803927024354405389506a', '8c89fe1a592c50a700b6d5cbb02dba8b')
('df0a70e2e6ca783f12461e8c82a26ad4', '9c4879fb550fded8be9ac03e95a1bf95')
('2898e4b3d3621b436641016499f4aafb', '5b575765bbe176e732b8eb4dc1bf2822')

for train, test in kfold.split(X, y):
    print((joblib.hash(train), joblib.hash(test)))

# Output
('6c30ee6a11803927024354405389506a', '8c89fe1a592c50a700b6d5cbb02dba8b')
('df0a70e2e6ca783f12461e8c82a26ad4', '9c4879fb550fded8be9ac03e95a1bf95')
('2898e4b3d3621b436641016499f4aafb', '5b575765bbe176e732b8eb4dc1bf2822')
```

I think this is problematic for two reasons:

1. Since BalancedClassificationDataCVSplit is the default splitter in automl, it means our pipelines are evaluated on different splits
2. Since `split` modifies the state of the data splitter it means we'll have different results between the sequential and parallel engines.

dsherry:
**Problem**
Currently the following fails. I would expect it to pass.
```python
undersampler = evalml.pipelines.components.Undersampler()
undersampler.fit(X, y)
X_undersampled_t, y_undersampled_t = undersampler.transform(X, y)
undersampler = evalml.pipelines.components.Undersampler()
X_undersampled_ft, y_undersampled_ft = undersampler.fit_transform(X, y)
assert X_undersampled_t == X_undersampled_ft
assert y_undersampled_t == y_undersampled_ft
```
This is because our samplers rely on the fact that the `ComponentGraph` impl just so happens to always call `fit_transform` at fit-time, and `transform` at predict-time. So in the above, the output of `.fit_transform()` will be downsampled but the output of `.fit().transform()` will not.

I think that level of subtlety in a component impl represents potentially dangerous behavior. If the `ComponentGraph` impl happens to change, this will break. Plus it makes it hard for users to look at the component code and understand what's happening.

**Changes**
I think we just need to shuffle the delegation a bit. I think all our transformers should be "dumb" in that they produce the same output regardless of whether you call `.fit().transform()` or `.fit_transform()`. Conversely, I think `ComponentGraph` should be smart enough to *only* include samplers at fit-time. This PR does that.

This PR also refactors the sampler code a bit. I don't think we need the undersampler to support more than one sampling method right now, for instance.

Finally, this PR also fixes #1982 by re-initializing every sampler during component `fit` instead of during component `__init__`, ensuring every invocation uses the same RNG seeding.

Labeled as spike until I can update the unit tests!


angela97lin:
Changes addressed by #2583 and #2595, closing!

PR #2155, opened 2021-04-19
User Specified Date Feature

ParthivNaresh:
Addresses #2070

PR #2214, opened 2021-04-30
Revert "User Specified Date Feature"

### LINKED ISSUE #2215: Build Conda Package Failure during pmdarima installation
Regarding #2155 and by extension #2009 

We're currently [facing a failure](https://github.com/alteryx/evalml/runs/2478486827?check_suite_focus=true) of `build_conda_pkg` on `main` where `pmdarima` isn't being installed despite `has_minimal_dependencies` being set to False.

`AssertionError: The test environment is missing expected extra dependency 'pmdarima'. Please either install all requirements in requirements.txt, or rerun the tests with '--has-minimal-dependencies'.`

ParthivNaresh:
Fixes #2215 
Reverts alteryx/evalml#2155

Reverting until we can figure out what's going on with `pmdarima` during installation

### DIFF
--- a/.github/workflows/linux_unit_tests.yml
+++ b/.github/workflows/linux_unit_tests.yml
@@ -61,7 +61,6 @@ jobs:
           source test_python/bin/activate
           make installdeps
           make installdeps-test
-          pip freeze
       - name: Erase Coverage
         run: |
           source test_python/bin/activate
--- a/.github/workflows/windows_unit_tests.yml
+++ b/.github/workflows/windows_unit_tests.yml
@@ -66,7 +66,6 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install .
           python -m pip install -r test-requirements.txt
-          pip freeze
       - name: Run unit tests
         run: |
           . $env:USERPROFILE\Miniconda3\shell\condabin\conda-hook.ps1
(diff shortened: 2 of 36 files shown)


dsherry (approved):
Thanks for keeping `main` green! Sorry you ran into trouble with conda. We can fix it 😁

freddyaboulton:
@ParthivNaresh @dsherry  What happened?

ParthivNaresh:
@freddyaboulton We moved ARIMA from the `statsmodel` implementation to `sktime` which has `pmdarima` as a soft dependency. So I guess this means we don't need `statsmodel?

freddyaboulton:
Ah I read the diff wrong! The original date feature pr changed the underlying library used by Arima. 

Maybe what we should do then is add pmdarima to the conda recipe? Might need to remove `statsmodels` from our recipe though. I see pmdarima has a max statsmodels [version](https://github.com/conda-forge/pmdarima-feedstock/blob/master/recipe/meta.yaml#L34) that's lower than our min.

freddyaboulton:
@ParthivNaresh Yea I don't think statsmodels is used by anything else. I think it's worth seeing if we can patch up the recipe before merging this pr?

ParthivNaresh:
@freddyaboulton Sorry I didn’t see your message before merging (dang refreshing). But I’ll take a look at the recipe next for sure

PR #2217, opened 2021-05-01
2070 user specified date feature

### LINKED ISSUE #2070: Time series: require user to specify a date feature
**Background**
Currently, our `DatetimeFeaturizer` generates a few indicator features from every datetime-type input feature. 

Our `TimeSeriesSplit` currently assumes the rows have been ordered by time, but otherwise it doesn't assume to have access to the datetime index itself.

**Goal**
For time series modeling, we need to require that users tag one of their input feature as the time index which EvalML should use throughout modeling. If this hasn't been provided but the problem type is time series, the user should get a clear error.

We need this in order for us to support data splitting for time series, and to support models like ARIMA which are aware of time, and things like seasonal detrending.

**Options**
I see two approaches to this. One is to set the woodwork/pandas index to be the datetime feature in question. Another is to require for time series modeling that a "time" feature is provided alongside the other features.

@ParthivNaresh @freddyaboulton 

I think a great way to add this check is via a data check error. The action to fix would be to halt, ask the user to choose their datetime index and try again. We'll also need a good `AutoMLSearch` error for this case though.

ParthivNaresh:
Fixes #2070 

This is a refiling of the old PR with the previous (main) version of ARIMA. After this is merged, ARIMA will be updated in a different PR using `sktime` as opposed to `statsmodels`.


.github/workflows/build_conda_pkg.yml
```diff
@@ -1,6 +1,8 @@
 name: Build Conda Package
 
 on:
+  pull_request:
+    types: [ opened, synchronize ]
```
ParthivNaresh, line 5: Do we want to keep this for feature builds moving forward?

dsherry:
@ParthivNaresh I see this in the description:
> After this is merged, ARIMA will be updated using sktime as opposed to statsmodels.
This is no longer true, correct? If so please delete :)

.github/workflows/build_conda_pkg.yml
freddyaboulton: We can try it out and if it's too slow we can undo the change?

evalml/tests/automl_tests/test_time_series_split.py
```diff
@@ -16,41 +17,85 @@ def test_time_series_split_init():
...
+            else:
+                pd.testing.assert_index_equal(X_train.index, answer[i][0])
+                pd.testing.assert_index_equal(X_test.index, answer[i][1])
         if not y_none:
             y_train, y_test = y.iloc[train], y.iloc[test]
             pd.testing.assert_index_equal(y_train.index, answer[i][0])
             pd.testing.assert_index_equal(y_test.index, answer[i][1])
+
+
+@pytest.mark.parametrize("max_delay,gap,date_index", [(0, 0, "Date"), (1, 0, "Date"), (0, 1, "Date")])
```
freddyaboulton, line 71: I think we can delete this test? The ts data splitter doesn't sort the data.

ParthivNaresh:
@dsherry Sorry I meant after this is merged, ARIMA will be updated in another PR :)

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -318,7 +318,7 @@ def __init__(self, dummy_int_parameter=0, dummy_categorical_parameter='dummy', d
 
     pipeline = BinaryClassificationPipeline([MockEstimator])
     algo = IterativeAlgorithm(allowed_pipelines=[pipeline, pipeline, pipeline],
-                              pipeline_params={'pipeline': {"gap": 2, "max_delay": 10}},
+                              pipeline_params={'pipeline': {'date_index': "Date", "gap": 2, "max_delay": 10}},
```
freddyaboulton, line 321: 👍

freddyaboulton (approved):
@ParthivNaresh Looks great!

.github/workflows/build_conda_pkg.yml
dsherry: Sure I am a fan!

docs/source/release_notes.rst
```diff
@@ -2,6 +2,7 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
+        * Added `date_index` as a required parameter for TimeSeries problems :pr:`2217`
```
dsherry, line 5: 👍

dsherry (approved):
Excellent!!

bchen1116 (approved):
👍

PR #2218, opened 2021-05-02
Delete `make_pipeline_from_components`

### LINKED ISSUE #2173: Delete make_pipeline_from_components
After #2091, our pipeline API now accepts lists/graphs of components as input during instance init rather than requiring they be specified in the class definition.

Before this was the case, we had added `make_pipeline_from_components` as a workaround to create a pipeline class definition from a list of components. With the API upgrade this is no longer necessary! 🥳 

Original comment: https://github.com/alteryx/evalml/pull/2091/files#r613584179

angela97lin:
Closes #2173


evalml/pipelines/utils.py
```diff
@@ -144,46 +145,6 @@ def make_pipeline(X, y, estimator, problem_type, parameters=None, custom_hyperpa
     return base_class(complete_component_graph, parameters=parameters, custom_hyperparameters=custom_hyperparameters)
 
 
-def make_pipeline_from_components(component_instances, problem_type, custom_name=None, random_seed=0):
```
freddyaboulton, line 147: I'm debating whether we should keep this around for another release and instead issue a deprecation warning in this pr.

I'm ok with deleting this util, I don't think it got a lot of traction from users but just noting that I think we use a different process for deleting functions/methods.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -529,21 +455,21 @@ def test_stacked_estimator_in_pipeline(problem_type, X_y_binary, X_y_multi, X_y_
...
-        input_pipelines = [make_pipeline_from_components([classifier], problem_type) for classifier in stackable_classifiers]
+        input_pipelines = [base_pipeline_class([classifier]) for classifier in stackable_classifiers]
         comparison_pipeline = logistic_regression_multiclass_pipeline_class(parameters={"Logistic Regression Classifier": {"n_jobs": 1}})
         objective = 'Log Loss Multiclass'
     elif problem_type == ProblemTypes.REGRESSION:
         X, y = X_y_regression
         base_pipeline_class = RegressionPipeline
         stacking_component_name = StackedEnsembleRegressor.name
-        input_pipelines = [make_pipeline_from_components([regressor], problem_type) for regressor in stackable_regressors]
+        input_pipelines = [base_pipeline_class([regressor]) for regressor in stackable_regressors]
```
freddyaboulton, line 472: nit-pick: I don't think there's a need for `base_pipeline_class` anymore.

freddyaboulton (approved):
Looks great @angela97lin ! I love the net deletion of lines hehe

bchen1116 (approved):
LGTM! Agree with Freddy that it might be useful to deprecate before immediately removing from EvalML

evalml/pipelines/utils.py
angela97lin: Interesting point, but I'm okay deleting immediately, since I think the main use of this was internal alignment :p Plus, it's not too difficult to do what `make_pipeline_from_components` with our new pipelines so not too much lost there!

evalml/tests/pipeline_tests/test_pipeline_utils.py
angela97lin: Hmmm, I think we still need it for the line after (depending on which problem type, we create a specific pipeline class) but I'm okay with updating this line to just use RegressionPipeline so it's more clear! 😁

PR #2222, opened 2021-05-03
Add null rows check to `HighlyNullDataCheck`

### LINKED ISSUE #2220: Add `HighlyNullRowsDataCheck`
Ask: we need a data check that returns the percentage of rows in a dataset with over 90% (or some other configurable threshold) null values.

jeremyliweishih:
Fixes #2220.


evalml/data_checks/highly_null_rows_data_check.py
```diff
@@ -0,0 +1,85 @@
...
+    DataCheckMessageCode,
+    DataCheckWarning
+)
+from evalml.utils import _convert_woodwork_types_wrapper, infer_feature_types
+
+
+class HighlyNullRowsDataCheck(DataCheck):
+    """Checks if there are any highly-null rows in the input."""
+
+    def __init__(self, pct_null_threshold=0.5):
```
jeremyliweishih, line 14: I set default to `0.5` because I think this can be more aggressive on removing individual samples/rows than removing a whole column in `HighlyNullDataCheck`.

evalml/data_checks/highly_null_rows_data_check.py
```diff
@@ -0,0 +1,85 @@
...
+            results["warnings"].extend([DataCheckWarning(message=warning_msg.format(row_name),
+                                                         data_check_name=self.name,
+                                                         message_code=DataCheckMessageCode.HIGHLY_NULL_ROWS,
+                                                         details={"row": row_name, "pct_null_cols": highly_null_rows[row_name]}).to_dict()
+                                        for row_name in highly_null_rows])
+        else:
+            highly_null_rows = {key: value for key, value in percent_null.items() if value >= self.pct_null_threshold}
+            warning_msg = "Row '{}' is {}% or more null"
+            results["warnings"].extend([DataCheckWarning(message=warning_msg.format(row_name, self.pct_null_threshold * 100),
+                                                         data_check_name=self.name,
```
chukarsten, line 77: Do 0% and > 0% need to be handled differently?

evalml/tests/data_checks_tests/test_highly_null_rows_data_check.py
```diff
@@ -0,0 +1,102 @@
...
+    assert highly_null_check.pct_null_threshold == 0.5
+
+    highly_null_check = HighlyNullRowsDataCheck(pct_null_threshold=1.0)
+    assert highly_null_check.pct_null_threshold == 1.0
+
+    with pytest.raises(ValueError, match="pct_null_threshold must be a float between 0 and 1, inclusive."):
+        HighlyNullRowsDataCheck(pct_null_threshold=-0.1)
+    with pytest.raises(ValueError, match="pct_null_threshold must be a float between 0 and 1, inclusive."):
+        HighlyNullRowsDataCheck(pct_null_threshold=1.1)
+
```
chukarsten, line 32: I really like these tests.  Super clear to read and infer the behavior of the datacheck.  Great job.

evalml/tests/data_checks_tests/test_highly_null_rows_data_check.py
```diff
@@ -0,0 +1,102 @@
...
+
+def test_highly_null_data_check_warnings():
+    data = pd.DataFrame({'a': [None, None, 10],
+                         'b': [None, "text", "text_1"]})
```
chukarsten, line 36: Is None the only "null" we're considering here?  We might want to consider similar tests with np.nan or other equivalent types of "null".  I think you could probably parameterize it:
```
@pytest.mark.parametrize("null_type", [None, np.nan])
def test_highly_null_data_check_warnings(null_type):
    data = pd.DataFrame({'a': [null_type, null_type, 10],
                         'b': [null_type, "text", "text_1"]})
```
Or something like that.

chukarsten (changes requested):
Great work here, Jeremy.  I think we should merge the logic for the 0% and the rest of the percentage empty in the data check .validate().  Unless there's a really compelling reason for this additional complexity, I think it'll be good to drop the if branching and merge them.  Additionally, we might want to talk about adding test coverage for other null types? This looks great though.  Really like the tests - very readable.

evalml/data_checks/default_data_checks.py
```diff
@@ -16,6 +17,7 @@ class DefaultDataChecks(DataChecks):
     Includes:
 
         - `HighlyNullDataCheck`
```
freddyaboulton, line 19: Maybe we should rename this to `HighlyNullColumnsDataCheck`? No need to handle in this pr. Probably best to issue a deprecation warning if we decide to do it.

evalml/data_checks/highly_null_rows_data_check.py
```diff
@@ -0,0 +1,85 @@
...
+    DataCheck,
+    DataCheckAction,
+    DataCheckActionCode,
+    DataCheckMessageCode,
+    DataCheckWarning
+)
+from evalml.utils import _convert_woodwork_types_wrapper, infer_feature_types
+
+
+class HighlyNullRowsDataCheck(DataCheck):
```
freddyaboulton, line 11: I wonder if we can share some code with `HighlyNullDataCheck`. The only computational difference I see is

`(X.isnull().mean()).to_dict()` vs `(X.isnull().mean(axis=1)).to_dict()`.

I'd be happy to punt on this though, doesn't seem super pressing.

freddyaboulton (approved):
@jeremyliweishih Looks great! I think this is good to merge pending @chukarsten 's comments.

evalml/data_checks/default_data_checks.py
jeremyliweishih: good call - I'll put up an issue for it 😄

evalml/data_checks/highly_null_rows_data_check.py
jeremyliweishih: I thought about this as well but I felt that it fits better into `DefaultDataChecks` and `DataChecks` better as separate data checks. I don't feel too strongly on this though - happy to see what others think!

bchen1116 (changes requested):
This implementation looks good! I did have a question that I think we should address, whether in this PR or in a future issue.

Currently, we find the null values with the original dataset, but I can see issues if we have a dataset like:
```python
X = pd.DataFrame({"col_1": [None, None, None],
                                  "col_2": [None, 1, 2],
                                  "col_3": [1, 2, None]})
```
Then the data check for null columns would suggest dropping col 1, and the data check for null rows would drop cols 2 and 3. I'm curious if we want to drop the null columns BEFORE finding the rows that satisfy this data check. 

It's mainly for larger datasets, where there might be 1000+ columns, and maybe half have empty columns. These two `highlyNull` datachecks would then drop all the rows and columns, which doesn't seem like the behavior we want. 

Just something that could raise problems in the future. Requested changes since I wanted to have this discussion before merging!

jeremyliweishih:
@bchen1116 good points! I'm not sure if there has been discussions on the order of operations with data check actions but I do like your suggestion. A solution now we could do would be to bump `pct_null_threshold` up so that the case you brought up won't happen. I can bump it up to something like `0.75` which may be more reasonable. What do you think?

bchen1116:
@jeremyliweishih yeah, I think that's a good short-term solution! Might need to discuss further for a longer-term solution

bchen1116 (approved):
Nice!

dsherry (changes requested):
@jeremyliweishih I'd like you to change the following:
* Add to existing `HighlyNullDataCheck` instead of creating a separate data check
* Return a single `DataCheckWarning` when one or more rows are highly null, instead of one warning for each row
* In the warning `details`, save info about which rows are highly null and what the percentage was using a single pandas `Series` instead of a dict.

evalml/data_checks/highly_null_data_check.py
```diff
@@ -58,25 +75,25 @@ def validate(self, X, y=None):
...
-                                                         details={"column": col_name, "pct_null_rows": highly_null_cols[col_name]}).to_dict()
-                                        for col_name in highly_null_cols])
+        percent_null_rows = X.isnull().mean(axis=1)
+        highly_null_rows = percent_null_rows[percent_null_rows >= self.pct_null_threshold]
+        if len(highly_null_rows) > 0:
+            warning_msg = f"{len(highly_null_rows)} out of {len(X)} rows are more than {self.pct_null_threshold*100}% null"
+            results["warnings"].append(DataCheckWarning(message=warning_msg,
+                                                        data_check_name=self.name,
+                                                        message_code=DataCheckMessageCode.HIGHLY_NULL_ROWS,
+                                                        details={"pct_null_cols": highly_null_rows}).to_dict())
```
dsherry, line 85: Thanks! Looks great

evalml/data_checks/highly_null_data_check.py
```diff
@@ -9,45 +9,62 @@
...
+    """Checks if there are any highly-null columns and rows in the input."""
 
     def __init__(self, pct_null_threshold=0.95):
-        """Checks if there are any highly-null columns in the input.
+        """Checks if there are any highly-null columns and rows in the input.
 
         Arguments:
             pct_null_threshold(float): If the percentage of NaN values in an input feature exceeds this amount,
-                that feature will be considered highly-null. Defaults to 0.95.
+                that column/row will be considered highly-null. Defaults to 0.95.
```
dsherry, line 19: Ah, I do think it would be helpful to have a separate threshold for rows vs columns.

I filed #2270 to track that separately.

dsherry (approved):
@jeremyliweishih thanks, nice going! 🚢 😁

PR #2223, opened 2021-05-04
Sort first batch of automl

### LINKED ISSUE #1341: AutoML first batch: define evaluation order
Currently, the order the first batch will be evaluated in is an artifact of the code--its determined based on the BFS performed [in the utils](https://github.com/alteryx/evalml/blob/main/evalml/utils/gen_utils.py#L146) (see `get_estimators` in the component utils for the stack, that's called in `AutoMLSearch` and ends up invoking the BFS in `_get_subclasses`).

The first automl batch currently runs the baseline pipeline, then runs one pipeline with default params for each allowed pipeline.

If a user selects `max_iterations<7`, they'd only get a subset of the first batch. For that use-case, it would be nice if we had a smart ordering to our pipeline evaluations. My recommendation would be to determine this ordering based on two considerations:
* Simplest models first. Linear/logreg should be first
* Fastest average fit-time first. Let's look at our perf test data, plot the models' fit-times relative to data size, and use that to inform the ordering.

This may be a moot point thanks to the parallel automl work happening in #144 , which will evaluate pipelines in parallel instead of in series.

dsherry:
Fix #1341 

Sort first batch of `IterativeAlgorithm` by model family. Linear models first. Anything which isn't listed will appear at the end of the batch.

This is helpful if someone specifies `max_iterations` which is less than the full number which automl would run by default (which is currently 9 for binary/multiclass/regression problem types).

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -9,6 +9,16 @@ from .automl_algorithm import AutoMLAlgorithm, AutoMLAlgorithmException
 from evalml.model_family import ModelFamily
 from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
 
+_ESTIMATOR_FAMILY_ORDER = [
+    ModelFamily.LINEAR_MODEL,
+    ModelFamily.DECISION_TREE,
+    ModelFamily.EXTRA_TREES,
+    ModelFamily.RANDOM_FOREST,
+    ModelFamily.XGBOOST,
+    ModelFamily.LIGHTGBM,
+    ModelFamily.CATBOOST
+]
+
 
 class IterativeAlgorithm(AutoMLAlgorithm):
     """An automl algorithm which first fits a base round of pipelines with default parameters, then does a round of parameter tuning on each pipeline in order of performance."""
@@ -23,7 +33,8 @@ class IterativeAlgorithm(AutoMLAlgorithm):
                  number_features=None,  # TODO remove
                  ensembling=False,
                  pipeline_params=None,
-                 _frozen_pipeline_parameters=None):
+                 _frozen_pipeline_parameters=None,
+                 _estimator_family_order=None):
         """An automl algorithm which first fits a base round of pipelines with default parameters, then does a round of parameter tuning on each pipeline in order of performance.
 
         Arguments:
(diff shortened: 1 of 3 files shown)


evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -9,6 +9,16 @@
...
 from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
 
+_ESTIMATOR_FAMILY_ORDER = [
+    ModelFamily.LINEAR_MODEL,
+    ModelFamily.DECISION_TREE,
+    ModelFamily.EXTRA_TREES,
+    ModelFamily.RANDOM_FOREST,
+    ModelFamily.XGBOOST,
+    ModelFamily.LIGHTGBM,
+    ModelFamily.CATBOOST
```
dsherry, line 19: Linear models should definitely go first. And I feel fine putting xgboost, lightgbm and catboost last.

I'm not sure we can say much a priori about the sort order between decision tree, extratrees and random forest at this time.

Current model families not shown here, which will therefore be pushed to the end of the first batch:
* K_NEIGHBORS: not currently included in automl
* SVM: not currently included in automl
* ARIMA
* ENSEMBLE: won't be included by iterative algo
* BASELINE: won't be included by iterative algo
* NONE: none of our estimators should have this label

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -37,7 +48,22 @@ def __init__(self,
...
+        pipelines_end = []
+        for pipeline in allowed_pipelines or []:
+            if pipeline.model_family in self._estimator_family_order:
+                indices.append(self._estimator_family_order.index(pipeline.model_family))
+                pipelines_to_sort.append(pipeline)
+            else:
+                pipelines_end.append(pipeline)
+        pipelines_start = [pipeline for _, pipeline in (sorted(zip(indices, pipelines_to_sort),
+                                                               key=lambda pair: pair[0]) or [])]
+        allowed_pipelines = pipelines_start + pipelines_end
```
dsherry, line 65: There were a few options for where to put this code:
1. In `IterativeAlgorithm.__init__`, when `allowed_pipelines` is passed in from `AutoMLAlgorithm.__init__`
2. In `AutoMLAlgorithm.__init__`, when `allowed_pipelines` is first created.
3. In `get_estimators` and the underlying utils, which define the order for the `allowed_pipelines` list created in `AutoMLAlgorithm.__init__`

I went with option 1.

My reasoning: the sort order of the first batch, and in fact the definition of what pipelines will be included in what automl batch, is owned by the automl algorithm instance. And `IterativeAlgorithm` defines our current default automl algorithm.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -370,12 +372,51 @@ def test_iterative_algorithm_results_best_pipeline_info_id(dummy_binary_pipeline
     scores = np.arange(0, len(next_batch))
     for pipeline_num, (score, pipeline) in enumerate(zip(scores, next_batch)):
         algo.add_result(score, pipeline, {"id": algo.pipeline_number + pipeline_num})
-    assert algo._best_pipeline_info[ModelFamily.RANDOM_FOREST]['id'] == 2
-    assert algo._best_pipeline_info[ModelFamily.LINEAR_MODEL]['id'] == 3
+    assert algo._best_pipeline_info[ModelFamily.RANDOM_FOREST]['id'] == 3
+    assert algo._best_pipeline_info[ModelFamily.LINEAR_MODEL]['id'] == 2
```
dsherry, line 376: This flipped because we placed linear model before random forest in the sort order.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -370,12 +372,53 @@ def test_iterative_algorithm_results_best_pipeline_info_id(dummy_binary_pipeline
...
 
     for i in range(1, 3):
         next_batch = algo.next_batch()
         scores = -np.arange(1, len(next_batch))  # Score always gets better with each pipeline
         for pipeline_num, (score, pipeline) in enumerate(zip(scores, next_batch)):
             algo.add_result(score, pipeline, {"id": algo.pipeline_number + pipeline_num})
             assert algo._best_pipeline_info[pipeline.model_family]['id'] == algo.pipeline_number + pipeline_num
+
+
+@pytest.mark.parametrize("problem_type", [ProblemTypes.REGRESSION, ProblemTypes.BINARY, ProblemTypes.MULTICLASS])
```
bchen1116, line 386: Can we add an additional test where we pass in an `_estimator_family_order` that's different from the default to ensure we use the new one rather than the default?

bchen1116 (approved):
Nice! Left a comment about an additional test, but looks good!

evalml/tests/automl_tests/test_iterative_algorithm.py
dsherry: We sure can! Good idea.

Given that the default works, I am gonna merge this now so that it gets into the release, and will address this comment in a separate PR.

PR #2242, opened 2021-05-09
Add Prophet Regressor

### LINKED ISSUE #1499: Time series regression: add Prophet estimator

ParthivNaresh:
Fixes #1499 

Thanks to a conversation with @dsherry , @kmax12 , and @tyler3991, we made the decision to include Prophet alongside the `cmdstanpy` backend. The reasoning behind this is that because we aren't distributing Prophet code (simply requiring the user to install it as a dependency) we are in the same position as Prophet is regarding its use of `pystan 2` and its GPLv3 license. As we aren't distributing the underlying code, we (and Prophet) aren't required to be under the same copyleft license.

However, because we have to think downstream to Tempo and whether it will be installed on-premises (which would require packaging up dependencies and distributing them), we can't rely on `pystan 2` as a backend. Because Prophet doesn't support `pystan 3` yet (which uses a permissive license), we will have to rely on `cmdstanpy`.

For more information read [here](https://alteryx.quip.com/aBgLAxP5WoiE/Licensing-Prophet)

This PR only deals with adding the estimator so we can get it into main. I'll be raising another PR to add it to AutoML alongside perf tests.

### DIFF
--- a/.github/workflows/linux_unit_tests_with_latest_deps.yml
+++ b/.github/workflows/linux_unit_tests_with_latest_deps.yml
@@ -65,7 +65,18 @@ jobs:
           # "!" negates return code. exit nonzero if any of these deps are found
           ! pip freeze | grep -E "xgboost|catboost|lightgbm|plotly|ipywidgets|category_encoders"
           exit $?
-      - if: ${{ !matrix.core_dependencies }}
+      - if: ${{ !matrix.core_dependencies && (matrix.command == 'git-test-automl' || matrix.command == 'git-test-other') }}
+        name: Installing Dependencies and Prophet
+        run: |
+          pip install virtualenv
+          virtualenv test_python -q
+          source test_python/bin/activate
+          pip install cmdstan-builder==0.0.4
+          make installdeps
+          make installdeps-test
+          pip uninstall pystan -y
+          pip freeze
+      - if: ${{ !matrix.core_dependencies && (matrix.command == 'git-test-modelunderstanding' || matrix.command == 'git-test-dask') }}
         name: Installing Dependencies
         run: |
           pip install virtualenv
@@ -73,6 +84,7 @@ jobs:
           source test_python/bin/activate
           make installdeps
           make installdeps-test
+          pip uninstall pystan -y
           pip freeze
       - name: Erase Coverage
         run: |
--- a/.github/workflows/linux_unit_tests_with_latest_deps.yml
+++ b/.github/workflows/linux_unit_tests_with_latest_deps.yml
@@ -84,7 +84,6 @@ jobs:
           source test_python/bin/activate
           make installdeps
           make installdeps-test
-          pip uninstall pystan -y
           pip freeze
       - name: Erase Coverage
         run: |
--- a/.github/workflows/linux_unit_tests_with_latest_deps.yml
+++ b/.github/workflows/linux_unit_tests_with_latest_deps.yml
@@ -74,7 +74,6 @@ jobs:
           pip install cmdstan-builder==0.0.4
           make installdeps
           make installdeps-test
-          pip uninstall pystan -y
           pip freeze
       - if: ${{ !matrix.core_dependencies && (matrix.command == 'git-test-modelunderstanding' || matrix.command == 'git-test-dask') }}
         name: Installing Dependencies
--- a/.github/workflows/linux_unit_tests_with_minimum_deps.yml
+++ b/.github/workflows/linux_unit_tests_with_minimum_deps.yml
@@ -35,7 +35,17 @@ jobs:
           virtualenv test_python -q
           source test_python/bin/activate
           python -m pip install --upgrade pip -q
-      - name: install evalml with test dependencies, core dependencies, & optional requirements
+      - if: ${{ matrix.command == 'git-test-automl' || matrix.command == 'git-test-other' }}
+        name: Install evalml with test dependencies, core dependencies, & optional requirements (Prophet)
+        run: |
+          source test_python/bin/activate
+          pip install cmdstan-builder==0.0.4
+          pip install -e . --no-dependencies
+          pip install -r evalml/tests/dependency_update_check/minimum_test_requirements.txt
+          pip install -r evalml/tests/dependency_update_check/minimum_core_requirements.txt
+          pip install -r evalml/tests/dependency_update_check/minimum_requirements.txt
+      - if: ${{ matrix.command == 'git-test-modelunderstanding' || matrix.command == 'git-test-dask' }}
+        name: Install evalml with test dependencies, core dependencies, & optional requirements
         run: |
           source test_python/bin/activate
           pip install -e . --no-dependencies
--- a/evalml/pipelines/components/estimators/regressors/prophet_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/prophet_regressor.py
@@ -1,0 +1,161 @@
+import copy
+
+import numpy as np
+import pandas as pd
+from skopt.space import Real
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+from evalml.utils import import_or_raise, infer_feature_types
+from evalml.utils.gen_utils import classproperty
+
+
+class ProphetRegressor(Estimator):
+    """
+    Prophet is a procedure for forecasting time series data based on an additive model where non-linear trends are fit with yearly, weekly, and daily seasonality, plus holiday effects.
+    It works best with time series that have strong seasonal effects and several seasons of historical data. Prophet is robust to missing data and shifts in the trend, and typically handles outliers well.
+
+    More information here: https://facebook.github.io/prophet/
+
+    """
+
+    name = "Prophet Regressor"
+    hyperparameter_ranges = {
+        "changepoint_prior_scale": Real(0.001, 0.5),
+        "seasonality_prior_scale": Real(0.01, 10),
+        "holidays_prior_scale": Real(0.01, 10),
+        "seasonality_mode": ["additive", "multiplicative"],
+    }
+    """{
+        "changepoint_prior_scale": Real(0.001, 0.5),
+        "seasonality_prior_scale": Real(0.01, 10),
+        "holidays_prior_scale": Real(0.01, 10),
+        "seasonality_mode": ["additive", "multiplicative"],
+    }"""
+    model_family = ModelFamily.PROPHET
+    """ModelFamily.PROPHET"""
+    supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
+    """[ProblemTypes.TIME_SERIES_REGRESSION]"""
+
+    def __init__(
+        self,
+        date_column="ds",
+        changepoint_prior_scale=0.05,
+        seasonality_prior_scale=10,
+        holidays_prior_scale=10,
+        seasonality_mode="additive",
+        random_seed=0,
+        stan_backend="CMDSTANPY",
+        **kwargs
+    ):
+        self.date_column = date_column
+
+        parameters = {
+            "changepoint_prior_scale": changepoint_prior_scale,
+            "seasonality_prior_scale": seasonality_prior_scale,
+            "holidays_prior_scale": holidays_prior_scale,
+            "seasonality_mode": seasonality_mode,
+            "stan_backend": stan_backend,
+        }
+
+        parameters.update(kwargs)
+
+        c_error_msg = (
+            "cmdstanpy is not installed. Please install using `pip install cmdstanpy`."
+        )
+        _ = import_or_raise("cmdstanpy", error_msg=c_error_msg)
+
+        p_error_msg = (
+            "prophet is not installed. Please install using `pip install prophet`."
+        )
+        prophet = import_or_raise("prophet", error_msg=p_error_msg)
+
+        prophet_regressor = prophet.Prophet(**parameters)
+        super().__init__(
+            parameters=parameters,
+            component_obj=prophet_regressor,
+            random_state=random_seed,
+        )
+
+    @staticmethod
+    def build_prophet_df(X, y=None, date_column="ds"):
+        if X is not None:
+            X = copy.deepcopy(X)
+        if y is not None:
+            y = copy.deepcopy(y)
+
+        if date_column in X.columns:
+            date_col = X[date_column]
+        elif isinstance(X.index, pd.DatetimeIndex):
+            date_col = X.reset_index()
+            date_col = date_col["index"]
+        elif isinstance(y.index, pd.DatetimeIndex):
+            date_col = y.reset_index()
+            date_col = date_col["index"]
+        else:
+            msg = "Prophet estimator requires input data X to have a datetime column specified by the 'date_column' parameter. If it doesn't find one, it will look for the datetime column in the index of X or y."
+            raise ValueError(msg)
+
+        date_col = date_col.rename("ds")
+        prophet_df = date_col.to_frame()
+        if y is not None:
+            y.index = prophet_df.index
+            prophet_df["y"] = y
+        return prophet_df
+
+    def fit(self, X, y=None):
+        if X is None:
+            X = pd.DataFrame()
+
+        X, y = super()._manage_woodwork(X, y)
+
+        prophet_df = ProphetRegressor.build_prophet_df(
+            X=X, y=y, date_column=self.date_column
+        )
+
+        self._component_obj.fit(prophet_df)
+        return self
+
+    def predict(self, X, y=None):
+        if X is None:
+            X = pd.DataFrame()
+
+        X = infer_feature_types(X)
+
+        prophet_df = ProphetRegressor.build_prophet_df(
+            X=X, y=y, date_column=self.date_column
+        )
+
+        y_pred = self._component_obj.predict(prophet_df)["yhat"]
+        return y_pred
+
+    def get_params(self):
+        return self.__dict__["_parameters"]
+
+    @property
+    def feature_importance(self):
+        """
+        Returns array of 0's with len(1) as feature_importance is not defined for Prophet regressor.
+        """
+        return np.zeros(1)
+
+    @classproperty
+    def default_parameters(cls):
+        """Returns the default parameters for this component.
+
+        Our convention is that Component.default_parameters == Component().parameters.
+
+        Returns:
+            dict: default parameters for this component.
+        """
+
+        parameters = {
+            "changepoint_prior_scale": 0.05,
+            "seasonality_prior_scale": 10,
+            "holidays_prior_scale": 10,
+            "seasonality_mode": "additive",
+            "stan_backend": "CMDSTANPY",
+        }
+
+        return parameters
--- a/evalml/tests/component_tests/test_components.py
+++ b/evalml/tests/component_tests/test_components.py
@@ -44,6 +44,7 @@ from evalml.pipelines.components import (
     OneHotEncoder,
     PerColumnImputer,
     PolynomialDetrender,
+    ProphetRegressor,
     RandomForestClassifier,
     RandomForestRegressor,
     RFClassifierSelectFromModel,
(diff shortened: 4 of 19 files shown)


ParthivNaresh:
A current issue I'm facing regarding the `cmdstanpy` backend is during following [these](https://github.com/facebook/prophet/blob/master/README.md) instructions.

After installation of the backend `cmdstanpy`, a function has to be run from the installed package: `install_cmdstan.py`
This is responsible for downloading CmdStan from GitHub and building the CmdStan utilities. After this is done, a new directory is create in the default $HOME location (or whatever directory was specified when `python install_cmdstan.py` was run) by the name of `cmdstan-2.26.1`,

The path to this folder has to be assigned to the environment variable CMDSTAN. It can either be assigned to it separately as 
`export CMDSTAN=<path/to/dir>` 
or as part of the final installation command for Prophet through: 
`CMDSTAN=<path/to/dir> STAN_BACKEND=CMDSTANPY pip install prophet` 

Prophet includes `pystan` naturally as part of the installation process. To specify using `cmdstanpy`, a parameter has to be passed during Prophet instantiation: `stan_backend='CMDSTANPY`. If this parameter isn't specified, it will revert to the `pystan` implementation.
```
m = prophet.forecaster.Prophet()
print(f'Using backend: {m.stan_backend.get_type()}')
>>> Using backend: PYSTAN
```

If the backend is specified as `cmdstanpy`, an error is thrown:
```
m = prophet.forecaster.Prophet(stan_backend='CMDSTANPY')
print(f'Using backend: {m.stan_backend.get_type()}')
>>> ValueError: no such file /Users/parthiv.naresh/miniconda3/envs/prophet/lib/python3.8/site-packages/prophet/stan_model/prophet_model.bin```

It appears to be looking for a file, `prophet_model.bin` which doesn't exist, whereas if the default `pystan` backend is used, then the file `prophet_model.pkl` is used, which does exist.

File: .github/workflows/linux_unit_tests_with_latest_deps.yml [hunk 1]
```diff
@@ -66,11 +66,16 @@ jobs:
           ! pip freeze | grep -E "xgboost|catboost|lightgbm|plotly|ipywidgets|category_encoders"
           exit $?
       - if: ${{ !matrix.core_dependencies }}
-        name: Installing Dependencies
+        name: Installing Dependencies and Prophet
```
freddyaboulton, line 69: Right now we're installing our dependencies (which installs prophet since it's in our requirements file) and then we have to reinstall prophet with the right backend.

Although that works, I think we can simplify it a bit. I tried this out locally, and it seems to work (in the sense the unit tests pass)

```
pip install cmdstanpy==0.9.68
<compile cmdstan>
export CMDSTAN=<path-to-cmdstan>
export STAN_BACKEND=CMDSTANPY
pip install -r dev-requirements.txt
```

I think this also makes it easier to communicate to users how to properly install prophet with the right backend.

This reminds me, we should add a section to the install page in the docs about prophet installation! I'm hoping it can be this:

```
pip install cmdstanpy==0.9.68
<compile cmdstan>
export CMDSTAN=<path-to-cmdstan>
export STAN_BACKEND=CMDSTANPY
pip install evalml
```

evalml/utils/gen_utils.py
```diff
@@ -490,3 +491,38 @@ def deprecate_arg(old_arg, new_arg, old_value, new_value):
         )
         value_to_use = old_value
     return value_to_use
+
+
+class suppress_stdout_stderr(object):
```
freddyaboulton, line 496: Omg prophet 😓  I don't know how I feel about opening files in the os from a component. Will this work if two prophet regressors are running simultaneously? 

I agree that supressing the stan output is prefered but is there something else we can do?

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def __init__(
+        self,
+        date_column="ds",
+        changepoint_prior_scale=0.05,
+        seasonality_prior_scale=10,
+        holidays_prior_scale=10,
+        seasonality_mode="additive",
+        random_seed=0,
+        stan_backend="CMDSTANPY",
```
freddyaboulton, line 41: This isn't meant to be changed right? I thought the plan was to only support prophet with the CMDSTANPY backend?

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,155 @@
...
+            "cmdstanpy is not installed. Please install using `pip install cmdstanpy`."
+        )
+        _ = import_or_raise("cmdstanpy", error_msg=c_error_msg)
+
+        p_error_msg = (
+            "prophet is not installed. Please install using `pip install prophet`."
+        )
+        prophet = import_or_raise("prophet", error_msg=p_error_msg)
+
+        prophet_regressor = prophet.Prophet(**parameters)
```
freddyaboulton, line 66: If a user does `pip install evalml` but they didn't go through the steps to install prophet, they would get a weird error about the model not being compiled right? 

Can we add some more statements like `import_or_raise` to make sure prophet was installed with the right backend? We may have to check that the `prophet_model.bin` file is present? 


I think the one flaw with our approach for setting up the installation right now is that it lets users get into this state by "default" if they do `pip install evalml`. I think this may be a good reason to do use the `extras_require` pattern like woodwork [does](https://github.com/alteryx/woodwork/blob/main/setup.py#L9).

FYI @dsherry

evalml/tests/component_tests/test_prophet_regressor.py
```diff
@@ -0,0 +1,154 @@
+import numpy as np
```
freddyaboulton, line 1: There are multiple tests taking about ~15 seconds. Can we tweak the prophet parameters to make them go faster. The data is already small enough lol 😨 

```
=========================================== slowest durations ===========================================
18.53s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_with_X_index
17.10s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_date_col
16.45s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_with_y_index
15.55s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_no_X
7.60s call     evalml/tests/component_tests/test_prophet_regressor.py::test_feature_importance
0.01s call     evalml/tests/component_tests/test_prophet_regressor.py::test_cmdstanpy_backend
```

freddyaboulton (commented):
@ParthivNaresh This is looking good! I have some broad comments on the installation process and how we can make that easier for users. Component logic looks good but I'll also take a closer look later!

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
dsherry: > If a user does pip install evalml but they didn't go through the steps to install prophet, they would get a weird error about the model not being compiled right?
> 
> Can we add some more statements like import_or_raise to make sure prophet was installed with the right backend? We may have to check that the prophet_model.bin file is present?

Our current install pattern is:
* Running `pip install evalml` installs everything from `core-requirements.txt` and the optional reqs in `requirements.txt`.
* We also support users manually installing only the stuff listed in `core-requirements.txt`. In which case, automl should not try to run xgboost/lightgbm/prophet, and users will get a clear error message if they try to run those estimators by hand.

I don't think this PR means we need to change that pattern. Prophet is another estimator like xgboost/lightgbm which we list as an "optional" dependency. Right?

freddyaboulton: @dsherry I think the one difference with prophet and xgboost is that if users do `pip install evalml` and then train/score with an xgboost estimator, nothing will error.

But if they do `pip install evalml` and then train/score with prophet, they will get an error because `pip install evalml` will not set up the `cmdstan` backend. To make it confusing, prophet is installed! So the `import_or_raise("prophet")` calls will not catch this. 

Given that prophet requires an extra step to properly set up the stan backend, my hope with using `extras_require` is that it won't install prophet by default and kind of force users to read our docs to find out how they should properly set up prophet. I think this is worth doing even if we get our `stan_ext` package off the ground, because users will still have to set up some env variables to properly tell prophet where the stan binaries are.

dsherry: Ah!!! Great point.

In that case, yes, I support this. Let's have clear install documentation for how to get prophet to work, clean behavior in automl and in the prophet component for our default installation (`pip install evalml`), and then also an extras install path to get prophet set up correctly. Right?

FWIW I don't think that blocks merging this PR, but I agree that we should get a fix merged.

freddyaboulton: Sounds good! Agree we can do it in a follow up issue.

.github/workflows/linux_unit_tests_with_latest_deps.yml
ParthivNaresh: Good call, hopefully we should be able to remove this once we can get the cmdstan-ext wheel up for an easy pip install

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
ParthivNaresh: So we need this to be specified so we can unit test it. Unfortunately the parameter doesn't get included in the attributes list unless it's passed into the Prophet component

.github/workflows/linux_unit_tests_with_latest_deps.yml
```diff
@@ -65,14 +65,34 @@ jobs:
...
+          python "${PYTHON_LIB}/cmdstanpy/install_cmdstan.py" -d "${PYTHON_LIB}"
+          export CMDSTAN="${CMDSTAN_LIB}"
+          export STAN_BACKEND=CMDSTANPY
+          pip install prophet
+          make installdeps
+          make installdeps
+          make installdeps-test
+          pip uninstall pystan -y
+          pip freeze
+      - if: ${{ !matrix.core_dependencies && (matrix.command == 'git-test-modelunderstanding' || matrix.command == 'git-test-dask') }}
```
ParthivNaresh, line 86: Don't install prophet for non-core dependencies and if the tests are `git-test-modelunderstanding` and `git-test-dask` because they don't need it

.github/workflows/linux_unit_tests_with_latest_deps.yml
ParthivNaresh: Update on this: we're not including Prophet in requirements because it's being installed as an extra-requirement in `setup.py`.  Therefore we'll be including the entire installation process for `Prophet` for `git-test-automl` and `git-test-other`.

.github/workflows/linux_unit_tests_with_minimum_deps.yml
```diff
@@ -35,7 +35,23 @@ jobs:
           virtualenv test_python -q
           source test_python/bin/activate
           python -m pip install --upgrade pip -q
-      - name: install evalml with test dependencies, core dependencies, & optional requirements
+      - if: ${{ matrix.command == 'git-test-automl' || matrix.command == 'git-test-other' }}
+        name: Install evalml with test dependencies, core dependencies, & optional requirements (Prophet)
```
ParthivNaresh, line 39: Include `Prophet` for `git-test-automl` and `git-test-other`.

chukarsten (approved):
Nice work, Parthiv!  I think that addressing @freddyaboulton 's concerns about the tests taking 15s each is very important to not bloating the CI time.  If you can reduce those and ping me with the results, I'd love to follow up.  Additionally, with respect to the suppression of stdout, I went through the ref and it seemed like some people had issues crop up with "running out of file descriptors" or something of the like.  For the sake of getting the Prophet regressor in, I'm willing to accept the risk of some weirdness here, but I'd like an issue filed to try and address this differently post-merge as I think this has the potential of being tech debt with a high interest rate, ala the weirdness going on with the matplotlib pngs.

.github/workflows/linux_unit_tests_with_latest_deps.yml
```diff
@@ -65,14 +65,34 @@ jobs:
...
+          source test_python/bin/activate
+          pip install cmdstanpy==0.9.68
+          PYTHON_LIB="/home/runner/work/evalml/evalml/test_python/lib/python${{ matrix.python_version }}/site-packages"
+          CMDSTAN_LIB="${PYTHON_LIB}/cmdstan-2.27.0"
+          python "${PYTHON_LIB}/cmdstanpy/install_cmdstan.py" -d "${PYTHON_LIB}"
+          export CMDSTAN="${CMDSTAN_LIB}"
+          export STAN_BACKEND=CMDSTANPY
+          pip install prophet
+          make installdeps
+          make installdeps
```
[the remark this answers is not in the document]
ParthivNaresh: Absolutely not!

evalml/utils/gen_utils.py
ParthivNaresh: I removed this bit, it seems like it was PyStan that was logging all of that output and we're relying on Cmdstanpy

evalml/tests/component_tests/test_prophet_regressor.py
ParthivNaresh: @freddyaboulton @chukarsten By changing the default parameters for `Prophet` to
`prophet.Prophet(uncertainty_samples=False, changepoint_prior_scale=2.0)`, 
the new times are
```
=========================================== slowest durations ===========================================
7.72s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_with_y_index
6.34s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_no_X
4.25s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_date_col
3.85s call     evalml/tests/component_tests/test_prophet_regressor.py::test_feature_importance
3.26s call     evalml/tests/component_tests/test_prophet_regressor.py::test_fit_predict_ts_with_X_index
```

setup.py
```diff
@@ -1,10 +1,12 @@
+from subprocess import call
 from setuptools import find_packages, setup
 
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
 extras_require = {
     'update_checker': ['alteryx-open-src-update-checker >= 2.0.0'],
+    'prophet': ['cmdstan-builder == 0.0.3']
```
ParthivNaresh, line 9: This links to the `cmdstan-builder` pip package.
The link for it on GitHub is [here](https://github.com/alteryx/cmdstan_ext), and the PyPI package is [here](https://pypi.org/project/cmdstan-builder/).
To learn more about the thought process behind this, please take a look at [this](https://alteryx.atlassian.net/wiki/spaces/PS/pages/551944193/Prophet+Installation), [this](https://alteryx.atlassian.net/wiki/spaces/PS/pages/901644585/Interaction+between+Stan+wrappers+and+Prophet), [this](https://alteryx.atlassian.net/wiki/spaces/PS/pages/884441406/Cmdstan+Installation+Flows), [this](https://alteryx.atlassian.net/wiki/spaces/PS/pages/629899414/Alternate+repository+for+cmdstan+building+and+publishing+a+wheel+for+EvalML+to+use), [this](https://alteryx.atlassian.net/wiki/spaces/PS/pages/561086672/Pip+installation+of+cmdstan) and [this](https://alteryx.quip.com/aBgLAxP5WoiE/Licensing-Prophet).

setup.py
```diff
@@ -31,3 +33,5 @@
                                        'evalml/tests/data/churn.csv',
                                        'evalml/tests/data/fraud_transactions.csv.gz'])],
 )
+
+call("pip uninstall pystan -y", shell=True)
```
ParthivNaresh, line 37: Uninstall `PyStan` to avoid licensing issues by mistake

freddyaboulton: I think this line belongs in `cmdstan-builder` as opposed to EvalML.

I commented out this line and added these lines:
```python
try:
    import pystan
    assert False, "pystan installed!"
except ModuleNotFoundError:
    print("pystan not installed!")
```

When I run
`python setup.py bdist_wheel` I see "pystan not installed!". Doing `pip freeze` verifies this. 

When I run `pip install dist/evalml-0.29.0-py3-none-any.whl`, `pystan` is also not installed.

However, when I run `pip install dist/evalml-0.29.0-py3-none-any.whl['prophet']`, pystan is installed!

image

evalml/utils/gen_utils.py
```diff
@@ -207,7 +209,7 @@ def get_importable_subclasses(base_class, used_in_automl=True):
         try:
             cls()
             classes.append(cls)
-        except (ImportError, MissingComponentError, TypeError):
+        except (ImportError, MissingComponentError, TypeError, ValueError):
```
freddyaboulton, line 212: What about Prophet would raise a ValueError when trying to instantiate the class? I think `ValueErrors` are pretty broad so I'm wondering if we can make the exception "narrower".

evalml/tests/utils_tests/test_dependencies.py
```diff
@@ -26,8 +29,11 @@ def test_has_minimal_deps(
     extra_deps += ["plotly.graph_objects"]
     for module in extra_deps:
         # We don't expect to install sktime in python 3.9. Let's verify it's not present:
-        if (module == "sktime" and is_running_py_39_or_above) or (
-            module == "pmdarima" and is_using_conda
+        if (
+            (module == "sktime" and is_running_py_39_or_above)
+            or (module == "pmdarima" and is_using_conda)
+            or (module == "cmdstanpy" and (is_using_windows or is_using_conda))
```
freddyaboulton, line 35: Do we need the extra cases for "cmdstanpy" and "prophet"? Those are not listed in the requirements.txt right?

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Prophet is a procedure for forecasting time series data based on an additive model where non-linear trends are fit with yearly, weekly, and daily seasonality, plus holiday effects.
+    It works best with time series that have strong seasonal effects and several seasons of historical data. Prophet is robust to missing data and shifts in the trend, and typically handles outliers well.
+
+    More information here: https://facebook.github.io/prophet/
+
+    """
+
+    name = "Prophet Regressor"
+    hyperparameter_ranges = {
```
freddyaboulton, line 24: Can you add the docstrings so they show up in the documentation?

evalml/tests/component_tests/test_components.py
```diff
@@ -1297,6 +1312,8 @@ def test_estimators_accept_all_kwargs(
         )
     if estimator_class.model_family == ModelFamily.ENSEMBLE:
         params = estimator.parameters
+    elif estimator_class.model_family == ModelFamily.PROPHET:
+        params = estimator.get_params()
```
freddyaboulton, line 1316: Would using `estimator.parameters` work for prophet?

ParthivNaresh: Yes that would work. I'm testing this in particular because Prophet has its own `_parameters` output that covers a wide range of parameters being passed into the component.

evalml/utils/gen_utils.py
ParthivNaresh: Hmm I think this was a different issue that's no longer applicable

evalml/tests/utils_tests/test_dependencies.py
ParthivNaresh: Oops these were leftover from when they were in requirements

.github/workflows/linux_unit_tests_with_minimum_deps.yml
ParthivNaresh: Great idea

bchen1116 (approved):
This looks great! Thanks for explaining how things work in the writeups/comments/calls! Glad we're able to add this to EvalML!

PR #2244, opened 2021-05-10
Address Sphinx 4.0 upgrade

bchen1116:
Sphinx v4.0.0 removes support for `add_javascript` and `add_stylesheet` and replaces them with other calls. 

We cap the version to <4.0.0 to avoid issues with [cross-referencing](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/629821/).  (depreciated API docs [here](https://www.sphinx-doc.org/ar/master/extdev/deprecated.html?highlight=add_javascript#deprecated-apis))


bchen1116:
New issue to unpin Sphinx [here](https://github.com/alteryx/evalml/issues/2246).

This PR will just cap Sphinx version <4.0.0 for now.

jeremyliweishih (approved):
looks good! Might need @chukarsten or @dsherry to merge in the dependency PR first so the dependency test passes here.

freddyaboulton (approved):
Thanks @bchen1116 !

angela97lin (approved):
LGTM!

Also filed https://github.com/alteryx/evalml/issues/2248 so we can automatically catch stuff like this in the future :)

ParthivNaresh (approved):

freddyaboulton (approved):
Thanks so much @bchen1116 !

evalml/tests/utils_tests/test_dependencies.py
```diff
@@ -5,12 +5,20 @@
 import requirements
 
 
+def _get_req_name(name):
+    if name == 'imbalanced-learn':
+        return 'imblearn'
+    elif name == 'MarkupSafe':
+        return 'markupsafe'
```
dsherry, line 12: Damn that's tough. Nice patch though.

requirements.txt
```diff
@@ -5,6 +5,7 @@ ipywidgets>=7.5
 xgboost>=0.82,<1.3.0
 catboost>=0.20
 lightgbm>=2.3.1,<3.1.0
+MarkupSafe==1.1.1
```
dsherry, line 8: @bchen1116 why add this to our requirements? Shouldn't this only be in our docs requirements?

dsherry (approved):
@bchen1116 thank you!

I approved so that we can unblock `main` builds. But I do want to understand why we have to add `MarkupSafe` to `requirements.txt`. I thought its only required to build the docs, not required to run evalml.

requirements.txt
bchen1116: When I just added this to the doc requirements, we still had failing tests. I'm not sure why it happened, but adding it to requirements solved the issue

freddyaboulton: RTD installs the docs requirements then evalml. evalml installs `MarkUpSafe` via `ipywidgets` so if we pin markupsafe in docs requirements, it'll still be overwritten when RTD installs evalml

dsherry: @freddyaboulton understood

[Here](https://github.com/alteryx/evalml/blob/main/.readthedocs.yaml#L21) is where we configure what gets installed in our RTD builds. I wonder if we now need another requirements file, which gets installed _after_ evalml is installed, and we can move the markupsafe pin there. I'm trying that out now in a draft PR #2261

PR #2253, opened 2021-05-11
Remove excess data splitter calls

bchen1116:
Noticed that my previous PR removing the sampling data splitter didn't remove some extra lines that called the data splitter.

Since we are now defaulting to StratifiedKFold and no longer have data splitters with a `transform_sample` method, I removed the excess code


jeremyliweishih (approved):
looks good!

angela97lin (commented):
Did codecov not catch this, if we don't have anything that calls `transform_sample`?

PR #2254, opened 2021-05-11
Set minimum required version for for pyzmq, colorama, and docutils

gsheni:
- Set minimum required version for for pyzmq, colorama, and docutils
- Pre-req to use minimum dependency generator


angela97lin (approved):
Looks good, thank you!!

freddyaboulton (approved):
Thanks @gsheni !

PR #2261, opened 2021-05-12
Docs build: refactor markupsafe dependency pin to be outside of requirements.txt

dsherry:
Follow-up from #2244 .

Since `MarkupSafe` is only needed for the docs, and is not needed to run evalml, it shouldn't be listed in our `requirements.txt`. So this PR moves it out of there.

Thanks to @rwedge for the ultimate solution!

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -22,6 +22,6 @@ python:
   version: 3.8
   install:
     - requirements: docs/readthedocs-requirements.txt
-    - requirements: docs-requirements.txt
     - method: pip
       path: .
+    - requirements: docs-requirements.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ ipywidgets>=7.5
 xgboost>=0.82,<1.3.0
 catboost>=0.20
 lightgbm>=2.3.1,<3.1.0
-MarkupSafe==1.1.1
 matplotlib>=3.3.3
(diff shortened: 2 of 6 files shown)


.readthedocs.yaml
```diff
@@ -22,6 +22,6 @@ python:
   version: 3.8
   install:
     - requirements: docs/readthedocs-requirements.txt
-    - requirements: docs-requirements.txt
     - method: pip
       path: .
+    - requirements: docs-requirements.txt
```
dsherry, line 27: This is what did it

requirements.txt
```diff
@@ -5,7 +5,6 @@ ipywidgets>=7.5
 xgboost>=0.82,<1.3.0
 catboost>=0.20
 lightgbm>=2.3.1,<3.1.0
-MarkupSafe==1.1.1
```
dsherry, line 8: 🥂

.readthedocs.yaml
rwedge: It should be ok to remove the MarkupSafe requirement from `docs-requirements` as well, since `nbsphinx>=0.8.5`, which is present in the docs requirements, also addresses the Markupsafe issue in its own requirements

dsherry: Ah thanks, will do

rwedge (approved):
looks good, assuming things pass

PR #2263, opened 2021-05-13
Update AutoMLSearch init Pipeline logging

### LINKED ISSUE #2092: Update logged information for AutoML init
Minor user experience nit-pick, but there have been several times when I just instantiate AutoMLSearch and forget to call `search`. The output of init looks like:
```
automl = AutoMLSearch(X, y, problem_type="binary")
```

```
Using default limit of max_batches=1.

Generating pipelines to search over...
```

As a user, "Generating pipelines to search over..." makes me wait in anticipation for something to happen. However, by the time init has returned, the pipelines have already been generated. Hence, we should update logged message to something else, such as "Generated pipelines to search over." or just remove it entirely.

ParthivNaresh:
Fixes #2092


evalml/automl/automl_search.py
```diff
@@ -383,6 +383,8 @@ def __init__(self,
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
+        else:
```
dsherry, line 386: Nit-pick: this `else` is not necessary, since the `if` raises an error

evalml/automl/automl_search.py
```diff
@@ -383,6 +383,8 @@ def __init__(self,
...
+            logger.info("Pipelines generated.")
```
dsherry, line 387: Is there anything else we can tell people here? Perhaps `f"{len(self.allowed_pipelines)} pipelines ready for search."`

dsherry (approved):
Cool! Left two nit picks

evalml/automl/automl_search.py
```diff
@@ -383,6 +383,8 @@ def __init__(self,
 
         if self.allowed_pipelines == []:
             raise ValueError("No allowed pipelines to search")
+
+        logger.info(f"{len(self.allowed_pipelines)} pipelines ready for search.")
```
bchen1116, line 387: nit, but can we add simple test coverage for this just to make sure the output is expected for diff numbers of pipelines/iterations/model families?

bchen1116 (approved):
LGTM! I think it'd be nice to add some simple test coverage, left a comment on suggestions but not necessary to implement all! I think just one test to check the output would be good

chukarsten (approved):
LGTM

PR #2266, opened 2021-05-13
Change number of cores for pytest from 4 to 2

gsheni:
- Change number of cores for pytest from 4 to 2


dsherry (approved):
LGTM, thanks. Did this because GH actions workers have 2 cores, whereas our old circleci builders had more. Looking at the unit test CI in this PR, I don't see a significant change in runtime vs current `main`, which is as expected.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

This reminds me that I think we're at the point where we're gonna need to have multiple unit test jobs which run in parallel!

PR #2267, opened 2021-05-13
Add minimum dependency generator GitHub Action

gsheni:
- Add a CI workflow to generate the minimum dependencies
- A correct PR was generated, as seen here:
  -  https://github.com/alteryx/evalml/pull/2268

### DIFF
--- a/.github/workflows/minimum_dependency_checker.yml
+++ b/.github/workflows/minimum_dependency_checker.yml
@@ -1,7 +1,5 @@
 name: Minimum Dependency Checker
 on:
-  pull_request:
-    types: [opened, synchronize]
   push:
     branches:
       - main
--- a/.github/workflows/minimum_dependency_checker.yml
+++ b/.github/workflows/minimum_dependency_checker.yml
@@ -1,5 +1,7 @@
 name: Minimum Dependency Checker
 on:
+  pull_request:
+    types: [opened, synchronize]
   push:
     branches:
       - main
--- a/.github/workflows/minimum_dependency_checker.yml
+++ b/.github/workflows/minimum_dependency_checker.yml
@@ -1,0 +1,77 @@
+name: Minimum Dependency Checker
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'requirements.txt'
+      - 'test-requirements.txt'
+      - 'core-requirements.txt'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Run min dep generator - test reqs
+        id: min_dep_gen_test
+        uses: alteryx/minimum-dependency-generator@v2
+        with:
+          requirements_paths: 'test-requirements.txt'
+      - name: Save min test deps and run diff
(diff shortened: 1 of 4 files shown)


.github/workflows/release_notes_updated.yml
```diff
@@ -28,7 +30,9 @@ jobs:
         run: |
           if [[ $(expr match "${{ github.event.pull_request.head.ref }}" "release_v[0-9.]\+") -gt 0 ]]; then
             exit 0;
-          elif [[ $(expr match "${{ github.event.pull_request.head.ref }}" "dep-update-[a-zA-Z0-9]*") -gt 0 ]]; then
+          elif [[ $(expr match "${{ github.event.pull_request.head.ref }}" "latest-dep-update-[a-zA-Z0-9]*") -gt 0 ]]; then
```
gsheni, line 33: Updated the release notes check to skip latest and minimum dependency branches

.github/workflows/minimum_dependency_checker.yml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Run min dep generator - test requirements
+        id: min_dep_gen_test
+        uses: alteryx/minimum-dependency-generator@v2
```
freddyaboulton, line 21: Does this action always list the packages in the same order? I ask cause in #2268, I notice that the packages are not listed in alphabetical order and `diff` cares about line order

freddyaboulton (approved):
@gsheni I think this looks good!

.github/workflows/minimum_dependency_checker.yml
gsheni: yes, the packages are always in the same order. I can update `minimum-dependency-generator` to generate them in alphabetical order if you want (thought it shouldn't block this MR).

freddyaboulton: No need, I was just wondering!

.github/workflows/minimum_dependency_checker.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: if min requirements have changed, write output file
+        if: steps.check_min_reqs.outcome != 'success'
+        run: |
+          printf "${{ steps.min_dep_gen_reqs.outputs.min_reqs }}" > evalml/tests/dependency_update_check/minimum_requirements.txt
+      - name: Run min dep generator - core requirements
+        id: min_dep_gen_core
+        uses: alteryx/minimum-dependency-generator@v2
+        with:
+          requirements_paths: 'core-requirements.txt'
+      - name: Save min core requirements and run diff
```
dsherry, line 55: There's a way in GH actions to define a "function" which could accept an arg like the requirements file name, and then execute the diff and file write, and then we could call that "function" 3x instead of copy-pasting the code, right? Not required for this PR lol just curious

.github/workflows/release_notes_updated.yml
dsherry: Good call
There's probably a fancy unix way to do this in one `if` check hahaha but I don't know what it is right away at least

dsherry (approved):
@gsheni thank you for adding this! I love it.

PR #2271, opened 2021-05-13
Dictionary support for Undersampler

### LINKED ISSUE #2105: Undersampler accept dictionary as sampling argument
Allow the Undersampler component to accept a dictionary argument. Original comment [here](https://github.com/alteryx/evalml/pull/2079#discussion_r608098815)

bchen1116:
fix #2105


evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -59,6 +59,49 @@ def transform(self, X, y=None):
             y = infer_feature_types(y)
         return X, y
 
+    def _convert_dictionary(self, sampling_dict, y):
+        """Expects the dictionary keys to be the target values y.
```
chukarsten, line 63: What do you mean by this?  Function description is a little unclear.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -59,6 +59,49 @@ def transform(self, X, y=None):
...
+
+        Returns:
+            dict: A dictionary with target values as keys and the number of samples as values
+        """
+        # check that the lengths of the dict and y are equal
+        y_unique = y.unique()
+        if len(sampling_dict) != len(y_unique):
+            raise ValueError("Lengths are different!")
+
+        if len(set(sampling_dict.keys()).intersection(set(y_unique))) != len(y_unique):
```
chukarsten, line 77: Nice.  I love Python sets and when people use them.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -59,6 +59,49 @@ def transform(self, X, y=None):
...
+            sampling_dict (dict): The input sampling dictionary passed in from user
+            y (ww.DataColumn): The target values
+
+        Returns:
+            dict: The parameters dictionary with the sampling_ratio_dict value replaced as necessary
+        """
+        param_copy = copy.copy(self.parameters)
+        if self.parameters['sampling_ratio_dict']:
+            new_dic = self._convert_dictionary(self.parameters['sampling_ratio_dict'], y)
+            param_copy['sampling_ratio_dict'] = new_dic
```
chukarsten, line 102: Placeholder thought: what happens if the sampling_ratio_dict is not defined?  Is it the desired behavior to add the key to the dictionary and assign it the new_dic value?

evalml/tests/component_tests/test_undersampler.py
```diff
@@ -74,3 +75,55 @@ def test_undersample_imbalanced(data_type, make_data_type):
...
+    undersampler = Undersampler(sampling_ratio_dict=dictionary)
+    with pytest.raises(ValueError, match=msg):
+        undersampler.fit_transform(X, y)
+
+
+@pytest.mark.parametrize("dictionary,dictionary_values", [({0: 1, 1: 0.5}, {0: 150, 1: 300}),
+                                                          ({0: 1, 1: 0.25}, {0: 150, 1: 600}),
+                                                          ({0: 1, 1: 0.1}, {0: 150, 1: 850}),
+                                                          ({0: 0.1, 1: 0.1}, {0: 150, 1: 850}),
+                                                          ({0: 0.1, 1: 1}, {0: 150, 1: 150})])
```
chukarsten, line 97: Can you explain these tests to me?

evalml/tests/component_tests/test_undersampler.py
```diff
@@ -74,3 +75,55 @@ def test_undersample_imbalanced(data_type, make_data_type):
...
+
+    assert len(new_X) == sum(expected_result.values())
+    assert new_y.to_series().value_counts().to_dict() == expected_result
+
+
+def test_undersampler_sampling_dict_strings():
+    X = np.array([[i] for i in range(1000)])
+    y = np.array(["minority"] * 150 + ["majority"] * 850)
+    dictionary = {"minority": 1, "majority": 0.5}
+    expected_result = {"minority": 150, "majority": 300}
```
chukarsten, line 124: this is cool. I like this functionality.

chukarsten (changes requested):
This is good work, Bryan.  I think I need some chat with what this means to have a ratio for both the majority and minority class.  That's a little confusing to me now that I'm looking at it.  I think we also need to drop an update to the docstring and figure out what _dictionary_to_params() should be doing without 'sampling_ratio_dict' defined in the params.

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: Updating the docstring! 

`Converts the provided sampling dictionary from a dictionary of ratios to a dictionary of number of samples.
        Expects the provided dictionary keys to be the target values y, and the associated values to be the min:max ratios.
        Converts and returns a dictionary with the same keys, but changes the values to be the number of samples rather than ratio.`

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: If there is no `sampling_ratio_dict` defined, we simply return the parameters we already have. We only do a conversion if the dic is passed in to the component in the init.

evalml/tests/component_tests/test_undersampler.py
bchen1116: Yep! The first dictionary is the ratio dict that a user will pass in, while the second dictionary is the resulting dictionary we want to post-conversion. The first one holds the ratios, while the second one has the number of samples that we want in the final result.

Renamed the dictionaries for clarity!

chukarsten (approved):
Solid work!  Ship it!

evalml/tests/component_tests/test_undersampler.py
```diff
@@ -74,3 +75,55 @@ def test_undersample_imbalanced(data_type, make_data_type):
 
     np.testing.assert_equal(X, transform_X.to_dataframe().values)
     np.testing.assert_equal(None, transform_y)
+
```
freddyaboulton, line 78: Can you add coverage the random seed is preserved when users pass in `sampling_dict`? Your implementation does that now but just suggesting so that we preserve that behavior during a refactor, especially since we init the component during `fit` and not `__init__`

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -59,6 +59,52 @@ def transform(self, X, y=None):
             y = infer_feature_types(y)
         return X, None
 
+    def _convert_dictionary(self, sampling_dict, y):
```
freddyaboulton, line 62: Should this method live in the base class? It would be different for undersamplers vs oversampler right?

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -59,6 +59,52 @@ def transform(self, X, y=None):
...
+        Arguments:
+            sampling_dict (dict): The input sampling dictionary passed in from user
+            y (ww.DataColumn): The target values
+
+        Returns:
+            dict: A dictionary with target values as keys and the number of samples as values
+        """
+        # check that the lengths of the dict and y are equal
+        y_unique = y.unique()
+        if len(sampling_dict) != len(y_unique):
```
freddyaboulton, line 76: I think the below check will catch this case?

freddyaboulton (approved):

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: I plan on doing an if-statement in the base class to handle the differences between over and undersamplers! Should be a one-line difference

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: Yeah, I wanted to include both since they are checking for slightly different things. Do you think this first case is unnecessary?

freddyaboulton: I see! Can we make the error message more detailed for the first case then?, e.g. sampling_dict contains different number of target values than are in the data or something like that?

freddyaboulton: Suggesting because if we keep the first case, I feel the error message should be more actionable than "Lengths are different"

bchen1116: Done!

freddyaboulton: 🚢

Issue #2285, opened 2021-05-17
Prediction explanations fail for catboost on titanic/fraud dataset


freddyaboulton:
Repro for fraud dataset because it's included in evalml but the same error happens for titanic:

```python
from evalml.demos import load_fraud
from evalml.pipelines import BinaryClassificationPipeline
from evalml.model_understanding import explain_predictions_best_worst
import pytest

X, y = load_fraud(1000)

catboost = BinaryClassificationPipeline(["Imputer", "DateTime Featurization Component", 
                                         "Text Featurization Component", "CatBoost Classifier"])
catboost.fit(X, y)
with pytest.raises(TypeError, match="Cannot convert IntegerArray to numpy.ndarray"):
    explain_predictions_best_worst(catboost, X, y)
```

The problem is that there are pandas-`Int64` typed features in the dataframe and we [do not convert these features](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/prediction_explanations/_algorithms.py#L57) to numpy `int64` for catboost pipelines. That is an oversight on our part but I've tested the same code on my feature branch for the woodwork update and we no longer see the error. 

So this will be fixed by https://github.com/alteryx/evalml/pull/2181

Issue #2281, opened 2021-05-17
SHAP test fails with the Elastic Net Classifier estimator


bchen1116:
In [this PR](https://github.com/alteryx/evalml/pull/2269), we change the init params for ENC in order to pass the shap tests (test_shap) in the `test_algoritthms.py` file. Leaving the init params as `alpha = 0.0001` and `l1_ratio=0.15` caused a `ZeroDivisionError` while calculating shap, which is likely linked to [this](https://github.com/slundberg/shap/issues/183). 

Filing this issue to determine why the test fails and figure out a good way to avoid this error.

freddyaboulton:
I think this is a shap issue. I filed https://github.com/slundberg/shap/issues/2000 for discussion. In the short term, I think we can do the following:

* Use link="identity" in the `KernelExplainer` for linear models
* Use [LinearExplainer](https://shap.readthedocs.io/en/latest/generated/shap.explainers.Linear.html#) with logit link, `explainer = shap.LinearExplainer(classifier, X, link=shap.links.logit)`

bchen1116:
A further test that failed:

On main, using `alpha=0.5, l1_ratio=0.5`, which is what we have prior to [this](https://github.com/alteryx/evalml/pull/2269), if we add this test into `evalml/tests/model_understanding_tests/prediction_explaination_tests/test_explainers.py`:

```python
@pytest.mark.parametrize("estimator", ["Extra Trees Classifier", "Elastic Net Classifier"])
def test_elastic_net(estimator, fraud_100):
    pytest.importorskip('imblearn', reason='Skipping test because imblearn not installed')
    X, y = fraud_100
    pipeline = BinaryClassificationPipeline(component_graph=["Imputer", "One Hot Encoder", "DateTime Featurization Component", estimator])
    pipeline.fit(X=X, y=y)
    pipeline.predict(X)
    importance = explain_predictions(pipeline, X, y, indices_to_explain=[0], top_k_features=4)
    assert report['feature_names'].isnull().sum() == 0
    assert report['feature_values'].isnull().sum() == 0

```

The test fails:
![image](https://user-images.githubusercontent.com/22552445/118707697-d5ab4880-b7e8-11eb-8d92-f70824d36d8e.png)


If we change the alpha and l1_ratio, it still fails.

chukarsten:
Ok @bchen1116 , go do it.

bchen1116:
Closing with [this PR](https://github.com/alteryx/evalml/pull/2420)

PR #2213, opened 2021-04-30
Add Oversampling Components to AutoMLSearch

bchen1116:
fix #2098 
Quip doc [here](https://alteryx.quip.com/TW2nANznZcpN/Oversampling-in-AutoML-Search)
perf test [part 1](https://alteryx.quip.com/0f9GAnwHxYnO/Sampler-Components-Perf-Test-Results-Part-1) and [part 2](https://alteryx.quip.com/2ARSAhbxbqou/Sampler-Component-Perf-Tests-Part-2)

This PR will be merged after adding the [Undersampler to AutoMLSearch](https://github.com/alteryx/evalml/pull/2128)

PR #2286, opened 2021-05-18
Add Model Understanding Tests

bchen1116:
Address comments from [this](https://github.com/alteryx/evalml/pull/2213#issuecomment-843236012)

We can't do `prediction_explanations` for Catboost or Elastic Net Classifier, and we already have issues related to the bugs [here](https://github.com/alteryx/evalml/issues/2285) and [here](https://github.com/alteryx/evalml/issues/2281)


evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -823,3 +823,14 @@ def test_explain_predictions_stacked_ensemble(problem_type, dummy_stacked_ensemb
 
     with pytest.raises(ValueError, match="Cannot explain predictions for a stacked ensemble pipeline"):
         explain_predictions_best_worst(pipeline, X, y)
+
+
+@pytest.mark.parametrize("estimator", ["Logistic Regression Classifier", "Decision Tree Classifier", "Extra Trees Classifier"])
+def test_explain_predictions_oversampler(estimator, fraud_100):
+    pytest.importorskip('imblearn.over_sampling', reason='Skipping test because imbalanced-learn not installed')
+    X, y = fraud_100
+    pipeline = BinaryClassificationPipeline(component_graph=["Imputer", "One Hot Encoder", "DateTime Featurization Component", estimator])
```
dsherry, line 832: @bchen1116 there's no oversampler in this pipeline! I'm guessing you forgot to add it? The other test looks 💯 🔥  haha

docs/source/release_notes.rst
```diff
@@ -2,6 +2,8 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
+        * Added oversamplers to AutoMLSearch :pr:`2213`
+        * Added tests for model understanding for oversamplers :pr:`2286`
```
dsherry, line 6: Mega nit-pick: you could lump these into one line if you want and our release notes job will still pass:
```
* Added oversamplers to AutoMLSearch :pr:`2213` :pr:`2286`
```

dsherry (changes requested):
You rock, thanks for this! This will cover our butts while we work on adding coverage for automl --> model understanding to our release test suite.

I just had one blocking comment.

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -823,3 +823,14 @@ def test_explain_predictions_stacked_ensemble(problem_type, dummy_stacked_ensemb
 
     with pytest.raises(ValueError, match="Cannot explain predictions for a stacked ensemble pipeline"):
         explain_predictions_best_worst(pipeline, X, y)
+
+
+@pytest.mark.parametrize("estimator", ["Logistic Regression Classifier", "Decision Tree Classifier", "Extra Trees Classifier"])
```
angela97lin, line 828: Rather than specify the estimators to test, it could be a good idea to pass in all of the estimators using all_estimators and then skipping the estimators that don't work :)

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -306,3 +306,14 @@ def test_undersampler(X_y_binary):
     pipeline.predict(X)
     test = calculate_permutation_importance(pipeline, X, y, objective="Log Loss Binary")
     assert test is not None
+
+
+@pytest.mark.parametrize("estimator", ["Logistic Regression Classifier", "Decision Tree Classifier", "Extra Trees Classifier"])
```
angela97lin, line 311: Same with this test :))

angela97lin (commented):
Looks good besides what @dsherry pointed out, but I think it could be worthwhile to try to all estimators and single out catboost / EN. Then, when those issues are resolved it'll be easier to catch here and update the tests accordingly.

ParthivNaresh (approved):
Looks great, no other comments besides the changes already requested!

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
bchen1116: Wow that's terrible. I'll fix!

dsherry: Lol its ok! This is why I love code review 😁

evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -823,3 +824,14 @@ def test_explain_predictions_stacked_ensemble(problem_type, dummy_stacked_ensemb
...
+
+@pytest.mark.parametrize("estimator", [e for e in _all_estimators() if ('Classifier' in e.name and not any(s in e.name for s in ["Baseline", "Cat", "Elastic", "KN", "Ensemble"]))])
+def test_explain_predictions_oversampler(estimator, fraud_100):
+    pytest.importorskip('imblearn.over_sampling', reason='Skipping test because imbalanced-learn not installed')
+    X, y = fraud_100
+    pipeline = BinaryClassificationPipeline(component_graph=["Imputer", "One Hot Encoder", "DateTime Featurization Component", "SMOTENC Oversampler", estimator])
+    pipeline.fit(X, y)
+    report = explain_predictions(pipeline, X, y, indices_to_explain=[0], output_format="dataframe", top_k_features=4)
+    assert report['feature_names'].isnull().sum() == 0
+    assert report['feature_values'].isnull().sum() == 0
```
dsherry, line 837: Yep this seems reasonable to me. We have other tests I believe which check that the output of prediction explanations is what we expect it to be. This test is just checking that something came out which matches the form we expect. 👍

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -306,3 +307,14 @@ def test_undersampler(X_y_binary):
...
+
+
+@pytest.mark.parametrize("estimator", [e for e in _all_estimators() if ('Classifier' in e.name and not any(s in e.name for s in ["Baseline", "Cat", "Elastic", "KN", "Ensemble"]))])
+def test_permutation_importance_oversampler(estimator, fraud_100):
+    pytest.importorskip('imblearn.over_sampling', reason='Skipping test because imbalanced-learn not installed')
+    X, y = fraud_100
+    pipeline = BinaryClassificationPipeline(component_graph=["Imputer", "One Hot Encoder", "DateTime Featurization Component", "SMOTENC Oversampler", estimator])
+    pipeline.fit(X=X, y=y)
+    pipeline.predict(X)
+    importance = calculate_permutation_importance(pipeline, X, y, objective="Log Loss Binary")
```
dsherry, line 319: @bchen1116 this can be a long-running computation. How long does this unit test take locally?

If its longer than a couple seconds total I suggest just doing it for one estimator. IMO I think that's acceptable because the goal for these tests is to prove that using a component which modifies the target during training (oversampler) doesn't break the model understanding tools, so iterating over the estimators is nice but not a requirement.

dsherry (approved):
Looking good!

very_cool

evalml/tests/model_understanding_tests/test_permutation_importance.py
bchen1116: Yep! Running all locally took ~36 seconds, while running with one estimator took ~4.5 seconds. Will change to using just 1 estimator!

PR #2288, opened 2021-05-18
Add dictionary support for oversamplers

### LINKED ISSUE #2142: Design and create dictionary input for Oversampler
In the [oversampler PR](https://github.com/alteryx/evalml/pull/2079), @freddyaboulton brought up 2 good points on the dictionary input for sampling for the oversamplers, [this](https://github.com/alteryx/evalml/pull/2079#discussion_r612611517) and [this](https://github.com/alteryx/evalml/pull/2079#discussion_r612608475). In short, we want to figure out the best way to design the dictionary input to allow users to smoothly pass their desired sample numbers to the component while avoiding issues with label encoding and not knowing absolute numbers of samples. 

This issue tracks the design and implementation of a solution for the two points above. Would be good to implement with [this issue](https://github.com/alteryx/evalml/issues/2105) in mind.

bchen1116:
fix #2142


evalml/automl/automl_search.py
```diff
@@ -358,7 +358,12 @@ def __init__(self,
             self._sampler_name = self.sampler_method
             if self.sampler_method in ['auto', 'Oversampler']:
                 self._sampler_name = get_best_sampler_for_data(self.X_train, self.y_train, self.sampler_method, self.sampler_balanced_ratio)
-            self._frozen_pipeline_parameters[self._sampler_name] = {"sampling_ratio": self.sampler_balanced_ratio}
+            if self._sampler_name not in parameters:
+                parameters[self._sampler_name] = {"sampling_ratio": self.sampler_balanced_ratio}
+                self._frozen_pipeline_parameters[self._sampler_name] = {"sampling_ratio": self.sampler_balanced_ratio}
+            elif self._sampler_name in parameters:
+                parameters[self._sampler_name].update({"sampling_ratio": self.sampler_balanced_ratio})
```
bchen1116, line 365: Decided to default to the `sampler_balanced_ratio` value that was provided for AutoMLSearch to set the `sampling_ratio`, which means it overrides the `sampling_ratio` that a user might input to the pipeline_params via
```python
pipeline_parameters = {sampler : {"sampling_ratio": 0.5}}
```

Updated the docstring above to mention this

chukarsten (approved):
Nice Bryan, this is good work.  I think you and I discussed a bunch of things in slack parallel to this and I like the addition of docs and the test for the string keys.  I still personally think the binary case with the key/val in the dict for the majority class in the oversampler case or the minority class in the undersampler case is a little confusing that they don't do anything.    Personally, I just struggle with the dictionary representation, but I think that's a personal problem X-D.

evalml/automl/automl_search.py
ParthivNaresh: Very cool! I don't think you need an `elif` here, it can just be an `else` since I don't think we'd have another logic path but total nitpick.

Also I like the way you set the logic for the `elif` block. You could do the same to the `if` block and move setting the `self._frozen_pipeline_parameters` after:
```
if self._sampler_name not in parameters:
     parameters[self._sampler_name] = {"sampling_ratio": self.sampler_balanced_ratio}
else:
     parameters[self._sampler_name].update({"sampling_ratio": self.sampler_balanced_ratio})
self._frozen_pipeline_parameters[self._sampler_name] = parameters[self._sampler_name]
```

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -78,9 +78,14 @@ def _convert_dictionary(self, sampling_dict, y):
...
-            # total samples for that class
-            new_dic[k] = int(min(y_counts.values[-1] / v, y_counts[k]))
+            if self.__class__.__name__ == "Undersampler":
+                # for undersampling, we make sure we never sample more than the
+                # total samples for that class
+                new_dic[k] = int(min(y_counts.values[-1] / v, y_counts[k]))
+            else:
+                # for oversampling, we need to make sure we never sample less than
+                # the total samples for that class
+                new_dic[k] = int(max(y_counts.values[0] * v, y_counts[k]))
```
ParthivNaresh, line 88: Nicely done

ParthivNaresh (approved):
Great job with this, excellent test coverage

PR #2290, opened 2021-05-19
Update `start_iteration_callback` to accept pipeline instance instead of pipeline class

### LINKED ISSUE #2282: AutoML search: start_iteration_callback is passed pipeline class instead of pipeline instance
Left over from the epic pipeline API update in #2091.
 
I think we just need to change this to provide `pipeline` as the first arg to the callback instead of `pipeline.__class__`. This is because our pipeline classes no longer contain the component graph! The pipeline instance holds this now.

https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L461

Thanks @ColinRTaylor for sharing this with us.

angela97lin:
Closes #2282

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,6 +2,7 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
+        * Added oversamplers to AutoMLSearch :pr:`2213` :pr:`2286`
         * Added dictionary input functionality for ``Undersampler`` component :pr:`2271`
         * Changed the default parameter values for ``Elastic Net Classifier`` and ``Elastic Net Regressor`` :pr:`2269`
     * Fixes
@@ -46,7 +47,6 @@ Release Notes
         * Updated prediction explanations functions to allow pipelines with XGBoost estimators :pr:`2162`
         * Added partial dependence for datetime columns :pr:`2180`
         * Update precision-recall curve with positive label index argument, and fix for 2d predicted probabilities :pr:`2090`
-        * Added oversamplers to AutoMLSearch :pr:`2213`
         * Add pct_null_rows to ``HighlyNullDataCheck`` :pr:`2211`
         * Added a standalone AutoML `search` method for convenience, which runs data checks and then runs automl :pr:`2152`
         * Make the first batch of AutoML have a predefined order, with linear models first and complex models last :pr:`2223` :pr:`2225`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
         * Changed the default parameter values for ``Elastic Net Classifier`` and ``Elastic Net Regressor`` :pr:`2269`
     * Fixes
     * Changes
+        * Updated ``start_iteration_callback`` to accept a pipeline instance instead of a pipeline class :pr:`2290`
     * Documentation Changes
     * Testing Changes
         * Use codecov action to update coverage reports :pr:`2238`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,7 @@ Release Notes
         * Changed the default parameter values for ``Elastic Net Classifier`` and ``Elastic Net Regressor`` :pr:`2269`
     * Fixes
     * Changes
-        * Updated ``start_iteration_callback`` to accept a pipeline instance instead of a pipeline class :pr:`2290`
+        * Updated ``start_iteration_callback`` to accept a pipeline instance instead of a pipeline class and no longer accept pipeline parameters as a parameter :pr:`2290`
     * Documentation Changes
     * Testing Changes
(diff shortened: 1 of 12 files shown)


evalml/tests/automl_tests/test_automl_search_regression.py
```diff
@@ -220,35 +220,29 @@ def test_automl_allowed_pipelines_init_allowed_both_specified(mock_fit, mock_sco
...
-    assert start_iteration_callback.call_count == 2
-    assert start_iteration_callback.call_args_list[0][0][0] == RegressionPipeline
-    assert start_iteration_callback.call_args_list[1][0][0] == dummy_regression_pipeline_class
-
+    if is_linear:
+        pipeline_class = dummy_regression_pipeline_class
+    else:
+        pipeline_class = nonlinear_regression_pipeline_class
 
-def test_automl_regression_nonlinear_pipeline_search(nonlinear_regression_pipeline_class, X_y_regression):
```
angela97lin, line 240: Combined :)

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -574,51 +574,41 @@ def test_automl_allowed_pipelines_init_allowed_both_specified_multi(mock_fit, mo
...
+        expected_mock_class = BinaryClassificationPipeline
+        pipeline_class = dummy_binary_pipeline_class if is_linear else nonlinear_binary_pipeline_class
+    else:
+        X, y = X_y_multi
+        mock_multi_score.return_value = {'Log Loss Multiclass': 1.0}
+        expected_mock_class = MulticlassClassificationPipeline
 
+        pipeline_class = dummy_multiclass_pipeline_class if is_linear else nonlinear_multiclass_pipeline_class
 
-def test_automl_multiclass_nonlinear_pipeline_search_more_iterations(nonlinear_multiclass_pipeline_class, X_y_multi):
```
angela97lin, line 609: Kind of confused by what this test was testing (that multiclass with more iterations will produce the right pipelines..?), but combined with above :)

docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Release Notes
         * Changed the default parameter values for ``Elastic Net Classifier`` and ``Elastic Net Regressor`` :pr:`2269`
     * Fixes
     * Changes
+        * Updated ``start_iteration_callback`` to accept a pipeline instance instead of a pipeline class :pr:`2290`
```
dsherry, line 10: This is a bit of a nit-pick, but technically we should add a breaking change entry too, because this is going out in a different release from the pipeline API change.

evalml/tests/automl_tests/test_automl_search_classification.py
dsherry: Sounds good! Yeah looking at this now, I'm not sure either haha. It feels like it could be an automl algo test which we put in automl search before those concepts had separation. Idk!

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -574,51 +574,41 @@ def test_automl_allowed_pipelines_init_allowed_both_specified_multi(mock_fit, mo
...
+                          max_iterations=5, start_iteration_callback=start_iteration_callback,
+                          allowed_pipelines=allowed_pipelines)
     automl.search()
 
-    assert start_iteration_callback.call_args_list[0][0][0] == MulticlassClassificationPipeline
-    assert start_iteration_callback.call_args_list[1][0][0] == nonlinear_multiclass_pipeline_class
-    assert start_iteration_callback.call_args_list[4][0][0] == nonlinear_multiclass_pipeline_class
+    assert isinstance(start_iteration_callback.call_args_list[0][0][0], expected_mock_class)
+    for i in range(1, 5):
+        assert isinstance(start_iteration_callback.call_args_list[i][0][0], pipeline_class)
```
dsherry, line 611: 👍

evalml/automl/automl_search.py
```diff
@@ -450,7 +450,7 @@ def _get_batch_number(self):
 
     def _pre_evaluation_callback(self, pipeline):
         if self.start_iteration_callback:
-            self.start_iteration_callback(pipeline.__class__, pipeline.parameters, self)
+            self.start_iteration_callback(pipeline, pipeline.parameters, self)
```
dsherry, line 453: While we're making a breaking change here: I think we should delete the `parameters` argument.

Why: now that we're providing a pipeline instance, the start iteration callback can access the parameters by calling `pipeline.parameters`, so there's no need for that to also be provided as a standalone arg.

dsherry (changes requested):
Looks great! I had two small changes to request: I think we should delete the parameters argument, and let's list this in breaking changes in the release notes.

evalml/automl/automl_search.py
angela97lin: Done!

dsherry (approved):
🙏 😁

PR #2295, opened 2021-05-20
Default `n_jobs` to 1 for `StackedEnsemblers` until we can write our own Ensembler

### LINKED ISSUE #2294: Titanic dataset with text columns fails on ensembling
Dataset: [titanic_text.csv](https://github.com/alteryx/evalml/files/6517249/titanic_text.csv)

Reproducer:
```
import pandas as pd
import woodwork as ww
data_set = pd.read_csv("titanic_alan.csv")
y = data_set['Survived']
X = data_set.drop(['Survived',], axis=1)
automl = AutoMLSearch(X, y, problem_type="binary", error_callback=raise_error_callback, max_batches=10, ensembling=True)
automl.search()
```

Errors with:

```

AttributeError                            Traceback (most recent call last)
<ipython-input-9-7e028fe6ad03> in <module>
      1 automl = AutoMLSearch(X, y, problem_type="binary", error_callback=raise_error_callback, max_batches=10, ensembling=True)
----> 2 automl.search()

~/Desktop/evalml/evalml/automl/automl_search.py in search(self, show_iteration_plot)
    598                     computation = computations[current_computation_index]
    599                     if computation.done():
--> 600                         evaluation = computation.get_result()
    601                         data, pipeline, job_log = evaluation.get('scores'), evaluation.get("pipeline"), evaluation.get("logger")
    602                         pipeline_id = self._post_evaluation_callback(pipeline, data, job_log)

~/Desktop/evalml/evalml/automl/engine/sequential_engine.py in get_result(self)
     33         Raises Exception: If computation fails. Returns traceback.
     34         """
---> 35         return self.work(**self.kwargs)
     36 
     37     def cancel(self):

~/Desktop/evalml/evalml/automl/engine/engine_base.py in evaluate_pipeline(pipeline, automl_config, X, y, logger)
    216     return train_and_score_pipeline(pipeline, automl_config=automl_config,
    217                                     full_X_train=X, full_y_train=y,
--> 218                                     logger=logger)
    219 
    220 

~/Desktop/evalml/evalml/automl/engine/engine_base.py in train_and_score_pipeline(pipeline, automl_config, full_X_train, full_y_train, logger)
    171             if automl_config.error_callback is not None:
    172                 automl_config.error_callback(exception=e, traceback=traceback.format_tb(sys.exc_info()[2]), automl=automl_config,
--> 173                                              fold_num=i, pipeline=pipeline)
    174             if isinstance(e, PipelineScoreError):
    175                 nan_scores = {objective: np.nan for objective in e.exceptions}

~/Desktop/evalml/evalml/automl/callbacks.py in raise_error_callback(exception, traceback, automl, **kwargs)
     13     logger.error(f'AutoML search raised a fatal exception: {str(exception)}')
     14     logger.error("\n".join(traceback))
---> 15     raise exception
     16 
     17 

~/Desktop/evalml/evalml/automl/engine/engine_base.py in train_and_score_pipeline(pipeline, automl_config, full_X_train, full_y_train, logger)
    160         try:
    161             logger.debug(f"\t\t\tFold {i}: starting training")
--> 162             cv_pipeline = train_pipeline(pipeline, X_train, y_train, automl_config.optimize_thresholds, automl_config.objective)
    163             logger.debug(f"\t\t\tFold {i}: finished training")
    164             if automl_config.optimize_thresholds and pipeline.can_tune_threshold_with_objective(automl_config.objective):

~/Desktop/evalml/evalml/automl/engine/engine_base.py in train_pipeline(pipeline, X, y, optimize_thresholds, objective)
    111                                                                   test_size=0.2, random_seed=pipeline.random_seed)
    112     cv_pipeline = pipeline.clone()
--> 113     cv_pipeline.fit(X, y)
    114     tune_binary_threshold(cv_pipeline, objective, cv_pipeline.problem_type,
    115                           X_threshold_tuning, y_threshold_tuning)

     16         @wraps(method)
     17         def _set_fit(self, X, y=None):
---> 18             return_value = method(self, X, y)
     19             self._is_fitted = True
     20             return return_value

~/Desktop/evalml/evalml/pipelines/classification_pipeline.py in fit(self, X, y)
     35         self._encoder.fit(y)
     36         y = self._encode_targets(y)
---> 37         self._fit(X, y)
     38         return self
     39 

~/Desktop/evalml/evalml/pipelines/pipeline_base.py in _fit(self, X, y)
    217     def _fit(self, X, y):
    218         self.input_target_name = y.name
--> 219         self._component_graph.fit(X, y)
    220         self.input_feature_names = self._component_graph.input_feature_names
    221 

~/Desktop/evalml/evalml/pipelines/component_graph.py in fit(self, X, y)
    122         X = infer_feature_types(X)
    123         X = _convert_woodwork_types_wrapper(X.to_dataframe())
--> 124         self._compute_features(self.compute_order, X, y, fit=True)
    125         self._feature_provenance = self._get_feature_provenance(X.columns)
    126         return self

~/Desktop/evalml/evalml/pipelines/component_graph.py in _compute_features(self, component_list, X, y, fit)
    249             else:
    250                 if fit:
--> 251                     component_instance.fit(input_x, input_y)
    252                 if not (fit and component_name == self.compute_order[-1]):  # Don't call predict on the final component during fit
    253                     output = component_instance.predict(input_x)

     16         @wraps(method)
     17         def _set_fit(self, X, y=None):
---> 18             return_value = method(self, X, y)
     19             self._is_fitted = True
     20             return return_value

~/Desktop/evalml/evalml/pipelines/components/estimators/estimator.py in fit(self, X, y)
     45         X, y = self._manage_woodwork(X, y)
     46         self.input_feature_names = list(X.columns)
---> 47         self._component_obj.fit(X, y)
     48         return self
     49 

~/Desktop/evalml_venv/lib/python3.7/site-packages/sklearn/ensemble/_stacking.py in fit(self, X, y, sample_weight)
    439         self._le = LabelEncoder().fit(y)
    440         self.classes_ = self._le.classes_
--> 441         return super().fit(X, self._le.transform(y), sample_weight)
    442 
    443     @if_delegate_has_method(delegate='final_estimator_')

~/Desktop/evalml_venv/lib/python3.7/site-packages/sklearn/ensemble/_stacking.py in fit(self, X, y, sample_weight)
    182                                        fit_params=fit_params,
    183                                        verbose=self.verbose)
--> 184             for est, meth in zip(all_estimators, self.stack_method_)
    185             if est != 'drop'
    186         )

~/Desktop/evalml_venv/lib/python3.7/site-packages/joblib/parallel.py in __call__(self, iterable)
   1052 
   1053             with self._backend.retrieval_context():
-> 1054                 self.retrieve()
   1055             # Make sure that we get a last message telling us we are done
   1056             elapsed_time = time.time() - self._start_time

~/Desktop/evalml_venv/lib/python3.7/site-packages/joblib/parallel.py in retrieve(self)
    931             try:
    932                 if getattr(self._backend, 'supports_timeout', False):
--> 933                     self._output.extend(job.get(timeout=self.timeout))
    934                 else:
    935                     self._output.extend(job.get())

~/Desktop/evalml_venv/lib/python3.7/site-packages/joblib/_parallel_backends.py in wrap_future_result(future, timeout)
    540         AsyncResults.get from multiprocessing."""
    541         try:
--> 542             return future.result(timeout=timeout)
    543         except CfTimeoutError as e:
    544             raise TimeoutError from e

~/.pyenv/versions/3.7.9/lib/python3.7/concurrent/futures/_base.py in result(self, timeout)
    433                 raise CancelledError()
    434             elif self._state == FINISHED:
--> 435                 return self.__get_result()
    436             else:
    437                 raise TimeoutError()

~/.pyenv/versions/3.7.9/lib/python3.7/concurrent/futures/_base.py in __get_result(self)
    382     def __get_result(self):
    383         if self._exception:
--> 384             raise self._exception
    385         else:
    386             return self._result

AttributeError: 'WordListCorpusReader' object has no attribute '_LazyCorpusLoader__args'

```

**Dropping all text columns allows the ensemble to run to completion:**

```
import pandas as pd
import woodwork as ww
data_set = pd.read_csv("titanic_alan.csv")
y = data_set['Survived']
X = data_set.drop(['Survived', 'Name', 'Embarked'], axis=1)
automl = AutoMLSearch(X, y, problem_type="binary", error_callback=raise_error_callback, max_batches=10, ensembling=True)
automl.search()
```

This is potentially relevant: https://github.com/nltk/nltk/issues/1576

ParthivNaresh:
Fixes #2294 

So it seems Angela and Freddy were most likely right regarding #[1576](https://github.com/nltk/nltk/issues/1576)’s relevance to this as well as this being primarily a `StackedEnsembler` issue. It seems to be an issue with an `nltk` backend implementation that was never solved. The underlying problem has to do with parallelism and how `nltk` manages large corpora of data. I believe it works like this:
1. We fit a `StackedEnsembler`, instantiating either `StackingClassifier` or `StackingRegressor` from `sklearn` depending on the problem type.
2. While transforming text related features, `nltk` is called by most likely a third party library.
3. `nltk` [leverages](https://www.nltk.org/_modules/nltk/corpus/util.html#LazyCorpusLoader) `LazyCorpusLoader`, a subclass of `WordListCorpusReader`, as a proxy object. It stands in for the corpus of text to be loaded to be more memory efficient.
4. Once `LazyCorpusLoader` is accessed for the first time, it's transformed into `WordListCorpusReader` and modifies its own `__class__` and `__dict__` attributes.
5. `sklearn` [passes](https://joblib.readthedocs.io/en/latest/parallel.html) `prefer="threads"` for `joblib.Parallel` fairly often, [especially for](https://github.com/scikit-learn/scikit-learn/blob/f6e6ad2d9e9172c55c778392b27b69c6af87bd98/sklearn/ensemble/_forest.py) `ensemble` libraries. I [don't think](https://github.com/scikit-learn/scikit-learn/blob/114616d9f6ce9eba7c1aacd3d4a254f868010e25/sklearn/ensemble/_stacking.py) it's doing that for `StackingClassifier` but I can't be sure and I don't know if the specification for the use of threads as a backend makes a difference.
6. The parallelism issue seems to be (as far as I can tell) that a second worker is attempting to access `LazyCorpusLoader` but because it hasn't converted into `WordListCorpusReader` yet, it begins the loading process all over.
7. By this point, the first worker has completed the transformation of `LazyCorpusLoader` into `WordListCorpusReader`.
8. Because the object is now `WordListCorpusReader`, it won't have the same attributes as `LazyCorpusLoader`, which leads to the error by the second worker of `AttributeError: 'WordListCorpusReader' object has no attribute '_LazyCorpusLoader__args'`

The short term solution here is to patch this by setting `n_jobs=1` in AutoMLSearch for ensembling problems that have text columns.

The long term solution is to build our own Ensembler that can leverage text based featurization in its own way without having to worry about worker conflicts.


ParthivNaresh:
Had to make a few changes, please refrain from reviewing just yet please!

evalml/automl/automl_search.py
```diff
@@ -385,6 +385,7 @@ def __init__(self,
         check_all_pipeline_names_unique(self.allowed_pipelines)
 
         run_ensembling = self.ensembling
+        text_in_ensembling = run_ensembling if infer_feature_types(X_train).select('natural_language').columns else False
```
dsherry, line 388: Ah right I forgot that under our current API, the automl algorithm doesn't have access to the data directly.

Is `run_ensembling` necessary to use here? Could you do this instead
```python
text_in_ensembling = len(infer_feature_types(X_train).select('natural_language').columns) > 0
```

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -105,9 +108,10 @@ def next_batch(self):
                 parameters = self._combine_parameters(pipeline, pipeline_params)
                 input_pipelines.append(pipeline.new(parameters=parameters,
                                                     random_seed=self.random_seed))
+            n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
```
dsherry, line 111: 👍

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1899,6 +1899,36 @@ def test_search_with_text(mock_fit, mock_score):
...
+                   'It was red and yellow and green and brown and scarlet and black and ochre and peach and ruby and olive and violet and fawn...']
+         })
+    if problem_type == 'binary':
+        y = [0, 1, 1, 0, 1, 0, 0, 1, 1, 0, 1, 0]
+    else:
+        y = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=problem_type, allowed_model_families=["catboost", "xgboost"],
+                          max_batches=4, ensembling=True)
+    automl.search()
+    assert automl.full_rankings.loc[automl.full_rankings['pipeline_name'] == pipeline_name].parameters.iloc[0][ensemble_name]['n_jobs'] == 1
```
dsherry, line 1929: @ParthivNaresh how long does this test take to run? With the `fit` and `score` mocking I bet its not too long.

An alternative organization would be to move the direct coverage down to `IterativeAlgorithm`:
* Move this test to instead cover `IterativeAlgorithm` directly. Set `ensembling` and `text_in_ensembling` to true, run a few batches and assert that the ensemble pipeline at the end has `n_jobs==1` in its parameters. Basically, what this tests already does.
* Add `AutoMLSearch` coverage which mocks the `IterativeAlgorithm` constructor, initializes `AutoMLSearch` and asserts that during initialization `IterativeAlgorithm` was given the right value for `text_in_ensembling`. Parametrize the test to use a grid on dataset with vs without a text feature and on `text_in_ensembling` set to true vs false.

If the current test runs longer than a few seconds, or if you like this organization more, let's do that!

dsherry: Ah, and I just saw you basically already have the first point done, below! Very cool. In that case, I think we should shrink this test to only cover the plumbing of the `text_in_ensembling` parameter, rather than checking the output of `AutoMLSearch.search`.

dsherry (changes requested):
Looking good! Left two comments.

evalml/automl/automl_search.py
ParthivNaresh: Much cleaner, thanks!

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: So I actually chose not to mock `fit` here because that's where the error takes place, so I wanted to make sure we caught that. As you mentioned the first point is covered in the `IterativeAlgorithm` tests. I like the idea of mocking the IterativeAlgorithm as per your second point!

ParthivNaresh: And this test does take too long for my liking so I'm sure it will speed it up

dsherry (approved):
Thanks, looks good to 🚢 !

PR #2181, opened 2021-04-22
Upgrade to ww 0.3.1

freddyaboulton:
### Pull Request Description

Fixes #2035 
Fixes #2019 
Fixes #2285 
Fixes #1955

Feature branch tracking the woodwork upgrade. Creating it now so that I can track if we ever fall behind `main`. 

Roadmap:

- [x] Demo datasets, utils, and preprocessing
- [x] Data Checks
- [x] Objectives (this will require updating the confusion matrix impl)
- [x] Components (will skip stacked ensembler unit tests since they take in pipelines)
- [x] Pipelines (will add stacked ensembler unit tests back in)
- [x] automl
- [x] model_understanding
- [x] Docs

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #2297, opened 2021-05-21
Component graph and text featurizer: preserve user-specified woodwork types

### LINKED ISSUE #2296: Converting Datetime to Natural Language produces TextFeaturizer error during pipeline fit
**Reproducer**
```python
import evalml
import pandas as pd
import woodwork as ww
import pytest

df = pd.read_csv('/Users/dylan.sherry/Downloads/fraud_detection_cleaned_v2_2500rows.csv')
dt = ww.DataTable(df)
y = dt.pop('fraud')
assert str(dt.logical_types['datetime']) == 'Datetime'

# manually change datetime to natural language. strange thing to do but it is possible.
dt = dt.set_types(logical_types={'datetime': 'NaturalLanguage'})
assert str(dt.logical_types['datetime']) == 'NaturalLanguage'

# define a pipeline _without_ a datetime featurizer, since there's no datetime-typed features in the data
pipeline = evalml.pipelines.BinaryClassificationPipeline([
    'Imputer', 'One Hot Encoder', 'Text Featurization Component', 'Elastic Net Classifier'])

# get error during pipeline fit
with pytest.raises(TypeError, match='invalid type promotion'):
    pipeline.fit(dt, y)
```

**Root Cause**
In our `TextFeaturizer`, we normalize to woodwork DataTable, get a pandas dataframe, run featuretools to compute some features, then run our LSA component to produce more features.

When we run the LSA component, we should be passing along the woodwork DataTable, because that contains the information that the "datetime" column happens to be NaturalLanguage-typed. However, instead we currently pass the pandas dataframe. This means the LSA component will run woodwork type inference, determine that "datetime" is a Datetime-typed feature, try to use that feature to generate TF-IDF features and error out.

[Here's the line](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py#L122) which causes the problem.

**Fix**
Short-term: update that line to provide the woodwork DataTable instead of the pandas dataframe, to preserve the user-specified feature types.
Long-term: hey @freddyaboulton does your woodwork accessor feature branch fix this problem? If not, we should definitely make a fix on there as well.

dsherry:
Fix #2296 

Repro and root cause are on the issue.

There was more to the problem than I had originally thought. In addition to the text featurizer not passing along user-specified woodwork types to the LSA component (which is used inside the text featurizer), our component graph also wasn't preserving custom types during `fit`. I shored up our unit test coverage of this case a bit.

@freddyaboulton I bet the component graph changes will cause conflicts with your woodwork accessor feature branch. but hopefully they will be minor. I tried to write the tests in a way which meant you could avoid huge conflicts there at least.


evalml/pipelines/component_graph.py
```diff
@@ -120,7 +121,7 @@ def fit(self, X, y):
             y (ww.DataColumn, pd.Series): The target training data of length [n_samples]
         """
         X = infer_feature_types(X)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
```
dsherry, line 123: First bug: there's no need to convert to pandas here. That's all handled in `_compute_features`.

Symptom: calls to pipeline `fit` always use whatever types woodwork infers for all the columns, overriding all user-specified values.

Explanation: the first thing we do in `_compute_features` is call `X = infer_feature_types(X)`. So, by converting to pandas here and then immediately converting back to woodwork first thing in `_compute_features`, we were effectively resetting the input to whatever woodwork infers by default.

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -119,7 +119,7 @@ def transform(self, X, y=None):
         if X_nlp_primitives.isnull().any().any():
             X_nlp_primitives.fillna(0, inplace=True)
 
-        X_lsa = self._lsa.transform(X[self._text_columns]).to_dataframe()
+        X_lsa = self._lsa.transform(X_ww[self._text_columns]).to_dataframe()
```
dsherry, line 122: This was the first fix required for this bug. We need to pass the woodwork type info to the LSA component, otherwise it will run woodwork inference again from scratch and the newly reset types will be used by all the downstream components!

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -683,8 +685,8 @@ def test_iteration(example_graph):
 
 
 def test_custom_input_feature_types(example_graph):
-    X = pd.DataFrame({'column_1': ['a', 'b', 'c', 'd', 'a', 'a', 'b', 'c', 'b'],
-                      'column_2': [1, 2, 3, 4, 5, 6, 5, 4, 3]})
+    X = pd.DataFrame({'column_1': ['a', 'a', 'a', 'b', 'b', 'b', 'c', 'c', 'd'],
+                      'column_2': [1, 2, 3, 3, 4, 4, 5, 5, 6]})
```
dsherry, line 689: No functional change, just reordered to make this test easier to understand

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -698,7 +700,7 @@ def test_custom_input_feature_types(example_graph):
     assert input_feature_names['Imputer'] == ['column_1', 'column_2']
     assert input_feature_names['OneHot_RandomForest'] == ['column_1', 'column_2']
     assert input_feature_names['OneHot_ElasticNet'] == ['column_1', 'column_2']
-    assert input_feature_names['Random Forest'] == ['column_1_a', 'column_1_b', 'column_2_4', 'column_2_5']
+    assert input_feature_names['Random Forest'] == ['column_1_a', 'column_1_b', 'column_2_3', 'column_2_4']
```
dsherry, line 703: This change did not occur because of the reordering above; it was happening before. I think this changed because "3", "4" and "5" each have two occurrences, but the RF OHE has `top_n` set to 2, so the OHE has to choose two categories from those top three tied categories.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -707,36 +709,53 @@ def test_component_graph_dataset_with_different_types():
...
+    assert input_feature_names['DateTime'] == ['column_3', 'column_4', 'column_5',
+                                               'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
                                                'column_2_1', 'column_2_2', 'column_2_3', 'column_2_4', 'column_2_5', 'column_2_6']
-    assert input_feature_names['Scaler'] == ['column_3', 'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
-                                             'column_2_1', 'column_2_2', 'column_2_3', 'column_2_4', 'column_2_5', 'column_2_6']
-    assert input_feature_names['Random Forest'] == ['column_3', 'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
-                                                    'column_2_1', 'column_2_2', 'column_2_3', 'column_2_4', 'column_2_5', 'column_2_6']
-    assert input_feature_names['Elastic Net'] == ['column_3', 'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
-                                                  'column_2_1', 'column_2_2', 'column_2_3', 'column_2_4', 'column_2_5', 'column_2_6']
+    assert input_feature_names['Text'] == ['column_3', 'column_5', 'column_1_a', 'column_1_b', 'column_1_c', 'column_1_d',
```
dsherry, line 742: Woodwork would infer the data in "column_5" to be datetime, but I set it to natural language above. We check here that it passes through the datetime featurizer untouched and enters the text featurizer. On the next assert, we check that it gets transformed into natural language features.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -707,36 +709,53 @@ def test_component_graph_dataset_with_different_types():
...
              'Elastic Net': [ElasticNetClassifier, 'Scaler.x'],
              'Logistic Regression': [LogisticRegressionClassifier, 'Random Forest', 'Elastic Net']}
 
     X = pd.DataFrame({'column_1': ['a', 'b', 'c', 'd', 'a', 'a', 'b', 'c', 'b'],
                       'column_2': [1, 2, 3, 4, 5, 6, 5, 4, 3],
                       'column_3': [True, False, True, False, True, False, True, False, False]})
+    X['column_4'] = [str((datetime(2021, 5, 21, 12, 0, 0) + timedelta(minutes=5 * x))) for x in range(len(X))]
+    X['column_5'] = X['column_4']
     y = pd.Series([1, 0, 1, 0, 1, 1, 0, 0, 0])
-    X = infer_feature_types(X, {"column_2": "categorical"})
```
dsherry, line 722: The confusing thing about this test is that it was passing before my changes. That's weird because here we set an integer column "column_2" to have categorical type, and saw that as expected one-hot encoded features based off of "column_2" show up in the datetime component's input.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -367,13 +369,13 @@ def test_fit_correct_inputs(mock_ohe_fit_transform, mock_imputer_fit_transform,
     X = pd.DataFrame(X)
     y = pd.Series(y)
     graph = {'Imputer': [Imputer], 'OHE': [OneHotEncoder, 'Imputer.x', 'Imputer.y']}
-    expected_x = ww.DataTable(pd.DataFrame(index=X.index, columns=X.index).fillna(1))
+    expected_x = ww.DataTable(pd.DataFrame(index=X.index, columns=X.columns).fillna(1))
```
dsherry, line 372: This isn't a "bug" per se, but the imputer will return one column for each input column, not more which is what this was doing

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -367,13 +369,13 @@ def test_fit_correct_inputs(mock_ohe_fit_transform, mock_imputer_fit_transform,
...
     expected_y = ww.DataColumn(pd.Series(index=y.index).fillna(0))
     mock_imputer_fit_transform.return_value = tuple((expected_x, expected_y))
     mock_ohe_fit_transform.return_value = expected_x
     component_graph = ComponentGraph(graph).instantiate({})
     component_graph.fit(X, y)
-    expected_x_df = expected_x.to_dataframe().astype("Int64")
+    expected_x_df = expected_x.to_dataframe().astype("float64")
```
dsherry, line 378: If you look at `X_y_binary`, all 20 columns in `X` have type `float64`.

evalml/pipelines/component_graph.py
```diff
@@ -319,13 +322,13 @@ def _consolidate_inputs(x_inputs, y_input, X, y):
...
         if len(x_inputs) == 0:
-            return_x = X
+            return_x = X.to_dataframe()
         else:
             return_x = pd.concat(x_inputs, axis=1)
         return_y = y
         if y_input is not None:
             return_y = y_input
-        return_x = infer_feature_types(return_x)
+        return_x = _retain_custom_types_and_initalize_woodwork(X, return_x)
```
dsherry, line 331: This method `_consolidate_inputs` currently only gets called inside `_compute_features`. If you look at the call site, you'll see the `x_inputs` is always a list of pandas dataframes. Therefore, after concatenating those pandas dataframes representing input features from the parent components in the graph, we need to call `_retain_custom_types_and_initalize_woodwork` to preserve any columns which appeared in the original woodwork input.

dsherry: I added `test_component_graph_types_multi_input_1` to check this is working properly.

evalml/pipelines/component_graph.py
```diff
@@ -120,7 +121,7 @@ def fit(self, X, y):
             y (ww.DataColumn, pd.Series): The target training data of length [n_samples]
         """
         X = infer_feature_types(X)
-        X = _convert_woodwork_types_wrapper(X.to_dataframe())
+        y = infer_feature_types(y)
```
dsherry, line 124: This, and the similar change in `_compute_features` are important. Otherwise, if the input provided to `fit` is not a woodwork structure, fit will fail!

evalml/pipelines/component_graph.py
```diff
@@ -319,13 +322,13 @@ def _consolidate_inputs(x_inputs, y_input, X, y):
             ww.DataTable, ww.DataColumn: The X and y transformed values to evaluate a component with
         """
         if len(x_inputs) == 0:
-            return_x = X
+            return_x = X.to_dataframe()
```
dsherry, line 325: Since we're now expecting `_compute_features` will be normalizing input to woodwork up the stack, we need to convert to pandas here. You can see that this X input is guaranteed to be woodwork by the code in `_compute_features`.

evalml/tests/automl_tests/test_dask_engine.py
```diff
@@ -154,8 +154,9 @@ def eval_pipelines(pipelines, engine):
         assert all([s._is_fitted for s in par_pipelines])
 
         # Ensure the scores in parallel and sequence are same
-        assert set(par_scores) == set(seq_scores)
```
dsherry, line 157: This test was flaking on me in the 17th decimal place, but only for [linux](https://github.com/alteryx/evalml/pull/2297/checks?check_run_id=2635860715)/[windows](https://github.com/alteryx/evalml/pull/2297/checks?check_run_id=2635667949) python 3.8 😆  not sure why exactly this PR would trigger that.

evalml/pipelines/component_graph.py
chukarsten: @dsherry this is very interesting and I wonder if it's part of what I was seeing in my investigation with Chris.

evalml/pipelines/component_graph.py
chukarsten: Ah, woodwork conversions

evalml/tests/component_tests/test_text_featurizer.py
```diff
@@ -218,6 +218,26 @@ def test_lsa_primitive_output():
...
+    expected_col_names = set(['col_2',
+                              'DIVERSITY_SCORE(col_1)',
+                              'LSA(col_1)[0]',
+                              'LSA(col_1)[1]',
+                              'MEAN_CHARACTERS_PER_WORD(col_1)',
+                              'POLARITY_SCORE(col_1)'])
+    X_t = tf.transform(X)
+    assert set(X_t.columns) == expected_col_names
+    assert len(X_t.columns) == 6
+    assert set(X_t.logical_types.values()) == {ww.logical_types.Double, ww.logical_types.Categorical}
```
chukarsten, line 238: Are you able to explicitly just check the logical type of "col_2"?  or do an ordered match?  I think it would just be nice to see that the specific user specified type and column were the same before and after.  It would be a little clearer to test readers that need to go through and realize that all the other fields will be doubles.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -707,39 +710,132 @@ def test_component_graph_dataset_with_different_types():
...
+    component_graph = ComponentGraph(graph)
+    # we don't have feature type selectors defined yet, so in order for the above graph to work we have to
+    # specify the types to select here.
+    # if the user-specified woodwork types are being respected, we should see the pass-through column_3 staying as numeric,
+    # meaning it won't cause a modeling error when it reaches the estimator
+    component_graph.instantiate({'Select numeric col_2': {'columns': ['column_2']},
+                                 'Select categorical col_1': {'columns': ['column_1']},
+                                 'Pass through col_3': {'columns': ['column_3']}})
+    assert component_graph.input_feature_names == {}
+    component_graph.fit(X, y)
```
chukarsten, line 789: So is this a smoke test for the estimator when fit() is called?  Where is the comment-mentioned test for column_3?

evalml/pipelines/component_graph.py
dsherry: Haha yep. This is necessary because there's no concatenate feature in woodwork `0.0.11`. I'm pretty sure this is handled on @freddyaboulton 's accessor feature branch

evalml/tests/component_tests/test_text_featurizer.py
dsherry: Oh yeah, that's true.
So, the fact that "col_2" did not end up being converted into text features implies that the text featurizer saw it as categorical-type, not as natlang type.
But you're right, the extra step would be to do an ordered match. I'll do it.

evalml/tests/pipeline_tests/test_component_graph.py
dsherry: Lol yeah this is a little confusing. The assert a couple lines down checks that column_3 was provided as an input to the RF estimator. Before the impl changes in this PR, this test would fail during `fit`, because the code where we concatenate the various inputs would fall over.

I'll see if I can find a clean way to have this test focus on the types which each component sees, rather than on the feature names. That would definitely make the value of this test easier to follow.

evalml/pipelines/component_graph.py
freddyaboulton: Yep, in the ww update branch we just do `return_x = X` hehe but fyi there still isn't a concatenate feature in ww

dsherry:
For those following along, this is still ready for review, but I'm waiting for #2181 to merge (woodwork accessor), then I'll rebase this and get it in.

bchen1116 (approved):
Nice! The tests look pretty thorough!

evalml/tests/pipeline_tests/test_component_graph.py
dsherry: Ok I think I got it

evalml/pipelines/component_graph.py
dsherry: Thanks for this @freddyaboulton . I'll try to file something for this before we merge this PR. Its a great point, and yep something we'll run into eventually.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -255,9 +255,9 @@ def round_dict_keys(dictionary, places=6):
     num_cat_features = len(set(X["currency"]))
     assert num_cat_features == 164
 
-    part_dep_ans = {0.1432616813857269: 154, 0.1502346349971562: 1, 0.14487916687594762: 1,
-                    0.1573183451314127: 1, 0.11695462432136654: 1, 0.07950579532536253: 1, 0.006794444792966759: 1,
-                    0.17745270478939879: 1, 0.1666874487986626: 1, 0.13357573073236878: 1, 0.06778096366056789: 1}
+    part_dep_ans = {0.19149451286750388: 154, 0.18249819137798806: 1, 0.24969722735684582: 1, 0.18542254073490247: 1,
```
freddyaboulton, line 258: What happened here is that since we updated the `ComponentGraph` to preserve the woodwork types throughout, card_id, store_id, customer_present get preserved as ints/bool so the predictions are different which makes the partial dependence results different.

This is a snapshot from the debugger of the ww types as soon as they make it to estimator.fit. Note that the ohe-columns are double instead of bool because of the standard scaler. 

### Main
woodwork_types_estimator_main


### This branch
image

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -363,18 +363,19 @@ def round_dict_keys(dictionary, places=6):
...
-        0.14487916687594762: 1,
-        0.1573183451314127: 1,
-        0.11695462432136654: 1,
-        0.07950579532536253: 1,
-        0.006794444792966759: 1,
-        0.17745270478939879: 1,
-        0.1666874487986626: 1,
-        0.13357573073236878: 1,
-        0.06778096366056789: 1,
+        0.19149451286750388: 154,
```
freddyaboulton, line 366: I had posted a [previous comment](https://github.com/alteryx/evalml/pull/2297#discussion_r645096770) explaining why this change is necessary but it got wiped from the diff view with the black pr. 

In summary, since we're preserving types we were previously not preserving (bool/int) the predicted probabilities are different which means the partial dependence is different.

freddyaboulton:
Perf tests [here](https://ayx.atlassian.net/wiki/spaces/AutoML/pages/513736705/Issue+2296+Preserve+Woodwork+Types+in+Component+Graph) 

Our linear estimators are doing worse so I think we should hold off on merging until we know why.

evalml/pipelines/component_graph.py
```diff
@@ -376,7 +383,22 @@ def _consolidate_inputs(x_inputs, y_input, X, y):
...
+        # type for that column. This requirement is important because the StandardScaler is expected to convert
+        # Ints to Doubles and converting back to an int would lose information.
+        # In order to meet this requirement, we start off with the user selected types (X.ww.logical_types) and then
+        # update them with the types created by components (if they are different).
+        # Components are not expected to create features with the same names
+        # so the only possible clash is between the types selected by the user and the types selected by a component.
+        # Because of that, the for-loop below is sufficient.
+
+        logical_types = {}
+        logical_types.update(X.ww.logical_types)
```
freddyaboulton, line 397: Before we would do this before the for-loop which was causing the performance regression in linear estimators

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1024,73 +1032,362 @@ def test_component_graph_dataset_with_different_types():
...
-        "column_2_6",
-    ]
-    assert input_feature_names["Logistic Regression"] == [
-        "Random Forest",
-        "Elastic Net",
+    assert input_feature_names["Random Forest"] == (
+        ["column_2", "column_1_a", "column_1_b", "column_1_c", "column_1_d", "column_3"]
+    )
+    assert mock_rf_fit.call_args[0][0].ww.logical_types["column_3"] == Integer
+    assert mock_rf_fit.call_args[0][0].ww.logical_types["column_2"] == Double
```
freddyaboulton, line 1226: Adding this to make sure we don't overwrite the types created by the standard scaler.

evalml/tests/model_understanding_tests/test_partial_dependence.py
freddyaboulton: Disregard this and the previous comment! This was actually a symptom of a bug hehe

freddyaboulton:
I've updated the performance tests [here](https://ayx.atlassian.net/wiki/spaces/AutoML/pages/513736705/Issue+2296+Preserve+Woodwork+Types+in+Component+Graph#Update-on-linear-estimators)! 

I was able to trace the issue to a bug where we were overriding the types set by the standard scaler. I updated the implementation so this should be good for review now!

freddyaboulton:
@dsherry I can't add you as a reviewer 😂  But your feedback would be appreciated!

evalml/pipelines/component_graph.py
```diff
@@ -376,7 +383,22 @@ def _consolidate_inputs(x_inputs, y_input, X, y):
...
-        return_x = infer_feature_types(return_x)
+
+        # Expected behavior: Preserve types selected by user unless one of the components updates the
+        # type for that column. This requirement is important because the StandardScaler is expected to convert
+        # Ints to Doubles and converting back to an int would lose information.
+        # In order to meet this requirement, we start off with the user selected types (X.ww.logical_types) and then
+        # update them with the types created by components (if they are different).
+        # Components are not expected to create features with the same names
+        # so the only possible clash is between the types selected by the user and the types selected by a component.
+        # Because of that, the for-loop below is sufficient.
```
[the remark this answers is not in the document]
dsherry: I wonder if woodwork could help manage this for us. I guess adding a `concatenate` ability would do it.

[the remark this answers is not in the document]
freddyaboulton: @dsherry I think you're right that ww.concat would definitely make this easier! Looks like it'll be in the next release so I'm really excited to try it out!

https://github.com/alteryx/woodwork/pull/932

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -118,15 +118,16 @@ def transform(self, X, y=None):
...
         es = self._make_entity_set(X_ww, self._text_columns)
         X_nlp_primitives = ft.calculate_feature_matrix(
             features=self._features, entityset=es
         )
         if X_nlp_primitives.isnull().any().any():
             X_nlp_primitives.fillna(0, inplace=True)
-
-        X_lsa = self._lsa.transform(X_ww[self._text_columns])
         X_nlp_primitives.set_index(X_ww.index, inplace=True)
+        X_lsa = self._lsa.transform(X_ww.ww[self._text_columns])
```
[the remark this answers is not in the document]
dsherry: Ohhhh. @freddyaboulton explained offline. Woodwork schema won't be copied by default pandas indexing operator, have to use woodwork namespace. Thanks @freddyaboulton !

PR #2314, opened 2021-05-28
Change minimum unit tests to run on all PRs

gsheni:


chukarsten (approved):

freddyaboulton:
@dsherry @ParthivNaresh I picked up where @gsheni left off. This should be ready for a second review now!

ParthivNaresh (approved):
Looks good! Quick question, does removing the `paths` make a difference here? Those paths would always be available for the CI to be triggered off of right?

gsheni:
@ParthivNaresh the `paths` made it so the workflow would only be trigger if it was a PR and those filepaths were changed

ParthivNaresh:
@gsheni Ahh correct I misread that [part](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) of the docs, sounds great!

Issue #2145, opened 2021-04-16
Investigate why Fraud/Lead scoring sometimes exhibits worse performance


chukarsten:
Following up on #2139 , it seems our notebook lead_scoring notebook sometimes shows contradictory modeling performance.  

From #2139 :

> Found that when we leave the data splitter as it was, the AUC for the LeadScoring-optimized search was 0.5 since it always guessed the same output. This is likely because balancing the data in this scenario leads to no learning done. If the data is balanced 1:4, as long as the gain outweighs the loss, we will always predict the same output. I changed the sampling to be smaller in order to highlight the point we wanted to convey.
> This previous version is similar to our current performance
> We note that we've had many iterations where the performance was very poor:
[v13](https://evalml.alteryx.com/en/v0.13.1/demos/lead_scoring.html#Describe-pipeline)
[v17](https://evalml.alteryx.com/en/v0.17.0/demos/lead_scoring.html#Evaluate-on-hold-out)
[v19](https://evalml.alteryx.com/en/v0.19.0/demos/lead_scoring.html#Evaluate-on-hold-out)
[v20](https://evalml.alteryx.com/en/v0.20.0/demos/lead_scoring.html#Evaluate-on-hold-out)

Closing Criteria: determine why this is happening and implement the long term solution going forward.

dsherry:
@bchen1116 will #2269 fix this issue? Or is there more which needs to happen before we can close it?

dsherry:
@bchen1116 and I spent some time digging into this.

**Summary**
* We discovered a cosmetic bug in automl search related to binary classification objective tuning. If the primary objective does not depend on classification threshold (e.g. AUC/log loss), but a secondary objective does depend on threshold, then the values for that second objective computed during automl will not be computed with a tuned threshold, and will therefore be artificially pessimistic about performance on that objective.
* This cosmetic bug could mean that whenever we set AUC or log loss as primary objective for automl, other objectives computed during automl are computed without a tuned threshold and are therefore worse than what is possible.
* API cleanup needed in `BinaryClassificationPipeline.optimize_threshold`: ideally we don't ask for `y_pred_proba` as an input, but rather compute that internally using the pipeline.
* Found minor bug in bin class threshold tuning plot: title is set to y axis instead

**Next steps**
* Figure out a fix for the additional objective scoring bug as described above.
* File issues for API cleanup and minor plotting bug
* Continue to look at threshold tuning graphs for lead scoring objective and see if we can fix the tutorial demo.

[...]

PR #2315, opened 2021-05-28
Fix Lead scoring demos

### LINKED ISSUE #2221: Fix Lead Scoring Demo
Currently, our lead_scoring demo doesn't use sampling in order to avoid a model that predicts all positive values, as seen [here](https://github.com/alteryx/evalml/pull/2128#discussion_r625120090). However, we have seen these trivial models in past docs, primarily [here](https://evalml.alteryx.com/en/v0.17.0/demos/lead_scoring.html) and [here](https://evalml.alteryx.com/en/v0.19.0/demos/lead_scoring.html).

This issue tracks looking into this demo, figuring out why the performance is poor, and setting up a fix for it.

bchen1116:
fixes #2221  and is part 1 of #2145 

New doc results:
image
image

Doc for how we chose these weights [here](https://alteryx.quip.com/9dUYAdLfZwCa/Fraud-and-Lead-scoring-analysis)
Perf test for binary pipeline is [here](https://ayx.atlassian.net/wiki/spaces/PS/pages/581337141/Bounded+Thresholding+Perf+Tests)


chukarsten (approved):

docs/source/demos/lead_scoring.ipynb
```diff
@@ -41,7 +41,7 @@
    "outputs": [],
    "source": [
     "lead_scoring_objective = LeadScoring(\n",
-    "    true_positives=1000,\n",
+    "    true_positives=100,\n",
```
freddyaboulton, line 44: @bchen1116 The lead scoring docs show a lead scoring score of 0 for both searches:

evalml/objectives/binary_classification_objective.py
```diff
@@ -43,7 +43,7 @@ def cost(threshold):
             cost = self.objective_function(y_true, y_predicted, X=X)
             return -cost if self.greater_is_better else cost
 
-        optimal = minimize_scalar(cost, method="Golden", options={"maxiter": 100})
+        optimal = minimize_scalar(cost, bounds=(0, 1), method="Bounded")
```
bchen1116, line 46: Using `bounded` rather than `Golden` in order to ensure that our threshold is between 0 and 1. `Bounded` uses a similar technique to `Brent`, which is preferable to `Golden`, according to [their docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize_scalar.html#scipy.optimize.minimize_scalar)

evalml/objectives/lead_scoring.py
```diff
@@ -43,5 +43,7 @@ def objective_function(self, y_true, y_predicted, X=None):
         profit += self.false_positives * false_positives
 
         profit_per_lead = profit / len(y_true)
+        # penalty if our estimator only predicts 1 output
+        same_class_penalty = (2 - len(set(y_predicted))) * (abs(profit_per_lead) / 2)
```
bchen1116, line 47: Adding a penalty if our estimator predicts only 1 output. Added this to try and avoid threshold graphs where the `0` or `1` threshold would result in the highest scores

angela97lin (approved):
I now see:
```
OrderedDict([('AUC', 0.6637428226050166), ('Lead Scoring', 0.034393809114359415)])
``` 
when we optimize for Lead Scoring and 
```
OrderedDict([('AUC', 0.8526896343306134), ('Lead Scoring', 0.0)])
```
when we optimize for AUC which aligns with what we're trying to say. Great work! :)

I wonder if we can catch this again if our AutoMLSearch changes and causes behavior in the docs that we don't expect. Maybe putting an assert in the docs to confirm that our auc-optimized lead scoring score is less than our lead-scoring optimized lead scoring score? (sorry, mouthful) 😁

bchen1116:
@angela97lin would failing assertions in the docs trigger any warnings for us, like in doc builds? We could make this a test, but that would add a fair bit of runtime to our tests. Otherwise, unsure how else we could catch this in the future beyond just revisiting the docs

angela97lin:
@bchen1116 Yup, or it'd just fail the docs test. Just having an assert between the two things we've already calculated above in a new cell shouldn't take too much runtime!

evalml/objectives/lead_scoring.py
dsherry: Got it.

Why the division by 2?

evalml/objectives/binary_classification_objective.py
dsherry: @bchen1116 please add a second entry to the release notes for this, under "Enhancements": "Updated threshold optimization method for binary classification"

docs/source/demos/lead_scoring.ipynb
dsherry: This comment is out of date, right? Resolving

evalml/tests/objective_tests/test_lead_scoring.py
```diff
@@ -105,3 +105,13 @@ def test_binary_more_than_two_unique_values():
     y_predicted = np.array([1, 0, 1])
     with pytest.raises(ValueError, match="y_true contains more than two unique values"):
         objective.score(y_true, y_predicted)
+
+
+@pytest.mark.parametrize(
+    "predicted,score", [([0, 1, 1, 1], 0.25), ([0, 0, 0, 0], 0), ([1, 1, 1, 1], 0)]
```
dsherry, line 111: 👍 !

dsherry (approved):
@bchen1116 this looks great. I left a question and a request for the release notes.

Since we'll be changing the optimization method we use for binary classification threshold tuning, please run the perf tests (at least on the binary classification datasets) and share the results before merging.

evalml/objectives/lead_scoring.py
bchen1116: @dsherry the division by 2 was just to make the penalty smaller. There was no real reason beyond that though. I can remove it if it makes it simpler?

freddyaboulton: I think that would make it so that the score of a model that predicts one label is `<= 0` right? Seems like a desirable property!

bchen1116:
@dsherry updated with the perf tests!

evalml/objectives/lead_scoring.py
dsherry: Thanks for explaining. That sounds good to me. Simpler is always good in my book :)

dsherry:
@bchen1116 looks good to me, let's 🚢 💨 !

bchen1116:
@angela97lin I added the assert to the docs! Hopefully it will allow us to catch this issue in the future

angela97lin:
@bchen1116 Perfect, looks great 😁

Issue #2166, opened 2021-04-20
Address need to duplicate pipeline parameters for time series pipelines in automl


angela97lin:
Post #2091, we pass pipeline instances to `allowed_pipelines` for AutoMLSearch. For time series pipelines, we have to specify `gap` and `max_delay`. However, in AutoMLSearch, we also have a `problem_configuration` parameter where we expect `max_delay` and `gap` as well. This means that the user has to duplicate these parameters (when initializing pipelines to pass to `allowed_pipelines` and when passing to AutoMLSearch). 

We should see if we can remove this redundancy. My guess is that we introduced `problem_configuration` before because we were working with pipeline classes and needed to store the parameters we wanted to use when creating our pipelines. Now that we're passing around pipeline instances, this should no longer be necessary, but we may need to verify this and change any logic that looks at `problem_configuration` to use the `gap` and `max_delay` of the input pipelines to `allowed_pipelines` instead.

freddyaboulton:
@angela97lin will a combination of #2159 and #1637 fix this problem?

dsherry:
@freddyaboulton I agree that #2159 will fix this. Let's keep it open until that issue is resolved.

Issue #2159, opened 2021-04-20
Update ``AutoMLSearch`` to accept ``allowed_component_graphs`` instead of ``allowed_pipelines``


angela97lin:

angela97lin:
Worth noting that this would affect the need for #2166: if we update to accept component graphs instead of pipelines, we would no longer be duplicating time series pipeline configurations and could close out #2166!

PR #2317, opened 2021-05-31
Separate `pipeline_parameters` from `custom_hyperparameters`

### LINKED ISSUE #2130: AutoMLSearch: clean up API for freezing parameters and constraining hyperparam ranges

ParthivNaresh:
Fixes #2130 

This PR deals with separating the `custom_hyperparameters` argument from its current pipeline implementation to `AutoMLSearch` instead. This is the first step before we [replace](https://github.com/alteryx/evalml/issues/2159) `allowed_pipelines` with `allowed_component_graphs` and [keep](https://github.com/alteryx/evalml/issues/2166) time series specifications within `problem_configuration`.

Document comparing the differences between the previous implementation and this one can be found [here](https://alteryx.quip.com/UvARAUYriWzb/Separating-pipeline-parameters-from-custom-hyperparameters).


evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -303,11 +328,7 @@ def test_iterative_algorithm_pipeline_params(parameters, dummy_binary_pipeline_c
     # make sure that future batches remain in the hyperparam range
     for i in range(1, 5):
         next_batch = algo.next_batch()
-        for p in next_batch:
-            if isinstance(parameters, Categorical):
-                assert p.parameters['Mock Classifier']['dummy_parameter'] in parameters
-            else:
-                assert p.parameters['Mock Classifier']['dummy_parameter'] == parameter
+        assert all([p.parameters['Mock Classifier']['dummy_parameter'] in hyperparameters for p in next_batch])
```
jeremyliweishih, line 331: might be nice to add a sanity check that the algorithm is producing different parameters (first example it goes thrugh 1, 3, and 5 as well).

jeremyliweishih (approved):
this is great! just had some clarifying questions but LGTM.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -33,7 +36,12 @@ def __init__(self,
         self._tuner_class = tuner_class or SKOptTuner
         self._tuners = {}
         for pipeline in self.allowed_pipelines:
-            self._tuners[pipeline.name] = self._tuner_class(pipeline.hyperparameters, random_seed=self.random_seed)
+            pipeline_hyperparameters = get_hyperparameter_ranges(pipeline.component_graph, custom_hyperparameters)
```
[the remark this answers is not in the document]
ParthivNaresh: Thanks for pointing it out!

evalml/tests/automl_tests/test_iterative_algorithm.py
ParthivNaresh: Great idea

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -164,25 +169,32 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
-
-            # Inspects each component and adds the following parameters when needed
-            if 'n_jobs' in init_params:
-                component_parameters['n_jobs'] = self.n_jobs
-            if 'number_features' in init_params:
-                component_parameters['number_features'] = self.number_features
             # For first batch, pass the pipeline params to the components that need them
-            if name in self._pipeline_params and self._batch_number == 0:
-                for param_name, value in self._pipeline_params[name].items():
+            if name in self._custom_hyperparameters and self._batch_number == 0:
```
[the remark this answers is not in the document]
ParthivNaresh: This is actually where the logic will differ slightly and circumvent the `Default parameters for components not in hyperparameter ranges` error in this scenario. An example of this is at `True - False - False - True` [here](https://alteryx.quip.com/UvARAUYriWzb/Separating-pipeline-parameters-from-custom-hyperparameters).

This allows the user to specify a custom hyperparameter range and expect that their pipelines will have parameters in those ranges. `pipeline_parameters` being passed to `AutoMLSearch` will still let them override the batch 0 parameters if they want to, but it gets rid of an error message that might not make sense to a new EvalML user.
i.e. I set the custom hyperparameters but it's giving me a ValueError that the default parameters are out of range, what do I do?

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -35,6 +35,7 @@ def __init__(self,
                  ensembling=False,
                  text_in_ensembling=False,
                  pipeline_params=None,
+                 custom_hyperparameters=None,
```
freddyaboulton, line 38: Do we still need the distinction between `pipeline_params` and `_frozen_pipeline_parameters`? I think it's adding some unnecessary complexity. From what I can tell, `pipeline_params` is used for passing the ts pipeline params (which are already included in `_frozen_pipeline_parameters`) and for overriding the parameters of the first batch.  Can we use only one of these? I think this is what's causing a bug in `test_automl_respects_pipeline_parameters_with_duplicate_components` (see comment below).

evalml/automl/utils.py
```diff
@@ -152,3 +154,24 @@ def get_best_sampler_for_data(X, y, sampler_method, sampler_balanced_ratio):
...
+
+    Returns:
+        dict: Dictionary of hyperparameter ranges for each component in the component graph.
+    """
+    linearized_component_graph = ComponentGraph.linearized_component_graph(component_graph)
+    hyperparameter_ranges = dict()
+    for component_name, component_class in linearized_component_graph:
+        component_hyperparameters = copy.copy(component_class.hyperparameter_ranges)
+        if custom_hyperparameters and component_name in custom_hyperparameters:
+            component_hyperparameters.update(custom_hyperparameters.get(component_name, {}))
```
freddyaboulton, line 175: Nit-pick: I don't think we need the `custom_hyperparameters.get` since we checked that the component name is in the dictionary.

freddyaboulton (changes requested):
@ParthivNaresh This looks great! I have two comments that I want to address before merging. The first I left in the diff for `iterative_algorithm` and the second I'll leave here:

Can we combine the `custom_hyperparameters` and `pipeline_parameters` arguments of AutoMLSearch?

One reason that we might want to do this is that I think it would be better for users. I think understanding the difference between what the two parameters means requires non-trivial understanding of automl. Secondly, I think it might actually simplify our code a bit if we only have one data structure that represents the parameters that need to override what the tuner proposes.

For example, right now if a user specifies `n_estimators` for a random forest as `pipeline_parameters` instead of `custom_hyperparameters` they can get a stack-trace:

```python
from evalml.model_family import ModelFamily

X, y = load_fraud(1000)

automl = AutoMLSearch(X, y, problem_type="binary", allowed_model_families=[ModelFamily.RANDOM_FOREST],
                      pipeline_parameters={"Random Forest Classifier": {"n_estimators": 5}},
                      max_batches=5)
automl.search()
```

If users specify the parameters as `custom_hyperparameters`, the search will finish but not as intended:
```python
from evalml.model_family import ModelFamily


X, y = load_fraud(1000)

automl = AutoMLSearch(X, y, problem_type="binary", allowed_model_families=[ModelFamily.RANDOM_FOREST],
                      custom_hyperparameters={"Random Forest Classifier": {"n_estimators": 5}},
                      max_batches=3)
automl.search()
n_estimators_used = automl.full_rankings.parameters.iloc[:11].map(lambda d: d["Random Forest Classifier"]['n_estimators'])
assert set(n_estimators_used) != {5}
```

 To wrap up, I think part of the problem is that we have multiple ways of trying to do the same thing. I propose only having one way (by consolidation pipeline_parameters and custom_hyperparameters) and focusing on making that one way work with lots of tests. I think there might be other solutions though! Happy to talk about it further.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2314,38 +2322,37 @@ def test_automl_respects_pipeline_parameters_with_duplicate_components(mock_scor
...
+                          pipeline_parameters={"One Hot Encoder": {"top_n": 15},
+                                               "One Hot Encoder_1": {"top_n": 25}},
                           max_batches=3)
     automl.search()
-    for i, row in automl.full_rankings.iterrows():
-        if "Mode Baseline Binary" in row['pipeline_name']:
-            continue
-        assert row["parameters"]["One Hot Encoder"]["top_n"] == 15
-        assert row["parameters"]["One Hot Encoder_1"]["top_n"] == 25
+    for row in automl.full_rankings.iloc[1:3].parameters:
```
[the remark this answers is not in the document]
ParthivNaresh: Since we'll be separating the logic, `pipeline_parameters` will be responsible for the first batch and as previously it would have represented the hyperparameters as well, now `custom_hyperparameters` will be responsible for that instead. So in this test only the initial batch is being set with these values, while the remaining batches revert to defaults in the case of `One Hot Encoder` (since `top_n` isn't a default hyperparameter range) or start iterating over the hyperparameter ranges in the case of `Random Forest Classifier`.

evalml/automl/automl_algorithm/iterative_algorithm.py
ParthivNaresh: So I was actually in the process of consolidating this as part of #2166 due to `problem_configuration` using time series parameters as well as the pipeline instances, since I agree that it would make more sense to have those parameters passed via one argument!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -242,6 +241,20 @@ def _transform_parameters(self, pipeline, proposed_parameters):
                         )
                     else:
                         component_parameters[param_name] = value
+            if name in self._pipeline_params and self._batch_number == 0:
+                for param_name, value in self._pipeline_params[name].items():
+                    if isinstance(value, (Integer, Real, Categorical)):
+                        raise ValueError(
+                            "Pipeline parameters should not contain skopt.Space variables, please pass them "
+                            "to custom_hyperparameters instead!"
+                        )
```
dsherry, line 250: This error is helpful!

I see `self._pipeline_params` here comes directly from `AutoMLSearch`, and is simply the `pipeline_params` input with the addition of any pipeline-level parameters (like time series config). I also see the `if` statements here only check the values of `self._pipeline_params`, and not the parameters proposed by the tuner.

So, we have a few options on where to do this check:
1. Don't do it
2. In `IterativeAlgorithm._transform_parameters` (current PR state)
3. In `IterativeAlgorithm.__init__`
4. In `AutoMLSearch.search`
5. In `AutoMLSearch.__init__`

Which do you think is best?

I think we should do this check, so 1 is out. I also think 3/4/5 are better than 2. Having this check in any of those spots is functionally equivalent -- if someone provides an skopt.Space in the `AutoMLSearch` `pipeline_parameters` input, this error will be triggered regardless. But in terms of code organization, it makes more sense to me to do this kind of check earlier. Since the check doesn't have anything to do with the proposed parameters here in `_transform_parameters`, I think its best to move it.

I think option 3 is my favorite. Doing it higher up the stack than `IterativeAlgorithm.__init__` means that initializing `IterativeAlgorithm` directly will skip the validation.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4328,9 +4339,12 @@ def test_automl_drop_index_columns(mock_train, mock_binary_score, X_y_binary):
     automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", max_batches=2)
     automl.search()
     for pipeline in automl.allowed_pipelines:
+        print(pipeline.parameters)
```
bchen1116, line 4342: forgot to remove?

evalml/automl/automl_search.py
```diff
@@ -411,11 +426,16 @@ def __init__(
         self.pipeline_parameters = (
             pipeline_parameters if pipeline_parameters is not None else {}
         )
+        self.custom_hyperparameters = (
+            custom_hyperparameters if custom_hyperparameters is not None else {}
+        )
```
dsherry, line 431: Style nit-pick:
```python
self.custom_hyperparameters = custom_hyperparameters or {}
```
I know this is mirroring the previous line... 🤷‍♂️ 😆

evalml/automl/automl_search.py
```diff
@@ -411,11 +426,16 @@ def __init__(
...
         self.search_iteration_plot = None
         self._interrupted = False
         self._frozen_pipeline_parameters = {}
 
         parameters = copy.copy(self.pipeline_parameters)
+        custom_hyperparameters = copy.copy(self.custom_hyperparameters)
```
dsherry, line 437: Why do a copy here? We copy `parameters` on the previous line because we then create a modified version where we add the pipeline-level parameters (time series config etc). But we don't do that for the custom hyperparameters, right?

evalml/automl/automl_search.py
```diff
@@ -471,19 +492,10 @@ def __init__(
...
                 for estimator in allowed_estimators
             ]
-        else:
-            for pipeline in self.allowed_pipelines:
-                if self.pipeline_parameters:
-                    if pipeline.custom_hyperparameters:
-                        for component_name, params in self.pipeline_parameters.items():
-                            pipeline.custom_hyperparameters[component_name] = params
-                    else:
-                        pipeline.custom_hyperparameters = self.pipeline_parameters
```
dsherry, line 486: Good riddance! What I wanna say to this code

get_outta_here

😆

evalml/automl/utils.py
```diff
@@ -194,3 +196,26 @@ def get_best_sampler_for_data(X, y, sampler_method, sampler_balanced_ratio):
                 return "SMOTENC Oversampler"
         except ImportError:
             return "Undersampler"
+
+
+def get_hyperparameter_ranges(component_graph, custom_hyperparameters):
```
dsherry, line 201: Got it. And in this case, `component_graph` is a list of strings / component classes, right? As opposed to a `ComponentGraph` instance.

Could you please update the args list to indicate this? I know this isn't the official syntax but you could say "component_graph (list(str, class))"

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2685,96 +2685,132 @@ def test_iterative_algorithm_pipeline_hyperparameters_make_pipeline_other_errors
...
+    "pipelines,pipeline_parameters", [(True, False), (True, True), (False, False)]
+)
+@pytest.mark.parametrize("automl_parameters", [True, False])
+@pytest.mark.parametrize("custom_hyperparameters", [True, False])
+@patch(
+    "evalml.pipelines.MulticlassClassificationPipeline.score",
+    return_value={"Log Loss Multiclass": 0.6},
+)
+@patch("evalml.pipelines.MulticlassClassificationPipeline.fit")
+def test_iterative_algorithm_pipeline_custom_hyperparameters_make_pipeline(
```
bchen1116, line 2698: big brain test

evalml/pipelines/pipeline_base.py
```diff
@@ -354,22 +341,6 @@ def model_family(self):
...
-            component_hyperparameters = copy.copy(component_class.hyperparameter_ranges)
-            if (
-                self.custom_hyperparameters
-                and component_name in self.custom_hyperparameters
-            ):
-                component_hyperparameters.update(
-                    self.custom_hyperparameters.get(component_name, {})
-                )
-            hyperparameter_ranges[component_name] = component_hyperparameters
-        return hyperparameter_ranges
```
dsherry, line 371: Glad to see this go!! 🎊

bchen1116 (approved):
Nice job, Parthiv! Left a few nits and testing suggestions.

evalml/pipelines/utils.py
```diff
@@ -189,16 +186,10 @@ def make_pipeline(
     )
     complete_component_graph = preprocessing_components + [estimator]
 
-    if custom_hyperparameters and not isinstance(custom_hyperparameters, dict):
-        raise ValueError(
-            f"if custom_hyperparameters provided, must be dictionary. Received {type(custom_hyperparameters)}"
-        )
```
dsherry, line 195: Could move this to `IterativeAlgorithm.__init__`

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -405,17 +405,66 @@ def test_iterative_algorithm_pipeline_params(parameters, dummy_binary_pipeline_c
...
-        parameter = parameter.rvs(random_state=0)
-    assert all(
-        [p.parameters["pipeline"] == {"gap": 2, "max_delay": 10} for p in next_batch]
+    if isinstance(parameter, (Categorical, Integer, Real)):
+        with pytest.raises(
+            ValueError,
+            match="Pipeline parameters should not contain skopt.Space variables",
+        ):
+            algo.next_batch()
+    else:
```
dsherry, line 415: Style nit-pick: flatten
```python
if isinstance(parameter, (Categorical, Integer, Real)):
    ...
    algo.next_batch()
    return
assert ...
```

evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -2,7 +2,7 @@ catboost==0.26
 click==8.0.1
 cloudpickle==1.6.0
 colorama==0.4.4
-dask==2021.5.1
+dask==2021.6.0
```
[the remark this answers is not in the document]
ParthivNaresh: Dask just had an upgrade 2 hours ago so it's necessary for the Detect Dependency Changes test

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -392,7 +392,7 @@ def test_iterative_algorithm_stacked_ensemble_n_jobs_regression(
 
 @pytest.mark.parametrize(
     "parameters",
-    [1, "hello", 1.3, -1.0006, Categorical([1, 3, 4]), Categorical((2, 3, 4))],
+    [1, "hello", 1.3, -1.0006, Categorical([1, 3, 4]), Integer(2, 4), Real(2, 6)],
```
dsherry, line 395: Good call

evalml/tests/dependency_update_check/latest_dependency_versions.txt
dsherry, line 5: Ah yeah, about to merge deps update to make this go away :)

dev-requirements.txt
```diff
@@ -4,4 +4,3 @@
 flake8==3.7.0
 black==21.5b1
 isort==5.0.0
-
```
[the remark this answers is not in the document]
ParthivNaresh: I'm not sure what this was, I think we had two new lines at the end of file and I just shortened it to 1 to conform to the rest of the requirement files? I have one line at the end of my local file but for some reason it's showing up as 0 here. It's not failing any lint checks so I'll just leave it as is!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -242,6 +241,20 @@ def _transform_parameters(self, pipeline, proposed_parameters):
                         )
                     else:
                         component_parameters[param_name] = value
+            if name in self._pipeline_params and self._batch_number == 0:
```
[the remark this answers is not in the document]
ParthivNaresh: Yes that's correct!

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -947,17 +878,15 @@ def __init__(self, random_arg=False, numpy_arg=[], random_seed=0):
...
-        ["Imputer", "Random Forest Classifier"],
-        custom_hyperparameters=custom_hyperparameters,
+        ["Imputer", "Random Forest Classifier"]
     )
     expected_code = (
         "from evalml.pipelines.binary_classification_pipeline import BinaryClassificationPipeline\n"
         "pipeline = BinaryClassificationPipeline(component_graph=['Imputer', 'Random Forest Classifier'], "
         "parameters={'Imputer':{'categorical_impute_strategy': 'most_frequent', 'numeric_impute_strategy': 'mean', 'categorical_fill_value': None, 'numeric_fill_value': None}, "
-        "'Random Forest Classifier':{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}}, custom_hyperparameters={'Imputer':{'numeric_impute_strategy': 'most_frequent'}}, random_seed=0)"
+        "'Random Forest Classifier':{'n_estimators': 100, 'max_depth': 6, 'n_jobs': -1}}, random_seed=0)"
```
dsherry, line 889: Awesome! Thanks for remembering to update codegen

evalml/tests/automl_tests/test_automl.py
dsherry: Yeah, this test is epic

evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -282,3 +286,26 @@ def test_get_best_sampler_for_data_sampler_method(
             assert name_output == "SMOTENC Oversampler"
         else:
             assert name_output == "SMOTEN Oversampler"
+
+
+def test_get_hyperparameter_ranges():
```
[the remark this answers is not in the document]
ParthivNaresh: Great idea! The expectation would be that no error is raised but that the hyperparameter ranges within the tuner don't include the nonexistent component.

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: most definitely forgot!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2685,96 +2685,132 @@ def test_iterative_algorithm_pipeline_hyperparameters_make_pipeline_other_errors
...
+        custom_hyperparameters=custom_hyperparameters_,
+        max_batches=4,
     )
     automl.search()
 
-    assert automl.best_pipeline.hyperparameters["Imputer"][
-        "numeric_impute_strategy"
-    ] == ["most_frequent", "mean"]
+    for i, row in automl.full_rankings.iterrows():
+        if "Random Forest Classifier" in row["pipeline_name"]:
```
dsherry, line 2764: I see we're only checking RF classifier here. But I'm guessing automl search ran other pipelines. I know we have mocking enabled which is good. But to save some runtime, could you set `model_family` to random forest in this test? That way we won't run other families. Could also turn down `max_batches` to 2.

evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -282,3 +286,26 @@ def test_get_best_sampler_for_data_sampler_method(
...
+        custom_hyperparameters=custom_hyperparameters_,
+    )
+    algo_ranges = algo._tuners[
+        "Random Forest Classifier w/ Imputer"
+    ]._pipeline_hyperparameter_ranges
+    hyper_ranges = get_hyperparameter_ranges(
+        pipeline_.component_graph, custom_hyperparameters_
+    )
+
+    assert algo_ranges == hyper_ranges
```
dsherry, line 311: Great call adding direct coverage of `get_hyperparameter_ranges`, we definitely need that now that a bunch of the old tests from `test_pipelines.py` are deleted!

This test currently checks that the output of `get_hyperparameter_ranges` matches the ranges used inside the automl algo (private variable). I think it would be more useful if it instead checked that the output of `get_hyperparameter_ranges` matches a hard-coded dictionary or something. You have other (great) behavioral tests in `test_automl.py` and the iterative algo tests which provide coverage of the algo behavior, so IMO there's no need to create an `IterativeAlgorithm` instance here and check a private field.

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -405,17 +405,66 @@ def test_iterative_algorithm_pipeline_params(parameters, dummy_binary_pipeline_c
...
+                    p.parameters["Mock Classifier"]["dummy_parameter"] == parameter
+                    for p in next_batch
+                ]
+            )
+
+
+@pytest.mark.parametrize(
+    "parameters,hyperparameters", [(1, Categorical([1, 3, 4])), (3, Integer(2, 4))]
+)
+def test_iterative_algorithm_custom_hyperparameters(
```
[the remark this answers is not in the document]
ParthivNaresh: The intended behaviour is actually that if `custom_hyperparameters` are passed alongside `pipeline_params` that don't exist in that range, an error will be thrown to the user that they have specified default parameters not in the hyperparameter ranges. I thought this was covered by `test_iterative_algorithm_pipeline_hyperparameters_make_pipeline_other_errors` but it doesn't look robust enough so I'll go ahead and add it here!

dsherry (changes requested):
@ParthivNaresh this is really excellent! And comprehensive. I left a few code organization things, in particular moving the `ValueError` for skopt parameters sneaking into `custom_parameters` to a different location, but not much.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -242,6 +241,20 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
+                for param_name, value in self._pipeline_params[name].items():
```
freddyaboulton, line 245: Tagging on to what @dsherry mentioned below, we may want to do a check for the "reverse" - specifying a custom hyperparameter without using `skopt.Space`. I think this may be helpful because not using `skopt.Space` for custom hyperparameters will not work as intended. I think it's an skopt issue since the tuner is properly created! 

```python
from evalml.model_family import ModelFamily


X, y = load_fraud(1000)

automl = AutoMLSearch(X, y, problem_type="binary", allowed_model_families=[ModelFamily.RANDOM_FOREST],
                      custom_hyperparameters={"Random Forest Classifier": {"n_estimators": 5}},
                      max_batches=3)
automl.search()
n_estimators_used = automl.full_rankings.parameters.iloc[:11].map(lambda d: d["Random Forest Classifier"]['n_estimators'])
assert set(n_estimators_used) != {5}
```

image

evalml/automl/callbacks.py
```diff
@@ -37,7 +37,5 @@ def log_error_callback(exception, traceback, automl, **kwargs):
     logger.info(
         f"\t\t\tFold {fold_num}: Exception during automl search: {str(exception)}"
     )
-    logger.debug(
-        f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.hyperparameters}"
-    )
+    logger.debug(f"\t\t\tFold {fold_num}: Hyperparameters:\n\t{pipeline.parameters}")
```
freddyaboulton, line 40: I like this change! Should we change the text to say `Parameters:\n\t{pipeline.parameters}`?

evalml/tests/automl_tests/test_automl.py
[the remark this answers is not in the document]
freddyaboulton: Thanks for explaining! We can preserve the previous behavior by setting `custom_hyperparameters`! I think we should list this as a breaking change. Not sure how many users were actively setting `pipeline_parameters` before, but might be good to give them a heads up to use `custom_hyperparameters` instead.

evalml/tests/automl_tests/test_automl_utils.py
freddyaboulton: Maybe we can recycle some of the deleted pipeline tests that checked that the pipeline `custom_hyperparameters` attribute was set correctly?

evalml/tests/automl_tests/test_iterative_algorithm.py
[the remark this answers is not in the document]
freddyaboulton: Great idea! In that case, the user would have to set new hyperparameter ranges that include the value set in `pipeline_parameters` right? Might be good to document that in a test.

freddyaboulton (approved):
@ParthivNaresh Looks great to me! I left a minor comment that's similar to @dsherry `s suggestion but other than that this is awesome.

evalml/automl/automl_search.py
ParthivNaresh: Ah thanks, initially I was considering modifying the `custom_hyperparameters` based on other criteria but forgot to change this back.

evalml/automl/automl_algorithm/iterative_algorithm.py
ParthivNaresh: Great point!

evalml/tests/automl_tests/test_automl.py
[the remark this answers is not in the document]
ParthivNaresh: Most definitely, I'll list this as a breaking change!

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -91,8 +94,29 @@ def __init__(
...
+                        "Pipeline parameters should not contain skopt.Space variables, please pass them "
+                        "to custom_hyperparameters instead!"
+                    )
+        for hyperparam_name_val in self._custom_hyperparameters.values():
+            for _, hyperparam_val in hyperparam_name_val.items():
+                if not isinstance(hyperparam_val, (Integer, Real, Categorical)):
+                    raise ValueError(
+                        "Custom hyperparameters should only contain skopt.Space variables such as Categorical, Integer,"
+                        " and Real!"
+                    )
```
dsherry, line 118: Marvelous

dsherry (approved):
🚢 🙏

Issue #2298, opened 2021-05-21
Test Profiling


chukarsten:
Goal:

1. Identify top 10 maximum run time tests
2. Identify top 10 memory usage tests
3. Evaluate the impact of streamlining fake test data into singular test fixtures within conftest
4. @dsherry what did I miss?

freddyaboulton:
Did some work for point 1 about a week ago:

https://github.com/alteryx/evalml/issues/1815#issuecomment-837379777

dsherry:
Looks good to me!

The primary goal of this spike is to speed up our unit tests. Memory usage is secondary.

The "top 10" is fuzzy, haha. Basically we wanna know which of our tests are the worst offenders when it comes to runtime.

chukarsten:
@freddyaboulton thanks a ton for doing that and bringing it back up.  That already seems to reveal a lot of pain points.  I think ultimately we're going to want to profile the tests themselves and see which function calls are the big hitters as well, so we'll need to drill down into that.  I look forward to it.

dsherry:
@freddyaboulton I believe you've completed this, correct?

freddyaboulton:
Yes! Closing.

Issue #1815, opened 2021-02-10
Unit Test Runtime


chukarsten:
This spike is intended to track the running of the unit tests.  Not sure whether it's worth tracking linux vs. windows separately here.  Unit tests are currently hitting ~20 minutes to complete locally and on CircleCI checks.  Even though we're moving to GitHub actions, that shouldn't really make the problem any better or any worse.  

The intended outcome of this spike is the following:

1. A documented methodology for profiling the unit tests.
2. Unit test profiling results posted (on a quip?) as a baseline.  (I know Steve Link showed a profiling tool that I've used before that is pretty visual and useful.)
3. Recommendations provided for the top 5 unit tests to address the performance of.
4. A quick discussion concerning the above.
5. Issues filed for the agreed results of the above.

dsherry:
@angela97lin I am remembering you did some unit test runtime profiling. Do you have any code to share for that and/or thoughts?

Primary goal: speed up CI runtime. So I'd recommend starting with the longest-running CI job (build_conda_package) and finding ways to speed that up.

angela97lin:
@dsherry Yup, I had to do some unit test runtime profiling for the WW PRs. I ended up writing a simple script that parses the XML files generated in our artifacts (example [here](https://app.circleci.com/pipelines/github/alteryx/evalml/9781/workflows/8f073d9f-8f2f-4952-a5a8-e7ec9485c732/jobs/115540/artifacts)) and comparing the one on the branch with the one in main. Then I just printed out the difference in time between the two for each test. Kinda messy but here's the script, super messy and specific to what I was doing, but maybe a good start for this:

```
import csv 
import xml.etree.ElementTree as ET 

           
def parse_xml(xmlfile): 
    tree = ET.parse(xmlfile) 
    root = tree.getroot() 
    results = {}
    for test_case in root.findall('testcase'): 
        attributes = test_case.attrib
        name = attributes['name']
        time = attributes['time']
        if float(time) > 1:
            results[name] = time
    return results


def compare_results(main_results, other_results):
    timing_results_dict = {}
    for test_case_name in main_results:
        try:
            time_diff = float(other_results[test_case_name]) - float(main_results[test_case_name])
            timing_results_dict[test_case_name] = time_diff
        except KeyError:
            continue
    return timing_results_dict
      
def main(): 
    main_results = parse_xml('main.xml') 
    woodwork_results = parse_xml('woodwork_36.xml') 
[...]

PR #2320, opened 2021-06-01
Add thresholding_objective argument to AutoMLSearch

### LINKED ISSUE #2301: Bin class: if automl objective is AUC/logloss, additional objective scores are low
If a user passes in a secondary objective that allows for thresholding, we should threshold using this objective, even if the primary objective doesn't use it. For instance, if a user uses 'log loss binary' as the primary, and 'binary accuracy` as an additional objective, we want to score on the secondary objective using thresholded pipelines.

This is part of the issue with the lead-scoring demo, as @dsherry mentioned [here](https://github.com/alteryx/evalml/issues/2145#issuecomment-846266275)

bchen1116:
fixes #2301 

Adds an additional thresholding_objective argument to AutoMLSearch. We use this to threshold binary classification problems when the original objective isn't thresholdable.

Original design doc [here](https://alteryx.quip.com/SizBAztQfCiZ/Threshold-tuning-on-additional-objectives)
Perf test results [HERE](https://alteryx.quip.com/PrBPAKanDZBc/Thresholding-Objective-Perf-tests)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -144,37 +144,37 @@ def test_pipeline_limits(mock_fit_binary, mock_score_binary,
     mock_score_multi.return_value = {'Log Loss Multiclass': 1.0}
     mock_score_regression.return_value = {'R2': 1.0}
 
-    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=automl_type, max_iterations=1)
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type=automl_type, optimize_thresholds=False, max_iterations=1)
```
bchen1116, line 147: For most of these tests, I choose to set `optimize_threhsholds=False` rather than patch `predict_proba`, `optimize_thresholds`, and `_encode_targets`.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -673,9 +673,14 @@ def test_automl_time_series_classification_threshold(mock_binary_fit, mock_binar
...
     assert isinstance(automl.data_splitter, TimeSeriesSplit)
     if objective == 'Log Loss Binary':
-        mock_optimize_threshold.assert_not_called()
-        assert automl.best_pipeline.threshold is None
-        mock_split_data.assert_not_called()
+        if not optimize:
+            mock_optimize_threshold.assert_not_called()
+            assert automl.best_pipeline.threshold is None
+            mock_split_data.assert_not_called()
+        else:
```
bchen1116, line 680: Check that we optimize the threshold even if the main objective isn't optimizable

evalml/automl/automl_search.py
```diff
@@ -423,11 +430,18 @@ def __init__(self,
...
-                                          self.objective, self.additional_objectives, self.optimize_thresholds,
+                                          self.objective, self.additional_objectives, self.thresholding_objective, self.optimize_thresholds,
                                           self.error_callback, self.random_seed,
                                           self.X_train.ww.schema,
                                           self.y_train.ww.schema)
-
+        self.threshold_automl_config = self.automl_config
+        if is_binary(self.problem_type) and self.optimize_thresholds and self.objective.score_needs_proba:
+            # use the thresholding_objective
+            self.threshold_automl_config = AutoMLConfig(self.data_splitter, self.problem_type,
```
bchen1116, line 440: This allows us to use the `thresholding_objective` as the objective to train the best pipeline with if we allow it.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -687,25 +692,6 @@ def test_automl_time_series_classification_threshold(mock_binary_fit, mock_binar
         mock_split_data.assert_not_called()
 
 
-@pytest.mark.parametrize("objective", ['F1', 'Log Loss Binary', 'AUC'])
-@patch('evalml.objectives.BinaryClassificationObjective.optimize_threshold')
-@patch('evalml.pipelines.BinaryClassificationPipeline._encode_targets', side_effect=lambda y: y)
-@patch('evalml.pipelines.BinaryClassificationPipeline.score')
-@patch('evalml.pipelines.BinaryClassificationPipeline.fit')
-@patch('evalml.pipelines.BinaryClassificationPipeline.predict_proba')
-def test_tuning_threshold_objective(mock_predict, mock_fit, mock_score, mock_encode_targets, mock_optimize_threshold, objective, X_y_binary):
```
bchen1116, line 696: The other tests cover this condition already, so removing this

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1165,9 +1173,14 @@ def test_automl_time_series_classification_threshold(
...
+            mock_optimize_threshold.assert_not_called()
+            assert automl.best_pipeline.threshold is None
+            mock_split_data.assert_not_called()
+        else:
+            mock_optimize_threshold.assert_called()
+            assert automl.best_pipeline.threshold == 0.62
+            mock_split_data.assert_called()
     elif optimize and objective == "F1":
         mock_optimize_threshold.assert_called()
         assert automl.best_pipeline.threshold == 0.62
```
chukarsten, line 1186: So should this move into the above else branch and be a split on mock_split_data being called or not?

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1443,6 +1425,7 @@ def test_automl_search_dictionary_oversampler(
         y_train=y,
         problem_type=problem_type,
         sampler_method="SMOTE Oversampler",
+        optimize_thresholds=False,
```
chukarsten, line 1428: Curious, what's going on here that all the optimize_thresholds need to be set to something off the default in all these tests that don't really have anything to do with the objective?

chukarsten (approved):
Hey Bryan, good work.  I don't think there's anything blocking, but I do think some lines can be removed, refactored just a little.  Let me know what you think.

evalml/automl/automl_search.py
```diff
@@ -556,13 +565,32 @@ def __init__(
...
+                self.problem_type,
+                self.thresholding_objective,
+                self.additional_objectives,
+                self.thresholding_objective,
+                self.optimize_thresholds,
+                self.error_callback,
+                self.random_seed,
+                self.X_train.ww.schema,
+                self.y_train.ww.schema,
+            )
```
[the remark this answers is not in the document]
bchen1116: Yep! The only difference is that the spot for `objective`, we use the `thresholding_objective` instead so that we can train and tune on this when necessary

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: @chukarsten Since we default to tuning the threshold, and this PR adds an objective to use for that even when the primary objective isn't tunable, we need to add this line in order to avoid tuning the threshold. When we tune, we call `predict_proba`, among other calls, and this fails since we're mocking `fit`, which results in a `pipeline not yet fitted` error. 

I chose to set `optimize_thresholds` to false rather than mocking `predict_proba` and the other methods that are involved in tuning.

evalml/automl/automl_search.py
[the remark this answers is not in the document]
chukarsten: Ah, sorry, I missed the difference there.

chukarsten (approved):

freddyaboulton:
@bchen1116 Are you planning on running performance tests? I think we should - looks like by default our pipelines will be fit with 20% less data from the tuning split and we'll also be tuning the threshold. I think it'd be good to quantify what the impact is on the scores and search time just so that we're aware !

bchen1116:
Good point @freddyaboulton! I'll run those and will updat this PR once they finish!

bchen1116:
@freddyaboulton updated this with the [perf test results](https://alteryx.quip.com/PrBPAKanDZBc/Thresholding-Objective-Perf-tests)!

evalml/automl/automl_search.py
```diff
@@ -281,6 +285,11 @@ def __init__(
             objective = get_default_primary_search_objective(self.problem_type.value)
         objective = get_objective(objective, return_instance=False)
         self.objective = self._validate_objective(objective)
+        self.thresholding_objective = get_objective(
+            thresholding_objective, return_instance=True
+        )
+        if self.thresholding_objective.score_needs_proba:
+            raise ValueError("Thresholding objective must be thresholdable!")
```
ParthivNaresh, line 292: Nitpick, I love this error message but should we perhaps specify that the thresholding objective cannot be one that needs proba so it might be more clear to the user? They might pass an additional objective not fully knowing what it entails.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -498,11 +526,21 @@ def test_automl_str_no_param_search(X_y_binary):
     assert "Search Results" not in str_rep
 
 
+@patch("evalml.objectives.BinaryClassificationObjective.optimize_threshold")
+@patch(
+    "evalml.pipelines.BinaryClassificationPipeline._encode_targets",
+    side_effect=lambda y: y,
```
ParthivNaresh, line 532: Very nice

ParthivNaresh (approved):
Excellent work on this!

evalml/automl/automl_search.py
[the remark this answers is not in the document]
dsherry: @bchen1116 hang on, I think @chukarsten 's point still stands. The logic you added in `engine_base.py` looks great. And I see you've added the alternate threshold tuning objective as an additional argument to `AutoMLConfig` above this code block, which looks great. So, why can't you delete this block entirely?

evalml/automl/automl_search.py
```diff
@@ -281,6 +285,13 @@ def __init__(
             objective = get_default_primary_search_objective(self.problem_type.value)
         objective = get_objective(objective, return_instance=False)
         self.objective = self._validate_objective(objective)
+        self.thresholding_objective = get_objective(
+            thresholding_objective, return_instance=True
+        )
+        if self.thresholding_objective.score_needs_proba:
+            raise ValueError(
+                "Thresholding objective must be thresholdable and cannot need probabilities!"
```
dsherry, line 293: This is a nit-pick but I don't think most people will understand what "thresholdable" means. Perhaps "Alternate threshold tuning objective must be a tuneable objective" ?

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -482,6 +482,7 @@ def test_describe_pipeline_objective_ordered(X_y_binary, caplog):
         problem_type="binary",
         objective="AUC",
         max_iterations=2,
+        optimize_thresholds=False,
```
dsherry, line 485: Why set these to `False`?

If you did it to avoid having the tests waste time calling the threshold optimizer, then, lol I agree we shouldn't run the optimizer in every test. I do wonder if there's a different way to accomplish this though. Could we mock `BinaryClassificationObjective.optimize_threshold` instead, in the same way we mock pipeline `fit` and `score` in many tests?

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1164,51 +1172,17 @@ def test_automl_time_series_classification_threshold(
...
-    automl.search()
-
-    if objective != "F1":
-        assert automl.best_pipeline.threshold is None
-    else:
-        assert automl.best_pipeline.threshold == 0.6
+        if objective == "Log Loss Binary":
+            assert automl.best_pipeline.threshold is None
+        else:
+            assert automl.best_pipeline.threshold == 0.5
```
dsherry, line 1185: Ah, so when "optimize" is true, that means automl will have run in this test with `optimize_thresholds` true, in which case the mock optimization value is expected, regardless of what the objective was. 👍 

If you wanna go above and beyond haha, I bet we could move most of this test to cover `engine_base.py` code directly instead of being an automl-level test. We wrote this test before the engine concept existed.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4474,3 +4688,69 @@ def test_automl_baseline_pipeline_predictions_and_scores_time_series(problem_typ
...
+            )
+    else:
+        automl = AutoMLSearch(
+            X_train=X,
+            y_train=y,
+            problem_type="binary",
+            thresholding_objective=objective,
+            max_iterations=2,
+        )
+        automl.search()
```
dsherry, line 4726: @bchen1116 how long does this test take to run?

Could you please mock `BinaryClassificationPipeline.fit` and `score` like we do in other tests? I bet that will reduce the runtime quite a bit, without changing the test behavior. Same for the other new test.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4474,3 +4688,69 @@ def test_automl_baseline_pipeline_predictions_and_scores_time_series(problem_typ
...
+        with pytest.raises(
+            ValueError, match="Thresholding objective must be thresholdable"
+        ):
+            AutoMLSearch(
+                X_train=X,
+                y_train=y,
+                problem_type="binary",
+                thresholding_objective=objective,
+            )
+    else:
```
dsherry, line 4718: Style nit-pick: flatten
```python
if errors:
    with pytest.raises(...)
        AutoMLSearch(...)
    return
automl = AutoMLSearch(...)
automl.search()
...
```

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4474,3 +4688,69 @@ def test_automl_baseline_pipeline_predictions_and_scores_time_series(problem_typ
...
+        ("AUC", True),
+        ("F1", False),
+        ("Accuracy Binary", False),
+    ],
+)
+@patch(
+    "evalml.objectives.binary_classification_objective.BinaryClassificationObjective.optimize_threshold",
+    return_value=0.65,
+)
+def test_automl_thresholding_objective(mock_optimize, objective, errors, X_y_binary):
```
dsherry, line 4706: Great test!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4474,3 +4688,69 @@ def test_automl_baseline_pipeline_predictions_and_scores_time_series(problem_typ
...
+            max_iterations=2,
+        )
+        automl.search()
+        mock_optimize.assert_called()
+        assert automl.best_pipeline.threshold == 0.65
+
+
+@pytest.mark.parametrize("threshold", [False, True])
+@patch("evalml.objectives.standard_metrics.F1.optimize_threshold", return_value=0.42)
+def test_automl_thresholding_train_pipelines(mock_objective, threshold, X_y_binary):
```
dsherry, line 4733: Awesome

dsherry (changes requested):
@bchen1116 looking good! I had one code change request, some comments on naming/docs and a couple points on the tests.

evalml/tests/automl_tests/test_automl.py
bchen1116: It takes about 4 seconds to run. I can set max_iterations even lower and get it to run in under 2 seconds. This is comparable to the runtime when we mock, but with this, and the following test, I wanted to make sure the thresholds were being set properly. Would it still be preferable to mock the `fit`, `score`, `predict_proba`, and other methods here?

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: Removed the param here!

evalml/automl/automl_search.py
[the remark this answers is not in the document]
bchen1116: Good point! I changed some code to remove this. Hopefully it's simpler!

evalml/automl/automl_search.py
```diff
@@ -216,6 +218,8 @@ def __init__(
             additional_objectives (list): Custom set of objectives to score on.
                 Will override default objectives for problem type if not empty.
 
+            thresholding_objective (str): The objective to use for thresholding binary classification pipelines. Defaults to F1.
```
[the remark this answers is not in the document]
bchen1116: @dsherry Renamed and fixed! Let me know if you think I should add more to the description

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: @dsherry What do you mean here?

evalml/automl/engine/engine_base.py
```diff
@@ -149,6 +157,14 @@ def train_and_score_pipeline(
     start = time.time()
     cv_data = []
     logger.info("\tStarting cross validation")
+    threshold_tuning_objective = automl_config.objective
+    if (
+        is_binary(automl_config.problem_type)
+        and automl_config.optimize_thresholds
+        and automl_config.objective.score_needs_proba
+    ):
+        # use the alternate_thresholding_objective
```
dsherry, line 166: Ultra nit-pick: this comment doesn't add any new information :)

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -715,6 +715,7 @@ def test_automl_allowed_pipelines_specified_allowed_pipelines_binary(
         X_train=X,
         y_train=y,
         problem_type="binary",
+        optimize_thresholds=False,
```
dsherry, line 718: Following up on [my outdated comment](https://github.com/alteryx/evalml/pull/2320#discussion_r646970921) on another one of these similar lines:
> Why set these to False?
> 
> If you did it to avoid having the tests waste time calling the threshold optimizer, then, lol I agree we shouldn't run the optimizer in every test. I do wonder if there's a different way to accomplish this though. Could we mock `BinaryClassificationObjective.optimize_threshold` instead, in the same way we mock pipeline fit and score in many tests?

I wasn't just referring to that specific test though heh, I was referring to every automl test where you've added `optimize_thresholds=False` in this PR.

evalml/tests/automl_tests/test_automl_search_classification.py
dsherry: Sorry lol. First part was me explaining to myself what this test does so that I understand it. Second part was a suggested improvement which you can certainly ignore if you'd like.

My point was that technically this test is checking the behavior of `EngineBase.train_and_score_pipelines` and doesn't have anything to do with `AutoMLSearch` itself, right? I guess its _also_ checking that the threshold values get saved and attached to `AutoMLSearch.best_pipeline`, but we could write a simpler test to check that using mocking.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4488,3 +4702,72 @@ def test_automl_baseline_pipeline_predictions_and_scores_time_series(problem_typ
...
+            )
+        return
+    automl = AutoMLSearch(
+        X_train=X,
+        y_train=y,
+        problem_type="binary",
+        alternate_thresholding_objective=objective,
+        max_iterations=1,
+    )
+    automl.search()
```
dsherry, line 4743: Following up on [outdated comment](https://github.com/alteryx/evalml/pull/2320#discussion_r646985865). @bchen1116 said:
> @dsherry It takes about 4 seconds to run. I can set max_iterations even lower and get it to run in under 2 seconds. This is comparable to the runtime when we mock, but with this, and the following test, I wanted to make sure the thresholds were being set properly. Would it still be preferable to mock the fit, score, predict_proba, and other methods here?

Oh lol so with `max_iterations=1` this is setting the threshold on the baseline pipeline. Sure that's fine. Just as long as we aren't burning test runtime training real models when we don't need to for this test!

dsherry (changes requested):
I left one request for refactor in the dask engine, to avoid duplicated code. After that, let's 🚢 !

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: Ah, I added `optimize_thresholds` to false to avoid tuning the thresholds. The other way would've been to patch `predict_proba`, `optimize_thresholds`, and `encode_targets` since so many of the tests patch the pipeline `fit`.

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: @dsherry I might just leave this test as is since it is particular to time series as well, and it seems to be a thorough enough test that shouldn't take much time. Seems like this test is just making sure that AutoMLSearch can threshold time series problems properly when needed.

evalml/tests/automl_tests/test_automl_search_classification.py
dsherry: Oh damn yeah I follow. Basically, now that we default to enabling threshold optimization, even if we mock pipeline `fit`/`predict`/`score`, optimization will still run and consume a bunch of test runtime.

@freddyaboulton FYI this could be relevant for #1815 #2298 

Thanks for explaining @bchen1116 SGTM

evalml/automl/engine/engine_base.py
```diff
@@ -149,6 +157,14 @@ def train_and_score_pipeline(
...
     cv_data = []
     logger.info("\tStarting cross validation")
+    threshold_tuning_objective = automl_config.objective
+    if (
+        is_binary(automl_config.problem_type)
+        and automl_config.optimize_thresholds
+        and automl_config.objective.score_needs_proba
+        and automl_config.alternate_thresholding_objective is not None
+    ):
+        threshold_tuning_objective = automl_config.alternate_thresholding_objective
```
dsherry, line 167: @bchen1116 why is this code necessary? You've already delegated this complexity down into `train_pipeline`! That's called here in `train_and_score_pipeline` below, so all you need to do is provide the automl config to `train_pipeline` unmodified.

dsherry: Update: ah I see you need the threshold tuning objective for a log message. I'm leaving another suggestion about that below

dsherry (approved):
@bchen1116 yep, well done! I left some comments about one last refactor, to minimize duplicated code.

Its great that our binary classification models will now stay tuned
stay_tuned

evalml/automl/engine/engine_base.py
bchen1116: Ah, that was to try and log the threshold properly, but I think I can get rid of it and do a different check!

evalml/automl/engine/engine_base.py
```diff
@@ -194,16 +210,14 @@ def train_and_score_pipeline(
...
-                automl_config.optimize_thresholds,
-                automl_config.objective,
+                pipeline, X_train, y_train, automl_config, schema=False
             )
             logger.debug(f"\t\t\tFold {i}: finished training")
             if (
                 automl_config.optimize_thresholds
-                and pipeline.can_tune_threshold_with_objective(automl_config.objective)
+                and pipeline.can_tune_threshold_with_objective(
+                    threshold_tuning_objective
```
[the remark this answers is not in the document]
bchen1116: @dsherry so rather than printing which fold, we just print that we found an optimal threshold whenever we call `train_pipeline`? This seems like unnecessary log statements if a user just called `train_pipelines` or something else outside of `AutoMLSearch.search()`.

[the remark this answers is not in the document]
dsherry: Yep, that's a good point.

My reasoning: if the choice is between duplicated code and extra debug-level log statements, personally I'd choose the latter, because duplicated code could lead to bugs later.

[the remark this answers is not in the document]
bchen1116: @dsherry I think I'll leave it out for now since `train_pipeline` doesn't include a `logger` parameter, and I think this would be an easy change to address in the future if we wanted to refactor some code.

PR #2344, opened 2021-06-07
Codecov Upload Reports fix

ParthivNaresh:
Fixes issue with Codecov where reports were not being uploaded properly. From now on the CODECOV_TOKEN will be passed to authorize the upload.

### DIFF
--- a/.github/workflows/linux_unit_tests_with_latest_deps.yml
+++ b/.github/workflows/linux_unit_tests_with_latest_deps.yml
@@ -83,4 +83,5 @@ jobs:
         name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
     * Enhancements
         * Removed self-reference from ``AutoMLSearch`` :pr:`2304`
     * Fixes
+        * Pass token to authorize uploading of codecov reports :pr:
     * Changes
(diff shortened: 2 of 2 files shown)


.github/workflows/linux_unit_tests_with_latest_deps.yml
```diff
@@ -83,4 +83,5 @@ jobs:
         name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
dsherry, line 86: So something changed with the `codecov/codecov-action@v1` action which now requires this token, whereas previously we didn't have to provide it?

dsherry (approved):
LGTM on green tests! 🚢 🙏

.github/workflows/linux_unit_tests_with_latest_deps.yml
chukarsten: Correct me if I'm wrong, but the @v1 stuff refers to which version of codecov-action we're checking out, right?  I think, if you're interested in previous versions of codecov action, you can change the 1 to a 2 or also specify the specific version of codecov-action you want to try to see whether the latest requires the token or not.  Still, very weird that it doesn't correspond to a recent release.

chukarsten (approved):

.github/workflows/linux_unit_tests_with_latest_deps.yml
ParthivNaresh: I'm not so sure that it's an issue with the codecov action, it doesn't seem to have released anything recently. It might be related to how secrets are accessed by GitHub Actions?

freddyaboulton (approved):
Thank you @ParthivNaresh !

PR #2345, opened 2021-06-07
AutoML: add "search order" to rankings and full_rankings

### LINKED ISSUE #1841: AutoML: add "search order" to rankings and full_rankings
Our `AutoMLSearch` object has a `search_order` field which indicates the pipeline IDs of the pipelines automl has tried, in the order in which automl tried them.

We should add that as another column to the `rankings` and `full_rankings` tables. That would make it easy for users to match the pipelines up with the progress on the search iteration plot. It could also be a coarse way to show whether or not tuning added value--if the best pipeline(s) have low `search_order` numbers, that means the earlier rounds outperformed the later ones.

angela97lin:
Closes #1841

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1252,6 +1252,7 @@ class AutoMLSearch:
 
         full_rankings_cols = [
             "id",
+            "search_order",
             "pipeline_name",
             "mean_cv_score",
             "standard_deviation_cv_score",
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1259,7 +1259,7 @@ class AutoMLSearch:
             "high_variance_cv",
             "parameters",
         ]
-        full_rankings_cols = ["search_order"] + pipeline_results_cols
+        full_rankings_cols = pipeline_results_cols + ["search_order"]
 
         if not self._results["pipeline_results"]:
             return pd.DataFrame(columns=full_rankings_cols)
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1249,10 +1249,8 @@ class AutoMLSearch:
         ascending = True
         if self.objective.greater_is_better:
             ascending = False
-
-        full_rankings_cols = [
+        pipeline_results_cols = [
             "id",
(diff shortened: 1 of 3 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -160,6 +160,7 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type, obje
                 np.dtype("float64"),
                 np.dtype("bool"),
                 np.dtype("O"),
```
jeremyliweishih, line 162: do any of these tests check if "search_order" has the correct values?

jeremyliweishih (approved):
Looks good but just have one testing question. If that case isn't covered we should add it.

evalml/tests/automl_tests/test_automl.py
angela97lin: @jeremyliweishih Line 109 checks: ``assert automl.results["search_order"] == [0, 1]`` :)

evalml/automl/automl_search.py
```diff
@@ -1262,11 +1261,14 @@ def full_rankings(self):
...
         ]
+        full_rankings_cols = pipeline_results_cols + ["search_order"]
+
         if not self._results["pipeline_results"]:
             return pd.DataFrame(columns=full_rankings_cols)
 
         rankings_df = pd.DataFrame(self._results["pipeline_results"].values())
-        rankings_df = rankings_df[full_rankings_cols]
+        rankings_df = rankings_df[pipeline_results_cols]
+        rankings_df["search_order"] = pd.Series(self._results["search_order"])
```
freddyaboulton, line 1271: Does this mean search order is the last column? Might make more sense to put it next to the pipeline name?

freddyaboulton (approved):
@angela97lin Looks good!

bchen1116 (approved):
LGTM!

evalml/automl/automl_search.py
angela97lin: Yeah :'( we get the other info from pipeline_results so I didn't want to break that up, but I think this makes more sense too

PR #2358, opened 2021-06-09
Fix ComponentGraph appending target to final_component_features if there is a component that returns both X and y

### LINKED ISSUE #2280: ComponentGraph appends target to final_component_features if there is a component that returns both X and y
Repro

```python
cg = ComponentGraph({"Imputer": ["Imputer"],
                     "DateTime": ["DateTime Featurization Component", "Imputer.x"],
                     "OneHot": ["One Hot Encoder", "DateTime.x"],
                     "TargetImputer": ["Target Imputer", "OneHot.x", "OneHot.y"],
                     "Logistic": ["Logistic Regression Classifier", "TargetImputer.x", "TargetImputer.y"]})
cg.instantiate({})
X, y = load_fraud(1000)
cg.fit(X, y)
final_features = cg.compute_final_component_features(X, y)
assert 'TargetImputer.y' in final_features.columns
```

The `TargetImputer.y` is not a "feature" of the final estimator so I would not expect it to show up in the dataframe. This is where we transform [DataColumns](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/component_graph.py#L240) to dataframes so we can concat along the other parent inputs. I think we need this for when we have non-linear pipelines with multiple estimators and we want to concat the predictions before passing to a final estimator (e.g. stacked ensemble). 

I think what we should do is only iterate over the parents that end with `.x` as opposed to both `.x` and `.y`

angela97lin:
Closes #2280

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -238,7 +238,13 @@ class ComponentGraph:
             self.compute_order[:-1], X, y=y, fit=needs_fitting
         )
         final_component_inputs = []
-        for parent in self.get_parents(self.compute_order[-1]):
+
+        parent_inputs = [
+            parent_input
+            for parent_input in self.get_parents(self.compute_order[-1])
+            if parent_input[-2:] != ".y"
+        ]
+        for parent in parent_inputs:
             parent_output = component_outputs.get(
                 parent, component_outputs.get(f"{parent}.x")
             )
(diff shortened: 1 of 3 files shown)


evalml/pipelines/component_graph.py
```diff
@@ -238,7 +238,13 @@ def _fit_transform_features_helper(self, needs_fitting, X, y=None):
             self.compute_order[:-1], X, y=y, fit=needs_fitting
         )
         final_component_inputs = []
-        for parent in self.get_parents(self.compute_order[-1]):
+
+        parent_inputs = [
+            parent_input
+            for parent_input in self.get_parents(self.compute_order[-1])
```
angela97lin, line 244: We default to component name --> pass X so just checking that we're not checking a .y parent input :)

jeremyliweishih (commented):
lgtm

PR #2359, opened 2021-06-09
Minor ``PipelineBase`` cleanup: remove ``_compute_features_during_fit``

angela97lin:
I removed `_compute_features_during_fit` because we called `self.input_feature_names = self.component_graph.input_feature_names` in `_compute_features_during_fit` and then again before returning in the time-series regression pipelines' `fit` method. Moving just these two lines here helps remove the redundancy and avoid bugs with setting the attribute twice.

### DIFF
--- a/evalml/pipelines/time_series_classification_pipelines.py
+++ b/evalml/pipelines/time_series_classification_pipelines.py
@@ -78,7 +78,10 @@ class TimeSeriesClassificationPipeline(
         X, y = self._convert_to_woodwork(X, y)
         self._encoder.fit(y)
         y = self._encode_targets(y)
-        X_t = self._compute_features_during_fit(X, y)
+
+        self.input_target_name = y.name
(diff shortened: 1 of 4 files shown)


evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -78,7 +78,10 @@ def fit(self, X, y):
         X, y = self._convert_to_woodwork(X, y)
         self._encoder.fit(y)
         y = self._encode_targets(y)
-        X_t = self._compute_features_during_fit(X, y)
```
angela97lin, line 81: I thought this was confusing because we called `self.input_feature_names = self.component_graph.input_feature_names` in `_compute_features_during_fit` and then again below! Moving just these two lines here helps remove the redundancy.

jeremyliweishih (approved):
I like it! Allows for easier reading for sure.

bchen1116 (approved):
Good catch!

PR #2365, opened 2021-06-10
Speed up unit tests

### LINKED ISSUE #1815: Unit Test Runtime

freddyaboulton:
### Pull Request Description
Fixes #1815 by speeding up some unit tests and splitting up some of the jobs. 

The changes follow the analysis in this [document](https://ayx.atlassian.net/wiki/spaces/PS/pages/533233674/1815+Speed+Up+Unit+Tests). 

I think this is about a >=70% speedup in the runtime of the tests.

Splitting up the jobs is not necessary. I am including it in this PR mainly to show how it could be done. I prefer to just use a beefier machine if possible rather than adding more checks. 

-----


Makefile
```diff
@@ -22,32 +22,40 @@ test:
...
-git-test:
-	pytest evalml/ -n 2 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure  --ignore=evalml/tests/automl_tests/dask_tests --timeout 300
+test-dask:
+	pytest evalml/tests/automl_tests/dask_tests/ --doctest-modules --doctest-continue-on-failure  --timeout 300 --durations 0
 
 .PHONY: git-test-dask
 git-test-dask:
-	pytest evalml/tests/automl_tests/dask_tests/ -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/ --junitxml=test-reports/junit.xml --doctest-continue-on-failure  --timeout 300
-
-.PHONY: git-test-nocov
```
freddyaboulton, line 36: These commands are no longer used.

evalml/automl/automl_search.py
```diff
@@ -283,7 +283,7 @@ def __init__(
                 "Time series support in evalml is still in beta, which means we are still actively building "
                 "its core features. Please be mindful of that when running search()."
             )
-
+        self._SLEEP_TIME = 0.1
```
freddyaboulton, line 286: One of the things I found is that the sleep time increases the run time for some unit tests that run many iterations. 

This is a quick way to manipulate the amount of sleep time for some tests but it's not the only one! Let me know if you think there's a better way.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -394,25 +394,25 @@ def round_dict_keys(dictionary, places=6):
...
-        0.14487916687594762: 1,
-        0.1573183451314127: 1,
-        0.11695462432136654: 1,
-        0.07950579532536253: 1,
-        0.006794444792966759: 1,
-        0.17745270478939879: 1,
-        0.1666874487986626: 1,
-        0.13357573073236878: 1,
-        0.06778096366056789: 1,
+        0.05824028901694482: 63,
```
freddyaboulton, line 405: Since the data we use is different, the expected partial dependence is now different too

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1496,9 +1513,10 @@ def test_describe_pipeline(mock_fit, mock_score, return_dict, caplog, X_y_binary
 
 @patch("evalml.pipelines.BinaryClassificationPipeline.score")
 @patch("evalml.pipelines.BinaryClassificationPipeline.fit")
+@patch("evalml.tuners.skopt_tuner.Optimizer.tell")
```
freddyaboulton, line 1516: Surprisingly, `Optimizer.tell` uses up a lot of the time for tests that run automl for many iterations.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -96,14 +97,14 @@ def test_fit_predict_ts_with_datetime_in_X_column(ts_data_seasonal):
     assert isinstance(y.index, pd.DatetimeIndex)
 
     m_clf = ARIMARegressor(d=None)
-    m_clf.fit(X=X[:250], y=y[:250])
-    y_pred = m_clf.predict(X=X[250:])
+    m_clf.fit(X=X[:25], y=y[:25])
```
freddyaboulton, line 100: I'm surprised this test took ~60 seconds since the total data size is 500 rows. 

I think we should file a separate issue to see where the bottleneck is.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -887,25 +887,28 @@ def test_partial_dependence_datetime(
 
     X = pd.DataFrame(X, columns=[str(i) for i in range(X.shape[1])])
     y = pd.Series(y)
-    X["dt_column"] = pd.Series(pd.date_range("20200101", periods=X.shape[0]))
-
+    random_dates = pd.Series(pd.date_range("20200101", periods=10)).sample(
```
freddyaboulton, line 890: In order to speed up these tests, I used fewer unique dates. I think this points to how our approach for handling dates does not scale super well. I think we should revisit this in the near future.

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -441,7 +477,7 @@ def test_get_permutation_importance_binary(
         parameters={"Logistic Regression Classifier": {"n_jobs": 1}}, random_seed=42
     )
     pipeline.fit(X, y)
-    for objective in binary_core_objectives:
+    for objective in ["Log Loss Binary", "AUC"]:
```
freddyaboulton, line 480: Running fewer objectives and columns

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -305,15 +307,15 @@ def test_fit_ts_without_y(ts_data):
...
 
     a_clf = sktime_arima.AutoARIMA()
-    a_clf.fit(y=y[:250])
+    a_clf.fit(y=y[:25])
     y_pred_sk = a_clf.predict(fh=fh_)
 
     m_clf = ARIMARegressor(d=None)
-    m_clf.fit(X=None, y=y[:250])
-    y_pred = m_clf.predict(X=None, y=y[250:])
+    m_clf.fit(X=None, y=y[:25])
```
chukarsten, line 317: Yea, I'm thinking just change the fixture to be 25 rows

evalml/tests/model_understanding_tests/test_permutation_importance.py
chukarsten: Curious, would it be better to replace `binary_core_objectives` with something like a `binary_test_objectives`, then we could just reduce the load on all the tests at once.  Additionally, is the fixture used after you pull it out?  Seems kinda weird that lint didn't pick it up...

chukarsten (approved):
I really like these changes and appreciate you taking the time to trudge through them.  I think there are a few changes that can move into fixtures, which would remove a few hardcoded values from the code base.  Also maybe start to pull more focus onto conf.py as a lever to turn up thoroughness of tests.

evalml/tests/component_tests/test_arima_regressor.py
angela97lin: Also maybe super optimization / nitpick but rather than having to update the number every time we want to change how much data we use, could be useful to set a variable... just in case we later decide we don't like the number 25 or something :3

evalml/tests/model_understanding_tests/test_partial_dependence.py
angela97lin: Honestly, I'm still not sure I understand the usefulness of checking against `part_dep_ans`... how are we getting these numbers anyhow 🤔

evalml/tests/model_understanding_tests/test_partial_dependence.py
angela97lin: Good to file an issue? :)

angela97lin (approved):
This looks great, and is so, so thorough!! Mostly nit-picky comments, but I'd like to revisit our idea to use `fail-fast` as False. I think it could be better to keep that as true and move the dask tests outside the matrix, if possible, given the increased number of jobs and the number of concurrent jobs we can have. Curious about your thoughts on that, but otherwise, thank you for doing this!! The increased runtimes are super exciting. 😁

evalml/automl/automl_search.py
freddyaboulton: This is the [thread](https://github.com/alteryx/evalml/pull/1975#discussion_r605132883) where we discussed the sleep in the dask pr. 

We picked 0.1 just cause it seemed like a good number. Decreasing the sleep time will not change the functionality especially since these tests use the sequential engine.

evalml/tests/model_understanding_tests/test_partial_dependence.py
freddyaboulton: Will do!

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Word I'd be down for changing the fixture! I just wanted to change the slowest tests first. What are good candidates for `binary_test_objectives`?

Makefile
```diff
@@ -22,32 +22,40 @@ test:
...
+git-test-automl:
+	pytest evalml/tests/automl_tests evalml/tests/tuner_tests -n 2 --ignore=evalml/tests/automl_tests/dask_tests --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
+
+.PHONY: git-test-modelunderstanding-core
+git-test-modelunderstanding-core:
+	pytest evalml/tests/model_understanding_tests -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies
+
+.PHONY: git-test-modelunderstanding
+git-test-modelunderstanding:
+	pytest evalml/tests/model_understanding_tests -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
```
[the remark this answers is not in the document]
freddyaboulton: I think we should measure against all evalml so that way we count the code that's defined outside of `model_understanding` that's run in the tests, e.g. pipelines, components, utils

evalml/tests/automl_tests/test_automl.py
```diff
@@ -389,6 +403,7 @@ def test_rankings(X_y_binary, X_y_regression):
         max_iterations=3,
         n_jobs=1,
     )
+    automl._SLEEP_TIME = 0.001
```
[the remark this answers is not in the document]
freddyaboulton: My thought was just to modify the slowest tests. Most tests run few iterations so the sleep time doesn't contribute that much to the total time. I do think it'd be nice to leave the `0.1` value in `AutoMLSearch` but then define a fixture or something that sets it small for all tests. I'm not sure what the easiest way to do that is though and that would require modifying all evalml unit tests to add the fixture.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -86,8 +86,9 @@ def test_format_dates(predict, dates_shape, ts_data):
 def test_feature_importance(ts_data):
     X, y = ts_data
     clf = ARIMARegressor()
-    clf.fit(X, y)
-    clf.feature_importance == np.zeros(1)
+    with patch.object(clf, "_component_obj"):
+        clf.fit(X, y)
+        clf.feature_importance == np.zeros(1)
```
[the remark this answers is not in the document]
freddyaboulton: Eagle eyes. Thank you hehehe

evalml/tests/component_tests/test_arima_regressor.py
freddyaboulton: I changed the fixture to only have 50 rows but the test still manually splits the data into two halves for fitting and predicting. So that's why the 25 is there. Do you want me to define another fixture to represent the second half of the data?

.github/workflows/linux_unit_tests_with_dask.yml
```diff
@@ -1,87 +0,0 @@
-name: Linux Unit Tests (Dask Only)
```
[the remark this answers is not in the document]
freddyaboulton: Correct.

.github/workflows/linux_unit_tests_with_latest_deps.yml
```diff
@@ -9,24 +9,76 @@ on:
 
 jobs:
   unit_tests:
-    name: ${{ matrix.python_version }} unit tests, Core Dependencies (${{ matrix.core_dependencies }}), Codecov (${{ matrix.codecov }})
+    name: ${{ matrix.python_version }} unit ${{matrix.dir}} tests, Core Dependencies (${{ matrix.core_dependencies }}), Codecov (${{ matrix.codecov }})
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
```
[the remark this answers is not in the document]
freddyaboulton: Do you just want to set `fail-fast` to `true` for all jobs? The thing I don't like about moving tests outside the test matrix is that it leads to a lot of duplicated code across the yamls because the install steps are the same. There may be a way to reuse code across actions but I'd rather just `fail-fast` to alleviate concerns over too many jobs running.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4539,8 +4576,8 @@ def test_automl_issues_beta_warning_for_time_series(problem_type, X_y_binary):
     "evalml.pipelines.BinaryClassificationPipeline.score",
     return_value={"Log Loss Binary": 0.3},
 )
-@patch("evalml.automl.engine.sequential_engine.train_pipeline")
-def test_automl_drop_index_columns(mock_train, mock_binary_score, X_y_binary):
+@patch("evalml.pipelines.BinaryClassificationPipeline.fit")
```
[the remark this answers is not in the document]
freddyaboulton: I don't think this was actually not mocking the pipeline fit before!

evalml/tests/model_understanding_tests/test_permutation_importance.py
ParthivNaresh: I think F1, Log Loss, and AUC could be good choices

ParthivNaresh (approved):
Looking great! The execution time looks so much better now!

.github/workflows/linux_nightlies.yml
```diff
@@ -0,0 +1,46 @@
+name: Nightly Linux Unit Tests
+
+on:
+  schedule:
```
freddyaboulton, line 4: These are what the nightlies look like. Before merge I'll remove the `on: pull_request` and update the `cron` to run at 3 am. I just added the `on: pull_request` to test it out. I'll also change the `ref` of the github action to say `main`.

So if this pr is merged, this is the expected behavior:

### On pushes to prs and merges to main run: 
- linux unit tests with latest versions python=3.8, core deps [True, False]. Measure coverage and upload to codecov.
- windows unit tests with python 3.8, core deps False. Don't measure coverage.
- linux unit tests with minimum versions with python 3.7, core deps False. Don't measure coverage.

### At 3am
- linux unit tests with latest versions with python [3.7, 3.9], core deps False. Don't measure coverage.
- windows unit tests with python [3.7, 3.9], core deps False. Don't measure coverage.
- Not running minimum dependency versions as a nightly because some of the minimum package versions are not available for 3.9! I think just testing minimum versions with 3.7 is fine (and that's what we do now).


Sound good? The total number of unit test checks would be 17. FYI @chukarsten @dsherry

.github/workflows/linux_nightlies.yml
```diff
@@ -0,0 +1,46 @@
...
+      - cron: '10 * * * *'
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  unit_tests:
+    name: Nightly ${{ matrix.python_version }} ${{matrix.command}}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
freddyaboulton, line 14: Setting `fail-fast` to be false so that we know what all the failures are when they occur.

.github/workflows/linux_unit_tests_with_minimum_deps.yml
```diff
@@ -9,16 +9,36 @@ on:
...
+    name: ${{ matrix.python_version }} min dep ${{ matrix.command }}, core deps (${{ matrix.core_dependencies }})
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         include:
           - python_version: "3.7"
             core_dependencies: false
+            command: 'git-test-automl'
```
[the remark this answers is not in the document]
freddyaboulton: No reason! `dir` is a leftover from a previous way I tried to parametrize the matrix. I changed them all to use `command` now.

evalml/tests/component_tests/test_arima_regressor.py
chukarsten: I think that might be best @freddyaboulton

evalml/tests/automl_tests/test_automl.py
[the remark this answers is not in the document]
chukarsten: Whatever you think is best!  I think it's weird having the sleep time to begin with, so without any context on why we have it to begin with, my natural inclination is to just drop it for everyone in the lowest touch way possible.  You ultimately make the decision on how you want to handle this, it's not blocking!

evalml/tests/component_tests/test_arima_regressor.py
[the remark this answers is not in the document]
chukarsten: lol you're just moving into the context manager....it was like that before you touched it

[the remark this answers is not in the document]
freddyaboulton: 😂  Anyways, should be an `assert` now.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2579,6 +2604,7 @@ def test_automl_one_allowed_pipeline_ensembling_disabled(
         ensembling=True,
     )
     automl.search()
+    automl._SLEEP_TIME = 0.0001
```
dsherry, line 2607: @freddyaboulton do you think there's a way to define a fixture for automl tests which handles stuff like `_SLEEP_TIME`, mocking pipeline methods and stuff like `Optimizer.tell`, etc?

freddyaboulton: @dsherry Good point! Per our discussion at stand-up today, I'll merge this as-is but then look into a way we can package our best practices into a fixture!

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -473,7 +509,7 @@ def test_get_permutation_importance_multiclass(
         parameters={"Logistic Regression Classifier": {"n_jobs": 1}}, random_seed=42
     )
     pipeline.fit(X, y)
-    for objective in multiclass_core_objectives:
+    for objective in ["Log Loss Multiclass", "AUC Micro"]:
```
bchen1116, line 512: Would be nice to do something similar to what @chukarsten pointed out here with `multiclass_test_objectives`

bchen1116 (approved):
Nice cleanup! Changes look good to me!

evalml/tests/automl_tests/test_automl.py
dsherry: Sounds great! Sorry for any confusion, this comment is definitely not blocking this PR, and we can discuss after this PR is merged.

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Done!

evalml/tests/model_understanding_tests/test_permutation_importance.py
freddyaboulton: Done!

evalml/tests/component_tests/test_arima_regressor.py
freddyaboulton: Done!

PR #2372, opened 2021-06-10
Fix partial dependence graph method failing on multiclass problems when the label is numeric

### LINKED ISSUE #2114: Partial Dependence Graph fails on multiclass problems when the label is numeric
Repro

```python
from sklearn import datasets
from evalml.pipelines import MulticlassClassificationPipeline
from evalml.model_understanding import graph_partial_dependence
import pytest

X, y = datasets.make_classification(n_samples=100, n_features=20, n_classes=3,
                                        n_informative=3, n_redundant=2, random_state=0)

class Pipeline(MulticlassClassificationPipeline):

pl = Pipeline({})
pl.fit(X, y)

with pytest.raises(ValueError,
                   match="Invalid value of type 'numpy.int64' received for the 'text' property of layout.annotation"):
    graph_partial_dependence(pl, X, features=2)
```

Converting the target to strings fixes this problem:

image

angela97lin:
Closes #2114

### DIFF
--- a/evalml/tests/model_understanding_tests/test_partial_dependence.py
+++ b/evalml/tests/model_understanding_tests/test_partial_dependence.py
@@ -287,6 +287,45 @@ def test_partial_dependence_multiclass(logistic_regression_multiclass_pipeline_c
     assert len(two_way_part_dep.columns) == grid_resolution + 1
 
 
+def test_partial_dependence_multiclass_numeric_labels(
+    logistic_regression_multiclass_pipeline_class, X_y_multi
+):
+    X, y = X_y_multi
+    X = pd.DataFrame(X)
+    y = pd.Series(y, dtype="int64")
+    pipeline = logistic_regression_multiclass_pipeline_class(
+        parameters={"Logistic Regression Classifier": {"n_jobs": 1}}
+    )
+    pipeline.fit(X, y)
+
+    num_classes = y.nunique()
+    grid_resolution = 20
+
+    one_way_part_dep = partial_dependence(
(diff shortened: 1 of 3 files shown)


evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -287,6 +287,45 @@ def test_partial_dependence_multiclass(logistic_regression_multiclass_pipeline_c
     assert len(two_way_part_dep.columns) == grid_resolution + 1
 
 
+def test_partial_dependence_multiclass_numeric_labels(
```
angela97lin, line 290: @freddyaboulton I'm sure you might have some opinions about me adding yet another long running test 😂 I wasn't confident in a way to mock things but definitely would love suggestions/opinions!

freddyaboulton: What's the runtime now? My takeaway is that using `grid_resolution=5` can trim the test time to a couple seconds down from like ~10. The next step would be using fewer rows.

chukarsten (approved):
Looks good.  I think dropping the grid res is a good idea to make it a shorter test.

evalml/tests/model_understanding_tests/test_partial_dependence.py
angela97lin: ~30-40 seconds before, 8 seconds after setting grid_resolution to 5 😁

freddyaboulton (approved):

PR #2375, opened 2021-06-11
Fix k_neighbors issue with SMOTE oversamplers

### LINKED ISSUE #2324: Fix SMOTE Oversampler k_neighbors arg
In some imbalanced tests, SMOTE ran into errors where the default `k_neighbors` value was too high (below)

We had initially chosen to use the default `k_neighbors` value provided by SMOTE, but it might be useful to lower this default. This issue tracks creating a fix for this issue, whether that be lowering the value and providing a better error message, or something else. 

**Repro**
```python
X_train, y_train = load_fraud(n_rows=100)
automl = AutoMLSearch(X_train=X_train, y_train=y_train, problem_type='binary')
automl.search()
```

image
image

bchen1116:
fix #2324 

Design doc [here](https://ayx.atlassian.net/l/c/Q6nwPL99)


evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -118,25 +118,26 @@ def __init__(
         sampler,
         sampling_ratio=0.25,
         sampling_ratio_dict=None,
-        k_neighbors=5,
+        k_neighbors_default=5,
         n_jobs=-1,
         random_seed=0,
-        **kwargs
+        **kwargs,
```
bchen1116, line 124: this got added by black

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -170,20 +171,32 @@ def _initialize_oversampler(self, X, y, sampler_class):
...
             sampler_class (imblearn.BaseSampler): The sampler we want to initialize
         """
-        _, y_ww = self._prepare_data(X, y)
+        _, y_pd = self._prepare_data(X, y)
         sampler_params = {
             k: v
             for k, v in copy.copy(self.parameters).items()
-            if k not in ["sampling_ratio", "sampling_ratio_dict"]
+            if k not in ["sampling_ratio", "sampling_ratio_dict", "k_neighbors_default"]
         }
```
chukarsten, line 179: Do we need the copying here?

```
sampler_params = {k: self.parameters[k] for k  ["sampling_ratio", "sampling_ratio_dict", "k_neighbors_default"]}
```

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -341,3 +341,29 @@ def test_oversampler_sampling_dict_strings(oversampler):
...
     assert len(new_X) == sum(expected_result.values())
     assert new_y.value_counts().to_dict() == expected_result
+
+
+@pytest.mark.parametrize("oversampler", [SMOTESampler, SMOTENCSampler, SMOTENSampler])
+@pytest.mark.parametrize(
+    "minority,expected,fails",
+    [(1, 0, True), (2, 1, False), (5, 4, False), (10, 5, False)],
+)
+def test_oversampler_sampling_k_neighbors(minority, expected, fails, oversampler):
```
chukarsten, line 351: Cool, I like the test.  It looks good.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -170,20 +171,32 @@ def _initialize_oversampler(self, X, y, sampler_class):
...
+
+        # check for k_neighbors value
+        neighbors = self.parameters["k_neighbors_default"]
+        min_counts = y_pd.value_counts().values[-1]
+        if min_counts <= 1:
+            raise ValueError(
+                f"Minority class needs more than 1 sample to use SMOTE!, received {min_counts} sample"
+            )
+        if min_counts <= neighbors:
+            neighbors = min_counts - 1
```
chukarsten, line 197: I like the solution.  I just would prefer if the user knew what was happening.  Do we have a way that we can communicate to the user, who thinks that k_neighbors_default is one value that it's being reduced due to their input?

chukarsten (approved):
Love it - it's quick, intuitive solution with a good test.  Is there anyway we can communicate what's happening to the user?  I'm still not sure how we're using warnings or logging, but I'd prefer, philosophically, for us to let the user know everytime we change the behavior they expect under the covers.

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: I could introduce a `k_neighbors` parameter that becomes the value we set it to in the parameters dictionary? This way, users can see the parameter when they look at the param dic, but I don't think our other components currently log any changes they make internally. We don't raise any changes [here](https://github.com/alteryx/evalml/blob/0ec9206c3d8f3214404782b3c280bd1798562878/evalml/pipelines/components/estimators/classifiers/catboost_classifier.py#L64) and [here](https://github.com/alteryx/evalml/blob/0ec9206c3d8f3214404782b3c280bd1798562878/evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py#L77).

Does this sound fine to you @chukarsten?

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: I think we technically don't, I had just added it to be safer. I can remove it though!

evalml/pipelines/components/transformers/samplers/base_sampler.py
chukarsten: Uh, how about just a change to a docstring on the k_neighbors_default parameter?  I think that would be sufficient.

chukarsten (approved):

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -118,25 +118,26 @@ def __init__(
...
+        **kwargs,
     ):
         """Initializes the oversampler component.
 
         Arguments:
             sampling_ratio (float): This is the goal ratio of the minority to majority class, with range (0, 1]. A value of 0.25 means we want a 1:4 ratio
                 of the minority to majority class after oversampling. We will create the a sampling dictionary using this ratio, with the keys corresponding to the class
                 and the values responding to the number of samples. Defaults to 0.25.
-            k_neighbors (int): The number of nearest neighbors to used to construct synthetic samples. Defaults to 5.
+            k_neighbors_default (int): The number of nearest neighbors to used to construct synthetic samples. This value will become smaller when there are less samples
```
angela97lin, line 132: Typo: to used --> to use

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -170,20 +171,33 @@ def _initialize_oversampler(self, X, y, sampler_class):
             y (pd.Series): Target features
             sampler_class (imblearn.BaseSampler): The sampler we want to initialize
         """
-        _, y_ww = self._prepare_data(X, y)
+        _, y_pd = self._prepare_data(X, y)
         sampler_params = {
             k: v
-            for k, v in copy.copy(self.parameters).items()
-            if k not in ["sampling_ratio", "sampling_ratio_dict"]
+            for k, v in self.parameters.items()
```
angela97lin, line 177: Do you know why we copied before and why its okay that we remove this?

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -170,20 +171,33 @@ def _initialize_oversampler(self, X, y, sampler_class):
...
             # create the sampling dictionary
             sampling_ratio = self.parameters["sampling_ratio"]
-            dic = make_balancing_dictionary(y_ww, sampling_ratio)
+            dic = make_balancing_dictionary(y_pd, sampling_ratio)
         sampler_params["sampling_strategy"] = dic
+
+        # check for k_neighbors value
+        neighbors = self.parameters["k_neighbors_default"]
+        min_counts = y_pd.value_counts().values[-1]
+        if min_counts <= 1:
```
angela97lin, line 192: Nit pick, but this could only be... one, right? Otherwise it wouldn't show up in `value_counts` to begin with?

evalml/pipelines/components/transformers/samplers/base_sampler.py
angela97lin: I think we could make this docstring more clear that this is the default used, but that the actual k_neighbors value used may be different if there are less samples? Currently a bit confusing to me or odd that it sounds like the parameter value will decrease :')

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1495,6 +1495,28 @@ def test_automl_search_sampler_k_neighbors_param(sampler, has_minimal_dependenci
...
+@pytest.mark.parametrize(
+    "parameters", [None, {"SMOTENC Oversampler": {"k_neighbors_default": 5}}]
+)
+def test_automl_search_sampler_k_neighbors_no_error(
+    parameters, has_minimal_dependencies, fraud_100
+):
+    # automatically uses SMOTE
+    if has_minimal_dependencies:
+        pytest.skip("Skipping tests since imblearn isn't installed")
+    # split this from the undersampler since the dictionaries are formatted differently
```
angela97lin, line 1512: are these docstrings correct? I'm confused what they refer to 😕

evalml/tests/component_tests/test_oversamplers.py
```diff
@@ -341,3 +341,30 @@ def test_oversampler_sampling_dict_strings(oversampler):
...
+    X_ww = infer_feature_types(X, feature_types={0: "Categorical", 1: "Categorical"})
+    y = np.array(["minority"] * minority + ["majority"] * (1000 - minority))
+    overs = oversampler(k_neighbors_default=5)
+    if fails:
+        with pytest.raises(
+            ValueError, match="Minority class needs more than 1 sample to use SMOTE"
+        ):
+            overs.fit_transform(X_ww, y)
+        return
+    _, _ = overs.fit_transform(X_ww, y)
```
angela97lin, line 368: nit pick, if we're not testing either output then maybe just calling `overs.fit_transform(X_ww, y)` is fine?

angela97lin (approved):
LGTM! left some nitpicking :]

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: Hahah yeah, this is true!

evalml/pipelines/components/transformers/samplers/base_sampler.py
bchen1116: This was something Karsten pointed out, but since we're making a copied dictionary from the parameters, I don't think it's necessary for us to copy the dictionary.

evalml/tests/automl_tests/test_automl_search_classification.py
bchen1116: Ah, mistake from copying!

evalml/pipelines/components/transformers/samplers/base_sampler.py
chukarsten: We do a lot of copying throughout the code base....I feel like it *might* be someone (rightfully) scared of overwriting the originals.

PR #2376, opened 2021-06-11
Separate dask job

chukarsten:


Makefile
```diff
@@ -19,23 +19,35 @@ lint-fix:
 
 .PHONY: test
 test:
-	pytest evalml/ --doctest-modules --doctest-continue-on-failure
+	pytest evalml/ --doctest-modules --doctest-continue-on-failure --ignore=evalml/tests/automl_tests/dask_tests
+
+.PHONY: test-dask
+test:
+	pytest evalml/tests/automl_tests/dask_tests/ --doctest-modules --doctest-continue-on-failure
```
freddyaboulton, line 26: Can you add the `--timeout 360` ?

Makefile
```diff
@@ -19,23 +19,35 @@ lint-fix:
 
 .PHONY: test
 test:
-	pytest evalml/ --doctest-modules --doctest-continue-on-failure
+	pytest evalml/ --doctest-modules --doctest-continue-on-failure --ignore=evalml/tests/automl_tests/dask_tests
```
freddyaboulton, line 22: I'd rather we still had one command to run all tests but this is not blocking merge.

.github/workflows/windows_unit_tests_with_dask.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+
+jobs:
+  win_unit_tests:
```
freddyaboulton, line 11: I'll file an issue for trying to reduce as many of the set-up steps as possible between the dask-only and non-dask tests. I think that'll pay off in the future.

freddyaboulton (approved):
Thanks @chukarsten !

Makefile
chukarsten: I'll do a quick PR to reintroduce this.  Good point

Issue #2341, opened 2021-06-04
Flaky Dask test


freddyaboulton:
I've seen the `test_automl_immediate_quit` dask test flake intermittently on my prs, on a [merge to main](https://github.com/alteryx/evalml/runs/2738729795), and on #2317 [link](https://github.com/alteryx/evalml/pull/2317/checks?check_run_id=2749351258). 

Rerunning the tests usually makes the failure go away but that's not ideal.

The expected outcome of this issue is an explanation of why the test is flaking and/or a solution to make the test not flake anymore!

chukarsten:
I can't seem to recover the specific test runs but this *seems* to always flake on Python 3.7.  I saw it at least 4 times flaking on 3.7.

chukarsten:
Just to add some more information on this, `build_conda_pkg` has failed on Python 3.7 on `test_submit_training_job_single()`, [here](https://github.com/alteryx/evalml/pull/2370/checks?check_run_id=2830544484).

chukarsten:
[Here](https://github.com/alteryx/evalml/pull/2382/checks?check_run_id=2832272083) is an instance of the flake happening on Python 3.8.  First time I've seen it on this version, so I guess the 3.7/dask version disagreement is out.

Issue #2354, opened 2021-06-09
Unit Test Timeouts (Dask Instability)


chukarsten:
We're currently seeing unit tests go to the GH Actions limit of 6hrs.  This is not good for obvious reasons.

[3.8 core deps](https://github.com/alteryx/evalml/pull/2320/checks?check_run_id=2784353628) 6 hr timeout (in progress)
[build_conda_pkg, 3.8 core deps, 3.7 non-core deps](https://github.com/alteryx/evalml/actions/runs/920331123) 6 hr timeout (in progress)
[3.7 non-core deps](https://github.com/alteryx/evalml/actions/runs/919587230) 6 hr timeout
[3.8 non-core deps](https://github.com/alteryx/evalml/runs/2778007146?check_suite_focus=true) 6 hr timeout
[3.7 non-core deps](https://github.com/alteryx/evalml/pull/2332/checks?check_run_id=2776014882) 1.5 hrs
[build_conda_pkg](https://github.com/alteryx/evalml/actions/runs/923208327)
[3.7 non-core deps](https://github.com/alteryx/evalml/actions/runs/926406335)
[3.8](https://github.com/alteryx/evalml/actions/runs/926403320)

chukarsten:
[Here's](https://github.com/alteryx/evalml/pull/2368/checks?check_run_id=2795432470) another for 3.9 non-core deps.

[github_unittests_2.txt](https://github.com/alteryx/evalml/files/6633702/github_unittests_2.txt)

dsherry:
Thanks for filing @chukarsten 

Thankfully we can rule out conda as a cause, since this happens for our normal unit test builds and not just for `build_conda_pkg`

Is there any other info we should collect which could help us figure this out? A few ideas below
* How reliably can we repro the timeout? Is it happening 50% of the time we run unit test job, more, less?
* Which test or tests is not completing properly? If we can get pytest to log the start and end of each test, we can look at the logs and deduce which test has not ended when the hanging occurs. [This](https://stackoverflow.com/questions/61437320/pytest-finding-when-each-test-started-and-ended) looked potentially useful.
* Do we still see these timeouts if we run tests without any pytest parallelization?
* This is just a hunch, but what happens if we disable the dask engine tests? I know we've seen some flakes with those recently #2341 
* What does CPU and memory utilization look like while the tests are running?

dsherry:
(@freddyaboulton I added you on here since this connects to #2298 and #1815)

[...]

PR #2377, opened 2021-06-11
Refactor dask tests

freddyaboulton:
### Pull Request Description
Refactor dask tests to not reuse a client across tests. Also use the "standard" pytest workflow of writing functions instead of classes. This was inspired by how lightgbm tests dask. 

https://github.com/microsoft/LightGBM/pull/4159/files

I pushed a commit every three hours over the weekend and the tests always passed. Note the following:

1. A lot of the red x's are because the `codecov/patch` check  did not pass (I believe this is a "bug" stemming from the fact there was no diff over repeated commits. The issue fixed itself though).

2. This does not fix this issue https://github.com/alteryx/evalml/issues/2341, and it's why the windows unit tests failed for commit `5d57ebb`. However, `dask_sends_woodwork_schema` did not fail. 

I still do not have a 100% confident explanation of what was causing the original error. But I don't think this was a gh problem after all since I was finally able to repro the timeout locally. 

My hand-wavy explanation is that the dask scheduler somehow is losing track of the data we've scattered. Since the [client is the scheduler](https://distributed.dask.org/en/latest/client.html#client) and we're reusing the same client across the test, it made me think that some global state was being corrupted. I saw that lightgbm and xgboost do not reuse clients across tests and it gave me the idea to try it here.

I filed an issue with `dask/distributed` as well https://github.com/dask/distributed/issues/4907

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Release Notes
     * Documentation Changes
     * Testing Changes
         * Add ``pytest-timeout``. All tests that run longer than 6 minutes will fail. :pr:`2374`
+        * Refactored dask tests :pr:`2377`
 
 .. warning::
 
--- a/evalml/tests/automl_tests/test_automl_dask.py
+++ b/evalml/tests/automl_tests/test_automl_dask.py
@@ -1,8 +1,6 @@
-import unittest
 
 import numpy as np
 import pytest
-from distributed import Client
 
 from evalml.automl import AutoMLSearch
 from evalml.automl.callbacks import raise_error_callback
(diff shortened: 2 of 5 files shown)


docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Release Notes
     * Documentation Changes
     * Testing Changes
         * Add ``pytest-timeout``. All tests that run longer than 6 minutes will fail. :pr:`2374`
+        * Refactored dask tests :pr:`2377`
```
angela97lin, line 10: omega nit-pick but maybe include why so that when we try to trace this back in the future we know? :P

angela97lin (approved):
LGTM, I like this refactor a lot and it aligns more with how our other unit tests work too. And of course, the almost-consistently green CI makes me happy 😁

chukarsten:
We should probably standby for a little bit to see whether turning off the broadcast option helps, like in PR #2376.  I've been tracking some of the speculation and resolution on the source issue, #2354 .  The dask issue that led me to try turning off broadcast was [here](https://github.com/dask/distributed/issues/2880).  I'd be hesitant to move away from the single client for testing only because it should work that way, haha.

bchen1116 (approved):
Nice changes!

PR #2379, opened 2021-06-14
Upgrade WW to 0.4.1

### LINKED ISSUE #2327: Upgrade to Woodwork v0.4.1
- We released a [new version](https://woodwork.alteryx.com/en/latest/release_notes.html#v0-4-0-may-26-2021) of Woodwork (with no breaking changes).

jeremyliweishih:
Fixes #2327.

This PR mostly consists of fixing unit tests and functionality where we deal with WW logical types since they are now instantiated instead of returning the class.

Perf tests look good with no changes in model performance and somelittle change in init time and fit time but all within a couple percentage points. Ran with 50 iterations and 3 trials.

[ww_upgrade_perf_results.zip](https://github.com/alteryx/evalml/files/6687309/ww_upgrade_perf_results.zip)


core-requirements.txt
```diff
@@ -11,7 +11,7 @@ psutil>=5.6.3
 requirements-parser>=0.2.0
 shap>=0.36.0
 texttable>=1.6.2
-woodwork==0.3.1
+woodwork==0.4.0
```
gsheni, line 14: I think you mean 0.4.1

jeremyliweishih: I'm taking it one step at a time!

evalml/tests/component_tests/test_datetime_featurizer.py
```diff
@@ -299,7 +299,7 @@ def test_datetime_featurizer_woodwork_custom_overrides_returned_by_components(
         try:
             X = X_df.copy()
             X.ww.init(logical_types={0: logical_type})
-        except (ww.exceptions.TypeConversionError, TypeError):
+        except (ww.exceptions.TypeConversionError, TypeError, ValueError):
```
jeremyliweishih, line 302: will be fixed by https://github.com/alteryx/woodwork/issues/991

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -80,14 +80,14 @@ def validate(self, X, y):
             return results
 
         y = infer_feature_types(y)
-        is_supported_type = y.ww.logical_type in numeric_and_boolean_ww + [
+        is_supported_type = type(y.ww.logical_type) in numeric_and_boolean_ww + [
```
tamargrey, line 83: I think this can be `isinstance(y.ww.logical_type, tuple(numeric_and_boolean_ww))` if you'd like here and in the other places where a similar check.

evalml/model_understanding/graphs.py
```diff
@@ -489,9 +489,9 @@ def graph_binary_objective_vs_threshold(pipeline, X, y, objective, steps=100):
 def _is_feature_of_type(feature, X, ltype):
     """Determine whether the feature the user passed in to partial dependence is a Woodwork logical type."""
     if isinstance(feature, int):
-        is_type = X.ww.logical_types[X.columns[feature]] == ltype
+        is_type = type(X.ww.logical_types[X.columns[feature]]) == ltype
     else:
-        is_type = X.ww.logical_types[feature] == ltype
+        is_type = type(X.ww.logical_types[feature]) == ltype
```
tamargrey, line 494: These are also places where we can use `isinstance`, which I think is slightly better from a timing perspective than `==` checks.

evalml/utils/woodwork_utils.py
```diff
@@ -101,6 +103,9 @@ def _retain_custom_types_and_initalize_woodwork(
         return ww.init_series(new_dataframe, old_logical_types)
     if ltypes_to_ignore is None:
         ltypes_to_ignore = []
+    old_logical_types = {
+        k: v if inspect.isclass(v) else type(v) for k, v in old_logical_types.items()
```
freddyaboulton, line 107: Do need `isclass` here? The types should always be instances from now on right?

evalml/data_checks/invalid_targets_data_check.py
freddyaboulton: @tamargrey I think the other option is to use `y.ww.logical_type.type_string` and then update our lists, equality/member checks to use the `type_strings`. What are your thoughts? I think both do the same thing but using `type_string` may be backwards compatible?

freddyaboulton (approved):
@jeremyliweishih Thanks for this! I think it looks great. I have a question on the diff for `_retain_woodwork_types...` and I think it may be better to follow @tamargrey 's suggestion to use `is_instance` or my suggestion to use `type_string` rather than wrapping everything in `type`. I'm ok with either of those options but @tamargrey 's opinion should take precedence over mine hehe.

freddyaboulton:
FYI I'll file a follow-up issue to use the new functionality added in this release:

```
Add option to return TableSchema instead of DataFrame from table accessor select method (#916)
Add dropping and renaming columns inplace (#920)
```

jeremyliweishih:
@freddyaboulton can you also add using `concat_columns` to that issue?

evalml/utils/woodwork_utils.py
jeremyliweishih: IIRC some tests were failing without this change but I'll double check this!

evalml/data_checks/invalid_targets_data_check.py
tamargrey: @freddyaboulton I like that idea! It does seem like it would be backwards compatible, which is nice.

It's definitely different from how we've tended to compare logical types inside of Woodwork, so my one worry is that it's using the `type_string` for something that it wasn't intended for. I'm not sure that that's actually a problem, though.

dsherry:
Taking a slight calculated risk and merging this before the docs build completes. We have a lot queued right now. I hope #2404 will help us here!

PR #2386, opened 2021-06-16
Added support for Individual Conditional Expectation plots to `graph_partial_dependencies()`

### LINKED ISSUE #2025: Adding Individual Conditional Expectation (ICE) Plots to model_understanding
Alongside `partial_dependence`, I would like to be able to leverage ICE plots as part of the `model_understanding` toolkit.

An ICE plot visualizes the dependence of the prediction on a feature for each instance separately, resulting in one line per instance, compared to one line overall in partial dependence plots. A PDP is the average of the lines of an ICE plot.
https://christophm.github.io/interpretable-ml-book/ice.html

http://scikit-learn.org/stable/auto_examples/inspection/plot_partial_dependence.html

christopherbunn:
This implementation extends `graph_partial_dependencies()` by adding support for plotting Individual Conditional Expectations (ICE) behind a Partial Dependency plot. In addition, `partial_dependence()` has been updated to be able to generate the ICE data (as `graph_partial_dependencies()` uses this function to get the plot data). 

Single Class:
image

Multiclass:
image

Multiclass ICE only:
image
 

By default, ICE plots are not shown when partial dependency plots are created. Setting `kind='individual'` or `kind='both'` will show the ICE plot.

Caveats: 
- Does not work for creating two-way partial dependency plots as there is no clean way to display ICE plots over a Contour plot. `partial_dependence()` is able to calculate the data behind the two-way partial dependency plot and return it as a list of DataFrames (with each DataFrame representing one sample). For now, a `ValueError` is thrown when an ICE plot is attempted for a two-way partial dependency. 

Once the implementation clears, I can update the documentation example to show an ICE plot.

Resolves #2025

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -762,7 +762,7 @@ def partial_dependence(
                         ind_data.append(ind_df)
                     else:
                         ind_data[i].concat(ind_df)
-            
+
             for sample in ind_data:
                 sample["class_label"] = np.repeat(classes, len(values[0]))
 
@@ -939,7 +939,7 @@ def graph_partial_dependence(
         )
 
     fig = _go.Figure(layout=layout)
-    if (isinstance(pipeline, evalml.pipelines.MulticlassClassificationPipeline)):
+    if isinstance(pipeline, evalml.pipelines.MulticlassClassificationPipeline):
         class_labels = [class_label] if class_label is not None else pipeline.classes_
         _subplots = import_or_raise(
             "plotly.subplots", error_msg="Cannot find dependency plotly.graph_objects"
@@ -959,10 +959,14 @@ def graph_partial_dependence(
         # Don't specify share_xaxis and share_yaxis so that we get tickmarks in each subplot
         fig = _subplots.make_subplots(rows=rows, cols=cols, subplot_titles=class_labels)
         for i, label in enumerate(class_labels):
-            label_df = part_dep.loc[part_dep.class_label == label] if part_dep is not None else ice_data.loc[ice_data.class_label == label]
+            label_df = (
+                part_dep.loc[part_dep.class_label == label]
+                if part_dep is not None
+                else ice_data.loc[ice_data.class_label == label]
+            )
             row = (i + 2) // 2
             col = (i % 2) + 1
-            if ice_data is not None and kind == 'individual':
+            if ice_data is not None and kind == "individual":
                 fig = _add_ice_plot(_go, fig, ice_data, row=row, col=col, label=label)
             else:
                 label_df.drop(columns=["class_label"], inplace=True)
@@ -985,12 +989,14 @@ def graph_partial_dependence(
                         trace = _go.Bar(x=x, y=y, name=label)
                     else:
                         if ice_data is not None:
-                            fig = _add_ice_plot(_go, fig, ice_data, row=row, col=col, label=label)
+                            fig = _add_ice_plot(
+                                _go, fig, ice_data, row=row, col=col, label=label
+                            )
                         trace = _go.Scatter(
                             x=x,
                             y=y,
                             line=dict(width=3, color="rgb(99,110,250)"),
-                            name='Partial Dependence: ' + class_labels_mapping[label],
+                            name="Partial Dependence: " + class_labels_mapping[label],
                         )
                     fig.add_trace(trace, row=row, col=col)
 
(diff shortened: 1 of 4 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -552,7 +552,7 @@ def _raise_value_error_if_mostly_one_value(df, percentile):
 
 
 def partial_dependence(
-    pipeline, X, features, percentiles=(0.05, 0.95), grid_resolution=100
+    pipeline, X, features, percentiles=(0.05, 0.95), grid_resolution=100, kind="average"
```
christopherbunn, line 555: We could potentially make `kind="both"` by default, but it would throw an error when doing a two-way PD. If we changed the behavior to be a silent change or a warning, I would be okay with doing this as a default

evalml/model_understanding/graphs.py
```diff
@@ -947,9 +1034,29 @@ def graph_partial_dependence(
                     x=part_dep["feature_values"],
                     y=part_dep["partial_dependence"],
                     name="Partial Dependence",
-                    line=dict(width=3),
+                    line=dict(width=3, color="rgb(99,110,250)"),
```
christopherbunn, line 1037: The RGB number corresponds to the default blue that I extracted from the default MacOS color picker. I don't think there's like an alias to this exact color in plotly, but I could be missing it from the docs

evalml/tests/model_understanding_tests/test_graphs.py
```diff
@@ -850,7 +850,7 @@ def test_jupyter_graph_check(
     jupyter_check.return_value = True
     with pytest.warns(None) as graph_valid:
         graph_partial_dependence(clf, X, features=0, grid_resolution=20)
-        assert len(graph_valid) == 1
+        assert len(graph_valid) == 0
```
christopherbunn, line 853: Is there a particular reason that `graph_partial_dependence` was set to 1 here while the other graphing functions were 0?

evalml/model_understanding/graphs.py
```diff
@@ -947,9 +1034,29 @@ def graph_partial_dependence(
...
+    x = ice_data["feature_values"]
+    if "class_label" in ice_data.columns:
+        ice_data.drop(columns=["class_label"], inplace=True)
+    ice_data.drop(columns=["feature_values"], inplace=True)
+    for i, sample in enumerate(ice_data):
+        fig.add_trace(
+            _go.Scatter(
+                x=x,
+                y=ice_data[sample],
+                line=dict(width=0.5, color="gray"),
```
christopherbunn, line 1053: There's some leeway to adjust this but from tweaking it a bit this is the best color and thickness combo for the ICE plot lines.

evalml/model_understanding/graphs.py
```diff
@@ -947,9 +1034,29 @@ def graph_partial_dependence(
...
+                name="Individual Conditional Expectation",
+                legendgroup="ICE",
+                showlegend=True if i == 0 else False,
```
christopherbunn, line 1056: These two make it so that the ICE plot lines are grouped together under one label in the figure.

evalml/model_understanding/graphs.py
```diff
@@ -818,6 +876,10 @@ def graph_partial_dependence(
         ]
         if any(is_categorical):
             features = _put_categorical_feature_first(features, is_categorical[0])
+        if kind == "individual" or kind == "both":
+            raise ValueError(
+                "Individual conditional expectation plot can only be created with a one-way partial dependence plot"
+            )
```
christopherbunn, line 882: Currently, I'm leaving out the ICE plot when plotting a two-way PD. I think the current behavior of raising an error is the best move, although I've considered putting in a warning message in the log. Open to suggestions for this behavior 🙂

christopherbunn (commented):
Some implementation details to consider when reviewing:

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1024,3 +1076,129 @@ def test_partial_dependence_respect_grid_resolution(fraud_100):
...
+            fig_dict["data"][-1]["y"], class_2_data
+        )
+
+    for i in range(len(X)):
+        assert np.array_equal(fig_dict["data"][i]["x"], ind_dep_data["feature_values"])
+        if problem_type == ProblemTypes.MULTICLASS:
+            window_length = len(ind_dep_data[f"Sample {i}"].values)
+            data = np.concatenate([fig_dict["data"][i]["y"][:window_length],
+                                   fig_dict["data"][i + len(X) + 1]["y"][:window_length],
+                                   fig_dict["data"][i + 2 * len(X) + 2]["y"][:window_length]])
```
christopherbunn, line 1136: For this particular test, I ended up doing some weird matrix math to wrangle the plotly data. For each sample, this snippet essentially combines the ICE data for each class to one array and compares it to the data for that particular sample in the `partial_dependence()` output data.  The +1 and +2 accounts for the additional PD trace that is in each plot.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1024,3 +1076,129 @@ def test_partial_dependence_respect_grid_resolution(fraud_100):
...
+        clf, X, features=feature, grid_resolution=20, kind="individual"
+    )
+
+    for i in range(len(X)):
+        assert np.array_equal(fig_dict["data"][i]["x"], ind_dep_data["feature_values"])
+        if problem_type == ProblemTypes.MULTICLASS:
+            window_length = len(ind_dep_data[f"Sample {i}"].values)
+            data = np.concatenate([fig_dict["data"][i]["y"][:window_length],
+                                   fig_dict["data"][i + len(X)]["y"][:window_length],
+                                   fig_dict["data"][i + 2 * len(X)]["y"][:window_length]])
```
christopherbunn, line 1168: Same as the other comment, but the +1 and +2 aren't here since we're only plotting the ICE data.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -517,7 +569,7 @@ def test_graph_partial_dependence_multiclass(
     for data, label in zip(fig_dict["data"], pipeline.classes_):
         assert len(data["x"]) == 5
         assert len(data["y"]) == 5
-        assert data["name"] == label
+        assert data["name"] == "Partial Dependence: " + label
```
christopherbunn, line 572: Ended up changing the PD trace name to mention that it is the partial dependence line to differentiate it from the ICE plots.

christopherbunn (commented):
Some more testing specific notes:

chukarsten (changes requested):
Awesome, great work.  I need to do some more due dilligence as I can't run this at the Pentagon and visualize it, but the code and tests look good.  I would just say let's stick with the lower grid resolutions, if possible, to try and reduce partial dependence run time.  If it can all be tied to one variable, that would be better but I can do that in another PR.  Also, I think shying away from the "if not" logic would be a bit more readable.  Really great work though for your first week back!

chukarsten (approved):
Thanks for the changes, friendo!

PR #2387, opened 2021-06-16
Update demos to download data from S3

frances-h:
Update EvalML demos to pull data from S3 instead of using local data files in the package. Tests that use the demos have been updated to use the `use_local` flag which pulls data from the local data directory rather than S3.


setup.py
```diff
@@ -21,6 +21,5 @@
           'evalml = evalml.__main__:cli'
         ]
     },
-    data_files=[('evalml/demos/data', ['evalml/demos/data/fraud_transactions.csv.gz', 'evalml/demos/data/churn.csv']),
```
freddyaboulton, line 24: If we delete `fraud_transactions.csv.gz` and `churn.csv` from this list, won't that mean that `load_fraud` and `load_churn` will fail for users who pip install and set `use_local=True`?

evalml/tests/demo_tests/test_datasets.py
```diff
@@ -1,10 +1,23 @@
...
 
+@pytest.fixture(autouse=True)
+def set_testing_headers():
+    opener = urllib.request.build_opener()
+    opener.addheaders = [("Testing", "True")]
+    urllib.request.install_opener(opener)
+
+
 def test_fraud():
     X, y = demos.load_fraud()
```
freddyaboulton, line 17: The purpose of keeping this line is to make sure the api requests work?

docs/source/conf.py
```diff
@@ -269,6 +269,7 @@ def setup(app):
     p = Path("/home/docs/.ipython/profile_default/startup")
     p.mkdir(parents=True, exist_ok=True)
     shutil.copy("disable-warnings.py", "/home/docs/.ipython/profile_default/startup/")
+    shutil.copy("set-headers.py", "/home/docs/.ipython/profile_default/startup")
```
freddyaboulton, line 272: Good catch. Would not have thought to do this!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1633,15 +1633,15 @@ def test_targets_pandas_data_types_classification(
...
+        X, y = load_breast_cancer(use_local=True)
         if "bool" in target_type:
             y = y.map({"malignant": False, "benign": True})
     elif automl_type == ProblemTypes.MULTICLASS:
         if "bool" in target_type:
             pytest.skip(
                 "Skipping test where problem type is multiclass but target type is boolean"
             )
-        X, y = load_wine()
+        X, y = load_wine(use_local=True)
```
freddyaboulton, line 1644: I think it may be best to just create `wine_data`, `breast_cancer_data` fixtures that set `use_local=True`. What do you think?

freddyaboulton (approved):
@frances-h Thanks for this! I have questions about how the change to `data_file` and a suggestion for creating fixtures that set `use_local=True` that I'd like to resolve before merge.

setup.py
frances-h: Yeah, `use_local` would fail unless they manually download the files. Since we're defaulting to downloading from S3, it seemed unnecessary to keep them in the package. They are small though, so keeping them in isn't a big deal if that's the way we want to go. 

Also, if we don't want to include the data files but do want to allow `use_local` for users/make it easier to use, I think there's two other options we could do:
1)  Change `use_local` from a boolean to a path. That way users could download a local copy and specify its location instead of downloading from S3 every time.
2) Add a `save_data` flag that saves the CSV after downloading from S3 to the default data directory so that `use_local` would work in the future.

evalml/tests/demo_tests/test_datasets.py
frances-h: Yep, and to catch any issues if for any reason the local and S3 versions ever get out of sync.

evalml/tests/automl_tests/test_automl.py
frances-h: Yeah that makes sense 👍

setup.py
freddyaboulton: I agree with you in that what we should do should depends on whether users are intended to be able to use the local datasets.

1. If we want them to be able to use the local data, then I think the simplest thing to do is to include `fraud_transactions.csv.gz` and `churn.csv` in the package. I don't think we need to add `save_data` or change `use_local` to be a path but let me know what you think.

2. If we want them to always go through s3, then I think what we can do is remove `use_local` from the `load_data` functions. In our tests, we'll create fixtures that load the data locally. Then we can remove the demo data from our pypi package.

I'm partial to 2 (users always go through s3) because it's consistent with what featuretools does and it lets us run the unit tests without an internet connection.

Let me know what you think!

evalml/tests/demo_tests/test_datasets.py
dsherry: So this unit test will fail if there's no network connection? Are there others which fail under that scenario?

dsherry (approved):
@frances-h this is super cool! Thanks for tackling it.

After this PR, can we still run all of our unit tests without a network connection and have them all pass? I'd like us to preserve this property. I see you and @freddyaboulton have already had some discussion about this.

If we are going to have a unit test or two which tests that the downloaded data matches the local data, which seems like a helpful thing to me, let's do what we can to make it clear why these tests will fail when there's no network connection. A couple ideas: add a note, move them to a separate file.

evalml/tests/demo_tests/test_datasets.py
frances-h: Yeah, these ones will fail if there's no network connection but they're the only ones that will. I can pull them out into a separate test file so it's clearer that it's a separate test. It might also be possible to conditionally xfail if the tests can't reach S3 so they still work offline without causing failures.

frances-h: @dsherry I split out checking that the local and downloaded versions match and marked them to be skipped if the update server is offline. The tests won't fail if you run them offline, and they should still catch if something's wrong with the individual S3 links. Let me know what you think?

setup.py
freddyaboulton: @frances-h Thanks for making the changes! This looks great. I thought about it this morning, and I think we'll need to still include fraud and churn in the pypi package. The reason is that we run our unit tests when we build the conda package for a new release and the conda build happens from the tar.gz on PyPi. 

Once that small change is in, feel free to merge! Thank you 😄 

FYI @dsherry @chukarsten

evalml/tests/demo_tests/test_datasets.py
```diff
@@ -1,28 +1,47 @@
...
+@pytest.fixture(autouse=True)
+def set_testing_headers():
+    opener = urllib.request.build_opener()
+    opener.addheaders = [("Testing", "True")]
+    urllib.request.install_opener(opener)
+
+
+@pytest.fixture(autouse=True)
+def skip_if_offline(request, set_testing_headers):
+    if request.node.get_closest_marker("skip_offline"):
```
dsherry, line 18: To make this easier to `grep`, please make the fixture name and the marker name identical

evalml/tests/conftest.py
```diff
@@ -34,9 +33,17 @@
...
 )
 from evalml.pipelines.components.utils import _all_estimators
+from evalml.preprocessing import load_data
 from evalml.problem_types import ProblemTypes, handle_problem_types
 
 
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers",
+        "skip_offline: mark test to be skipped if offline (https://api.featurelabs.com cannot be reached)",
```
dsherry, line 43: Whoa neat 👍  this is cool

evalml/tests/demo_tests/test_datasets.py
```diff
@@ -1,28 +1,47 @@
...
-def test_fraud():
-    X, y = demos.load_fraud()
+@pytest.fixture(autouse=True)
+def set_testing_headers():
+    opener = urllib.request.build_opener()
+    opener.addheaders = [("Testing", "True")]
+    urllib.request.install_opener(opener)
+
+
+@pytest.fixture(autouse=True)
```
dsherry, line 16: Can we run this fixture in session scope? So that it is only run once per pytest run? This will minimize the chance of network flakes

https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

PR #2395, opened 2021-06-17
Miscellaneous CI updates

dsherry:
Follow-on to #2365 which was just merged.

Changes
* Rename tests so they're a little easier to read in the list which appears at the bottom of each PR
* Run lint on python 3.9
* Run nightlies on python 3.8 too (why not?)
* Have the nightlies run at 3am EST instead of 7am, just in case anyone is working early 😆

### DIFF
--- a/.github/workflows/build_conda_pkg.yml
+++ b/.github/workflows/build_conda_pkg.yml
@@ -1,4 +1,4 @@
-name: Build Conda Package
+name: Build conda package
 
 on:
   pull_request:
--- a/.github/workflows/linux_nightlies.yml
+++ b/.github/workflows/linux_nightlies.yml
@@ -1,8 +1,8 @@
-name: Nightly Linux Unit Tests
+name: Nightly unit tests, linux
 
 on:
   schedule:
-      - cron: '0 7 * * *'
+      - cron: '0 3 * * *'
 
 jobs:
   unit_tests:
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ['3.7', '3.9']
+        python_version: ['3.7', '3.8', '3.9']
         command: ['git-test-automl', 'git-test-modelunderstanding', 'git-test-other', 'git-test-dask']
     steps:
       - name: Set up Python ${{ matrix.python_version }}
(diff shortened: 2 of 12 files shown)


.github/workflows/build_conda_pkg.yml
```diff
@@ -1,4 +1,4 @@
-name: Build Conda Package
+name: Build conda package
```
dsherry, line 1: Call me out if there's no point updating this lol, I just figured it'd be easier to read in the list

.github/workflows/linux_unit_tests_with_latest_deps.yml
```diff
@@ -1,4 +1,4 @@
-name: Linux Unit Tests with Latest Dependencies
+name: Unit tests, linux, latest dependencies
```
dsherry, line 1: Listing the job type first and platform second makes the sort better I think, and makes it easier to distinguish between the jobs.

.github/workflows/linux_nightlies.yml
```diff
@@ -1,8 +1,8 @@
-name: Nightly Linux Unit Tests
+name: Nightly unit tests, linux
 
 on:
   schedule:
-      - cron: '0 7 * * *'
+      - cron: '0 3 * * *'
```
freddyaboulton, line 5: FYI the cron is in UTC so this is actually 11 pm EST. I think before it was set to 3 am EST.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

.github/workflows/linux_nightlies.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ['3.7', '3.9']
+        python_version: ['3.7', '3.8', '3.9']
```
freddyaboulton, line 14: Sounds good to me! 👍

freddyaboulton (approved):
@dsherry Looks good to me! I don't have a preference for when the nightlies are run but right now I think it's set to 11 pm because of UTC.

jeremyliweishih (approved):
LGTM - I think these changes make it more clear!

.github/workflows/linux_nightlies.yml
dsherry: Oh 💩  I'm sorry! I'll change it back

PR #2398, opened 2021-06-17
Revert nightly run time

dsherry:
Thanks @freddyaboulton for catching that. Cron time is in UTC

https://github.com/alteryx/evalml/pull/2395#discussion_r653975612


jeremyliweishih (approved):
good catch

freddyaboulton (approved):
Thanks @dsherry !

PR #2406, opened 2021-06-18
Fixture to handle mocking for automl tests

freddyaboulton:
### Pull Request Description
Using a fixture that handles mocking for you. The idea is that we will have one place that sets all the `patch` calls to reduce boilerplate.

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -66,7 +66,6 @@ Release Notes
         * Sped up unit tests and split into separate jobs :pr:`2365`
         * Change CI job names, run lint for python 3.9, run nightlies on python 3.8 at 3am EST :pr:`2395` :pr:`2398`
         * Set fail-fast to false for CI jobs that run for PRs :pr:`2402`
-        * Added ``AutoMLTestEnv`` test fixture :pr:`2406`
 
 .. warning::
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -16,6 +16,7 @@ Release Notes
         * Added slack integration for nightlies tests :pr:`2436`
         * Changed ``build_conda_pkg`` CI job to run only when dependencies are updates :pr:`2446`
         * Updated workflows to store pytest runtimes as test artifacts :pr:`2448`
+        * Added ``AutoMLTestEnv`` test fixture for making it easy to mock automl tests :pr:`2406`
 
 .. warning::
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -65,6 +65,7 @@ Release Notes
         * Sped up unit tests and split into separate jobs :pr:`2365`
         * Change CI job names, run lint for python 3.9, run nightlies on python 3.8 at 3am EST :pr:`2395` :pr:`2398`
         * Set fail-fast to false for CI jobs that run for PRs :pr:`2402`
+        * Added ``AutoMLTestEnv`` test fixture :pr:`2406`
 
 .. warning::
 
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -1049,7 +1049,7 @@ def test_add_to_rankings(
         allowed_component_graphs=dummy_classifier_linear_component_graph,
     )
     env = AutoMLTestEnv("binary")
-    with env.test_env(score_return_value={"Log Loss Binary": 1.0, "F1": 0.5}):
+    with env.test_context(score_return_value={"Log Loss Binary": 1.0, "F1": 0.5}):
         automl.search()
     assert len(automl.rankings) == 1
     assert len(automl.full_rankings) == 1
@@ -2680,7 +2680,7 @@ def test_max_batches_output(max_batches, AutoMLTestEnv, X_y_binary, caplog):
         max_batches=max_batches,
     )
     env = AutoMLTestEnv("binary")
-    with env.test_context(automl, score_return_value={"Log Loss Binary": 0.3}):
+    with env.test_context(score_return_value={"Log Loss Binary": 0.3}):
         automl.search()
 
     output = caplog.text
@@ -2777,19 +2777,22 @@ def test_data_splitter_binary(AutoMLTestEnv, X_y_binary):
     env = AutoMLTestEnv("binary")
     with pytest.raises(Exception, match="Missing target values in the"):
         with pytest.warns(UserWarning):
-            env.run_search(automl, score_return_value={"Log Loss Binary": 1.0})
+            with env.test_context(score_return_value={"Log Loss Binary": 1.0}):
+                automl.search()
 
     y[1] = 1
     automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", n_jobs=1)
     with pytest.raises(Exception, match="Missing target values in the"):
         with pytest.warns(UserWarning):
-            env.run_search(automl, score_return_value={"Log Loss Binary": 1.0})
+            with env.test_context(score_return_value={"Log Loss Binary": 1.0}):
+                automl.search()
 
     y[2] = 1
     automl = AutoMLSearch(
         X_train=X, y_train=y, problem_type="binary", optimize_thresholds=False, n_jobs=1
     )
-    env.run_search(automl, score_return_value={"Log Loss Binary": 1.0})
+    with env.test_context(score_return_value={"Log Loss Binary": 1.0}):
+        automl.search()
 
 
 def test_data_splitter_multi(AutoMLTestEnv, X_y_multi):
@@ -2801,7 +2804,8 @@ def test_data_splitter_multi(AutoMLTestEnv, X_y_multi):
     env = AutoMLTestEnv("multiclass")
     with pytest.raises(Exception, match="Missing target values"):
         with pytest.warns(UserWarning):
-            env.run_search(automl, score_return_value={"Log Loss Multiclass": 1.0})
+            with env.test_context(score_return_value={"Log Loss Multiclass": 1.0}):
+                automl.search()
 
     y[1] = 2
     # match based on regex, since data split doesn't have a random seed for reproducibility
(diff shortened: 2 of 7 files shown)


evalml/tests/conftest.py
```diff
@@ -1016,3 +1023,170 @@ def _imbalanced_data_X_y(problem_type, categorical_columns, size):
...
+        mock_fit = self._patch_method(
+            "fit", side_effect=mock_fit_side_effect, return_value=mock_fit_return_value
+        )
+        mock_score = self._patch_method(
+            "score", side_effect=mock_score_side_effect, return_value=score_return_value
+        )
+
+        # For simplicity, we will always mock predict_proba and _encode_targets even if the problem is not a binary
+        # problem. For time series problems, we will mock these methods in the time series class (self._pipeline_class)
+        # and for non-time-series problems we will use BinaryClassificationPipeline
```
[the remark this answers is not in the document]
freddyaboulton: Good point. We don't **_have_** to because the only time we call these methods outside of `score` is for threshold tuning. Since mutliclass pipelines don't support threshold tuning and we mock `score` for the `_pipeline_class`, everything works. But the implementation is nicer if we patch the multiclass methods for multiclass pipelines.

chukarsten (approved):
I think this is a very ambitious project and there's a lot I like about it.  I like the consolidation of the patching into the TestEnv.  I think the change, overall, is a good one.  I think it does make writing tests that are properly mocked (for performance sake), easier, at the cost of some clarity when it comes to what the test is doing.  I think it's certainly a price that I'm willing to pay if we can integrate this into a unified philosophy of how to write tests for AutoML stuff.  

It would certainly be good to get everyone's eyes on it as devs will need to use it to conform to this paradigm and reviewers will need to know it exists and enforce its usage.  

I didn't see anything blocking but there's an extra automl.search() in there.  Great work and initiative.

evalml/tests/conftest.py
```diff
@@ -1088,3 +1094,170 @@ def _imbalanced_data_X_y(problem_type, categorical_columns, size):
...
+        """Short-hand for creating test_context and running search within that test_context."""
+        with self.test_context(
+            score_return_value=score_return_value,
+            mock_score_side_effect=mock_score_side_effect,
+            mock_fit_side_effect=mock_fit_side_effect,
+            mock_fit_return_value=mock_fit_return_value,
+            predict_proba_return_value=predict_proba_return_value,
+            optimize_threshold_return_value=optimize_threshold_return_value,
+        ):
+            automl._SLEEP_TIME = 0.0000001
```
jeremyliweishih, line 1257: is this to decrease the sleep time we already have?

jeremyliweishih (approved):
This is amazing @freddyaboulton, should make mocking tests much easier. Thanks for this!

evalml/tests/conftest.py
```diff
@@ -1088,3 +1094,170 @@ def _imbalanced_data_X_y(problem_type, categorical_columns, size):
...
+            "evalml.objectives.BinaryClassificationObjective.optimize_threshold",
+            return_value=optimize_threshold_return_value,
+        )
+
+        mock_tell = patch("evalml.tuners.skopt_tuner.Optimizer.tell")
+
+        # Unfortunately, in order to set the MagicMock instances as class attributes we need to use the
+        # `with ... ` syntax.
+        with mock_fit as fit, mock_score as score, mock_encode_targets as encode, mock_predict_proba as proba, mock_tell as tell, mock_optimize as optimize:
+            # Can think of `yield` as blocking this method until the computation finishes running
```
bchen1116, line 1229: Great comment!

bchen1116 (approved):
Big fan of the comments you left in the `_AutoMLTestEnv` class, as well as how much shorter/cleaner the tests look! Great change!

ParthivNaresh (approved):
Excellent work, it definitely looks cleaner! I don't know how common having a centralized mocking fixture is normally in projects but after seeing this, I'd be surprised if I don't see it more often.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -410,7 +389,8 @@ def test_rankings(
         max_iterations=3,
         n_jobs=1,
     )
-    automl._SLEEP_TIME = 0.001
+    env = AutoMLTestEnv("binary")
+    env.run_search(automl, score_return_value={"Log Loss Binary": 0.03})
     automl.search()
```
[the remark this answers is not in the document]
freddyaboulton: Yes!

evalml/tests/automl_tests/test_engine_base.py
```diff
@@ -127,26 +117,25 @@ def test_train_pipeline_trains_and_tunes_threshold(
...
+    env = AutoMLTestEnv("binary")
+    with env.test_context():
+        _ = train_pipeline(
+            dummy_binary_pipeline_class({}), X, y, automl_config=automl_config
+        )
 
-    mock_pipeline_fit.assert_called_once()
-    mock_optimize.assert_not_called()
+    env.mock_fit.assert_called_once()
+    env.mock_optimize_threshold.assert_not_called()
```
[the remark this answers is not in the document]
freddyaboulton: This is a great observation. The mocks are not set as class attributes until after the `yield` in the context manager, i.e until after the computation we're mocking has finished running. So the mocks have to be accessed outside `test_manager`!

I modified the implementation so that a `ValueError` is raised if mock is accessed within `test_context`. Hopefully that'll help any future confusion.

evalml/tests/conftest.py
freddyaboulton: Yes!

evalml/tests/conftest.py
```diff
@@ -1088,3 +1094,217 @@ def _imbalanced_data_X_y(problem_type, categorical_columns, size):
...
+            yield
+            self._mock_fit = fit
+            self._mock_tell = tell
+            self._mock_score = score
+            self._mock_encode_targets = encode
+            self._mock_predict_proba = proba
+            self._mock_optimize_threshold = optimize
+
+    def run_search(
```
dsherry, line 1285: What is the value of using this method over running
```python
with self.test_context():
    automl.search()
```
?
It seems like the main value is that `automl._SLEEP_TIME` is getting set. That's pretty great; I'm on board with standardizing that.

I think one disadvantage of using this method is that for an outside reader, there's more cognitive load to go look up with `_AutoMLTestEnv.run_search` does and see it runs `AutoMLSearch.search`, rather than just seeing that right away and understanding what the test is doing.

Assuming I'm right about the value here, is there some way we can set that sleep time in `test_context`, delete `run_search` and have all the test code still call `AutoMLSearch.search` directly?

One idea for doing that:
* Change the way `AutoMLSearch._SLEEP_TIME` is defined and used in `AutoMLSearch` so that the usage of it in `AutoMLSearch.search` could be mocked
* Update `test_context` to mock `SLEEP_TIME` to your value, 1e-7
* Optionally, add a `sleep_time` argument to `test_context` alongside the other args, if there are tests which want to set that value to something else.

I guess an alternative would be to make the sleep time a private `AutoMLSearch` init param 🤷‍♂️

Thoughts?

docs/source/release_notes.rst
```diff
@@ -64,6 +65,7 @@ Release Notes
         * Sped up unit tests and split into separate jobs :pr:`2365`
         * Change CI job names, run lint for python 3.9, run nightlies on python 3.8 at 3am EST :pr:`2395` :pr:`2398`
         * Set fail-fast to false for CI jobs that run for PRs :pr:`2402`
+        * Added ``AutoMLTestEnv`` test fixture :pr:`2406`
```
dsherry, line 68: Please remove this entry, wrong section

evalml/tests/conftest.py
```diff
@@ -1088,3 +1094,217 @@ def _imbalanced_data_X_y(problem_type, categorical_columns, size):
         return X, y
 
     return _imbalanced_data_X_y
+
+
+class _AutoMLTestEnv:
```
dsherry, line 1099: Please feel free to ignore this comment entirely. But...

You have me wondering how hard it would be to separate the automl-specific parts of this fixture out from the mocking/context parts. And then put up a PR into pytest with the mocking fixture, and several examples of how to use it in different contexts.

Again, not at all necessary for this PR, just food for thought. It could be real cool to give people a tool which makes it easy to define fixtures like this which perform mocking!!!!

dsherry (approved):
@freddyaboulton wow this PR is total 🔥🔥🔥 ! I am impressed, this is pretty much exactly what I was hoping for when we chatted about this on [your previous PR](https://github.com/alteryx/evalml/pull/2365#discussion_r653691447). Using a fixture to standardize the mocking pattern(s) we use for running general unit tests of our `AutoMLSearch.search` functionality is a huge win for future-proofing our tests and making it easy for others to contribute. And using the context manager like this is nifty, I haven't done that before personally. Really well done!

One blocking comment: please remove duplicate release notes entry

I left some other comments. One was a thought about whether or not the `run_search` method is necessary. Another is a thought about generalizing this pattern.

👏👏😁

evalml/tests/automl_tests/test_engine_base.py
```diff
@@ -127,26 +117,25 @@ def test_train_pipeline_trains_and_tunes_threshold(
...
         None, "binary", LogLossBinary(), [], None, True, None, 0, None, None
     )
-    _ = train_pipeline(
-        dummy_binary_pipeline_class({}), X, y, automl_config=automl_config
-    )
+    env = AutoMLTestEnv("binary")
+    with env.test_context():
+        _ = train_pipeline(
+            dummy_binary_pipeline_class({}), X, y, automl_config=automl_config
+        )
```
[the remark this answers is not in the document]
freddyaboulton: Correct!

docs/source/release_notes.rst
freddyaboulton: Great catch!

evalml/tests/conftest.py
freddyaboulton: @dsherry I think this is a great point! Yes, originally `run_search` was added to set the sleep time and because I thought that `create context then run search` was such a common operation that it deserved its own short-hand but I agree that it's clearer if we create the context for every test!

I'm implementing the first bullet you suggested here.

evalml/tests/conftest.py
freddyaboulton: This is such a cool idea! I think the one thing that we need an abstraction to create multiple nested contexts so that it can be more general. Right now, we're manually combining all the patches into the same context, e.g. `with mock_fit, mock_score, ...`. 

I looked at [ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) when I was first designing this but it wasn't working like I wanted it to. Could be that I was not using it correctly! So I'll get this merged in and think about this further.

PR #2408, opened 2021-06-20
Update `compute_order` in ComponentGraph to be a read-only property

### LINKED ISSUE #2342: Update `compute_order` in ComponentGraph to be a read-only property
Following up with the discussion in https://github.com/alteryx/evalml/pull/2332#discussion_r645722170: the compute order is something we use, but not something that users should necessarily be able to change (as it should be defined by edges and nodes the the graph). As such, we should make it a read-only parameter, since it'd be weird if the ComponentGraph had specific edges but a compute order that conflicts.

By doing so, we no longer have to explicitly check for compute_order in ComponentGraph's `__eq__` as added in #2332.

angela97lin:
Closes #2342


chukarsten (approved):
Looks good.

PR #2410, opened 2021-06-21
Exposes `thread_count` for Catboost estimators as `n_jobs` parameters and n_jobs as a keyword argument for XGBoost

### LINKED ISSUE #1475: Add thread count parameter to Catboost/XGBoost estimators
Catboost allows users to specify a[ `thread_count` parameter](https://catboost.ai/docs/concepts/python-reference_parameters-list.html#python-reference_parameters-list) on init. Out Catboost estimators can accept `thread_count` as a kwarg, but since it is not a named argument, our automl algorithm won't create the component with the [user's desired level of parallelism](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L137). 

It looks like the default value of `thread_count` is -1 for all available threads. I think in most cases this is what's desired but I think it would be useful for users if we properly document this parameter and make sure automl sets it to the right value.

`thread_count` can help if the user specifies a large number of estimators:
image

evalml's Catboost estimators fail automl's `n_jobs` check:

```python
import inspect
params = inspect.signature(EvalmlCatBoost.__init__).parameters
assert 'n_jobs' not in params
```

The same can be said for [XGBoost](https://xgboost.readthedocs.io/en/latest/parameter.html)

angela97lin:
Closes #1475.

Some UX questions and concerns:
- What happens if user specifies `thread_count` and `n_jobs`? Currently, will only use the `n_jobs` parameter. That means if a user sets `thread_count` to 2, but does *not* set `n_jobs` which defaults to -1, we will still use -1. I think this is okay given that our API specifies `n_jobs` as the parameter to use for multicore processing, but this could be a bit confusing (especially given XGBoost/CatBoost use threads, scikit-learn uses processes).

-------

Image after exposing `n_jobs` for CatBoost:
image

XGBoost: Since we use the python scikit-learn interface, `n_jobs` is already implemented. This PR just exposes n_jobs as a named argument :)

--------
It's interesting to note that XGBoost takes more time using all threads than with just two. There is some documentation about how thread contention slows down the algorithm which probably plays a role in this. See image below for how the runtime slows down after setting >16 threads. I filed #2437 to further track this, as it could be useful to investigate this with more datasets and runs.

Relevant links: https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.core; https://github.com/dmlc/xgboost/issues/3810

image


docs/source/release_notes.rst
```diff
@@ -20,6 +20,7 @@ Release Notes
         * Add optional automatic update checker :pr:`2350`
         * Added ``search_order`` to ``AutoMLSearch``'s ``rankings`` and ``full_rankings`` tables :pr:`2345`
         * Updated threshold optimization method for binary classification :pr:`2315`
+        * Exposed ``thread_count`` for Catboost estimators as ``n_jobs`` parameter :pr:`2410`
```
angela97lin, line 23: Will move after release :)

evalml/pipelines/components/estimators/classifiers/catboost_classifier.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         silent=True,
         allow_writing_files=False,
         random_seed=0,
+        n_jobs=-1,
```
freddyaboulton, line 45: Might be time to finally write a docstring for catboost and xgboost? 😅  Especially since there's a risk users can pass in `n_threads` to Catboost and assume it will work.

evalml/pipelines/components/estimators/regressors/catboost_regressor.py
```diff
@@ -57,9 +58,11 @@ def __init__(
         cb_parameters = copy.copy(parameters)
         if bootstrap_type is None:
             cb_parameters.pop("bootstrap_type")
+        cb_parameters["thread_count"] = n_jobs
```
freddyaboulton, line 61: Regarding your UX question, it might be good to let users know via a warning that the `thread_count` parameter is ignored if passed in.

We used to do something similar for the `loss` parameter of the old elastic net parameter.

https://github.com/alteryx/evalml/blob/v0.25.0/evalml/pipelines/components/estimators/classifiers/elasticnet_classifier.py#L32

jeremyliweishih (approved):
I think this looks good! RE your UX question I think its fine the way it is but maybe we can add an explanation in the docstring of the catboost components for clarity.

freddyaboulton (approved):
@angela97lin I think this looks great! I have two points I'd like to resolve before merging:

1. Can you add coverage to the `AutoMLSearch`/`IterativeAlgorithm` that xgboost and catboost are initialized with the right value of `n_jobs`?  I don't think this line was working as intended for those estimators before this pr right? https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L258

2. Since we'll now be running xgboost with `n_jobs=-1` by default and you noted that can slow down xgboost, how much slower will `AutoMLSearch` after this PR? Might be good to quantify that via perf tests so we can tag https://github.com/alteryx/evalml/issues/2437 as high priority.

Let me know what you think!

evalml/pipelines/components/estimators/regressors/xgboost_regressor.py
```diff
@@ -37,23 +37,35 @@ def __init__(
...
+            "n_jobs": n_jobs,
         }
         parameters.update(kwargs)
 
         xgb_error_msg = (
             "XGBoost is not installed. Please install using `pip install xgboost.`"
         )
         xgb = import_or_raise("xgboost", error_msg=xgb_error_msg)
-        xgb_Regressor = xgb.XGBRegressor(random_state=random_seed, **parameters)
+        xgb_regressor = xgb.XGBRegressor(random_state=random_seed, **parameters)
```
bchen1116, line 66: nice catch!

bchen1116 (approved):
LGTM! Nice tests and documentation!

evalml/pipelines/components/estimators/regressors/catboost_regressor.py
angela97lin: I love this suggestion, added!!

chukarsten (approved):
Looks good to me!

PR #2414, opened 2021-06-21
Set CV folds to 10 for datasets under size threshold

### LINKED ISSUE #1783: Small dataset support: update data splitting strategy
I bet we'll see better generalizability on some small datasets (low number of rows) if we increase the default number of CV folds.

This issue tracks choosing and implementing a policy for small datasets.

My gut recommendation is to do 10 CV folds for datasets under a size threshold. Personally I'd set this threshold somewhere around 100 rows. [This writeup](https://scikit-learn.org/stable/modules/cross_validation.html#leave-one-out-loo) was a good read (note: I'm not suggesting we do leave-one-out CV; the text I'm linking to happened to be in that section).

@rpeck

### LINKED ISSUE #2312: Small data: use 10-fold CV instead of 3-fold
TBD: size threshold. Perhaps 500 rows?

christopherbunn:
Resolves #1783
Resolves #2312


angela97lin:
@christopherbunn This resolves #2312? Will unassign myself :)

christopherbunn:
> @christopherbunn This resolves #2312? Will unassign myself :)

Looks like it, I can add the issue number to the PR. Thanks!

dsherry:
Ah thanks both for catching!

@christopherbunn please assign yourself to #2312 and #1783.

@christopherbunn thanks for finding #1783 as well, sounds like we had a duplicate there!

evalml/automl/utils.py
```diff
@@ -77,6 +80,11 @@ def make_data_splitter(
         sklearn.model_selection.BaseCrossValidator: Data splitting method.
     """
     random_seed = random_seed
+    if n_splits is None:
```
freddyaboulton, line 83: @christopherbunn I think it's worth running perf tests before merge. What do you think? 

We may need to artificially shorten our datasets to meet the threshold but I'm curious by how much this slows us down and if it will make our small-dataset models better?

christopherbunn: For sure, I was planning on kicking off a set today. There's none of them that are below 100 rows, but there are a good chunk that are 500 rows or less (which is the threshold suggested in #1783). If we want to increase the threshold to 500, I should be able to assemble a set of small datasets easily but I might have to make a new artificial set for <100 rows.

freddyaboulton: Sweet, sounds good! I don't have a strong opinion on what the threshold should be so happy to keep with `100` and artificially shorten some of the existing datasets to meet the threshold.

chukarsten: I agree with freddy with respect to artificially shortening.  If you're going to do that, I'd suggest tying the max number of rows in the dataset to a variable so we can adjust that lever later on all the tests at once.

evalml/automl/utils.py
```diff
@@ -77,6 +80,11 @@ def make_data_splitter(
...
+        if X.shape[0] < _SMALL_DATA_ROW_THRESHOLD and not is_time_series(problem_type):
+            n_splits = _SMALL_DATA_CV_FOLDS
+        else:
+            n_splits = 3
```
chukarsten, line 87: Perhaps we could just leave the default to be 3 and reduce the conditional to just the if branch?  Functionally, it should be the same, yes?

evalml/tests/automl_tests/test_automl_utils.py
```diff
@@ -93,6 +98,20 @@ def test_make_data_splitter_default(problem_type, large_data):
...
+    ]:
+        assert isinstance(
+            data_splitter,
+            KFold if problem_type == ProblemTypes.REGRESSION else StratifiedKFold,
+        )
+        assert data_splitter.n_splits == _SMALL_DATA_CV_FOLDS
+        assert data_splitter.shuffle
+        assert data_splitter.random_state == 0
+        return
+
```
chukarsten, line 114: Perhaps this branch should merge with the top branch as all of them branch off the same group of problem types?

```
    if problem_type in [
        ProblemTypes.REGRESSION,
        ProblemTypes.BINARY,
        ProblemTypes.MULTICLASS,
    ]:
        if data_size == "large":
            assert isinstance(data_splitter, TrainingValidationSplit)
            assert data_splitter.stratify is None
            assert data_splitter.random_seed == 0
            assert data_splitter.shuffle
            assert data_splitter.test_size == _LARGE_DATA_PERCENT_VALIDATION
        elif data_size == "small" 
            assert isinstance(
                data_splitter,
                KFold if problem_type == ProblemTypes.REGRESSION else StratifiedKFold,
            )
            assert data_splitter.n_splits == _SMALL_DATA_CV_FOLDS
            assert data_splitter.random_state == 0
            assert data_splitter.shuffle
```

chukarsten (approved):
I think the suggestions I left are mostly line-reducing nits.  Nothing blocking!  Nice work!

christopherbunn:
Closing due to a lack of performance improvement on smaller datasets. More info on #2414.

PR #2415, opened 2021-06-21
Latest dep rebase

chukarsten:
### Pull Request Description
(replace this text with your description)

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #2389, opened 2021-06-16
Unpinning PyZMQ

christopherbunn:
Testing to see if the Windows issues are resolved...
Resolves #1759

PR #2427, opened 2021-06-22
Automated Minimum Dependency Updates

machineFL:
This is an auto-generated PR with **minimum** dependency updates. Please do not delete the `min-dep-update` branch because it's needed by the auto-dependency bot.

### DIFF
--- a/evalml/tests/dependency_update_check/minimum_core_requirements.txt
+++ b/evalml/tests/dependency_update_check/minimum_core_requirements.txt
@@ -3,6 +3,7 @@ pandas==1.2.1
 scipy==1.3.3
 scikit-learn==0.24.0
 scikit-optimize==0.8.1
+pyzmq==20.0.0
 colorama==0.4.4
 cloudpickle==1.5.0
 click==7.1.2
--- a/evalml/tests/dependency_update_check/minimum_requirements.txt
+++ b/evalml/tests/dependency_update_check/minimum_requirements.txt
@@ -1,4 +1,4 @@
-plotly==5.0.0
+plotly==4.14.0
 kaleido==0.1.0
 ipywidgets==7.5
 xgboost==1.1.0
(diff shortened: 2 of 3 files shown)


evalml/tests/dependency_update_check/minimum_requirements.txt
```diff
@@ -1,4 +1,4 @@
-plotly==5.0.0
+plotly==4.14.0
```
dsherry, line 1: Wait... how did the deps update bot choose to lower a min version?? I thought it would only modify the latest requirement versions. I could be misunderstanding though. @gsheni do you happen to know the answer? Tagging @freddyaboulton too

evalml/tests/dependency_update_check/minimum_core_requirements.txt
```diff
@@ -3,6 +3,7 @@ pandas==1.2.1
 scipy==1.3.3
 scikit-learn==0.24.0
 scikit-optimize==0.8.1
+pyzmq==20.0.0
```
dsherry, line 6: I guess we must have missed this in #2389. The confusing thing is that was merged 5 days ago, so why is this showing up now?

gsheni: I think its showing up now because the minimum dependency checker was failing 5 days ago, when `pytest-timeout` was added without a minimum version. 
- https://github.com/alteryx/evalml/pull/2374/files

evalml/tests/dependency_update_check/minimum_requirements.txt
gsheni: I am unsure why this happened, but I imagine its similar reason as above. It is the correct version, based on whats in `main` correctly.

freddyaboulton: @chukarsten had to modify our unit tests to be compatible with plotly 5.0.0. So I think what we should do is change `requirements.txt` to be `plotly>=5.0.0` and disregard this change. 

https://github.com/alteryx/evalml/pull/2415/files

freddyaboulton:
Closed by #2440

PR #2430, opened 2021-06-23
Speed up docs build

freddyaboulton:
### Pull Request Description

Following the process described in this [document](https://ayx.atlassian.net/wiki/spaces/PS/pages/586809580/WIP+2404+-+Speed+Up+Docs+Build)

I think this is about a 45% reduction in total run time, so it's about twice as fast now! 

It's hard to look at the internals of sphinx process but I think there are two main buckets:

1. Running our notebooks and converting them to html
2. Creating our api ref from our source code by importing our modules

The sphinx build time is about 8 minutes total. It takes about 5 minutes to run all of the notebooks. So the other 3 minutes are sphinx creating our api reference according to our templates. 

I scaled the notebook computations as much as possible while still being reasonable, so that 5 minutes looks hard to beat (feel free to make suggestions!). I'm not exactly sure how to speed up the api reference beyond setting `collapse_navigation` to `True`. It's not clear how we can profile that to see where the bottleneck is (if it exists). One thing that comes to mind is using the `-j` parameter to build the api ref in parallel, but RTD does not let us control that as far as I know (once again, correct me if I'm wrong!).

So I think this 8 minute build time is about as low as we can go if we still want to have dynamically generated docs, which I think we do! 

-----


docs-requirements.txt
```diff
@@ -3,3 +3,4 @@ pydata-sphinx-theme>=0.3.1
 Sphinx>=2.0.1,<4.0.0
 nbconvert>=5.5.0
 nbsphinx>=0.8.5
+ipython-autotime
```
freddyaboulton, line 6: This is just to time the notebook cells. I will delete before merge.

docs/source/conf.py
```diff
@@ -101,6 +101,8 @@
 html_theme_options = {
     "github_url": "https://github.com/alteryx/evalml",
     "twitter_url": "https://twitter.com/AlteryxOSS",
+    "collapse_navigation": True,
```
freddyaboulton, line 104: This turns out to be pretty important. 

https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-collapse_navigation

docs/source/demos/lead_scoring.ipynb
```diff
@@ -133,7 +133,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Because the lead scoring labels are binary, we will use `AutoMLSearch(X_train=X_train, y_train=y_train, problem_type='binary')`. When we call `.search()`, the search for the best pipeline will begin. "
+    "Because the lead scoring labels are binary, we will use set the problem type to \"binary\". When we call `.search()`, the search for the best pipeline will begin. "
```
freddyaboulton, line 136: It was bugging me that `AutoMLSearch(X_train=X_train, y_train=y_train, problem_type='binary')` is not how we actually create automl in the next cell.

docs/source/demos/lead_scoring.ipynb
```diff
@@ -146,6 +146,7 @@
     "                      problem_type='binary',\n",
     "                      objective=lead_scoring_objective,\n",
     "                      additional_objectives=['auc'],\n",
+    "                      allowed_model_families=[\"catboost\", \"random_forest\", \"linear_model\"],\n",
```
freddyaboulton, line 149: Seems like the only way to not fail the lead_score >= auc score assertion is to train catboost on the entire data. I think this is the cleanest way to do that while also cutting out some pipelines to not spend about a minute on this search.

docs/source/conf.py
```diff
@@ -270,6 +272,7 @@ def setup(app):
     p.mkdir(parents=True, exist_ok=True)
     shutil.copy("disable-warnings.py", "/home/docs/.ipython/profile_default/startup/")
     shutil.copy("set-headers.py", "/home/docs/.ipython/profile_default/startup")
+    shutil.copy("ipython_config.py", "/home/docs/.ipython/profile_default")
```
freddyaboulton, line 275: Will also delete before merge.

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -123,7 +127,7 @@
    "outputs": [],
    "source": [
     "from evalml.model_understanding.graphs import partial_dependence\n",
-    "partial_dependence(pipeline_binary, X, features='mean radius')"
+    "partial_dependence(pipeline_binary, X_holdout, features='mean radius', grid_resolution=5)"
```
chukarsten, line 130: classic grid resolution 👍

chukarsten (approved):
I love this @freddyaboulton , thanks for making such an impactful change so quickly.

docs/source/conf.py
dsherry: 🤯  amazing.

I wonder if the navigation depth also helped

docs/source/demos/cost_benefit_matrix.ipynb
```diff
@@ -122,7 +122,8 @@
    "outputs": [],
    "source": [
     "from evalml import AutoMLSearch\n",
-    "automl = AutoMLSearch(X_train=X_train, y_train=y_train, problem_type='binary', objective='log loss binary')\n",
+    "automl = AutoMLSearch(X_train=X_train, y_train=y_train, problem_type='binary', objective='log loss binary',\n",
+    "                      max_iterations=5)\n",
```
dsherry, line 126: Seems fine. We'll probably have to revisit this when the new automl algo comes in because `max_iterations` may no longer be supported at that point.

docs/source/demos/fraud.ipynb
```diff
@@ -68,7 +68,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "X, y = evalml.demos.load_fraud(n_rows=5000)"
+    "X, y = evalml.demos.load_fraud(n_rows=1000)"
```
dsherry, line 71: Does the fraud demo still read OK with this change? I seem to remember us changing this because the numbers in the demo were screwy when the data size was too small. But I could be wrong.

docs/source/demos/text_input.ipynb
```diff
@@ -38,7 +38,7 @@
     "import pandas as pd\n",
     "\n",
     "input_data = urlopen('https://featurelabs-static.s3.amazonaws.com/spam_text_messages_modified.csv')\n",
-    "data = pd.read_csv(input_data)\n",
+    "data = pd.read_csv(input_data)[:750]\n",
```
dsherry, line 41: Nice. Dang how big was this dataset?

docs/source/ipython_config.py
```diff
@@ -0,0 +1,2 @@
+
+c.InteractiveShellApp.extensions = ['autotime']
```
dsherry, line 2: I'm guessing this will also be deleted before merge?

docs/source/user_guide/automl.ipynb
```diff
@@ -607,10 +607,12 @@
    "outputs": [],
    "source": [
     "X, y = evalml.demos.load_breast_cancer()\n",
+    "X, y = X[:250], y[:250]\n",
```
dsherry, line 610: Is this necessary? You already set `n_rows` at the top

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -40,9 +40,13 @@
     "from evalml.pipelines import BinaryClassificationPipeline\n",
     "X, y = evalml.demos.load_breast_cancer()\n",
     "\n",
+    "X_train, X_holdout, y_train, y_holdout = evalml.preprocessing.split_data(X, y, problem_type='binary',\n",
+    "                                                                         test_size=0.2, random_seed=0)\n",
+    "\n",
+    "\n",
```
dsherry, line 46: Niiiice good call

docs/source/demos/text_input.ipynb
dsherry: Ah, 3k rows. Not too bad. Truncating seems good to me tho

dsherry (approved):
👏  incredible

can_you_believe_this_guy

docs/source/demos/lead_scoring.ipynb
jeremyliweishih: good idea!

jeremyliweishih (approved):
great work @freddyaboulton! I'll double check the docs itself but this looks good to me.

jeremyliweishih:
An observation @freddyaboulton:
- Looks like for the [fraud demo ](https://feature-labs-inc-evalml--2430.com.readthedocs.build/en/2430/demos/fraud.html
)- both pipelines score evenly on fraud. AUC is actually higher on the fraud optimized pipeline as well. Not sure if this is a change in behavior but could make this section a little confusing.

docs/source/demos/fraud.ipynb
jeremyliweishih: @freddyaboulton
Looks like for the [fraud demo ](https://feature-labs-inc-evalml--2430.com.readthedocs.build/en/2430/demos/fraud.html
)- both pipelines score evenly on fraud. AUC is actually higher on the fraud optimized pipeline as well. Not sure if this is a change in behavior but could make this section a little confusing.

docs/source/demos/cost_benefit_matrix.ipynb
freddyaboulton: Agreed! Hopefully there's a "super fast" mode hehe

docs/source/demos/fraud.ipynb
freddyaboulton: I think the fraud demo is currently broken which is why @bchen1116 is reworking the fraud objective? It's possible we'll have to increase the size of the data once the fraud objective "works" again but when we get there, we can do what we're doing now for lead scoring and limit the model families to keep the computation fast.

So I think this is the simplest way to not spend a lot of time in this notebook while we fix it. Let me know what you think!

image

jeremyliweishih: @freddyaboulton works for me! thanks for explaining 😄

freddyaboulton: Ohhh nvm, lower scores are better for fraud! I'll limit the model families now!

docs/source/demos/text_input.ipynb
freddyaboulton: Yea 3k isn't too bad but the problem is that the text featurizer takes up a lot of the run time. Perhaps when we add component caching, we can go back to the full 3k rows!

docs/source/ipython_config.py
freddyaboulton: Yes!

docs/source/user_guide/automl.ipynb
freddyaboulton: This is a different dataset from fraud! Let me time if I can use the full breast cancer data.

docs/source/demos/fraud.ipynb
freddyaboulton: Ohh, that screen shot was from stable, not latest. Looks like it is broken on latest/main hehe

https://evalml.alteryx.com/en/latest/demos/fraud.html

image

docs/source/user_guide/automl.ipynb
freddyaboulton: Yea I think shrinking the model families and doing one less batch are sufficient so I'll use the full dataset!

Issue #2431, opened 2021-06-23
Test compatibility with upcoming pandas release 1.3.0 


gsheni:
- The [upcoming pandas release](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.3.0.html) should be tested with EvalML to make sure upcoming breaking changes are addressed.
- Woodwork plans to make sure that we support pandas 1.3.0. Here is the pull request in progress
  - https://github.com/alteryx/woodwork/pull/987

chukarsten:
EvalML 0.27.0 is incompatible as-is with Pandas 1.3.0rc1.  Will be working on an update to identify and fix some minor bugs introduced.

One potential incompatibility might be in XGBoost's use of DMatrix() applied to the new pandas DataFrame.  When the dataframe is handed to the DMatrix constructor with the columns attribute being a RangeIndex object rather than a list of int64, an exception is raised.

PR #2442, opened 2021-06-24
Pandas 1.3.0 Upgrade

chukarsten:
Addresses issues introduced with Pandas 1.3.0 rc1. #2431 Preliminary work required to prepare for Woodwork 0.5.0.

### DIFF
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -91,10 +91,10 @@ class ClassImbalanceDataCheck(DataCheck):
             error_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
             DataCheck._add_message(
                 DataCheckError(
-                    message=error_msg.format(self.cv_folds, below_threshold_values),
+                    message=error_msg.format(self.cv_folds, sorted(below_threshold_values)),
                     data_check_name=self.name,
                     message_code=DataCheckMessageCode.CLASS_IMBALANCE_BELOW_FOLDS,
-                    details={"target_values": below_threshold_values},
+                    details={"target_values": sorted(below_threshold_values)},
                 ),
                 results,
             )
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -91,7 +91,9 @@ class ClassImbalanceDataCheck(DataCheck):
             error_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
             DataCheck._add_message(
                 DataCheckError(
-                    message=error_msg.format(self.cv_folds, sorted(below_threshold_values)),
+                    message=error_msg.format(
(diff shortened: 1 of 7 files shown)


evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -91,10 +91,12 @@ def validate(self, X, y):
             error_msg = "The number of instances of these targets is less than 2 * the number of cross folds = {} instances: {}"
             DataCheck._add_message(
                 DataCheckError(
-                    message=error_msg.format(self.cv_folds, below_threshold_values),
+                    message=error_msg.format(
```
chukarsten, line 94: Something about the new pandas caused the order in the datachecks details to change.  I sorted the ones that got changed so it shouldn't be an issue going forward.

evalml/pipelines/components/estimators/estimator.py
```diff
@@ -62,6 +64,8 @@ def predict(self, X):
         """
         try:
             X = infer_feature_types(X)
```
chukarsten, line 66: For some reason, passing the equivalent RangeIndex to XGBoost's DMatrix results in the columns receiving an appended space when converted to strings.  e.g. ["0 ", "1 ", "2 "...]

chukarsten (commented):
The two places that the pandas upgrade from 1.2.5 to 1.3.0 manifested themselves were: 

1. The columns of XGBoost's DMatrix created from the new DataFrame resulted in the integer column labels being prepended extra spaces so that they would be strings with exactly two characters.  Not sure why this happens with the new Pandas, but I think this is a pandas/XGBoost interaction, so might file an issue there.
2. The order, but not the values, of some of the datacheck validation results has changed.  The affected tests are rewritten so the order doesn't matter.

jeremyliweishih (approved):
nice!

evalml/data_checks/class_imbalance_data_check.py
freddyaboulton: Must be `value_counts`? It's called in both this data check and InvalidTargetDataCheck

evalml/pipelines/components/estimators/estimator.py
freddyaboulton: Bizzare.

freddyaboulton (approved):
Looks good to me @chukarsten ! Thanks for doing this.

bchen1116 (approved):
LGTM!

evalml/data_checks/class_imbalance_data_check.py
angela97lin: Frustrating, I believe they changed the behavior of `value_counts` not too many releases ago too 🤔

angela97lin (approved):
Not that you needed another green seal of approval, but LGTM!

evalml/pipelines/components/estimators/estimator.py
chukarsten: Ultimately this was an issue with xgboost and DMatrix, resolved in xgboost > 1.4.2.

PR #2448, opened 2021-06-28
Updated workflows to store pytest runtimes as test artifacts 

### LINKED ISSUE #2396: Record runtime of each unit test as a test artifact
It would be awesome if we generated a junit yml or something which recorded the runtime of each test. I am pretty sure pytest has a flag for this.

This would make it easy for us to check if a test is running longer than we expect. It also opens the door for us to write a job which scans for changes in test runtimes and errors out if any tests start running long for some reason. There are certainly other ways to solve the problem of unit test runtime creep--we can discuss that later, independently of this issue.

angela97lin:
Closes #2396.

Example of artifact: https://github.com/alteryx/evalml/actions/runs/980439245

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,35 +31,35 @@ test-dask:
 
 .PHONY: git-test-dask
 git-test-dask:
-	pytest evalml/tests/automl_tests/dask_tests/ -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/ --junitxml=test-reports/junit.xml --doctest-continue-on-failure  --timeout 300 --durations 0
+	pytest evalml/tests/automl_tests/dask_tests/ -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/ --junitxml=test-reports/git-test-dask-junit.xml --doctest-continue-on-failure  --timeout 300 --durations 0
 
 .PHONY: git-test-minimal-deps-dask
 git-test-minimal-deps-dask:
-	pytest evalml/tests/automl_tests/dask_tests/  -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/  --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies --timeout 300 --durations 0
+	pytest evalml/tests/automl_tests/dask_tests/  -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/  --junitxml=test-reports/git-test-minimal-deps-dask-junit.xml --doctest-continue-on-failure --has-minimal-dependencies --timeout 300 --durations 0
 
 .PHONY: git-test-automl-core
 git-test-automl-core:
-	pytest evalml/tests/automl_tests evalml/tests/tuner_tests -n 2 --ignore=evalml/tests/automl_tests/dask_tests --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies
+	pytest evalml/tests/automl_tests evalml/tests/tuner_tests -n 2 --ignore=evalml/tests/automl_tests/dask_tests --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-automl-core-junit.xml --doctest-continue-on-failure --has-minimal-dependencies
 
 .PHONY: git-test-automl
 git-test-automl:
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -3,6 +3,7 @@ python_files = evalml/tests/*
 filterwarnings =
     ignore::DeprecationWarning
     ignore::PendingDeprecationWarning
+junit_duration_report = call
```
angela97lin, line 6: This flag is set so that we don't report setup/teardown time, just the call time: https://docs.pytest.org/en/6.2.x/reference.html#confval-junit_duration_report

Makefile
```diff
@@ -31,35 +31,35 @@ test-dask:
...
 
 .PHONY: git-test-modelunderstanding
 git-test-modelunderstanding:
-	pytest evalml/tests/model_understanding_tests -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure
+	pytest evalml/tests/model_understanding_tests -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-modelunderstanding-junit.xml --doctest-continue-on-failure
 
 .PHONY: git-test-other-core
 git-test-other-core:
-	pytest evalml/tests --ignore evalml/tests/automl_tests/ --ignore evalml/tests/tuner_tests/ --ignore evalml/tests/model_understanding_tests/ -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies
+	pytest evalml/tests --ignore evalml/tests/automl_tests/ --ignore evalml/tests/tuner_tests/ --ignore evalml/tests/model_understanding_tests/ -n 2 --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-other-core-junit.xml --doctest-continue-on-failure --has-minimal-dependencies
```
angela97lin, line 58: Maybe we don't need it for core tests? 🤷‍♀️

chukarsten (approved):
LGTM!  Thanks for fixing my copypasta lol

Makefile
```diff
@@ -31,35 +31,35 @@ test-dask:
...
 
 .PHONY: git-test-minimal-deps-dask
 git-test-minimal-deps-dask:
-	pytest evalml/tests/automl_tests/dask_tests/  -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/  --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies --timeout 300 --durations 0
+	pytest evalml/tests/automl_tests/dask_tests/  -n 1 --doctest-modules --cov=evalml/tests/automl_tests/dask_tests/  --junitxml=test-reports/git-test-minimal-deps-dask-junit.xml --doctest-continue-on-failure --has-minimal-dependencies --timeout 300 --durations 0
 
 .PHONY: git-test-automl-core
 git-test-automl-core:
-	pytest evalml/tests/automl_tests evalml/tests/tuner_tests -n 2 --ignore=evalml/tests/automl_tests/dask_tests --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/junit.xml --doctest-continue-on-failure --has-minimal-dependencies
+	pytest evalml/tests/automl_tests evalml/tests/tuner_tests -n 2 --ignore=evalml/tests/automl_tests/dask_tests --durations 100 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-automl-core-junit.xml --doctest-continue-on-failure --has-minimal-dependencies
```
freddyaboulton, line 42: Thoughts on setting `durations` to 0 so that we record every test? Before, I knew we could do something cool like this I limited durations to 100 because I didn't want to flood the console lol

Makefile
freddyaboulton: Why not though?

freddyaboulton (approved):
Super cool @angela97lin ! Curious what you think the next step is? Maybe you can share some of your scripts you had written a while ago to parse the xmls into more human-readable formats?

Makefile
angela97lin: Hehe I guess theoretically there shouldn't be too much of a difference in runtime for core-tests that isn't captured in our non-core tests, unless catboost / xgboost / other dependencies are doing something fishy. But touche, por que no when it doesn't cost us much? 😛

angela97lin:
@freddyaboulton For sure! I filed #2453 with some of my initial scrappy code,  but definitely worth having the scripts in the repo so it's accessible to the whole team whenever we want to try to debug unit test runtimes. I also think it could also be super cool if we could, during a job, calculate the difference in time it took to run on that branch vs the current artifact on main and raise a warning / flag if the runtime significantly increased, but that might be asking for a little bit more 😝

Makefile
angela97lin: I like it! 😁

PR #2450, opened 2021-06-28
Fix Fraud Cost

### LINKED ISSUE #2318: Fix Fraud Objective
Through [experimenting](https://alteryx.quip.com/9dUYAdLfZwCa/Fraud-and-Lead-scoring-analysis#KPcACABcR1k) with the fraud objective, in relation to [this issue](https://github.com/alteryx/evalml/issues/2145), we found that the fraud objective will choose a 0 threshold, or always opt for a naive estimator, in order to obtain the best results in fraud. 

This issue tracks changing the fraud objective such that it no longer chooses a simple naive estimator as the best estimator.

bchen1116:
fix #2318 

In this PR, I revert the minimization objective back to `Golden`.  I did this because I found that `Brent` and `Bounded` methods didn't perform well with the optimization process. While `Brent` and `Bounded` are preferred and suggested by `scipy` due to the speed increase they have over the `Golden` method, I found that these two weren't able to find the threshold properly when I was testing fraud. As we can see from the previous [perf test](https://alteryx.atlassian.net/wiki/spaces/PS/pages/581337141/Bounded+Thresholding+Perf+Tests), these methods performed similarly here, although `Bounded` did take longer to run compared to `Golden`.  

Fraud Doc: 
image
Lead scoring doc build passes without triggering asserts!

Quick examples of thresholding [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/612466759/Fraud+Objective+Results)

Perf test [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/594870288/Golden+vs+Bounded+minimizing+functions)


evalml/objectives/fraud_cost.py
```diff
@@ -36,23 +36,6 @@ def __init__(
         self.fraud_payout_percentage = fraud_payout_percentage
         self.amount_col = amount_col
 
-    def decision_function(self, ypred_proba, threshold=0.0, X=None):
-        """Determine if a transaction is fraud given predicted probabilities, threshold, and dataframe with transaction amount.
-
-        Arguments:
-            ypred_proba (pd.Series): Predicted probablities
-            threshold (float): Dollar threshold to determine if transaction is fraud
```
bchen1116, line 44: This `threshold` parameter didn't really make sense to me. For our fraud prediction models, we are predicting whether or not a transaction is fraud, so the `predicted_proba` output should be a probability on whether the transaction was fraud or not. Multiplying this probability by the amount spent shouldn't be helpful in determine if there was fraud or not, so I don't think we should threshold based on some spent amount. 

I removed this so that we use the probabilities like our other objectives. Let me know if I'm misinterpreting!

evalml/objectives/fraud_cost.py
```diff
@@ -90,7 +73,9 @@ def objective_function(self, y_true, y_predicted, X, sample_weight=None):
         # calculate money lost from fees
         false_positives = (~y_true & y_predicted) * interchange_cost
 
-        loss = false_negatives.sum() + false_positives.sum()
+        # add a penalty if we output naive predictions
+        all_one_prediction_cost = (2 - len(set(y_predicted))) * fraud_cost.sum()
```
bchen1116, line 77: Add a penalty to naive predictions. Without this penalty, we could incur small fraud costs if we predict every transaction as fraudulent.

evalml/objectives/fraud_cost.py
```diff
@@ -23,7 +23,7 @@ def __init__(
             retry_percentage (float): What percentage of customers that will retry a transaction if it
                 is declined. Between 0 and 1. Defaults to .5
 
-            interchange_fee (float): How much of each successful transaction you can collect.
+            interchange_fee (float): How much of each successful transaction you pay.
```
bchen1116, line 26: interchange fees are fees that a merchant pays when a customer uses their card

evalml/objectives/fraud_cost.py
freddyaboulton: I guess the idea was to label as fraud the transactions whose expected loss due to fraud was larger than some threshold.

Maybe this would be helpful in prioritizing larger transactions?

I'm just thinking out loud. I think this is ok from my point of view.

evalml/objectives/fraud_cost.py
freddyaboulton: And this "small fraud cost" is currently not captured by the objective, right? I guess I'm wondering why this isn't currently captured by how the objective is computed.

freddyaboulton (approved):
@bchen1116 I think this looks good! I have a question on the penalty you're adding.

evalml/objectives/binary_classification_objective.py
```diff
@@ -43,7 +43,10 @@ def cost(threshold):
             cost = self.objective_function(y_true, y_predicted, X=X)
             return -cost if self.greater_is_better else cost
 
-        optimal = minimize_scalar(cost, bounds=(0, 1), method="Bounded")
+        optimal = minimize_scalar(
+            cost, bracket=(0, 1), method="Golden", options={"maxiter": 250}
```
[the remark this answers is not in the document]
bchen1116: No worries, I was referencing the [old perf test](https://alteryx.atlassian.net/wiki/spaces/PS/pages/581337141/Bounded+Thresholding+Perf+Tests) I ran when I switched from Golden to Bounded. I saw the perforamcne was the same, but the time took a little longer with these new minimization methods. I assumed the reverse would hold true here

evalml/objectives/fraud_cost.py
bchen1116: Yeah, that seemed reasonable. This also definitely didn't work well with the `Bounded` method, since we bound search from [0, 1], but it also didn't work well with `Brent` and `Golden` methods since those resulting thresholds generally stayed <2, and oftentimes went negative! 

It makes sense to prioritize larger transactions, but that should hopefully be something the model can find rather than relying on the objective, imo

evalml/objectives/fraud_cost.py
bchen1116: It is currently captured, but that would only be the cost coming from `false_positive` cases, which small, especially when `interchange_cost` is small. I wanted to add a big penalization to prevent us from choosing a naive estimator.

evalml/tests/objective_tests/test_fraud_detection.py
```diff
@@ -116,35 +116,29 @@ def test_binary_more_than_two_unique_values():
         fraud_cost.score(y_true, y_predicted)
 
 
-def test_fraud_objective_score(X_y_binary):
-    X, y = X_y_binary
+def test_fraud_objective_score():
     fraud_cost = FraudCost(amount_col="value")
 
-    y_predicted = pd.Series([0.1, 0.5, 0.5])
+    y_predicted = pd.Series([0.5, 0.1, 0.5])
```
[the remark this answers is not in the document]
bchen1116: @chukarsten Previously, we thresholded in FraudCost by doing `predicted_proba * extra_columns`, and the thresholding was `5`. This meant the values we thresholded on looked like
```python
[0.1, 0.5, 0.5] * [100, 5, 200] = [10, 2.5, 100]
```
Thresholding with 5 on this would result in an array that matches `y_true`. Now that I changed thresholding to be on the `predicted_proba` directly, I needed to change the `y_predicted` values to then match `y_true` post-thresholding with the 0.45 threshold.

evalml/objectives/binary_classification_objective.py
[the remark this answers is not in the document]
jeremyliweishih: As a general rule of thumb - I would say optimization methods and their performance on different objective functions and datasets may vary depending on the characteristics of the dataset and luck as well (there's generally some randomness). Might be a good idea to run performance tests just to make sure but I don't think performance would be greatly impacted.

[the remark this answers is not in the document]
jeremyliweishih: Might be cool to play around with `maxiter` as well! My hunch is that 250 might be a little low but thats just a guess 😄

bchen1116:
@jeremyliweishih @freddyaboulton @chukarsten perf test has been [updated](https://alteryx.atlassian.net/wiki/spaces/PS/pages/594870288/Golden+vs+Bounded+minimizing+functions) using `numiter=250`.

evalml/tests/objective_tests/test_fraud_detection.py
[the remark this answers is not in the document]
chukarsten: Awesome, thanks for the explanation!

chukarsten (approved):
Looks good to me!  Nice work!

Issue #2336, opened 2021-06-02
Partial dependence fails if computed percentiles are too close together


freddyaboulton:
Repro

```python
from evalml.pipelines import RegressionPipeline
from evalml.model_understanding import partial_dependence
import pandas as pd
import pytest

pl = RegressionPipeline(["Random Forest Regressor"])
X = pd.DataFrame({"a": list(range(30)), "b": list(range(-10, 20))})
y = 10*X['a'] + X['b']

pl.fit(X, y)

X_pd = X.copy()
X_pd['a'] = X['a'] * 1.e-10

with pytest.raises(ValueError, match="percentiles are too close to each other"):
    partial_dependence(pl, X_pd, "a", grid_resolution=5)
```

This is similar to https://github.com/alteryx/evalml/issues/1923. The error is being raised by [sklearn](https://github.com/scikit-learn/scikit-learn/blob/15a949460dbf19e5e196b8ef48f9712b72a3b3c3/sklearn/inspection/_partial_dependence.py#L94). 

I think we should still raise an error but perhaps it should be its own exception class and make it clear that partial dependence cannot be computed for these kinds of features. I think the same change should be made for the logic we added for #1923.

chukarsten:
I think we need to do a little investigation into `partial_dependence()` to determine why the scale is an issue.   The outcome of this spike should be a recommendation about how to properly handle/or communicate the issue to the user.

PR #2455, opened 2021-06-28
Partial Dependence Scale Error

chukarsten:
Addresses #2336 

The origin of the original error is within sklearn and is the result of `mquantile()` being called on certain features whose scales are too close together.  "Too close" is defined by a call of numpy's `allclose()` function, which compares the values defining, in this case, the .05 and .95 percentiles.  If `np.allclose(five_percentile, ninety_five_percentile)`, the ValueError for the original issue will be raised.  The tolerance for the `allclose()` function that triggers the exception is, by default, 1E-08, so when features are closer together than this, we experience the error.


docs/source/release_notes.rst
```diff
@@ -5,6 +5,7 @@ Release Notes
         * Added support for showing a Individual Conditional Expectations plot when graphing Partial Dependence :pr:`2386`
         * Updated Objectives API to allow for sample weighting :pr:`2433`
     * Fixes
+        * Added custom exception to address features with scales to small for partial dependence :pr:`2455`
```
angela97lin, line 8: too small

angela97lin (approved):
LGTM! 😁

evalml/model_understanding/graphs.py
```diff
@@ -690,14 +690,24 @@ def partial_dependence(
...
+            X=X,
+            features=features,
+            percentiles=percentiles,
+            grid_resolution=grid_resolution,
+            kind=kind,
+        )
+    except ValueError as e:
+        if "percentiles are too close to each other" in str(e):
+            raise ValueError(
+                "The scale of these features is too small and results in"
```
bchen1116, line 705: Is there a way that users would be able to run PDP on this small data? ie change the percentiles? Or is it that if the data is too small/close together, partial dependence won't work at all?

If it's the former, I think it'd be nice if we can include a comment to give the user suggestions on how to change their params. Otherwise, this looks fine!

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1257,3 +1257,24 @@ def test_graph_partial_dependence_ice_plot_two_way_error(
...
+    pl.fit(X, y)
+
+    X_pd = X.copy()
+    X_pd["a"] = X["a"] * 1.0e-10
+
+    with pytest.raises(ValueError, match="scale of these features is too small"):
+        partial_dependence(pl, X_pd, "a", grid_resolution=5)
+
+    # Ensure that sklearn partial_dependence exceptions are still caught as expected
+    with pytest.raises(
```
bchen1116, line 1277: nice!

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1257,3 +1257,24 @@ def test_graph_partial_dependence_ice_plot_two_way_error(
             grid_resolution=5,
             kind="individual",
         )
+
+
+def test_partial_dependence_scale_error():
+
```
bchen1116, line 1263: is this space here on purpose?

bchen1116 (approved):
LGTM! Left a comment on something that can be added if there is a solution to get PDP to work with small scales. Otherwise, 👌

evalml/model_understanding/graphs.py
```diff
@@ -690,14 +690,24 @@ def partial_dependence(
...
+    try:
+        preds = sk_partial_dependence(
+            wrapped,
+            X=X,
+            features=features,
+            percentiles=percentiles,
+            grid_resolution=grid_resolution,
+            kind=kind,
+        )
+    except ValueError as e:
```
freddyaboulton, line 702: Thoughts on us doing the check rather than relying on sklearn? My reasoning is that with @bchen1116 's #2454 , we're going to be computing the grid ourselves for datetimes and I think we'd want to have this same check in place for dates as well?

Maybe the right thing to do is modify #2454 to compute the grid ourselves for all features rather than having separate branches for datetimes vs other features. That may make it easier to put the "can we compute a valid grid for these features?" checks in one place? Curious what you think!

chukarsten: My first crack was reproducing the check myself, rather than just let sklearn's partial dependence fail.  I decided against that, ultimately, as I didn't want to take the risk that sklearn changes the current criteria behind the check and I wanted the catch to be a little more resilient to that.

I'm not even sure how the same check that's failing within partial dependence would handle date times.  I guess you could define a relative tolerance of 1E-5 and it mean something if you convert a datetime to like linux time or something.  

I'm also not against rejecting this PR if you don't think catching the existing sklearn exception is still the right move.  If we're more in favor of computing, for sure, a functional grid, then we can go that route and I can flush this sucker down the toilet.

evalml/model_understanding/graphs.py
chukarsten: I think you could artificially scale the feature by perhaps multiplying by 1E6 or something like that to defeat the absolute tolerance check.  That might fall within the realm of our current transformation components, but I think the story was just for a more informative error message.  Let me add a little bit to the error message.

evalml/tests/model_understanding_tests/test_partial_dependence.py
chukarsten: yea, because I'm a jerk lol

ParthivNaresh (approved):
LGTM! Do you think that this warrants a potential DataCheck in the future? I get the feeling that `uniqueness_data_check` could have an extension to it that checks if the 5th and 95th percentiles are too close together.

PR #2458, opened 2021-06-29
Use sphinx-autoapi to generate api reference

### LINKED ISSUE #601: Use sphinx-autoapi to build API modules statically
During the flurry of testing that happened to get the improved API objectives documentation build to succeed on ReadtheDocs, I came across the `sphinx-autoapi` extension from [RtD's suggestions on cutting build resources](https://docs.readthedocs.io/en/stable/guides/build-using-too-many-resources.html) (a good doc to come back to if we face this again!). As the documentation states:

**If you are installing a lot of Python dependencies just to document your Python modules API using sphinx.ext.autodoc, you can give a try to sphinx-autoapi Sphinx’s extension instead which should produce the exact same output but running statically. This could drastically reduce the memory and bandwidth required to build your docs.**

If we choose to pursue this, we may also want to check on all of the warnings about circular dependencies that the extension produces, as well as if it actually decreases our build time (https://readthedocs.com/projects/feature-labs-inc-evalml/builds/313458/ --> 600+ seconds isn't really an improvement, but this may be because I hadn't actually replaced our previous mechanism of using `autodoc` for building modules?)

### LINKED ISSUE #1877: Rename Objective Functions to Objectives in docs
To maintain consistency, we should rename Objective Functions to Objectives in the table of contents of the docs.

https://evalml.alteryx.com/en/stable/user_guide/objectives.html

freddyaboulton:
### Pull Request Description
Fixes #601 

Fixes #1877 

See this [doc](https://alteryx.atlassian.net/wiki/spaces/PS/pages/642875540/601-+Use+sphinx-autoapi+to+generate+the+api+reference) for the pros and cons of making this change.

-----


docs/source/_auto_api_templates/python/attribute.rst
```diff
@@ -0,0 +1,60 @@
+{% if obj.display %}
```
freddyaboulton, line 1: This is to get the docstrings for `hyperparameter_ranges` etc to show up

docs/source/_auto_api_templates/python/module.rst
```diff
@@ -0,0 +1,120 @@
+{% if not obj.display %}
+:orphan:
+
+{% endif %}
+{% if obj.name.split('.')[-1][0] == '_' %}
+{{ obj.name }}
+{% elif obj.name.split('.')|length <= 2 %}
+{{ (' '.join(obj.name.split('.')[-1].split('_'))|title) }}
```
freddyaboulton, line 8: This template is to modify the display names of the modules, e.g. `evalml.model_understanding` to `Model Understanding`

evalml/automl/automl_search.py
```diff
@@ -124,9 +124,6 @@ class AutoMLSearch:
 
     _MAX_NAME_LEN = 40
 
-    # Necessary for "Plotting" documentation, since Sphinx does not work well with instance attributes.
-    plot = PipelineSearchPlots
```
freddyaboulton, line 128: `plot` shows up as an attribute in the docs without this

https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/automl/index.html#evalml.automl.AutoMLSearch.plot

angela97lin:
Just a quick comment but the "Show Value" dropdown is interesting yet quite confusing 😂 
For example, for the EN Classifier I see this, which looks a little odd?:

image


(https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/pipelines/components/index.html#evalml.pipelines.components.ElasticNetClassifier)

docs/source/_auto_api_templates/python/attribute.rst
angela97lin: My comment is probably related to this :d

freddyaboulton: Good catch with the original comment. I just pushed up a change to fix this! This is how it looks locally now:

image

Made the same change for the linear regressor!

bchen1116:
Some of the formatting in the docs is weird, like [here](https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/api_index.html#preprocessing), [here](https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/api_index.html#automl-utils), [here](https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/api_index.html#prediction-explanations), and [here](https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/api_index.html#model-understanding) where the description is pushed far to the side in comparison to the rest of the docstring. 

Also I'm not sure if I like the look of the docs, like [here](https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/objectives/index.html#evalml.objectives.FraudCost). It feels a little overwhelming and kinda messy, even though it does include a lot of info, but if other people are fine with it, then it's fine!

freddyaboulton:
@bchen1116 Great points! Regarding the description being pushed to the side, I can make it so that the function arguments are not displayed in the api reference page but only when users click on them. That matches the behavior we say on latest/stable:

image

Regarding `FraudCost` looking messy, do you have any suggestions for improvement? I can remove the `Bases: evalml.objectives.binary_classification_objective.BinaryClassificationObjective` which isn't in `latest` or `stable`. Anything else?

bchen1116:
@freddyaboulton The fixes you suggested sound good! 

For me, the main thing that makes it harder to follow is the fact that the attributes are the same size / bigger than the class:
image

For instance, above where the font size of `name` or `greater_is_better` is the same as `FraudCost`. Very nitpicky, but it doesn't feel as clean or easy to read. What do you think?

docs/source/_auto_api_templates/python/module.rst
```diff
@@ -0,0 +1,120 @@
...
+{% endif %}
+{% endblock %}
+
+{% block functions scoped %}
+{% if visible_functions %}
+Functions
+~~~~~~~~~
+
+.. autoapisummary::
+
```
chukarsten, line 92: I legit don't know how to review this lol

evalml/model_understanding/force_plots.py
```diff
@@ -57,7 +57,7 @@ def gen_force_plot(shap_values, training_data, expected_value, matplotlib):
 
 
 def force_plot(pipeline, rows_to_explain, training_data, y):
-    """Function to generate the data required to build a force plot.
+    r"""Function to generate the data required to build a force plot.
```
chukarsten, line 60: Is there a reason this 'r' is here?  I have no idea how I caught this.

evalml/pipelines/components/estimators/classifiers/decision_tree_classifier.py
```diff
@@ -15,13 +15,25 @@ class DecisionTreeClassifier(Estimator):
...
         ProblemTypes.MULTICLASS,
         ProblemTypes.TIME_SERIES_BINARY,
         ProblemTypes.TIME_SERIES_MULTICLASS,
     ]
+    """[
+        ProblemTypes.BINARY,
+        ProblemTypes.MULTICLASS,
+        ProblemTypes.TIME_SERIES_BINARY,
+        ProblemTypes.TIME_SERIES_MULTICLASS,
+    ]"""
```
chukarsten, line 36: This is really weird and I'm not sure I like it.  It adds a lot of bloat, it seems.  Probably still worth the performance benefit, but will take some getting used to.

chukarsten (approved):
I'm cool with this if we get the stated performance benefit.  I think that makes it worth it.  I do think it looks kind of weird and the documentation of the default values of the class variables is strange.  It will take some getting used to.

angela97lin:
@freddyaboulton Great work figuring out how to get this package to work, and putting up the doc of pros and cons! It makes this much more clear to review 😁 

I think what I say might need to be taken with a grain of salt, because I personally don't like how this change looks 😅  As we talked about in OH, this feels very developer-focus, with focus on packages and modules as the way to organize things. Still, as you pointed out, there's really cool benefits to this. In particular, I really like the decrease in runtime (though at this point, what runs slower? Docs or windows tests? Are we still limited by the runtime of window tests?), that docs are generated automatically (since I know I'm not the only one guilty of not adding the most comprehensive docstrings all the time 😛), and the new component class attributes showing up as "code" (though I like less that it requires clicking the dropdown and "multiline value" is not as informative lol). I also like that the autogenerated docs specify what are properties, abstract methods, etc.


Here are some more specific details:
- Methods now display parameters, which means the description width may vary or not show up at all:
image
(another example: https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/model_understanding/index.html)
- It's nice that everything is autogenerated... but it looks like for each class, it just takes the order that its defined in our file. Since we don't do a great job of listing out properties, then methods, we end up with a messy collection. I listed AutoML as an example (https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/automl/index.html):

image
- I agree that the dropdown looks nicer in your PR but arguably, what we have on the right side of our current docs is comparable :d 
image
- I honestly find the dump of everything in a class overwhelming LOL as opposed to our current behavior, which simply lists each of the methods. I looked at scikit-learn's docs (ex: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html) and realized they also list everything, but for some reason it's less overwhelming... so my guess is, it's the formatting. Before, we take advantage of lines as visible borders to separate methods. Now, we use whitespace. Given the bold and larger font we use for method/property/attribute names, it feels like a lot is thrown on the page at once. Is there a way to change the formatting more? (I know I'm asking for a lot now lolol). Maybe if there's a way to list all of the methods and then allow clicking to bring you to the method details like scikit-learn does, that could also be helpful as a way to summarize methods?
- I'm not a fan of how because everything is loaded, clicking on something just brings you to the index. For example, clicking on `MulticlassClassificationObjective` brings you to the top of the `MulticlassClassificationObjective` class, but since it's tiny, most of my screen actually shows the `ObjectiveBase` class. That's a bit confusing 😂 (https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/objectives/index.html#evalml.objectives.MulticlassClassificationObjective)
image


Other specifics:
- https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/objectives/index.html#evalml.objectives.MulticlassClassificationObjective: `problem_type` is specified but not the value?


Oddly enough, I actually like that our current docs mean that we have to think really specifically about how our code is structured, and what we want to display to users. I realized this while looking at https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/tuners/index.html#evalml.tuners.Tuner: it lists the Exceptions we added to tuner_exception.py, I'm guessing, right smack in the middle of the page of all of the tuner classes because that's how our code is organized (tuner.py --> tuner_exception.py --> random...). This makes sense from a code perspective, sure, but would be a scavenger hunt for a user to find 😂 

Again, take what I'm saying with a grain of salt--my dislike for the look might have me nitpicking at a lot of random things 😂

docs/source/_auto_api_templates/python/module.rst
angela97lin: relatable

freddyaboulton:
@angela97lin @bchen1116 @chukarsten Thanks for the detailed feedback! I think I've addressed most of your specific points, detailed below. Before reviewing the docs **clear your browser cache** because I updated the css file and if your browser cached it, you won't see the changes, namely method/function name code highlighting. 

> The drop-down menu for class attributes adds bloat/is non-itutitive
- We're no longer using the drop down menu for class attributes.

> Methods now display parameters, which means the description width may vary or not show up at all
- Methods and functions no longer display parameters, which is what our current docs do! 

> Method/function order in classes or modules is messy
- Methods and functions are now displayed in alphabetical order, which is what our current docs do!

> Displaying all methods in a class looks messy. The font size of the class attributes is so big that things also look messy. 
- We're now listing all methods in a table with a link to the full method documentation. This is what sklearn does. We also display all the attributes in a table so that the font size is smaller. I also added code highlighting to the method and class names to break things up a little bit more. This is also what sklearn does.

image

> I'm not a fan of how because everything is loaded, clicking on something just brings you to the index. For example, clicking on MulticlassClassificationObjective brings you to the top of the MulticlassClassificationObjective class, but since it's tiny, most of my screen actually shows the ObjectiveBase class.

-  That was because before we were not documenting inherited methods/attributes. Now we are. Clicking on the multiclassification objective link will show all the methods/attributes for the objective. It doesn't look as confusing as before! See the image I linked above. 

Still tweaking some stuff but let me know if this is in the right direction!

angela97lin:
@freddyaboulton Only took a quick peek but I think it looks much cleaner already 🤩

bchen1116:
@freddyaboulton the docs look great! Thanks for making the changes, I think this flow is much cleaner and easier to follow! Will look more thoroughly tomorrow

docs/source/conf.py
```diff
@@ -60,7 +59,13 @@
 ]
 
 # Add any paths that contain templates here, relative to this directory.
-templates_path = ["_templates"]
+#templates_path = ["_templates"]
```
bchen1116, line 62: should we delete this comment?

docs/source/conf.py
```diff
@@ -213,8 +221,8 @@
 }
 
 # autoclass_content = 'both'
-autosummary_generate = ["api_reference.rst"]
-templates_path = ["_templates"]
+#autosummary_generate = ["api_reference.rst"]
```
bchen1116, line 224: and these?

bchen1116 (approved):
Wasn't quite sure how to review the rst files, but the docs look great! Thanks for cleaning them up so much and putting in the work to make them much cleaner and easy to follow!

Left a couple comments about things we could delete, but otherwise 🚢

docs/source/conf.py
freddyaboulton: For sure! Thanks for pointing this out!

ParthivNaresh (approved):
Great work on this, and after all the improvements based on the others' suggestions, I think this is solid.
I had a couple of questions about the formatting. 
For `hyperparameter_ranges` for the transformers, I think the format is getting a little messed up:
https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/pipelines/components/index.html#evalml.pipelines.components.SimpleImputer
https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/pipelines/components/index.html#evalml.pipelines.components.TargetImputer
https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/pipelines/components/index.html#evalml.pipelines.components.PolynomialDetrender

Additionally, Exceptions have a method called `with_traceback` that seems to be missing the second half of its docstring `with_traceback   Exception.with_traceback(tb) -`
https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/exceptions/index.html#evalml.exceptions.MissingComponentError.with_traceback
https://feature-labs-inc-evalml--2458.com.readthedocs.build/en/2458/autoapi/evalml/tuners/tuner_exceptions/index.html
I think it's because they're inherited from the `BaseException` class in `builtins` and the new line is confusing Sphinx:
```
"""
     Exception.with_traceback(tb) --
          set self.__traceback__ to tb and return self.
"""
```

freddyaboulton:
@ParthivNaresh Thanks for pointing out those issues! I fixed the formatting for the hyperparameter ranges in the components you listed. I also modified the template to not display `with_traceback` and `args` for exceptions. Those were being pulled in from the base class and don't add a lot of info imo.

This is what it looks like now! You'll need to clear your browser cache so that you pull in the new css:

image

Issue #2459, opened 2021-06-29
Add support for Unknown Logical Type


gsheni:
- The **Unknown** Logical Type has been merged into `main` of Woodwork. 
- It is the default inferred Logical Type, and uses string as the representation.
  - Before NaturalLanguage was the default inferred type. 
- EvalML should be able to handle this Logical Type, and correctly ignore it during AutoML search.

freddyaboulton:
@gsheni I think this is a dup of #2426 !

gsheni:
Dpu

PR #2460, opened 2021-06-29
Pinned WW top at 0.4.2 and bottom at 0.4.1

chukarsten:
In preparation for woodwork 0.5.0 release and evaluate the new type in #2459 .

### DIFF
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -11,7 +11,7 @@ psutil>=5.6.6
 requirements-parser>=0.2.0
 shap>=0.36.0
 texttable>=1.6.2
-woodwork>=0.4.1,<=0.4.2
+woodwork>=0.4.1,<0.5.0
 dask>=2.12.0
 featuretools>=0.21.0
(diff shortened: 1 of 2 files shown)


core-requirements.txt
```diff
@@ -11,7 +11,7 @@ psutil>=5.6.6
 requirements-parser>=0.2.0
 shap>=0.36.0
 texttable>=1.6.2
-woodwork>=0.4.1
+woodwork>=0.4.1,<=0.4.2
```
dsherry, line 14: Should we do `woodwork>=0.4.1,<0.5.0`? If they follow semantic versioning, 0.5.0 will be the next breaking change

@gsheni

gsheni: Agreed, `<0.5.0` is the right approach, and will avoid the upcoming breaking change.

freddyaboulton (approved):
@chukarsten Looks good to me once we set the upper version at < 5.0.0! Thank you!

Issue #2456, opened 2021-06-29
Ordinal Logical Type cause error to EvalML


iki77:
I got an error when using Ordinal Logical Type with EvalML, the data I used is from [default credit card](https://archive.ics.uci.edu/ml/datasets/default+of+credit+card+clients)
I have defined the order values when I initiate it with woodowoek, but suddenly the order values are gone when used with EvalML.

```python
order = sorted(features["EDUCATION"].unique())
logical_types = {
    "AGE": "Age",
    "EDUCATION": Ordinal(order=order)
}
features.ww.init(logical_types=logical_types)

features_train, features_holdout, target_train, target_holdout = evalml.preprocessing.split_data(
    features, 
    target, 
    problem_type="binary",
    test_size=0.2,
    random_seed=42
)

model_families= ["linear_model", "random_forest"]

automl = evalml.AutoMLSearch(
    X_train=features_train, 
    y_train=target_train,
    problem_type="binary",
    objective="f1",
    max_batches=1,
    random_seed=42
)
automl.search()
```

The error log:
```
*****************************
* Beginning pipeline search *
*****************************

Optimizing for F1. 
Greater score is better.

Using SequentialEngine to train and score pipelines.
Searching up to 1 batches for a total of 4 pipelines. 
Will stop searching for new pipelines after 180 seconds.

Allowed model families: linear_model, random_forest

Evaluating Baseline Pipeline: Mode Baseline Binary Classification Pipeline
			Fold 0: Encountered an error.
			Fold 0: All scores will be replaced with nan.
			Fold 0: Please check d:\credit_default\notebooks\evalml_debug.log for the current hyperparameters and stack trace.
			Fold 0: Exception during automl search: Must use an Ordinal instance with order values defined
			Fold 1: Encountered an error.
			Fold 1: All scores will be replaced with nan.
			Fold 1: Please check d:\credit_default\notebooks\evalml_debug.log for the current hyperparameters and stack trace.
			Fold 1: Exception during automl search: Must use an Ordinal instance with order values defined
			Fold 2: Encountered an error.
			Fold 2: All scores will be replaced with nan.
			Fold 2: Please check d:\credit_default\notebooks\evalml_debug.log for the current hyperparameters and stack trace.
			Fold 2: Exception during automl search: Must use an Ordinal instance with order values defined
Mode Baseline Binary Classification Pipeline:
	Starting cross validation
	Finished cross validation - mean F1: nan

```

chukarsten:
[...]

PR #2461, opened 2021-06-30
Quick fix to retain typing

chukarsten:
Addresses: #2456 

This modifies the for-loop to handle Ordinal and Datetime ltypes to pass through the properties that were missing.  The other ltypes are still handled the same way.  When Woodwork addresses this, we can revert this change.


dsherry:
@chukarsten did you mean to say "Fixes #2456 " in the description above?

evalml/tests/utils_tests/test_woodwork_utils.py
```diff
@@ -203,3 +208,49 @@ def test_infer_feature_types_raises_invalid_schema_error():
...
+        }
+    )
+    logical_types = {
+        "non-ordinal": "Age",
+        "ordinal": Ordinal(order=[0, 1, 2]),
+        "categorical": Categorical(encoding="Encoding"),
+        "datetime": Datetime(datetime_format="%Y-%m-%d"),
+    }
+    features.ww.init(logical_types=logical_types)
+
```
chukarsten, line 236: The testing strategy here will be to make sure that the values set in the WW init, `order` for Ordinal and `datetime` for Datetime, remain after they pass through our utility function.  With the exception of categorical, which accepts a value for its init but does nothing with it.

evalml/utils/woodwork_utils.py
```diff
@@ -110,13 +111,20 @@ def _retain_custom_types_and_initalize_woodwork(
         return ww.init_series(new_dataframe, old_logical_types)
     if ltypes_to_ignore is None:
```
chukarsten, line 112: @freddyaboulton do we have a plan for phasing out `_retain_custom_types_and_initialize_woodwork`?  I know your PR dropped some usages of it.  If we need to file an issue with WW to move some functionality over, I'll do it.

evalml/tests/utils_tests/test_woodwork_utils.py
angela97lin: This makes sense to me, but I'm a little confused about the comments in the code. For example, "    # Ordinal type doesn't retain the 'order' property" --> aren't we testing that the ordinal type *does* retain the order property? Or am I totally confused lol

angela97lin (approved):
LGTM, thanks for doing this! Do we have an issue tracking removing this code once WW makes their changes? 😮

evalml/utils/woodwork_utils.py
freddyaboulton: @chukarsten I think we need it for the time being for the Imputer because the imputer will "break" the schema and we need a way of recreating it while allowing some of the types to change. I've been told that https://github.com/alteryx/woodwork/issues/965 will help with that but I think we need to think about that further.

evalml/tests/utils_tests/test_woodwork_utils.py
```diff
@@ -203,3 +208,50 @@ def test_infer_feature_types_raises_invalid_schema_error():
...
+    # Ordinal type should now pass through the function without issue and retain the 'order' property
+    ordinal_subset = _retain_custom_types_and_initalize_woodwork(
+        old_logical_types=logical_types, new_dataframe=features[["ordinal"]]
+    )
+    ltypes = ordinal_subset.ww.logical_types
+    assert ltypes["ordinal"].order is not None
```
freddyaboulton, line 242: Might as well verify the order and date format match the user input?

freddyaboulton (approved):
Looks good to me @chukarsten !

evalml/tests/utils_tests/test_woodwork_utils.py
bchen1116, line 242: Can we not assert that this order is `[0, 1, 2]`? Or do the new logical types not use the old logical types?

bchen1116 (approved):
Looks good! Left a question, but nothing blocking. Would be nice to file an issue to track removing this once WW fixes!

PR #2463, opened 2021-06-30
Add ability to use built-in `pickle` for saving AutoMLSearch

### LINKED ISSUE #2174: Ability to pickle AutoMLSearch
Our `AutoMLSearch.save` and `load` methods support serializing/deserializing `AutoMLSearch` instances using `cloudpickle`. This issue supports doing the same using python's `pickle`.

The pickling method could be an argument to `save` and `load`. Note that whatever pattern we use here, we should use the same pattern elsewhere in our code, like for `PipelineBase.save` and `load`.

christopherbunn:
Resolves #2174

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1330,7 +1330,7 @@ class AutoMLSearch:
             pkl_lib = pickle
         else:
             raise ValueError(
-                f"`pickle_type` must be either 'pickle' or 'cloudpickle'. Received: {pickle_type}"
+                f"`pickle_type` must be either 'pickle' or 'cloudpickle'. Received {pickle_type}"
             )
 
         with open(file_path, "wb") as f:
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,4 +1,5 @@
 import copy
+import pickle
 import sys
 import time
 import traceback
@@ -55,12 +56,7 @@ from evalml.problem_types import (
 )
 from evalml.tuners import SKOptTuner
 from evalml.utils import convert_to_seconds, infer_feature_types
-from evalml.utils.logger import (
-    get_logger,
-    log_subtitle,
-    log_title,
-    time_elapsed,
-)
+from evalml.utils.logger import get_logger, log_subtitle, log_title, time_elapsed
 
 logger = get_logger(__file__)
 
@@ -1307,18 +1303,26 @@ class AutoMLSearch:
 
         return self._best_pipeline
 
-    def save(self, file_path, pickle_protocol=cloudpickle.DEFAULT_PROTOCOL):
+    def save(self, file_path, pickle_type="cloudpickle", pickle_protocol=cloudpickle.DEFAULT_PROTOCOL):
         """Saves AutoML object at file path
 
         Arguments:
             file_path (str): location to save file
+            pickle_type {"pickle", "cloudpickle"}: the pickling library to use.
             pickle_protocol (int): the pickle data stream format.
 
         Returns:
             None
         """
+        if pickle_type == "cloudpickle":
+            pkl_lib = cloudpickle
+        elif pickle_type == "pickle":
+            pkl_lib = pickle
+        else:
+            raise ValueError(f"`pickle_type` must be either 'pickle' or 'cloudpickle'. Received: {pickle_type}")
+
         with open(file_path, "wb") as f:
-            cloudpickle.dump(self, f, protocol=pickle_protocol)
+            pkl_lib.dump(self, f, protocol=pickle_protocol)
 
     @staticmethod
     def load(file_path):
@@ -1331,7 +1335,7 @@ class AutoMLSearch:
             AutoSearchBase object
         """
         with open(file_path, "rb") as f:
-            return cloudpickle.load(f)
+            return pickle.load(f)
 
     def train_pipelines(self, pipelines):
         """Train a list of pipelines on the training data.
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -56,7 +56,12 @@ from evalml.problem_types import (
 )
 from evalml.tuners import SKOptTuner
 from evalml.utils import convert_to_seconds, infer_feature_types
-from evalml.utils.logger import get_logger, log_subtitle, log_title, time_elapsed
+from evalml.utils.logger import (
+    get_logger,
+    log_subtitle,
+    log_title,
+    time_elapsed,
+)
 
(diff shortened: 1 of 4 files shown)


evalml/automl/automl_search.py
```diff
@@ -1331,7 +1347,7 @@ def load(file_path):
             AutoSearchBase object
         """
         with open(file_path, "rb") as f:
-            return cloudpickle.load(f)
+            return pickle.load(f)
```
freddyaboulton, line 1350: So we can save with cloudpickle and read with pickle?! Wow, I had no idea that works hehe.

I feel like we should accept an argument here for the "pickle type"? Feels weird to offer a choice of library for `save` but not respect that in `load`. Not blocking though.

freddyaboulton (approved):
@christopherbunn Looks good! The original issue also mentions doing something similar for pipelines. Is the plan is to do that in this PR or a follow up?

evalml/automl/automl_search.py
```diff
@@ -1307,18 +1308,33 @@ def best_pipeline(self):
...
+        elif pickle_type == "pickle":
+            pkl_lib = pickle
+        else:
+            raise ValueError(
+                f"`pickle_type` must be either 'pickle' or 'cloudpickle'. Received {pickle_type}"
+            )
+
         with open(file_path, "wb") as f:
-            cloudpickle.dump(self, f, protocol=pickle_protocol)
+            pkl_lib.dump(self, f, protocol=pickle_protocol)
```
angela97lin, line 1337: What if the user uses a cloudpickle protocol while trying to use pickle? 🤔

angela97lin (commented):
I'm trying to pickle using a jupyter notebook and I'm getting the following, am I doing something wrong?:

```
automl_ = AutoMLSearch(X, y, problem_type="regression")
automl_.search()
automl_.save("test.pkl", pickle_type="pickle")
```


I get:
```
<ipython-input-12-e1254c6fb959> in <module>
----> 1 automl_.save("test.pkl", pickle_type="pickle")

~/Desktop/evalml/evalml/automl/automl_search.py in save(self, file_path, pickle_type, pickle_protocol)
   1335 
   1336         with open(file_path, "wb") as f:
-> 1337             pkl_lib.dump(self, f, protocol=pickle_protocol)
   1338 
   1339     @staticmethod

TypeError: can't pickle module objects
```

It works if I use cloudpickle 🤔

evalml/automl/automl_search.py
christopherbunn: Apparently cloudpickles can be opened by the regular pickling library (according to their example on their [README.md](https://github.com/cloudpipe/cloudpickle) and the doc string for `cloudpickle.py` in the attached screenshot)! I didn't know this before so that's pretty neat.

image

freddyaboulton:
@angela97lin Are you running in a jupyter notebook? 

The following works for me when running from a script:

```python

X, y = load_diabetes()

automl = AutoMLSearch(X, y, "regression")
automl.search()
automl.save("automl.pkl", pickle_type="pickle")
automl2 = automl.load("automl.pkl")
```

christopherbunn:
> I'm trying to pickle using a jupyter notebook and I'm getting the following, am I doing something wrong?:
> 
> ```
> automl_ = AutoMLSearch(X, y, problem_type="regression")
> automl_.search()
> automl_.save("test.pkl", pickle_type="pickle")
> ```
> 
> I get:
> 
> ```
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-12-e1254c6fb959> in <module>
> ----> 1 automl_.save("test.pkl", pickle_type="pickle")
> 
> ~/Desktop/evalml/evalml/automl/automl_search.py in save(self, file_path, pickle_type, pickle_protocol)
>    1335 
>    1336         with open(file_path, "wb") as f:
> -> 1337             pkl_lib.dump(self, f, protocol=pickle_protocol)
>    1338 
>    1339     @staticmethod
> 
> TypeError: can't pickle module objects
> ```
> 
> It works if I use cloudpickle 🤔

Hmm, that's so odd! I am able to generate a pickle with the test and through a script version of my test Jupyter notebook. At the same time, I am also running into the same error as you when I try to run it interactively in a Jupyter notebook. Let me look into this...

evalml/automl/automl_search.py
angela97lin: So cool 🤩

angela97lin:
@freddyaboulton @christopherbunn Yup, running via jupyter notebook. What I was curious about was if it was possible to save in a file and load in completely different file so I wanted to have two separate notebooks to test but ran into this, heh.

christopherbunn:
I think I figured out the issue: in the notebook environment, we have the search iteration plot appear. However, in the test and script environment, we aren't showing a search iteration plot. As a result, the jupyter notebook version has the plot stored in `automl.search_iteration_plot` whereas the script version has None in this slot. I am able to pickle the search in Jupyter when i set `automl.search_iteration_plot=None`. 

I'm guessing that for some reason the plot doesn't support pickling? This is a bit odd considering that plotly has support for pickling per [this PR on their repo](https://github.com/plotly/plotly.py/pull/1191). I'll see if I can find a workaround or a deeper root cause in Plotly.

angela97lin:
@christopherbunn Wow, good find! If you can't find a good workaround, I think it could also be fine to say that the plots will not be pickled and set `automl.search_iteration_plot=None` before saving, not sure if others have different opinions on that 🤷‍♀️

freddyaboulton:
I think the issue is that the `PipelineSearchPlot` has a reference to `plotly.graph_objects`, which is a module. Maybe we can refactor that? But I think not saving the plot is also a valid approach.

image

We might want to update the AutoMLSearch User Guide to talk about also using pickle to save automl and the limitations. Would also give us "coverage" for pickling in a jupyter notebook env which our unit tests don't cover.

christopherbunn:
Turns out I was wrong, the Jupyter notebook version uses a `SearchIterationPlot` object whereas the script/test version stores a plotly figure (which is pickleable). The error is raised because the `_go` class variable in `SearchIterationPlot` stores the module representing the `plotly.graph_objects` class.  

I think the best path is to actually clear out the `_go` class variable after the init since this is the only place where this is used. Initial testing show that pickling works once this is done. I'll test a few edge cases but if it looks good I'll push it up.

evalml/automl/automl_search.py
christopherbunn: Hmm yeah I agree, it does feel symmetrical to do so. I think it would be a no-op though since it looks like the doc for cloudpickle just recommends using the standard python pickler for loading.

PR #2465, opened 2021-07-01
Use latest ww features

### LINKED ISSUE #2457: Use some of woodwork's latest features
In #2379, we upgraded to the latest woodwork version and made our version requirements less restrictive.

This opens the door to use some of the latest features they've added. From their release notes:

```
Add option to return TableSchema instead of DataFrame from table accessor select method (#916)
Add dropping and renaming columns inplace (#920)
Add concat_columns util function to concatenate multiple Woodwork objects into one, retaining typing information (#932)
```

This should clean up some debt in our repo.

Closing criteria:
- [ ] Use `concat_columns` in our ComponentGraph and components, e.g. DateTimeFeaturizer, OneHotEncoder
- [ ] See if we can delete some of the special code added to #2297 to make sure we preserve types through concat
- [ ] Use the option to return a schema instead of a datframe from `select` to avoid creating new dataframes
- [ ] Drop and rename columns in place in components
- [ ] Since we're touching a lot of components, perf tests to verify no unintended changes are introduced

freddyaboulton:
### Pull Request Description
Fixes #2457

Acceptance Criteria
- [x] Use `concat_columns` in our ComponentGraph and components, e.g. DateTimeFeaturizer, OneHotEncoder
- [x] See if we can delete some of the special code added to #2297 to make sure we preserve types through concat
- [x] Use the option to return a schema instead of a datframe from `select` to avoid creating new dataframes
- [x] Drop and rename columns in place in components
- [x] Since we're touching a lot of components, perf tests to verify no unintended changes are introduced

Performance tests are [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/625180765/2457+Use+new+features+in+ww+0.4.1)

-----

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -1,5 +1,6 @@
 import networkx as nx
 import pandas as pd
+import woodwork as ww
 from networkx.algorithms.dag import topological_sort
 from networkx.exception import NetworkXUnfeasible
 
@@ -8,12 +9,7 @@ from evalml.pipelines.components.transformers.transformer import (
     TargetTransformer,
 )
 from evalml.pipelines.components.utils import handle_component_class
-from evalml.utils import (
-    _retain_custom_types_and_initalize_woodwork,
-    get_logger,
-    import_or_raise,
-    infer_feature_types,
-)
+from evalml.utils import get_logger, import_or_raise, infer_feature_types
 
 logger = get_logger(__file__)
 
@@ -258,14 +254,14 @@ class ComponentGraph:
                 parent_output = infer_feature_types(parent_output)
             if parent_output is not None:
                 final_component_inputs.append(parent_output)
-        concatted = pd.concat(
-            [component_input for component_input in final_component_inputs], axis=1
+        concatted = ww.utils.concat_columns(
+            [component_input for component_input in final_component_inputs]
         )
         if needs_fitting:
             self.input_feature_names.update(
                 {self.compute_order[-1]: list(concatted.columns)}
             )
-        return infer_feature_types(concatted)
+        return concatted
 
     def predict(self, X):
         """Make predictions using selected features.
@@ -436,26 +432,11 @@ class ComponentGraph:
         if len(x_inputs) == 0:
             return_x = X
         else:
-            return_x = pd.concat(x_inputs, axis=1)
+            return_x = ww.concat_columns(x_inputs)
         return_y = y
         if y_input is not None:
             return_y = y_input
 
-        # Expected behavior: Preserve types selected by user unless one of the components updates the
-        # type for that column. This requirement is important because the StandardScaler is expected to convert
-        # Ints to Doubles and converting back to an int would lose information.
-        # In order to meet this requirement, we start off with the user selected types (X.ww.logical_types) and then
-        # update them with the types created by components (if they are different).
-        # Components are not expected to create features with the same names
-        # so the only possible clash is between the types selected by the user and the types selected by a component.
-        # Because of that, the for-loop below is sufficient.
-
-        logical_types = {}
-        logical_types.update(X.ww.logical_types)
-        for x_input in x_inputs:
-            if x_input.ww.schema is not None:
-                logical_types.update(x_input.ww.logical_types)
-        return_x = _retain_custom_types_and_initalize_woodwork(logical_types, return_x)
         if return_y is not None:
             return_y = infer_feature_types(return_y)
         return return_x, return_y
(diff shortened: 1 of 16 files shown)


evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -512,15 +512,15 @@ def test_fit_correct_inputs(
     X = pd.DataFrame(X)
     y = pd.Series(y)
     graph = {"Imputer": [Imputer], "OHE": [OneHotEncoder, "Imputer.x", "Imputer.y"]}
-    expected_x = pd.DataFrame(index=X.index, columns=X.columns).fillna(1)
```
freddyaboulton, line 515: The original data has float-type features but the mock returns int-type features. This meant we were technically changing the dtypes between components. The component graph previously changed the int-types back to floats with this code block:

```python
  logical_types = {}
  logical_types.update(X.ww.logical_types)
  for x_input in x_inputs:
      if x_input.ww.schema is not None:
          logical_types.update(x_input.ww.logical_types)
  return_x = _retain_custom_types_and_initalize_woodwork(logical_types, return_x)
```
which is why we need the `astype("float64")` call below.

Since we now use `ww.concat_columns`, we would be preserving the int-types returned by the mocks because that's `ww.concat_columns` is supposed to do! 

So in order to get this test to pass, we either need to make the mocks return floats or get rid of the `astype` call below. I opted for the former because it makes it clearer that the component graph does not overwrite the types returned by the components.

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -91,7 +89,9 @@ def __init__(
     @staticmethod
     def _get_cat_cols(X):
         """Get names of categorical columns in the input DataFrame."""
-        return list(X.ww.select(include=["category"]).columns)
+        return _put_into_original_order(
```
freddyaboulton, line 92: I'm putting in the same order as they're found in the dataframe to keep the behavior the same.

evalml/pipelines/component_graph.py
```diff
@@ -436,26 +432,11 @@ def _consolidate_inputs(x_inputs, y_input, X, y):
         if len(x_inputs) == 0:
             return_x = X
         else:
-            return_x = pd.concat(x_inputs, axis=1)
+            return_x = ww.concat_columns(x_inputs)
         return_y = y
         if y_input is not None:
             return_y = y_input
 
-        # Expected behavior: Preserve types selected by user unless one of the components updates the
```
freddyaboulton, line 444: `ww.concat_columns` handles this logic for us now

ParthivNaresh (approved):
Looks great, nothing scary in the perf tests so looks good to go!

bchen1116 (approved):
LGTM! I like the changes!

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
freddyaboulton: I was surprised by this! Yea because `.columns` on the schema returns a dictionary keyed by column name. So when you convert that to a list the column order could be different from time to time.

evalml/utils/woodwork_utils.py
```diff
@@ -131,3 +131,8 @@ def _convert_numeric_dataset_pandas(X, y):
         )
     y_ww = infer_feature_types(y)
     return X_ww, y_ww
+
+
+def _put_into_original_order(X, columns_to_subset):
```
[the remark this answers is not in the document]
chukarsten: I submitted [this issue](https://github.com/alteryx/woodwork/issues/1039) to back this out later.

chukarsten (approved):
Submitted [WW issue](https://github.com/alteryx/woodwork/issues/1039) and [EvalML issue](https://github.com/alteryx/evalml/issues/2473) to revert the _put_in_original_order() when they're ready.  Looks good, hope we can bury woodwork_utils.

evalml/utils/woodwork_utils.py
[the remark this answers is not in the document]
freddyaboulton: Thanks @chukarsten !

evalml/pipelines/utils.py
```diff
@@ -80,17 +80,17 @@ def _get_preprocessing_components(
...
+    index_columns = list(X.ww.select("index", return_schema=True).columns)
     if len(index_columns) > 0:
         pp_components.append(DropColumns)
 
-    datetime_cols = X.ww.select(["Datetime"])
+    datetime_cols = list(X.ww.select(["Datetime"], return_schema=True).columns)
 
-    add_datetime_featurizer = len(datetime_cols.columns) > 0
+    add_datetime_featurizer = len(datetime_cols) > 0
     if add_datetime_featurizer and estimator_class.model_family != ModelFamily.ARIMA:
```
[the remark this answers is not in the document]
freddyaboulton: Yes, a stylistic choice 😂

PR #2471, opened 2021-07-06
Fix dask flaky tests

### LINKED ISSUE #2341: Flaky Dask test

bchen1116:
Fix #2341 

Previously, running this test 10 times resulted in some >1 test failures:
image

This seemed to be an issue related to the parallelization of the search. By removing the `threads_per_worker` parameter, we default the threads to 16, which is more than enough to handle AutoMLSearch.

50 runs of this test:
image

100 runs of this test:
image

Interestingly, 8 threads fails, but 12 threads passes.

### DIFF
--- a/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
+++ b/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
@@ -23,7 +23,7 @@ def sequential_engine():
 @pytest.fixture(scope="module")
 def cluster():
     dask_cluster = LocalCluster(
-        n_workers=1, threads_per_worker=4, dashboard_address=None
+        n_workers=1, dashboard_address=None
     )
     yield dask_cluster
     dask_cluster.close()
--- a/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
+++ b/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
@@ -23,7 +23,7 @@ def sequential_engine():
 @pytest.fixture(scope="module")
 def cluster():
     dask_cluster = LocalCluster(
-        n_workers=1, threads_per_worker=2, dashboard_address=None
+        n_workers=1, threads_per_worker=4, dashboard_address=None
     )
     yield dask_cluster
     dask_cluster.close()
--- a/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
+++ b/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
@@ -148,8 +148,8 @@ def test_automl_immediate_quit(X_y_binary_cls, cluster, caplog):
 
         pipelines = [
             TestPipelineFast({}),
-            TestPipelineWithFitError({}),
             TestPipelineSlow({}),
+            TestPipelineWithFitError({}),
         ]
         automl = AutoMLSearch(
             X_train=X,
--- a/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
+++ b/evalml/tests/automl_tests/dask_tests/test_automl_dask.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pytest
 from dask.distributed import Client, LocalCluster
+from dask.system import CPU_COUNT
 
 from evalml.automl import AutoMLSearch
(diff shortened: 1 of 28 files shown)


jeremyliweishih (approved):
looks interesting! any idea why 8 fails but 12 doesnt?

evalml/tests/automl_tests/dask_tests/test_automl_dask.py
```diff
@@ -22,9 +22,7 @@ def sequential_engine():
 
 @pytest.fixture(scope="module")
 def cluster():
-    dask_cluster = LocalCluster(
-        n_workers=1, threads_per_worker=2, dashboard_address=None
-    )
+    dask_cluster = LocalCluster(n_workers=1, dashboard_address=None)
```
chukarsten, line 25: lol but why

freddyaboulton, line 25: Great find! I guess the thinking is that the test relies on `TestPipelineFast` finishing before `TestPipelineWithFitError` but since we set `threads_per_worker` to 2 and there are three pipelines, we can't guarantee that all three pipelines start evaluating at the same time.

To be extra safe, can you verify what the value of `CPU_COUNT` (referenced in the dask [source code](https://github.com/dask/distributed/blob/main/distributed/deploy/local.py#L195)) in our gh workers is? 

One way to do that may be to add a test that fails like this:

```python
from dask.system import CPU_COUNT
assert CPU_COUNT == None
```

The benefit of doing this is verifying that threads_per_worker gets set to 16 as opposed to something else. Like you said, even going with 8 can cause failures.

bchen1116:
@jeremyliweishih @chukarsten Unfortunately, I wasn't sure why 8 fails. I initially thought that having 4 threads would be enough (since there were 3 pipelnes), but it seems like dask's distributed might not work like that or there's additional stuff happening. I got some performance results here:
image

But I'm not sure how to further evaluate what is happening. Would appreciate any advice or tips if yall have any!

evalml/tests/automl_tests/dask_tests/test_automl_dask.py
freddyaboulton: FWIW, I think I'm good with merging this in and seeing how it goes. I just want to know what `CPU_COUNT` is on gh so we know what to expect with regards to future flakes. If this fails, we can try to rewrite the test to not rely a specific pipeline evaluation order?

bchen1116: It's 16!

freddyaboulton: In that case, let's get this baby merged and see what happens 😂

freddyaboulton (approved):
Thanks for the good work digging into this @bchen1116 !

evalml/tests/automl_tests/dask_tests/test_automl_dask.py
bchen1116: It's 2 lol
image

freddyaboulton: 😨  😂  🙈

freddyaboulton: In that case, how about we change the test? The test is supposed to verify that all computations stop after we hit an exception with the `raise_error_callback`.

I think this line that we already have in the test (`assert len(automl.full_rankings) < len(pipelines)`) is sufficient. Whether or not `TestPipelineFast` actually finishes is not really relevant.

What do you think? FYI @chukarsten

bchen1116: I think that sounds good to me. I agree with not requiring the fast pipeline be in the results

PR #2477, opened 2021-07-08
Allow AutoMLSearch to handle Unknown type

### LINKED ISSUE #2426: Handle woodwork's Unknown Logical Type
Woodwork is planning on releasing an `Unknown` logical type soon (https://github.com/alteryx/woodwork/pull/992).

Some our of unit tests and demos may break:

> The PR that adds the Unknown logical type will also come with the behavior that no columns will be inferred as NaturalLanguage since we don’t have an inference function for NaturalLanguage currently.
> So, if you have any tests/demos/code that relies on the NaturalLanguage logical type being set by default by Woodwork, an update will be needed to explicitly set the appropriate columns as NaturalLanguage after the Unknown update is released.

I also think `AutoMLSearch` will break if users pass in data with `Unknown` logical types given that our components don't know how to handle them. 

Closing criteria:
- [ ] Green CI on a branch that installs the latest woodwork release with Unknown logical type.
- [ ] Design and implementation of how evalml should handle `Unknown` logical type. I think we should ignore these features but down to discuss alternatives and/or how to best achieve that!

bchen1116:
fix #2426 

Perf tests [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/669253968/Unknown+type+in+AutoMLSearch)


evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -262,7 +262,7 @@ def test_per_column_imputer_woodwork_custom_overrides_returned_by_components(
     override_types = [Integer, Double, Categorical, NaturalLanguage, Boolean]
     for logical_type in override_types:
         # Column with Nans to boolean used to fail. Now it doesn't
-        if has_nan and logical_type == Boolean:
+        if has_nan and logical_type in [Boolean, NaturalLanguage]:
```
bchen1116, line 265: Leave out NaturalLanguage since casting this will result in `np.nan` becoming `pd.NA`, which fails the imputer

evalml/tests/data_checks_tests/test_natural_language_nan_data_check.py
```diff
@@ -165,14 +169,3 @@ def test_nl_nan_check_input_formats():
...
-    #  test 2D list
-    nl_col_without_nan = [
-        "string_that_is_long_enough_for_natural_language",
-        "string_that_is_long_enough_for_natural_language",
-        "string_that_is_long_enough_for_natural_language",
-    ]
-    assert nl_nan_check.validate([nl_col, nl_col_without_nan]) == expected
-
-    # test np.array
-    assert nl_nan_check.validate(np.array([nl_col, nl_col_without_nan])) == expected
```
bchen1116, line 178: Since we can't cast this to NL type without ww, we remove this from the tests. They'll be cast as `Unknown`

evalml/tests/component_tests/test_per_column_imputer.py
freddyaboulton: But how come this wouldn't happen before? Woodwork doesn't convert `np.nan` to `pd.NA` in 0.4.2? Is this because of the pandas upgrade?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4787,3 +4789,43 @@ def test_automl_thresholding_train_pipelines(mock_objective, threshold, X_y_bina
     else:
         mock_objective.assert_not_called()
         assert all([p.threshold is None for p in pipes.values()])
+
+
+@pytest.mark.parametrize("columns", [[], ["unknown_col"], ["unknown1, unknown2"]])
+def test_automl_drop_unknown_columns(columns, AutoMLTestEnv, X_y_binary, caplog):
```
freddyaboulton, line 4795: Solid 👍

evalml/tests/component_tests/test_text_featurizer.py
```diff
@@ -381,6 +396,7 @@ def test_featurizer_with_custom_indices(text_df):
 
 def test_text_featurizer_does_not_modify_input_data(text_df):
     X = text_df
+    X.ww.init(logical_types={"col_1": "NaturalLanguage", "col_2": "NaturalLanguage"})
```
freddyaboulton, line 399: I feel like we should just make this change in the `text_df` fixture?

freddyaboulton (approved):
@bchen1116 Thank you for doing this! The changes look good to me and I'm glad the performance looks better with this change for a lot of the datasets we commonly test one.

One thing I'd like to find out/cover with (small/fast) unit tests before merge though - Do our model understanding methods support `Unknown` features?

I think we should focus on 
- [ ] partial dependence
- [ ] permutation importance
- [ ] prediction explanations

My hope is yes but I'm not 100% sure.

evalml/tests/component_tests/test_per_column_imputer.py
bchen1116: @freddyaboulton I believe it's due to the upgrade!
image
Looking at the 1.3.0 release docs, seems like there's a [lot of changes](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html) with NaN handling, and they're using [`<NA>` for scalar types](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#experimental-na-scalar-to-denote-missing-values)

freddyaboulton: Thanks for looking into this @bchen1116 Let's list this as a breaking change for now. I imagine we might want to file an issue to discuss if there are any changes we need to make to the simpleimputer? If users run it on natural language after this pr they'll get a stacktrace they didn't get before.

bchen1116: @freddyaboulton updated the release notes with the breaking change and filed the issue [here](https://github.com/alteryx/evalml/issues/2515)!

evalml/automl/automl_search.py
```diff
@@ -494,9 +494,20 @@ def __init__(
             index_columns = list(
                 self.X_train.ww.select("index", return_schema=True).columns
             )
-            index_columns = _put_into_original_order(self.X_train, index_columns)
-            if len(index_columns) > 0 and drop_columns is None:
-                parameters["Drop Columns Transformer"] = {"columns": index_columns}
+            unknown_columns = list(
+                self.X_train.ww.select("Unknown", return_schema=True).columns
+            )
```
chukarsten, line 499: Is it possible to select the unknown and index columns in one call?  I ask because after this next WW release, select will return to us the columns in their original order so that we can probably refactor out `_put_into_original_order`.  Merging the select of index and unknown gets us ready for that.

evalml/model_understanding/graphs.py
```diff
@@ -652,6 +652,12 @@ def partial_dependence(
         X_features = (
             X.ww.iloc[:, [features]] if isinstance(features, int) else X.ww[[features]]
         )
+    X_unknown = X_features.ww.select("unknown")
+    if len(X_unknown.columns):
+        # We drop the unknown columns in the pipelines, so we cannot calculate partial dependence for these
+        raise ValueError(
+            "Columns included for partial dependence cannot be of type 'Unknown'"
```
chukarsten, line 659: I think one helpful tidbit here would probably be to tell the user in the exception which columns are Unknown

evalml/pipelines/utils.py
```diff
@@ -84,8 +84,10 @@ def _get_preprocessing_components(
     if len(text_columns) > 0:
         pp_components.append(TextFeaturizer)
 
-    index_columns = list(X.ww.select("index", return_schema=True).columns)
-    if len(index_columns) > 0:
+    index_and_unknown_columns = list(
+        X.ww.select(["index", "Unknown"], return_schema=True).columns
```
chukarsten, line 88: super-nit: consistency in the capitalization of 'index' and 'unknown'...all one way or another, if you please.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1296,6 +1296,25 @@ def test_partial_dependence_scale_error():
...
+    pl = BinaryClassificationPipeline(["Random Forest Classifier"])
+    pl.fit(X, y)
+    if error:
+        with pytest.raises(
+            ValueError,
+            match="Columns included for partial dependence cannot be of type 'Unknown'",
+        ):
+            partial_dependence(pl, X, indices, grid_resolution=10)
+        return
+    s = partial_dependence(pl, X, indices, grid_resolution=10)
```
chukarsten, line 1314: How low can we keep this grid_resolution?

chukarsten (approved):
A nit here and there, but overall, I like where this is headed.  Some of the things I'd like for you to consider, particularly with respect to the explanation predictions, partial dependence and permutation importance is whether we're adding back in more run time?  I need to reacquaint myself with Freddy's testing environment/context manager to see whether that could help us there, but I'd just like to be cognizant of whether we're walking back any of his work by adding in these tests.  

Overall, though, this is solid.  Thanks a lot for addressing it so quickly from design to impl and getting it out there.

evalml/tests/model_understanding_tests/test_partial_dependence.py
bchen1116: Lowest is 2

bchen1116:
@chukarsten two of the tests take about 1 second, while the permutation importance test add about 3 seconds. I think it's important to have the coverage on these tests, and it doesn't take too long to raise concern for now imo. Let me know what you think!

PR #2487, opened 2021-07-09
Log transformer data check

### LINKED ISSUE #914: AutoML: apply nonlinear transformation to regression target during search
https://scikit-learn.org/stable/auto_examples/compose/plot_transformed_target.html#sphx-glr-auto-examples-compose-plot-transformed-target-py

Could be useful to further affirm that this is useful and improves the accuracy of our models first!

ParthivNaresh:
Fixes #914 
Also addresses: https://alteryx.atlassian.net/browse/EVALML-44 and https://alteryx.atlassian.net/browse/EVALML-53

Documentation and notes are [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/569212979/Detecting+Lognormal+distributions+in+target+data)

Perf tests are [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/900071444/LogTransformer+Performance+Tests)


core-requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.20.0
 pandas>=1.2.1
-scipy>=1.3.3
+scipy>=1.5.0
```
ParthivNaresh, line 3: Added because scipy changed how they return values for the Shapiro test (as a named tuple)

evalml/pipelines/utils.py
```diff
@@ -65,6 +66,14 @@ def _get_preprocessing_components(
     """
 
     pp_components = []
+
+    if problem_type in [ProblemTypes.REGRESSION, ProblemTypes.TIME_SERIES_REGRESSION]:
+        for each_action in TargetDistributionDataCheck(problem_type).validate(X, y)[
```
bchen1116, line 71: Gotcha, this makes sense. Personally, I'm not sure how much I like doing this check, since it seems we need to iterate through the datacheck to figure out if we need to add this component, but I can't think of another way to do it. Looks fine for now though!

evalml/tests/component_tests/test_components.py
```diff
@@ -813,7 +816,7 @@ def test_transformer_transform_output_type(X_y_binary):
     ]
 
     for component_class in _all_transformers():
-        if component_class == PolynomialDetrender:
+        if component_class in [PolynomialDetrender, LogTransformer]:
             # Skipping because this test is handled in test_polynomial_detrender
```
bchen1116, line 820: can we update this comment to say these tests are handled in their respective test files?

evalml/tests/component_tests/test_log_transformer.py
```diff
@@ -0,0 +1,71 @@
...
+        y_ = -np.abs(y_)
+
+    if y_.min() <= 0:
+        y_ = y_ + abs(y_.min()) + 1
+    expected_log = np.log(y_)
+
+    X_ = pd.DataFrame(X_)
+    y_ = pd.Series(y_)
+
+    X, y = X_, y_
```
bchen1116, line 39: I think you should copy X and y earlier in this test. Right now, if the `data_type` is mixed or negative, you change `y_` to become all positive (line 33), and when you set `y = y_`, it sets `y` to the newly-changed value of `y_`.

I think you should copy `X, y` after line 30, where you've altered the `y_` data to be in sync with the `data_type` but haven't yet made the values all positive.

evalml/tests/component_tests/test_log_transformer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if data_type == "positive":
+        y = np.abs(y)
+    elif data_type == "negative":
+        y = -np.abs(y)
+
+    log_ = LogTransformer()
+    output_X, output_y = log_.fit_transform(X, y)
+    output_inverse_y = log_.inverse_transform(output_y)
+    pd.testing.assert_series_equal(pd.Series(y), output_inverse_y, check_dtype=False)
```
bchen1116, line 71: Would it be useful to do these tests with the time series regression data as well?

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
```diff
@@ -0,0 +1,170 @@
...
+    X, y = X_y_regression
+
+    target_dist_check = TargetDistributionDataCheck("regression")
+
+    if distribution == "normal":
+        y = norm.rvs(loc=3, size=10000)
+    elif distribution == "lognormal":
+        y = lognorm.rvs(0.4, size=10000)
+    else:
+        # Will have a p-value of 0 the null hypothesis even after log transforming
```
bchen1116, line 125: This comment is confusing..

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
```diff
@@ -0,0 +1,170 @@
...
+                    },
+                ).to_dict()
+            ],
+            "actions": [],
+        }
+
+
+@pytest.mark.parametrize("data_type", ["positive", "mixed", "negative"])
+@pytest.mark.parametrize("distribution", ["normal", "lognormal", "lognormal_"])
+def test_target_distribution_data_check_warning_action(
```
bchen1116, line 113: So with `lognormal` vs `lognormal_`, is the difference that one is much more lognormal (`lognormal_`) compared to the other? I don't understand why having the `p_value=0` is important here, since I assume the `p_value` is low enough to reject the null hypothesis with `lognormal`?

bchen1116 (changes requested):
The implementation looks pretty good! Left some comments on things that I think we could fix, as well as clarifying questions for my own understanding! I don't think we need to include the `problem_type` as an arg for this datacheck, but I'm open to discussing it!

docs/source/user_guide/data_checks.ipynb
```diff
@@ -419,6 +420,40 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Target Distribution\n",
+    "\n",
+    "Target data can come in a variety of distributions, such as Gaussian or Lognormal. When we work with machine learning models, we feed data into an estimator that learns from the training data provided. Sometimes the data can be significantly spread out with a long tail or outliers, which could lead to a lognormal distribution. This can cause machine learning model performance to suffer.\n",
+    "\n",
+    "To help the estimators better understand the underlying relationships in the data between the features and the target, we can use the `TargetDistributionDataCheck`to identify such a distribution. If you use `AutoML.search` to try and find the best pipeline, it will automatically run this data check for you and if a lognormal distribution is detected, will add a `LogTransformer` to your pipeline to help the model performance!"
```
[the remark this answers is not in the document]
ParthivNaresh: I like the way you think

evalml/pipelines/regression_pipeline.py
```diff
@@ -54,3 +54,10 @@ def score(self, X, y, objectives):
         return self._score_all_objectives(
             X, y, y_predicted, y_pred_proba=None, objectives=objectives
         )
+
+    def predict(self, X, objective=None):
```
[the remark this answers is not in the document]
ParthivNaresh: Yep that's right!

evalml/pipelines/utils.py
ParthivNaresh: Yeah the reason it's set up like this is if we add more target checks in the future

evalml/tests/component_tests/test_log_transformer.py
ParthivNaresh: Oops reworded!

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
ParthivNaresh: Oops reworded!

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
ParthivNaresh: I've updated it with a comment that clarifies it. Basically this is meant to check the `=` part of `shapiro_test_log.pvalue >= shapiro_test_og.pvalue`. In the event that the target data is log transformed and doesn't reduce its `p-value`, it falls under the assumption that it's as normal if not more `normal` than before. That more or less implies that the underlying distribution had some log normality to it. The `p-value` is low enough that we can reject the null hypothesis that the distribution is `normal`, but a distribution that maintains or increases its `p-value` post log transformation would give us enough of a reason to believe that log normality exists.

evalml/tests/component_tests/test_log_transformer.py
ParthivNaresh: Good point!

evalml/data_checks/default_data_checks.py
```diff
@@ -47,12 +49,13 @@ def __init__(self, problem_type, objective, n_splits=3):
...
             super().__init__(
-                self._DEFAULT_DATA_CHECK_CLASSES,
+                self._DEFAULT_DATA_CHECK_CLASSES + [TargetDistributionDataCheck],
                 data_check_params={
                     "InvalidTargetDataCheck": {
                         "problem_type": problem_type,
                         "objective": objective,
-                    }
+                    },
+                    "TargetDistributionDataCheck": {"problem_type": problem_type},
```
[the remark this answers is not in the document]
ParthivNaresh: So these were the reasons I felt it was necessary:

- I had originally added it in case open source users try leveraging it against their data outside of `DefaultDataChecks`. A binary or multiclass problem that consists of a few unique numbers in the target data could actually be interpreted by the `TargetDistributionDataCheck` as having a lognormal distribution although that data check message wouldn't make much sense for non regression problems, and a user without much data science experience might not know what to do with it.
- I don't think most DataChecks need `problem_type` in their implementation, so we leave it out. However I do feel like `ClassImbalanceDataCheck` not having it is more of an argument in favour of including it for `DataChecks` that rely on `problem_type` like `UniquenessDataCheck` rather than against including it here.

I don't mind removing it at all, since adding it here makes me feel like we should add it for any `DataChecks` that rely on the exclusion of a problem type. What do you think?

chukarsten (changes requested):
Even though I'm requesting changes, they're super minor.  I think that instead of doing that list addition, you might as well just add the datacheck to the defaults, unless there's a reason I'm missing.  If there's a good reason for doing it that way, let me know and I'll approve.  But this looks good - good work.

evalml/data_checks/default_data_checks.py
```diff
@@ -47,12 +49,12 @@ def __init__(self, problem_type, objective, n_splits=3):
             ProblemTypes.TIME_SERIES_REGRESSION,
         ]:
             super().__init__(
-                self._DEFAULT_DATA_CHECK_CLASSES,
+                self._DEFAULT_DATA_CHECK_CLASSES + [TargetDistributionDataCheck],
```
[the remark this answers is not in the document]
ParthivNaresh: This Data Check (as it exists currently anyways) only checks for a regression distribution. Just like how `ClassImbalanceDataCheck` is classification only, it wouldn't make much sense to run `TargetDistributionDataCheck` for classification problems.

chukarsten (approved):
Once CodeCov is good, I'm good

evalml/data_checks/default_data_checks.py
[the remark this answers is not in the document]
bchen1116: @ParthivNaresh This makes sense. I think this does set a new precedent of having to add the `problem_type` arg to `DataChecks` that rely on specific problem types, but that isn't necessarily a bad thing. If this seems right to you, we can move forward with it!

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
bchen1116: Thanks for the clarification!

bchen1116 (approved):
Nice!, LGTM! Thanks for addressing all the comments. Left one lil nit, but great work on this one!

PR #2488, opened 2021-07-09
Add a DeprecationWarning about deprecating the list API for ComponentGraphs

### LINKED ISSUE #2483: Add a DeprecationWarning about deprecating the list API for ComponentGraphs
Blocks #2482, per history, we should add a DeprecationWarning a sprint ahead about our plan to deprecate the list API for ComponentGraphs.

angela97lin:
Closes #2483

### DIFF
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -3,6 +3,7 @@ import inspect
 import os
 import sys
 import traceback
+import warnings
 from abc import ABC, abstractmethod
 from collections import OrderedDict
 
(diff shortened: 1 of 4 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -66,9 +67,14 @@ def __init__(
         self.random_seed = random_seed
 
         if isinstance(component_graph, list):  # Backwards compatibility
-            self.component_graph = ComponentGraph().from_list(
-                component_graph, random_seed=self.random_seed
-            )
+            with warnings.catch_warnings():
```
angela97lin, line 70: I added this here to filter for the DeprecationWarning because users who are using AutoMLSearch or PipelineBase might be confused as to why they're receiving this. I think the warning should only surface if the user directly initializes a ComponentGraph using a list directly, Pipelines should still handle lists of components :)

jeremyliweishih (approved):
nice! I like the warning catch.

evalml/pipelines/pipeline_base.py
dsherry: Great call

dsherry (approved):
🚢 😁

PR #2490, opened 2021-07-09
Deprecate the list API for ComponentGraph

### LINKED ISSUE #2481: Update ComponentGraph to enforce needing .x and .y for each component specified in the graph
Part of https://alteryx.atlassian.net/wiki/spaces/PS/pages/616431882/Component+Graph+Updates+Design+and+Plan+Final, we want to update the ComponentGraph definition s.t. all edges are explicit.

### LINKED ISSUE #2482: Deprecate the list API for ComponentGraph
Part of https://alteryx.atlassian.net/wiki/spaces/PS/pages/616431882/Component+Graph+Updates+Design+and+Plan+Final, we want to deprecate use of the list API for ComponentGraphs.

angela97lin:
Closes #2481 and closes #2482.

I won't merge this until a release after #2483 😛 

This PR does the following:
- ~Moves `get_hyperparameter_ranges` from automl/utils to a method on `PipelineBase`. The method was only called in `AutoMLAlgorithm`, and was passed the pipeline's component graph anyway, so this isn't a big lift. Main reason for this was because it called ComponentGraph's `linearized_component_graph` method which is removed in this PR.~ Work done in #2546 
- ~Removes `from_list`~ Work done in #2549 
- ~Rename `get_parents` to `get_inputs`; open to other suggestions but I think this is a little more inclusive given that we no longer only take in parent components but potentially also X and y, and that now all inputs have to be specified with .x/.y (estimator names alone do not work)~ Work done in #2540
- Enforces that a component dict needs to specify .x and .y for each component. Component names are no longer allowed, since it could lead to some confusion about what that means.
- ~Moved `_make_component_dict_from_component_list` from `pipeline/utils.py` to `PipelineBase` as a static method. This was to enable using it in `PipelineBase` without causing a circular dependency 😬 (since `pipeline/utils.py` imports the base pipeline classes)~ Work done in #2549 

TODO:
- Remove component name == get X from component (see estimator)
- If y is specified and component does not produce y output, error out. Ex: "Imputer.y"
- Test that estimator names alone do not work, need .x/.y
- Update notebooks to discuss how to initialize component graph and requirements.


evalml/automl/utils.py
```diff
@@ -207,29 +205,6 @@ def get_best_sampler_for_data(X, y, sampler_method, sampler_balanced_ratio):
             return "Undersampler"
 
 
-def get_hyperparameter_ranges(component_graph, custom_hyperparameters):
```
angela97lin, line 210: Moved to `PipelineBase` so we don't have to use `ComponentGraph.linearized_component_graph`, which I deleted

evalml/pipelines/component_graph.py
```diff
@@ -366,7 +282,7 @@ def _compute_features(self, component_list, X, y=None, fit=False):
                     output = component_instance.predict(input_x)
                 else:
                     output = None
-                output_cache[component_name] = output
+                output_cache[f"{component_name}.x"] = output
```
angela97lin, line 285: This change means that Estimators' outputs will be specified via `EstimatorName.x` rather than `EstimatorName`. I think this is clearer, since we are moving away from the expectation that just using a component name will default to the features / .x, now that we have more components which return valid .y values :)

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -157,73 +157,68 @@ def test_init_str_components():
...
 
 def test_invalid_init():
-    invalid_graph = {"Imputer": [Imputer], "OHE": OneHotEncoder}
+    invalid_graph = {"Imputer": [Imputer, "X", "y"], "OHE": OneHotEncoder}
     with pytest.raises(
         ValueError, match="All component information should be passed in as a list"
     ):
         ComponentGraph(invalid_graph)
 
-    with pytest.raises(
```
angela97lin, line 166: Duplicate test of below, so deleting :P

angela97lin:
Closing in favor of #2563

PR #2495, opened 2021-07-12

machineFL:


dsherry (approved):

freddyaboulton:
Windows binaries have not been uploaded yet so I think this update is breaking our windows tests:

https://pypi.org/project/sktime/0.7.0/#files

Looks like they also refactored their module structure so we need to update one of our tests

angela97lin (approved):
🚢

PR #2496, opened 2021-07-12
Update components with attribute that describes if they can be used in list API for Pipelines

### LINKED ISSUE #2494: Update components with attribute that describes if they can be used in list API for Pipeline
Part of https://alteryx.atlassian.net/wiki/spaces/PS/pages/616431882/Component+Graph+Updates+Design+and+Plan+Final, we want to update components with an attribute that describes if they can be used in list API for Pipeline. If not, users must define pipelines using the graph/dict representation instead.

angela97lin:
Closes #2494 

Adds a private variable to be used in deprecating our list API for complex pipelines. I chose to be more restrictive by setting  `_supported_by_list_API = False` as the default and requiring components to override this. This is so when we add new components such as the LogTransformer (where `_supported_by_list_API` should be set to False), there is not a chance of accidentally forgetting to set the value.

On the other hand, this means for other components where ``_supported_by_list_API` should be True and we forget to set it, we have a bug of being too restrictive.

I like this behavior better than setting the default to True and then accidentally enabling some components where `_supported_by_list_API` should have been set to False to run with undefined / inaccurate behavior, but open to discussion of course 😁 

--
Alternatively: Do all `_supported_by_list_API = False` components also happen to be components that could potentially modify the target? Would it be safe and more descriptive to call the attribute `modifies_target`?

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
+++ b/evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
@@ -17,8 +17,8 @@ class PolynomialDetrender(TargetTransformer):
     """
 
     name = "Polynomial Detrender"
-
     hyperparameter_ranges = {"degree": Integer(1, 3)}
+    _supported_by_list_API = False
 
     def __init__(self, degree=1, random_seed=0, **kwargs):
         if not isinstance(degree, int):
--- a/evalml/tests/component_tests/test_components.py
+++ b/evalml/tests/component_tests/test_components.py
@@ -68,6 +68,12 @@ from evalml.pipelines.components.ensemble import (
     StackedEnsembleClassifier,
     StackedEnsembleRegressor,
 )
+from evalml.pipelines.components.transformers.samplers.base_sampler import (
+    BaseSampler,
+)
+from evalml.pipelines.components.transformers.transformer import (
+    TargetTransformer,
+)
 from evalml.pipelines.components.utils import (
     _all_estimators,
     _all_transformers,
@@ -1628,3 +1634,15 @@ def test_estimator_fit_respects_custom_indices(
     estimator.fit(X, y)
     pd.testing.assert_index_equal(X.index, X_original_index)
     pd.testing.assert_index_equal(y.index, y_original_index)
+
+
+def test_component_parameters_supported_by_list_API():
+    for component_class in all_components():
+        if (
+            issubclass(component_class, BaseSampler)
+            or issubclass(component_class, TargetTransformer)
+            or component_class in [TargetImputer]
+        ):
+            assert not component_class._supported_by_list_API
+        else:
+            assert component_class._supported_by_list_API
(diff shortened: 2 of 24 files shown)


requirements.txt
```diff
@@ -12,4 +12,4 @@ category_encoders>=2.2.2
 statsmodels >= 0.12.2
 imbalanced-learn>=0.8.0
 pmdarima==1.8.0
-sktime>=0.5.3;python_version<"3.9"
+sktime>=0.5.3,<0.7.0;python_version<"3.9"
```
angela97lin, line 15: Will revert later :)

evalml/tests/component_tests/test_components.py
```diff
@@ -1628,3 +1634,15 @@ def test_estimator_fit_respects_custom_indices(
     estimator.fit(X, y)
     pd.testing.assert_index_equal(X.index, X_original_index)
     pd.testing.assert_index_equal(y.index, y_original_index)
+
+
+def test_component_parameters_supported_by_list_API():
```
angela97lin, line 1639: Testing private attribute... but I think it could be good to have this so that we're aware each time a new component is added :)

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -17,8 +17,8 @@ class PolynomialDetrender(TargetTransformer):
     """
 
     name = "Polynomial Detrender"
-
     hyperparameter_ranges = {"degree": Integer(1, 3)}
+    _supported_by_list_API = False
```
freddyaboulton, line 21: We technically get this from the base class so this is just to be explicit?

evalml/tests/component_tests/test_components.py
freddyaboulton: Sounds good to me!

chukarsten: Would we want to just smoke test the component to make sure that those that support the list API do and those that don't, do not?

chukarsten (approved):
ngl, I got nervous when I saw 24 files changed

freddyaboulton (approved):
@angela97lin This looks good to me! Regarding your comments about having a more descriptive name and what the default value should be - have you considered the following design:

1. Call the property `modifies_target` and makes it abstract

```python
    @property
    @classmethod
    @abstractmethod
    def modifies_target(cls):
        """Returns whether this component modifies (subsets or transforms) the target variable."""
```

2. Make the `_supported_by_list_api` a `classproperty`

```python
    @classproperty
    def _supported_by_list_api(cls):
        return not cls.modifies_target
```

The benefits of this approach are that we never forget to set it for a new component and users can use the list api with their custom components by answering a simple question about their component. If the logic for determining the list api gets more complicated in the future, we can add more classproperties to capture those cases or have a different `supports_list_api` implementation for those components.

bchen1116 (approved):
Nice! This doesn't yet enforce that pipelines can't be defined as lists with the components that have `_supported_by_list_API=False`. Is that an issue filed for the future?

requirements.txt
ParthivNaresh: Was this related to the PR? Just curious :)

ParthivNaresh (approved):
Love it!

angela97lin:
Closing, addressed by #2504 :0

PR #2504, opened 2021-07-13
Update make_pipelines for AutoML to create component dictionaries, not lists, to initialize pipelines

### LINKED ISSUE #2493: Update make_pipelines for AutoML to create ComponentGraphs, not lists, to initialize pipelines
In AutoML, we create pipelines to search over via `make_pipelines`. We should update this to generate component graphs to pass to pipelines, not lists, once our API is updated. This way, we don’t end up with any tricky bugs in AutoML stemming from our list logic.

### LINKED ISSUE #2494: Update components with attribute that describes if they can be used in list API for Pipeline

angela97lin:
Closes #2493, closes #2494 (in lieu of https://github.com/alteryx/evalml/pull/2496).

This PR does a few things:
- Adds a private variable to be used in deprecating our list API for complex pipelines (`_supported_by_list_API`). Also adds two class properties, `modifies_target` and `modifies_features`. `_supported_by_list_API` will be used in #2490 when we actually deprecate the list API, but adding here because of its current connection to `modifies_target`.
- Updates  our component graph to first generate the appropriate component dict from a list of components internally for `make_pipeline`, rather than passing in the list of components directly. This is also related to eventually deprecating the list API.

Q:
- We wanted to deprecate the list API because of ambiguity, but in `make_pipeline`, we needed a way to convert a list of components to a ComponentGraph. Is this reliable enough that we don't have to deprecate the list API after all? My hunch says it is okay to be more restrictive for now (still deprecate list API) since it relies on the user to have  `modifies_target` and `modifies_features` properly set for their components, and this is a restraint we can always loosen later, but curious about thoughts 😁 

Also welcome any other graph examples you guys think of for testing 😁


evalml/tests/component_tests/test_components.py
```diff
@@ -1628,3 +1646,33 @@ def test_estimator_fit_respects_custom_indices(
...
+            assert not component_class.modifies_target
+        if issubclass(component_class, TargetTransformer) or component_class in [
+            TargetImputer
+        ]:
+            assert not component_class.modifies_features
+        else:
+            assert component_class.modifies_features
+
+
+def test_component_parameters_supported_by_list_API():
```
angela97lin, line 1669: Will smoke test in #2490. Right now, all components are still "supported" in list API, until it's deprecated in #2490 :)

evalml/tests/component_tests/test_components.py
```diff
@@ -1628,3 +1646,33 @@ def test_estimator_fit_respects_custom_indices(
     estimator.fit(X, y)
     pd.testing.assert_index_equal(X.index, X_original_index)
     pd.testing.assert_index_equal(y.index, y_original_index)
+
+
+def test_component_parameters_modifies_feature_target():
```
chukarsten, line 1651: I'm not sure I understand the name of this test

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -986,3 +993,110 @@ def __init__(self, random_arg=False, random_seed=0):
     )
     pipeline = generate_pipeline_code(mock_pipeline_with_custom_components)
     assert pipeline == expected_code
+
+
+def test_make_component_dict_from_component_list():
```
chukarsten, line 998: I like this test, nice.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -986,3 +993,110 @@ def __init__(self, random_arg=False, random_seed=0):
...
+        [Undersampler, Undersampler, RandomForestClassifier]
+    ) == {
+        "Undersampler": [Undersampler, "X", "y"],
+        "Undersampler_1": [Undersampler, "Undersampler.x", "Undersampler.y"],
+        "Random Forest Classifier": [
+            RandomForestClassifier,
+            "Undersampler_1.x",
+            "Undersampler_1.y",
+        ],
+    }
```
chukarsten, line 1102: Very thorough, nice.

chukarsten (approved):
Awesomesauce.  Looks good to me!  Nice work.

ParthivNaresh (approved):
Looks great, and I love the diversity of tests. What would happen if you added a non-estimator as the last component of the component list?
Regarding your question, I think it's perfectly fine to deprecate the list API for the reason you gave. However if we choose not to, I think it would be helpful to update the [docs](https://evalml.alteryx.com/en/stable/user_guide/components.html#Defining-Custom-Components) with a line about the inclusion of `modifies_target` and `modifies_features`.

evalml/pipelines/components/component_base.py
```diff
@@ -52,6 +52,20 @@ def name(cls):
     def model_family(cls):
         """Returns ModelFamily of this component"""
 
+    @property
+    @classmethod
+    @abstractmethod
+    def modifies_features(cls):
+        """Returns whether this component modifies (subsets or transforms) the features variable during transform.
+        For Estimator objectss, this attribute determines if the return value from `predict` or `predict_proba` should be used as features or targets."""
```
bchen1116, line 60: typo: `objects`

evalml/pipelines/utils.py
```diff
@@ -131,6 +131,33 @@ def _get_preprocessing_components(
...
+        component_name = component.name
+        if component_name in seen:
+            component_name = f"{component_name}_{idx}"
+        seen.add(component_name)
+        components_with_names.append((component_name, component))
+
+    component_dict = {}
+    most_recent_target = "y"
+    most_recent_features = "X"
+    for component_name, component_class in components_with_names:
```
bchen1116, line 148: nit, but curious if it'd be possible to combine these for loops into 1 loop?

```python
seen = set()
component_dict = {}
most_recent_target = "y"
most_recent_features = "X"
for idx, component in enumerate(component_list):
    component_name = component.name
    if component_name in seen:
        component_name = f"{component_name}_{idx}"
    seen.add(component_name)

    component_dict[component_name] = [
        component,
        most_recent_features,
        most_recent_target,
    ]
    if component.modifies_target:
        most_recent_target = f"{component_name}.y"
    if component.modifies_features:
        most_recent_features = f"{component_name}.x"
return component_dict
```

evalml/tests/pipeline_tests/test_pipeline_utils.py
bchen1116: Seconded, I like this test!

bchen1116 (approved):
Great job with the clean implementation and the easy-to-follow and thorough tests! LGTM! Left a suggestion on something we can maybe simplify, but not blocking.

Agreed with @ParthivNaresh about updating the docs with `modifies_features` and `modifies_targets`.

evalml/pipelines/components/component_base.py
```diff
@@ -80,6 +94,10 @@ def default_parameters(cls):
 
         return cls._default_parameters
 
+    @classproperty
+    def _supported_by_list_API(cls):
```
freddyaboulton, line 98: If the plan is move ahead with deprecating the list api for pipelines, I'm wondering if we even need this anymore?

It's currently only used in `make_pipeline` and it only called on the components from `_get_preprocessing_components`. That's an implementation we control - so those components should always support the list api right?

I guess what I'm trying to get at is that we decide to get rid of the list api, then the only "list to dict" conversion that would exist is in `make_pipeline` but that's only because `_make_preprocessing_components` was written way before the `ComponentGraph` abstraction! If you're with me so far, then I think what we should do is have `_make_preprocessing_components` generate a component graph dict directly. 

That being said, I think this suggestion would only apply if we decide to deprecate the list api completely. If we go with the "half-half" solution from last week, I think the way you have this laid out currently is good!

I'm curious what you think about all this. It seems we're still on the fence about whether to get rid of the list api? 😅 Maybe this is something to broach with more members of the team?

evalml/tests/component_tests/test_components.py
angela97lin: Relatable, copy pasta failure :)

evalml/pipelines/utils.py
```diff
@@ -131,6 +131,33 @@ def _get_preprocessing_components(
...
+        component_dict[component_name] = [
+            component_class,
+            most_recent_features,
+            most_recent_target,
+        ]
+        if component_class.modifies_target:
+            most_recent_target = f"{component_name}.y"
```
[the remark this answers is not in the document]
angela97lin: @ParthivNaresh you can't have multiple y parents as input! So if you have multiple target transformers, you could specify the target transformer's "X" input, if any or useful, but only one y :)

evalml/pipelines/utils.py
angela97lin: Yup, I like this! 😁

angela97lin:
@ParthivNaresh Added a test with transformer after estimator, and answered your question. LMK if there's still any confusion! :) Thanks for the doc suggestion, added! (@bchen1116)

@freddyaboulton Ahh yes, there's so much back and forth 😂 I ultimately decided that since post-conversation last week, we decided to do the half-half solution, that's what I'll do in #2490 (in the process of updating right now, RIP). With that in mind, I'll keep `_supported_by_list_API` with its simple impl right now. `make_pipeline` still has some logic of grabbing all of the components in order first, then generating the graph rather than generating the graph as we add components because when I tried to impl the dict as we went along, the logic was not any simpler :/ With that in mind, going to keep with our half-half sol and as Parthiv suggested, add the public attributes to our doc 😁

PR #2505, opened 2021-07-13
Clean up files written to tmpdir during tests

### LINKED ISSUE #2337: Allow test_saving_png_file to run during build_conda_pkg job
In #2323, we disabled `test_saving_png_file` during the conda build because we were getting `out of memory` from `Popen` commands issued by graphviz (see image below). This is pretty confusing and my guess is that the directory where the png files are stored in the docker container (pytest tmpdir fixture) is running out of space. The fact that this is all happening within docker may make debugging a bit more complicated.

One possible thing to try is to have each test delete any files it creates once the test finishes and see if it fixes the problem.

image

freddyaboulton:
### Pull Request Description
Fixes #2337

`build_conda_pkg` has passed three times already. It used to fail consistently when `test_save_png_file` was enabled. I ran locally with `--basetemp mydir` for `main` and `--basetemp mydir-new-fixture` for this branch. The total space on this branch is 0 bytes compared to 22 mb on main.

image

-----

### DIFF
--- a/.github/workflows/build_conda_pkg.yml
+++ b/.github/workflows/build_conda_pkg.yml
@@ -6,10 +6,6 @@ on:
       - main
   pull_request:
     types: [ opened, synchronize ]
-    paths:
-      - 'requirements.txt'
-      - 'core-requirements.txt'
-      - 'evalml/tests/dependency_update_check/*.txt'
 
 jobs:
   shellcheck:
--- a/.github/workflows/build_conda_pkg.yml
+++ b/.github/workflows/build_conda_pkg.yml
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
     types: [ opened, synchronize ]
+    paths:
+      - 'requirements.txt'
+      - 'core-requirements.txt'
+      - 'evalml/tests/dependency_update_check/*.txt'
 
 jobs:
   shellcheck:
--- a/evalml/tests/conftest.py
+++ b/evalml/tests/conftest.py
@@ -1291,3 +1291,9 @@ class _AutoMLTestEnv:
 @pytest.fixture
 def AutoMLTestEnv():
     return _AutoMLTestEnv
+
+
+@pytest.fixture
+def tmpdir_with_cleanup(tmpdir):
+    yield tmpdir
+    tmpdir.remove(ignore_errors=True)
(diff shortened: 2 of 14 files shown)


evalml/tests/conftest.py
```diff
@@ -1291,3 +1291,9 @@ def test_context(
 @pytest.fixture
 def AutoMLTestEnv():
     return _AutoMLTestEnv
+
+
+@pytest.fixture
+def tmpdir_with_cleanup(tmpdir):
```
freddyaboulton, line 1297: There are other ways to clean up files after a test that don't involve using a new fixture (see [here](https://stackoverflow.com/questions/37563396/deleting-py-test-tmpdir-directory-after-successful-test-case)) but this seems simpler to me

.github/workflows/build_conda_pkg.yml
```diff
@@ -6,10 +6,6 @@ on:
       - main
   pull_request:
     types: [ opened, synchronize ]
-    paths:
```
freddyaboulton, line 9: Will add back before merge

evalml/tests/conftest.py
```diff
@@ -1291,3 +1291,9 @@ def test_context(
...
+    yield tmpdir
+    tmpdir.remove(ignore_errors=True)
```
freddyaboulton, line 1299: I tested locally that this removes the directory even if the test fails.

evalml/tests/conftest.py
jeremyliweishih: why not just use one fixture instead of having one for the tmpdir and one for cleaning up? Are there some tests we don't want the clean up functionality?

jeremyliweishih (approved):
looks good! just one question.

evalml/tests/conftest.py
freddyaboulton: I thought about this but tbh

```python
@pytest.fixture
def tmpdir(tmpdir):
    yield tmpdir
    tmpdir.remove(ignore_errors=True)
```

looked weird to me. That being said, I will try the following:

```python
@pytest.fixture
def tmpdir(tmp_path):
    dir = py.path.local(tmp_path)
    yield dir
    dir.remove(ignore_errors=True)
```

which I think will work and will involve less code changes!

freddyaboulton: Pushed this up! Thanks for the suggestion - it's working nicely

angela97lin (approved):
I'm surprised that this isn't something built into tmpdir :( but looks good, thank you!

PR #2516, opened 2021-07-15
Fix the windows 3.7 nightlies

freddyaboulton:
### Pull Request Description
Fix the windows 3.7 nightlies

Our windows ci installs graphviz with conda and then with pip when we run `pip install .`

If you look at a successful 3.7 run from [before last night](https://github.com/alteryx/evalml/runs/3062986346?check_suite_focus=true), you should see `Requirement already satisfied: graphviz>=0.13 in c:\users\runneradmin\miniconda3\envs\curr_py\lib\site-packages (from evalml==0.28.0) (0.16)
`

If you look at the run that [failed last night](https://github.com/alteryx/evalml/runs/3072898319?check_suite_focus=true), you will see: 
`Collecting graphviz>=0.13 Downloading graphviz-0.17-py3-none-any.whl (18 kB)`. 

So for some reason, the pip version is starting to override the conda version. This only happens for 3.7! The logs for the successful 3.8 runs from[ last night](https://github.com/alteryx/evalml/runs/3072898500?check_suite_focus=true) say `Requirement already satisfied: graphviz>=0.13 in c:\users\runneradmin\miniconda3\envs\curr_py\lib\site-packages (from evalml==0.28.0) (0.16)`

Not exactly sure why this is happening, but since conda is able to successfully install graphviz in a way that makes our unit tests pass I think we can avoid the problem if we skip the pip install for windows.

Our docs already tell users to install graphviz with windows so I think we're covered there. 

-----


angela97lin (approved):
Definitely odd, but the explanation makes sense. Thanks for looking into this!
Even more odd that I couldn't find anything on the graphviz side about this 🤔

requirements.txt
```diff
@@ -6,7 +6,7 @@ xgboost>=1.4.2
 catboost>=0.20
 lightgbm>=2.3.1
 matplotlib>=3.3.3
-graphviz>=0.13
+graphviz>=0.13;platform_system!='Windows'
```
bchen1116, line 9: This is really cool!

bchen1116 (approved):
LGTM! Thanks for hunting this down, definitely a weird bug!

PR #2519, opened 2021-07-15
Fix formatting of objective API reference

angela97lin:
Closes ##2447


angela97lin:
This works (fix is to leave a newline between description and Arguments:) but it's so insignificant I'm going to close and combine with #2520 lol

Issue #2439, opened 2021-06-24
Design spike: automl and featuretools feature engineering


dsherry:
**Background**
There's been [discussion recently](https://alteryx.quip.com/120hAOlhXOg8/AutoML-Upgrade-Feature-EngineeringFeature-Selection) about how featuretools fits in with automl. We currently ask users to run featuretools before running automl. We've been exploring ways to have featuretools run as part of automl, or for a feature store to allow information to be shared between the two.

**Challenges**
One of the things we'd like to do in our new automl algorithm is to run a baseline pipeline without any engineered features. Doing so will give us the ability to show users how much adding featuretools' feature engineering will improve modeling. In order to do this, automl needs a) access to all original unmodified features b) access to all featuretools' engineered features c) a clear way to tell which is which. Q: How can we give automl the ability to ?

An additional related challenge has to do with where we run certain types of feature engineering. Currently, automl adds our text featurizer and datetime featurizer components to our pipelines if needed, and under the hood those components use featuretools primitives. Q: Should we be doing all our feature engineering in the pipelines created by automl, should we be doing all our feature engineering before automl runs, or is it ok to continue to have a hybrid approach?

There may also be implications here from time series. Automl owns data splitting during model training. When we add time series support ( @freddyaboulton ), we'll need to run time-based operations like windowed moving averages and delayed features, which share information between the rows. Q: do our goals for time series mean that we must run those feature engineering steps in automl, after data splitting?

freddyaboulton:
@jeremyliweishih FYI I'm adding this to our "Supporting All WW Types" Epic (#2443) because featuretools will soon adopt ww typing and the question of where we should run featuretools computations is directly related to supporting the full set of ww types. 

This spike is also related to Spike #2423

dsherry:
Another illustrative example for this: #2444 . FT is planning to add url and email address primitives.

dsherry:
One more point to drop: we don't run featuretools in LG currently. We should discuss that.

gsheni:
@jeremyliweishih We have added URL and EmailAddress primitives to Featuretools and released them.
- https://github.com/alteryx/featuretools/releases/tag/v0.26.0

jeremyliweishih:
Initial design spike has been completed [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/654442499/Feature+engineering+featuretools+and+evalml). Will continue and fully flesh out a design doc moving forward.

PR #2525, opened 2021-07-19
Add `EvalMLAlgorithm`

### LINKED ISSUE #2347: AutoML Algorithm Upgrade: new algorithm
https://alteryx.quip.com/CDiuAtU8WpUc/AutoML-Upgrade-Design-Doc

* Algorithm implementation (2 weeks)
    * update RF feature selectors to take in estimator

Define new automl algorithm, without changing the behavior of the existing one.

jeremyliweishih:
Fixes #2347. 

Still need do the following:

- [x] ensure test coverage
- [x] remove duplicate code
- [x] remove unnecessary impl (FS custom estimator mainly)

Saving custom_hyperparameter and pipeline params for AutoMLSearch integration.


evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,367 @@
...
+    _make_stacked_ensemble_pipeline,
+    make_pipeline,
+)
+from evalml.problem_types import is_regression
+
+
+class EvalMLAlgorithm(AutoMLAlgorithm):
+    """An automl algorithm that consists of two modes: fast and long. Where fast is a subset of long.
+
+    1. Naive pipelines:
```
jeremyliweishih, line 31: Thought it would be nice to list it out here - can remove as well.

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,367 @@
...
+        X,
+        y,
+        problem_type,
+        _sampler_name,
+        tuner_class=None,
+        random_seed=0,
+        pipeline_params=None,
+        custom_hyperparameters=None,
+        n_jobs=-1,
+        number_features=None,
```
jeremyliweishih, line 70: Not sure if `number_features` is still necessary anymore. I still see it in `IterativeAlgorithm` but don't see it mentioned anywhere else in the codebase.

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,367 @@
...
+
+        self.n_jobs = n_jobs
+        self.number_features = number_features
+        self._best_pipeline_info = {}
+        self.text_in_ensembling = text_in_ensembling
+        self._pipeline_params = pipeline_params or {}
+        self._custom_hyperparameters = custom_hyperparameters or {}
+        self._selected_cols = None
+        self._top_n_pipelines = None
+        if custom_hyperparameters and not isinstance(custom_hyperparameters, dict):
```
jeremyliweishih, line 108: There's some duplicate logic from `IterativeAlgorithm` that we can refactor into a base class later on.

evalml/automl/automl_algorithm/evalml_algorithm.py
jeremyliweishih: Note: this current algorithm doesn't have any feature engineering steps as that is blocked by #2439.

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,365 @@
...
+        )
+        return [ensemble]
+
+    def _create_long_top_n(self, n):
+        estimators = [
+            (pipeline_dict["pipeline"].estimator, pipeline_dict["mean_cv_score"])
+            for pipeline_dict in self._best_pipeline_info.values()
+        ]
+        estimators.sort(key=lambda pipeline: pipeline[1])
+        estimators = estimators[:n]
```
bchen1116, line 238: I see that this is choosing the best estimators based on the `mean_cv_score`, which makes sense. I think the sorting should be done based on what objective is being used, since `R2==1` is the best for regression problems but `log loss == 0` is the best for classification problems.

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,365 @@
...
+            next_batch = self._create_naive_pipelines_with_feature_selection()
+        elif self._batch_number == 2:
+            next_batch = self._create_fast_final()
+        elif self.batch_number == 3:
+            next_batch = self._create_ensemble()
+        elif self.batch_number == 4:
+            next_batch = self._create_long_top_n(n=3)
+        elif self.batch_number % 2 != 0:
+            next_batch = self._create_ensemble()
+        else:
```
bchen1116, line 287: Nice! I like this if/elif/else logic

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,365 @@
...
+            next_batch = []
+            for _ in range(10):
```
bchen1116, line 289: Im curious if it'd be beneficial to make this for loop its own separate helper function. It could then make the code cleaner/less repetitive for both this spot and when you're finding the 50 instances of each pipeline.

bchen1116 (commented):
Did a quick look-over and it's looking pretty good! Left a few suggestions on ways to clean up the code.

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,365 @@
...
+            self._best_pipeline_info.update(
+                {
+                    pipeline.model_family: {
+                        "mean_cv_score": score_to_minimize,
+                        "pipeline": pipeline,
+                        "parameters": pipeline.parameters,
+                        "id": trained_pipeline_results["id"],
+                    }
+                }
+            )
```
chukarsten, line 349: I'm not sure how I feel about all this being in `add_result()`, not like you really have much other choice, but I think maybe some of what's happening might want to make its way into the docstring.  I'm not 100% sure what _selected_cols is, or why it's being updated and it's also good to know that the _best_pipeline_info is being updated here.

evalml/pipelines/utils.py
```diff
@@ -190,7 +191,12 @@ def make_pipeline(
     preprocessing_components = _get_preprocessing_components(
         X, y, problem_type, estimator, sampler_name
     )
-    complete_component_list = preprocessing_components + [estimator]
+
+    complete_component_list = (
+        preprocessing_components + extra_components + [estimator]
+        if extra_components
+        else preprocessing_components + [estimator]
+    )
```
chukarsten, line 199: I'm not sure you need to do this.  

```
>>> x = [1,2,3]
>>> x + []
[1, 2, 3]
```

evalml/tests/automl_tests/test_evalml_algorithm.py
```diff
@@ -0,0 +1,169 @@
...
+    mock_get_names.return_value = ["0", "1", "2"]
+    problem_type = automl_type
+    sampler_name = None
+    algo = EvalMLAlgorithm(X, y, problem_type, sampler_name)
+    naive_model_families = [ModelFamily.LINEAR_MODEL, ModelFamily.RANDOM_FOREST]
+
+    first_batch = algo.next_batch()
+    assert len(first_batch) == 2
+    for pipeline in first_batch:
+        assert pipeline.model_family in naive_model_families
```
chukarsten, line 115: I don't think this is blocking, but it's possible naive_model_families has multiple different families but each batch only has one of those types.  Might need to do a set equality in addition to the length equality to be sure.

chukarsten (changes requested):
Jeremy, so sorry I delayed you for such a long time.  This is really good work.  Not going to lie - I was intimidated by the IterativeAlgorithm, but this really feels and looks a lot cleaner and easier to understand.  I think there's a few places that can use some cleanup or refactor.

1. add_result() *maybe* needs a breakup.  It's a little thick and could use a bit more addition to the docstring.
2. The hardcoded 10 and 50 should be replaced with variables and possibly exposed as params for the EvalMLAlgorithm
3. EvalMLAlgorithm should be renamed "JeremyAlgorithm"
4. The use of the pipeline's new function in various places should be explored and determined whether it's necessary or not.  I'd recommend doing a commit that just pulls out the new calls and see what happens.  I feel like those are done because they're an ancient and hallowed tradition or we're uncertain whether we need yet another copy of the pipeline.
5. File an issue for creating another common baseclass with Iterative Algorithm.

But yea, this is really good work and super clean.  I can see why it took such a long time as it's polished.  Well done.

evalml/automl/automl_algorithm/evalml_algorithm.py
jeremyliweishih: all the scores coming in should have been [converted to "lower is better"](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L1129) so this order should be ok!

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -0,0 +1,365 @@
...
+        n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
+        ensemble = _make_stacked_ensemble_pipeline(
+            input_pipelines,
+            input_pipelines[0].problem_type,
+            random_seed=self.random_seed,
+            n_jobs=n_jobs_ensemble,
+        )
+        return [ensemble]
+
+    def _create_long_top_n(self, n):
```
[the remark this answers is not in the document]
jeremyliweishih: It's this step: 
```
Run top 3 estimators:
        a. Generate 50 random parameter sets. Run all 150 in one batch
 ```
How does `_create_long_exploration` sound? I see it as the exploration phase by diving deeper into the search space for the top 3 pipelines.

File: evalml/automl/automl_algorithm/evalml_algorithm.py [hunk 11]
```diff
@@ -0,0 +1,365 @@
...
+            for pipeline in pipelines:
+                if pipeline.name not in self._tuners:
+                    self._create_tuner(pipeline)
+                proposed_parameters = self._tuners[pipeline.name].propose()
+                parameters = self._transform_parameters(pipeline, proposed_parameters)
+                parameters.update(
+                    {"Select Columns Transformer": {"columns": self._selected_cols}}
+                )
+                next_batch.append(
+                    pipeline.new(parameters=parameters, random_seed=self.random_seed)
```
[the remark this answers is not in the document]
jeremyliweishih: it's not necessary in this case as @bchen1116 pointed out but it is when the we're creating new pipelines from a tuner proposed parameter set.

evalml/automl/automl_algorithm/evalml_algorithm.py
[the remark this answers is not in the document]
chukarsten: Sounds good.

chukarsten (approved):
LGTM!  Thanks for making the changes!

PR #2531, opened 2021-07-20
Add SelectDtypeColumns component

### LINKED ISSUE #2479: Create a new component which selects using feature types
We currently have the `SelectColumns` component which allows us to select specific columns based on column names. However, it would be useful to also have a component which filters / selects columns by (Woodwork logical) type. This could help us enable specific subgraphs / pipelines that handle certain data types (ex: select categorical cols --> pass to OHE).

eccabay:
Closes #2479 

I'm totally open to changing the name of the class, it was just the first thing I thought of


evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,52 @@ def transform(self, X, y=None):
...
+        cols = self.parameters.get("columns") or []
+
+        if len(cols) > 0 and not isinstance(cols[0], str):
+            column_types = [
+                logical_type.__class__ for logical_type in X.ww.logical_types.values()
+            ]
+        else:
+            column_types = [
+                logical_type.type_string for logical_type in X.ww.logical_types.values()
+            ]
```
eccabay, line 150: This logic is definitely not the cleanest it could be - if anyone knows a more efficient way to pull out this information, please let me know!

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,52 @@ def transform(self, X, y=None):
...
+        columns (list(string), list(ww.LogicalType)): List of datatypes, used to determine which columns to select.
+        random_seed (int): Seed for the random number generator. Defaults to 0.
+    """
+
+    name = "Select Datatype Columns Transformer"
+    hyperparameter_ranges = {}
+    needs_fitting = False
+
+    def _check_input_for_columns(self, X):
+        cols = self.parameters.get("columns") or []
```
freddyaboulton, line 141: What are your thoughts on making this a no-op? 

If we do that, then we get selection by semantic tag for free. I feel like that would be useful for workflows that don't use the "standard" set of types. For example, a user might be interested in selecting all numeric features (including Age and AgeNullable) or all categorical features (CountryCode and Ordinal). It would also make it easy for users to select based on their own criteria if they set the tags correctly. For example, users may be interested in selecting based on whether a feature was created by another component or not.

Another argument for doing that is that `check_input_for_columns` was implemented because `X.ww[cols]` would error if the columns were not in the dataframe. However, `Xww.select` does not error out if the resulting dataframe is empty. I think it's ok if this returns an empty dataframe - maybe we can just raise a warning instead?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,52 @@ def transform(self, X, y=None):
...
+                "Column(s) of type {} not found in input data".format(
+                    ", ".join(f"'{col_name}'" for col_name in missing_cols)
+                )
+            )
+
+    def _modify_columns(self, cols, X, y=None):
+        return X.ww.select(cols)
+
+    def transform(self, X, y=None):
+        """Transforms data X by dropping columns.
```
freddyaboulton, line 164: I think this got carried over from the `DropColumns` transformer!

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,52 @@ def transform(self, X, y=None):
             pd.DataFrame: Transformed X.
         """
         return super().transform(X, y)
+
+
+class SelectDtypeColumns(ColumnSelector):
```
freddyaboulton, line 127: Regarding the name, I feel like the right thing to do refactor `SelectColumns` to allow for logical type selection - that way there's only one `SelectColumns` transformer - but I don't think that's in scope for this issue.

If we follow my suggestion to also allow selecting by semantic tags, I think `SelectByTypeorTag` is a good alternate name. If not, I think `SelectByLogicalType` is also a good name! I don't like `Dtype` because  that's more pandas terminology and I feel it obscures the fact it's done via woodwork.

freddyaboulton (approved):
@eccabay Looks good! Thanks for using the existing `ColumnSelector` abstraction!

One comment I'd like to resolve before merge is that if we make `_check_input_for_columns` a no-op we get semantic tag selection for free! I think that'll make this more flexible.

evalml/pipelines/components/transformers/column_selectors.py
bchen1116: Agreed. I think this could definitely be a no-op, since it does allow us to select by semantic tags automatically.

Additionally,, currently it doesn't allow for selection if you use the capital string logical types (ie "NaturalLanguage" vs "natural_language")
image
I would think this should work since I can call `X.ww.select()` with both and get valid results

evalml/tests/component_tests/test_column_selector_transformers.py
```diff
@@ -168,3 +235,33 @@ def test_column_transformer_int_col_names_np_array(class_to_test, answers):
...
+
+    transformer = SelectDtypeColumns(columns=[ww.logical_types.Age])
+    with pytest.raises(ValueError, match=" not found in input data"):
+        transformer.fit(X)
+    with pytest.raises(ValueError, match=" not found in input data"):
+        transformer.transform(X)
+    with pytest.raises(ValueError, match=" not found in input data"):
+        transformer.fit_transform(X)
+
+    # X_t = SelectDtypeColumns(columns=[ww.logical_types.Integer]).fit_transform(X)
```
bchen1116, line 257: Is this supposed to remain commented?

evalml/tests/component_tests/test_column_selector_transformers.py
```diff
@@ -168,3 +235,33 @@ def test_column_transformer_int_col_names_np_array(class_to_test, answers):
 
     transformer = class_to_test(columns=[])
     assert_frame_equal(answer3, transformer.transform(X))
+
+
+def test_dtype_column_transformer_ww_types():
```
bchen1116, line 240: Can you add a test that selects based on capitalized types? ie `"NaturalLanguage"`, `"Double"`

bchen1116 (changes requested):
Looking good! Agreed with Freddy, might be nice to not limit it to selecting by logical types. Left some comments on additional tests that could be helpful, as well as some clean up

evalml/tests/component_tests/test_column_selector_transformers.py
eccabay: It is not, thanks for the catch!

ParthivNaresh (approved):
Looks great! After incorporating Freddy's and Bryan's changes I'd say this is solid. Nice job expanding the test to cover Double as well. I just left a minor comment regarding the docstring.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,52 @@ def transform(self, X, y=None):
...
+    """
+    Selects columns by specified datatype in input data.
+
+    Arguments:
+        columns (list(string), list(ww.LogicalType)): List of datatypes, used to determine which columns to select.
+        random_seed (int): Seed for the random number generator. Defaults to 0.
+    """
+
+    name = "Select Datatype Columns Transformer"
+    hyperparameter_ranges = {}
```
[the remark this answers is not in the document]
freddyaboulton: Great point @ParthivNaresh !!

bchen1116 (approved):
Nice! LGTM!

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
+        This check is not required, since ww.select does not throw an error if the requested types are not in the dataset.
+        """
+        return
+
+    def _modify_columns(self, cols, X, y=None):
+        columns = X.ww.select(cols)
+        if len(cols) > 0 and columns.empty:
+            warnings.warn(
+                "No columns of the selected type(s) were found in the input data. SelectByTypeorTag will return an empty DataFrame"
+            )
```
chukarsten, line 153: Do we want this behavior?  I would think that we might want to take the same route as the sibling classes and raise an exception if a bad selection is made.  @angela97lin what use cases are we trying to satisfy with this?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
             pd.DataFrame: Transformed X.
         """
         return super().transform(X, y)
+
+
+class SelectByTypeorTag(ColumnSelector):
```
chukarsten, line 128: Mega-nit: `SelectByTypeOrTag`?  The lowercase or was bugging me

chukarsten (approved):
Nice!  I'd personally be a fan of upper-casing the 'or' in the class name.  I'd also like to resolve whether we should error or warn if a selection is attempted on data without the expected type present.  For some reason, our components seem a little more intentional, rather than explorational, to me, so I would expect if this component is added to select a certain type, that that certain type is there.  Also, I feel like I've run into issues with our partial dependence code that warns when mixed nan/str features are detected, but our downstream consumers don't see the error and the trouble manifests in strange ways.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
+    """
+    Selects columns by specified Woodwork logical type or semantic tag in input data.
+
+    Arguments:
+        columns (list(string), list(ww.LogicalType)): List of datatypes, used to determine which columns to select.
+        random_seed (int): Seed for the random number generator. Defaults to 0.
+    """
+
+    name = "Select Columns By Type or Tag Transformer"
```
angela97lin, line 137: Just food for thought, but if we want to allow users to use this transformer via the name, this is a mighty long name that they'd have to not make a typo for... lolol

evalml/pipelines/components/transformers/column_selectors.py
dsherry: I think `SelectByType` would be slightly easier to read without losing much meaning. Thoughts?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
+        """Transforms data X by selecting columns.
+
+        Arguments:
+            X (pd.DataFrame): Data to transform.
+            y (pd.Series, optional): Targets.
+
+        Returns:
+            pd.DataFrame: Transformed X.
+        """
+        return super().transform(X, y)
```
dsherry, line 166: If all the `transform` method for this class is doing is calling `super`, please delete it!

evalml/pipelines/components/transformers/column_selectors.py
dsherry: Good thinking @eccabay and great point @chukarsten .

I think we should raise an exception here.

Example: graph filters down to categoricals only, then passes those to OHE and to estimator. If there are no categoricals in the dataset, OHE will be passed empty DF (and possibly error out), estimator will be passed empty DF (and possibly error out) and if for some reason there weren't errors in training the model will not have any value because it didn't receive any input features! This feels to me like grounds for a clear error message from this component.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
+    name = "Select Columns By Type or Tag Transformer"
+    hyperparameter_ranges = {}
+    """{}"""
+    needs_fitting = False
+
+    def _check_input_for_columns(self, X):
+        """
+        This check is not required, since ww.select does not throw an error if the requested types are not in the dataset.
+        """
+        return
```
dsherry, line 146: Why define this method as a no-op? Why not simply omit it?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
         """
         return super().transform(X, y)
+
+
+class SelectByTypeorTag(ColumnSelector):
+    """
+    Selects columns by specified Woodwork logical type or semantic tag in input data.
+
+    Arguments:
+        columns (list(string), list(ww.LogicalType)): List of datatypes, used to determine which columns to select.
```
dsherry, line 133: Don't we need an `__init__` method defined for this class in order to accept the types/tag(s) to use for the filter?

dsherry (changes requested):
@eccabay nice going, left a few comments.

Also, open question in case it hasn't already been discussed: should we define this functionality as a separate class (as is currently being done) or as simply a different option or way to use the existing `SelectColumns` transformer? I think defining a separate component makes sense, as long as there isn't too much code duplication.

evalml/pipelines/components/transformers/column_selectors.py
eccabay: The function is implemented in the parent class - I defined it here in order to overwrite the parent function.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +123,44 @@ def transform(self, X, y=None):
...
+    needs_fitting = False
+
+    def _check_input_for_columns(self, X):
+        """
+        This check is not required, since ww.select does not throw an error if the requested types are not in the dataset.
+        """
+        return
+
+    def _modify_columns(self, cols, X, y=None):
+        columns = X.ww.select(cols)
```
angela97lin, line 149: Nit pick on readability but `cols` vs `columns` as names is hard to differentiate; even just `selected_cols` could be easier to read?

evalml/pipelines/components/transformers/column_selectors.py
angela97lin: I like the idea of raising an exception or doing nothing at all:

Doing nothing at all: for the component independently, it can arguably be okay if the user wants to filter for a certain type and there happens to be no columns with that type. This is trickier with the component graph, though, as @dsherry pointed out, it is unclear what our other components do with an empty df. If our other components are able to handle empty dfs, I feel like it could be okay to do nothing: in the example above, we could try to have a separate branch for selecting categoricals for the OHE. If there are none, then the OHE doesn't do anything to the empty DF. That can be concatted with the non-categoricals and passed to the estimator. 

TLDR: safety net can be just to raise error but I feel like there's some validity in not doing anything at all /shrug

evalml/pipelines/components/transformers/column_selectors.py
angela97lin: Edit: I like @dsherry's suggestion of `SelectByType` :)

evalml/pipelines/components/transformers/column_selectors.py
eccabay: This implementation just piggybacks off of the same argument name as defined in the parent class, so there's no need for an explicit `__init__`. That being said, it may make sense to rename this argument to differentiate between column names and column types. Thoughts?

evalml/pipelines/components/transformers/column_selectors.py
eccabay: Does anyone know for certain that a downstream component would error out if passed an empty DataFrame? I ask since the default parameter for `SelectColumns` selects no columns and returns an empty DF (and we test to ensure this). Obviously, the situation is a little different if the user attempted to select a certain type and that failed, but I'm unclear how worried we should be about just passing empty DFs through pipelines.

evalml/pipelines/components/transformers/column_selectors.py
eccabay: Since the docstring mentions either logical type or semantic tag is possible, I'm fine renaming it to `SelectByType`!

evalml/pipelines/components/transformers/column_selectors.py
eccabay: Good point @angela97lin, I didn't think about that! Do you think just "Select By Type Transformer" would be concise but also informative enough?

evalml/pipelines/components/transformers/column_selectors.py
freddyaboulton: The downside is that we lose the docstring! But since this docstring isn't very specific to this class, that's ok I think

evalml/pipelines/components/transformers/column_selectors.py
freddyaboulton: Yea it would error out. Our components are not written to handle empty features so I think it's ok to raise an exception! My reasoning for raising a warning instead was that ww lets you select for types/tags that are not present and I figured that our component should not get in the way of that. I don't think we need to worry about passing empty features around. The pipelines created by AutoMLSearch don't do that so that would only happen if the user is doing something custom - we typically let users encounter failures in those cases. Making our components resilient to empty features could be an interesting project, but not one that we need to embark on now?

image

evalml/tests/component_tests/test_utils.py
```diff
@@ -29,13 +29,13 @@ def test_all_components(
     has_minimal_dependencies, is_running_py_39_or_above, is_using_conda
 ):
     if has_minimal_dependencies:
-        n_components = 37
+        n_components = 38
     elif is_using_conda:
         n_components = 48
```
ParthivNaresh, line 34: Putting a pin in this after our conversation during office hours 😅

evalml/pipelines/components/transformers/column_selectors.py
eccabay: Post office hours discussion, moving forward by throwing an error if the resulting DataFrame is empty. I've filed an issue to track further discussion/work in #2554

evalml/tests/component_tests/test_utils.py
ParthivNaresh: Issue filed #2555

evalml/pipelines/components/transformers/column_selectors.py
angela97lin: Yup, I think that's great 😁

ParthivNaresh:
@eccabay I triggered `build_conda_pkg` on the PR just to verify that it [fails](https://github.com/alteryx/evalml/pull/2531/checks?check_run_id=3164589089#step:5:2763) without the bump in `n_components`. I went ahead and added +1 to `is_using_conda` and `is_running_py_39_or_above`.

evalml/pipelines/components/transformers/column_selectors.py
dsherry: Got it, thanks for explaining.

I think the most pythonic thing to do here would be to add a second argument called "column_types" which is a list of types corresponding 1x1 positionally to the list of columns.

Overriding the type definition of an argument in the inheritance structure adds complexity. I don't think it'll cause problems in this particular case, since the inheritance structure is currently quite simple. But I do think using a separate argument would effectively future-proof this API against any unanticipated changes we may want to make to these components later, and is generally a useful pattern to follow for that reason.

evalml/pipelines/components/transformers/column_selectors.py
dsherry: I see that `ColumnSelector._check_input_for_columns` does the following:
```python
        cols = self.parameters.get("columns") or []

        column_names = X.columns

        missing_cols = set(cols) - set(column_names)
        if missing_cols:
            raise ValueError(
                "Columns {} not found in input data".format(
                    ", ".join(f"'{col_name}'" for col_name in missing_cols)
                )
            )
```
And that the above is called during `ColumnSelector.fit`. And, since `SelectByType` doesn't define a `fit` method, `SelectByType.fit` will implicitly call `ColumnSelector.fit`, which is good 👍 

So, is there a reason why we should *not* also do that during `SelectByType.fit`?

I propose
* Make the change I suggested about defining an init method for `SelectByType` and accepting `column_types` as a separate arg. So the `columns` arg will have the same structure in both `ColumnSelector` and `SelectByType` init
* Delete `SelectByType._check_input_for_columns`

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +122,30 @@ def transform(self, X, y=None):
...
+    def _check_input_for_columns(self, X):
+        """
+        This check is not required, since ww.select does not throw an error if the requested types are not in the dataset.
+        """
+        return
+
+    def _modify_columns(self, cols, X, y=None):
+        selected_columns = X.ww.select(cols)
+        if len(cols) > 0 and selected_columns.empty:
+            raise ValueError(f"Column(s) of type {cols} not found in input data")
```
dsherry, line 150: This is the same type of error we raise in `ColumnSelector._check_input_for_columns`. Do we need to define it in two places? I think the intent with the current layout of `ColumnSelector` methods was for `_check_input_for_columns` to do this sort of validation, and for `_modify_columns` to only include logic to make modifications (i.e. dropping columns for the `DropColumns` transformer)

I think we should define this error only in `_check_input_for_columns`.

If you want the error to include the column types for `SelectByType`, but omit the column types for `ColumnSelector` and other subclasses, I would suggest something like:
* Define a class attribute like `_VALIDATION_ERROR_FORMAT_STRING = "Columns {columns} not found in input data."`
* Override the value of `_VALIDATION_ERROR_FORMAT_STRING` in the `SelectByType` class definition to `"Columns {columns} of type {column_types} not found in input data"`
* Update `ColumnSelector._check_input_for_columns` to do
```python
raise ValueError(
                self._VALIDATION_ERROR_FORMAT_STRING.format(**self.parameters)
            )
```
* Delete `SelectByType._modify_columns`

Alternatively, we could simplify this, just delete `SelectColumns._modify_columns` and not have the error message say anything about types.

Make sense?

dsherry (approved):
@eccabay looks good! I left some specific suggestions on how to reduce the complexity of this code change even further. Not blocking merge, feel free to merge this PR and get another up to address, if you are interested in doing so.

dsherry:
@ParthivNaresh thanks for the heads up. Thankfully, since we made the codecov patch check optional, and that is the one which is going red on this PR, @eccabay doesn't need me to do the merge! The codecov project check is still green here.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +125,41 @@ def transform(self, X, y=None):
...
+            component_obj=None,
+            random_seed=random_seed,
+        )
+
+    def _modify_columns(self, cols, X, y=None):
+        return X.ww.select(cols)
+
+    def transform(self, X, y=None):
+        X = infer_feature_types(X)
+        self._check_input_for_columns(X)
```
angela97lin, line 162: Interesting... does `_check_input_for_columns` do anything here? Since we don't define it, I'm assuming it uses `ColumnSelector`'s, which has the line:
```
        cols = self.parameters.get("columns") or []
```
That returns an empty list for this component and therefore doesn't do any checks, right?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -122,3 +125,41 @@ def transform(self, X, y=None):
...
+    """{}"""
+    needs_fitting = False
+    _VALIDATION_ERROR_FORMAT_STRING = (
+        "Columns of type {column_types} not found in input data."
+    )
+
+    def __init__(self, column_types=None, random_seed=0, **kwargs):
+        parameters = {"column_types": column_types}
+        parameters.update(kwargs)
+        Transformer.__init__(
```
angela97lin, line 150: This, and the following comment, has me questioning if it's worth subclassing `ColumnSelector` rather than just `Transformer`; it seems like we're not gaining much value from `ColumnSelector` 😂

evalml/tests/component_tests/test_column_selector_transformers.py
```diff
@@ -31,7 +47,7 @@ def test_column_transformer_empty_X(class_to_test):
     assert_frame_equal(X, transformer.fit_transform(X))
 
     transformer = class_to_test(columns=["not in data"])
-    with pytest.raises(ValueError, match="'not in data' not found in input data"):
+    with pytest.raises(ValueError, match="not found in input data"):
```
angela97lin, line 50: This test doesn't actually test that a column type does not exist, but rather that a column does not exist! Since we're passing in the keyword `columns`, we trigger `self._check_input_for_columns(X)` to raise a ValueError, but don't we actually want to test the following:
```
transformer = SelectByType(column_types=["some fake column type"])
```
and see that the ValueError is raised? I think this is related to my previous comment about using `_check_input_for_columns` :)

angela97lin (commented):
Thanks for cleaning this up and taking in all of our comments, Becca! This looks great 😁 

I do have some final comments before this is merged regarding `_check_input_for_columns`--I'm not certain it's doing exactly what we want right now in SelectByType. It checks for column names via the `columns` parameter, but SelectByType checks for column types, so I'm not sure we can just use it as is?

evalml/pipelines/components/transformers/column_selectors.py
eccabay: It's true that that line will return an empty list and therefore the missing column name check will do nothing - that's by design here. This new bit of code should check if there are column types and if they match though:
```
col_types = self.parameters.get("column_types")
if col_types:
       missing_cols = X.ww.select(col_types).empty
```
I agree, it does feel a little weird to do this here, I'm thinking it may make more sense to move this to `check_input_for_columns` within `SelectByType` itself.

evalml/pipelines/components/transformers/column_selectors.py
eccabay: That's true, the way this is written there's not a lot of benefit to having `SelectByType` subclass this implementation of `ColumnSelector`. That being said, it still makes logical sense to have it be a subclass, since it does function to select columns. With that, I see this more as a reason to consider refactoring `ColumnSelector` rather than removing `SelectByType` as a subclass.

PR #2532, opened 2021-07-20
Address NaN values in Text Featurizer

### LINKED ISSUE #1587: Impute NaNs for natural language features
In #1492, we realized the `LSA` transformer errors when there are NaNs in one of the feature columns. Should the `LSA` transformer handle NaNs "gracefully" like the `nlp_primitives` used in the `TextFeaturizer` or should we update the `Imputer` to impute text columns?

This is the reproducer/original comment on #1492 that lead us to open this issue:

I think the root cause is that there are NaNs in the `job` column and this causes the `TfidfVectorizer` in LSA to crash. You can reproduce this locally:

```python
from sklearn.feature_extraction.text import TfidfVectorizer
import pandas as pd
import woodwork as ww
from urllib.request import urlopen
customers_data = urlopen('https://featurelabs-static.s3.amazonaws.com/lead_scoring_ml_apps/customers.csv')
interactions_data = urlopen('https://featurelabs-static.s3.amazonaws.com/lead_scoring_ml_apps/interactions.csv')
leads_data = urlopen('https://featurelabs-static.s3.amazonaws.com/lead_scoring_ml_apps/previous_leads.csv')
customers = pd.read_csv(customers_data)
interactions = pd.read_csv(interactions_data)
leads = pd.read_csv(leads_data)

X = customers.merge(interactions, on='customer_id').merge(leads, on='customer_id')
y = X['label']
X = X.drop(['customer_id', 'date_registered', 'birthday','phone', 'email',
            'owner', 'company', 'id', 'time_x',
            'session', 'referrer', 'time_y', 'label', 'country'], axis=1)
TfidfVectorizer().fit_transform(X['job'])
```

Dropping NaNs fixes the issue:
image

All of our AutoML pipelines have an Imputer. So I think the true bug is that the imputer does not impute NaNs for the `job` column!

image

This is because the Imputer will only impute [categorical columns, not text](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/imputers/imputer.py#L67). When we convert to pandas from ww, text columns get converted to `object`. 

The question now is "where should NaNs be handled for text columns"?. I think we have two options:
1. Update the Imputer to handle text columns.
2. Interestingly, the `LSA` transformer comes after the `nlp_primitives` are finished computing features. That makes me think that the `nlp_primitives` DFS naturally handle NaNs  whereas `LSA` does not. Should that feature be added to LSA @eccabay ? 
3. Both

What do you guys think?

_Originally posted by @freddyaboulton in https://github.com/alteryx/evalml/issues/1492#issuecomment-745414068_

bchen1116:
fix #1587


evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -126,16 +130,48 @@ def transform(self, X, y=None):
         if X_nlp_primitives.isnull().any().any():
             X_nlp_primitives.fillna(0, inplace=True)
 
-        X_lsa = self._lsa.transform(X_ww.ww[self._text_columns])
+        X_ww_altered = infer_feature_types(
+            X_ww.ww[self._text_columns].fillna(""),
+            {s: "NaturalLanguage" for s in self._text_columns},
```
bchen1116, line 135: We use this to keep the natural language type after filling in the nans with empty strings

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -126,16 +130,48 @@ def transform(self, X, y=None):
...
         X_nlp_primitives.set_index(X_ww.index, inplace=True)
 
         X_ww = X_ww.ww.drop(self._text_columns)
+        all_columns = np.append(X_nlp_primitives.columns.values, X_lsa.columns.values)
         for col in X_nlp_primitives:
             X_ww.ww[col] = X_nlp_primitives[col]
         for col in X_lsa:
             X_ww.ww[col] = X_lsa[col]
+
+        if X.isna().any().any():
```
bchen1116, line 147: We only convert empty to nans if there were nans in the original data

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -126,16 +130,48 @@ def transform(self, X, y=None):
         if X_nlp_primitives.isnull().any().any():
             X_nlp_primitives.fillna(0, inplace=True)
 
-        X_lsa = self._lsa.transform(X_ww.ww[self._text_columns])
+        X_ww_altered = infer_feature_types(
```
freddyaboulton, line 133: @bchen1116 I think we can simplify this implementation if we keep a mask of the rows that were NaN. I'm posting it here but down to sit down and talk through it together as well!

```python
        """Transforms data X by creating new features using existing text columns

        Arguments:
            X (pd.DataFrame): The data to transform.
            y (pd.Series, optional): Ignored.

        Returns:
            pd.DataFrame: Transformed X
        """
        X_ww = infer_feature_types(X)
        if self._features is None or len(self._features) == 0:
            return X_ww

        es = self._make_entity_set(X_ww, self._text_columns)
        nan_mask = X[self._text_columns].isna()
        any_nans = nan_mask.any().any()
        X_nlp_primitives = ft.calculate_feature_matrix(
            features=self._features, entityset=es
        )
        if X_nlp_primitives.isnull().any().any():
            X_nlp_primitives.fillna(0, inplace=True)

        if any_nans:
            primitive_features = self._get_primitives_provenance(self._features)
            for column, derived_features in primitive_features.items():
                X_nlp_primitives.loc[nan_mask[column], derived_features] = pd.NA

        X_ww_altered = infer_feature_types(
            X_ww.ww[self._text_columns].fillna(""),
            {s: "NaturalLanguage" for s in self._text_columns},
        )
        X_lsa = self._lsa.transform(X_ww_altered)
        if any_nans:
            lsa_features = self._lsa._get_feature_provenance()
            for column, derived_features in lsa_features.items():
                X_lsa.loc[nan_mask[column], derived_features] = pd.NA

        X_nlp_primitives.set_index(X_ww.index, inplace=True)

        X_ww = X_ww.ww.drop(self._text_columns)
        for col in X_nlp_primitives:
            X_ww.ww[col] = X_nlp_primitives[col]
        for col in X_lsa:
            X_ww.ww[col] = X_lsa[col]
        return X_ww
```

I think it's simpler because I think it's easier to understand what's happening (no regex, fewer lines of code) and it uses abstractions we already have in place (the feature provenance). It also doesn't require us to call `infer_feature_types` at the end - I'm afraid that may cause a bug down the line? At least that's what happened in https://github.com/alteryx/evalml/pull/2297 .

It may also be more performant? Haven't quantified this but with this implementation there's less space used (yes we keep a mask of NaNs but there are two calls to copy in `_empty_to_nan`) and the operation of replacing values in place is vectorized whereas calling `df.apply` is not.

evalml/tests/component_tests/test_text_featurizer.py
```diff
@@ -460,3 +460,41 @@ def test_text_featurizer_sets_max_depth_1(mock_dfs):
     tf.fit(X)
     _, kwargs = mock_dfs.call_args
     assert kwargs["max_depth"] == 1
+
+
+@pytest.mark.parametrize("nones", [np.nan, pd.NA, None])
+def test_nan_allowed(nones):
```
freddyaboulton, line 466: We should also add a test where there are nans in more than one column and in different indices

evalml/pipelines/utils.py
```diff
@@ -77,13 +77,13 @@ def _get_preprocessing_components(
...
     }
 
-    if len(input_logical_types.intersection(types_imputer_handles)) > 0:
-        pp_components.append(Imputer)
-
     text_columns = list(X.ww.select("NaturalLanguage", return_schema=True).columns)
     if len(text_columns) > 0:
         pp_components.append(TextFeaturizer)
 
+    if len(input_logical_types.intersection(types_imputer_handles)) > 0:
```
freddyaboulton, line 84: I think this should be `len(input_logical_types.intersection(types_imputer_handles)) or len(text_columns)`. If there are only text features we want to also add an imputer to impute any possible missing values.

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
chukarsten: @freddyaboulton I'm disappointed you didn't submit a PR to the PR.

freddyaboulton: 🙈

bchen1116: @freddyaboulton thanks for the suggestion! Yep, can definitely use this!

chukarsten (approved):
I think if you take the simplification per Freddy's request, I feel pretty good about this!  Nice work.

PR #2541, opened 2021-07-22

machineFL:

### DIFF
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -11,7 +11,7 @@ psutil>=5.6.6
 requirements-parser>=0.2.0
 shap>=0.36.0
 texttable>=1.6.2
-woodwork>=0.5.0
+woodwork==0.5.0
 dask>=2.12.0
 featuretools>=0.21.0
 nlp-primitives>=1.1.0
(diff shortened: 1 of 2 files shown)


freddyaboulton:
@alteryx/evalml Thoughts on pinning woodwork until we figure out how to support the new version? I think the change that's causing the issue is https://github.com/alteryx/woodwork/pull/1043

I think it's tricky to fix because: 
* Prior to 0.5.0, partial dependence would work on numeric columns with missing values. It would fill the entire column with NaNs but the physical type would still be float so the imputer could handle it. Now the physical type changes to string and the Imputer cannot handle missing values in string physical types. I think this one requires discussion?
*  Our data checks action for all null columns are broken cause they also pull in the "Unknown" data check message.
* The imputers can no longer handle all null columns for the same reason as why the partial dependence is broken

Seems a lot of changes are required it may make sense to prepare a design doc? Let me know what you think!

chukarsten (dismissed):

dsherry:
@freddyaboulton yes sounds good to me! Thanks for looking into it. You wanna file an issue to track upgrading to 0.5.1 and we can centralize discussion there?

freddyaboulton (approved):

PR #2544, opened 2021-07-23
Add gini coefficient as an objective

### LINKED ISSUE #2102: Add Gini coefficient as an objective
Mentioned during standup, the Gini coefficient would be a good metric to add to EvalML.

eccabay:
Closes #2102

### DIFF
--- a/evalml/tests/objective_tests/test_objectives.py
+++ b/evalml/tests/objective_tests/test_objectives.py
@@ -99,7 +99,7 @@ def test_can_get_only_core_and_all_objective_names():
 
 def test_get_core_objectives_types():
     assert len(get_core_objectives(ProblemTypes.MULTICLASS)) == 13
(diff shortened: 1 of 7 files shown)


evalml/tests/objective_tests/test_objectives.py
```diff
@@ -99,7 +99,7 @@ def test_can_get_only_core_and_all_objective_names():
```
freddyaboulton, line 99: Gini coefficient is also being tested in `test_objective_outputs` so I think it's in line with how we test our other objectives.

freddyaboulton (approved):
@eccabay Thanks for the write-up on the original issue! I agree with the conclusion/implementation proposed there!

angela97lin (approved):
Clean and simple, LGTM! 😁

PR #2546, opened 2021-07-23
Move `get_hyperparameter_ranges` to `PipelineBase` class from automl/utils module

angela97lin:
More breaking up #2490 into smaller and cleaner PRs :)

This PR moves `get_hyperparameter_ranges` from automl/utils to a method in `PipelineBase`. The method was only called in AutoMLAlgorithm, and was passed the pipeline's component graph anyway, so this isn't a big lift. Main reason for this was because it called `ComponentGraph`'s linearized_component_graph method, which I'm thinking about removing since it's only used in a list definition of the component graph.

### DIFF
--- a/docs/source/api_index.rst
+++ b/docs/source/api_index.rst
@@ -95,8 +95,8 @@ Pipeline Base Classes
 .. autoapisummary::
 
     evalml.pipelines.PipelineBase
-    evalml.pipelines.ClassificationPipeline
-    evalml.pipelines.BinaryClassificationPipeline
+    evalml.pipelines.classification_pipeline.ClassificationPipeline
+    evalml.pipelines.binary_classification_pipeline.BinaryClassificationPipeline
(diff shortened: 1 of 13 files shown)


docs/source/api_index.rst
```diff
@@ -95,8 +95,8 @@ Pipeline Base Classes
 .. autoapisummary::
 
     evalml.pipelines.PipelineBase
-    evalml.pipelines.ClassificationPipeline
-    evalml.pipelines.BinaryClassificationPipeline
+    evalml.pipelines.classification_pipeline.ClassificationPipeline
```
angela97lin, line 98: Docs break without this update (see [here](https://readthedocs.com/projects/feature-labs-inc-evalml/builds/683782/)); filed an issue to track this long term as #2548

freddyaboulton (approved):

chukarsten (approved):
LGTM!

PR #2549, opened 2021-07-23
Remove `ComponentGraph.from_list` and update PipelineBase impl creating from lists

angela97lin:
More breaking up #2490 into smaller and cleaner PRs :)

This PR removes `from_list` from ComponentGraph and updates `PipelineBase`'s implementation to handle creating a component dictionary to pass along to the ComponentGraph.

This is slightly different from our behavior today because it puts the burden of converting from a list of components --> component dict on the pipeline. This means that ComponentGraphs do not need to understand how to handle lists.

Note that `_make_component_dict_from_component_list` updates the generated component graph to include X and y edges for each component, but it is not enforced yet! That will come in a later PR :)

PR #2556, opened 2021-07-26
Remove `linearized_component_graph` and `ComponentGraph.from_list` 

angela97lin:
#2549 removed almost all usages of `ComponentGraph.from_list` but didn't actually remove the method 😂 

This PR does that, and removes the idea of `linearized_component_graph`. I think this is a good idea because this method assumes that there is a way to make something "linear", which is suspicious in DAG cases. In addition, most places where our code currently uses this can be easily replaced with `component_instances`, which makes me suspicious that order did not really matter; instead, `linearized_component_graph` was used as a convenient way to get component name mapping to the component class. If this is useful, it should be revamped into a simple dictionary, but for now, functionality is covered by `component_instances`.


docs/source/release_notes.rst
```diff
@@ -4,18 +4,19 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
-        * Removed ``ComponentGraph.from_list`` and update PipelineBase implementation for creating pipelines from a list of components :pr:`2549`
+        * Updated ``PipelineBase`` implementation for creating pipelines from a list of components :pr:`2549`
```
angela97lin, line 7: Updating release notes to make clear that this is actually the PR that removed `from_list` :)

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -417,7 +367,7 @@ def test_get_estimators(example_graph):
         LogisticRegressionClassifier(),
     ]
 
-    component_graph = ComponentGraph.from_list(["Imputer", "One Hot Encoder"])
+    component_graph = ComponentGraph({"Imputer": ["Imputer", "X", "y"]})
```
angela97lin, line 370: I thought this test still had validity: it was testing that `get_estimators` when a component graph has no estimators will return an empty list. I'm just changing how the graph is initialized :)

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1513,16 +1385,25 @@ def test_component_graph_sampler_y_passes(mock_estimator_fit):
...
     X = pd.DataFrame({"a": [i for i in range(100)], "b": [i % 3 for i in range(100)]})
     y = pd.Series([0] * 90 + [1] * 10)
-    component_list = [
-        "Imputer",
-        "SMOTE Oversampler",
-        "Standard Scaler",
-        "Logistic Regression Classifier",
-    ]
-    component_graph = ComponentGraph.from_list(component_list)
+    component_graph = {
```
angela97lin, line 1391: Again, I thought this could be valuable to keep around even though the original test (does y value get passed even though edge from standard scaler is unclear?) is different; still valuable to double check that we are indeed getting the oversampled y :)

ParthivNaresh: Just a tad confused by `does y value get passed even though edge from standard scaler is unclear?`, we're not expecting it to be passed right?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1531,31 +1412,6 @@ def test_component_graph_sampler_y_passes(mock_estimator_fit):
     assert len(mock_estimator_fit.call_args[0][0]) == int(1.25 * 90)
 
 
-@patch("evalml.pipelines.components.estimators.RandomForestClassifier.fit")
-@patch("evalml.pipelines.components.estimators.DecisionTreeClassifier.fit")
-def test_component_graph_sampler_same_given_components(mock_dt_fit, mock_rf_fit):
```
ParthivNaresh, line 1536: I'm not sure what the original purpose of this test was, just verifying that `SMOTE Oversampler` was included in both graphs?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1921,114 +1777,6 @@ def test_component_graph_inverse_transform(
     pd.testing.assert_series_equal(answer, expected)
 
 
-lists = [
```
ParthivNaresh, line 1924: Goodnight dear lists

ParthivNaresh (approved):
Looks great!

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Before this PR, we expect that the oversampler's y value gets passed to the estimator because standard scaler doesn't produce a y value so the "most recent y" value the estimator sees is still the oversampler's y, but this isn't really clear using the list API.

Now that we're specifying edges, it's more clear that the estimator gets the oversampler's y edge, which would have meant that this test isn't necessary to confirm that anymore, but I wanted to keep it around anyways in the case that the estimator does not get the oversampler's y :)

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: I think it was just testing that with different estimators that the oversampler was the same and received the same input :)

PR #2562, opened 2021-07-27
Separate thresholds for pct null rows and columns in `HighlyNullDataCheck`

### LINKED ISSUE #2270: Highly null data check: separate thresholds for pct null cols vs rows
**Background**
In #2222 @jeremyliweishih updated the highly null data check to warn if individual rows exceed a threshold percentage of nulls. Currently that threshold is the same as the one used for checking columns in the same manner.

**Proposal**
Let's have separate thresholds for pct null rows vs pct null columns. That'll make it easy to tinker with the behavior on real data, in case we decide we want different mechanics.

eccabay:
Closes #2270


docs/source/release_notes.rst
```diff
@@ -3,6 +3,7 @@ Release Notes
 **Future Release**
     * Enhancements
         * Added Gini coefficient as an objective :pr:`2544`
+        * Added separate thresholds for pct null rows and columns in ``HighlyNullDataCheck`` :pr:`2562`
```
bchen1116, line 6: nit: pct -> percent

evalml/data_checks/highly_null_data_check.py
```diff
@@ -12,17 +12,25 @@ class HighlyNullDataCheck(DataCheck):
...
+    def __init__(self, pct_null_col_threshold=0.95, pct_null_row_threshold=0.95):
+        if (
+            pct_null_col_threshold < 0
+            or pct_null_col_threshold > 1
+            or pct_null_row_threshold < 0
+            or pct_null_row_threshold > 1
+        ):
             raise ValueError(
-                "pct_null_threshold must be a float between 0 and 1, inclusive."
+                "pct null thresholds must be a float between 0 and 1, inclusive."
```
bchen1116, line 30: I think it'd be nice to separate the checks for cols vs rows to raise error specifically on the value needed

bchen1116 (approved):
Looks good! I left a nit and a comment about potentially separating the check for valid thresholds on rows vs columns, but nothing blocking!

evalml/data_checks/highly_null_data_check.py
```diff
@@ -12,17 +12,25 @@ class HighlyNullDataCheck(DataCheck):
     """Checks if there are any highly-null columns and rows in the input.
 
     Arguments:
-        pct_null_threshold(float): If the percentage of NaN values in an input feature exceeds this amount,
-            that column/row will be considered highly-null. Defaults to 0.95.
+        pct_null_col_threshold(float): If the percentage of NaN values in an input feature exceeds this amount,
+            that column will be considered highly-null. Defaults to 0.95.
+        pct_null_row_threshold(float): If the percentage of NaN values in an input feature exceeds this amount,
```
angela97lin, line 17: Docstring nitpick: If the percentage of NaN values in an input row (not feature, which refers to col)

angela97lin (approved):
Left a nitpicky comment but LGTM, thanks for this! 😁

evalml/tests/data_checks_tests/test_highly_null_data_check.py
```diff
@@ -155,8 +177,84 @@ def test_highly_null_data_check_warnings():
...
+            DataCheckAction(
+                DataCheckActionCode.DROP_COL, metadata={"column": "all_null"}
+            ).to_dict(),
+        ],
+    }
+
+    col_null_check = HighlyNullDataCheck(
+        pct_null_col_threshold=0.0, pct_null_row_threshold=0.9
+    )
+    highly_null_rows = SeriesWrap(pd.Series([2 / 3, 2 / 3, 2 / 3, 2 / 3, 1 / 3]))
```
chukarsten, line 225: I feel like I'm missing something, where is `highly_null_rows` used?

chukarsten: yea i think this is pasta

evalml/tests/data_checks_tests/test_highly_null_data_check.py
```diff
@@ -155,8 +177,84 @@ def test_highly_null_data_check_warnings():
     }
 
 
+def test_highly_null_data_check_separate_rows_cols():
```
chukarsten, line 180: Perhaps we want to adopt a similar testing strategy where we create no-null rows, all null rows and some null rows and just test that?  It's not immediately clear what this test is setting out to accomplish and it looks a lot like the other tests.

chukarsten (changes requested):
Awesome, this looks like a solid extension of the highlynulldatacheck.  I think we need to clean up the added test a little bit.  I think it carries over some copy pasta.  We might want to consider a `test_null_rows` that just goes through and tests the rows similarly to the columns, with the all null, some null, all full and shifting the threshold around to play with that.

evalml/tests/data_checks_tests/test_highly_null_data_check.py
eccabay: Could you expand more about what you're thinking of? The previous test checks both null rows and columns at thresholds of 1, 0.5, and 0. This added test just checks that having different values for the row vs column thresholds will throw/not throw warnings correctly, I'm not quite sure why we'd need to add more explicit null row coverage.

evalml/tests/data_checks_tests/test_highly_null_data_check.py
eccabay: Great catch, thank you 😅

evalml/tests/data_checks_tests/test_highly_null_data_check.py
chukarsten: Ok, I didn't understand what the purpose of this specific test was.  Good to go.

chukarsten (approved):

PR #2563, opened 2021-07-27
Update ComponentGraph to enforce needing .x and .y for each component specified in the graph

### LINKED ISSUE #2481: Update ComponentGraph to enforce needing .x and .y for each component specified in the graph

angela97lin:
Closes #2481, based on #2490 :)

The output of estimators (and all other components) must be accessed via `.x` or `.y`; simply using the component name will no longer suffice. The convention to defaulting to `.x` if not specified was chosen when we did not have many (if any) components that outputted an appropriate .y value. Now that we do have more target transformers, I think enforcing explicit edges is helpful in clarity.

This PR also deletes some fixtures and does some cleanup for our tests.


evalml/pipelines/component_graph.py
```diff
@@ -271,7 +286,7 @@ def _compute_features(self, component_list, X, y=None, fit=False):
                     output = component_instance.predict(input_x)
                 else:
                     output = None
-                output_cache[component_name] = output
+                output_cache[f"{component_name}.x"] = output
```
angela97lin, line 289: This means that components will use estimator outputs as `estimator_name.x`, not `estimator_name` to align with how users need to specify :)

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -576,36 +583,7 @@ def test_multi_format_creation(X_y_binary):
     assert not clf.feature_importance.isnull().all().all()
 
 
-def test_multiple_feature_selectors(X_y_binary):
```
angela97lin, line 579: I'm not really sure what exactly this was testing... just that we could have multiple feature selectors defined in a component graph? I deleted because I'm not sure how this differs from simply testing that we have duplicate components in a pipeline, but feel free to point out something I'm missing!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -384,19 +383,9 @@ def test_describe_pipeline(
         assert component.name in out
 
 
-def test_nonlinear_model_family():
+def test_nonlinear_model_family(example_graph):
```
angela97lin, line 386: Updated this test to just use our example graph, rather than defining a similar component graph below. As such, new expected model family should be linear, not RF as before :)

evalml/tests/conftest.py
```diff
@@ -403,14 +390,35 @@ def dummy_classifier_dict_component_graph(dummy_classifier_estimator_class):
...
+                dummy_regressor_estimator_class,
+                "Imputer.x",
+                "y",
+            ],
         }
     }
     return component_graph_dict
 
 
+@pytest.fixture
```
angela97lin, line 404: Moved from `test_component_graph.py` to be used here, too!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1064,7 +1063,6 @@ def test_add_to_rankings(
         y_train=y,
         problem_type="binary",
         max_iterations=1,
-        allowed_component_graphs=dummy_classifier_linear_component_graph,
```
angela97lin, line 1067: `allowed_component_graphs` plays no role in the validity of this test, so removing :)

evalml/tests/conftest.py
```diff
@@ -372,43 +372,38 @@ def fit(self, X, y):
...
-@pytest.fixture
-def dummy_regressor_linear_component_graph(dummy_regressor_estimator_class):
-    component_graph_linear = {
-        "Name": ["Imputer", "One Hot Encoder", dummy_regressor_estimator_class]
-    }
-    return component_graph_linear
-
-
-@pytest.fixture
-def dummy_classifier_dict_component_graph(dummy_classifier_estimator_class):
```
angela97lin, line 391: Removed these in favor of `example_graph` and `example_regression_graph` since they were confusing; they used dummy classifier but were labeled as Random Forest. Didn't provide much, if any value over the example_graph fixture

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -91,23 +91,6 @@ def dummy_components():
     return TransformerA, TransformerB, TransformerC, EstimatorA, EstimatorB, EstimatorC
 
 
-@pytest.fixture
```
angela97lin, line 94: Moved to conftest so that it could be used in other files

evalml/pipelines/component_graph.py
```diff
@@ -514,8 +531,8 @@ def generate_order(cls, component_dict):
             return []
         digraph = nx.DiGraph()
         digraph.add_edges_from(edges)
-        if not nx.is_weakly_connected(digraph):
```
angela97lin, line 517: I don't think this is necessary anymore given that edges have to be defined..? But open to discussion in case I've missed anything!

angela97lin:
@chukarsten I know you keep hoping for a large PR but hopefully, after splitting out the original PR, this is as big as it gets 😂

evalml/pipelines/component_graph.py
```diff
@@ -227,6 +244,9 @@ def _compute_features(self, component_list, X, y=None, fit=False):
             y_input = None
 
             for parent_input in self.get_inputs(component_name):
+                import pdb
+
+                pdb.set_trace()
```
chukarsten: NoooooooOOOooooooOOoooo

evalml/tests/automl_tests/test_automl.py
```diff
@@ -3744,99 +3748,81 @@ def test_automl_respects_pipeline_parameters_with_duplicate_components(
...
-        component_graph = {
-            "Name_dict": {
-                "Imputer": ["Imputer"],
-                "Imputer_1": ["Imputer", "Imputer"],
-                "Random Forest Classifier": ["Random Forest Classifier", "Imputer_1"],
-            }
+    custom_hyperparameters = {
+        "Imputer": {"numeric_impute_strategy": Categorical(["most_frequent", "mean"])},
+        "Imputer_1": {"numeric_impute_strategy": Categorical(["median", "mean"])},
+        "Random Forest Classifier": {"n_estimators": Categorical([50, 100])},
```
chukarsten, line 3813: How'd you get this past black without it exploding into 20+ lines?

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1008,20 +1005,11 @@ def test_automl_component_graphs_search(
...
     else:
         X, y = X_y_multi
         score_return_value = {"Log Loss Multiclass": 1.0}
         expected_mock_class = MulticlassClassificationPipeline
-        component_graph = (
-            dummy_classifier_linear_component_graph
-            if is_linear
-            else dummy_classifier_dict_component_graph
-        )
+    component_graph = {"CG": example_graph}
```
chukarsten, line 1012: nice

chukarsten (approved):
Awesome, this is a really great PR and I really appreciate your cleanup as you go, philosophy.  Looks really good.  Thank you.

evalml/pipelines/component_graph.py
freddyaboulton: We may have introduced a bug somewhere? I would expect the following graph to raise this error on `main` but it doesn't:

```python

cg = ComponentGraph({"Imputer": ["Imputer", "X", "y"],
                     "DateTime": ["DateTime Featurization Component", "X", "y"],
                     "Logistic Regression Classifier": ["Logistic Regression Classifier", "DateTime.x", "DateTime.y"]})
```

It's not completely connected!

image

evalml/tests/automl_tests/test_automl.py
angela97lin: LOL I just double checked: my line is 88 chars long, and black's limit is 88 chars per line--so I just bareeeely made it 😅

evalml/pipelines/component_graph.py
angela97lin: Yes, I think you're right--gonna look into this, thank you!

angela97lin: @freddyaboulton Added the nodes to the DiGraph that we're using to determine connectivity so that if a node does not have an edge, it raises a `ValueError`, and added back the test in `test_init_bad_graphs`. Thanks for your sharp eyes 👀

freddyaboulton: Thanks for digging into this!

PR #2564, opened 2021-07-27
Warn users when pipeline parameters aren't used when initializing pipelines

### LINKED ISSUE #1839: Pipeline Parameters: Warn/Raise on Extra Parameters
When parameters are passed in as a dictionary to instantiate a pipeline, the pipeline takes what it needs and passes the parameters up to the __init__ of it's parent class.  The parent class then takes that dictionary of parameters and iterates through the components the user has placed in the pipeline, pulling the parameters for the component that match the key in the parameters dictionary.  

I've run into one case that I don't think exhibits proper behavior: if the supplied parameters contain excess parameters for components that are not in the pipeline.

Repro:
```

class TSPipeline(TimeSeriesRegressionPipeline):
    component_graph = ["Delayed Feature Transformer", "Random Forest Regressor"]
    name = "time series pipeline"

tspipeline = TSPipeline({"pipeline": {"gap": 1, "max_delay": 2},
                             "Linear Regressor": {"n_jobs": 1}})
```

Options for desired behavior:

1. Continue to do nothing.  Sub-optimal as a user has tried to set parameters on something and the default parameters will be used instead.  This could lead to very tough to catch misbehavior for unaware users (like me).
2. Communicate a warning indicating that the supplied parameters will be ignored and the pointing out which components will be instantiated with defaults.
3. (Preferred) Raise a ValueError indicating that parameters were provided for non-existent components.

A good place to start looking is in evalml/pipeplines/component_graph.py::ComponentGraph.instantiate().  A simple comparison of keys to the component graph's components will probably lead to 2 or 3 being complete.

bchen1116:
fix #1839 

I opted to use warnings rather than raise errors since pipeline initialization occurs during AutoMLSearch as well. In the case where users pass in a valid `pipeline_param`, we don't want to error it out since we pass the entirety of `pipeline_parameters` to every pipeline we initialize. 
```python
automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary', 
                      pipeline_parameters={"CatBoost Classifier": {'n_estimators': 20}}), max_batches=2)
```
It would also be too much logic to include in `AutoMLSearch` to filter out only the appropriate components we need from pipeline parameters into the pipelines we are initializing. (ie determining which components are necessary in which pipelines before `make_pipelines` so we can trim and pass the appropriate pipeline parameters)

In this scenario, we don't want to raise an error since that would prevent `AutoMLSearch` from happening. 

However, raising a warning allows the search to continue while also giving the user feedback on whether or not the component parameters are being used.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,6 +1,5 @@
 import copy
 import pickle
-import re
 import sys
 import time
 import traceback
@@ -28,6 +27,7 @@ from evalml.automl.utils import (
 from evalml.data_checks import DefaultDataChecks
 from evalml.exceptions import (
     AutoMLSearchException,
+    ParameterNotUsedWarning,
     PipelineNotFoundError,
     PipelineScoreError,
 )
@@ -508,7 +508,7 @@ class AutoMLSearch:
                     )
 
             with warnings.catch_warnings(record=True) as w:
-                warnings.filterwarnings("always", message="Parameters for components")
+                warnings.filterwarnings("always", category=ParameterNotUsedWarning)
                 self.allowed_pipelines = [
                     make_pipeline(
                         self.X_train,
@@ -524,7 +524,7 @@ class AutoMLSearch:
         else:
 
             with warnings.catch_warnings(record=True) as w:
-                warnings.filterwarnings("always", message="Parameters for components")
+                warnings.filterwarnings("always", category=ParameterNotUsedWarning)
                 self.allowed_pipelines = get_pipelines_from_component_graphs(
                     self.allowed_component_graphs,
                     self.problem_type,
@@ -639,23 +639,13 @@ class AutoMLSearch:
     def _catch_warnings(self, warning_list):
         if len(warning_list) == len(self.allowed_pipelines) and len(warning_list) > 0:
             # we find the value(s) that we must throw the warning for
-            final_message = ""
+            final_message = set([])
             for idx, msg in enumerate(warning_list):
                 if idx == 0:
-                    val = re.search(r"\{(.*?)\}", str(msg.message))
-                    params = set(val.group(1).split(", "))
-                    final_message = str(msg.message)
+                    final_message = final_message.union(msg.message.components)
                 else:
-                    params = params.intersection(
-                        set(
-                            re.search(r"\{(.*?)\}", str(msg.message))
-                            .group(1)
-                            .split(", ")
-                        )
-                    )
-            warnings.warn(
-                final_message.replace(val.group(0), str(params).replace('"', ""))
-            )
+                    final_message = final_message.intersection(msg.message.components)
+            warnings.warn(ParameterNotUsedWarning(final_message))
 
     def _get_batch_number(self):
         batch_number = 1
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1,5 +1,6 @@
 import copy
 import pickle
+import re
 import sys
 import time
 import traceback
@@ -466,11 +467,11 @@ class AutoMLSearch:
                     self.sampler_method,
                     self.sampler_balanced_ratio,
                 )
-            if self._sampler_name not in parameters:
+            if self._sampler_name not in parameters and self._sampler_name is not None:
                 parameters[self._sampler_name] = {
                     "sampling_ratio": self.sampler_balanced_ratio
                 }
-            else:
+            elif self._sampler_name is not None:
                 parameters[self._sampler_name].update(
                     {"sampling_ratio": self.sampler_balanced_ratio}
                 )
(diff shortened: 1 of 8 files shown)


evalml/automl/automl_search.py
```diff
@@ -466,11 +467,11 @@ def __init__(
                     self.sampler_method,
                     self.sampler_balanced_ratio,
                 )
-            if self._sampler_name not in parameters:
+            if self._sampler_name not in parameters and self._sampler_name is not None:
```
bchen1116, line 470: We use this to make sure we don't add the sampling ratio to the param dictionary if the `sampler_name` is None

evalml/automl/automl_search.py
```diff
@@ -627,6 +636,27 @@ def __init__(
             custom_hyperparameters=custom_hyperparameters,
         )
 
+    def _catch_warnings(self, warning_list):
```
bchen1116, line 639: Allows us to catch all warnings thrown during pipeline initialization and only throw the warnings for the component names that appear in all warnings. For instance, if we had pipelines:
```python
p1 = ['Imputer', 'Standard Scaler', 'Logistic Regression Classifier']
p2 = ['Imputer', 'Random Forest Classifier']
params_passed_in = {"Logistic Regression Classifier": {"some value": 2}, 
                   {"Undersampler": {"another val": 1}}
```
then we would only want to raise warning for `Undersampler` since it isn't used in any of the pipelines

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1881,3 +1882,48 @@ def test_component_graph_with_X_y_inputs_y(mock_fit, mock_fit_transform):
     assert_series_equal(mock_fit_transform.call_args[0][1], y)
     # Check that we use "Log.y" for RF
     assert_series_equal(mock_fit.call_args[0][1], infer_feature_types(np.log(y)))
+
+
+@pytest.mark.parametrize(
+    "pipeline_parameters,set_values",
+    [
+        ({"Logistic Regression Classifier": {"penalty": "l1"}}, {}),
+        ({"Logistic Regression": {"penalty": "l1"}}, {"Logistic Regression"}),
```
bchen1116, line 1891: This should issue a warning since the component name we've defined in the graph is `Logistic Regression Classifier` rather than `Logistic Regression`

evalml/automl/automl_search.py
```diff
@@ -627,6 +636,27 @@ def __init__(
...
+        if len(warning_list) == len(self.allowed_pipelines) and len(warning_list) > 0:
```
freddyaboulton, line 640: @bchen1116 I think we can simplify this implementation a bit if we raise our own custom warning that subclasses `UserWarning`. 

```python
class ParameterNotUsedWarning(UserWarning):

    def __init__(self, components):
        self.components = components

        msg = f"Parameters for components {components} will not be used to instantiate the pipeline since they don't appear in the pipeline"
        super().__init__(msg)
```

ComponentGraph now becomes:

```python
        diff = param_set.difference(set(self.component_instances.keys()))
        if len(diff):
            warnings.warn(ParameterNotUsedWarning(diff))
```

_catch_warnings now becomes:

```python
    def _catch_warnings(self, warning_list):
        if len(warning_list) == len(self.allowed_pipelines) and len(warning_list) > 0:
            # we find the value(s) that we must throw the warning for
            final_message = set([])
            for idx, msg in enumerate(warning_list):
                if idx == 0:
                    final_message = final_message.union(msg.message.components)
                else:
                    final_message.intersection(msg.message.components)
            warnings.warn(ParameterNotUsedWarning(list(final_message)))
```

I think this is cleaner because it defines an api for accessing the components we want to warn about rather than messing with the string. I worry that'll be harder to maintain if we ever decide to change the message. I think the implementation is easier to understand as well. We can also explicitly filter for our custom warning in `AutoMLSearch` which seems cleaner too.

Something like: 
```python
            with warnings.catch_warnings(record=True) as w:
                warnings.filterwarnings("always", category=ParameterNotUsedWarning)
```

Let me know what you think!

evalml/automl/automl_search.py
freddyaboulton: Was this a bug?

evalml/automl/automl_search.py
bchen1116: This is a good idea! Yeah, I can change the impl here

bchen1116: @freddyaboulton added the suggestion you raised. Thanks!

evalml/exceptions/exceptions.py
```diff
@@ -87,3 +87,11 @@ class ObjectiveCreationError(Exception):
 
 class NoPositiveLabelException(Exception):
     """Exception when a particular classification label for the 'positive' class cannot be found in the column index or unique values"""
+
+
+class ParameterNotUsedWarning(UserWarning):
```
freddyaboulton, line 92: It shows up on the documentation automatically! https://feature-labs-inc-evalml--2564.com.readthedocs.build/en/2564/autoapi/evalml/exceptions/index.html#evalml.exceptions.ParameterNotUsedWarning

image

evalml/automl/automl_search.py
freddyaboulton: Thanks @bchen1116 I will take a look

evalml/automl/automl_search.py
freddyaboulton: Tbh when I first read the issue I didn't foresee we would need this in AutoMLSearch but I think it's the right call!

freddyaboulton (approved):
@bchen1116 This looks great to me!

chukarsten (approved):
Awesome, Bryan.  I am sorry this scope creeded up so much.  Thanks for being flexible and talking me through it.

evalml/automl/automl_search.py
bchen1116: @freddyaboulton I think so. It wasn't doing anything except adding `{None: {sampler_ratio: 0.25}}` to the param dictionary, which doesn't do anything but also isn't necessary and could cause problems in the future (like with this PR). It should be fixed now!

evalml/automl/automl_search.py
```diff
@@ -505,24 +506,32 @@ def __init__(
...
+                        self.y_train,
+                        estimator,
+                        self.problem_type,
+                        parameters=parameters,
+                        sampler_name=self._sampler_name,
+                    )
+                    for estimator in allowed_estimators
+                ]
+            self._catch_warnings(w)
+        else:
```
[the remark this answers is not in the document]
bchen1116: yep!

PR #2565, opened 2021-07-27
Add repr to ComponentGraph and use in PipelineBase

### LINKED ISSUE #2204: Printing component graph fails
Trying to print the component graph for pipelines in AutoMLSearch fails. This probably is related to the `repr` method for the components generated through `make_pipelines`, since we don't see the same behavior when calling a component graph in regular pipelines. This is run with Jupyter Lab

image

image

angela97lin:
Closes #2204 by adding `repr` to the ComponentGraph class and then cleaning up `PipelineBase`'s `repr` to reuse `ComponentGraph`'s `repr`.

On main:
image

After this PR:
image

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,6 +3,7 @@ Release Notes
 **Future Release**
     * Enhancements
         * Added Gini coefficient as an objective :pr:`2544`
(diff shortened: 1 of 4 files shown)


chukarsten (approved):
Awesome sauce.  I always get sad when the built ins aren't defined.

PR #2572, opened 2021-07-29
Refactor and fix make_pipeline tests to assert correct order and address new Woodwork Unknown type

### LINKED ISSUE #2553: Our make_pipeline tests don't actually verify the pipeline is correct!
I just noticed that `test_make_pipeline_*` doesn't actually assert the pipeline is correct! These tests could fail and we wouldn't know. Let's add the `assert` back in.

https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L149
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L206
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L261
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L309
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L370
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L438
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L492
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L535
https://github.com/alteryx/evalml/blob/main/evalml/tests/pipeline_tests/test_pipeline_utils.py#L591

angela97lin:
Closes #2553. 

- Tests needed to be updated to handle Woodwork's defaulting to "Unknown" rather than "NaturalLanguage". This is the case when our input is pandas type, since calling `infer_feature_type` will convert text columns to Unknown type, which returns a different pipeline.
- Pipeline tests needed to address the new `LogTransformer` component that could be added to pipeline
- Complete refactor of if statements and tests: many of the `make_pipeline` tests had similar logic, but different data. Having to update these tests for the updates in `make_pipeline` logic meant I had to update it in many places. It may be hard to tell from the diff, but I refactored all the tests into one parameterized test and made a fixture for the data to use. Although having one large parameterized method may seem scary, it's definitely easier to add new test cases via new columns in the df. I primarily kept all of the existing tests as is with some touchups but open to changing / removing :) 
- Removed `Time Series Baseline Estimator` in parameters dict passed. Thanks to @bchen1116's great work, a warning was raised indicating that the parameters for this component weren't used since it's not in the pipeline! 😁


evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -212,74 +227,19 @@ def test_make_pipeline_no_nulls(input_type, problem_type):
...
                 )
-            pipeline.component_graph.compute_order == [
+            assert pipeline.component_graph.compute_order == [
                 component.name for component in expected_components
             ]
 
 
 @pytest.mark.parametrize("input_type", ["pd", "ww"])
 @pytest.mark.parametrize("problem_type", ProblemTypes.all_problem_types)
-def test_make_pipeline_no_datetimes(input_type, problem_type):
```
angela97lin, line 222: I didn't see value in this compared to `test_make_pipeline_text_columns` so deleting :)

evalml/pipelines/utils.py
```diff
@@ -181,7 +180,7 @@ def make_pipeline(
     problem_type,
     parameters=None,
     sampler_name=None,
-    extra_components=[],
+    extra_components=None,
```
angela97lin, line 183: I think it'd be a good idea to change this default value to None, rather than an empty list. Right now, we don't modify this parameter but if we did, it could lead to unexpected behavior:

From https://docs.python.org/3/reference/compound_stmts.html#function-definitions:

Default parameter values are evaluated from left to right when the function definition is executed. This means that the expression is evaluated once, when the function is defined, and that the same “pre-computed” value is used for each call. This is especially important to understand when a default parameter is a mutable object, such as a list or a dictionary: if the function modifies the object (e.g. by appending an item to a list), the default value is in effect modified. This is generally not what was intended. A way around this is to use None as the default, and explicitly test for it in the body of the function...

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -528,76 +472,75 @@ def test_make_pipeline_numpy_input(problem_type):
...
             pipeline = make_pipeline(X, y, estimator_class, problem_type, parameters)
             assert isinstance(pipeline, pipeline_class)
-            delayed_features = []
-            if is_time_series(problem_type):
-                delayed_features = [DelayedFeatureTransformer]
-            if estimator_class.model_family == ModelFamily.LINEAR_MODEL:
-                estimator_components = [StandardScaler, estimator_class]
-            elif estimator_class.model_family == ModelFamily.CATBOOST:
-                estimator_components = [estimator_class]
-            else:
```
angela97lin, line 590: I found myself being confused by the different if/else statements (some of which are related, others which are not!) so I thought this would be a clearer breakdown but let me know if you have thoughts/opinions!

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,165 @@
...
-            if estimator_class.model_family == ModelFamily.LINEAR_MODEL:
-                estimator_components = [OneHotEncoder, StandardScaler, estimator_class]
-            elif estimator_class.model_family == ModelFamily.CATBOOST:
-                estimator_components = [estimator_class]
-            else:
-                estimator_components = [OneHotEncoder, estimator_class]
-            if estimator_class.model_family == ModelFamily.ARIMA:
-                expected_components = [Imputer] + estimator_components
+@pytest.fixture
+def get_test_data_from_configuration():
```
angela97lin, line 44: Fixture that contains the data for the make_pipeline test!

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -512,92 +251,37 @@ def test_make_pipeline_numpy_input(problem_type):
...
-    if problem_type == ProblemTypes.MULTICLASS:
-        y = pd.Series([0, 2, 1, 2])
-
-    for estimator_class in estimators:
-        if problem_type in estimator_class.supported_problem_types:
-            parameters = {}
-            if is_time_series(problem_type):
-                parameters = {
-                    "pipeline": {"date_index": "soem dates", "gap": 1, "max_delay": 1},
-                    "Time Series Baseline Estimator": {
```
angela97lin, line 574: Not using Time Series Baseline Estimator so this is unnecessary!

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,204 @@
...
+        ("only dates", ["dates"]),
+        ("only numerical", ["numerical"]),
+        ("only all_null", ["all_null"]),
+        ("only categorical", ["categorical"]),
+        ("text with other features", ["text", "numerical", "categorical"]),
+        ("url with other features", ["url", "numerical", "categorical"]),
+        ("email with other features", ["email", "numerical", "categorical"]),
+    ],
+)
+def test_make_pipeline_master(
```
angela97lin, line 133: `test_description` isn't being used by the code but since we no longer have the title of tests to tell us what is going on, I thought this would be useful to quickly understand what each test case is for.

evalml/tests/pipeline_tests/test_pipeline_utils.py
chukarsten: Love this, anything that consolidates fixtures is good.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,204 @@
...
+                "all_null": [np.nan, np.nan, np.nan, np.nan, np.nan, np.nan, np.nan],
+                "numerical": range(7),
+                "categorical": ["a", "b", "a", "c", "c", "a", "b"],
+                "dates": pd.date_range("2000-02-03", periods=7, freq="W"),
+                "text": [
+                    "this is a string",
+                    "this is another string",
+                    "this is just another string",
+                    "evalml should handle string input",
+                    "cats are gr8",
```
chukarsten, line 59: you're a woman of both taste and distinction

chukarsten (approved):
Solid, solid PR.  I love the mass deletion and consolidation into the fixture.  Very clean, very nice.

evalml/tests/pipeline_tests/test_pipeline_utils.py
freddyaboulton: This is awesome!

evalml/tests/pipeline_tests/test_pipeline_utils.py
freddyaboulton: Can we use it in the assert error message?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,204 @@
...
-                standard_scaler = [StandardScaler]
-            if estimator_class.model_family == ModelFamily.ARIMA:
-                expected_components = standard_scaler + [estimator_class]
-            else:
-                expected_components = (
-                    [DateTimeFeaturizer]
-                    + delayed_features
-                    + standard_scaler
-                    + [estimator_class]
+            delayed_features = (
```
freddyaboulton, line 160: There are two definitions of `delayed_features` in the same scope?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,204 @@
...
-                expected_components = [TextFeaturizer, Imputer] + estimator_components
-            else:
-                expected_components = (
-                    [TextFeaturizer, Imputer] + delayed_features + estimator_components
-                )
-            pipeline.component_graph.compute_order == [
-                component.name for component in expected_components
-            ]
-
+@pytest.mark.parametrize(
```
freddyaboulton, line 116: This is awesome! Non-blocking food-for-thought comment hehe but did you consider including the expected answer in the parametrization rather than computing the answer within the test?

Pro: It'd be much easier to add future test cases that are independent of the other ones. I remember having to "fight" with the answer-generation-logic when I was  adding test cases for URL and Email featurizers. 

Con: There are a lot of test cases so it's a lot of data to specify. There will also be some duplicate answers across these tests.

I think this is great as-is but just wanted to put that out there.

evalml/tests/pipeline_tests/test_pipeline_utils.py
freddyaboulton: Hot take: We shouldn't support plain numpy arrays anymore.

freddyaboulton (approved):
@angela97lin Thank you for this!

ParthivNaresh (approved):
Great work, I love the parametrization of test cases to reduce code and improve readability!

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -40,465 +40,204 @@
...
+    "test_description, column_names",
```
bchen1116, line 117: is `test_description` just to allow people who look at the test to understand what each test is param iteration is testing?

bchen1116 (approved):
Love the cleanup you did here! The new test is very through. Left a comment on some extra simplification, but also left a comment just for my understanding.

evalml/tests/pipeline_tests/test_pipeline_utils.py
angela97lin: Yup! Now that we can't use the test names to understand, I thought having `test_description` would make it easier to see at a quick glance what each test case is for 😁

evalml/tests/pipeline_tests/test_pipeline_utils.py
angela97lin: Gosh, I love this idea a lot. The reason I decided to go with logic, though, is because the estimator also affects what the pipeline looks like, and we currently iterate through each of the estimators in the test. There's room to argue whether or not it's really necessary to loop through all of the estimators as we do or if we should just write out test cases for the special estimators such as CatBoost which can handle categorical cols natively and thus don't require a OneHotEncoder... but maybe this will be worth another iteration of refactoring. 

I do agree though, I had to fight and wrangle the logic quite some bit to get all of these tests to pass, and if we add more logic, I'm not sure how well everything will continue to hold up 😂

PR #2573, opened 2021-07-29
Don't allow certain logical types in partial_dependence

### LINKED ISSUE #2547: Do not Allow Partial Dependence for NaturalLanguage, URL, EmailAddress
Our partial dependence implementation function allows computing partial dependence on NaturalLanguage features.

I was discussing with @rpeck and this doesn't make sense for some features.

Computing the partial dependence requires calculating a grid of quantiles for the feature in question to approximate taking a sample of all possible values for the feature. This grid computation does not make sense for features where a quantile calculation is not well defined, i.e. NaturalLanguage, URL, EmailAddress.  So in short, there is no good way to compute the "x-axis" of the partial dependence plot for these features.

Not that this reasoning does not apply for categorical and datetime features because there are other calculations we do to get a sample of the feature space, e.g. converting to unix epoch time or using all possible values. These operations do not make sense for NaturalLanguage, URL, EmailAddress.

bchen1116:
fix #2547


evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1407,3 +1409,30 @@ def test_partial_dependence_datetime_extra(
...
         assert len(part_dep["feature_values"]) == 10
     assert not part_dep.isnull().any(axis=None)
+
+
+@pytest.mark.parametrize(
+    "cols,expected_cols", [(0, [0]), ([0, 1], [0, 1]), ([0, 2], [0]), (2, [])]
+)
+@pytest.mark.parametrize("types", ["URL", "EmailAddress", "NaturalLanguage"])
+def test_partial_dependence_not_allowed_types(types, cols, expected_cols, X_y_binary):
+    # test to see if we can get partial dependence fine with a dataset that has unknown features
```
chukarsten, line 1419: copy pasta?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1407,3 +1409,30 @@ def test_partial_dependence_datetime_extra(
...
+    X, y = X_y_binary
+    X = pd.DataFrame(X)
+    X.ww.init(logical_types={0: types, 1: "URL"})
```
chukarsten, line 1422: Is the intent here for "URL" to be the parametrized type?

chukarsten (changes requested):
I think this looks good.  The test needs to be changed to actually parameterize over the different disallowed types, though, I think.

eccabay (commented):
Looks pretty good to me! Just a few places where I think you can clean up the logic a bit to make it more readable.

evalml/tests/model_understanding_tests/test_partial_dependence.py
bchen1116: @chukarsten This makes col 0 parametrized with the passed-in types, and col 1 be of type 'URL'. This allows me to make sure if I pass in 2 non-allowed columns, (0, 1) to partial dependence, I get the message that both columns aren't allowed and that both types are reported.

chukarsten: I see now, I totally misread this.

chukarsten (approved):
LGTM! Thanks for pulling that copy pasta.

evalml/model_understanding/graphs.py
```diff
@@ -659,6 +659,14 @@ def partial_dependence(
             f"Columns {X_unknown.columns.values} are of type 'Unknown', which cannot be used for partial dependence"
         )
 
+    X_not_allowed = X_features.ww.select(["URL", "EmailAddress", "NaturalLanguage"])
+    if len(X_not_allowed.columns):
+        # these three logical types aren't allowed for partial dependence
```
angela97lin, line 664: Nit picky but if we want to keep this comment, maybe annotate `X_not_allowed` instead? :')

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1407,3 +1409,29 @@ def test_partial_dependence_datetime_extra(
...
         assert len(part_dep["feature_values"]) == 10
     assert not part_dep.isnull().any(axis=None)
+
+
+@pytest.mark.parametrize(
+    "cols,expected_cols", [(0, [0]), ([0, 1], [0, 1]), ([0, 2], [0]), (2, [])]
+)
+@pytest.mark.parametrize("types", ["URL", "EmailAddress", "NaturalLanguage"])
+def test_partial_dependence_not_allowed_types(types, cols, expected_cols, X_y_binary):
+    X, y = X_y_binary
```
angela97lin, line 1419: I think this is fine as is, but I wonder if we could make it more readable by either just using a X that is explicitly 3 cols or indicating that we're choosing col 2 because it's expected to be numeric? Just thinking that if we ever change X_y_binary to be not fully numeric, this could start changing the expected behavior of this test in unexpected ways.

angela97lin (approved):
LGTM! Left some nitpicky comments but nothing blocking 😁

PR #2577, opened 2021-08-02
Support for NaN types for NLP in AutoMLSearch

### LINKED ISSUE #2515: Support for missing data in Natural Language types
As of the pandas 1.3.0 upgrade, when we cast columns to Natural language type, the imputer fails. This is due to pandas changing the `NaN` value to `<NA>` instead (pd.NA), which fails on the imputer when finding the mask of missing values.
repro:
```python
import pandas as pd
import woodwork as ww
import numpy as np

X = pd.DataFrame({"a": [np.nan] + [i for i in range(99)],
                  "b": [np.nan] + [f"string {i} is valid" for i in range(99)]})
X.ww.init(logical_types={"b": "NaturalLanguage"})
y = pd.Series([0] * 25 + [1] * 75)

automl = AutoMLSearch(X_train=X, y_train=y, problem_type='binary')
automl.search()
```
image

This issue tracks fixing this so that missing values are once again supported

bchen1116:
close #2515 

Added some support to make sure missing natural language types are supported in `AutoMLSearch`. We change `pd.NA` to `None` to ensure we can preserve the `float64` type that we expect from the text featurizer.

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
@@ -150,7 +150,9 @@ class TextFeaturizer(TextTransformer):
             for column, derived_features in lsa_features.items():
                 X_lsa.loc[nan_mask[column], derived_features] = None
         X_lsa.ww.init(logical_types={col: "Double" for col in X_lsa.columns})
-        X_nlp_primitives.ww.init(logical_types={col: "Double" for col in X_nlp_primitives.columns})
+        X_nlp_primitives.ww.init(
+            logical_types={col: "Double" for col in X_nlp_primitives.columns}
+        )
         X_ww = X_ww.ww.drop(self._text_columns)
         for col in X_nlp_primitives:
             X_ww.ww[col] = X_nlp_primitives[col]
--- a/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
@@ -2,7 +2,6 @@ import string
 
 import featuretools as ft
 import nlp_primitives
-import pandas as pd
 
 from evalml.pipelines.components.transformers.preprocessing import (
     LSA,
@@ -141,17 +140,17 @@ class TextFeaturizer(TextTransformer):
             {s: "NaturalLanguage" for s in self._text_columns},
         )
         X_lsa = self._lsa.transform(X_ww_altered)
+        X_nlp_primitives.set_index(X_ww.index, inplace=True)
         if any_nans:
             primitive_features = self._get_primitives_provenance(self._features)
             for column, derived_features in primitive_features.items():
-                X_nlp_primitives.loc[nan_mask[column], derived_features] = pd.NA
+                X_nlp_primitives.loc[nan_mask[column], derived_features] = None
 
             lsa_features = self._lsa._get_feature_provenance()
             for column, derived_features in lsa_features.items():
-                X_lsa.loc[nan_mask[column], derived_features] = pd.NA
-
-        X_nlp_primitives.set_index(X_ww.index, inplace=True)
-
+                X_lsa.loc[nan_mask[column], derived_features] = None
+        X_lsa = X_lsa.astype("float64")
+        X_nlp_primitives = X_nlp_primitives.astype("float64")
         X_ww = X_ww.ww.drop(self._text_columns)
         for col in X_nlp_primitives:
             X_ww.ww[col] = X_nlp_primitives[col]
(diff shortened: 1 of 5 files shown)


evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -141,17 +140,17 @@ def transform(self, X, y=None):
             {s: "NaturalLanguage" for s in self._text_columns},
         )
         X_lsa = self._lsa.transform(X_ww_altered)
+        X_nlp_primitives.set_index(X_ww.index, inplace=True)
```
bchen1116, line 143: Reset the index to be equal to the input X

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
```diff
@@ -141,17 +140,17 @@ def transform(self, X, y=None):
...
+                X_nlp_primitives.loc[nan_mask[column], derived_features] = None
 
             lsa_features = self._lsa._get_feature_provenance()
             for column, derived_features in lsa_features.items():
-                X_lsa.loc[nan_mask[column], derived_features] = pd.NA
-
-        X_nlp_primitives.set_index(X_ww.index, inplace=True)
-
+                X_lsa.loc[nan_mask[column], derived_features] = None
+        X_lsa = X_lsa.astype("float64")
```
bchen1116, line 152: Enforce that the types are 'float64', in case pandas decides to treat `None` similarly to `pd.NA` or `np.nan`. The latter two cause the dtypes to become objects (and thus treats these columns as categorical)

freddyaboulton: IMO we should use `X_lsa.ww.init(logical_types={col: "Double" for col in X_lsa.columns})`. The reason is that if we don't set a ww type explicitly, `ww` will still infer types for the features in `X_lsa`. Right now, that inference works in our favor because ww uses `pd.dtypes.is_float_dtype`. I think the safer thing to do here is to set an explicit woodwork logical type and avoid the type inference.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4933,3 +4933,18 @@ def test_pipeline_parameter_warnings_component_graphs(
...
+    X = pd.DataFrame(
+        {
+            "a": [np.nan] + [i for i in range(99)],
+            "b": [np.nan] + [f"string {i} is valid" for i in range(99)],
+        }
+    )
+    X.ww.init(logical_types={"b": "NaturalLanguage"})
+    y = pd.Series([0] * 25 + [1] * 75)
+
+    automl = AutoMLSearch(X_train=X, y_train=y, problem_type="binary", max_iterations=2)
```
freddyaboulton, line 4949: I think the follow-up bug that we encountered was that the types were being set to categorical by woodwork and only the LightGBMClassifier failed because of it right?. 

I think it may be better to use the `AutoMLSearchEnv` and check all of the woodwork types are only Double/Integer in `mock_fit`. I see we also check that in `test_text_featurizer.py` now but if we're going to add an AutoMLSearch test, I think it's worth it to run a full batch and check what we're after with mocks. Correct me if I'm wrong but this test wouldn't catch the LGBM problem right?

freddyaboulton (approved):
Thanks @bchen1116 ! I left a suggestion for `test_search_text_with_nans` but otherwise this looks fantastic!

evalml/tests/automl_tests/test_automl.py
bchen1116: Yeah, good point! I'll patch that

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
angela97lin: Just curious, why is this necessary for this PR?

angela97lin (approved):
This looks great! Just had a question about why we need to reset index--if it's related to this PR or if it's something you noticed but LGTM 😁

evalml/pipelines/components/transformers/preprocessing/text_featurizer.py
bchen1116: During AutoMLSearch, when we do the data splits, we get input indices that aren't in consecutive order. Seems like `X_nlp_primitives` completely resets the index (so it becomes 0, 1, 2, ...), which causes errors when we try to set the `None` values using the mask. I just moved this line from below to above the `for` loop so that the `.loc` method doesn't fail

freddyaboulton: We don't have to do it for the lsa features because they are computed by the lsa component, which preserves the index because our components preserve indices!!

PR #2578, opened 2021-08-02
Rename existing ensembler implementation from `StackedEnsemblers` to `SklearnStackedEnsemblers`

christopherbunn:
Renaming existing ensembler component to clarify it is sklearn based and to differentiate it from the new ensembler component.


chukarsten (approved):
LGTM.

Issue #2581, opened 2021-08-03
Flaky test in nightlies: target distribution data check


dsherry:
I saw the following fail in [the linux nightlies yesterday](https://github.com/alteryx/evalml/runs/3228179479)
```
FAILED evalml/tests/data_checks_tests/test_target_distribution_data_check.py::test_target_distribution_data_check_unsupported_target_type[double]
...
        if target_type in ["integer", "double"]:
>           assert target_dist_check.validate(X, y) == {
                "warnings": [],
                "errors": [],
                "actions": [],
            }
E           AssertionError: assert {'actions': [...rning', ...}]} == {'actions': [...warnings': []}
E             Omitting 1 identical items, use -vv to show
E             Differing items:
E             {'actions': [{'code': 'TRANSFORM_TARGET', 'metadata': {'column': None, 'is_target': True, 'transformation_strategy': 'lognormal'}}]} != {'actions': []}
E             {'warnings': [{'code': 'TARGET_LOGNORMAL_DISTRIBUTION', 'data_check_name': 'TargetDistributionDataCheck', 'details': {'shapiro-statistic/pvalue': '0.969/0.018'}, 'level': 'warning', ...}]} != {'warnings': []}
E             Use -v to get the full diff

evalml/tests/data_checks_tests/test_target_distribution_data_check.py:59: AssertionError
```

Tagging @ParthivNaresh since I think he touched this code recently and may have an idea on the fix

freddyaboulton:
I think we just got a flake on a different test for `build_conda_pkg`: 

https://github.com/alteryx/evalml/pull/2585/checks?check_run_id=3242371966

dsherry:
Got it @freddyaboulton , so it was a different test but from the same file / area.

freddyaboulton:
Looks like we saw a similar flake due to rounding in the 3.9 nightlies: https://github.com/alteryx/evalml/runs/3503548207

Reopening this issue to track if it happens again!

ParthivNaresh:
I have a fix for this, will be putting it up now

PR #2598, opened 2021-08-05
Install prophet during linux nightlies

freddyaboulton:
### Pull Request Description

Fixes nightlies by installing prophet . They originally failed because `test_all_components` was expecting Prophet to be installed but it was missing from the build script. 

You can see the nightlies passing here: https://github.com/alteryx/evalml/actions/runs/1102244065

-----

### DIFF
--- a/evalml/tests/component_tests/test_utils.py
+++ b/evalml/tests/component_tests/test_utils.py
@@ -35,8 +35,12 @@ def test_all_components(
         n_components = 41
     elif is_using_conda:
         n_components = 52
-    elif is_running_py_39_or_above or is_using_windows:
+    elif is_using_windows and not is_running_py_39_or_above:
         n_components = 53
+    elif is_using_windows and is_running_py_39_or_above:
+        n_components = 51
+    elif not is_using_windows and is_running_py_39_or_above:
+        n_components = 52
     else:
         n_components = 54
     assert len(all_components()) == n_components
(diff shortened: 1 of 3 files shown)


evalml/tests/component_tests/test_utils.py
```diff
@@ -35,8 +35,12 @@ def test_all_components(
         n_components = 41
     elif is_using_conda:
         n_components = 52
-    elif is_running_py_39_or_above or is_using_windows:
+    elif is_using_windows and not is_running_py_39_or_above:
```
freddyaboulton, line 38: There was a slight bug in the prophet pr I think. This is what the number of components were before prophet and the new number of components depending on the environment:

image

freddyaboulton:
Test that is failing is same as #2581 so it's unrelated to this change

evalml/tests/component_tests/test_utils.py
```diff
@@ -35,8 +35,12 @@ def test_all_components(
...
         n_components = 53
+    elif is_using_windows and is_running_py_39_or_above:
```
angela97lin, line 40: I'm starting to get confused by all of these +/-1 numbers, lol... Do you think you could add an in-line comment stating which components we expect not to get installed? 🙈

angela97lin (approved):
Looks good! Thank you :)

I'm starting to get confused by all of these different cases in `test_all_components`. Do you think you could add an in-line comment stating which components we expect not to get installed so it's more clear when things are not as expected? 🙈

evalml/tests/component_tests/test_utils.py
freddyaboulton: Yes!

freddyaboulton: This will be cleaner once sktime supports 3.9 hehe

PR #2583, opened 2021-08-03
Update ``ComponentGraph`` to handle not calling `transform` during `predict`, and update samplers' `transform` methods s.t. `fit_transform` is equivalent to `fit(X, y).transform(X, y)`

angela97lin:
- Closes #2273 
- Removes the need to keep track of the most recent y value and consolidating inputs. We should be able to just grab the parent output now that ComponentGraphs must be fully specified.
   -  This is related to sampler behavior because we currently call "Undersampler.y" expecting that it'll grab the parent input. Since we're changing the undersampler to no longer return None, the component graph will take the undersampler output, which is not what we want.
- General cleanup of ``BaseSampler`` and subclasses 

A little writeup of sampler and ComponentGraph behavior here: https://alteryx.atlassian.net/wiki/spaces/PS/pages/958890019/Samplers+and+Component+Graph

PR #2599, opened 2021-08-05
Update ComponentGraph `_validate_component_dict` logic

angela97lin:
- Remove "Cannot have multiple y parents for a single component" check in _compute_features, all checking is done in `_validate_component_dict` now.
- Cleaned up and added edge case to `_validate_component_dict` (where user specifies .x, .y, but also "thisisnotvalid" 😂). Before, this check would pass as long as there was one .x, one .y properly. This update ensures that all specified values must be .x/.y/X, y :)

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -74,14 +74,22 @@ class ComponentGraph:
                     "All components must have exactly one target (.y/y) edge."
                 )
 
-            def f(i):
+            def check_all_inputs_have_correct_syntax(edge):
                 return not (
-                    i.endswith(".y") or i == "y" or i.endswith(".x") or i == "X"
+                    edge.endswith(".y")
+                    or edge == "y"
+                    or edge.endswith(".x")
+                    or edge == "X"
                 )
 
-            if len(list(filter(f, component_inputs))) != 0:
+            if (
+                len(
+                    list(filter(check_all_inputs_have_correct_syntax, component_inputs))
+                )
+                != 0
+            ):
                 raise ValueError(
-                    "All edges must be specified as either an input feature (.x) or input target (.y)."
+                    "All edges must be specified as either an input feature ('X'/.x) or input target ('y'/.y)."
                 )
 
     @property
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -61,11 +61,25 @@ class ComponentGraph:
                 component_input.endswith(".x") or component_input == "X"
                 for component_input in component_inputs
             )
-            has_target_input = any(
+            has_one_target_input = sum(
                 component_input.endswith(".y") or component_input == "y"
                 for component_input in component_inputs
             )
-            if not (has_feature_input and has_target_input):
+            if not has_feature_input:
+                raise ValueError(
+                    "All components must have at least one input feature (.x/X) edge."
+                )
+            if has_one_target_input != 1:
+                raise ValueError(
+                    "All components must have exactly one target (.y/y) edge."
+                )
+
+            def f(i):
+                return not (
+                    i.endswith(".y") or i == "y" or i.endswith(".x") or i == "X"
+                )
+
+            if len(list(filter(f, component_inputs))) != 0:
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


evalml/pipelines/component_graph.py
```diff
@@ -61,13 +61,35 @@ def _validate_component_dict(self):
                 component_input.endswith(".x") or component_input == "X"
                 for component_input in component_inputs
             )
-            has_target_input = any(
+            has_one_target_input = sum(
                 component_input.endswith(".y") or component_input == "y"
                 for component_input in component_inputs
             )
-            if not (has_feature_input and has_target_input):
+            if not has_feature_input:
```
angela97lin, line 68: Kind of redundant / can be covered by case below. Kept for the specific ValueError, but open to just removing and having it encapsulated below 😂

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1917,9 +1914,35 @@ def test_component_graph_does_not_define_all_edges():
...
 
 
+def test_component_graph_defines_edges_with_bad_syntax():
+    # Graph does not define an X edge
+    with pytest.raises(
+        ValueError, match="All edges must be specified as either an input feature"
+    ):
+        ComponentGraph(
+            {
+                "Imputer": [Imputer, "X", "y"],  # offending line
```
freddyaboulton, line 1924: This is not actually the offending line right?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1917,9 +1914,35 @@ def test_component_graph_does_not_define_all_edges():
...
+                "One Hot Encoder": [OneHotEncoder, "Imputer.x", "y"],
+                "Target Imputer": [
+                    TargetImputer,
+                    "Imputer",
+                    "One Hot Encoder.x",
+                    "y",
+                ],  # offending line: Imputer not allowed
```
freddyaboulton, line 1931: nit-pick: I think the comment would make more sense if it was above "Target Imputer"

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -700,7 +693,7 @@ def test_component_graph_evaluation_plumbing(
     graph = {
         "transformer a": [TransformerA, "X", "y"],
         "transformer b": [TransformerB, "transformer a.x", "y"],
-        "transformer c": [TransformerC, "transformer a", "transformer b.x", "y"],
+        "transformer c": [TransformerC, "transformer a.x", "transformer b.x", "y"],
```
freddyaboulton, line 696: Ok so before "transformer a" was being silently ignored? I'm curious why this change doesn't mean we also have to change the actual expected call args for `mock_transc`

freddyaboulton (approved):
Thank you @angela97lin !!

evalml/tests/pipeline_tests/test_component_graph.py
freddyaboulton: Ohh it's cause the output of transformer b already had the output of transformer a. I wonder if we should change the test so that the column names don't overlap? Duplicate column names would cause ww to error out anyways.

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Wow I swore I updated this but I must have dreamed it. Thank you 😂

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Going to put it next to the "Imputer" :)

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Hehe so right now, `parent_x = output_cache.get(parent_input, output_cache.get(f"{parent_input}.x"))` means that it'll try either getting `transformer a` or `transformer a.x` as a backup, which is why this worked. We're no longer going to have to default to this anymore but but https://github.com/alteryx/evalml/pull/2583 is going to clean that up further :)

Issue #2502, opened 2021-07-13
Implement our own partial dependence method


bchen1116:
With [this PR](https://github.com/alteryx/evalml/pull/2454), we add a section of code to handle datetime support for partial dependence. @chukarsten [brings up a good point](https://github.com/alteryx/evalml/pull/2454#pullrequestreview-703171806) about risks involved in using the private functions of sklearn's partial dependence, `_grid_from_X` and `_partial_dependence_brute`.

In my look into the issue, I found the following:
- sklearn's `partial_dependence`  makes calls to `_grid_from_X` and `_partial_dependence_brute`
- `_grid_from_X` doesn't accept datetime features. If a datetime is passed in, it will result in the original datetimes to be returned, ignoring the `grid_resolution` parameter. We resolve this by turning the datetime into seconds (an int value), then pass it forward
- `_partial_dependence_brute` calls the `pipeline.predict/predict_proba` methods on the trained pipelines. This means that we can't alter the dataset in only partial dependence if it isn't altered similarly for training (ie we cannot introduce a new column or alter the X data)

If we wanted to not use the the private functions, we need to find a way to properly pass the data to sklearn's `partial_dependence`. The issue here is that we somehow need to handle converting the datetime to seconds for obtaining the grid, but we'd need to use datetime for fitting, predicting, and computing the partial dependence otherwise.

It might be beneficial to build our own partial dependence feature or to find another way to pass this through without necessarily relying on private methods. This issue tracks finding the next best-steps to clean up our current partial dependence implementation.

fyi @chukarsten @freddyaboulton

dsherry:
Agreed. Goal is to avoid more issues like #2475 , make our code easier to read and maintain.

Seems like we have two options. 1) clean up our current impl 2) don't use sklearn impl any more, write our own.

PR #2604, opened 2021-08-06
Raise a custom exception for partial dependence errors

### LINKED ISSUE #2539: Raise a custom exception for partial dependence errors
Our partial dependence function raises a ValueError in a variety of different [cases](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/graphs.py#L607). This is great, but since sklearn also raises ValueErrors, we're lumping together "expected" errors that are caused by user input with possible bugs in our implementation that are causing sklearn to raise an exception.

This makes it hard for us as developers to know what went wrong when a user encounters an error. This also makes it hard for users that leverage our partial dependence implementation in their own libraries to know whether the error is caused by their input or by EvalML.

It would be better if our partial dependence implementation raised its own exception class that will catch our expected failure cases and provide a code so that users can know which "case" of error they fell in.

```python
from enum import Enum

class PartialDependenceErrorCode(Enum):
  TOO_MANY_FEATURES = "too_many_features"
  FEATURES_ARGUMENT_INCORRECT_TYPES = "features_argument_too_many_types"
  UNFITTED_PIPELINE = "unfitted_pipeline"
  PIPELINE_IS_BASELINE = "pipeline_is_baseline"
  TWO_WAY_REQUESTED_FOR_DATES = "two_way_requested_for_dates"
  FEATURE_IS_ALL_NANS = "feature_is_all_nans"
  FEATURE_IS_MOSTLY_ONE_VALUE = "feature_is_mostly_one_value"
  COMPUTED_PERCENTILES_TOO_CLOSE = "computed_percentiles_too_close"
  ALL_OTHER_ERRORS = "all_other_errors"

class PartialDependenceError(Exception):

  def __init__(self, message, error_code: PartialDependenceErrorCode):
    self.error_code = error_code
    super().__init__(message)
```

freddyaboulton:
### Pull Request Description

Fixes #2539 

Design document [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/654573616/Surfacing+Informative+Errors+for+Partial+Dependence+in+AML) .

-----


evalml/model_understanding/graphs.py
```diff
@@ -605,221 +612,250 @@ def partial_dependence(
...
-            _is_feature_of_type(features, X, ww.logical_types.Categorical)
-        ]
-        is_datetime = [_is_feature_of_type(features, X, ww.logical_types.Datetime)]
-
-    if isinstance(features, (list, tuple)):
-        if len(features) != 2:
-            raise ValueError(
-                "Too many features given to graph_partial_dependence.  Only one or two-way partial "
-                "dependence is supported."
+    try:
```
freddyaboulton, line 623: Adding this `try` unfortunately makes the diff seem bigger than it is. I think it's important to catch all possible errors that this can raise so that users don't have to resort to `except Exception` or `except PartialDependenceError, ValueError` and hope that'll catch everything.

freddyaboulton: We have an outstanding issue to clean up partial dependence (https://github.com/alteryx/evalml/issues/2502) so maybe this will be cleaner once we redesign this with this requirement in mind.

chukarsten (approved):
Very awesome, great work.  Thanks for consolidating all these exceptions and making it super tidy.  I think this will also help downstream consumers of EvalML a LOT.

bchen1116 (approved):
LGTM! Thanks for cleaning this up and making it easier to tell users what the issue is. The error docs don't appear on the [docs](https://feature-labs-inc-evalml--2604.com.readthedocs.build/en/2604/api_index.html#exceptions) though, unless you explicitly type it into the [search](https://feature-labs-inc-evalml--2604.com.readthedocs.build/en/2604/autoapi/evalml/exceptions/exceptions/index.html?highlight=partialdep#module-contents)?

ParthivNaresh (approved):
Solid work, love the Enum implementation

evalml/model_understanding/graphs.py
```diff
@@ -605,221 +612,250 @@ def partial_dependence(
...
+            if "percentiles are too close to each other" in str(e):
+                raise PartialDependenceError(
+                    "The scale of these features is too small and results in"
+                    "percentiles that are too close together.  Partial dependence"
+                    "cannot be computed for these types of features.  Consider"
+                    "scaling the features so that they differ by > 10E-7",
+                    code=PartialDependenceErrorCode.COMPUTED_PERCENTILES_TOO_CLOSE,
+                )
+            else:
+                raise e
```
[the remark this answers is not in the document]
freddyaboulton: Yea it's caused by something else in sklearn. Raising here makes sure the final `except` catches it and then we can wrap as `PartialDependenceError` with the "all other errors code"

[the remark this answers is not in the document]
freddyaboulton: Also, about the docs page. Yea you have to click on the drop down on the left hand side and then click on exceptions.

The "API Reference" is a manually curated list of functions/classes but it's not exhaustive.

angela97lin (approved):
I love this change. LGTM!

evalml/model_understanding/graphs.py
```diff
@@ -605,221 +612,250 @@ def partial_dependence(
...
-        grid_resolution = max([max_num_cats + 1, grid_resolution])
-
-    X_dt = X_features.ww.select("datetime")
-
-    if isinstance(features, (list, tuple)):
-        feature_names = _get_feature_names_from_str_or_col_index(X, features)
-        if any(is_datetime):
-            raise ValueError(
-                "Two-way partial dependence is not supported for datetime columns."
+        X_unknown = X_features.ww.select("unknown")
```
angela97lin, line 668: Can we combine this with the following check for "URL", "EmailAddress", "NaturalLanguage"? They all have the `INVALID_FEATURE_TYPE` code :)

PR #2608, opened 2021-08-09
Fix Inconsistent Test Outcome from TargetDistributionDataCheck

### LINKED ISSUE #2581: Flaky test in nightlies: target distribution data check

ParthivNaresh:
Fixes #2581 
-----

### DIFF
--- a/evalml/data_checks/target_distribution_data_check.py
+++ b/evalml/data_checks/target_distribution_data_check.py
@@ -30,14 +30,14 @@ class TargetDistributionDataCheck(DataCheck):
         Example:
             >>> from scipy.stats import lognorm
             >>> X = None
-            >>> y = lognorm.rvs(0.1, size=1000)
-            >>> target_checdk = TargetDistributionDataCheck()
+            >>> y = [0.946, 0.972, 1.154, 0.954, 0.969, 1.222, 1.038, 0.999, 0.973, 0.897]
+            >>> target_check = TargetDistributionDataCheck()
             >>> assert target_check.validate(X, y) == {"errors": [],\
                                                        "warnings": [{"message": "Target may have a lognormal distribution.",\
                                                                     "data_check_name": "TargetDistributionDataCheck",\
                                                                     "level": "warning",\
                                                                     "code": "TARGET_LOGNORMAL_DISTRIBUTION",\
-                                                                    "details": {"shapiro-statistic/pvalue": '0.993/0.0'}}],\
+                                                                    "details": {"shapiro-statistic/pvalue": '0.84/0.045'}}],\
                                                         "actions": [{'code': 'TRANSFORM_TARGET', 'metadata': {'column': None, 'transformation_strategy': 'lognormal', 'is_target': True}}]}
         """
         results = {"warnings": [], "errors": [], "actions": []}
@@ -78,7 +78,7 @@ class TargetDistributionDataCheck(DataCheck):
         if shapiro(y).pvalue >= 0.05:
             return results
 
-        y_new = y
+        y_new = round(y, 6)
         if any(y <= 0):
             y_new = y + abs(y.min()) + 1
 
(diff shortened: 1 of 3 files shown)


evalml/data_checks/target_distribution_data_check.py
```diff
@@ -97,7 +97,7 @@ def validate(self, X, y):
 
         if log_detected:
             details = {
-                "shapiro-statistic/pvalue": f"{round(shapiro_test_og.statistic, 3)}/{round(shapiro_test_og.pvalue, 3)}"
+                "shapiro-statistic/pvalue": f"{round(shapiro_test_og.statistic, 2)}/{round(shapiro_test_og.pvalue, 3)}"
```
ParthivNaresh, line 100: Based on the previous implementation, 4-5 out of 100 runs would result in slightly different values due to how rounding would occur after a certain number of decimal places. This would throw off the statistic to the thousandths place by 1, resulting in a failure. I'm rounding all values here to keep the results the same in the data check and the test block.

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
```diff
@@ -37,19 +37,17 @@ def test_target_distribution_data_check_no_y(X_y_regression):
...
+        y = pd.Series([True, False] * 5)
     elif target_type == "categorical":
-        y = pd.Series(["One", "Two", "Three", "Four"] * 25)
+        y = pd.Series(["One", "Two", "Three", "Four", "Five"] * 2)
     elif target_type == "integer":
-        y = pd.Series(np.random.randint(1, 25, 100))
+        y = [-1, -3, -5, 4, -2, 4, -4, 2, 1, 1]
     else:
-        y = np.random.normal(4, 1, 100)
+        y = [9.2, 7.66, 4.93, 3.29, 4.06, -1.28, 4.95, 6.77, 9.07, 7.67]
```
ParthivNaresh, line 50: Both these test blocks would previously throw errors 7-8 times out of a 100 runs. Even after changing the values to be more gaussian in distribution with larger sample sizes, they still failed 1-2 times out of the 100,000 simulations I ran. Randomness leads to a somewhat lognormal distribution every once in a while. I decided to replace these with constant smaller samples of a lognormal distribution so we don't have to worry about inconsistency again.

chukarsten (approved):
LGTM

evalml/data_checks/target_distribution_data_check.py
```diff
@@ -30,14 +30,14 @@ def validate(self, X, y):
         Example:
             >>> from scipy.stats import lognorm
             >>> X = None
-            >>> y = lognorm.rvs(0.1, size=1000)
-            >>> target_checdk = TargetDistributionDataCheck()
+            >>> y = [0.946, 0.972, 1.154, 0.954, 0.969, 1.222, 1.038, 0.999, 0.973, 0.897]
```
freddyaboulton, line 33: 👍 agree that it's best to not rely on random vaues for this

evalml/data_checks/target_distribution_data_check.py
freddyaboulton: Sweet. If the value of this ratio impacted whether or not we did the log transform, then I would have suggested we instead do the rounding in the test as opposed to the implementation. Since this is just to provide extra details to users, I think it's ok to round however we want!

evalml/tests/data_checks_tests/test_target_distribution_data_check.py
freddyaboulton: Agreed!

freddyaboulton (approved):
Thank you @ParthivNaresh !!

PR #2609, opened 2021-08-09
Remove `component_graphs` as a global variable in `test_component_graphs.py`

### LINKED ISSUE #2571: Remove `component_graphs` as a global variable in `test_component_graphs.py`
I noticed that we have a variable `component_graphs` in `test_component_graphs.py` set to a list of component graphs which is then used in `test_component_graph_inverse_transform`. I propose that we make this "local" to `test_component_graph_inverse_transform`. Otherwise, trying to debug anything in `ComponentGraph` will always try to initialize these graphs before running the actual test in question, which can be confusing.

angela97lin:
Closes #2571. Simple change, just moves `component_graph` to the parametrize argument.


jeremyliweishih (approved):
nice! good fix 😄

freddyaboulton (approved):
Looks good to me!

bchen1116 (approved):
Nice!

PR #2610, opened 2021-08-09
Updates to support Woodwork 0.5.1

### LINKED ISSUE #2543: Upgrade to Woodwork 0.5.1
WW release 0.5.1 introduced a breaking change regarding inference for columns that are all NaN. Here is a possibly  incomplete list of the breakages:

- Prior to 0.5.0, partial dependence would work on numeric columns with missing values. It would fill the entire column with NaNs but the physical type would still be float so the imputer could handle it. Now the physical type changes to string and the Imputer cannot handle missing values in string physical types. I think this one requires discussion?
- Our data checks action for all null columns are broken cause they also pull in the "Unknown" data check message.
- The imputers can no longer handle all null columns for the same reason as why the partial dependence is broken.

I think someone should go through the tests that fail and propose some changes to our behavior regarding unknown columns!

chukarsten:
Fixes #2543

### DIFF
--- a/evalml/data_checks/invalid_targets_data_check.py
+++ b/evalml/data_checks/invalid_targets_data_check.py
@@ -80,7 +80,8 @@ class InvalidTargetDataCheck(DataCheck):
 
         y = infer_feature_types(y)
         is_supported_type = y.ww.logical_type.type_string in numeric_and_boolean_ww + [
-            ww.logical_types.Categorical.type_string
+            ww.logical_types.Categorical.type_string,
+            ww.logical_types.Unknown.type_string,
         ]
         if not is_supported_type:
             results["errors"].append(
--- a/evalml/data_checks/invalid_targets_data_check.py
+++ b/evalml/data_checks/invalid_targets_data_check.py
@@ -81,7 +81,6 @@ class InvalidTargetDataCheck(DataCheck):
         y = infer_feature_types(y)
         is_supported_type = y.ww.logical_type.type_string in numeric_and_boolean_ww + [
             ww.logical_types.Categorical.type_string,
-            ww.logical_types.Unknown.type_string,
         ]
         if not is_supported_type:
             results["errors"].append(
--- a/evalml/pipelines/components/transformers/imputers/target_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/target_imputer.py
@@ -74,9 +74,15 @@ class TargetImputer(Transformer, metaclass=TargetImputerMeta):
         Returns:
             self
         """
+        from woodwork.logical_types import Unknown
+
         if y is None:
             return self
-        y = infer_feature_types(y).to_frame()
+        y = infer_feature_types(y)
+        if isinstance(y.ww.logical_type, Unknown):
+            raise TypeError("Provided target full of pd.NA.")
+        y = y.to_frame()
+        # should y be an un-inited dataframe?
 
         # Convert all bool dtypes to category for fitting
         if (y.dtypes == bool).all():
@@ -110,8 +116,6 @@ class TargetImputer(Transformer, metaclass=TargetImputerMeta):
             )
 
         transformed = self._component_obj.transform(y_df)
-        if transformed.shape[1] == 0:
-            raise RuntimeError("Transformed data is empty")
         y_t = pd.Series(transformed[:, 0], index=y_ww.index)
         return X, _retain_custom_types_and_initalize_woodwork(y_ww.ww.logical_type, y_t)
 
--- a/evalml/pipelines/components/transformers/imputers/target_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/target_imputer.py
@@ -74,15 +74,12 @@ class TargetImputer(Transformer, metaclass=TargetImputerMeta):
         Returns:
             self
         """
-        from woodwork.logical_types import Unknown
-
         if y is None:
             return self
         y = infer_feature_types(y)
-        if isinstance(y.ww.logical_type, Unknown):
-            raise TypeError("Provided target full of pd.NA.")
+        if all(y.isnull()):
+            raise TypeError("Provided target full of nulls.")
         y = y.to_frame()
-        # should y be an un-inited dataframe?
 
         # Convert all bool dtypes to category for fitting
         if (y.dtypes == bool).all():
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -79,8 +79,8 @@ def _get_preprocessing_components(
     if len(all_null_cols) > 0:
         pp_components.append(DropNullColumns)
 
-    unknown_columns = list(X.ww.select(["unknown"], return_schema=True).columns)
-    if len(unknown_columns) and not set(all_null_cols) == set(unknown_columns):
+    unknown_columns = set(X.ww.select(["unknown"], return_schema=True).columns)
+    if len(unknown_columns) and not set(all_null_cols) == unknown_columns:
         pp_components.append(DropColumns)
 
     email_columns = list(X.ww.select("EmailAddress", return_schema=True).columns)
@@ -109,8 +109,8 @@ def _get_preprocessing_components(
         text_columns
     ):
         pp_components.append(Imputer)
-    index_columns = list(X.ww.select(["index"], return_schema=True).columns)
-    if len(index_columns) > 0:
+
+    if X.ww.index is not None:
         pp_components.append(DropColumns)
 
     datetime_cols = list(X.ww.select(["Datetime"], return_schema=True).columns)
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -79,8 +79,10 @@ def _get_preprocessing_components(
     if len(all_null_cols) > 0:
         pp_components.append(DropNullColumns)
 
-    unknown_columns = set(X.ww.select(["unknown"], return_schema=True).columns)
-    if len(unknown_columns) and not set(all_null_cols) == unknown_columns:
+    index_and_unknown_columns = list(
+        X.ww.select(["index", "unknown"], return_schema=True).columns
+    )
+    if len(index_and_unknown_columns) > 0:
         pp_components.append(DropColumns)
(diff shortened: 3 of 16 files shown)


evalml/pipelines/utils.py
```diff
@@ -105,10 +109,8 @@ def _get_preprocessing_components(
         text_columns
     ):
         pp_components.append(Imputer)
-    index_and_unknown_columns = list(
-        X.ww.select(["index", "unknown"], return_schema=True).columns
-    )
-    if len(index_and_unknown_columns) > 0:
+    index_columns = list(X.ww.select(["index"], return_schema=True).columns)
+    if len(index_columns) > 0:
```
thehomebrewnerd, line 113: Woodwork will only ever have one index column, or it will be `None`, so I think you can combine these two lines into just this:
```python
if X.ww.index:
```

thehomebrewnerd (commented):
Was a bit curious about the updates required for WW so took a quick look at this. Noticed a couple things that I thought I'd mention. Feel free to ignore or act upon as you see fit.

evalml/pipelines/utils.py
```diff
@@ -105,10 +109,8 @@ def _get_preprocessing_components(
         text_columns
     ):
         pp_components.append(Imputer)
-    index_and_unknown_columns = list(
-        X.ww.select(["index", "unknown"], return_schema=True).columns
-    )
-    if len(index_and_unknown_columns) > 0:
+
+    if X.ww.index is not None:
```
freddyaboulton, line 113: So if the pipeline has an index and unknown columns, it will have two drop columns transformers? I think that'll work but it'll be confusing because AutoMLSearch will only update the parameters for the first drop columns transformer.

freddyaboulton: So the second one won't "do" anything I believe

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -64,8 +64,8 @@ def _get_test_data_from_configuration(
                     "abalone_0@gmail.com",
                     "AbaloneRings@yahoo.com",
                     "abalone_2@abalone.com",
-                    "$titanic_data%&@hotmail.com",
-                    "foo*EMAIL@email.org",
+                    "titanic_data@hotmail.com",
```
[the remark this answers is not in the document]
davesque: @chukarsten Yeah, never seen email addresses like that before :).  I think it's safe to delete them from test data to accommodate the woodwork update.

evalml/tests/component_tests/test_lsa.py
```diff
@@ -72,15 +72,17 @@ def test_some_missing_col_names(text_df, caplog):
     }
 
 
-def test_lsa_empty_text_column():
-    X = pd.DataFrame({"col_1": []})
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
ParthivNaresh: Sorry if I'm confused about this but is this related to behaviour as it pertains to an empty column or if there aren't any text columns?

evalml/data_checks/invalid_targets_data_check.py
```diff
@@ -80,7 +80,8 @@ def validate(self, X, y):
 
         y = infer_feature_types(y)
         is_supported_type = y.ww.logical_type.type_string in numeric_and_boolean_ww + [
-            ww.logical_types.Categorical.type_string
+            ww.logical_types.Categorical.type_string,
+            ww.logical_types.Unknown.type_string,
```
bchen1116, line 84: Why add `unknown` type as a supported type for `y`? Seems like if it's a classification problem, the type would be guessed as `categorical` if it was string. Are there cases where it could be `unknown`?

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -76,7 +76,10 @@ def fit(self, X, y):
         """
         if y is None:
             return self
-        y = infer_feature_types(y).to_frame()
+        y = infer_feature_types(y)
```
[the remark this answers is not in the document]
bchen1116: nit, but couldn't you do
```python
y = infer_feature_types(y).to_frame()
if all(y.isnull()):
        raise TypeError("Provided target full of nulls.")
```
just to shorten/simplify slightly?

evalml/pipelines/utils.py
bchen1116: Agreed, I don't think we should have 2 of the same components if we don't need/use both

angela97lin: Agreed--the code in AutoML search here means that we should be covered without needing two DropColumn components, right?

This makes me question if there was a way to pass component instances instead of component classes along, s.t. we could have a DropColumn(index) and DropColumn(unknown_cols) component passed to AutoML rather than having automl handle the parameters to initialize separately. But OOS for this discussion I guess 🤷‍♀️ 


```
            index_and_unknown_columns = list(
                self.X_train.ww.select(["index", "unknown"], return_schema=True).columns
            )
            unknown_columns = list(
                self.X_train.ww.select("unknown", return_schema=True).columns
            )
            index_and_unknown_columns = _put_into_original_order(
                self.X_train, index_and_unknown_columns
            )
            if len(index_and_unknown_columns) > 0 and drop_columns is None:
                parameters["Drop Columns Transformer"] = {
                    "columns": index_and_unknown_columns
                }
                if len(unknown_columns):
                    logger.info(
                        f"Removing columns {unknown_columns} because they are of 'Unknown' type"
 ```

**Review by angela97lin — APPROVED** [review]
I think this is good to go, thanks Karsten! Agreed w/ Freddy on the double DropColumn, I think we're set on AutoML side but it's maybe worth discussing outside this context why we have AutoML do this in the first place :P

**Review by bchen1116 — CHANGES_REQUESTED** [review]
Looking good! I left a few questions/nits, but agreed with @freddyaboulton that we should try not to append 2 `DropColumn` components to the pipeline, especially since it's likely we only set 1 of them.

File: evalml/utils/woodwork_utils.py [hunk 7]
```diff
@@ -82,9 +82,20 @@ def infer_feature_types(data, feature_types=None):
...
 
     if isinstance(data, pd.Series):
+        if all([isinstance(x, type(pd.NA)) for x in data]):
+            data = data.replace(pd.NA, np.nan)
+            feature_types = "Double"
         return ww.init_series(data, logical_type=feature_types)
     else:
         ww_data = data.copy()
+        # Revert the inference of all nulls to the unknown type and change it back to double.
+        all_null_cols = ww_data.columns[ww_data.isnull().all(0)]
```
freddyaboulton, line 92: I don't think this will work if ww is initialized before being passed into one of our components?

image

I think this might also cause a problem with partial dependence but I have not verified. Down to talk about it after stand-up!

freddyaboulton (changes requested):

evalml/pipelines/utils.py
chukarsten: Should be good to go now.  I think I misunderstood what was happening!

evalml/utils/woodwork_utils.py
chukarsten: I handled that specific case.  Feel free to check it out.  I was considering maybe adding a similar test for the target imputer with the y series being pre-inited.

evalml/data_checks/invalid_targets_data_check.py
chukarsten: you're right, I took the wrong tack on that.

evalml/utils/woodwork_utils.py
```diff
@@ -79,14 +97,17 @@ def infer_feature_types(data, feature_types=None):
                 ww_error = f"{ww_error}. Please initialize ww with df.ww.init() to get rid of this message."
             raise ValueError(ww_error)
         data.ww.init(schema=data.ww.schema)
-        return data
+        return convert_all_nan_unknown_to_double(data)
 
     if isinstance(data, pd.Series):
+        if all([isinstance(x, type(pd.NA)) for x in data]):
```
freddyaboulton, line 103: Can we use the vectorized form of this, e.g. `(x == pd.NA).all()`, here and in `is_column_pd_na` ?

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -54,12 +54,19 @@ def test_invalid_target_data_check_nan_error():
     assert invalid_targets_check.validate(X, y=pd.Series([np.nan, np.nan, np.nan])) == {
         "warnings": [],
         "errors": [
+            DataCheckError(
```
freddyaboulton, line 57: 👍

evalml/tests/pipeline_tests/test_pipeline_utils.py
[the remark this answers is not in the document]
freddyaboulton: Agreed. I think https://github.com/alteryx/woodwork/issues/1080 will help cover against users passing impossible email values by manually specifying the email type.

freddyaboulton (approved):
Thank you for your work on this @chukarsten !!

evalml/tests/utils_tests/test_woodwork_utils.py
```diff
@@ -257,3 +265,49 @@ def test_ordinal_retains_order_min():
...
+            [None, None, None],
+            [np.nan, np.nan, np.nan],
+            [pd.NA, pd.NA, pd.NA],
+            ["ax23n9ck23l", "1,28&*_%*&&xejc", "xnmvz@@Dcmeods-0"],
+        ],
+        [True, False],
+    ),
+)
+def test_infer_feature_types_NA_to_nan(null_col, already_inited):
+    """A short test to make sure that columnds with all null values
```
bchen1116, line 283: typo: columns

bchen1116 (approved):
Nice! I like the new test! left one comment on a typo, but LGTM

evalml/tests/utils_tests/test_woodwork_utils.py
chukarsten: classic :)

PR #2612, opened 2021-08-10
Clean up `ComponentGraph` implementation

### LINKED ISSUE #2518: Refactor underlying ComponentGraph implementation
`fit_features` (used in TimeSeries pipelines' `fit` method) and `compute_final_component_features` (used in PipelineBase's `compute_estimator_features`) use `_fit_transform_features_helper` while ComponentGraph's `fit` and `predict` uses `_compute_features`, which `_fit_transform_features_helper` also calls. 

`_fit_transform_features_helper` and `_compute_features` have a suspicious amount of overlapping code. I have a feeling we can refactor this so that we don't have two large codepaths to deal with.

angela97lin:
Closes #2518

I'm still skeptical about `fit_features` and `compute_final_component_features` being useful but at this point, they're basically just the public facing API for our pipelines to use `_fit_transform_features_helper`. 🤷‍♀️

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -221,6 +221,28 @@ class ComponentGraph:
             )
         return x_inputs
 
+    def consolidate_inputs(self, component_outputs, component, X, y=None):
+        x_inputs = []
+        y_input = None
+        for parent_input in self.get_inputs(component):
+            if parent_input.endswith(".y"):
+                y_input = component_outputs[parent_input]
+            elif parent_input == "y":
+                y_input = y
+            if parent_input == "X":
+                x_inputs.append(X)
+            elif parent_input.endswith(".x"):  # must end in .x
+                parent_x = component_outputs[parent_input]
+                if isinstance(parent_x, pd.Series):
+                    parent_x = parent_x.rename(parent_input)
+                x_inputs.append(parent_x)
+        x_inputs = ww.concat_columns(x_inputs)
+        # if needs_fitting:
+        #     self.input_feature_names.update(
+        #         {self.compute_order[-1]: list(x_inputs.columns)}
+        #     )
+        return x_inputs, y_input
+
     def predict(self, X):
         """Make predictions using selected features.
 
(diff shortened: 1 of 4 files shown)


evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -621,7 +621,7 @@ def test_time_series_pipeline_not_fitted_error(
         X, y = X_y_regression
         clf = time_series_regression_pipeline_class(
             parameters={
-                "Linear Regressor": {"n_jobs": 1},
+                "Random Forest Regressor": {"n_jobs": 1},
```
bchen1116, line 624: Why change this to RFR?

evalml/pipelines/component_graph.py
```diff
@@ -205,36 +205,40 @@ def _fit_transform_features_helper(self, needs_fitting, X, y=None):
...
-            parent_input
-            for parent_input in self.get_inputs(self.compute_order[-1])
-            if parent_input[-2:] != ".y"
-        ]
-        for parent in parent_inputs:
-            parent_output = component_outputs.get(
-                parent, component_outputs.get(f"{parent}.x")
-            )
-            if isinstance(parent_output, pd.Series):
-                parent_output = pd.DataFrame(parent_output, columns=[parent])
```
bchen1116, line 220: Do we no longer need to change the series to a dataframe?

bchen1116 (approved):
Nice cleanup! Left some questions, but idt anything blocking.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
ParthivNaresh: @bchen1116 I can't be 100% sure but due to the occasional issues we're running into with LinearRegressor for time series problems, it might be better to default to a more consistent estimator, but I could be wrong.

evalml/pipelines/component_graph.py
angela97lin: I think no, this might have been outdated--we use ww's `concat_columns` and it takes both series and dfs :)

https://woodwork.alteryx.com/en/stable/generated/woodwork.utils.concat_columns.html#woodwork.utils.concat_columns

Issue #2592, opened 2021-08-04
Withholding ARIMA from AutoML if time unit is not supported by sktime


ParthivNaresh:
Certain frequencies are not supported in `sktime` due to inconsistency, such as month (28 vs 30 vs 31 days) and year (365 vs 366 days).

This issue tracks two things:

1. Researching which frequencies are allowed in `sktime`
2. Checking the frequency of the time series dataset and if it isn't supported, then disabling ARIMA from running in AutoML.

freddyaboulton:
@ParthivNaresh I wonder if this would still be a problem if we used a relative forecasting frequency?

https://www.sktime.org/en/latest/examples/01_forecasting.html#step-2---specifying-the-forecasting-horizon

ParthivNaresh:
@freddyaboulton We actually currently leverage a ForecastingHorizon for ARIMA. I think the fix that `sktime` put out doesn't address this issue with ForecastingHorizon yet, but I need to dig deeper.

freddyaboulton:
I guess I'm wondering if the error would go away if we used:

```
fh_ = forecasting_.ForecastingHorizon(dates, is_relative=True)
```

rather than

```
fh_ = forecasting_.ForecastingHorizon(dates, is_relative=False)
```

As far as I understand it, a relative forecasting horizon specifies "forecast the next k points in the series". I think that's in line with our requirements.

PR #2613, opened 2021-08-10
Support relative forecasting for ARIMA

ParthivNaresh:
Fixes: #2592 

The original issue was filed under the assumption that `sktime` would not be able to support certain time units, such as those that had a frequency other than 1, such as 2 days, 4 minutes, 3 months, etc. This manifests when exogenous variables (features) are included in `fit` and `predict`. This problem doesn't seem to persist when the `is_relative` parameter in `ForecastingHorizon` is set to `True`. By setting this to `True`, predictions will be made on a set of absolute indices starting from after the last date used when fitting the estimator, as opposed to being made on specific dates.

```
time_index_1 = pd.date_range("1/1/21", periods=100, freq="3T")
X = pd.DataFrame(range(100), index=time_index_1)
y = pd.Series(y_values, index=time_index_1)
```

With the flag is set to `False`
```
# The actual dates have to be passed in to the ForecastingHorizon for when predictions need to be made
fh_ = ForecastingHorizon(X.index[75:], is_relative=False)
a_clf = AutoARIMA(suppress_warnings=True)
clf = a_clf.fit(y=y[:75])
y_pred_sk = clf.predict(fh=fh_)
print(y_pred_sk)
---------------------------------------------------
2021-03-17    0.371662
2021-03-18    0.776146
...
2021-04-09   -0.486227
2021-04-10   -0.000040
```

This does not work when features are passed
```
fh_ = ForecastingHorizon(X.index[75:], is_relative=False)
a_clf = AutoARIMA(suppress_warnings=True)
clf = a_clf.fit(X=X[:75], y=y[:75])
y_pred_sk = clf.predict(fh=fh_, X=X[75:])
print(y_pred_sk)
---------------------------------------------------
668         X = self._check_exog(X)  # type: np.ndarray
669         if X is not None and X.shape[0] != n_periods:
--> 670             raise ValueError('X array dims (n_rows) != n_periods')
671 
672         # f = self.arima_res_.forecast(steps=n_periods, exog=X)

ValueError: X array dims (n_rows) != n_periods
```

With the flag is set to `True`
```
# Only the "periods" or "absolute indices" after the last date provided when fitting the model are needed
fh_ = ForecastingHorizon([i+1 for i in range(len(X.index[75:]))], is_relative=True)
a_clf = AutoARIMA(suppress_warnings=True)
clf = a_clf.fit(y=y[:75])
y_pred_sk = clf.predict(fh=fh_)
print(y_pred_sk)
---------------------------------------------------
2021-03-17    0.371662
2021-03-18    0.776146
...
2021-04-09   -0.486227
2021-04-10   -0.000040
```

This works even when features are passed
```
fh_ = ForecastingHorizon([i+1 for i in range(len(X.index[75:]))], is_relative=True)
a_clf = AutoARIMA(suppress_warnings=True)
clf = a_clf.fit(X=X[:75], y=y[:75])
y_pred_sk = clf.predict(fh=fh_, X=X[75:])
print(y_pred_sk)
---------------------------------------------------
2021-01-01 03:45:00    0.369964
2021-01-01 03:48:00    0.768921
...
2021-01-01 04:54:00   -0.458856
2021-01-01 04:57:00   -0.012400
```

More information can be found [here](https://www.sktime.org/en/latest/examples/01_forecasting.html#section_1_2_2).

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -143,7 +143,7 @@ class ARIMARegressor(Estimator):
                 f" Found {dates.shape[1]} columns."
             )
         freq = "M" if pd.infer_freq(dates) == "MS" else pd.infer_freq(dates)
-        dates = dates.to_period(freq=freq)
+        dates = pd.DatetimeIndex(dates, freq=freq)
         X, y = self._match_indices(X, y, dates)
         if predict:
             arima_model_msg = (
@@ -152,7 +152,9 @@ class ARIMARegressor(Estimator):
             forecasting_ = import_or_raise(
                 "sktime.forecasting.base", error_msg=arima_model_msg
             )
-            fh_ = forecasting_.ForecastingHorizon([i+1 for i in range(len(dates))], is_relative=True)
+            fh_ = forecasting_.ForecastingHorizon(
+                [i + 1 for i in range(len(dates))], is_relative=True
+            )
             return X, y, fh_
         else:
             return X, y, None
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -152,7 +152,7 @@ class ARIMARegressor(Estimator):
             forecasting_ = import_or_raise(
                 "sktime.forecasting.base", error_msg=arima_model_msg
             )
-            fh_ = forecasting_.ForecastingHorizon(dates, is_relative=False)
+            fh_ = forecasting_.ForecastingHorizon([i+1 for i in range(len(dates))], is_relative=True)
             return X, y, fh_
         else:
             return X, y, None
--- a/evalml/tests/component_tests/test_arima_regressor.py
+++ b/evalml/tests/component_tests/test_arima_regressor.py
@@ -121,7 +121,7 @@ def test_fit_predict_ts_with_only_datetime_column_in_X(
     assert isinstance(X.index, pd.DatetimeIndex)
     assert isinstance(y.index, pd.DatetimeIndex)
 
-    fh_ = forecasting.ForecastingHorizon(y_test.index, is_relative=False)
+    fh_ = forecasting.ForecastingHorizon([i+1 for i in range(len(y_test))], is_relative=True)
 
     a_clf = sktime_arima.AutoARIMA()
     clf = a_clf.fit(y=y)
@@ -144,7 +144,7 @@ def test_fit_predict_ts_with_X_and_y_index_out_of_sample(
     assert isinstance(X.index, pd.DatetimeIndex)
     assert isinstance(y.index, pd.DatetimeIndex)
 
-    fh_ = forecasting.ForecastingHorizon(y_test.index, is_relative=False)
+    fh_ = forecasting.ForecastingHorizon([i+1 for i in range(len(y_test))], is_relative=True)
 
     a_clf = sktime_arima.AutoARIMA()
     clf = a_clf.fit(X=X, y=y)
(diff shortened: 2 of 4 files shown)


evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -361,4 +375,27 @@ def test_fit_predict_date_index_named_out_of_sample(
...
+    datetime_ = pd.date_range("1/1/1870", periods=20, freq=freq_num + freq_str)
+
+    X = pd.DataFrame(range(20), index=datetime_)
+    y = pd.Series(np.sin(np.linspace(-8 * np.pi, 8 * np.pi, 20)), index=datetime_)
+
+    fh_ = forecasting.ForecastingHorizon(
+        [i + 1 for i in range(len(y[15:]))], is_relative=True
+    )
+
+    a_clf = sktime_arima.AutoARIMA(start_p=2, start_q=2, max_p=2, max_q=2)
```
ParthivNaresh, line 393: These parameters reduce this test run time from 1m 14s to 20s

chukarsten: great call!

chukarsten: Do you think it would be the same coverage to just reduce the freq_str to maybe half or so?  I appreciate the thoroughness, but I feel like maybe we can trust sklearn to handle uniformity between the different time periods.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -152,7 +152,9 @@ def _format_dates(self, dates, X, y, predict=False):
             forecasting_ = import_or_raise(
                 "sktime.forecasting.base", error_msg=arima_model_msg
             )
-            fh_ = forecasting_.ForecastingHorizon(dates, is_relative=False)
+            fh_ = forecasting_.ForecastingHorizon(
+                [i + 1 for i in range(len(dates))], is_relative=True
+            )
```
chukarsten, line 157: So, dumb question, it seems that right now relative forecasting is all we support with this change.  Is there any benefit to changing the api by having `is_relative` be assigned during the regressor's creation?   I'm not super familiar, so maybe we only want relative forecasting, but it seems to the luddite like me that "both" might be desirable.

chukarsten (approved):
Nice job, Parthiv!  I'm not 100% sure whether we only want only relative forecasting or we want to enable both relative and absolute.  If the answer is the former, then I think we're good to go.  If the latter, we might need to rework this a little bit.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
ParthivNaresh: Not dumb at all and this is certainly something we can discuss further. Based on the time series conversations that have been going on, we want to make sure that however we implement this, it should work well with the moving window strategy we want to leverage in Rolling Origin Cross Validation. I think especially [considering](https://alteryx.atlassian.net/wiki/spaces/PS/pages/931004419/Draft+Version+1+of+Time+Series+AutoMLSearch+API#In-depth-look-of-pipeline.predict-for-the-test-set) `gap` and `forecast_horizon`, we'd want to look a certain number of steps ahead when making our predicts. Having relative forecasting would make this easier since we can determine how far ahead we want to look, and what those indices should be.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -152,7 +152,9 @@ def _format_dates(self, dates, X, y, predict=False):
             forecasting_ = import_or_raise(
                 "sktime.forecasting.base", error_msg=arima_model_msg
             )
-            fh_ = forecasting_.ForecastingHorizon(dates, is_relative=False)
+            fh_ = forecasting_.ForecastingHorizon(
+                [i + 1 for i in range(len(dates))], is_relative=True
```
bchen1116, line 156: Why are we forecasting on this array of ints rather than the dates? Do the dates become irrelevant when we use `is_relative`?

bchen1116 (approved):
Left a question for my own understanding, but rest of it looks good!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
ParthivNaresh: @bchen1116 Correct, the ints represent how many "periods" after relative to the end of the training dates we want to predict for.

PR #2617, opened 2021-08-10
Only add docs build files if ipython 'startup' profile exists

### LINKED ISSUE #2567: Make it easy to build docs locally
@eccabay had to comment out a few lines in the docs setup.py in order to build the docs locally. It would be great if we had a make target for local docs build and another for RTD docs build.

@eccabay could you please share which lines you had to comment out to get local build to work?

Thanks @freddyaboulton for the suggestion

dsherry:
Fixes #2567 

This will be true on the RTD builder, but false locally. This does mean warnings will show up in the docs for local builds, but I think its better than local docs build simply failing with a path error.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -53,6 +53,8 @@ Note that if you're on Mac, there are a few extra steps you'll want to keep trac
   open build/html/index.html
   ```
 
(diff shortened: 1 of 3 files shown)


bchen1116 (approved):
LGTM!

eccabay (approved):
LGTM, thanks for fixing this!

PR #2622, opened 2021-08-12
Use range-based approach for high variance check in AutoMLSearch

### LINKED ISSUE #2621: Use range-based approach for detecting High Variance
Use the approach detailed [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/957546651/High+Variance+calculations?focusedCommentId=971312633#comment-971312633) to tackle detecting high variance rather than using the coefficient of variation that we currently use.

bchen1116:
fix #2621 
Design doc and discussion [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/957546651/High+Variance+calculations?focusedCommentId=971312633#comment-971312633)


evalml/objectives/standard_metrics.py
```diff
@@ -369,6 +390,7 @@ class LogLossBinary(BinaryClassificationObjective):
     score_needs_proba = True
     perfect_score = 0.0
     is_bounded_like_percentage = False  # Range [0, Inf)
+    expected_range = [0, 1]
```
bchen1116, line 393: Although the range of log loss is (0, inf), we choose to set its expected range much smaller. I chose [0, 1] since we ideally expect the models to output very similar scores throughout the different CV folds. If the scores are too different (in this case, difference >= 0.5), we want to warn users that there could be high variance

evalml/tests/automl_tests/test_automl.py
```diff
@@ -3968,28 +3985,33 @@ def test_automl_pipeline_random_seed(AutoMLTestEnv, random_seed, X_y_multi):
...
+    for cv_scores in [
+        pd.Series([0, 1, np.nan, np.nan]),
+        pd.Series([0, 1, 2, 3]),
+        pd.Series([0, -1, -1, -1]),
+        pd.Series([10, 0, -1, -10]),
+    ]:
+        if objectives == "Log Loss Binary":
+            assert automl._check_for_high_variance(pipeline, cv_scores)
+        else:
+            assert not automl._check_for_high_variance(pipeline, cv_scores)
```
bchen1116, line 4014: This objective shouldn't raise any high variance warnings

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1918,6 +1918,22 @@ class CustomClassificationObjective(BinaryClassificationObjective):
...
+
+class CustomClassificationObjectiveInf(BinaryClassificationObjective):
+    """Accuracy score for binary and multiclass classification."""
+
+    name = "Classification Accuracy"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = False
+    expected_range = [float("-inf"), float("inf")]
```
ParthivNaresh, line 1936: Might be good to check a custom objective that has an `expected_range` which is only bounded on one side like `0, inf` but I'm fine if there isn't one

ParthivNaresh (approved):
Looks great!

evalml/objectives/standard_metrics.py
```diff
@@ -494,6 +522,7 @@ class R2(RegressionObjective):
     score_needs_proba = False
     perfect_score = 1
     is_bounded_like_percentage = False  # Range (-Inf, 1]
+    expected_range = [-1, 1]
```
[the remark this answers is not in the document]
bchen1116: @ParthivNaresh That's a good point! When there are outliers, this will likely raise a high variance warning more often, but I think that makes sense to do as well. In scenarios where each fold can get different enough variation of data where the models perform fairly differently, raising a high variance warning would be beneficial

chukarsten (approved):
This LGTM!  Thanks!  I think I just put up a small comment that's a nit, but no big deal.  Sorry, I reviewed this yesterday but somehow forgot to hit approve.

evalml/objectives/objective_base.py
```diff
@@ -42,6 +42,12 @@ def perfect_score(cls):
     def is_bounded_like_percentage(cls):
         """Returns whether this objective is bounded between 0 and 1, inclusive."""
 
+    @property
+    @classmethod
+    @abstractmethod
+    def expected_range(cls):
+        """Returns the expected range of the objective, which is not necessarily the possible ranges."""
```
eccabay, line 49: Nit: Could you add a little more detail here? If a user is trying to define their own objective, they may not quite understand what this should be

eccabay (approved):
LGTM! My only request would be to add a small section in the [defining custom objectives section of our docs](https://evalml.alteryx.com/en/stable/user_guide/objectives.html) to reflect the addition of the `expected_range`.

Issue #2614, opened 2021-08-10
Refactor time series pipeline inheritance


angela97lin:
``TimeSeriesClassificationPipeline`` inherits from ``ClassificationPipeline`` and ``TimeSeriesRegressionPipeline`` inherits from ``RegressionPipeline``. However, these classes override almost all of the methods in these base classes (predict, score, predict_proba, fit). In addition, ``TimeSeriesClassificationPipeline`` and ``TimeSeriesRegressionPipeline`` have suspiciously similar code for each of these methods.

I think we could refactor these classes s.t. we have a TimeSeriesBasePipeline class which has predict, score, predict_proba, fit as defined in both classes. The classification time series pipelines will have to do some extra encoding, but at least we could share the logic for padding.

If we want to take advantage of any functionality in our normal pipelines, we can use mixins, as we already are doing for `TimeSeriesBinaryClassificationPipeline` via `BinaryClassificationPipelineMixin`.

PR #2625, opened 2021-08-13
Add support for creating pipelines without an estimator as the final component and add ``transform()`` to component graphs and pipelines

### LINKED ISSUE #2103: Add support for creating pipelines without an estimator as the final component
Currently, our pipelines assume that the final component in the component graph is an estimator. We should allow for pipelines that are just a series of transformations (ex: preprocessing pipeline) that do not have an estimator.

This issue tracks adding support for pipelines without an estimator. Right now, `fit` does not error out, but it doesn't make sense to call `predict` or `predict_proba` when we don't have an estimator. Perhaps we should define or allow a `transform` method?

This came up in #2058, where we want to create pipelines from data check actions. These pipelines would only consist of transformers, and no estimators. However, until #2103 is complete, there's not a good way to create these valid preprocessing pipelines based on actions.

angela97lin:
Closes #2103

Design doc here: https://alteryx.atlassian.net/wiki/spaces/PS/pages/541622308/Component+Graph+Updated+Design+and+Notes

- Adds a `transform` method for ComponentGraph and pipelines
   - Similar to `transform` for components, `transform` for pipelines / component graphs will return a tuple if transforming y, otherwise just one df (transformed X)
   - Will raise a ValueError if the final component is an Estimator

- `predict` and `predict_proba` will raise a ValueError if the final component is a Transformer


evalml/pipelines/pipeline_meta.py
```diff
@@ -14,20 +14,23 @@ def check_for_fit(cls, method):
...
 
         @wraps(method)
-        def _check_for_fit(self, X=None, objective=None):
+        def _check_for_fit(self, *args, **kwargs):
             klass = type(self).__name__
             if not self._is_fitted:
                 raise PipelineNotYetFittedError(
                     f"This {klass} is not fitted yet. You must fit {klass} before calling {method.__name__}."
                 )
-            if method.__name__ == "predict_proba":
```
angela97lin, line 23: Because we want to support multiple methods now with different signatures (ex: predict takes X, objective but transform takes X, y), we need to make this more general.

evalml/pipelines/component_graph.py
```diff
@@ -220,32 +220,65 @@ def _consolidate_inputs_for_component(
...
-            elif parent_input == "y":
+            if parent_input == "y":
                 y_input = y
-            if parent_input == "X":
+            elif parent_input == "X":
                 x_inputs.append(X)
-            elif parent_input.endswith(".x"):  # must end in .x
+            elif parent_input.endswith(".y"):
+                y_input = component_outputs[parent_input]
+            elif parent_input.endswith(".x"):
```
angela97lin, line 229: Not related, just cleaning up logic :)

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -83,7 +83,11 @@ def fit(self, X, y):
 
         y_shifted = y.shift(-self.gap)
         X_t, y_shifted = drop_rows_with_nans(X_t, y_shifted)
-        self.estimator.fit(X_t, y_shifted)
+
+        if self.estimator is not None:
+            self.estimator.fit(X_t, y_shifted)
+        else:
+            self.component_graph.get_last_component().fit(X_t, y)
```
angela97lin, line 90: If this pipeline does not have an estimator, this pipeline is just a series of transformations. As such, just using the original, unshifted `y` here

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -75,7 +75,10 @@ def fit(self, X, y):
 
         y_shifted = y.shift(-self.gap)
         X_t, y_shifted = drop_rows_with_nans(X_t, y_shifted)
-        self.estimator.fit(X_t, y_shifted)
+        if self.estimator is not None:
```
angela97lin, line 78: The logic here and in the other time series classification pipelines makes a greater case for https://github.com/alteryx/evalml/issues/2614 :)

evalml/tests/pipeline_tests/classification_pipeline_tests/test_binary_classification.py
```diff
@@ -198,7 +198,7 @@ def test_pipeline_thresholding_errors(
         parameters={"Logistic Regression Classifier": {"n_jobs": 1}}
     )
     pipeline.fit(X, y)
-    pred_proba = pipeline.predict_proba(X, y).iloc[:, 1]
+    pred_proba = pipeline.predict_proba(X).iloc[:, 1]
```
angela97lin, line 201: Not time series, shouldn't have y!

evalml/pipelines/component_graph.py
```diff
@@ -220,32 +220,63 @@ def _consolidate_inputs_for_component(
...
         return x_inputs, y_input
 
+    def transform(self, X, y=None):
+        """Transform the input using the component graph.
+
+        Arguments:
+            X (pd.DataFrame): Input features of shape [n_samples, n_features].
+            y (pd.Series): The target data of length [n_samples]. Defaults to None.
+
+        Returns:
```
chukarsten, line 244: nit: how do we doc string variable returns?  Like separate lines?
```
Returns:
    rtype_1: <desc>
    rtype_2: <desc>
```

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -2104,3 +2094,95 @@ def test_component_graph_compute_final_component_features_with_sampler(
     assert len(mock_estimator_fit.call_args[0][0]) == expected_length
     features_for_estimator = component_graph.compute_final_component_features(X, y)
     assert len(features_for_estimator) == len(y)
+
+
+@patch("evalml.pipelines.components.Imputer.transform")
+@patch("evalml.pipelines.components.OneHotEncoder.transform")
+def test_component_graph_transform(
```
chukarsten, line 2101: What exactly is this testing?  The components are all mocked out to just pass through X.  Wouldn't it actual demonstrate that the new `component_graph.transform()` is working by actually having it do the transformation(s)?

chukarsten (approved):
I love the super thorough testing.  It's very clear and easy to understand.  A few questions left, but nothing blocking.  I do agree that it would be good to get the refactor PR up ASAP after this merges in.  Feel free to bring it up in sprint kickoff.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1,3 +1,4 @@
+import re
```
angela97lin, line 1: something I just thought of to test: adding estimator within cg but not at end and calling `transform` :0

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: My philosophy here was that it doesn't matter what the components individual `transform` methods return, hence why I'm mocking each of the transform. The component graph should just be responsible for and tested for being able to string those along... but maybe there is a danger in keeping the original value, X, though, rather than returning some dummy value. I can change that if it makes it more clear!

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2580,3 +2576,123 @@ def test_get_hyperparameter_ranges():
...
+    X = make_data_type("ww", X)
+    y = make_data_type("ww", y)
+    pipeline_class = _get_pipeline_base_class(problem_type)
+    pipeline = pipeline_class(
+        component_graph={
+            "Imputer": ["Imputer", "X", "y"],
+            "OHE": ["One Hot Encoder", "Imputer.x", "y"],
+        },
+        parameters={
+            "pipeline": {"gap": 0, "max_delay": 0, "date_index": None},
```
ParthivNaresh, line 2602: Do you need these for the problem_types this test covers?

evalml/tests/pipeline_tests/test_component_graph.py
ParthivNaresh: This is interesting, but I was only able to see this being used on tests in which the estimator was the final component?

ParthivNaresh (approved):
Awesome work and great test coverage!

bchen1116 (approved):
LGTM! Have the changes already been made for samplers to have `fit_transform()` be the same as `.fit().transform()`?

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Oh I just added this comment here to remind myself when I looked over this file, has nothing to do with the import 😂 

Importing `re` is just used for matching expected exceptions!

evalml/tests/pipeline_tests/test_component_graph.py
angela97lin: Yup!

angela97lin:
@bchen1116 

> LGTM! Have the changes already been made for samplers to have `fit_transform()` be the same as `.fit().transform()`?

Yup, changes were made in https://github.com/alteryx/evalml/pull/2583 :)

evalml/pipelines/pipeline_meta.py
angela97lin: Interesting question! I think the difficulty is with the signature of `_check_for_fit` not being the most flexible since that's wrapped around all of our methods. I didn't see any downsides of being more general--if the user passes in the wrong number of arguments, the exception thrown is pretty clear--so I think this was mostly an implementation detail and a nice cleanup! Open to hearing concerns though 😅

PR #2626, opened 2021-08-13
Patch pre-Release v0.30.1

chukarsten:
A small update to do a better job of detecting all-pd.NA values.  I guess.

### DIFF
--- a/evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
+++ b/evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
@@ -55,13 +55,6 @@ def test_invalid_target_data_check_nan_error():
         "warnings": [],
         "errors": [
             DataCheckError(
-                message="Target is unsupported Unknown type. Valid Woodwork "
-                "logical types include: integer, double, boolean",
-                data_check_name=invalid_targets_data_check_name,
-                message_code=DataCheckMessageCode.TARGET_UNSUPPORTED_TYPE,
-                details={"unsupported_type": "unknown"},
-            ).to_dict(),
-            DataCheckError(
                 message="Target is either empty or fully null.",
                 data_check_name=invalid_targets_data_check_name,
                 message_code=DataCheckMessageCode.TARGET_IS_EMPTY_OR_FULLY_NULL,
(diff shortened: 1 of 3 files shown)


evalml/utils/woodwork_utils.py
```diff
@@ -66,7 +66,7 @@ def infer_feature_types(data, feature_types=None):
 
     def convert_all_nan_unknown_to_double(data):
         def is_column_pd_na(data, col):
-            return all([isinstance(x, type(pd.NA)) for x in data[col]])
+            return all(data[col].isna())
```
chukarsten, line 69: @freddyaboulton and @bchen1116 I think you recommended both this and `all(data[col] == pd.NA)`.  What happened was I tried `all(data[col] == pd.NA)`, which created weird behavior in as much as a column full of `pd.NA` returns a column full of `pd.NA` when compared to `pd.NA`.  I found a little bit more about this [here](https://github.com/pandas-dev/pandas/issues/38063), where they discuss some of the design decisions about how arithmetic and comparisons using `pd.NA` result in `pd.NA` to "propagate pd.NA".  I ended up just going with what I had and it was super slow - as you foresaw.  I'm now trying to atone for my sins.

evalml/tests/data_checks_tests/test_invalid_targets_data_check.py
```diff
@@ -54,13 +54,6 @@ def test_invalid_target_data_check_nan_error():
     assert invalid_targets_check.validate(X, y=pd.Series([np.nan, np.nan, np.nan])) == {
         "warnings": [],
         "errors": [
-            DataCheckError(
```
chukarsten, line 57: The slightly different logic catches fully null Unknown columns and changes them to np.nan Double columns both in dataframes and in series now.

bchen1116 (approved):
LGTM! Thanks for making the changes. Do we need to update the release note with this PR # (i think you can just tack it on to the end of the ww update PR)

evalml/utils/woodwork_utils.py
dsherry: Great catch.

dsherry (approved):
Well done @chukarsten , I'm so glad we caught this pre-release 👏

Issue #1930, opened 2021-03-04
Build ensembler component


dsherry:
**Background**
We currently wrap sklearn's impl.

**Proposal**
Build our own.

**Advantages over current approach**
* Use whatever metalearner we want
* Have full control over what data the pipelines and the metalearner are computed on
* Caching: avoid recomputing pipeline (or component) predictions during ensembler training
* Feature selection in ensembler

@rpeck FYI

rpeck:
Even more important:

- optimize the way we do the splits for training the ensemble for different sizes of input data (see my text in https://alteryx.quip.com/daHLAQyVIaUC/Addressing-Ensembling-Overfitting)
- optimize scoring on the component pipelines, by doing common feature engineering steps like target encoding, date expansion, NLP processing, etc only once

PR #2629, opened 2021-08-13
Stored `predict_proba` results in `.x` for intermediate estimators in ComponentGraph 

christopherbunn:
As part of the component graph changes necessary to support building our new ensembler (#1930), this PR stores the prediction probabilities (when available) for non-final estimators in a component graph. If `predict_proba` is not available or if the final estimator in a component graph is being evaluated, then `predict` is used.

### DIFF
--- a/evalml/tests/pipeline_tests/test_component_graph.py
+++ b/evalml/tests/pipeline_tests/test_component_graph.py
@@ -542,16 +542,19 @@ def test_compute_final_component_features(
     X, y = X_y_binary
     mock_imputer.return_value = pd.DataFrame(X)
     mock_ohe.return_value = pd.DataFrame(X)
-    mock_en_predict_proba.return_value = pd.DataFrame((
-        {0: np.zeros(X.shape[0]), 1: np.ones(X.shape[0])}
-    ))
+    mock_en_predict_proba.return_value = pd.DataFrame(
+        ({0: np.zeros(X.shape[0]), 1: np.ones(X.shape[0])})
+    )
     mock_en_predict_proba.return_value.ww.init()
-    mock_rf_predict_proba.return_value = pd.DataFrame((
-        {0: np.ones(X.shape[0]), 1: np.zeros(X.shape[0])}
-    ))
+    mock_rf_predict_proba.return_value = pd.DataFrame(
+        ({0: np.ones(X.shape[0]), 1: np.zeros(X.shape[0])})
+    )
     mock_rf_predict_proba.return_value.ww.init()
     X_expected = pd.DataFrame(
-        {"1_Random Forest.x": np.zeros(X.shape[0]), "1_Elastic Net.x": np.ones(X.shape[0])}
+        {
+            "Col 1 Random Forest.x": np.zeros(X.shape[0]),
+            "Col 1 Elastic Net.x": np.ones(X.shape[0]),
+        }
     )
     component_graph = ComponentGraph(example_graph).instantiate({})
     component_graph.fit(X, y)
@@ -825,8 +828,8 @@ def test_input_feature_names(example_graph):
         "column_1_c",
     ]
     assert input_feature_names["Logistic Regression"] == [
-        "1_Random Forest.x",
-        "1_Elastic Net.x",
+        "Col 1 Random Forest.x",
+        "Col 1 Elastic Net.x",
     ]
 
 
@@ -893,8 +896,8 @@ def test_custom_input_feature_types(example_graph):
         "column_2_5",
     ]
     assert input_feature_names["Logistic Regression"] == [
-        "1_Random Forest.x",
-        "1_Elastic Net.x",
+        "Col 1 Random Forest.x",
+        "Col 1 Elastic Net.x",
     ]
 
 
@@ -1063,8 +1066,8 @@ def test_component_graph_dataset_with_different_types():
             )
         )
         assert input_feature_names["Logistic Regression"] == [
-            "1_Random Forest.x",
-            "1_Elastic Net.x",
+            "Col 1 Random Forest.x",
+            "Col 1 Elastic Net.x",
         ]
 
     check_feature_names(component_graph.input_feature_names)
--- a/evalml/tests/pipeline_tests/test_component_graph.py
+++ b/evalml/tests/pipeline_tests/test_component_graph.py
@@ -12,7 +12,11 @@ from pandas.testing import (
 )
 from woodwork.logical_types import Double, Integer
 
-from evalml.exceptions import MissingComponentError, ParameterNotUsedWarning
+from evalml.exceptions import (
+    MethodPropertyNotFoundError,
+    MissingComponentError,
+    ParameterNotUsedWarning,
+)
 from evalml.pipelines import ComponentGraph
 from evalml.pipelines.components import (
     DateTimeFeaturizer,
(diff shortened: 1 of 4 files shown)


evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -458,36 +458,43 @@ def test_component_graph_fit_features(
     mock_X_t = pd.DataFrame(np.ones(pd.DataFrame(X).shape))
     mock_fit_transform.return_value = mock_X_t
     mock_fit.return_value = Estimator
-    mock_predict.return_value = pd.Series(y)
+    mock_predict_proba.return_value = pd.Series(y)
```
angela97lin, line 461: Should we make this a DataFrame instead to mimic and make sure that when we have an output with more than one column?

angela97lin: (oops, pressed add single comment too quickly!)

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -870,8 +874,23 @@ def test_compute_estimator_features_nonlinear(
...
+    mock_en_predict_proba_df = pd.DataFrame(
+        {0: np.zeros(X.shape[0]), 1: np.ones(X.shape[0])}
+    )
+    mock_en_predict_proba_df.ww.init()
+    mock_en_predict_proba.return_value = mock_en_predict_proba_df
+    mock_rf_predict_proba_df = pd.DataFrame(
+        {0: np.ones(X.shape[0]), 1: np.zeros(X.shape[0])}
+    )
+    mock_rf_predict_proba_df.ww.init()
+    mock_rf_predict_proba.return_value = mock_rf_predict_proba_df
```
angela97lin, line 887: This is the same as `mock_en_predict_proba_df`, right? Do we need to initialize two separate dfs or can we just use the same one?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -406,19 +406,19 @@ def test_get_last_component(example_graph):
...
-@patch("evalml.pipelines.components.Estimator.predict")
+@patch("evalml.pipelines.components.Estimator.predict_proba")
 def test_fit_component_graph(
-    mock_predict, mock_fit, mock_fit_transform, example_graph, X_y_binary
+    mock_predict_proba, mock_fit, mock_fit_transform, example_graph, X_y_binary
 ):
     X, y = X_y_binary
     mock_fit_transform.return_value = pd.DataFrame(X)
-    mock_predict.return_value = pd.Series(y)
+    mock_predict_proba.return_value = pd.Series(y)
```
angela97lin, line 415: Same comment as below, let's change this to a DataFrame?

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -511,22 +518,28 @@ def test_predict_repeat_estimator(mock_predict, mock_fit, X_y_binary):
...
+    mock_imputer,
+    example_graph,
+    X_y_binary,
 ):
     X, y = X_y_binary
     mock_imputer.return_value = pd.DataFrame(X)
     mock_ohe.return_value = pd.DataFrame(X)
-    mock_en_predict.return_value = pd.Series(np.ones(X.shape[0]))
-    mock_rf_predict.return_value = pd.Series(np.zeros(X.shape[0]))
+    mock_en_predict_proba.return_value = pd.Series(np.ones(X.shape[0]))
```
angela97lin, line 541: Same comment, let's return dfs!

angela97lin (changes requested):
LGTM! 😁 

I had a suggestion: we should convert our `mock_predict_proba` to a Dataframe to closer mimic real output. I commented directly on some but not all the places 😂 

I'm curious if we have tests for calling `_compute_features` for a `RegressionPipeline` (via fit/predict/compute_features/or other public methods) and for `MulticlassClassificationPipeline` (since predict_proba would not drop a column). 

Are there tests that already exist (and hence that test not erroring on this PR is confirmation) for multiclass/regression? Otherwise, it could be good to write some :)

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -802,8 +815,8 @@ def test_input_feature_names(example_graph):
         "column_1_c",
     ]
     assert input_feature_names["Logistic Regression"] == [
-        "Random Forest.x",
-        "Elastic Net.x",
+        "1_Random Forest.x",
```
chukarsten, line 818: Do we perhaps have any other thoughts for options to name this?  The prepending with the underscore is kinda weird, especially with the space after it.  I think an appen....sion? might look better and adding like a "col" or something to it might be a bit clearer.

chukarsten (approved):
I think this looks good and addressing Angela's concerns and perhaps changing the naming scheme might be a good move.

evalml/tests/pipeline_tests/test_component_graph.py
christopherbunn: I chose to do `column_name_component` scheme since to my knowledge we aren't using underscores anywhere else for the pipeline name. However, in general I don't really have any strong preferences for what the naming convention should be and I'm definitely open to ideas. Your idea of something like [`Col 0 Random Forest.x`, `Col 1 Random Forest.x`, etc.] makes sense to me and I'm good to move forward with this if there aren't any other suggestions.

evalml/tests/pipeline_tests/test_pipelines.py
christopherbunn: They're slightly different, `en` is a column of all zeros and then a column of all ones, whereas `rf` is a column of all ones then a column of all zeros. It's more to make sure that we're choosing the correct columns and appending values correctly.

christopherbunn:
> LGTM! 😁
> 
> I had a suggestion: we should convert our `mock_predict_proba` to a Dataframe to closer mimic real output. I commented directly on some but not all the places 😂

Thanks! I covered most of these test cases, lmk if I missed some but we should be using dfs as the output for intermediate estimators in tests now.


> I'm curious if we have tests for calling `_compute_features` for a `RegressionPipeline` (via fit/predict/compute_features/or other public methods) and for `MulticlassClassificationPipeline` (since predict_proba would not drop a column).
> 
> Are there tests that already exist (and hence that test not erroring on this PR is confirmation) for multiclass/regression? Otherwise, it could be good to write some :)

For `RegressionPipeline`, there's some coverage with tests that use the `nonlinear_regression_pipeline_class` fixture (like `test_score_nonlinear_regression`) since they use the `example_regression_graph` that calculates `predict_proba` for intermediate estimators. This the same with `MulticlassClassificationPipeline`, `nonlinear_multiclass_pipeline_class`, and `test_score_nonlinear_multiclass`. I think the `score()` test case covers the fit/predict test case for these pipeline types, but lmk if you think it would be worth having a more explicit case for these.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -458,36 +459,46 @@ def test_component_graph_fit_features(
...
+    assert mock_predict_proba.call_count == 2
 
 
 @patch("evalml.pipelines.components.Estimator.fit")
+@patch("evalml.pipelines.components.Estimator.predict_proba")
 @patch("evalml.pipelines.components.Estimator.predict")
-def test_predict(mock_predict, mock_fit, example_graph, X_y_binary):
+def test_predict(mock_predict, mock_predict_proba, mock_fit, example_graph, X_y_binary):
     X, y = X_y_binary
+    mock_predict_proba.return_value = pd.DataFrame(y)
```
angela97lin, line 477: Bahaha RE comment to change output to dataframes, I'm curious about when it's a df and has multiple columns! but not a big change :)

angela97lin (approved):
LGTM, thanks for making the changes!

RE the output as dataframes, it'd be nice if they had multiple cols like the output of predict_proba usually would, but not a blocking change 😁

evalml/tests/pipeline_tests/test_component_graph.py
christopherbunn: Oops I see 😅. I added the two test cases with different `predict_proba` outputs to address this.

PR #2632, opened 2021-08-16
Drop ARIMA when `month_start` freq is detected

### LINKED ISSUE #2631: ARIMA doesn't support MS as a frequency
Pandas doesn't support `MS` (month beginning) as a frequency that can be translated into a `PeriodIndex`. Sktime attempts to call `_coerce_to_period` in order to turn the datetime information into a `PeriodIndex`, raising this `ValueError`.

```
a = pd.date_range("1/11/21", periods=10, freq="MS")
pd.PeriodIndex(a)
------------------------------------------
1079 
1080     freq = Period._maybe_convert_freq(freq)
-> 1081     base = freq._period_dtype_code
1082     return c_dt64arr_to_periodarr(data.view("i8"), base, tz), freq
1083 

AttributeError: 'pandas._libs.tslibs.offsets.MonthBegin' object has no attribute '_period_dtype_code'
```

If `MS` is inferred as a frequency in the data then we won't include `ArimaRegressor` during `search`.

ParthivNaresh:
Fixes #2631

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -511,13 +511,14 @@ class AutoMLSearch:
             )
             if (
                 is_time_series(self.problem_type)
-                and pd.infer_freq(X_train[parameters["pipeline"]["date_index"]]) == "MS"
+                and parameters["pipeline"]["date_index"]
             ):
-                allowed_estimators = [
-                    estimator
-                    for estimator in allowed_estimators
-                    if estimator.name != "ARIMA Regressor"
-                ]
+                if pd.infer_freq(X_train[parameters["pipeline"]["date_index"]]) == "MS":
+                    allowed_estimators = [
+                        estimator
+                        for estimator in allowed_estimators
+                        if estimator.name != "ARIMA Regressor"
+                    ]
             logger.debug(
                 f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}"
             )
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -509,6 +509,15 @@ class AutoMLSearch:
             allowed_estimators = get_estimators(
                 self.problem_type, self.allowed_model_families
             )
+            if (
+                is_time_series(self.problem_type)
+                and pd.infer_freq(X_train[parameters["pipeline"]["date_index"]]) == "MS"
+            ):
+                allowed_estimators = [
+                    estimator
+                    for estimator in allowed_estimators
+                    if estimator.name != "ARIMA Regressor"
+                ]
             logger.debug(
                 f"allowed_estimators set to {[estimator.name for estimator in allowed_estimators]}"
             )
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -3417,8 +3417,9 @@ def test_automl_woodwork_user_types_preserved(
             )
 
 
-def test_automl_validates_problem_configuration(X_y_binary, ts_data):
-    X, y = X_y_binary
+def test_automl_validates_problem_configuration(ts_data):
+    _, y = ts_data
+    X = pd.DataFrame(pd.date_range("2020-10-01", "2020-10-31"), columns=["Date"])
     assert (
         AutoMLSearch(X_train=X, y_train=y, problem_type="binary").problem_configuration
         == {}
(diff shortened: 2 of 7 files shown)


evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -379,7 +379,7 @@ def test_fit_predict_date_index_named_out_of_sample(
 
 
 @pytest.mark.parametrize("freq_num", ["1", "2"])
-@pytest.mark.parametrize("freq_str", ["S", "T", "H", "D", "M", "Y"])
+@pytest.mark.parametrize("freq_str", ["T", "M", "Y"])
```
ParthivNaresh, line 382: Sneaking this in here because it didn't go through in my last ARIMA PR

evalml/tests/automl_tests/test_automl.py
```diff
@@ -3452,7 +3452,8 @@ def test_automl_validates_problem_configuration(X_y_binary):
             problem_type="time series regression",
             problem_configuration={"max_delay": 2, "gap": 3},
         )
-
+    _, y = ts_data
+    X = pd.DataFrame(pd.date_range("2020-10-01", "2020-10-31"), columns=["Date"])
```
bchen1116, line 3456: Why did we need to add this to the test here?

bchen1116 (approved):
LGTM! Just left a question about one of the tests

freddyaboulton (commented):
@ParthivNaresh When I run this repro for Arima with MS frequency on main I get a value error but it's coming from our own arima implementation.

```
import pandas as pd

X, y = load_diabetes()
X.ww["Date"] = pd.Series(pd.date_range(start="1/1/2018", periods=X.shape[0], freq="MS"))
arima = ARIMARegressor(date_index="Date")
arima.fit(X, y)
```

When I delete the line that sets the frequency to `M` for `MS`, i.e. `freq = pd.infer_freq(dates)`, the above repro doesn't throw an error anymore and the arima is able to fit:

image


Unit tests also pass. So I'm wondering whether we need to make the changes to AutoMLSearch?

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: Originally I had added it because the code I included in `AutoMLSearch` would throw an error if the data didn't have a `date_index` column specified, which `X_y_binary` didn't have. Decided to just clean up the test and replace the data with what I added, thanks!

ParthivNaresh:
@freddyaboulton  If that line is changed to `freq = pd.infer_freq(dates)`  the error just gets pushed back to `predict`.
```
import pandas as pd
import numpy as np
X = pd.DataFrame()
y = pd.Series(np.random.randint(1, 10, 10))
X["Date"] = pd.Series(pd.date_range(start="1/1/2018", periods=10, freq="MS"))

arima = ARIMARegressor(date_index="Date")
arima.fit(X, y)
arima.predict(X)
------------------------------------------
   481         msg = str(e)
    482         if "Invalid frequency" in msg or "_period_dtype_code" in msg:
--> 483             raise ValueError(
    484                 "Invalid frequency. Please select a frequency that can "
    485                 "be converted to a regular `pd.PeriodIndex`. For other "

ValueError: Invalid frequency. Please select a frequency that can be converted to a regular `pd.PeriodIndex`. For other frequencies, basic arithmetic operation to compute durations currently do not work reliably.
```

evalml/automl/automl_search.py
```diff
@@ -505,6 +505,16 @@ def __init__(
             allowed_estimators = get_estimators(
                 self.problem_type, self.allowed_model_families
             )
+            if (
```
freddyaboulton, line 508: I'm on the fence as to whether this should live here or in `get_estimators`. I'm ok keeping this here. Is this a bug that sktime can fix in the future or is it a limitation of the `MS` frequency?

freddyaboulton (approved):
Thank you @ParthivNaresh !

evalml/automl/automl_search.py
ParthivNaresh: The only reason I chose to put it here instead of `get_estimators` was to avoid passing `problem_configuration` to `get_estimators` just for this check

freddyaboulton: Yea let's not change the api of `get_estimators` yet! Ok, I think there will be other frequencies with that limitation so I will file an issue for following up about that. Might make sense to add yet another data check for this hehe

PR #2637, opened 2021-08-16
Use woodwork outlier detection

### LINKED ISSUE #2529: Update `OutlierDataCheck` to use Woodwork's implementation of outlier detection
https://github.com/alteryx/woodwork/pull/1048 has been merged, so once Woodwork gets a release out, we should update EvalML to use Woodwork's outlier detection implementation :)

bchen1116:
fix #2529

### DIFF
--- a/evalml/data_checks/outliers_data_check.py
+++ b/evalml/data_checks/outliers_data_check.py
@@ -1,6 +1,4 @@
-import numpy as np
-from scipy.stats import gamma
-
+import woodwork as ww
 from evalml.data_checks import (
     DataCheck,
     DataCheckMessageCode,
@@ -42,17 +40,18 @@ class OutliersDataCheck(DataCheck):
 
         X = infer_feature_types(X)
         X = X.ww.select("numeric")
+        X.ww.drop(X.columns[X.isna().all()].tolist(), inplace=True)
 
         if len(X.columns) == 0:
             return results
 
         has_outliers = []
         for col in X.columns:
-            outlier_results = OutliersDataCheck._outlier_score(X[col], False)
-            if (
-                outlier_results is not None and outlier_results["score"] <= 0.9
-            ):  # 0.9 is threshold indicating data needs improvement
+            col_series = ww.init_series(X[col])
+            box_plot_dict = col_series.ww.box_plot_dict()
+            if len(box_plot_dict['low_values']) or len(box_plot_dict['high_values']):
                 has_outliers.append(col)
+
         warning_msg = "Column(s) {} are likely to have outlier data.".format(
             ", ".join([f"'{col}'" for col in has_outliers])
         )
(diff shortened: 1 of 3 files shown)


evalml/data_checks/outliers_data_check.py
```diff
@@ -42,17 +41,18 @@ def validate(self, X, y=None):
 
         X = infer_feature_types(X)
         X = X.ww.select("numeric")
+        X.ww.drop(X.columns[X.isna().all()].tolist(), inplace=True)
```
tamargrey, line 44: Is this necessary because `box_plot_dict` was causing issues with fully null columns?

If so, there may be a bug in Woodwork, because I would expect outlier detection be able to [handle fully null columns](https://github.com/alteryx/woodwork/blob/main/woodwork/statistics_utils.py#L389-L402). There wouldn't be any low or high values, so the check on line 53 would not add them to `has_outliers`

evalml/data_checks/outliers_data_check.py
```diff
@@ -42,17 +41,18 @@ def validate(self, X, y=None):
...
 
         if len(X.columns) == 0:
             return results
 
         has_outliers = []
         for col in X.columns:
-            outlier_results = OutliersDataCheck._outlier_score(X[col], False)
-            if (
-                outlier_results is not None and outlier_results["score"] <= 0.9
```
tamargrey, line 53: Woodwork doesn't perform this kind of outlier score calculation, but it does return the necessary information to calculate that, if you still only want to add the column when it's below `0.9`

```python
pct_outliers = (len(box_plot_dict["low_values"]) + len(box_plot_dict["high_values"])) / len(column_nonan)

num_records = len(column_nonan)
score = OutliersDataCheck._no_outlier_prob(num_records, pct_outliers)
```

evalml/data_checks/outliers_data_check.py
```diff
@@ -42,17 +41,18 @@ def validate(self, X, y=None):
...
-            ):  # 0.9 is threshold indicating data needs improvement
+            col_series = ww.init_series(X[col])
```
tamargrey, line 51: You can get this series with less overhead by doing `col_series = X.ww[col]`. It'll retain all the typing information that's already present, avoiding type inference

evalml/data_checks/outliers_data_check.py
bchen1116: Yep, when I pass in `nan` data otherwise, I get `TypeError: Cannot calculate box plot statistics for non-numeric column` as the error

tamargrey: Ah, this is because we now infer fully null columns as the `Unknown` logical type. 

I believe that the `init_series` call reinferring the logical type as `Unknown` is causing this at the `box_plot_dict` call. Switching to `X.ww[col]` should fix that, since the logical type will remain whatever numeric logical type it is in the DataFrame

evalml/data_checks/outliers_data_check.py
bchen1116: Thanks!

evalml/data_checks/outliers_data_check.py
bchen1116: Ah, in our incorporation of ww 0.5.1, we changed the [nan/null inference](https://github.com/alteryx/evalml/blob/948e3b44dc69f2bd27add1ab9ed24256e2970ea8/evalml/utils/woodwork_utils.py#L67) to be type `Double` rather than `Unknown` in order to handle how we build our components and pipelines. Using `X.ww[col]` caused it to become a Series object rather than a ww DataColumn obj, which then raised the `AttributeError: 'Series' object has no attribute 'box_plot_dict'` error here

tamargrey: Not sure I'm following.

Both `ww.init_series(X[col])` and `X.ww[col]` should return a Series that has Woodwork initialized on it. If Woodwork was not initialized on a Series and you did `col_series.ww.box_plot_dict()`, the error would be `WoodworkNotInitError: Woodwork not initialized for this Series. Initialize by calling Series.ww.init`, and if the attribute did not exist in Woodwork, it would look more like `AttributeError: Woodwork has no attribute 'doesnt_exist'`.

The attribute error you're showing seems more like what you get if you try and access a Woodwork attribute not through the Woodwork accessor (so `df.box_plot_dict()` gets this exact error, when `df.ww.box_plot_dict()` would work)

bchen1116: Ah yeah, I was able to get it to work! It was a mistake on my end. Does the implementation look better to you?

tamargrey: Looks great! Love the chaining with `X.ww[col].ww.box_plot_dict()`.

bchen1116: Thanks for the feedback and help!

tamargrey: So happy to help! And I'd love to hear if you have any feedback on if there's anything we can be doing to make Woodwork's `ww` usage a little easier to understand.

chukarsten (approved):
Nice job, Bryan!  This looks good.  Its kind of cool that it's relatively simple to swap outlier detection like this!

angela97lin (approved):
Nice, I like how clean this is now. Great work, LGTM!

Issue #2587, opened 2021-08-04
Custom Ensembler Component


christopherbunn:
Currently, we are using a wrapped sklearn component to perform ensembling. 

This epic tracks the work items needed to implement our own ensembling component. This will allow a comparable CV score and enhanced compatibility with our component graph. 

[Design doc](https://alteryx.atlassian.net/l/c/KJ53M22F)

christopherbunn:
Current ensembler deprecation plan:
- Change current ensembler implementation's name to`SklearnStackedEnsembler`
- Create new custom ensembler with the name `StackedEnsembler`
- Change default ensembler in AutoML search from `SklearnStackedEnsembler` to new `StackedEnsembler`
- Do deprecation path for `SklearnStackedEnsembler`

PR #2643, opened 2021-08-16
Added ability to utilize instantiated components when creating a pipeline

christopherbunn:
As part of #2587, we will need to be able to create new pipelines using already trained components. When creating new pipelines ensemble pipelines, the AutoML algorithm will be choosing from the best previously trained pipelines to create one large ensembling pipeline. Rather than creating an entirely new ensembling pipeline with the best parameters and retraining the entire pipeline, we can just utilize the pre-trained components and train only the final estimator. 

This PR adds support for using already instantiated components when creating a new pipeline. When an instantiated component is passed through a dictionary, the existing object is used rather than the current behavior of raising an error.

### DIFF
--- a/evalml/tests/pipeline_tests/test_component_graph.py
+++ b/evalml/tests/pipeline_tests/test_component_graph.py
@@ -143,6 +143,18 @@ def test_init_str_components():
     assert comp_graph.compute_order == expected_order
 
 
+def test_init_instantiated():
+    graph = {
+        "Imputer": [
+            Imputer(numeric_impute_strategy="constant", numeric_fill_value=0),
+            "X",
+            "y",
+        ]
+    }
+    cg = ComponentGraph(graph)
+    assert graph['Imputer'][0] == cg.get_component('Imputer')
+
+
 def test_invalid_init():
     invalid_graph = {"Imputer": [Imputer, "X", "y"], "OHE": OneHotEncoder}
     with pytest.raises(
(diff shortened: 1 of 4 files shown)


chukarsten (approved):
This looks good to me!  Should be a pretty solid upgrade to performance.  Do you have perf tests to see what the time to fit consequences are for ensembling pipelines?  This would be an AMAZING demo.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -139,6 +139,19 @@ def test_init_str_components():
...
+    graph = {
+        "Imputer": [
+            Imputer(numeric_impute_strategy="constant", numeric_fill_value=0),
+            "X",
+            "y",
+        ]
+    }
+    cg = ComponentGraph(graph)
+    cg.instantiate({"Imputer": {}})
+    assert graph["Imputer"][0] == cg.get_component("Imputer")
```
angela97lin, line 152: Do we have a test that the component instance's set parameters are still kept in the component graph / pipeline?

angela97lin (approved):
I think this looks really good! Left one testing suggestion, to make sure that on the CG/pipeline level, all of the parameters are as we expect but 🚢

evalml/tests/pipeline_tests/test_component_graph.py
christopherbunn: Updated test to make sure that existing parameters are not changed during instantiation.

Issue #2588, opened 2021-08-04
Prophet: speed up CI


freddyaboulton:
With the merge of Prophet in #2242, cmdstanpy will install and compile stan for the automl and "other" test jobs. This can add about 6 minutes to the installation time. 

Is it worth it to separate prophet tests into their own job?

If we only install stan for the tests that need it, the run time of the prophet tests will be 6-8 minutes, which is on par with the run time of the automl tests before the Prophet merge. 

This would mean there would not be a noticeable increase in the time to go through a full CI cycle.

Of course there could be downsides, like increasing the total number of jobs and hitting the limit of number of concurrent github actions.

The desired outcome of this issue is to investigate whether it's worth it to split the Prophet tests into their job and doing the split if so.

angela97lin:
I like this idea a lot! I don't think it'd be too different from the pattern we've already adopted with dask tests, even if the reason for separating out the tests are different here.

dsherry:
Next steps
* Parthiv will look into cmdstan-builder and see if that can speed up our CI overall
* If not, we should look into splitting prophet tests into a separate job to avoid slowing down the other jobs and overall CI time

PR #2644, opened 2021-08-17
Reduce CI Time by separating Prophet tests

ParthivNaresh:
Fixes: #2588 

This will bring `git-test-other` down from 19/20 min to 5/6 min. In anticipation of integrating Prophet into AutoML, running `git-test-automl` means that many of our tests that rely on instantiating every component/estimator will still need to instantiate Prophet, so I don't think removing it from `git-test-automl` is a good idea.


docs/source/release_notes.rst
```diff
@@ -8,6 +8,7 @@ Release Notes
         * Deleted ``_put_into_original_order`` helper function :pr:`2639`
     * Documentation Changes
     * Testing Changes
+        * Speed up CI time by splitting Prophet tests into a separate workflow in GitHUB :pr:`2644`
```
gsheni, line 11:
```suggestion
        * Speed up CI time by splitting Prophet tests into a separate workflow in GitHub :pr:`2644`
```

Makefile
```diff
@@ -59,7 +59,11 @@ git-test-other-core:
...
+	pytest evalml/tests --ignore evalml/tests/automl_tests/ --ignore evalml/tests/tuner_tests/ --ignore evalml/tests/model_understanding_tests/ --ignore evalml/tests/pipeline_tests/ --ignore evalml/tests/objective_tests/ --ignore evalml/tests/utils_tests/ --ignore evalml/tests/component_tests/test_prophet_regressor.py --ignore evalml/tests/component_tests/test_components.py --ignore evalml/tests/component_tests/test_utils.py -n 2 --durations 0 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-other-junit.xml --doctest-continue-on-failure
+
+.PHONY: git-test-prophet
+git-test-prophet:
+	pytest evalml/tests/component_tests/test_prophet_regressor.py evalml/tests/component_tests/test_components.py evalml/tests/component_tests/test_utils.py evalml/tests/pipeline_tests/ evalml/tests/objective_tests/ evalml/tests/utils_tests/ -n 2 --durations 0 --timeout 300 --doctest-modules --cov=evalml --junitxml=test-reports/git-test-prophet-junit.xml --doctest-continue-on-failure
```
ParthivNaresh, line 66: Due to how many of our tests are constructed, we end up instantiating every estimator including Prophet, so I had to break up `git-test-other` to focus on the files that end up requiring `Prophet`. This is also the reason that I'm leaving `git-test-automl` as it is.

chukarsten (approved):
This looks good to merge to me

bchen1116 (approved):
LGTM!

Makefile
freddyaboulton: Shouldn't the `objective_tests` be in `test-other`? They don't depend on prophet right?

freddyaboulton (approved):
Thank you @ParthivNaresh !

PR #2645, opened 2021-08-17
Simplify logging behavior

### LINKED ISSUE #1174: Simplify evalml's logging behavior

eccabay:
Closes #1174 - design doc [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/972456879/Simplify+EvalML+logging?focusedCommentId=983204675#comment-983204675)


evalml/automl/automl_search.py
```diff
@@ -791,8 +803,12 @@ def search(self, show_iteration_plot=True):
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
         """
+        if self.verbose:
+            logger = get_logger(__name__)
+        else:
+            logger = logging.getLogger(__name__)
```
dsherry, line 809: Is there a way we can avoid duplicate code? Over time the duplicate paths could drift.

One idea: do
```python
        if self.verbose:
            self.logger = get_logger(__name__)
        else:
            self.logger = logging.getLogger(__name__)
```
in `AutoMLSearch.__init__` as you are already doing. And then in the rest of the `AutoMLSearch` code, i.e. `search` and `_should_continue`, reference `self.logger` instead of `logger`

evalml/automl/automl_search.py
```diff
@@ -831,7 +847,7 @@ def search(self, show_iteration_plot=True):
             % ", ".join([model.value for model in self.allowed_model_families])
         )
         self.search_iteration_plot = None
-        if self.plot:
+        if self.plot and self.verbose:
```
dsherry, line 850: If I follow correctly, this means we will only show the search iteration plot in verbose mode.

However, I thought we already had code set up which would only show this plot in the jupyter console. Meaning, if you run evalml outside of jupyter, it will never show up.

Is this correct? If so, why tie this to the verbose flag?

evalml/automl/automl_search.py
```diff
@@ -667,6 +676,9 @@ def __init__(
             pipeline_params=parameters,
             custom_hyperparameters=custom_hyperparameters,
         )
+        if self.verbose:
+            logger.handlers.pop()
+            logger.setLevel(logging.WARNING)
```
dsherry, line 681: Why do this at the end of init? Why not put it with the other logger init code at the top of the init method?

dsherry: Also, calling `pop` implies we're assuming the last handler in the list on `logger.handlers` is one that we added and has to be deleted. What's the reason for doing so here?

evalml/pipelines/pipeline_base.py
```diff
@@ -222,6 +223,7 @@ def describe(self, return_dict=False):
         Returns:
             dict: Dictionary of all component parameters if return_dict is True, else None
         """
+        logger = get_logger(f"{__name__}.describe")
```
dsherry, line 226: Why do this? We're already getting the logger using `logging.getLogger(__name__)`

evalml/utils/logger.py
```diff
@@ -1,10 +1,6 @@
 import logging
-import os
 import sys
 import time
-import traceback
-from logging.handlers import RotatingFileHandler
-from pathlib import Path
 
 
 def get_logger(name):
```
dsherry, line 6: Now that we're not using this `get_logger` method in our source code, do we have unit test coverage for it?

evalml/automl/automl_search.py
eccabay: I had that implementation when I started out - the issue that I ran into was with reverting the change afterwards. 

We want to make sure our logging configuration is removed as soon as it's no longer necessary, so that (for example) if you run search with `verbose=True` followed by `verbose=False`, the second run does not log anything. However, simply reverting the logging at the end of search does not guarantee this reversion - a user might not run search after instantiation, and may change the verbose flag in between those runs (I ran into this issue when running our tests!). Therefore to make this as flexible as possible, it makes the most sense to set up and tear down the logger within each function independently, so we don't carry over unwanted logging functionality anywhere.

That being said, I'm more than happy to throw this logic into a helper function to clean up code duplication!

evalml/automl/automl_search.py
eccabay: I referenced this in my above comment, but it's to ensure we undo our logging configuration as soon as we no longer need it! (Briefly discussed in the design doc [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/972456879/Simplify+EvalML+logging?focusedCommentId=983695807#comment-983695807))

evalml/automl/automl_search.py
eccabay: I'm happy to change this if people think it's important! I added this limitation since I was testing my changes in jupyter and it felt weird to have just a graph updating at random times, when there was no other output coming during search.

evalml/utils/logger.py
eccabay: Yep, tests still exist in `get_logger`, just without the tests for file configuration!

We do still use this `get_logger` method when in verbose mode for AutoMLSearch and when describing pipelines and components, with associated tests.

dsherry (changes requested):
Nice work! Looking solid

I left some questions and suggestions about the automl search code, and asked about unit test coverage for the get logger method

evalml/automl/automl_search.py
dsherry: Ah right, understood, thanks for explaining that.

Hmm, perhaps this complexity around reversion is an argument for removing the verbosity flag altogether. What do you think, would that be helpful here?

evalml/automl/automl_search.py
dsherry: Ah right, that makes sense. If we're gonna keep the verbosity flag, I suppose it would be strange to see a plot appear without any log output.

docs/source/start.ipynb
```diff
@@ -135,6 +135,24 @@
     "automl.search()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you would like to see real time updates as `AutoMLSearch` runs, you can use the argument `verbose=True`."
```
dsherry, line 142: Should we update our docs examples to all use `verbose=True` by default, and then have this section say "If you'd like to suppress stdout output, turn off verbose" ?

evalml/automl/automl_search.py
```diff
@@ -667,6 +673,18 @@ def __init__(
...
+    def _activate_logger(self):
+        if self.verbose:
+            self.logger = get_logger(__name__)
+        else:
+            self.logger = logging.getLogger(__name__)
+
+    def _deactivate_logger(self):
+        if self.verbose and len(self.logger.handlers):
+            self.logger.handlers.pop()
+            self.logger.setLevel(logging.WARNING)
```
dsherry, line 687: Remind me once more why we reset the logger level to WARNING here?

Suppose a user has an application where, somewhere down in the stack, they run `automl=AutoMLSearch(..., verbose=True); automl.search()` and then delete the automl object. And let's say first they set the global log level to INFO. By setting the log level here, is it possible this line would reset the global log level to WARNING, overriding their original setting?

evalml/automl/automl_search.py
```diff
@@ -771,7 +789,9 @@ def _handle_keyboard_interrupt(self):
                 .lower()
             )
             if choice == "y":
-                logger.info("Exiting AutoMLSearch.")
+                self._activate_logger
+                self.logger.info("Exiting AutoMLSearch.")
+                self._deactivate_logger
```
dsherry, line 794: Warning: missing parentheses to invoke the methods here

I'm surprised our lint didn't catch this! I wonder if there's a way / file an issue to have lint fail if we try to invoke a non-property as a property. Tagging @angela97lin since she's been messing with linting / style recently

evalml/automl/automl_search.py
```diff
@@ -791,10 +811,12 @@ def search(self, show_iteration_plot=True):
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
         """
+        self._activate_logger()
         if self._searched:
-            logger.info(
+            self.logger.error(
                 "AutoMLSearch.search() has already been run and will not run again on the same instance. Re-initialize AutoMLSearch to search again."
             )
+            self._deactivate_logger()
```
dsherry, line 819: What would happen if we deleted `_deactivate_logger` entirely from this PR?

evalml/automl/automl_search.py
angela97lin: That's a really good question @dsherry! I'm not sure I've run into anything that would have caught this--probably tricky to do since calling this without the paraentheses still returns the method and is "valid" 🤔 . Will keep an eye out though!

evalml/automl/automl_search.py
eccabay: Deleting `_deactivate_logger` would mean we can no longer run AutoML search with `verbose=False` after running with `verbose=True`, since our `get_logger` function (run when verbose is True) modifies the same logger that's grabbed by `logging.getLogger` when verbose is False.

Just now in reading your comments I had a thought - it doesn't have to be the case that the logger obtained when verbose is True vs when verbose is False have the same name. Instead, we could create another sub-logger for when verbose is True (like I did with `pipeline.describe()` and others) so that it would maintain any user-configured logging behavior without needing to be reset with `_deactivate_logger` at the end of each method that uses it. This would also resolve your question above about setting the logging level back to `logging.WARNING` in `_deactivate_logger`, as that would no longer be necessary. Thoughts?

docs/source/start.ipynb
angela97lin: I like this idea, could be helpful in the doc examples to be loud about the steps that we're taking!

evalml/automl/automl_search.py
```diff
@@ -1230,6 +1256,7 @@ def describe_pipeline(self, pipeline_id, return_dict=False):
             Description of specified pipeline. Includes information such as
             type of pipeline components, problem, training time, cross validation, etc.
         """
+        logger = get_logger(f"{__name__}.describe_pipeline")
```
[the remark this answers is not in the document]
angela97lin: To clarify, why do we need a separate sub-logger here, rather than using `self.logger`? :o

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1252,6 +1253,8 @@ def test_automl_search_sampler_method(
     has_minimal_dependencies,
     caplog,
 ):
+    logger = logging.getLogger("evalml")
+    logger.setLevel(logging.DEBUG)
```
angela97lin, line 1257: Why is this necessary here? :0

evalml/automl/automl_search.py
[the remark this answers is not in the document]
eccabay: `describe_pipeline` uses logging to display the pipeline information - if we used `self.logger` and the logger was set to only WARNING level or higher (as is default now), the method wouldn't display anything. To make sure we still display something, we have to set up a logger ourselves, but need the sublogger to make sure we don't override the user's choice to log only higher-level logs.

evalml/tests/automl_tests/test_automl_search_classification.py
eccabay: Aha, good catch! It was necessary before, since we run `assert "Could not import imblearn.over_sampling" in caplog.text` on line 1273 and that was previously logged at the debug level (so caplog didn't catch it). Since I switched that "Could not import" to log at a warning level, the caplog catches it without needing to increase the logging level. Will delete!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -298,33 +316,38 @@ def test_pipeline_limits(
...
+        verbose=verbose,
     )
     env = AutoMLTestEnv("binary")
     with env.test_context(
         mock_fit_side_effect=Exception("all your model are belong to us")
     ):
         automl.search()
     out = caplog.text
-    assert "Exception during automl search" in out
+    "Exception during automl search" in out
```
bchen1116, line 350: I don't think this is testing the assert?

bchen1116 (commented):
Left a question and a few nits, but this was great work!

evalml/automl/automl_search.py
```diff
@@ -1358,7 +1369,7 @@ def _check_for_high_variance(self, pipeline, cv_scores, threshold=0.5):
             return high_variance_cv
         cv_range = max(cv_scores) - min(cv_scores)
         if cv_range >= threshold * allowed_range:
-            logger.warning(
+            self.logger.warning(
```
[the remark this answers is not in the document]
eccabay: The warning is still there in both cases, it just appears in different ways! When `verbose=False`, the default logging is used, so warnings and higher are printed in the default format of red highlighting. When `verbose=True`, we override the default formatting settings, so while warnings are still printed to stdout (along with all the info level information), they are no longer highlighted in red.

You can see the same warning in the `verbose=True` case where the report of the Decision Tree cv scores are.

chukarsten (approved):
This is a great PR.  I like what's going on here and the simplicity of it.  

I think the one thing we should address before merge is the testing strategy.  Does it make sense to test this feature by running a variety of tests with verbose on and off?  I think this is OK for tests that don't drastically increase the test runtime by much (don't run automl.search).  What are your thoughts on keeping the tests as written before this PR, but adding verbose=True to all of them, then have just a single test that verifies that nothing is logged when verbose=False.  It seems that if verbose=False, a single test should capture that entirely.

Let me know your thoughts.  If you don't want to do that, I think it would be sufficient to show the runtime of the tests isn't drastically affected!  

Another nice to have would be if you could silence the dask/cfengine stuff in the docs.  Out of scope, so feel free to tell me to pound sand.

eccabay:
@chukarsten thanks for the testing suggestions! I took another look at the tests with parameterizations of verbose values, and I agree, a lot of them weren't necessary. I've filtered through them, according to a few criteria:
- Any tests that didn't change behavior whether verbose was True or False (i.e. the message logged was either warning or error level) don't need parameterization. If they pass when verbose is False, they will pass when verbose is True by definition.
- I removed some redundant tests, i.e. testing verbose/not in methods that other tests already covered. For example, parameterizing `verbose=[True, False]` in `test_pipeline_limits` made sure the distinction between the values worked correctly in both `__init__` and `search`. Therefore, we didn't need to parameterize verbose in `test_automl_drop_unknown_columns`, since the logging differences there only occur in `__init__`.
At this point, every test that includes `@pytest.mark.parametrize("verbose", [True, False])` should be testing something unique with respect to the verbose flag's functionality and behavior.

PR #2649, opened 2021-08-17
Refactor time series pipeline code via time series pipeline base class

### LINKED ISSUE #2614: Refactor time series pipeline inheritance

angela97lin:
Closes #2614


chukarsten (approved):
LGTM!

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -10,9 +7,7 @@
 )
 
 
-class TimeSeriesRegressionPipeline(
-    RegressionPipeline, metaclass=TimeSeriesPipelineBaseMeta
-):
+class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
```
freddyaboulton, line 10: Can you please mark that `TimeSeriesRegressionPipeline` does not inherit from `RegressionPipeline` as a breaking change?

freddyaboulton (approved):
Thank you @angela97lin !

jeremyliweishih (approved):
looks GREAT to me!

evalml/pipelines/time_series_regression_pipeline.py
angela97lin: sure thing 😁

PR #2651, opened 2021-08-18
Improve SVM hyperparameters

### LINKED ISSUE #2615: Handle scaling for SVM
**Issue**:
We currently allow a linear kernel as a tunable parameter for SVM classifiers. However, allowing a linear kernel turns the SVM model into a linear model. Linear models require input data with features that are on the same order of magnitude as each other, which is not a requirement for other types of functions (and therefore not a requirement for other SVM kernels).

**Repro**:
Run `fit` with `kernel='linear'` set in an `SVMClassifier` on [this dataset](https://evalml-looking-glass-datasets.s3.amazonaws.com/space_ga_1.csv), the computation will hang and never complete. Note that the `XCOORD` column is on the 10^7 order of magnitude, while all others are in single digits.

**Options**: 
- Remove `"linear"` from the kernel hyperparameter range for our SVM classifier
- Add a `StandardScaler` to any SVM pipelines with `"linear"` as the kernel

Note: the same probably holds for SVM regressors, but I haven't done that experimenting explicitly.

eccabay:
Closes #2615 by removing "linear" as an option from `SVMClassifier` and `SVMRegressor`, and swaps "auto" to be SVM's default gamma parameter for increased first-guess performance (as discussed in performance test results [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/981991444/Adding+SVM+to+AutoML))


evalml/pipelines/components/estimators/classifiers/svm_classifier.py
```diff
@@ -14,23 +14,23 @@ class SVMClassifier(Estimator):
...
         C (float): The regularization parameter. The strength of the regularization is inversely proportional to C.
             Must be strictly positive. The penalty is a squared l2 penalty. Defaults to 1.0.
-        kernel ({"linear", "poly", "rbf", "sigmoid", "precomputed"}): Specifies the kernel type to be used in the algorithm. Defaults to "rbf".
-        gamma ({"scale", "auto"} or float): Kernel coefficient for "rbf", "poly" and "sigmoid". Defaults to "scale".
-            - If gamma='scale' (default) is passed then it uses 1 / (n_features * X.var()) as value of gamma
-            - if "auto", uses 1 / n_features
+        kernel ({"poly", "rbf", "sigmoid", "precomputed"}): Specifies the kernel type to be used in the algorithm. Defaults to "rbf".
+        gamma ({"scale", "auto"} or float): Kernel coefficient for "rbf", "poly" and "sigmoid". Defaults to "auto".
+            - If gamma='scale' is passed then it uses 1 / (n_features * X.var()) as value of gamma
+            - if "auto" (default), uses 1 / n_features
```
chukarsten, line 20: meganit - either "If" here or "if" on the line before, haha.

evalml/pipelines/components/estimators/regressors/svm_regressor.py
```diff
@@ -13,22 +13,22 @@ class SVMRegressor(Estimator):
...
         C (float): The regularization parameter. The strength of the regularization is inversely proportional to C.
             Must be strictly positive. The penalty is a squared l2 penalty. Defaults to 1.0.
-        kernel ({"linear", "poly", "rbf", "sigmoid", "precomputed"}): Specifies the kernel type to be used in the algorithm. Defaults to "rbf".
-        gamma ({"scale", "auto"} or float): Kernel coefficient for "rbf", "poly" and "sigmoid". Defaults to "scale".
-            - If gamma='scale' (default) is passed then it uses 1 / (n_features * X.var()) as value of gamma
-            - if "auto", uses 1 / n_features
+        kernel ({"poly", "rbf", "sigmoid", "precomputed"}): Specifies the kernel type to be used in the algorithm. Defaults to "rbf".
+        gamma ({"scale", "auto"} or float): Kernel coefficient for "rbf", "poly" and "sigmoid". Defaults to "auto".
+            - If gamma='scale' is passed then it uses 1 / (n_features * X.var()) as value of gamma
+            - if "auto" (default), uses 1 / n_features
```
chukarsten, line 19: Same here!

chukarsten (approved):
Looks good!  Just two hyper nits about the doc strings!

evalml/pipelines/components/estimators/regressors/svm_regressor.py
```diff
@@ -42,7 +42,7 @@ class SVMRegressor(Estimator):
         ProblemTypes.TIME_SERIES_REGRESSION,
     ]"""
 
-    def __init__(self, C=1.0, kernel="rbf", gamma="scale", random_seed=0, **kwargs):
+    def __init__(self, C=1.0, kernel="rbf", gamma="auto", random_seed=0, **kwargs):
```
freddyaboulton, line 45: Should we make this change? The perf tests only considered binary classification problems.

freddyaboulton (commented):
@eccabay Thank you for this! I agree with removing the linear kernel for both regression and classification but not sure if we should tweak the other hyperparameters for regression since the perf tests you only ran on binary classification.

I left a comment on your results as well about how much better SVM is than the next best estimator. The fit time is only slightly slower for most datasets but for some it's like 4x slower. If the SVM more than 4x better, there is an argument for including it in AutoMLSearch.

I'd like to continue the discussion on your perf test doc before approving!

evalml/pipelines/components/estimators/regressors/svm_regressor.py
eccabay: This is a fair point! It's hard to say since we don't have very many regression datasets in looking glass. I'll run a few tests with what we have and see if the results are consistent or not!

eccabay:
One thing I forgot to implement/mention before publishing this PR is that `kernel=precomputed` also needs to be removed - I had it removed in my testing branch and forgot to carry it over into this one. With this parameter, Sklearn's SVC expects a precomputed kernel of shape (n_samples_test, n_samples_train), which we don't give it, so it errors out and doesn't run in the first place.

bchen1116 (approved):
LGTM! Also curious about the change of `gamma` from `scale` to `auto`, but will wait for the results on that!

evalml/pipelines/components/estimators/regressors/svm_regressor.py
eccabay: @freddyaboulton Results from the regression testing is now in the performance test doc! On the very small number of datasets we have, "auto" performs better significantly more often than "scale", so I think this change should happen.

freddyaboulton (approved):
Thank you @eccabay !

Issue #2620, opened 2021-08-12
Deprecate Scikit-Learn based Ensembling Component


christopherbunn:
As part of our implementation of our [new ensembler component](#1930), we will need to deprecate the current scikit-learn based StackingEnsembler. Once we finish the new ensembler component, it will be used alongside the scikit-learn version during an ensembling AutoML search. However, because the new ensembler component's will contain the same functionality of the scikit-learn based ensembler, there isn't a need to keep both. This issue tracks deprecating the scikit-learn based ensembler.

Current implementation plan:
- Keep both ensemblers in AutoML search, but raise a DeprecationWarning during search that the sklearn version will not be included in search during the next release. 
- Remove the scikit-learn based ensembler from AutoML search, but still retain
   - This will allow users to create new pipelines with this component manually if necessary
- Eventually, remove the scikit-learn based ensembler component from EvalML completely

PR #2653, opened 2021-08-18
Add new ensembler component

### LINKED ISSUE #1930: Build ensembler component

christopherbunn:
Resolves #1930

### DIFF
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -1,6 +1,7 @@
 """Base class for the AutoML algorithms which power EvalML."""
 from abc import ABC, abstractmethod
 
+from evalml import model_family
 from evalml.exceptions import PipelineNotFoundError
 from evalml.pipelines.components.utils import handle_component_class
 from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
@@ -93,45 +94,20 @@ class AutoMLAlgorithm(ABC):
         next_batch = []
 
         # Custom Stacked Pipelines
-        ensembler_component_graph = {}
-        final_components = []
-        problem_type = None
+        problem_type = list(self._best_pipeline_info.values())[0][
+            "pipeline"
+        ].problem_type
         n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
-
-        for model_type, best_info in self._best_pipeline_info.items():
-
-            def _make_new_component_name(component_name):
-                return str(model_type) + " Pipeline - " + component_name
-
-            pipeline = best_info["pipeline"]
-            if problem_type is None:
-                problem_type = pipeline.problem_type
-            final_component = None
-            ensemble_y = "y"
-            for name, component_list in pipeline.component_graph.component_dict.items():
-                new_component_list = []
-                new_component_name = _make_new_component_name(name)
-                for i, item in enumerate(component_list):
-                    if i == 0:
-                        fitted_comp = handle_component_class(item)
-                        new_component_list.append(fitted_comp)
-                    elif isinstance(item, str) and item not in ["X", "y"]:
-                        new_component_list.append(_make_new_component_name(item))
-                    else:
-                        new_component_list.append(item)
-                    if i != 0 and item.endswith(".y"):
-                        ensemble_y = _make_new_component_name(item)
-                ensembler_component_graph[new_component_name] = new_component_list
-                final_component = new_component_name
-            final_components.append(final_component)
+        input_pipelines = [
+            best_info["pipeline"]
+            for best_info in list(self._best_pipeline_info.values())
+        ]
 
         ensemble = _make_stacked_ensemble_pipeline(
+            input_pipelines,
             problem_type,
-            component_graph=ensembler_component_graph,
-            final_components=final_components,
             random_seed=self.random_seed,
             n_jobs=n_jobs_ensemble,
-            ensemble_y=ensemble_y,
         )
         next_batch.append(ensemble)
 
@@ -154,8 +130,8 @@ class AutoMLAlgorithm(ABC):
                 pipeline.new(parameters=parameters, random_seed=self.random_seed)
             )
         ensemble = _make_stacked_ensemble_pipeline(
+            input_pipelines,
             problem_type,
-            input_pipelines=input_pipelines,
             random_seed=self.random_seed,
             n_jobs=n_jobs_ensemble,
             use_sklearn=True,
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -34,6 +34,7 @@ class AutoMLAlgorithm(ABC):
         max_iterations=None,
         tuner_class=None,
         random_seed=0,
+        n_jobs=-1,
     ):
         self.random_seed = random_seed
         self.allowed_pipelines = allowed_pipelines or []
@@ -42,7 +43,8 @@ class AutoMLAlgorithm(ABC):
         self._tuners = {}
         self._best_pipeline_info = {}
         self.text_in_ensembling = False
-        self.n_jobs = -1
+        self.n_jobs = n_jobs
+        self._selected_cols = None
         for pipeline in self.allowed_pipelines:
             pipeline_hyperparameters = pipeline.get_hyperparameter_ranges(
                 custom_hyperparameters
@@ -59,6 +61,15 @@ class AutoMLAlgorithm(ABC):
 
         Returns:
             list[PipelineBase]: A list of instances of PipelineBase subclasses, ready to be trained and evaluated.
+        """
+
+    @abstractmethod
+    def _transform_parameters(self, pipeline, proposed_parameters):
+        """Given a pipeline parameters dict, make sure pipeline_params, custom_hyperparameters, n_jobs are set properly.
+
+        Arguments:
+            pipeline (PipelineBase): The pipeline object to update the parameters.
+            proposed_parameters (dict): Parameters to use when updating the pipeline.
         """
 
     def add_result(self, score_to_minimize, pipeline, trained_pipeline_results):
@@ -98,17 +109,16 @@ class AutoMLAlgorithm(ABC):
         input_pipelines = []
         for pipeline_dict in best_pipelines:
             pipeline = pipeline_dict["pipeline"]
-            pipeline_params = pipeline_dict["parameters"]
-            if hasattr(self, "_transform_parameters"):
-                pipeline_params = self._transform_parameters(pipeline, pipeline_params)
-            if hasattr(self, "_selected_cols"):
-                if (
-                    "Select Columns Transformer"
-                    in pipeline.component_graph.component_instances
-                ):
-                    pipeline_params.update(
-                        {"Select Columns Transformer": {"columns": self._selected_cols}}
-                    )
+            pipeline_params = self._transform_parameters(
+                pipeline, pipeline_dict["parameters"]
+            )
+            if (
+                "Select Columns Transformer"
+                in pipeline.component_graph.component_instances
+            ):
+                pipeline_params.update(
+                    {"Select Columns Transformer": {"columns": self._selected_cols}}
+                )
             input_pipelines.append(
                 pipeline.new(parameters=pipeline_params, random_seed=self.random_seed)
             )
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -92,14 +92,26 @@ class AutoMLAlgorithm(ABC):
         next_batch = []
 
         # Custom Stacked Pipelines
-        problem_type = list(self._best_pipeline_info.values())[0][
-            "pipeline"
-        ].problem_type
+        best_pipelines = list(self._best_pipeline_info.values())
+        problem_type = best_pipelines[0]["pipeline"].problem_type
         n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
-        input_pipelines = [
-            best_info["pipeline"]
-            for best_info in list(self._best_pipeline_info.values())
-        ]
+        input_pipelines = []
+        for pipeline_dict in best_pipelines:
+            pipeline = pipeline_dict["pipeline"]
+            pipeline_params = pipeline_dict["parameters"]
+            if hasattr(self, "_transform_parameters"):
+                pipeline_params = self._transform_parameters(pipeline, pipeline_params)
+            if hasattr(self, "_selected_cols"):
+                if (
+                    "Select Columns Transformer"
+                    in pipeline.component_graph.component_instances
+                ):
+                    pipeline_params.update(
+                        {"Select Columns Transformer": {"columns": self._selected_cols}}
+                    )
+            input_pipelines.append(
+                pipeline.new(parameters=pipeline_params, random_seed=self.random_seed)
+            )
 
         ensemble = _make_stacked_ensemble_pipeline(
             input_pipelines,
@@ -110,23 +122,6 @@ class AutoMLAlgorithm(ABC):
         next_batch.append(ensemble)
 
         # Sklearn Stacked Pipelines
-        input_pipelines = []
-        for pipeline_dict in self._best_pipeline_info.values():
-            pipeline = pipeline_dict["pipeline"]
-            pipeline_params = pipeline_dict["parameters"]
-            if hasattr(self, "_transform_parameters"):
-                parameters = self._transform_parameters(pipeline, pipeline_params)
-            if hasattr(self, "_selected_cols"):
-                if (
-                    "Select Columns Transformer"
-                    in pipeline.component_graph.component_instances
-                ):
-                    parameters.update(
-                        {"Select Columns Transformer": {"columns": self._selected_cols}}
-                    )
-            input_pipelines.append(
-                pipeline.new(parameters=parameters, random_seed=self.random_seed)
-            )
         ensemble = _make_stacked_ensemble_pipeline(
             input_pipelines,
             problem_type,
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -1,9 +1,7 @@
 """Base class for the AutoML algorithms which power EvalML."""
 from abc import ABC, abstractmethod
 
-from evalml import model_family
 from evalml.exceptions import PipelineNotFoundError
-from evalml.pipelines.components.utils import handle_component_class
 from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
 from evalml.tuners import SKOptTuner
 
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -2,6 +2,8 @@
 from abc import ABC, abstractmethod
 
 from evalml.exceptions import PipelineNotFoundError
+from evalml.pipelines.components.utils import handle_component_class
+from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
 from evalml.tuners import SKOptTuner
 
 
@@ -39,6 +41,9 @@ class AutoMLAlgorithm(ABC):
         self.max_iterations = max_iterations
         self._tuner_class = tuner_class or SKOptTuner
         self._tuners = {}
+        self._best_pipeline_info = {}
+        self.text_in_ensembling = False
+        self.n_jobs = -1
         for pipeline in self.allowed_pipelines:
             pipeline_hyperparameters = pipeline.get_hyperparameter_ranges(
                 custom_hyperparameters
@@ -83,3 +88,77 @@ class AutoMLAlgorithm(ABC):
     def batch_number(self):
         """Returns the number of batches which have been recommended so far."""
         return self._batch_number
+
+    def _create_ensemble(self):
+        next_batch = []
+
+        # Custom Stacked Pipelines
+        ensembler_component_graph = {}
+        final_components = []
+        problem_type = None
+        n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
+
+        for model_type, best_info in self._best_pipeline_info.items():
+
+            def _make_new_component_name(component_name):
+                return str(model_type) + " Pipeline - " + component_name
+
+            pipeline = best_info["pipeline"]
+            if problem_type is None:
+                problem_type = pipeline.problem_type
+            final_component = None
+            ensemble_y = "y"
+            for name, component_list in pipeline.component_graph.component_dict.items():
+                new_component_list = []
+                new_component_name = _make_new_component_name(name)
+                for i, item in enumerate(component_list):
+                    if i == 0:
+                        fitted_comp = handle_component_class(item)
+                        new_component_list.append(fitted_comp)
+                    elif isinstance(item, str) and item not in ["X", "y"]:
+                        new_component_list.append(_make_new_component_name(item))
+                    else:
+                        new_component_list.append(item)
+                    if i != 0 and item.endswith(".y"):
+                        ensemble_y = _make_new_component_name(item)
+                ensembler_component_graph[new_component_name] = new_component_list
+                final_component = new_component_name
+            final_components.append(final_component)
+
+        ensemble = _make_stacked_ensemble_pipeline(
+            problem_type,
+            component_graph=ensembler_component_graph,
+            final_components=final_components,
+            random_seed=self.random_seed,
+            n_jobs=n_jobs_ensemble,
+            ensemble_y=ensemble_y,
+        )
+        next_batch.append(ensemble)
+
+        # Sklearn Stacked Pipelines
+        input_pipelines = []
+        for pipeline_dict in self._best_pipeline_info.values():
+            pipeline = pipeline_dict["pipeline"]
+            pipeline_params = pipeline_dict["parameters"]
+            if hasattr(self, "_transform_parameters"):
+                parameters = self._transform_parameters(pipeline, pipeline_params)
+            if hasattr(self, "_selected_cols"):
+                if (
+                    "Select Columns Transformer"
+                    in pipeline.component_graph.component_instances
+                ):
+                    parameters.update(
+                        {"Select Columns Transformer": {"columns": self._selected_cols}}
+                    )
+            input_pipelines.append(
+                pipeline.new(parameters=parameters, random_seed=self.random_seed)
+            )
+        ensemble = _make_stacked_ensemble_pipeline(
+            problem_type,
+            input_pipelines=input_pipelines,
+            random_seed=self.random_seed,
+            n_jobs=n_jobs_ensemble,
+            use_sklearn=True,
+        )
+        next_batch.append(ensemble)
+        return next_batch
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -734,7 +734,7 @@ class AutoMLSearch:
                 1
                 + len(self.allowed_pipelines)
                 + len(self.allowed_pipelines) * self._pipelines_per_batch
-                + 1
+                + 2
             )
             if self.max_iterations < first_ensembling_iteration:
                 run_ensembling = False
@@ -766,7 +766,7 @@ class AutoMLSearch:
                     + len(self.allowed_pipelines)
                     + self._pipelines_per_batch
                     * (self.max_batches - 1 - num_ensemble_batches)
-                    + num_ensemble_batches
+                    + num_ensemble_batches * 2
                 )
             else:
                 self.max_iterations = (
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -734,7 +734,7 @@ class AutoMLSearch:
                 1
                 + len(self.allowed_pipelines)
                 + len(self.allowed_pipelines) * self._pipelines_per_batch
-                + 2
+                + 1
             )
             if self.max_iterations < first_ensembling_iteration:
                 run_ensembling = False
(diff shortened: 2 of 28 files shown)


evalml/tests/component_tests/test_utils.py
```diff
@@ -31,26 +31,26 @@ def test_all_components(
...
-        n_components = 52
+        n_components = 54
     elif is_using_windows and not is_running_py_39_or_above:
         # No prophet
-        n_components = 53
+        n_components = 55
     elif is_using_windows and is_running_py_39_or_above:
         # No detrender, no arima, no prophet
-        n_components = 51
+        n_components = 53
```
christopherbunn, line 48: Since these are Windows/Conda tests, we might just not get coverage for this section until we merge to main.

freddyaboulton (dismissed):

evalml/tests/automl_tests/test_evalml_algorithm.py
```diff
@@ -12,6 +12,12 @@
     SklearnStackedEnsembleClassifier,
     SklearnStackedEnsembleRegressor,
 )
+from evalml.pipelines.components.ensemble.stacked_ensemble_classifier import (
+    StackedEnsembleClassifier,
+)
+from evalml.pipelines.components.ensemble.stacked_ensemble_regressor import (
+    StackedEnsembleRegressor,
+)
```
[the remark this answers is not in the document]
christopherbunn: Good catch, I think this was a VSCode autofill thing 😬

evalml/automl/automl_algorithm/evalml_algorithm.py
```diff
@@ -202,6 +202,52 @@ def _create_fast_final(self):
         return pipelines
 
     def _create_ensemble(self):
+        next_batch = []
```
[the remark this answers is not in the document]
christopherbunn: Moved this to `AutoMLAlgorithm`. I ended up having to do attribute checks for this, but the upside is that there is now one source of truth to how ensemble pipelines are created.

christopherbunn:
[Here ](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1012335076/Ensembler+Performance+Results) is the ensembler performance test doc that I have so far. I think that once I get in this PR for looking glass we can get into a deeper analysis of the results and see where the new ensembler ends up relative to the old one.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -80,3 +84,48 @@ def pipeline_number(self):
...
         """Returns the number of batches which have been recommended so far."""
         return self._batch_number
+
+    def _create_ensemble(self):
+        next_batch = []
+
+        # Custom Stacked Pipelines
+        problem_type = list(self._best_pipeline_info.values())[0][
+            "pipeline"
+        ].problem_type
```
eccabay, line 94: This is a hard-to-read way of fetching the problem type, and the long call of `list(self._best_pipeline_info.values())` is used more than once - it might be worth it to factor that bit out into its own descriptively-named variable!

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -80,3 +84,48 @@ def pipeline_number(self):
...
+            pipeline = pipeline_dict["pipeline"]
+            pipeline_params = pipeline_dict["parameters"]
+            if hasattr(self, "_transform_parameters"):
+                parameters = self._transform_parameters(pipeline, pipeline_params)
+            if hasattr(self, "_selected_cols"):
+                if (
+                    "Select Columns Transformer"
+                    in pipeline.component_graph.component_instances
+                ):
+                    parameters.update(
```
eccabay, line 107: If I'm reading this correctly, this relies on `hasattr(self, "_transformer_parameters")` being True in order to create the `parameters` object being updated here and used in creating a new pipeline down below. Is there a situation where `self._transform_parameters` would not exist?

evalml/automl/engine/engine_base.py
```diff
@@ -170,7 +172,7 @@ def train_and_score_pipeline(
     for i, (train, valid) in enumerate(
         automl_config.data_splitter.split(full_X_train, full_y_train)
     ):
-        if pipeline.model_family == ModelFamily.ENSEMBLE and i > 0:
+        if isinstance(pipeline.estimator, SklearnStackedEnsembleBase) and i > 0:
             # Stacked ensembles do CV internally, so we do not run CV here for performance reasons.
             logger.debug(
                 f"Skipping fold {i} because CV for stacked ensembles is not supported."
```
eccabay, line 178: Could you update the comment/debug message here since this is now only applicable for sklearn ensembles?

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -0,0 +1,69 @@
...
+
+        super().__init__(
+            parameters=parameters,
+            component_obj=final_estimator,
+            random_seed=random_seed,
+        )
+
+    @property
+    def feature_importance(self):
+        """Not implemented for SklearnStackedEnsembleClassifier and SklearnStackedEnsembleRegressor"""
```
eccabay, line 54: Nit: `"""Not implemented for StackedEnsembleClassifier and StackedEnsembleRegressor"""`

evalml/pipelines/utils.py
```diff
@@ -264,48 +274,109 @@ def _make_stacked_ensemble_pipeline(
...
     if is_classification(problem_type):
-        parameters = {
-            "Sklearn Stacked Ensemble Classifier": {
-                "input_pipelines": input_pipelines,
-                "n_jobs": n_jobs,
+        if use_sklearn:
+            parameters = {
+                "Sklearn Stacked Ensemble Classifier": {
+                    "input_pipelines": input_pipelines,
+                    "n_jobs": n_jobs,
```
eccabay, line 292: Is there a reason the regression pipelines take a `final_estimator` here but the classifiers don't?

evalml/tests/component_tests/test_utils.py
```diff
@@ -31,26 +31,26 @@ def test_all_components(
     is_using_conda,
     is_using_windows,
 ):
-    # The total number of minimal components is 42
-    # The total number of components is 53
+    # The total number of minimal components is 43
+    # The total number of components is 56
```
eccabay, line 35: Not sure 56 is the correct value?

eccabay (approved):
Nice, this continues to be awesome work! Just left a couple small suggestions and nits.

evalml/automl/automl_algorithm/automl_algorithm.py
christopherbunn: Currently, we have `self._transform_parameters` in both the default and iterative algorithm. However, I put this check in since `self._transform_parameters` and `self._selected_cols` doesn't exist in the AutoMLAlgorithm abstract base class. I updated this to create an empty parameters dictionary if this attribute doesn't exist.

evalml/pipelines/utils.py
christopherbunn: Oops, I think it might have been left over from the previous copy/paste job. I updated this so that `final_estimator` is a param in sklearn ensemblers and not in the new one.

angela97lin (commented):
Just some comments on what I've read thus far, will add more later!

angela97lin:
Just trying out some examples here of potentially more complicated graphs:

```

graph_en = {
        "Imputer": ["Imputer", "X", "y"],
        "Target Imputer": ["Target Imputer", "X", "y"],
        "OneHot_RandomForest": ["One Hot Encoder", "Imputer.x", "Target Imputer.y"],
        "OneHot_ElasticNet": ["One Hot Encoder", "Imputer.x", "y"],
        "Random Forest": ["Random Forest Regressor", "OneHot_RandomForest.x", "y"],
        "Elastic Net": ["Elastic Net Regressor", "OneHot_ElasticNet.x", "y"],
        "EN": [
            "Elastic Net Regressor",
            "Random Forest.x",
            "Elastic Net.x",
            "y",
        ],
}


graph_rf = {
        "Imputer_1": ["Imputer", "X", "y"],
        "Target Imputer": ["Target Imputer", "X", "y"],
        "OneHot_RandomForest": ["One Hot Encoder", "Imputer_1.x", "Target Imputer.y"],
        "OneHot_ElasticNet": ["One Hot Encoder", "Imputer_1.x", "y"],
        "Random Forest": ["Random Forest Regressor", "OneHot_RandomForest.x", "y"],
        "Elastic Net": ["Elastic Net Regressor", "OneHot_ElasticNet.x", "y"],
        "RF": [
            "Linear Regressor",
            "Random Forest.x",
            "Elastic Net.x",
            "y",
        ],
}

RegressionPipeline(component_graph=graph_en)
RegressionPipeline(component_graph=graph_rf)

input_pipelines = [
    RegressionPipeline(component_graph=graph_en),
    RegressionPipeline(component_graph=graph_rf),
]

pipeline = _make_stacked_ensemble_pipeline(
        input_pipelines=input_pipelines,
        problem_type=ProblemTypes.REGRESSION,
        use_sklearn=False,
)

```

results in 

```
<ipython-input-45-fc88c214b3a2> in <module>
     46         input_pipelines=input_pipelines,
     47         problem_type=ProblemTypes.REGRESSION,
---> 48         use_sklearn=False,
     49 )
     50 

~/Desktop/evalml/evalml/pipelines/utils.py in _make_stacked_ensemble_pipeline(input_pipelines, problem_type, final_estimator, n_jobs, random_seed, use_sklearn)
    380         parameters=parameters,
    381         custom_name=pipeline_name,
--> 382         random_seed=random_seed,
    383     )
    384 

~/Desktop/evalml/evalml/pipelines/pipeline_base.py in __init__(self, component_graph, parameters, custom_name, random_seed)
     81         elif isinstance(component_graph, dict):
     82             self.component_graph = ComponentGraph(
---> 83                 component_dict=component_graph, random_seed=self.random_seed
     84             )
     85         elif isinstance(component_graph, ComponentGraph):

~/Desktop/evalml/evalml/pipelines/component_graph.py in __init__(self, component_dict, random_seed)
     58         self._feature_provenance = {}
     59         self._i = 0
---> 60         self._compute_order = self.generate_order(self.component_dict)
     61 
     62     def _validate_component_dict(self):

~/Desktop/evalml/evalml/pipelines/component_graph.py in generate_order(cls, component_dict)
    628         digraph.add_edges_from(edges)
    629         if not nx.is_weakly_connected(digraph):
--> 630             raise ValueError("The given graph is not completely connected")
    631         try:
    632             compute_order = list(topological_sort(digraph))

ValueError: The given graph is not completely connected
```

Interestingly enough, this disappears when I try not to differentiate between the two imputers in the given graphs:
```


graph_en = {
        "Imputer": ["Imputer", "X", "y"],
        "Target Imputer": ["Target Imputer", "X", "y"],
        "OneHot_RandomForest": ["One Hot Encoder", "Imputer.x", "Target Imputer.y"],
        "OneHot_ElasticNet": ["One Hot Encoder", "Imputer.x", "y"],
        "Random Forest": ["Random Forest Regressor", "OneHot_RandomForest.x", "y"],
        "Elastic Net": ["Elastic Net Regressor", "OneHot_ElasticNet.x", "y"],
        "EN": [
            "Elastic Net Regressor",
            "Random Forest.x",
            "Elastic Net.x",
            "y",
        ],
}


graph_rf = {
        "Imputer": ["Imputer", "X", "y"], # no longer `Imputer_1`
        "Target Imputer": ["Target Imputer", "X", "y"],
        "OneHot_RandomForest": ["One Hot Encoder", "Imputer.x", "Target Imputer.y"],
        "OneHot_ElasticNet": ["One Hot Encoder", "Imputer.x", "y"],
        "Random Forest": ["Random Forest Regressor", "OneHot_RandomForest.x", "y"],
        "Elastic Net": ["Elastic Net Regressor", "OneHot_ElasticNet.x", "y"],
        "Linear Regressor": [
            "Linear Regressor",
            "Random Forest.x",
            "Elastic Net.x",
            "y",
        ],
}

RegressionPipeline(component_graph=graph_en)
RegressionPipeline(component_graph=graph_rf)

input_pipelines = [
    RegressionPipeline(component_graph=graph_en),
    RegressionPipeline(component_graph=graph_rf),
]

pipeline = _make_stacked_ensemble_pipeline(
        input_pipelines=input_pipelines,
        problem_type=ProblemTypes.REGRESSION,
        use_sklearn=False,
)

```
Calling `pipeline.graph()` shows that both of these graphs end up with the same preprocessing steps--is this intended?

chukarsten (approved):
Really coming along!

evalml/automl/automl_search.py
```diff
@@ -763,7 +763,7 @@ def __init__(
                     + len(self.allowed_pipelines)
                     + self._pipelines_per_batch
                     * (self.max_batches - 1 - num_ensemble_batches)
-                    + num_ensemble_batches
+                    + num_ensemble_batches * 2
```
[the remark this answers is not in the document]
christopherbunn: We are planning on ultimately just using our component. This will most likely be revered when we do #2620.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -80,3 +84,47 @@ def pipeline_number(self):
...
+        ensemble = _make_stacked_ensemble_pipeline(
+            input_pipelines,
+            problem_type,
+            random_seed=self.random_seed,
+            n_jobs=n_jobs_ensemble,
+        )
+        next_batch.append(ensemble)
+
+        # Sklearn Stacked Pipelines
+        ensemble = _make_stacked_ensemble_pipeline(
```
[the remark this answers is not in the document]
christopherbunn: Yep, we're keeping both. The deprecation path for this is in #2620.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5037,3 +5062,70 @@ def test_automl_chooses_engine(engine_choice, X_y_binary):
...
+    }
+    automl = AutoMLSearch(
+        X,
+        y,
+        "regression",
+        allowed_component_graphs=component_graphs,
+        ensembling=True,
+        max_batches=4,
+    )
+    automl.search()
```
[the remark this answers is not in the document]
christopherbunn: From skimming the docstring for this mock, I saw [this line](https://github.com/alteryx/evalml/blob/27198716f31a0e1661b0d4733c960be0d94c1acf/evalml/tests/conftest.py#L1101): 
> but it may not be suitable for all tests, such as tests that patch Estimator.fit instead of Pipeline.fit

Since most of the tests are mocking `predict_proba` and `predict` from the Estimator, I'm assuming that we can't unless something has changed.

evalml/pipelines/components/ensemble/stacked_ensemble_base.py
```diff
@@ -0,0 +1,69 @@
...
+    _default_final_estimator = None
+
+    def __init__(
+        self,
+        final_estimator=None,
+        n_jobs=-1,
+        random_seed=0,
+        **kwargs,
+    ):
+        final_estimator = final_estimator or self._default_final_estimator()
```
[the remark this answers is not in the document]
christopherbunn: The main reason is because of serialization support. We can't serialize objects that store modules as an attribute. We had a similar issue with [imblearn](https://github.com/alteryx/evalml/pull/2755) and [plotly](https://github.com/alteryx/evalml/pull/2463)

christopherbunn:
@angela97lin RE: your pipeline graphs, the way that `_make_stacked_ensemble_pipelines` is structured is that it assumes that each pipeline has a unique model family. In the context of AutoMLSearch, this works since the input pipelines are all from different model families. 

In your examples, both of the input pipelines have the same model family. I've updated `_make_new_component_name` so that we add add an index value when there are multiple of the same type of model family.  The graph of your first example is below. Notice the new name for the second linear pipeline: 
image

And here is the second example. Here, the two separate imputers are shown and are from separate pipelines:
image

Let me know what you think! I am also open to setting a different component renaming scheme if you would like to as well.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -80,3 +84,47 @@ def pipeline_number(self):
...
+
+        # Custom Stacked Pipelines
+        best_pipelines = list(self._best_pipeline_info.values())
+        problem_type = best_pipelines[0]["pipeline"].problem_type
+        n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
+        input_pipelines = []
+        for pipeline_dict in best_pipelines:
+            pipeline = pipeline_dict["pipeline"]
+            pipeline_params = pipeline_dict["parameters"]
+            if hasattr(self, "_transform_parameters"):
```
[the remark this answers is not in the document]
christopherbunn: Added new abstract method for `_transfrom_parameters` and added `_selected_columns` as a default attribute.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -39,6 +40,9 @@ def __init__(
         self.max_iterations = max_iterations
         self._tuner_class = tuner_class or SKOptTuner
         self._tuners = {}
+        self._best_pipeline_info = {}
+        self.text_in_ensembling = False
+        self.n_jobs = -1
```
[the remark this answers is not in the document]
christopherbunn: Added as a new init param

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1521,30 +1527,49 @@ def test_describe_pipeline_with_ensembling(
...
+            True if "Sklearn" in automl.get_pipeline(ensemble_id).name else False
+        )
         caplog.clear()
         automl_dict = automl.describe_pipeline(ensemble_id, return_dict=return_dict)
         out = caplog.text
-        assert "Sklearn Stacked Ensemble Classification Pipeline" in out
+        if sklearn_pl:
+            assert "Sklearn Stacked Ensemble Classification Pipeline" in out
+            assert "* final_estimator : None" in out
+            num_sklearn_pl += 1
```
angela97lin, line 1546: Why are we adding 1 here? 👀 Do we use this variable anywhere else? / do we want to check the value of this?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5214,3 +5239,72 @@ def test_automl_chooses_engine(engine_choice, X_y_binary):
             automl = AutoMLSearch(
                 X_train=X, y_train=y, problem_type="binary", engine=engine_choice
             )
+
+
+@patch("evalml.pipelines.components.RandomForestClassifier.predict_proba")
+@patch("evalml.pipelines.components.RandomForestClassifier.predict")
+@patch("evalml.pipelines.components.ElasticNetClassifier.predict_proba")
+@patch("evalml.pipelines.components.ElasticNetClassifier.predict")
+def test_automl_enesmbler_allowed_component_graphs(
```
angela97lin, line 5248: Typo: test_automl_enesmbler_allowed_component_graphs --> test_automl_ensembler_allowed_component_graphs

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5214,3 +5239,72 @@ def test_automl_chooses_engine(engine_choice, X_y_binary):
...
+        X,
+        y,
+        "regression",
+        allowed_component_graphs=component_graphs,
+        ensembling=True,
+        max_batches=4,
+        verbose=True,
+    )
+    automl.search()
+    assert "Stacked Ensemble Regression Pipeline" in caplog.text
```
angela97lin, line 5310: Food for thought: do you think we need to run automl search to test for this? Is there a way to test without running search / training all the other pipelines?

evalml/tests/automl_tests/test_automl_algorithm.py
```diff
@@ -16,6 +16,9 @@ def next_batch(self):
             return self._dummy_pipelines.pop()
         raise StopIteration("No more pipelines!")
 
+    def _transform_parameters(self, pipeline, proposed_parameters):
```
angela97lin, line 19: So I'm understanding correctly: this is overridden since we're using a dummy algo that would otherwise inherit our complicated impl?

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
```diff
@@ -4,169 +4,139 @@
...
+    pytest.importorskip(
+        "imblearn.over_sampling",
+        reason="Skipping nondefault y test because imblearn not installed",
     )
-    clf = SklearnStackedEnsembleClassifier(
-        input_pipelines=[logistic_regression_binary_pipeline_class(parameters={})],
-        final_estimator=base,
-        n_jobs=1,
+    X, y = X_y_binary
+    input_pipelines = [
```
angela97lin, line 87: Interesting... how does `fit()` work here? Doesn't the first pipeline's oversampler cause the two input pipelines to have different number of rows for the final estimator?

angela97lin (commented):
This is looking great! I left a few notes and questions, but for sure, I'd love to see updated perf test results to see how this new ensemble compares 😁

evalml/tests/automl_tests/test_automl.py
christopherbunn: Forgot an assert, I'll add it at the end.

evalml/tests/automl_tests/test_automl_algorithm.py
christopherbunn: I made `_transform_parameters` an abstract method, so it needs to be implemented otherwise it'll throw an error. We are testing that in other places so I figured we can just have it be a pass here.

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
christopherbunn: When we had this bug previously (i.e. when we used to just use the input y blindly for the metalearner rather than take from an oversampler), we would get a mismatched input data error for this estimator. The `fit()` makes sure that it'll fit the pipeline without raising a RuntimeError and the `predict()` ensures that when we run that function, the input size match out the expected output size.

evalml/tests/automl_tests/test_automl.py
christopherbunn: This test was moreso to make sure that we can we can set pipelines in `allowed_component_graphs` and that it will create ensemble pipelines. Testing the case where we need to ensemble string/class based pipelines was more secondary. I can split off the second case into it's own test if you think that makes sense.

evalml/pipelines/component_graph.py
```diff
@@ -21,8 +21,9 @@
     TargetTransformer,
 )
 from evalml.pipelines.components.utils import handle_component_class
-from evalml.utils import import_or_raise, infer_feature_types
-from evalml.utils.logger import get_logger
+from evalml.utils import get_logger, import_or_raise, infer_feature_types
+
+logger = get_logger(__file__)
```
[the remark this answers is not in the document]
christopherbunn: This was a VSCode isort change I think. I can revert it if it's out of scope for this PR.

evalml/pipelines/components/ensemble/stacked_ensemble_classifier.py
```diff
@@ -0,0 +1,36 @@
...
+    ]
+    """[
+        ProblemTypes.BINARY,
+        ProblemTypes.MULTICLASS,
+        ProblemTypes.TIME_SERIES_BINARY,
+        ProblemTypes.TIME_SERIES_MULTICLASS,
+    ]"""
+    hyperparameter_ranges = {}
+    """{}"""
+    _default_final_estimator = ElasticNetClassifier
```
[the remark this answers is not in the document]
christopherbunn: Tests are in progress, I'll update this when they're done 😅

evalml/tests/automl_tests/test_automl.py
angela97lin: Yeah, that might help give this more clarity :)

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
angela97lin: Yeah, I'm curious how we're able to fit when the two input pipelines produce different number of predictions? I just want to clarify what's going on for my own sake 😂 

Using this example, let's say our dataset is 100 rows. The first input pipeline has a sampler that will oversample to 125 rows, while the second pipeline does not have any sampler. 

Is this not an issue because these two input pipelines are already fitted so the sampler does not modify the data, but instead we just call `predict` using the two input pipeline's estimators to create data for the metalearner? Is that why this isn't an issue? 🤔

angela97lin (commented):
Amazing! Let's sync about perf tests results, but thanks for addressing all my comments--I just left a few more for clarification and then 🚢

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
christopherbunn: Chatted with @angela97lin about this. We decided to just update this test case for now so both pipelines have an oversampler and to make detecting pipelines with a different amount of rows into an estimator a separate issue.

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
```diff
@@ -3,170 +3,177 @@
...
+def test_stacked_ensemble_keep_estimator_parameters(X_y_binary):
+    pytest.importorskip(
+        "imblearn.over_sampling",
+        reason="Skipping nondefault y test because imblearn not installed",
+    )
+    X, y = X_y_binary
+    input_pipelines = [
+        BinaryClassificationPipeline(
+            {
+                "OS": ["Oversampler", "X", "y"],
```
angela97lin, line 131: Omega nitpick but if we're just testing that the ensemble parameters are kept, we don't need to use the oversampler which only runs when it's imported right? We can use any of our other components? :)

evalml/tests/component_tests/test_stacked_ensemble_regressor.py
```diff
@@ -4,191 +4,232 @@
...
+        "Linear Regressor": [
+            "Linear Regressor",
+            "Random Forest.x",
+            "Elastic Net.x",
+            "y",
+        ],
+    }
+
+    RegressionPipeline(component_graph=graph_en)
+    RegressionPipeline(component_graph=graph_rf)
```
angela97lin, line 154: Unnecessary lines? 😂

angela97lin: Also, this might be incorrectly named? :P Since the final estimator is a linear regressor...

evalml/tests/component_tests/test_stacked_ensemble_regressor.py
```diff
@@ -4,191 +4,232 @@
...
+
+    assert "Linear Pipeline - Imputer" in pipeline.component_graph.compute_order
+    assert "Linear Pipeline 2 - Imputer" in pipeline.component_graph.compute_order
+
+    assert "Linear Pipeline - Elastic Net" in pipeline.component_graph.compute_order
+    assert "Linear Pipeline 2 - Elastic Net" in pipeline.component_graph.compute_order
+    assert "Linear Pipeline - EN" in pipeline.component_graph.compute_order
+
+
+def test_automl_ensembler_str_and_classes():
```
angela97lin, line 175: Probably don't need automl in this name? Even though it seems a little silly, maybe we should have a stacked ensemble classifier with this test too! (aka mirror tests in regression / classification when possible)

angela97lin (approved):
Looks great! Left a couple last nit-picky comments 😁  

The performance results are interesting. I'm glad to see that we're not incurring a big change in runtime, but I wonder why we're not saving in runtime by not running CV (hence prompting my question below about retraining). The performance flip-flopping makes me think that we'll need to file some additional issues to increase the performance of the ensemble via some spikes / more investigation--can you file an issue for that, if there isn't already? :) But as a first step into implementing our own ensemble, this is massive work! Great job 👏 

I have two final big things / questions / tests:
- Are we *actually* not retraining the input pipelines? If yes, can we add a test to confirm this? If no, is it in the scope of this PR or are we pushing it as a separate issue?
- Can we confirm that the meta learner’s input for `fitting` is the predictions generated by the input estimators? Could be as simple as mocking the final estimator, and then confirming the arguments passed in are predictions (and not accidentally the input X to the input pipelines or something)

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
christopherbunn: True, I'll update this to use an imputer instead!

christopherbunn:
Ty @angela97lin! To answer your questions: 

>The performance flip-flopping makes me think that we'll need to file some additional issues to increase the performance of the ensemble via some spikes / more investigation--can you file an issue for that, if there isn't already?

I filed [this issue](https://github.com/alteryx/evalml/issues/2805), which tracks tuning the metalearner's hyperparameters. I think this will be one of the most impactful changes to improving performance. I also filed this issue for other possible ideas [here](https://github.com/alteryx/evalml/issues/2835).  

> I'm glad to see that we're not incurring a big change in runtime, but I wonder why we're not saving in runtime by not running CV (hence prompting my question below about retraining)
> Are we actually not retraining the input pipelines? If yes, can we add a test to confirm this? If no, is it in the scope of this PR or are we pushing it as a separate issue?

We are currently not retraining the input pipelines. We're pushing this to another issue for the future. It is mentioned [here](https://github.com/alteryx/evalml/issues/2835) as another possible path for improving performance. I also noted what we have tried and why that did not work.

> Can we confirm that the meta learner’s input for fitting is the predictions generated by the input estimators? Could be as simple as mocking the final estimator, and then confirming the arguments passed in are predictions (and not accidentally the input X to the input pipelines or something)

Updated to include `test_ensembler_use_component_preds`, `test_ensembler_use_component_preds_binary`, and `test_ensembler_use_component_preds_multi` that tests for this.

PR #2661, opened 2021-08-19
Switch order of boolean checks in infer_feature_types

### LINKED ISSUE #2642: Spike: Investigate increase in fit time for `KDDCup` dataset
In our 0.31.1 release performance tests, @chukarsten noted that the `KDDCup` dataset is 10% slower. https://alteryx.atlassian.net/wiki/spaces/PS/pages/975536762/EvalML+v0.30.0+-+v0.30.1+Upgrade

@chukarsten Thinks it's due to the changes we made to `infer_feature_types` in #2610 and I agree.

That being said, someone should take a closer look at the `KDDCup` dataset and find the real culprit. If `infer_feature_types` is causing the slowdown, we should speed it up. A 10% penalty in performance is too steep a price to pay for the corner case of all null columns.

freddyaboulton:
### Pull Request Description
Fixes #2642 

I was able to track the regression down to `infer_feature_types` like @chukarsten  and I suspected.

Steps to identify problem:

1. Profile this script locally for both v0.30.0 and main
```python
import pandas as pd

X = pd.read_csv("/Users/freddy.boulton/Downloads/KDDCup09_churn.csv")
y = X.pop("CHURN")

X_train, X_validation, y_train, y_validation = split_data(X, y, problem_type="binary",test_size=0.5, random_seed=0)
automl = AutoMLSearch(X_train, y_train, "binary")
automl.search()
```

2. Visualize with snakeviz. I saw that main took 5 minutes 3 seconds while v0.30.0 took 4 minutes 30 seconds. There's pretty much an exact 30 second increase in total time spent in `infer_feature_types` corresponding to `is_pd_na`

### main
image

### 0.30.0
image

3. We need to perform `is_pd_na` but the trick is to only do it when the column is `Unknown`, that's the only case we care about. In python, the booleans are evaluated from left to right so we'll put the `is_column_unknown` check first since that one is near instant. Example:

Screen Shot 2021-08-19 at 12 00 41 PM

4. With the new order, `infer_feature_types` now only takes 10 seconds longer compared to 0.30.0. This is coming from running `convert_all_nan_unknown_to_double` and checking if the columns are Unknown. 

image

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,7 @@ Release Notes
         * Updated the high variance check in AutoMLSearch to be robust to a variety of objectives and cv scores :pr:`2622`
         * Use Woodwork's outlier detection for the ``OutliersDataCheck`` :pr:`2637`
         * Added ability to utilize instantiated components when creating a pipeline :pr:`2643`
+        * Sped up the all Nan and unknown check in ``infer_feature_types`` :pr:`2661`
     * Fixes
     * Changes
         * Deleted ``_put_into_original_order`` helper function :pr:`2639`
--- a/evalml/utils/woodwork_utils.py
+++ b/evalml/utils/woodwork_utils.py
@@ -66,7 +66,7 @@ def infer_feature_types(data, feature_types=None):
 
     def convert_all_nan_unknown_to_double(data):
         def is_column_pd_na(data, col):
-            return all(data[col].isna())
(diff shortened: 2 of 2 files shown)


chukarsten (approved):
LGTM!  Thanks for looking into this!

angela97lin (approved):
This is really cool and as always, your writeup is super clear. LGTM! 🚢 👏

PR #2662, opened 2021-08-19
Raise `ValueError` when user attempts to use .y for a component that does not produce a y output

### LINKED ISSUE #2569: Define behavior when user attempts to use .y for a component that does not produce a y output
We should define the behavior when user attempts to use .y for a component that does not produce a y output. For example, if the user specifies:
```
ComponentGraph(
            {
                "Imputer": [Imputer, "X", "y"],
                "One Hot Encoder": [OneHotEncoder, "Imputer.x", "Imputer.y"],
                "Random Forest Classifier": [
                    RandomForestClassifier,
                    "One Hot Encoder.x",
                    "One Hot Encoder.y",
                ],
            }
        )
```
using `Imputer.y` when the Imputer does not produce a y output, what should happen? I think we should error out given that the graph is not properly formed. Note that this might not be necessary if we tackle #1945, but this issue tracks ensuring proper behavior, as well as adding the appropriate tests.

angela97lin:
Closes #2569

### DIFF
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -65,45 +65,8 @@ class ComponentGraph:
                 raise ValueError(
                     "All component information should be passed in as a list"
                 )
-        # for _, component_inputs in self.component_dict.items():
-        #     component_inputs = component_inputs[1:]
-        #     has_feature_input = any(
-        #         component_input.endswith(".x") or component_input == "X"
-        #         for component_input in component_inputs
-        #     )
-        #     num_target_inputs = sum(
-        #         component_input.endswith(".y") or component_input == "y"
-        #         for component_input in component_inputs
-        #     )
-        #     if not has_feature_input:
-        #         raise ValueError(
-        #             "All components must have at least one input feature (.x/X) edge."
-        #         )
-        #     if num_target_inputs != 1:
-        #         raise ValueError(
-        #             "All components must have exactly one target (.y/y) edge."
-        #         )
-
-        #     def check_all_inputs_have_correct_syntax(edge):
-        #         return not (
-        #             edge.endswith(".y")
-        #             or edge == "y"
-        #             or edge.endswith(".x")
-        #             or edge == "X"
-        #         )
-
-        #     if (
-        #         len(
-        #             list(filter(check_all_inputs_have_correct_syntax, component_inputs))
-        #         )
-        #         != 0
-        #     ):
-        #         raise ValueError(
-        #             "All edges must be specified as either an input feature ('X'/.x) or input target ('y'/.y)."
-        #         )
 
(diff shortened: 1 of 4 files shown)


evalml/pipelines/component_graph.py
```diff
@@ -45,11 +45,15 @@ def __init__(self, component_dict=None, random_seed=0):
...
             )
         self._validate_component_dict()
+
         self.component_instances = {}
         self._is_instantiated = False
         for component_name, component_info in self.component_dict.items():
             component_class = handle_component_class(component_info[0])
             self.component_instances[component_name] = component_class
+
+        self._validate_component_dict_edges()
```
angela97lin, line 55: Since this requires instantiation of components (to more easily handle detecting if the component is a subclass of a target transformer that would return a tuple), I had to create a separate helper method that was separate from `self._validate_component_dict()`

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -230,7 +230,7 @@ def test_init_bad_graphs():
 def test_order_x_and_y():
     graph = {
         "Imputer": [Imputer, "X", "y"],
-        "OHE": [OneHotEncoder, "Imputer.x", "Imputer.y"],
+        "OHE": [OneHotEncoder, "Imputer.x", "y"],
```
angela97lin, line 233: The new logic catches some cases where we created a graph and used a component's .y when it doesn't produce such!

eccabay (approved):
LGTM!

evalml/pipelines/component_graph.py
```diff
@@ -97,6 +104,20 @@ def check_all_inputs_have_correct_syntax(edge):
...
+            target_inputs = [
+                component
+                for component in component_inputs
+                if (component.endswith(".y"))
+            ]
+            if target_inputs:
+                target_component_name = target_inputs[0][:-2]
+                target_component_class = self.get_component(target_component_name)
+                if (
+                    not issubclass(target_component_class, TargetTransformer)
```
freddyaboulton, line 116: Why not just use `not target_component_class.modifies_target`? All TargetTransformers modify the target by definition. 

In fact, I think we can get rid of the TargetTransformer abstraction in the future. I added that because at the time our component graph would allow you to specify `.y` for a component that did not modify the target so the component graph `inverse_transform` could error out. I will file an issue.

freddyaboulton (approved):
Thank you @angela97lin !

bchen1116 (approved):
LGTM! Agreed with Freddy's comments!

evalml/pipelines/component_graph.py
angela97lin: I think you're absolutely right--thanks for filing the issue! 😁

PR #2668, opened 2021-08-20

machineFL:

### DIFF
--- a/evalml/tests/dependency_update_check/latest_dependency_versions.txt
+++ b/evalml/tests/dependency_update_check/latest_dependency_versions.txt
@@ -13,7 +13,8 @@ matplotlib==3.4.3
 matplotlib-inline==0.1.2
 networkx==2.5.1
 nlp-primitives==1.1.0
-numpy==1.20.3
+numpy==1.21.2
+numba==0.53.0
 pandas==1.3.2
 plotly==5.2.1
 pmdarima==1.8.0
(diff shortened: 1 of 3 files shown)


evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -13,7 +13,7 @@ matplotlib==3.4.3
 matplotlib-inline==0.1.2
 networkx==2.5.1
 nlp-primitives==1.1.0
-numpy==1.21.2
```
freddyaboulton, line 16: This seems like something we should look into? The latest dask release downgraded numpy? 

That's not clear from their [setup script](https://github.com/dask/dask/blob/2021.08.1/setup.py) or [changelog](https://docs.dask.org/en/latest/changelog.html)

angela97lin: Odd, based on the commit history though, the numpy change happened before the dask one:
https://github.com/alteryx/evalml/commit/f4c5e8604ae63bf8cceb10df6507bd7757e2aba2
https://github.com/alteryx/evalml/commit/605540391f7fd26d625f944d7fc055dc547a9656

freddyaboulton: I think I figured it out - [numba](https://pypi.org/project/numba/0.54.0/) released on friday and they do not support 1.21.0. Numba gets pulled in via [shap](https://github.com/slundberg/shap/blob/master/setup.py) and they pull in the latest numba. 

I guess that means we only support numpy 1.20.0. FYI @alteryx/evalml

chukarsten (approved):

PR #2701, opened 2021-08-26
Add label encoder to XGBoost

### LINKED ISSUE #2517: Add label encoder to XGBoostClassifier
I currently get the following DeprecationWarning with XGBoost==1.4.2:
```
  The use of label encoder in XGBClassifier is deprecated and will be removed in a future release. To remove this warning, do the following: 1) Pass option use_label_encoder=False when constructing XGBClassifier object; and 2) Encode your labels (y) as integers starting with 0, i.e. 0, 1, 2, ..., [num_class - 1].
```
Since I know we want to upgrade XGBoost to >1.4.2 for the pandas 1.3.0 release, it could be a good idea to address this too, sooner rather than later :)

bchen1116:
fix #2517 

Now, we have this in search:
image

As opposed to these warnings from before:
image

**Question**: Should I address the remaining warning message (specifying 'error' versus 'logloss' as the default objective for XGBoost, or should a separate issue be filed?


chukarsten (approved):

evalml/tests/component_tests/test_components.py
```diff
@@ -1289,6 +1289,8 @@ def test_estimators_accept_all_kwargs(
         params = estimator._component_obj.get_params()
         if "random_state" in params:
             del params["random_state"]
+        if estimator.name == "XGBoost Classifier":
```
freddyaboulton, line 1292: nit-pick: Why not pop `use_label_encoder` from the kwargs rather than changing the test?

evalml/pipelines/components/estimators/classifiers/xgboost_classifier.py
```diff
@@ -88,11 +94,18 @@ def _convert_bool_to_int(X):
...
+            self._label_encoder = LabelEncoder()
+            y = pd.Series(self._label_encoder.fit_transform(y), dtype="int64")
+        return y
+
     def fit(self, X, y=None):
         X, y = super()._manage_woodwork(X, y)
         X.ww.set_types(self._convert_bool_to_int(X))
         self.input_feature_names = list(X.columns)
         X = _rename_column_names_to_numeric(X, flatten_tuples=False)
+        y = self._label_encode(y)
```
freddyaboulton, line 108: I think we need to decode the predictions if we've encoded them?

freddyaboulton (approved):
Thanks @bchen1116 ! Left a couple of questions about decoding but I'm glad we won't be seeing that warning anymore!

evalml/pipelines/components/estimators/classifiers/xgboost_classifier.py
bchen1116: Good call!

PR #2713, opened 2021-08-31
Update installation docs for Prophet

### LINKED ISSUE #2688: Update docs for Prophet Installation
Update installation instructions for users to install Prophet

ParthivNaresh:
Fixes #2688


README.md
```diff
@@ -30,6 +30,23 @@ Prophet is currently only supported via pip installation in EvalML.
 ```shell
 pip install evalml[prophet]
 ```
+Another option for installing Prophet with CmdStan as a backend is to use `make installdeps-prophet`.
```
gsheni, line 33: You might want to think about moving this instructions to **Install page on the docs**. This currently will make the README.md much longer. You could link from the README.md to the Install documentation page.

We put graphviz install instructions in our Install docs:
https://featuretools.alteryx.com/en/latest/install.html#installing-graphviz

ParthivNaresh: @gsheni Thanks for the recommendation!

Makefile
```diff
@@ -70,6 +70,14 @@ installdeps:
 	pip install --upgrade pip -q
 	pip install -e . -q
 
+SITE_PACKAGES_DIR=$$(python -c 'import site; print(site.getsitepackages()[0])')
```
freddyaboulton, line 73: When I first tried this out in a fresh environment, I got the infamous `no such file ...prophet_model.bin` error when I tried running the unit tests. 

I was really confused about why this happened, but when I looked at the pip install, I saw:

```
Collecting prophet==1.0.1
  Using cached prophet-1.0.1-py3-none-any.whl
```

This stood out to me because as we all know by now, prophet does not publish any wheels! So I was confused why we weren't building from source.

Then I read the [pip documentation](https://pip.pypa.io/en/stable/topics/caching/#locally-built-wheels), and I learned that locally built wheels are added to the cache. 

So my guess as to what went wrong for me is that I had installed prophet beforehand without a cmdstan backend and then subsequent installs with a cmdstan backend were using the cached version so it was not properly setting up cmdstan.

So in short, I think we need to do `pip install --no-cache-dir prophet==1.0.1` instead. That properly set up the cmdstan backend and I was able to get the unit tests to pass.

 Apparently, `--no-cache-dir` is discouraged, but the [workaround in the pip docs](https://pip.pypa.io/en/stable/topics/caching/#avoiding-caching) did not work for me.

README.md
freddyaboulton: I also agree with this recommendation @ParthivNaresh !

README.md
```diff
@@ -30,6 +30,23 @@ Prophet is currently only supported via pip installation in EvalML.
...
+This command will do the following:
```
angela97lin, line 34: As a beginner user, I'm a bit confused on how to use this command. This assumes that a user actually has the evalml repo; that is, I can't just create a new virtual env, pip install evalml, and then run this. I'd need to have cloned the evalml repo and run this while in the evalml repo, correct?

If that's true, we should make that clear! (though also makes me question how often people would take this route)

chukarsten (approved):
This seemed clear and it worked for me.

freddyaboulton (approved):
Thank you @ParthivNaresh !

Issue #2705, opened 2021-08-27
Add `DropNaNRowsTransformer`


angela97lin:
https://github.com/alteryx/evalml/pull/2692 introduced a generic `DropRowsTransformer` but based on the thread [here](https://github.com/alteryx/evalml/issues/2100#issuecomment-905875874), it'd be a good idea to introduce a data-agnostic component which detects and drops nan rows.

This could probably be done by subclassing `DropRowsTransformer` and adding nan detection logic!

angela97lin:
One pro of this approach is that we don't need to store the indices to remove in the output of the data check, since we need to recalculate here. This means that it is incredibly important that the two methods of detecting nan rows are at parity at all times.

PR #2737, opened 2021-09-01
Delete `drop_nan_target_rows` util method

### LINKED ISSUE #2684: Delete `drop_nan_target_rows` util method
There is a preprocessing util called `drop_nan_target_rows` but it isn't used anywhere. I'm not sure how useful this is, but it sounds like it should be replaced by data check action work. This issue tracks removing `drop_nan_target_rows`!

angela97lin:
Closes #2684


jeremyliweishih:
instead of just deleting it, is there any value in changing it into a component?

angela97lin:
> instead of just deleting it, is there any value in changing it into a component?

Good question, we have https://github.com/alteryx/evalml/issues/2705 to track that :) (Not explicitly dropping just NaN targets, but also rows in the features too)

jeremyliweishih:
@angela97lin sounds good! it might be good to link this PR to that issue if we ever want to repurpose some of the code.

jeremyliweishih (approved):
👏

chukarsten (approved):
Love it and love stories like this.

Issue #2744, opened 2021-09-03
Modify our code to initialize woodwork with partial schemas


freddyaboulton:
Woodwork introduced the ability to initialize typing information with a partial schema in https://github.com/alteryx/woodwork/pull/1100

This means that we can potentially get rid of `_retain_custom_types_and_intialize_woodwork` and therefore retain more than just the logical types after some components modify the input data in a way that breaks the woodwork schema.

For example, this is how the `TargetEncoder` would look:

```python
    X_t_df = pd.DataFrame(X_t, columns=X_ww.columns, index=X_ww.index)
    cat_columns = X_ww.ww.select("categorical", return_schema=True).columns
    columns = [c for c in X_t_df.columns if c not in cat_columns]
    # Infer types on newly created columns
    X_t_df.ww.init(schema=original_schema._get_subset_schema(columns))
```

chukarsten:
As acceptance criteria for completing this issue, let's evaluate the ability to remove `_retain_custom_types_and_intialize_woodwork` and refactor using the new WW capabilities.

chukarsten:
@freddyaboulton do we need to look at the fix in [#2752](https://app.zenhub.com/workspaces/evalml-5d7698439dde650001eaf87c/issues/alteryx/evalml/2752) to refactor to use this as well?

freddyaboulton:
@chukarsten If you're asking if we're going to have to revisit the `Imputer` as part of this refactor, then you're spot on. I think anywhere `_retain_custom_types_and_initalize_woodwork` is in scope for this issue!

PR #2752, opened 2021-09-07
Fix Imputer so that it does not erase ww info during transform

### LINKED ISSUE #2751: Imputer.transform erases woodwork typing information
Repro

```python
import pandas as pd
import woodwork as ww
from evalml.pipelines.components import Imputer
import pytest

df_train = pd.DataFrame({"a": [1, 2, 3, 2], "b": ["a", "b", "b", "c"]})
df_holdout = pd.DataFrame({"a": [2], "b": [None]})
df_train.ww.init(logical_types={"a": "Double", "b": "Categorical"})
df_holdout.ww.init(logical_types={"a": "Double", "b": "Categorical"})

imputer = Imputer()
imputer.fit(df_train, None)

with pytest.raises(ValueError, match="could not convert string to float: 'b'"):
    imputer.transform(df_holdout, None)
```

The problem is that the imputer selects the categorical columns but it does not go through the ww accessor to do so. This means that the `SimpleImputer` used for categoricals has to do type inference. Since `b` is all Nulls, the inferred type is `Unknown`, which causes `infer_feature_types` to convert it to a double logical type. 

If the imputer selected the categorical columns through the ww accessor, the `SimpleImputer` would not infer `b` as `Unknown` so we would not try to convert `b` to `Double`.

freddyaboulton:
### Pull Request Description
Fixes #2751 

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
         * Set ``eval_metric`` to ``logloss`` for ``XGBoostClassifier`` :pr:`2741`
         * Added support for ``woodwork`` versions ``0.7.0`` and ``0.7.1`` :pr:`2743`
     * Fixes
+        * Fixed bug where ``Imputer.transform`` would erase ww typing information prior to handing data to the ``SimpleImputer`` :pr:`2752`
     * Changes
     * Documentation Changes
         * Specified installation steps for Prophet :pr:`2713`
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -123,15 +123,17 @@ class Imputer(Transformer):
             df = pd.DataFrame(index=X.index)
             return _retain_custom_types_and_initalize_woodwork(original_ltypes, df)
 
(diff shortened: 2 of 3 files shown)


chukarsten (approved):
Nice!  Thanks Freddy!

angela97lin (approved):
Looks good, thanks for filing and fixing this! 

This, along with the few other bugs that have popped up re ww typing, makes me wonder what we could do to make this easier--right now, our components could have potentially different behavior if they weren't fully updated to use ww like the imputer was. No real action, just food for thought 😅

freddyaboulton:
@angela97lin Agreed that it's not comforting to uncover these bugs one-by-one. I will say that the reason that the Imputer did not go through the ww accessor prior to handing the data to the simple imputer is that it didn't need to! The ww type inference used to be almost entirely dependent on the pandas dtype so the simple imputer inference would match the Imputer logical types. Maybe it was always a bug waiting to happen but this implementation got us pretty far. 

What I'm saying is that as ww inference evolves and matures so will our code because new corner cases that couldn't exist before will begin to exist. 

I think it could help if we always went through the accessor and never "broke" the schema, i think https://github.com/alteryx/evalml/issues/2744 can help with that.

PR #2753, opened 2021-09-08
Catch only `ParameterNotUsed` warnings in `_catch_warnings`

### LINKED ISSUE #2652: AutoMLSearch raises AttributeError when TargetDistributionDataCheck issues UserWarning and only one estimator family is allowed
Repro

```python
import pandas as pd
import woodwork as ww
import pytest

X = pd.read_csv("/Users/freddy.boulton/sources/tempo2-server/testData/stones_encoded.csv")
y = X.pop("total_sales_price")
X_train, X_test, y_train, y_test = split_data(X, y, "regression")

with pytest.raises(AttributeError, match="'UserWarning' object has no attribute 'components'"):
    automl = AutoMLSearch(X_train, y_train, "regression", allowed_model_families=["xgboost"],
                      max_batches=3)
```

The problem is that the `TargetDistributionDataCheck` (called during make_pipeline) is issuing a UserWarning about the p-value. This then gets caught by our [warnings filter](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L543) and then `_catch_warnings` tries to access the `components` attribute of the warning (which doesn't exist because it's not a ParameterNotUsedWarning). 

We need to improve `_catch_warnings` to only look at `ParameterNotUsedWarnings`.

eccabay:
Closes #2652


evalml/tests/automl_tests/test_automl.py
```diff
@@ -4945,6 +4945,29 @@ def test_pipeline_parameter_warnings_component_graphs(
         assert w[0].message.components == set_values
 
 
+@patch(
+    "evalml.data_checks.target_distribution_data_check.TargetDistributionDataCheck.validate"
+)
+def test_pipeline_parameter_warnings_only_parameternotused(
+    mock_data_check, X_y_regression
+):
+    mock_data_check.side_effect = UserWarning("Big bad warning")
```
angela97lin, line 4954: Love it 😂

angela97lin (approved):
Love it, LGTM! 😁

evalml/automl/automl_search.py
```diff
@@ -848,14 +848,18 @@ def close_engine(self):
...
-            # we find the value(s) that we must throw the warning for
+            # we find the value(s) that we must throw any ParameterNotUsedWarnings for
             final_message = set([])
             for idx, msg in enumerate(warning_list):
-                if idx == 0:
-                    final_message = final_message.union(msg.message.components)
-                else:
-                    final_message = final_message.intersection(msg.message.components)
-            warnings.warn(ParameterNotUsedWarning(final_message))
+                if isinstance(msg.message, ParameterNotUsedWarning):
```
freddyaboulton, line 854: There might be a bug here? What if the warning list is `[RandomWarning(...), ParameterNotUsedWarning(...)]`.
Then `idx` will be 1 when the `isinstance` check is true so we'll do the intersection of an empty set and the `ParameterNotUsedWarning`, so we'll wipe out the warning we want to log?

eccabay: Oh shoot, I just merged this but you're totally right 🤦‍♀️ 

It should be a pretty simple fix to replace the idx check with checking if the set is empty or not instead - I'll put up a PR for that shortly!

eccabay: Actually, there is a bug here, but it's not that! The check of `if len(warning_list) == len(self.allowed_pipelines) and len(warning_list) > 0:` would prevent the warning list of `[RandomWarning(...), ParameterNotUsedWarning(...)]` from ever reaching that point in the code, since in that case `len(warning_list != len(self.allowed_pipelines)`.

So in this case we would indeed fail to log the `ParameterNotUsedWarning`, but not for the reason highlighted. 

Also, this highlights another issue where we currently catch and then do not raise all non-`ParameterNotUsedWarning` warnings during `make_pipeline`. Some sleuthing found that PR #2564 added the warning catches where they were not before, so this behavior was unwanted but introduced then. I'm going to file an issue and put up a PR for raising all warnings here again, since the size of this change just grew.

PR #2755, opened 2021-09-08
Remove imblearn from Oversampler attribute

### LINKED ISSUE #2754: Remove `imblearn` module from being an attribute of `Oversampler`
@christopherbunn and I discovered that the merging of #2695 introduced a bug when attempting to copy automl objects to post results. 
image
image (1)
image (2)

This is because the new Oversampler class stores the `imblearn` module as an object during `Oversampler.__init__`, in order to assign the correct SMOTE class during `_get_best_oversampler`.

eccabay:
Closes #2754


christopherbunn (approved):
LGTM! Thanks for getting this in so quickly.

bchen1116 (approved):
LGTM! Nice catch!

PR #2757, opened 2021-09-09
Change default n_jobs value for XGBoost

### LINKED ISSUE #2437: Update default n_jobs used by XGBoost to not be -1
https://github.com/alteryx/evalml/pull/2410 updates XGBoost by exposing the `nthread` parameter passed to XGBoost as `n_jobs`. However, while profiling I noticed that the default value of nthreads=-1 (use all threads) performs slower than using 2 threads. Upon further testing, it seems like after a certain number of threads, the performance drops significantly. In my case, the performance dropped after 16 threads (probably because I have 8-cores, 2 threads per core).

XGBoost docs mention that thread contention could significantly slow down performance of the algorithm: https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.core

This issue tracks investigating this and determining if we should change our default value of -1 to something generally more performant. It alarms me that in my example, having just two threads cut the runtime of `fit` in half.

image

I initially had this issue tracking CatBoost too, but after running a few more tests, I think the CatBoost differences are just due to variance, and not too concerning.

bchen1116:
fix #2437 

Original Perf test results [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/881066744/XGBoost+njobs+parameter+testing?focusedCommentId=1034321996#comment-1034321996)


jeremyliweishih (approved):
good work!

PR #2765, opened 2021-09-10
Raise all warnings in `_catch_warnings`

### LINKED ISSUE #2761: `_catch_warnings` suppresses non-`ParameterNotUsed` warnings
PR #2753 fixed a bug where non-`ParameterNotUsed` warnings would break `_catch_warnings`, a method originally designed to collapse multiple `ParameterNotUsed` warnings into a single warning. However, a side effect of this was that any other types of warnings thrown in this section of code are now caught but never re-raised. 

Some discussion of this from [here](https://github.com/alteryx/evalml/pull/2753/files#r705657398)

eccabay:
Closes #2761


jeremyliweishih (approved):
looks good to me!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5102,6 +5102,25 @@ def test_pipeline_parameter_warnings_only_parameternotused(
     assert str(warning) == "Big bad warning"
 
 
+def test_pipeline_parameter_warnings_with_other_types():
```
bchen1116, line 5105: Is there test coverage on if we get multiple warnings with the same message? ie coverage for
```python
elif str(msg.message) not in raised_messages:
                warnings.warn(msg.message)
                raised_messages.append(str(msg.message))
```
?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5102,6 +5102,25 @@ def test_pipeline_parameter_warnings_only_parameternotused(
...
+    # Data that will trigger a UserWarning from `TargetLeakageDataCheck`
+    X = pd.DataFrame({"a": [1, 2, 2, 1, 2, 1, 2, 2, 1, 1] * 505})
+    y = [0.946, 0.972, 1.154, 0.954, 0.969, 1.222, 1.038, 0.999, 0.973, 0.897] * 505
+
+    with pytest.warns(None) as w:
+        AutoMLSearch(
+            X_train=X,
+            y_train=y,
+            problem_type="regression",
+            allowed_model_families=["random_forest", "linear_model"],
```
bchen1116, line 5115: We're using random forest and linear model families, so I think the params you passed for Decision Tree won't be used.

I think we can shorten this to just set `max_iterations=2` rather than defining `allowed_model_families`, `max_batches`, and `pipeline_parmaeters`

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5102,6 +5102,25 @@ def test_pipeline_parameter_warnings_only_parameternotused(
...
+            max_batches=1,
+            pipeline_parameters={"Decision Tree Classifier": {"max_depth": 1}},
+        )
+    assert len(w) == 2
+    assert isinstance(w[0].message, UserWarning)
+    assert isinstance(w[1].message, ParameterNotUsedWarning)
+
+
```
bchen1116, line 5123: Is there a test to see the logger outputs are expected when we allow `verbose=True` vs `False` in respect to these warnings with PNU?

bchen1116 (changes requested):
Great catch! I left a few nits and a few questions, plus a suggestion to improve and fix tests. Looks good otherwise!

evalml/tests/automl_tests/test_automl.py
eccabay: Yep, this test covers that! The `TargetLeakage` warning actually throws the same warning for each generated pipeline, just like the `ParameterNotUsed` warnings come up. This test is actually how I discovered we needed to filter duplicate warnings.

evalml/tests/automl_tests/test_automl.py
eccabay: The params for Decision Tree is part of the test, making sure that `ParameterNotUsed` warnings and other kinds of warnings that come up on the same run will both be thrown, since that was the original issue.

evalml/tests/automl_tests/test_automl.py
eccabay: Since these are warnings, they are logged to stdout regardless of whether verbose is True or False. Since search isn't running though (which would add a nontrivial amount of time), I can still add the check for redundancy's sake.

evalml/tests/automl_tests/test_automl.py
bchen1116: Wow that completely slipped my mind. Thanks for clarifying!

bchen1116 (approved):
LGTM once all tests pass! Thanks for clarifying!

PR #2776, opened 2021-09-13
Support training-only components in pipelines and component graphs

### LINKED ISSUE #2681: Support training-only components in ComponentGraph
Right now, we support samplers as training-only components by checking if the component is a sampler component and not taking any action if so: https://github.com/alteryx/evalml/blob/4afe6dde123c59386d792f17c07664f1c836bcfd/evalml/pipelines/component_graph.py#L323

Once we add other components that are train-only (ex: drop rows), we should verify our behavior holds for other components by updating this line to be more general.

angela97lin:
Closes #2681


evalml/pipelines/component_graph.py
```diff
@@ -288,7 +285,9 @@ def transform(self, X, y=None):
                 "Cannot call transform() on a component graph because the final component is not a Transformer."
             )
 
-        outputs = self._compute_features(self.compute_order, X, y, False)
+        outputs = self._compute_features(
+            self.compute_order, X, y, fit=False, evaluate_training=True
```
eccabay, line 289: I may be misunderstanding something here - why should `evaluate_training` be true during `transform`, but not during `fit`?

angela97lin: @eccabay Good question! I think the language here is a bit confusing, but `evaluate_training` is a flag for determining whether or not we want to compute "training-only" components, such as the samplers or the DropRowsComponent when we are *not* fitting. 

We decided that there should be a difference between `transform`, which takes the pipeline and calls `transform` on every component vs `predict`, where we want predictions on unseen data and should not evaluate training-only components. This enables us to keep our current behavior, but also allow users to use pipelines as just a sequence of transformations.

Does that make sense? (or did I confuse you further?) 😄

eccabay: Ah yeah, that makes a lot of sense. Thanks! 😃

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2754,3 +2755,80 @@ def test_pipeline_transform_with_final_estimator(
         ),
     ):
         pipeline.transform(X, y)
+
+
+@patch("evalml.pipelines.components.LogisticRegressionClassifier.fit")
```
freddyaboulton, line 2760: @angela97lin Can you please add coverage for `fit_features` and `compute_estimator_features`?

freddyaboulton: The expectation is that the training only components are run during `fit_features` and not `compute_estimator_features`

freddyaboulton (approved):
@angela97lin Thank you for doing this! I'm surprised this wasn't a bigger diff honestly! 

I left some coverage suggestions as well as a comment on @eccabay 's original question - maybe we can make the parameter name/docstring clearer about what's happening?

I do think this change (the addition of training-only components) is the kind of thing that will eventually merit a redesign/refactor of the component graph. I think right now we encode whether we should run training-only components via the method that's called:

1. `fit` - yes
2. `transform` - yes
3. `predict` - no
4. `compute_estimator_features` - no
5. `fit_features` - yes

That kind of behavior is not really clear from the method names and we have a lot of methods for generating features that it's hard to keep it all straight (`fit_features` vs `compute_estimator_features` vs `_fit_transform_features_helper` vs `_compute_features`) especially in light of the training-only requirement. 

I think this is good to merge but I wonder what your thoughts are on addressing the tech debt that I think has been accruing.

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2754,3 +2755,80 @@ def test_pipeline_transform_with_final_estimator(
...
+def test_training_only_component_in_pipeline_fit(mock_fit):
+    X = pd.DataFrame(
+        {
+            "a": [i for i in range(9)] + [np.nan],
+            "b": [i % 3 for i in range(10)],
+            "c": [i % 7 for i in range(10)],
+        }
+    )
+    y = pd.Series([0] * 5 + [1] * 5)
```
chukarsten, line 2769: Admittedly this is on my backburner's backburner - but are able to either use an existing X/y pair for these tests or create a module level test fixture for these additional tests?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2754,3 +2755,80 @@ def test_pipeline_transform_with_final_estimator(
...
+            "Logistic Regression Classifier": [
+                "Logistic Regression Classifier",
+                "Drop Rows Transformer.x",
+                "Drop Rows Transformer.y",
+            ],
+        },
+        parameters={"Drop Rows Transformer": {"indices_to_drop": [0, 9]}},
+    )
+    pipeline.fit(X, y)
+    assert len(mock_fit.call_args[0][0]) == 8
```
chukarsten, line 2783: There's a time when your brains are too big for me.  This is one of them, haha.  Why should this be true if the issue is completed successfully?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2754,3 +2755,80 @@ def test_pipeline_transform_with_final_estimator(
...
+                "Logistic Regression Classifier",
+                "Drop Rows Transformer.x",
+                "Drop Rows Transformer.y",
+            ],
+        },
+        parameters={"Drop Rows Transformer": {"indices_to_drop": [9]}},
+    )
+    pipeline.fit(X, y)
+    preds = pipeline.predict(X)
+    assert len(preds) == 10
```
chukarsten, line 2810: Is the jist here that no rows have been dropped during prediction since DropRowsTransformer is `evaluate_training=True`?

chukarsten (approved):
LGTM.  Just a nit about using the same X,y three times, but very clear overall.  Thanks for doing this!

evalml/pipelines/components/transformers/preprocessing/drop_rows_transformer.py
```diff
@@ -13,6 +13,7 @@ class DropRowsTransformer(Transformer):
 
     name = "Drop Rows Transformer"
     modifies_target = True
+    training_only = True
```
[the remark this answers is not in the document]
angela97lin: I've made this abstract, but `Transformer` still sets this value to False, so any subclass of `Transformer` will have a default value of False... which might defeat the purpose of having this as abstract 😅 

I *think* this is okay though, since most of our components will have `training_only` as False for now... but open to thoughts and concerns :)

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Hahaha, I think this issue is pretty confusing! But it's 8 here because during `fit` / training time, we evaluate all components, regardless of whether or not they're training_only. So the DropRowsTransformer will drop the two rows during fit --> a total of 8 rows left!

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Yup!

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Thank you for holding me accountable, done!

angela97lin:
Thank you for doing this! I'm surprised this wasn't a bigger diff honestly!

I left some coverage suggestions as well as a comment on @eccabay 's original question - maybe we can make the parameter name/docstring clearer about what's happening?

I do think this change (the addition of training-only components) is the kind of thing that will eventually merit a redesign/refactor of the component graph. I think right now we encode whether we should run training-only components via the method that's called:

@freddyaboulton I think your instinct is on point! It seems like there's a lot of confusion amongst everyone, which probably indicates that something is not super clear, either in the implementation or in the design 😅 

I renamed `evaluate_training` to `evaluate_training_only_components` to be as explicit as possible about what the flag is, and filed https://github.com/alteryx/evalml/issues/2795 with some of my thoughts about the different methods. TLDR is, I wonder if there's a need for all of these methods, when in reality, we've been using some for just time series work and internal implementations. I'll add test coverage in this PR for those methods regardless though since we still have them around 😁

evalml/tests/pipeline_tests/test_pipelines.py
angela97lin: Of course, great idea 😁

PR #2782, opened 2021-09-14
Fix permutation importance when pipeline has target transformer

### LINKED ISSUE #2781: Permutation importance not properly calculated when there is a target transformer
Repro
```python
from unittest.mock import PropertyMock, patch
from evalml.model_understanding.permutation_importance import calculate_permutation_importance_one_column

X, y = load_diabetes()
X_train, X_test, y_train, y_test = split_data(X, y, "regression")


pipeline = RegressionPipeline({"Log": ["Log Transformer", "X", "y"],
                               "RF": ["Random Forest Regressor", "X", "Log.y"]})


fast = calculate_permutation_importance_one_column(pipeline, X_test, y_test,
                                            "bmi", "R2", fast=True, precomputed_features=X_test)
slow = calculate_permutation_importance_one_column(pipeline, X_test, y_test,
                                            "bmi", "R2", fast=False)

assert round(fast, 3) != round(slow, 3)
print(f"Fast: {fast} vs Slow: {slow}")
```

```
Fast: 0.002107686575141976 vs Slow: 0.15713380334051302
```

freddyaboulton:
### Pull Request Description
Fixes #2781 

This can go in the release in two weeks. Just opening up for review. 

-----

### DIFF
--- a/evalml/model_understanding/permutation_importance.py
+++ b/evalml/model_understanding/permutation_importance.py
@@ -325,5 +325,6 @@ def _fast_scorer(pipeline, features, X, y, objective):
         preds = pipeline.estimator.predict_proba(features)
     else:
         preds = pipeline.estimator.predict(features)
+    preds = pipeline.inverse_transform(preds)
     score = pipeline._score(X, y, preds, objective)
     return score if objective.greater_is_better else -score
--- a/evalml/tests/model_understanding_tests/test_permutation_importance.py
+++ b/evalml/tests/model_understanding_tests/test_permutation_importance.py
@@ -8,7 +8,11 @@ from evalml.model_understanding.permutation_importance import (
     calculate_permutation_importance,
     calculate_permutation_importance_one_column,
 )
-from evalml.pipelines import BinaryClassificationPipeline, Transformer
+from evalml.pipelines import (
+    BinaryClassificationPipeline,
+    RegressionPipeline,
+    Transformer,
+)
 from evalml.pipelines.components import (
     PCA,
     DateTimeFeaturizer,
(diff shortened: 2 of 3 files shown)


evalml/model_understanding/permutation_importance.py
```diff
@@ -325,5 +325,6 @@ def _fast_scorer(pipeline, features, X, y, objective):
         preds = pipeline.estimator.predict_proba(features)
     else:
         preds = pipeline.estimator.predict(features)
+    preds = pipeline.inverse_transform(preds)
```
angela97lin, line 328: If I'm following correctly, we only need to change the fast_scorer because in the slow_scorer we call score --> predict --> predict_in_sample which does inverse_transform, correct?

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -343,14 +363,17 @@ def test_fast_permutation_importance_matches_slow_output(
         "region",
         "amount",
     ]:
+        if col == "amount" and pipeline_class == PipelineWithTargetTransformer:
```
angela97lin, line 366: Interesting way to use a classification dataset for regression 😂

angela97lin (approved):
Looks good to me, thank you! I'm curious how you found this bug and determined which permutation importance value was correct 😂

evalml/model_understanding/permutation_importance.py
freddyaboulton: Yea - this is why the slow score is the correct one. In the "slow" method we basically treat the whole pipeline as an estimator. So we call `pipeline.predict`, which does `pipeline.inverse_transform`. In the fast method, we optimize by only doing the feature engineering once and get the predictions by doing `estimator.predict`. Since the estimator has no idea it's part of a pipeline, we need to remember to call `inverse_transform`.

chukarsten (approved):
Looks good to me!

PR #2794, opened 2021-09-16
Delete EmptyDataChecks

### LINKED ISSUE #2702: Remove `EmptyDataChecks` class
`EmptyDataChecks` was a class we used when data checks ran as part of AutoMLSearch. It was used to indicate that users did not want data checks to run. Now that data checks are separated from automl, there's no use for this class, so we should delete it!

### LINKED ISSUE #2703: Have fit_features return the transformed target
This came up during #2697

Our time series pipelines rely on `compute_estimator_features` to transform the data, then do some custom-time-series specific data transformations before passing the data to the estimator. 

Excerpt from our [time series regression pipeline](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/time_series_regression_pipeline.py#L45):

```python
        features = self.compute_estimator_features(X, y)
        features_no_nan, y = drop_rows_with_nans(features, y)
        predictions = self._estimator_predict(features_no_nan, y)
```

This means that the detrender is not working as expected for time series regression pipelines.

This touches more than just time series pipelines. We use `compute_estimator_features` in our `permutation_importance` optimization [algorithm](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/permutation_importance.py#L37). 

This means that the estimator is outputting transformed predictions but it's being compared against the untransformed target variable in [scoring](https://github.com/alteryx/evalml/blob/main/evalml/model_understanding/permutation_importance.py#L313). This means our permutation importance algorithm is not accurate for pipelines with target transformers (not including over/undersamplers)

freddyaboulton:
### Pull Request Description
Fixes #2702

-----


jeremyliweishih (approved):
nice!

angela97lin (approved):
Thank you! 😁

chukarsten (approved):
#canceled

PR #2796, opened 2021-09-17
Add default None arg to ComponentGraph.instantiate

### LINKED ISSUE #2698: Set `None` as a default value for `ComponentGraph.instantiate`
Right now, the signature for `ComponentGraph`'s `instantiate` method is:
```    
def instantiate(self, parameters):
```

In order to not pass any specific parameters, the user has to pass an empty dictionary or None. We should make this a keyword argument with a default of None:
```    
def instantiate(self, parameters=None):
```

bchen1116:
fix #2698


angela97lin (approved):
Love it! Thank you 😁

Issue #2679, opened 2021-08-23
Add ``preprocessing_components_to_prepend`` to AutoMLSearch


angela97lin:
We should support a parameter of preprocessing components to prepend to each pipeline in AutoMLSearch, enabling components that are produced as part of data check actions that should be applied at both train and predict time.

PR #2806, opened 2021-09-17
Remove data check for log transformation in `make_pipeline`

### LINKED ISSUE #2601: Use data check actions to apply log transformer
#2487  added a data check for when we should apply a log transform to a regression target. It also updated automl to run that data check under the hood, and use the output to automatically add log transform to the pipeline when applicable. This is useful and really cool!

We should find a way to use the data check actions API to do this instead of hard-coding it into our `make_pipeline` method. This will allow users to have control over when log transform will or will not be added to the pipeline. It will make it easier for projects which have integrated with the data check actions pattern to do so as well. And it will also make it easy for us to update our automl algo in the future if we want to try different things, like trying pipelines both with and without log transform.

angela97lin:
Closes #2601. #2679 tracks adding a parameter which would enable users to add preprocessing components to automl search, which users can use to pass the log transformer recommended by an action!

Note that while this may have some scoring implications that may lead to a decrease in performance for datasets with log distributions, this better aligns with the API / flow we have for data check actions.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
     * Enhancements
         * Added validation that ``training_data`` and ``training_target`` are not ``None`` in prediction explanations :pr:`2787`
         * Added support for training-only components in pipelines and component graphs :pr:`2776`
+        * Added default argument for the parameters value for ``ComponentGraph.instantiate`` :pr:`2796`
     * Fixes
         * Fixed bug where ``calculate_permutation_importance`` was not calculating the right value for pipelines with target transformers :pr:`2782`
         * Fixed bug where transformed target values were not used in ``fit`` for time series pipelines :pr:`2780`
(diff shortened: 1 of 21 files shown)


ParthivNaresh (approved):
My beautiful data check 😭  just kidding, this is a good cleanup and data check actions are definitely the way to go if we want to leverage any of these data check outcomes in pipelines in the future.

docs/source/release_notes.rst
```diff
@@ -14,6 +14,7 @@ Release Notes
         * Fixed bug where ``score_pipelines`` method of ``AutoMLSearch`` would not work for time series problems :pr:`2786`
     * Changes
         * Deleted ``EmptyDataChecks`` class :pr:`2794`
+        * Removed data check for log distributions in ``make_pipeline`` :pr:`2806`
```
freddyaboulton, line 17: I think this is technically a breaking change.

freddyaboulton (approved):
Thank you @angela97lin !

bchen1116 (approved):
LGTM!

docs/source/release_notes.rst
angela97lin: I always think of a breaking change as something that would cause our API to change, not the implementation, but I'm down to add it so that we're loud about the behavior of AutoML changing 😁

eccabay (approved):
Nice! I really like the solution you found with the warnings test 😄

PR #2812, opened 2021-09-20
Return JSON from Pipeline DAG structure

### LINKED ISSUE #1969: Pipeline graph: return JSON with nodes and edges
If we define a method which returns a list of nodes stating the component names and parameters, and edges stating directed connections between the components, users can build visualizations with a variety of different graphical tools.

This would be a great complement to `PipelineBase.graph()`. We've already separated visualization and data generation for our other model understanding tools.

ParthivNaresh:
Fixes #1969


evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,60 @@ def feature_importance(self):
         df = pd.DataFrame(importance, columns=["feature", "importance"])
         return df
 
+    def graph_json(self):
+        """Generates a JSON with nodes consisting of the component names and parameters, and edges detailing component relationships.
+
+        Returns:
+            dag_json (str): A serialized JSON representation of a DAG structure.
+        """
+        dag_json = {"Nodes": {"X": "X", "y": "y"}, "Edges": {}}
```
ParthivNaresh, line 435: Added X and y as default nodes

evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,60 @@ def feature_importance(self):
...
+            component_name,
+            component_info,
+        ) in self.component_graph.component_dict.items():
+            dag_json["Nodes"][component_name] = {
+                "class": component_info[0]
+                if isinstance(component_info[0], str)
+                else component_info[0].name,
+                "attributes": self.parameters[component_name]
+                if component_name in self.parameters
+                else None,
```
ParthivNaresh, line 447: Some classes don't have any parameters that get passed through like `StandardScaler`

evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,60 @@ def feature_importance(self):
...
+            component_name,
+            component_info,
+        ) in self.component_graph.component_dict.items():
+            for input_ in component_info[1:]:
+                from_comp = input_[:-2]
+                if input_.endswith(".x"):
+                    dag_json["Edges"][f"from_{from_comp}_to_{component_name}"] = {
+                        "from": from_comp,
+                        "to": component_name,
+                        "data": f"X modified by {from_comp}",
```
ParthivNaresh, line 460: `data` is meant to provide a little more detail as to what's being passed through that edge even though it can be inferred based on the `from` and `to` fields. OS users might find this useful in understanding this output.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5214,3 +5215,44 @@ def test_automl_chooses_engine(engine_choice, X_y_binary):
...
+        problem_type="multiclass",
+        allowed_component_graphs={"Name_0": ComponentGraph(component_graph)},
+        pipeline_parameters=automl_parameters_,
+        max_batches=2,
+    )
+
+    dag_str = automl.allowed_pipelines[0].graph_json()
+    dag_json = json.loads(dag_str)
+    for node_, params_ in automl_parameters_.items():
+        for key_, val_ in params_.items():
```
ParthivNaresh, line 5254: The specific structure is tested in `test_graphs`, this is primarily to check the `self.parameters` as they get passed through the pipeline

evalml/tests/conftest.py
```diff
@@ -395,6 +395,25 @@ def example_graph():
     return component_graph
 
 
+@pytest.fixture
+def example_pass_target_graph():
+    component_graph = {
+        "Imputer": ["Imputer", "X", "y"],
+        "Target Imputer": ["Target Imputer", "X", "y"],
+        "OneHot_RandomForest": ["One Hot Encoder", "Imputer.x", "Target Imputer.y"],
```
ParthivNaresh, line 403: Wanted a text fixture that provided a graph with a modification of `y`

evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,60 @@ def feature_importance(self):
         df = pd.DataFrame(importance, columns=["feature", "importance"])
         return df
 
+    def graph_json(self):
+        """Generates a JSON with nodes consisting of the component names and parameters, and edges detailing component relationships.
+
+        Returns:
+            dag_json (str): A serialized JSON representation of a DAG structure.
+        """
```
eccabay, line 434: The explanation/template in the description for this PR was really helpful. Could you add that to the docs somewhere, either here in the docstring or somewhere in the User Guide?

ParthivNaresh: I'll add it in both!

evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,58 @@ def feature_importance(self):
         df = pd.DataFrame(importance, columns=["feature", "importance"])
         return df
 
+    def graph_json(self):
```
angela97lin, line 429: General question: what do you think about having this method *not* attached to the PipelineBase class? This feels like it could be its standalone util method that takes in a pipeline as an argument. Thoughts? 👀

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5214,3 +5215,44 @@ def test_automl_chooses_engine(engine_choice, X_y_binary):
             automl = AutoMLSearch(
                 X_train=X, y_train=y, problem_type="binary", engine=engine_choice
             )
+
+
+def test_graph_automl(X_y_multi):
```
angela97lin, line 5220: I'm not sure we need to test this on the automl level as a unit test. It seems like if we can test it on the pipeline level, that should be sufficient coverage?

evalml/pipelines/pipeline_base.py
```diff
@@ -425,6 +426,58 @@ def feature_importance(self):
...
+            }
+
+        for (
+            component_name,
+            component_info,
+        ) in self.component_graph.component_dict.items():
+            for input_ in component_info[1:]:
+                from_comp = input_[:-2]
+                if input_.endswith(".x") or input_.endswith(".y"):
+                    dag_json["Edges"][f"from_{from_comp}_to_{component_name}"] = {
```
angela97lin, line 467: Curious, what do you think about (`from_{from_comp}_to_{component_name}`)? Do we need it? It seems like something that the users could construct from the to / from, and feels a little more cluttered to me 🤔

angela97lin (commented):
Just left some general design-y questions I was curious to hear your thoughts on!

freddyaboulton (commented):
@ParthivNaresh Great job on this! This is a tough feature that I think is under-specified in terms of requirements. There are lots of ways to represent a graph as json. I'm curious what your thoughts are on this alternate format/implementation:

```python
from evalml.pipelines.component_graph import ComponentGraph

cg = ComponentGraph({"One Hot Encoder Special": ["One Hot Encoder", "X", "y"],
                     "Imputer": ["Imputer", "One Hot Encoder Special.x", "y"],
                     "Standard Scaler": ["Standard Scaler", "Imputer.x", "y"],
                     "Linear Regressor": ["Linear Regressor", "Standard Scaler.x", "y"]})
nodes = list(cg.component_instances.keys())
x_edges = cg._get_edges(cg.component_dict, "features")
y_edges = cg._get_edges(cg.component_dict, "target")
for component_name, component_info in cg.component_dict.items():
    for parent in component_info[1:]:
        if parent == "X":
            x_edges.append(("X", component_name))
        elif parent == "y":
            y_edges.append(("y", component_name))
graph_as_json = {"nodes": nodes,
                 "x_edges": x_edges,
                 "y_edges": y_edges}
graph_as_json
```
Produces
```
{'nodes': ['One Hot Encoder Special',
  'Imputer',
  'Standard Scaler',
  'Linear Regressor'],
 'x_edges': [('One Hot Encoder Special', 'Imputer'),
  ('Imputer', 'Standard Scaler'),
  ('Standard Scaler', 'Linear Regressor'),
  ('X', 'One Hot Encoder Special')],
 'y_edges': [('y', 'One Hot Encoder Special'),
  ('y', 'Imputer'),
  ('y', 'Standard Scaler'),
  ('y', 'Linear Regressor')]}
```

What I like about this approach is that it leverages the same logic for constructing the graph that the `ComponentGraph.graph` method does (what's missing is the parameters/attributes of each component). If we use this format we could have one method for generating the json and then `ComponentGraph.graph` can call that method and only worry about making the right graphviz calls.

Apologies for bringing this up once you got the PR up.

evalml/pipelines/pipeline_base.py
chukarsten: Agreed.  I think @freddyaboulton 's proposal for an alternate structure is a little leaner.  I think we could do away with that key and just have the edges be either a list of tuples where the meaning of the data is positional and documented, e.g. [(from, to), (from, to)...] or, if you want to retain some of your explicit structure, a list of dict: [{"from": from, "to": to, "data":data}...]

chukarsten (commented):
Nice work!  That test_component_as_json test is mighty thorough.  I think we might want to move towards removing that intermediary key from the json, though, as it doesn't really add too much and might actually clutter up the json a little bit.  I think also that we owe you a much better specification for the JSON next time.  I'll add this to our sprint feedback.

evalml/pipelines/pipeline_base.py
ParthivNaresh: So my reasoning for putting this here was:
1. I was looking at this from the perspective of running `AutoML.search()` and then choosing the best pipeline and being able to make a quick function call to the `graph_json` to export it to a third party tool for visualization.
2. If the user has access to any pipeline, then calling this function would be just as easy of a step as passing it to a util function.
3. Just as `model_understanding` has functions broken up into calculations and visualization, even though `PipelineBase.graph` won't be using `graph_json`, having them together seemed appropriate.

I like your idea and I'm totally down for discussing it further. It just seems to me that since the function requires `component_dict`, it might be better to keep in in `PipelineBase`.

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: The purpose of this test is to make sure that the `automl_parameters` that were set in `AutoMLSearch` are being picked up properly by `graph_json` in `{"Nodes": {"Attributes": ...}}`

PR #2813, opened 2021-09-20
Add `verbose` flag to top level search method

### LINKED ISSUE #2791: Add `verbose` flag to `search()`
Currently there is no way to allow AutoMLSearch output to show using the top level `search()` method. This issue tracks adding the `verbose` flag to `search()` and passing it down.

jeremyliweishih:
Fixes #2791.


evalml/automl/automl_search.py
```diff
@@ -12,9 +12,10 @@
 import cloudpickle
 import numpy as np
 import pandas as pd
-from dask import distributed as dd
 from sklearn.model_selection import BaseCrossValidator
 
+from dask import distributed as dd
```
freddyaboulton, line 17: This is a typo right?

freddyaboulton (approved):
Thanks @jeremyliweishih ! I think this is good to merge but I'd like your input on these two points beforehand:

1. Should we add test coverage?
2. Why not also make the change for `search_iterative` ?

jeremyliweishih:
@freddyaboulton `search_iterative` takes in `**kwargs` so it doesn't need this explicit flag. I'll fix the typo but I think the coverage in `test_automl.py` is enough for the verbose flag.

angela97lin (approved):
LGTM!

PR #2818, opened 2021-09-20
Add information about the row indices that are outliers in `OutliersDataCheck`

### LINKED ISSUE #2817: Add row information for `OutliersDataCheck`
Right now, `OutliersDataCheck` only reports which columns have outliers. It would also be useful if we could report, for each column, which rows are where the outliers are. This information could then be used to inform which rows to drop for data check actions.

angela97lin:
Closes #2817


freddyaboulton (approved):

PR #2819, opened 2021-09-20
Remove Sklearn Ensembler

### LINKED ISSUE #2620: Deprecate Scikit-Learn based Ensembling Component

christopherbunn:
Resolves #2620


evalml/tests/model_understanding_tests/prediction_explanations_tests/test_explainers.py
```diff
@@ -1647,22 +1651,19 @@ def test_categories_aggregated_when_some_are_dropped(
...
-    dummy_stacked_ensemble_multiclass_estimator,
-    dummy_stacked_ensemble_regressor_estimator,
     X_y_binary,
     X_y_multi,
     X_y_regression,
 ):
     if is_binary(problem_type):
         X, y = X_y_binary
-        pipeline = dummy_stacked_ensemble_binary_estimator
+        pipeline = StackedEnsembleClassifier(random_seed=0)
```
angela97lin, line 1660: Also pre-existing but i'm confused: it's called pipeline but is an estimator? Does this code work if its a pipeline with an ensemble estimator? Is that what we're trying to test right now?

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -449,36 +449,19 @@ def __init__(self, parameters, random_seed=0):
...
-            random_seed=random_seed,
-        )
-
-
 pipelines_that_do_not_support_fast_permutation_importance = [
     PipelineWithDimReduction,
     PipelineWithDFS,
     PipelineWithCustomComponent,
     EnsembleDag,
-    SklearnStackedEnsemblePipeline,
```
angela97lin, line 468: Our new impl also cannot be used for permutation importance right? `EnsembleDag` is that example?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -330,9 +318,7 @@ def test_stacked_estimator_in_pipeline(
 
     if problem_type == ProblemTypes.BINARY or problem_type == ProblemTypes.MULTICLASS:
         assert not np.isnan(pipeline.predict_proba(X)).values.any()
-        assert pipeline_score <= comparison_pipeline_score
-    else:
-        assert pipeline_score >= comparison_pipeline_score
```
angela97lin, line 335: why was this deleted?

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -128,9 +128,9 @@ def test_all_estimators(
         assert len((_all_estimators_used_in_search())) == 9
     else:
         if is_using_conda:
-            n_estimators = 15
+            n_estimators = 14
```
angela97lin, line 131: Why is only -1?

angela97lin (approved):
LGTM! Nothing blocking, but left some questions for clarifications! 😁 👏

evalml/tests/pipeline_tests/test_pipelines.py
christopherbunn: Hmm, it looks like we actually don't need to change this actually since both implementations were excluded as part of `_not_used_in_automl`. Thus, removing the sklearn versions actually doesn't have an impact.

PR #2820, opened 2021-09-21
Add "DROP_ROWS" action to output of `OutliersDataCheck`

### LINKED ISSUE #2677: Add "DROP_ROW" action to ``OutliersDataCheck``
We should add an action to drop outlier rows in ``OutliersDataCheck``

angela97lin:
Closes #2677

### DIFF
--- a/evalml/tests/data_checks_tests/test_outliers_data_check.py
+++ b/evalml/tests/data_checks_tests/test_outliers_data_check.py
@@ -41,6 +41,35 @@ def test_outliers_data_check_warnings():
     }
 
 
+def test_outliers_data_check_warnings_with_duplicate_outlier_indices():
+    a = np.arange(10) * 0.01
+    data = np.tile(a, (100, 10))
+
+    X = pd.DataFrame(data=data)
+    X.iloc[0, 3] = 1000
+    X.iloc[3, 25] = 1000
+    X.iloc[0, 55] = 10000
+    X.iloc[0, 72] = -1000
+    X.iloc[:, 90] = "string_values"
+
+    outliers_check = OutliersDataCheck()
+    assert outliers_check.validate(X) == {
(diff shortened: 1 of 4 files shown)


evalml/tests/data_checks_tests/test_outliers_data_check.py
```diff
@@ -37,7 +37,7 @@ def test_outliers_data_check_warnings():
             ).to_dict()
         ],
         "errors": [],
-        "actions": [],
+        "actions": [{"code": "DROP_ROWS", "metadata": {"indices": [0, 3, 5, 10]}}],
```
chukarsten, line 40: Can we add a second outlier in the same row, but different column?  I was going to write up an issue to increase coverage this way, but we could just do it here if you're game!  If not, let me know and I'll press Go on the issue.

Additionally, do we intend to handle that case with multiple outliers in the same row here?  Or keep the duplicate rows in DataCheckAction and let it handle that?

chukarsten (dismissed):
It's clear that you started with clear code and a solid design as this was super easy to complete!  Very nice.  I'm really excited about Datacheck/actions and how it's progressing.

I think the only thing I'd like to see resolved would be the issue with duplicate rows going into the Datacheck.  That might already be determined so I'm going to go back through the issue now, so if it is, ping me and I'll re-review.

evalml/tests/data_checks_tests/test_outliers_data_check.py
angela97lin: That's a great idea! I think we should filter for duplicates, updated impl to handle this case and added a new test :)

chukarsten (approved):
Love it.

bchen1116 (approved):
Nice! LGTM

PR #2821, opened 2021-09-21
Use label encoder component instead of doing encoding on the pipeline level

### LINKED ISSUE #2648: Do label encoding at the component level rather than pipeline level
Currently, our classification pipelines manually [encode](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/classification_pipeline.py#L52)/[decode](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/classification_pipeline.py#L104) the target during pipeline fitting/predicting. 

Why not do this with a component? It moves us towards our goal of encoding all the data transformations in the pipeline structure. This should also not be technically challenging because have a mechanism in place for doing inverse transformations(`inverse_transform`) so we can make sure the pipelines output unencoded targets. We have also made great strides in our component graph abstraction to better handle target transformations.

The one downside I see to this is that inexperienced users may not know to do label encoding automatically. It may make it tougher to specify classification pipelines quickly since we won't be able to use the list api. There are trade-offs, but I favor doing the encoding in a component.

Filing this to start the conversation!

angela97lin:
Closes #2648 

Perf test docs here: https://alteryx.atlassian.net/wiki/spaces/PS/pages/1088684090/Using+label+encoder+as+a+component+in+pipelines

### DIFF
--- a/evalml/pipelines/classification_pipeline.py
+++ b/evalml/pipelines/classification_pipeline.py
@@ -54,8 +54,7 @@ class ClassificationPipeline(PipelineBase):
         X = infer_feature_types(X)
         y = infer_feature_types(y)
         self._fit(X, y)
-        self._classes_ = np.unique(y)
-        self._classes_ = list(ww.init_series(self._classes_))
+        self._classes_ = list(ww.init_series(np.unique(y)))
         return self
 
     def _encode_targets(self, y):
--- a/evalml/pipelines/time_series_classification_pipelines.py
+++ b/evalml/pipelines/time_series_classification_pipelines.py
@@ -43,8 +43,7 @@ class TimeSeriesClassificationPipeline(TimeSeriesPipelineBase, ClassificationPip
         """
         X, y = self._convert_to_woodwork(X, y)
         self._fit(X, y)
-        self._classes_ = np.unique(y)
-        self._classes_ = list(ww.init_series(self._classes_))
+        self._classes_ = list(ww.init_series(np.unique(y)))
         return self
 
     def _estimator_predict_proba(self, features, y):


evalml/objectives/cost_benefit_matrix.py
```diff
@@ -46,7 +46,9 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
         Returns:
             float: Cost-benefit matrix score
         """
-        conf_matrix = confusion_matrix(y_true, y_predicted, normalize_method="all")
+        conf_matrix = evalml.model_understanding.graphs.confusion_matrix(
```
angela97lin, line 49: To avoid circular dependencies with model understanding tools and objectives :)

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -335,7 +335,8 @@ def test_fast_permutation_importance_matches_slow_output(
         objective = "R2"
 
     mock_supports_fast_importance.return_value = True
-    parameters["Random Forest Classifier"] = {"n_jobs": 1}
+    parameters["Estimator"] = {"n_jobs": 1}
```
angela97lin, line 338: This was incorrect before. We need to use the name of the component as it is found in dictionary keys, not the default name of the component.

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -49,16 +49,15 @@ def test_time_series_baseline(
     forecast_horizon, gap, problem_type, X_y_regression, X_y_binary, X_y_multi
 ):
 
-    if problem_type.TIME_SERIES_REGRESSION:
```
angela97lin, line 52: This was incorrect and always went to the else statement 😅

PR #2829, opened 2021-09-22
Add lint to fix jupyter executed cells

### LINKED ISSUE #2566: Local dev: update lint tools to check/remove notebook cell output
As a developer, If I want to add content to our notebook-based documentation, I typically modify the notebooks locally and then push changes up and open a PR. 

The problem is that we don't want to check-in notebooks with cell outputs into our repo because it makes it harder to analyze the diff during PR reviews and it makes the notebooks larger. 

This makes developing the notebooks a pain because I need to hit `Restart Kernel and Clear Outputs` on jupyter and then `Save` before I want to commit a change.

The proposal is for `make lint` to throw an error if any of the jupyter notebook cells in our docs have outputs. This will make it harder for developers to check in notebooks with output. The second part is for `make lint-fix` to automatically clear the outputs from the notebook cells.

For inspiration, look at the `notebook-version-standardizer`: https://github.com/alteryx/evalml/blob/main/docs/notebook_version_standardizer.py

bchen1116:
fix #2566 

Example usage:
Running 2 different notebooks in different directories:
diff --git adocssourcedemosfraud ipynb bdocssourcedemosfraud ipynb
diff git adocssourceuser_guidedata_actions ipynb bdocssourceuser_data_nsn

# lint output
make lint

# post lint-fix
Removed the outputs for

The notebooks then become cleaned and are no longer different from the originals

### DIFF
--- a/docs/notebook_version_standardizer.py
+++ b/docs/notebook_version_standardizer.py
@@ -58,7 +58,7 @@ def _fix_execution_and_output(notebook):
     with open(notebook, "r") as f:
         source = json.load(f)
         for cells in source["cells"]:
-            if cells["cell_type"] == "code" and cells["execution_count"] != 'null':
+            if cells["cell_type"] == "code" and cells["execution_count"] != None:
                 cells["execution_count"] = None
                 cells["outputs"] = []
         source["metadata"]["kernelspec"]["display_name"] = "Python 3"
--- a/docs/notebook_version_standardizer.py
+++ b/docs/notebook_version_standardizer.py
@@ -138,9 +138,9 @@ def check_execution():
     executed_notebooks = _get_notebooks_with_executions(notebooks)
     if executed_notebooks:
         executed_notebooks = ["\t" + notebook for notebook in executed_notebooks]
-        different_versions = "\n".join(executed_notebooks)
+        executed_notebooks = "\n".join(executed_notebooks)
         raise SystemExit(
-            f"The following notebooks have executed outputs:\n {different_versions}\n"
+            f"The following notebooks have executed outputs:\n {executed_notebooks}\n"
             "Please run make lint-fix to fix this."
         )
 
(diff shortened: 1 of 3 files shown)


docs/notebook_version_standardizer.py
```diff
@@ -81,13 +115,34 @@ def standardize(desired_version):
...
+        )
+
+
+@cli.command()
+def check_execution():
+    notebooks = _get_ipython_notebooks(DOCS_PATH)
+    executed_notebooks = _get_notebooks_with_executions(notebooks)
+    if executed_notebooks:
+        executed_notebooks = ["\t" + notebook for notebook in executed_notebooks]
+        different_versions = "\n".join(executed_notebooks)
```
jeremyliweishih, line 141: super nit pick but could we rename different versions here?

jeremyliweishih (approved):
beautiful

docs/notebook_version_standardizer.py
```diff
@@ -45,6 +45,40 @@ def _standardize_versions(notebooks, desired_version="3.8.6"):
...
+            if cells["cell_type"] == "code" and cells["execution_count"] != None:
+                return False
+    return True
+
+
+def _fix_execution_and_output(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and cells["execution_count"] != 'null':
```
freddyaboulton, line 61: If we parse the file into json, don't the 'null's get convered to `None`?

freddyaboulton (approved):
Thank you @bchen1116 ! This will be super useful.

docs/notebook_version_standardizer.py
bchen1116: I think you're right! A typo leftover

PR #2833, opened 2021-09-22
Remove TargetTransformer class

### LINKED ISSUE #2687: Delete TargetTransformer class
The TargetTransformer parent class was added when the pipeline/component graph `inverse_transform` method was added. We felt it was necessary to introduce this parent class because our component graph was not 100% ready to handle components that modified the target, e.g. we still supported `from_list` for the component graph even though lists are an ambiguous representation for pipelines with target transformers and it was possible to grab the `.y` from a component that did not modify the target.

You can see this thread from the original PR for context: https://github.com/alteryx/evalml/pull/2256#discussion_r630597217

The TargetTransformer class is no longer necessary because of:
https://github.com/alteryx/evalml/pull/2563
https://github.com/alteryx/evalml/pull/2556
https://github.com/alteryx/evalml/pull/2662

bchen1116:
fix #2687

### DIFF
--- a/evalml/tests/component_tests/test_components.py
+++ b/evalml/tests/component_tests/test_components.py
@@ -75,9 +75,6 @@ from evalml.pipelines.components.transformers.preprocessing.log_transformer impo
 from evalml.pipelines.components.transformers.samplers.base_sampler import (
     BaseSampler,
 )
-from evalml.pipelines.components.transformers.transformer import (
-    TargetTransformer,
-)
 from evalml.pipelines.components.utils import (
     _all_estimators,
     _all_transformers,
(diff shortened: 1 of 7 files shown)


evalml/tests/component_tests/test_components.py
```diff
@@ -1631,13 +1628,13 @@ def test_component_modifies_feature_or_target():
     for component_class in all_components():
         if (
             issubclass(component_class, BaseSampler)
-            or issubclass(component_class, TargetTransformer)
+            or hasattr(component_class, "inverse_transform")
```
angela97lin, line 1631: Open question: Do you think this will hold true long term? It's true for our components now but... 👀

angela97lin (approved):
Love these cleanup PRs, keep them coming 😂 ! LGTM, thank you! 👍

evalml/tests/component_tests/test_components.py
bchen1116: Hmm, I do think that if we have an `inverse_transform` method, then it'll modify the target rather than the features. It doesn't seem to make sense to have a transformer have an 'inverse_tranform` method for the features. I also don't think it'd make sense to introduce an `inverse_transform` method if we don't alter the target at all. Logically, it seems like this should still hold long term.

freddyaboulton: Should we add `inverse_transform` as a method to `Transformer`. The "default" behavior is a no-op and if we want to enforce that transformers that modify the target override that behavior we can do something like

```python
def inverse_transform(self, y):
    if self.modifies_target:
        raise NotImplementedError("Transformers that modify the target must override inverse_transform")
    return y
```

Or we don't make this change? Maybe I was being crazy when I filed the issue lol

chukarsten: I think Freddy's suggestion is a good one!

chukarsten (approved):
LGTM!

PR #2834, opened 2021-09-23
Our own Partial Dependence Implementation

### LINKED ISSUE #2475: Partial dependence errors with column with string and NaN values
Partial dependence errors when a column contains both string values and `NaN. There seems to be some failure in generating the grid as `np.unique()` cannot run on mixed values.

```python
import pandas as pd
from evalml.model_understanding import partial_dependence

df = pd.read_csv('1625078186889-mushroom_subset.csv')
y_train = df['class']
X_train = df.drop('class', axis=1)

aml = AutoMLSearch(X_train, y_train, 'binary')
aml.search()

pipeline = aml.best_pipeline

holdout = pd.read_csv('mushroom_holdout.csv')
partial_dependence(pipeline, holdout, 3)
```

Screen Shot 2021-07-08 at 11 24 07 AM

Notebook and datasets:
[string_nan_ex.zip](https://github.com/alteryx/evalml/files/6785630/string_nan_ex.zip)

### LINKED ISSUE #2502: Implement our own partial dependence method

freddyaboulton:
### Pull Request Description

Fixes #2502 
Fixes #2475

Same run-time as `main` for model understanding tests: 
[main](https://github.com/alteryx/evalml/runs/3689418420?check_suite_focus=true): 9m 59s
[this branch](https://github.com/alteryx/evalml/pull/2834/checks?check_run_id=3689441521): 9m 14s

Plots match between this branch and main
## This branch
image
## Main
image

## This branch
image
## Main
image

-----

### DIFF
--- a/evalml/model_understanding/_partial_dependence.py
+++ b/evalml/model_understanding/_partial_dependence.py
@@ -142,10 +142,10 @@ def _partial_dependence_calculation(pipeline, grid, features, X):
     else:
         prediction_method = pipeline.predict_proba
 
-    X_eval = X.ww.copy()
+    X_eval = X.copy()
     for _, new_values in grid.iterrows():
         for i, variable in enumerate(features):
-            X_eval.ww[variable] = pd.Series([new_values[i]] * X_eval.shape[0])
+            X_eval.loc[:, variable] = new_values[i]
 
         pred = prediction_method(X_eval)
 
--- a/evalml/model_understanding/_partial_dependence.py
+++ b/evalml/model_understanding/_partial_dependence.py
@@ -1,0 +1,185 @@
+import numpy as np
+import pandas as pd
+from scipy.stats.mstats import mquantiles
+
+from evalml.problem_types import is_regression
+
+
+def _grid_for_dates(X_dt, percentiles, grid_resolution):
+    timestamps = np.array(
+        [X_dt - pd.Timestamp("1970-01-01")] // np.timedelta64(1, "s")
+    ).reshape(-1, 1)
+    timestamps = pd.DataFrame(timestamps)
+    grid, values = _grid_from_X(
+        timestamps,
+        percentiles=percentiles,
+        grid_resolution=grid_resolution,
+        custom_range={},
+    )
+    grid_dates = pd.to_datetime(pd.Series(grid.squeeze()), unit="s")
+    return grid_dates
+
+
+def _grid_from_X(X, percentiles, grid_resolution, custom_range):
+    values = []
+    for feature in X.columns:
+        if feature in custom_range:
+            # Use values in the custom range
+            feature_range = custom_range[feature]
+            if not isinstance(feature_range, (np.ndarray, pd.Series)):
+                feature_range = np.array(feature_range)
+            if feature_range.ndim != 1:
+                raise ValueError(
+                    "Grid for feature {} is not a one-dimensional array. Got {}"
+                    " dimensions".format(feature, feature_range.ndim)
+                )
+            axis = feature_range
+        else:
+            uniques = np.unique(X.loc[:, feature])
+            if uniques.shape[0] < grid_resolution:
+                # feature has low resolution use unique vals
+                axis = uniques
+            else:
+                # create axis based on percentiles and grid resolution
+                emp_percentiles = mquantiles(
+                    X.loc[:, feature], prob=percentiles, axis=0
+                )
+                if np.allclose(emp_percentiles[0], emp_percentiles[1]):
+                    raise ValueError(
+                        "percentiles are too close to each other, "
+                        "unable to build the grid. Please choose percentiles "
+                        "that are further apart."
+                    )
+                axis = np.linspace(
+                    emp_percentiles[0],
+                    emp_percentiles[1],
+                    num=grid_resolution,
+                    endpoint=True,
+                )
+        values.append(axis)
+
+    return _cartesian(values), values
+
+
+def _cartesian(arrays):
+
+    arrays = [np.asarray(x) for x in arrays]
+    shape = (len(x) for x in arrays)
+
+    ix = np.indices(shape)
+    ix = ix.reshape(len(arrays), -1).T
+
+    out = pd.DataFrame()
+
+    for n, arr in enumerate(arrays):
+        out[n] = arrays[n][ix[:, n]]
+
+    return out
+
+
+def _partial_dependence_calculation(pipeline, grid, features, X):
+
+    predictions = []
+    averaged_predictions = []
+
+    if is_regression(pipeline.problem_type):
+        prediction_method = pipeline.predict
+    else:
+        prediction_method = pipeline.predict_proba
+
+    for _, new_values in grid.iterrows():
+        X_eval = X.copy()
+        for i, variable in enumerate(features):
+            X_eval.loc[:, variable] = new_values[i]
+
+        pred = prediction_method(X_eval)
+
+        predictions.append(pred)
+        # average over samples
+        averaged_predictions.append(np.mean(pred, axis=0))
+
+    n_samples = X.shape[0]
+
+    # reshape to (n_targets, n_instances, n_points) where n_targets is:
+    # - 1 for non-multioutput regression and binary classification (shape is
+    #   already correct in those cases)
+    # - n_tasks for multi-output regression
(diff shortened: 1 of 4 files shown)


evalml/model_understanding/_partial_dependence.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    values = []
+    for feature in X.columns:
+        if feature in custom_range:
+            # Use values in the custom range
+            feature_range = custom_range[feature]
+            if not isinstance(feature_range, (np.ndarray, pd.Series)):
+                feature_range = np.array(feature_range)
+            if feature_range.ndim != 1:
+                raise ValueError(
```
freddyaboulton, line 67: I'm ok if this isn't covered. It's impossible to trigger this as a user because `custom_range` is not a public parameter. But I'd like to keep this check in case we refactor this in the future. Helped catch a couple of bugs during development.

evalml/model_understanding/graphs.py
```diff
@@ -653,6 +652,11 @@ def partial_dependence(
             is_datetime = [_is_feature_of_type(features, X, ww.logical_types.Datetime)]
 
         if isinstance(features, (list, tuple)):
+            if any(is_datetime) and len(features) > 1:
```
freddyaboulton, line 655: There should to be two `if isinstance(features, (list, tuple))` checks. Consolidating into one now.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1519,3 +1519,30 @@ def test_partial_dependence_not_allowed_types(types, cols, expected_cols):
...
+
+    assert dep.shape[0] == X["provider"].dropna().nunique()
+    assert not dep["feature_values"].isna().any()
+    assert not dep["partial_dependence"].isna().any()
+
+    dep2way = partial_dependence(
+        pl, X, features=("amount", "provider"), grid_resolution=5
+    )
+    assert not dep2way.isna().any().any()
+    # Minus 1 in the columns because there is `class_label`
```
angela97lin, line 1547: +1, not minus?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1519,3 +1519,30 @@ def test_partial_dependence_not_allowed_types(types, cols, expected_cols):
...
+    assert dep2way.shape == (5, X["provider"].dropna().nunique() + 1)
```
angela97lin, line 1548: Omega nitpick, but I think it'd be a good idea to set grid_resolution as a variable and use it above / here, where``assert dep2way.shape == (grid_resolution_variable,...)``? Just so its more clear where this 5 value is coming from :)

evalml/model_understanding/_partial_dependence.py
```diff
@@ -0,0 +1,251 @@
...
+
+    ix = np.indices(shape)
+    ix = ix.reshape(len(arrays), -1).T
+
+    out = pd.DataFrame()
+
+    for n, arr in enumerate(arrays):
+        out[n] = arrays[n][ix[:, n]]
+
+    return out
```
angela97lin, line 121: This seems to be the same as https://github.com/scikit-learn/scikit-learn/blob/844b4be24d20fc42cc13b957374c718956a0db39/sklearn/utils/extmath.py#L655 except we return a dataframe, and since it's a public sklearn method, we could just import it--whatcha think? Also totally down to take their impl 😂

angela97lin (approved):
This is epic 👏!

I left some nitpicky comments but nothing blocking. Great work @freddyaboulton!

Also, the speedups are a cherry on top :)

chukarsten (approved):
wow @freddyaboulton , this is amazing.  I am really impressed.  I feel like you cleaned up the code substantially, improve performance and enhanced functionality.  This is a great PR.  I had a question about the handling of the times, but that isn't blocking.

evalml/model_understanding/_partial_dependence.py
chukarsten: This is a great idea.  I know as we first tredged through partial dependence that we borrowed a lot.  Perhaps a bit more from some private methods than I would like, but it was necessary.  If we can refactor to use their public methods, that's great.

freddyaboulton: @angela97lin Great point. Originally I wanted to use their method but the problem is that numpy arrays cannot handle mixed-types very well. So if we want to have a grid of categoricals and datetimes, the conversion storing it in a numpy array won't really work.

image

There may be a way around it I'm not seeing (maybe [this](https://numpy.org/doc/stable/user/basics.rec.html)) but IMO that's a nice to have?

evalml/tests/model_understanding_tests/test_partial_dependence.py
freddyaboulton: Thank you!!

evalml/model_understanding/_partial_dependence.py
```diff
@@ -0,0 +1,251 @@
...
+    predictions = []
+    averaged_predictions = []
+
+    if is_regression(pipeline.problem_type):
+        prediction_method = pipeline.predict
+    else:
+        prediction_method = pipeline.predict_proba
+
+    for _, new_values in grid.iterrows():
+        X_eval = X.copy()
```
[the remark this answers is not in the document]
freddyaboulton: Great point. I think we can move it out the loop. Will test it out!

evalml/model_understanding/_partial_dependence.py
```diff
@@ -0,0 +1,251 @@
...
+        X_dt (pd.DataFrame): Datetime features in original data. We currently
+            only support X_dt having a single column.
+        percentiles (tuple float): Percentiles to interpolate between.
+        grid_resolution (int): Number of points in range.
+
+    Returns:
+        pd.Series: Range of dates between percentiles.
+    """
+    timestamps = np.array(
+        [X_dt - pd.Timestamp("1970-01-01")] // np.timedelta64(1, "s")
```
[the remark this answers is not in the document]
freddyaboulton: Fantastic point @chukarsten ! I think what this is getting at is making our `custom_range` internal parameter public. I think there can be value in letting users specify how the grid for their features is computed! 

I will file a separate issue for tracking that.

evalml/tests/model_understanding_tests/test_partial_dependence.py
freddyaboulton: I completely agree!

freddyaboulton:
@chukarsten @angela97lin Thank you so much for the reviews! I didn't think this would get into the coming release.

Kicked off perf tests out of paranoia to make sure none of the datasets error out on partial dependence. Will merge if those look good.

freddyaboulton:
Perf tests [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1059915614/2502+Our+Own+Partial+Dependence+Implementation) and they look good to me!

PR #2837, opened 2021-09-23
Remove empty last cell in jupyter notebooks

bchen1116:
A quick change to also handle empty cells in jupyter notebooks in our docs.
Example usage can be seen [here](https://github.com/alteryx/woodwork/pull/1153#issuecomment-926084592) and
image


jeremyliweishih (approved):
nice!

PR #2846, opened 2021-09-27
Add Vowpal Wabbit estimators

### LINKED ISSUE #770: Add vowpal wabbit component
Vowpal wabbit is a package which supports [a variety of simple but useful algorithms](https://en.wikipedia.org/wiki/Vowpal_Wabbit). Its known for its quick training time. We should add a pipeline for it.

Links:
* https://github.com/VowpalWabbit/vowpal_wabbit
* https://vowpalwabbit.org/features.html

Note it has a non-pandas data input format. That should be the first thing we look at when adding this. I found a couple gists which convert pandas dicts to the VW input format: [one](https://gist.github.com/vlivashkin/b837c5c2ca5e55ab6ddba1d051d4ce25), [two](https://gist.github.com/maxpagels/6685efe304180e5a499540c85158b287).

We should be able to add regression, binary and multiclass classification.

angela97lin:
Reviving https://github.com/alteryx/evalml/issues/770 as part of inno days experimentation.

(Closes #770)

### DIFF
--- a/evalml/model_family/model_family.py
+++ b/evalml/model_family/model_family.py
@@ -44,6 +44,9 @@ class ModelFamily(Enum):
     PROPHET = "prophet"
     """Prophet model family."""
 
+    VOWPAL_WABBIT = "vowpal_wabbit"
+    """Vowpal Wabbit model family."""
+
     NONE = "none"
     """None"""
 
--- a/evalml/pipelines/__init__.py
+++ b/evalml/pipelines/__init__.py
@@ -37,6 +37,7 @@ from .components import (
     SVMRegressor,
     ARIMARegressor,
     ProphetRegressor,
+    VowpalWabbitClassifier,
 )
 
 from .component_graph import ComponentGraph
--- a/evalml/pipelines/components/__init__.py
+++ b/evalml/pipelines/components/__init__.py
@@ -26,6 +26,7 @@ from .estimators import (
     SVMClassifier,
     SVMRegressor,
     ARIMARegressor,
+    VowpalWabbitClassifier,
 )
 from .transformers import (
     Transformer,
--- a/evalml/pipelines/components/estimators/__init__.py
+++ b/evalml/pipelines/components/estimators/__init__.py
@@ -12,6 +12,7 @@ from .classifiers import (
     DecisionTreeClassifier,
     KNeighborsClassifier,
     SVMClassifier,
+    VowpalWabbitClassifier,
 )
 from .regressors import (
     LinearRegressor,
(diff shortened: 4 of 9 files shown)


eccabay (approved):
LGTM!

Just curious, where did the default parameters come from? I see they're not set in the source code. Not of much importance since we're not adding this to AutoML, but did you run any perf tests to see if those parameters were a good starting point?

evalml/pipelines/components/estimators/classifiers/vowpal_wabbit_classifiers.py
```diff
@@ -0,0 +1,118 @@
...
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+from evalml.utils.gen_utils import import_or_raise
+
+
+class VowpalWabbitBaseClassifier(Estimator):
+    """Vowpal Wabbit Base Classifier.
+
+    Args:
+        learning_rate (float): Boosting learning rate. Defaults to 0.5.
```
bchen1116, line 16: add `loss_function` to the args docstrings

evalml/tests/component_tests/test_components.py
```diff
@@ -1226,10 +1236,18 @@ def test_serialization(X_y_binary, ts_data, tmpdir, helper_functions):
...
+                        StackedEnsembleClassifier,
+                        StackedEnsembleRegressor,
+                        VowpalWabbitBinaryClassifier,
+                        VowpalWabbitMulticlassClassifier,
+                        VowpalWabbitRegressor,
+                    ),
+                )
             ):
                 assert (
                     component.feature_importance == loaded_component.feature_importance
```
bchen1116, line 1253: I know you test for default params in the vowpal respective test files, but why not add vowpal to `test_components.py::test_describe_component`?

bchen1116 (approved):
Left a question and a documentation addition, but looks good! The perf test doc was nice and easy to follow, agreed that we should hold off on adding to AutoMLSearch.

evalml/tests/component_tests/test_components.py
angela97lin: Wow, this test is *not* it. Will add, but also filed https://github.com/alteryx/evalml/issues/2876 😛

angela97lin:
Update: looks like the vowpal wabbit 8.8.1 version (aka the [latest conda version](https://anaconda.org/conda-forge/vowpalwabbit)) has a very different API from the latest (8.11.0) version which I based my code on. Filed https://github.com/VowpalWabbit/vowpal_wabbit/issues/3406, but otherwise may consider looking into using the older API or holding off entirely.

PR #2851, opened 2021-09-28
Check conda versions match pip versions

freddyaboulton:
### Pull Request Description

Add a CI job to check that the versions listed in our requirements files (used to install via pip) match the versions listed in our conda recipe (used to install via conda).

This is important because these should be equal to avoid the conda recipe getting stale. XGBoost is a good example of this. In pip, we list `>=1.4.2` but in conda, we have [`<1.3.0`](https://github.com/conda-forge/evalml-core-feedstock/blob/master/recipe/meta.yaml#L70). 

The goal is to check that our pip dependencies match the conda dependencies in `latest_release_changes` whenever we update one of our dependencies.

We check `latest_release_changes` instead of `master` because the `latest_release_changes` branch keeps track of all the changes we need to make prior to making the next conda release.

When we go do the next conda release in the feedstock, [CI will run](https://github.com/conda-forge/evalml-core-feedstock/pull/79) to check that the PR has the same dependencies as `latest_release_changes`

### Example where the check failed
https://github.com/alteryx/evalml/runs/3735325116?check_suite_focus=true

### Example where the check passed
https://github.com/alteryx/evalml/runs/3735846986?check_suite_focus=true

### Example of build_conda_pkg passing
https://github.com/alteryx/evalml/runs/3735847310?check_suite_focus=true

-----

### DIFF
--- a/.github/conda_version_check.py
+++ b/.github/conda_version_check.py
@@ -7,7 +7,7 @@ import os
 CONDA_RECIPE = "/Users/freddy.boulton/sources/personal/evalml-core-feedstock/recipe/meta.yaml"
 EVALML_PATH = "/Users/freddy.boulton/sources/evalml/"
 
-IGNORE_PACKAGES = {"python", "pmdarima"}
+IGNORE_PACKAGES = {"python", "pmdarima", "pyzmq"}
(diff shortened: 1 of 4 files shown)


chukarsten (approved):
This is really cool.  Thanks for doing this Freddy.  I guess you got tired of all of us asking you how to handle the feedstock X-D.

PR #2853, opened 2021-09-28
Add label encoder component to EvalML

### LINKED ISSUE #2852: Add label encoder component to EvalML
Breaking down #2648 to two different issues: 
- one which adds a new LabelEncoder component
- one which uses the label encoder in our pipelines in lieu of using a label encoder attribute

This issue simply tracks the former. :)

angela97lin:
Closes #2852

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/label_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/label_encoder.py
@@ -1,21 +1,16 @@
 """A transformer that encodes target labels using values between 0 and num_classes - 1."""
 import pandas as pd
-from woodwork.logical_types import Categorical
+from sklearn.preprocessing import LabelEncoder as SKLabelEncoder
 
 from ..transformer import Transformer
 
-from evalml.pipelines.components.transformers.encoders.onehot_encoder import (
-    OneHotEncoderMeta,
-)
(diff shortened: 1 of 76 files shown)


freddyaboulton (approved):

evalml/pipelines/components/transformers/encoders/label_encoder.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        if y is None:
+            raise ValueError("y cannot be None!")
+
+        y_ww = infer_feature_types(y)
+        y_t = self._component_obj.transform(y_ww)
+        y_t = pd.Series(y_t, index=y_ww.index)
+        return X, ww.init_series(y_t)
+
+    def fit_transform(self, X, y):
```
eccabay, line 75: Is there a reason we don't just use the parent `Transformer`'s implementation for this?

eccabay (approved):
Nice, LGTM!

evalml/pipelines/components/transformers/encoders/label_encoder.py
angela97lin: I think nope, except we want a different docstring 😅 (which was incorrect, updating!)

PR #2854, opened 2021-09-28
Js 2656 pipeline building

### LINKED ISSUE #2656: Move pipeline building logic into `IterativeAlgorithm`
Currently there is a discrepancy between how `EvalMLAlgorithm` and `IterativeAlgorithm`  builds pipelines. `IterativeAlgorithm` leans on `AutoMLSearch` to build pipelines using `make_pipeline()` and information such as `allowed_estimators`. However, `EvalMLAlgorithm` takes a more biased approach and creates its own pipelines in `EvalMLAlgorithm` itself. This issue tracks the following:

- move pipeline building logic out of `AutoMLSearch` and into `IterativeAlgorithm`
- move ensembling creation logic into `IterativeAlgorithm`
- move algorithm specific tests out of `test_automl.py`, `test_automl_search_classification.py` and `test_automl_search_regression.py` by mocking out `next_batch` and algorithm specific methods

jeremyliweishih:
Fixes #2656.

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -1,5 +1,7 @@
 """An automl algorithm which first fits a base round of pipelines with default parameters, then does a round of parameter tuning on each pipeline in order of performance."""
 import inspect
+import logging
+import warnings
 from operator import itemgetter
 
 import numpy as np
@@ -7,8 +9,16 @@ from skopt.space import Categorical, Integer, Real
 
 from .automl_algorithm import AutoMLAlgorithm, AutoMLAlgorithmException
 
+from evalml.automl.utils import get_pipelines_from_component_graphs
+from evalml.exceptions import ParameterNotUsedWarning
 from evalml.model_family import ModelFamily
-from evalml.pipelines.utils import _make_stacked_ensemble_pipeline
+from evalml.pipelines.components.utils import get_estimators
+from evalml.pipelines.utils import (
+    _make_stacked_ensemble_pipeline,
+    make_pipeline,
+)
+from evalml.problem_types import is_time_series
+from evalml.utils.logger import get_logger
 
 _ESTIMATOR_FAMILY_ORDER = [
     ModelFamily.LINEAR_MODEL,
@@ -42,7 +52,12 @@ class IterativeAlgorithm(AutoMLAlgorithm):
 
     def __init__(
         self,
-        allowed_pipelines=None,
+        X,
+        y,
+        problem_type,
+        sampler_name=None,
+        allowed_model_families=None,
+        allowed_component_graphs=None,
         max_iterations=None,
         tuner_class=None,
         random_seed=0,
@@ -54,6 +69,7 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         pipeline_params=None,
         custom_hyperparameters=None,
         _estimator_family_order=None,
+        verbose=False,
     ):
         """An automl algorithm which first fits a base round of pipelines with default parameters, then does a round of parameter tuning on each pipeline in order of performance.
 
@@ -71,13 +87,37 @@ class IterativeAlgorithm(AutoMLAlgorithm):
             custom_hyperparameters (dict or None): Custom hyperparameter ranges specified for pipelines to iterate over. Defaults to None.
             _estimator_family_order (list[ModelFamily]): specify the sort order for the first batch. Defaults to None, which uses _ESTIMATOR_FAMILY_ORDER.
         """
+        self.X = X
+        self.y = y
+        self.problem_type = problem_type
+        self.random_seed = random_seed
+        self.sampler_name = sampler_name
+        self.allowed_model_families = allowed_model_families
+        self.pipelines_per_batch = pipelines_per_batch
+        self.n_jobs = n_jobs
+        self.number_features = number_features
+        self._first_batch_results = []
+        self._best_pipeline_info = {}
+        self.ensembling = ensembling and len(self.allowed_pipelines) > 1
+        self._pipeline_params = pipeline_params or {}
+        self._custom_hyperparameters = custom_hyperparameters or {}
+
+        if verbose:
+            self.logger = get_logger(f"{__name__}.verbose")
+        else:
+            self.logger = logging.getLogger(__name__)
+
         self._estimator_family_order = (
             _estimator_family_order or _ESTIMATOR_FAMILY_ORDER
         )
         indices = []
         pipelines_to_sort = []
         pipelines_end = []
-        for pipeline in allowed_pipelines or []:
+
+        self.allowed_component_graphs = allowed_component_graphs
+        self._create_pipelines()
+
+        for pipeline in self.allowed_pipelines or []:
             if pipeline.model_family in self._estimator_family_order:
                 indices.append(
                     self._estimator_family_order.index(pipeline.model_family)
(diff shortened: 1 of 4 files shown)


evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -129,6 +193,88 @@ def __init__(
                         " and Real!"
                     )
 
+    def _create_pipelines(self):
```
jeremyliweishih, line 196: This logic is ripped out of `AutoMLSearch` and the API changes in `IterativeAlgorithm` accommodates this. Now `DefaultAlgorithm` and `IterativeAlgorithm` have more similar APIs.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -129,6 +196,141 @@ def __init__(
...
+                    self._pipeline_params,
+                    self.random_seed,
+                )
+            self._catch_warnings(w)
+
+        if self.allowed_pipelines == []:
+            raise ValueError("No allowed pipelines to search")
+
+        if self.ensembling and len(self.allowed_pipelines) == 1:
+            self.logger.warning(
```
jeremyliweishih, line 274: I also opted to move all the logging in as well. From my understanding of the logger, there will be no change in output.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -279,3 +481,27 @@ def _transform_parameters(self, pipeline, proposed_parameters):
                         component_parameters[param_name] = value
             parameters[name] = component_parameters
         return parameters
+
+    def _catch_warnings(self, warning_list):
```
jeremyliweishih, line 485: This was only used in pipeline building so I moved it in as well.

evalml/automl/automl_search.py
```diff
@@ -796,30 +661,32 @@ def __init__(
...
-        self.allowed_model_families = list(
-            set([p.model_family for p in (self.allowed_pipelines)])
-        )
 
-        self.logger.debug(
-            f"allowed_pipelines set to {[pipeline.name for pipeline in self.allowed_pipelines]}"
-        )
-        self.logger.debug(
-            f"allowed_model_families set to {self.allowed_model_families}"
+        text_in_ensembling = (
```
jeremyliweishih, line 665: Opted to leave this out as both `IterativeAlgorithm` and `DefaultAlgorithm` take in `text_in_ensembling` as arguments.

evalml/automl/automl_search.py
```diff
@@ -836,34 +703,17 @@ def __init__(
         else:
             raise ValueError("Please specify a valid automl algorithm.")
 
+        self.allowed_model_families = [
```
jeremyliweishih, line 706: Added this logic for continuity but `self.allowed_model_families` is only used in search output and `self.allowed_pipelines` is used `__str__`.

evalml/tests/conftest.py
```diff
@@ -193,6 +193,8 @@ def assert_allowed_pipelines_equal_helper():
     def assert_allowed_pipelines_equal_helper(
         actual_allowed_pipelines, expected_allowed_pipelines
     ):
+        actual_allowed_pipelines.sort(key=lambda p: p.name)
```
jeremyliweishih, line 196: Since pipelines used to be built in `AutoMLSearch`, tests would compare against an unsorted list of pipelines. Changed due to `self.allowed_pipelines` in `IterativeAlgorithm` sorting due to `_ESTIMATOR_FAMILY_ORDER`.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -53,30 +85,73 @@ def __init__(
...
         custom_hyperparameters=None,
         _estimator_family_order=None,
+        verbose=False,
     ):
         """An automl algorithm which first fits a base round of pipelines with default parameters, then does a round of parameter tuning on each pipeline in order of performance.
 
         Args:
-            allowed_pipelines (list(class)): A list of PipelineBase instances indicating the pipelines allowed in the search. The default of None indicates all pipelines for this problem type are allowed.
-            max_iterations (int): The maximum number of iterations to be evaluated.
+            X (pd.DataFrame): Training data.
```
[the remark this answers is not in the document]
jeremyliweishih: I'll remove it completely 😄

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -327,23 +433,34 @@ def test_iterative_algorithm_passes_n_jobs_catboost_xgboost(
...
+    X_y_binary, ensembling_value, dummy_binary_pipeline_classes
 ):
+    X, y = X_y_binary
+
+    (
+        dummy_binary_pipeline_classes,
+        allowed_component_graphs,
+    ) = dummy_binary_pipeline_classes()
+    dummy_binary_pipeline_classes = [dummy_binary_pipeline_classes[0]]
+    allowed_component_graphs = {"graph_1": allowed_component_graphs["graph_1"]}
```
angela97lin, line 445: Nitpick: I saw that you added `logistic_regression_binary_component_graph`, could be used here instead w/ keeping the original `logistic_regression_binary_pipeline_class`? Just noticed since you're taking the fixture and only taking the first element anyways.

evalml/tests/conftest.py
```diff
@@ -589,14 +591,21 @@ def clone(self):
 
 
 @pytest.fixture
-def logistic_regression_binary_pipeline_class():
+def logistic_regression_binary_component_graph():
+    return [
```
angela97lin, line 595: I think it's fine to create a new fixture, but if we're using it in conjunction with logistic_regression_binary_pipeline_class, we could either remove this and call logistic_regression_binary_pipeline_class.component_graph or have this be defined as `logistic_regression_binary_pipeline_class.component_graph`.

evalml/tests/conftest.py
angela97lin: If we want to do this, we have tests in place then to confirm that the order of allowed_pipelines as as expected? My concern here is that by sorting both lists, we're able to confirm that the types of pipelines match, but not the order in which they're executed anymore.

evalml/automl/automl_algorithm/iterative_algorithm.py
jeremyliweishih: I meant the `IterativeAlgorithm.__init__()` parameters!

evalml/automl/automl_search.py
jeremyliweishih: great call 😺

evalml/tests/conftest.py
jeremyliweishih: Yes I agree, but from what I could tell `assert_allowed_pipelines_equal_helper` isn't used in any tests that are checking for order and we have tests in `test_iterative_algorithm.py` like `test_iterative_algorithm_first_batch_order_param` that do account for the case!

jeremyliweishih: To clarify: since most of the tests in `test_automl` etc use the default iterative algorithm, the pipelines will be sorted in the order defined by `_ESTIMATOR_FAMILY_ORDER` in iterative algorithm. But since these tests directly compare against `make_pipeline`, the pipelines compared are in a different order. Another solution to this would be to sort the order into `_ESTIMATOR_FAMILY_ORDER` but I opted for a simpler solution.

bchen1116 (approved):
LGTM! Left two nits about docs

PR #2859, opened 2021-09-29
Show DeprecationWarning when Sklearn Ensemblers are called

christopherbunn:
When a sklearn ensembler is instantiated, show a warning that it will removed in the next release.

### DIFF
--- a/evalml/tests/component_tests/test_sklearn_stacked_ensemble_regressor.py
+++ b/evalml/tests/component_tests/test_sklearn_stacked_ensemble_regressor.py
@@ -197,7 +197,7 @@ def test_sklearn_stacked_feature_importance(
         clf.feature_importance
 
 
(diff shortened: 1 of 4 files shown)


evalml/tests/component_tests/test_sklearn_stacked_ensemble_regressor.py
```diff
@@ -194,3 +195,18 @@ def test_sklearn_stacked_feature_importance(
         NotImplementedError, match="feature_importance is not implemented"
     ):
         clf.feature_importance
+
+
+def test_component_graph_from_list_deprecation_warning(
```
angela97lin, line 200: Test name might be wrong? 😂

angela97lin (approved):
LGTM, thanks! Just left a comment about the test name 😅

PR #2875, opened 2021-10-05
Add max batches stopping criterion logic

### LINKED ISSUE #2800: Long mode of search() does not go past fast batches unless specified with max_time
@jeremyliweishih commented on [Thu Sep 16 2021](https://github.com/alteryx/evalml_looking_glass/issues/386)

Currently, if a user specifies `mode="long"` using the top level `search()` method, it will only run for the default `AutoMLSearch.search()` run time and does not even make it past the `fast` batches. This issue tracks fixing the logic in `search()` so that if a user specifies `long` it will complete at least the initial long mode batches.

A complication with this issue is the current logic behind `max_batches` and `max_iterations`. My initial understanding was that by specifying `max_batches` AutoMLSearch will end after exactly n batches. However, the actual stopping criterion is based off of a calculation of max_pipelines and the length of allowed pipelines. This works for the current algorithm and its ensembling implementation but does not work with the new default algorithm. This might be resolved by #2656.

jeremyliweishih:
Fixes #2800.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -179,7 +179,11 @@ def search(
 
     max_batches = None
     if mode == "fast":
-        max_batches = 3
+        max_batches = 4 # corresponds to end of 'fast' mode
+    elif mode == "long" and max_time:
+        max_batches = 999 # defers to stopping criterion
+    elif mode == "long" and max_time is None:
+        max_batches = 6 # corresponds to end of 'long' exploration phase
 
     automl_config = {
         "X_train": X_train,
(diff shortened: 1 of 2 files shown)


evalml/automl/automl_search.py
```diff
@@ -1148,12 +1158,14 @@ def _should_continue(self):
...
         num_pipelines = self._num_pipelines()
 
-        # check max_time and max_iterations
+        # check max_time, max_iterations and max_batches
         elapsed = time.time() - self._start
         if self.max_time and elapsed >= self.max_time:
             return False
         elif self.max_iterations and num_pipelines >= self.max_iterations:
             return False
+        elif self.max_batches and self._get_batch_number() > self.max_batches:
```
jeremyliweishih, line 1167: This logic would only be invoked for `DefaultAlgorithm` as `max_iterations` is set when it is None when using `IterativeAlgorith`.

PR #2883, opened 2021-10-07
Add integration tests for end to end flow for data checks --> data check actions

### LINKED ISSUE #1998: Data check actions: don't produce duplicate components
Currently, we could add multiple duplicate components from the same actions. We should try to prune duplicate components, if possible:
```
# current behavior. But this might cause errors, trying to drop 'some col' multiple times
actions = [DataCheckAction(DataCheckActionCode.DROP_COL, {"columns": ['some col']}),
                  DataCheckAction(DataCheckActionCode.DROP_COL, {"columns": ['some col']})]
_make_component_list_from_actions(actions) == [DropColumns(columns=['some col']),
                                                                    DropColumns(columns=['some col'])] 
```

```
# desired behavior
actions = [DataCheckAction(DataCheckActionCode.DROP_COL, {"columns": ['some col']}),
                  DataCheckAction(DataCheckActionCode.DROP_COL, {"columns": ['some col']})]
_make_component_list_from_actions(actions) == [DropColumns(columns=['some col'])] # only once
```

There's some support in the `DataChecks` class to not return duplicate actions, but if a user just used multiple DataCheck objects, we will still end up with duplicates in `_make_component_list_from_actions`

### LINKED ISSUE #2815: Add integration tests for end to end flow for data checks --> data check actions
There are currently unit tests for both data checks and data check actions, but there's no test that takes the output of a data check and passes it to the code for data check actions. This has caused some issues with API consistency, found and written about in #2792. It could be good to create an end-to-end test flow that runs data checks, and feeds the output of `validate` to creating data check actions components.

This issue tracks adding integration tests (subfolder of `tests`) and creating a separate job for integration test suites?

angela97lin:
Closes #2815, closes #1998.

Along the way, I noticed that our data checks return dictionary forms of the data check messages / actions (since those are more easily serializable), but `_make_component_list_from_actions` takes in actual data check action objects. Added a helper method to go from one to another.  I also noticed it's not super easy to use a list of components, so going to work on https://github.com/alteryx/evalml/pull/2968 sooner rather than later to make transformations easier :)

I think it's fine to just one python version for now and linux only, but curious if there are other opinions!


evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
```diff
@@ -0,0 +1,132 @@
...
+        }
+    )
+    X_expected = pd.DataFrame(
+        {
+            "lots_of_null": [None, 2, None, 3, 5],
+            "no_null": [1, 2, 3, 4, 5],
+        }
+    )
+    for component in action_components:
+        X_t = component.fit_transform(X_t)
```
chukarsten, line 68: This test highlights how nicely DataCheck/Actions fits into the fit/transform syntax.  Very nicely done.  I like this, very intuitive.

evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
```diff
@@ -0,0 +1,132 @@
...
+    ]
+    action_components = _make_component_list_from_actions(actions)
+    assert action_components == [DropRowsTransformer()]
+
+    X_t = pd.DataFrame(data=data)
+    X_t.iloc[0, 3] = 1000
+    X_t.iloc[3, 25] = 1000
+    X_t.iloc[5, 55] = 10000
+    X_t.iloc[10, 72] = -1000
+    X_t.iloc[:, 90] = "string_values"
```
chukarsten, line 122: Maybe I'm missing something, but can we not just use X?  X isn't changing is it?

chukarsten (approved):
Angela, this looks good to me!  This will lead to some difficult questions about what constitutes an integration tests and how we handle all of them, but I think this is a good starting point.

evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
angela97lin: Yeah... I think you're right--I had made a separate variable in case we were changing X somewhere / wanted to check the original value, but that was probably unnecessarily cautious :). Removing!

PR #2933, opened 2021-10-18
Update CI to run doctest (docstring tests) and apply necessary fixes to docstrings

angela97lin:
Options:
- Add separate job for doctests. Seems like overkill for now, given that we only have a handful? Could be good to revisit during https://github.com/alteryx/evalml/issues/2936
- Currently this PR just adds to the `git-test-others`. If we expand the suite, could be worth to add to each individual git-test command as we have for pytest. :)


angela97lin:
Great idea @freddyaboulton! Updated internal docs, and will add a brief line in our contributing guidelines to add doctests if applicable.

PR #2958, opened 2021-10-25
Add `NoSplit` data splitter

### LINKED ISSUE #2956: Remove requirement of data splitting in `search`
One of the first steps to supporting unsupervised learning. Clustering has no concept of cross-validation, since the point is not to predict on unseen data but rather gather insights about the training data. Right now, `AutoMLSearch` requires a cross-validation method, so we need to remove this requirement before adding any clustering pipelines. I see two ways of doing this: 
- Remove the requirement of a data splitter altogether - allow running pipelines that do not use cross validation or data splitting at all, the step is just skipped.
- Add a `NoSplit` data splitter - for pipelines that shouldn't use CV, use a placeholder data splitter that does nothing.

I am moving forward with the second option, since it's significantly less of a refactor of `search`.

eccabay:
Closes #2956 

First step of adding unsupervised learning to evalml, this code is lifted directly from my [innovation days branch](https://github.com/alteryx/evalml/compare/inno_days_clustering).


chukarsten (approved):
Thanks for making the changes!

PR #2979, opened 2021-10-27
Preserve input indices on estimators and pipelines' predict/predict_proba/transform/inverse_transform

### LINKED ISSUE #1639: Pipeline predict should preserve the data index
Currently, calling `pipeline.predict(X)` will return a series with an index with values from 0 to `X.shape[0] - 1`. This has can cause bugs  (#1634) and can be confusing to users if they pass in data with a custom index.

Should we change this behavior?

angela97lin:
Closes #1639


evalml/model_understanding/prediction_explanations/_user_interface.py
```diff
@@ -741,7 +741,9 @@ def make_dict(self, index, y_pred, y_true, scores, dataframe_index):
 
         return {
             "probabilities": pred_values,
-            "predicted_value": _make_json_serializable(self.predicted_values[index]),
+            "predicted_value": _make_json_serializable(
```
angela97lin, line 744: Calling iloc will get the positional index. Since `.predict()` now preserves the original index, which may be non-numeric, we need to call `iloc` here!

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -146,9 +146,12 @@ def predict(self, X, y=None):
...
         )
 
-        y_pred = self._component_obj.predict(prophet_df)["yhat"]
-        y_pred = y_pred.rename(None)
-        return y_pred
+        prophet_output = self._component_obj.predict(prophet_df)
+        predictions = prophet_output["yhat"]
+        predictions = infer_feature_types(predictions)
+        predictions = predictions.rename(None)
+        predictions.index = prophet_output["ds"]
```
angela97lin, line 153: Reading the docs [here](https://facebook.github.io/prophet/docs/quick_start.html#python-api), it looks like the date index information is stored in "ds".

evalml/tests/component_tests/test_components.py
```diff
@@ -1562,6 +1573,12 @@ def test_transformer_fit_and_transform_respect_custom_indices(
         pd.testing.assert_index_equal(
             y.index, y_original_index, check_names=check_names
         )
+
+    if hasattr(transformer_class, "inverse_transform"):
```
freddyaboulton, line 1577: Nice thanks for adding this!

evalml/tests/component_tests/test_estimators.py
```diff
@@ -297,3 +299,52 @@ class MockEstimator(Estimator):
...
     assert isinstance(y, pd.Series)
+
+
+@pytest.mark.parametrize("estimator_class", _all_estimators())
+@pytest.mark.parametrize("use_custom_index", [True, False])
+@pytest.mark.parametrize(
+    "problem_type",
+    [ProblemTypes.BINARY, ProblemTypes.MULTICLASS, ProblemTypes.REGRESSION],
+)
+def test_estimator_fit_predict_and_predict_proba_respect_custom_indices(
```
freddyaboulton, line 310: Why doesn't this test catch this?

```python
import pandas as pd
from evalml.pipelines.components import LightGBMClassifier

X, y = load_breast_cancer()
lgbm = LightGBMClassifier()
X.index = range(25, 25 + X.shape[0]
lgbm.fit(X, y)

pd.testing.assert_index_equal(lgbm.predict(X).index, X.index)
```

AssertionError: Index are different

```
Index values are different (100.0 %)
[left]:  RangeIndex(start=0, stop=569, step=1)
[right]: RangeIndex(start=25, stop=594, step=1)
```

`predict_proba` does preserve the index though

freddyaboulton (changes requested):
@angela97lin Thank you this looks great! I think there's a bug in the test cause LightGBM.predict is not preserving the index?

evalml/tests/component_tests/test_estimators.py
angela97lin: Wow, this is a really good catch! Tracking this down, it's because we have separate logic for LightGBM and XGBoost's predict methods depending on whether or not we need to encode the targets (they internally have label encoders). Since my test uses the X_y_binary fixture, it doesn't use the label encoder. In the case where we *do* use the label encoder, the label encoder's inverse_transform wipes the index so we have to set it again:

```
        X, _ = super()._manage_woodwork(X)
        X.ww.set_types(self._convert_bool_to_int(X))
        X = _rename_column_names_to_numeric(X, flatten_tuples=False)
        predictions = super().predict(X)
        if not self._label_encoder:
            return predictions
        predictions = pd.Series(
            self._label_encoder.inverse_transform(predictions.astype(np.int64)),
            index=predictions.index,
        )
        return infer_feature_types(predictions)
```


Adding another set of tests for numeric vs non-numeric targets 😅

evalml/pipelines/components/transformers/encoders/label_encoder.py
```diff
@@ -97,5 +97,7 @@ def inverse_transform(self, y):
         """
         if y is None:
             raise ValueError("y cannot be None!")
+        y_ww = infer_feature_types(y)
```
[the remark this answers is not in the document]
angela97lin: Yeah, not really related to index work, just standardizing since we do so for other components :')

evalml/pipelines/classification_pipeline.py
```diff
@@ -101,9 +101,11 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
         Returns:
             pd.Series: Estimated labels.
         """
-        predictions = self._predict(X, objective=objective)
-        predictions = self.inverse_transform(predictions.astype(int))
-        predictions = pd.Series(predictions, name=self.input_target_name)
+        _predictions = self._predict(X, objective=objective)
+        predictions = self.inverse_transform(_predictions.astype(int))
+        predictions = pd.Series(
+            predictions, name=self.input_target_name, index=_predictions.index
```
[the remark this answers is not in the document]
angela97lin: Theoretically if all of the components properly handle preserving indices including the label encoder then no... but I figured it would be good to have on the pipelines just in case!

bchen1116 (approved):
LGTM! Glad we're now able to preserve the index in our estimators and pipelines

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
```diff
@@ -112,7 +112,7 @@ def inverse_transform(self, y):
         """
         if y is None:
             raise ValueError("y cannot be None for PolynomialDetrender!")
-        y_dt = infer_feature_types(y)
-        y_t = self._component_obj.inverse_transform(y_dt)
-        y_t = infer_feature_types(pd.Series(y_t, index=y_dt.index))
+        y_ww = infer_feature_types(y)
```
chukarsten, line 115: Is part of the underlying problem here that `infer_feature_types()` should be preserving the index?  Should we do it in there?

evalml/tests/component_tests/test_estimators.py
```diff
@@ -297,3 +299,60 @@ class MockEstimator(Estimator):
...
+        X, y = X_y_regression
+
+    X = pd.DataFrame(X)
+    y = pd.Series(y)
+
+    if is_classification(problem_type) and not use_numerical_targets_for_classification:
+        y = y.map({val: str(val) for val in np.unique(y)})
+
+    if use_custom_index:
+        gen = np.random.default_rng(seed=0)
```
chukarsten, line 339: I know we do this in similar tests, but it's bold.  I'm always hesitant to see pseudo-randomness added into a testing suite.  I think it opens the door for potential flakiness and I'm not sure about the implications across platforms.  I'm pretty sure this usage is fine, though.

chukarsten (approved):
Angela, this looks great and I think it greatly enhances our test coverage with respect to index preservation.  I don't see anything that blocks, but I am curious whether we should do any index preservation within infer_feature_types?  I know that would cut out a few lines, from the individual pipeline classes, and it also seems like something we should expect of the function.  If not, also, nbd.

evalml/pipelines/components/transformers/preprocessing/polynomial_detrender.py
angela97lin: @chukarsten Unfortunately, I think regardless of our `infer_feature_types` behavior (which I think should preserve indices but I'm not positive), the issue is that a lot of third party libraries might return data such as np.arrays or new pandas objects that don't have the indices attached to them. 🥲

evalml/tests/component_tests/test_estimators.py
angela97lin: Heh, I had taken inspiration from our other tests but agreed! The pseudo-randomness isn't really necessary here so I'll remove it from this instance and our other tests that use this :)

Issue #2906, opened 2021-10-14
Update label encoder in pipeline implementation to look for LabelEncoder class instead of looking for "Label Encoder"


angela97lin:
https://github.com/alteryx/evalml/pull/2821/ introduces the label encoder as a component in the pipeline, rather than an attribute. However, we still set a pipeline's `_encoder` attribute to a label encoder in the pipeline, if it's available.

As documented in [this comment](https://github.com/alteryx/evalml/pull/2821/files#r723692569), there is an assumption that the Label Encoder component has a name "Label Encoder". However, we should be more flexible and accept the label encoder even if its name was "my cool label encoder". 

One suggestion is to find the component by checking for the component class instead. This work would be made much easier by https://github.com/alteryx/evalml/issues/2878.

One edge case that may be important to think about the case where there may be multiple label encoders. What do we do then?

chukarsten:
I added the needs design tag to this to cover the discussion of the edge case.  I think when we agree on a way forward for the multiple label encoders case, we can begin work.

PR #2986, opened 2021-10-28
Handle categorical columns in a separate chain for `DefaultAlgorithm` pipelines

jeremyliweishih:
Couple major changes:

1. When applicable (i.e categorical columns exist), the pipeline structure changes s.t categoricals are handled in a separate preprocessing chain.  Heres, an example of what the change looks like.  Note: the real pipeline has an additional label encoder which is detailed by #2987.

graphviz (1)

 
This requires some additional logic in `DefaultAlgorithm` due to renaming of components and handling when to create this two-pronged pipeline.

2. I made the same changes I made to `Select Columns Transformer` to `Drop Columns Transformer` as well. This was because if certain columns were to be dropped, sometimes it would only exist in one chain or different columns would be in different chains. Thus, we only drop whats available.

3. Added `_make_pipeline_from_multiple_graphs` which essentially has the same implementation of the stacked ensembler helper but with a couple logic changes on naming and selecting which final y output to take for the estimator. I left this as a separate function due to the discussions in #2987 and #2906 and those limitations apply here as well.

Perf tests:
[report.html.zip](https://github.com/alteryx/evalml/files/7463113/report.html.zip)


evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -81,8 +81,14 @@ class DropColumns(ColumnSelector):
     """{}"""
     needs_fitting = False
 
+    def _check_input_for_columns(self, X):
```
jeremyliweishih, line 84: Needed the same changes that I made to `SelectColumns` in #2944 due to the split in pipeline structure.

evalml/pipelines/utils.py
```diff
@@ -369,6 +386,126 @@ def _make_new_component_name(model_type, component_name, idx=None):
     )
 
 
+def _make_pipeline_from_multiple_graphs(
```
jeremyliweishih, line 389: Adding this new function but most of it is the same as `_make_stacked_ensemble_pipeline` with some new logic around which final y input to take. I can put up an issue to refactor this moving forward but mainly kept this separate due to the conversations surrounding #2987.

evalml/pipelines/utils.py
```diff
@@ -158,7 +158,7 @@ def _get_preprocessing_components(
             )
             pp_components.append(Undersampler)
 
-    if estimator_class.model_family == ModelFamily.LINEAR_MODEL:
+    if estimator_class and estimator_class.model_family == ModelFamily.LINEAR_MODEL:
```
jeremyliweishih, line 161: Due to `make_pipeline` changes below.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -316,6 +351,20 @@ def add_result(self, score_to_minimize, pipeline, trained_pipeline_results):
                     "RF Classifier Select From Model"
                 ).get_names()
 
+            try:
+                ohe = pipeline.get_component("One Hot Encoder")
```
freddyaboulton, line 355: Rather than the try except can we just check if the One Hot Encoder component exists in the pipeline?

I think since this lives in the default algorithm, and the default algorithm has full control over the pipelines it creates (unlike iterative algo), I think it's ok since the One Hot Encoder will always have the same name?

freddyaboulton (commented):
@jeremyliweishih Thank you for this! 🙏  I left some comments, will come back for a full review later!

evalml/pipelines/utils.py
jeremyliweishih: @freddyaboulton the implementation for the stacked ensembler assumes that each input pipeline has an estimator so that the final y input chosen to the stacked estimator is the y input going into each pipelines estimator. This usually means that the label encoder or the original y is selected as the final y input. This implementation always chooses the last component that modifies y of the last pipeline. @christopherbunn could give a better explanation of how the stacked implementation works.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -369,3 +418,51 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
+                self.y,
+                estimator,
+                self.problem_type,
+                sampler_name=self.sampler_name,
+                parameters=categorical_pipeline_parameters,
+                extra_components=[SelectColumns],
+                extra_components_position="before_preprocessing",
+                use_estimator=False,
+            )
+            input_pipelines = [basic_pipeline, categorical_pipeline]
```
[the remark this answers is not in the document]
jeremyliweishih: each sampler _should_ have the same parameters so theoretically it should be ok as it'll have the same outputs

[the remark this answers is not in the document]
freddyaboulton: My question is whether sampling parts of the data independently is materially different than sampling the whole data once. For Undersampling, I think the answer is no but for oversampling I'm not sure. If the perf tests come back ok, then we can leave this question for a follow-up issue.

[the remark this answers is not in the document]
bchen1116: Agreed here. Oversampling is handled differently for Categorical data versus Numeric data, so this could crunch out different results if done separately (SMOTE/SMOTEN vs SMOTENC) versus together. However, I do think the sampling strategy should be the same overall? In that, if the data was numeric + categorical, I think sampling with `SMOTE` and `SMOTEN` on the data separately uses the same techniques as sampling with `SMOTENC`, which should lead to very similar/same sampling results. 

Definitely something worth testing, however!

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -369,3 +418,51 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
+            use_estimator=False if self._selected_cat_cols else True,
+        )
+
+        if self._selected_cat_cols:
+            self._split = True
+            categorical_pipeline_parameters = {
+                "Select Columns Transformer": {"columns": self._selected_cat_cols}
+            }
+            categorical_pipeline = make_pipeline(
+                self.X,
```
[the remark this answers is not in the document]
jeremyliweishih: will probably file an followup issue for this @freddyaboulton.

evalml/pipelines/utils.py
```diff
@@ -369,6 +387,119 @@ def _make_new_component_name(model_type, component_name, idx=None):
     )
 
 
+def _make_pipeline_from_multiple_graphs(
+    input_pipelines,
+    estimator,
+    problem_type,
+    parameters={},
```
freddyaboulton, line 394: I think mutable defaults are dangerous, let's use None for the default?

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -179,65 +184,105 @@ def _create_naive_pipelines(self, use_features=False):
...
             for estimator in estimators
         ]
 
-        pipelines = self._create_pipelines_with_params(
-            pipelines, {"Select Columns Transformer": {"columns": self._selected_cols}}
-        )
+        if self._split:
+            # rename all pipeline params with correct component names
+            names_to_value = {}
+            for component_name in self._pipeline_params:
```
freddyaboulton, line 241: We should probably be doing the same for `_custom_hyperparameters` right?

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -179,65 +184,105 @@ def _create_naive_pipelines(self, use_features=False):
...
-                extra_components=[SelectColumns],
             )
             for estimator in estimators
         ]
 
-        pipelines = self._create_pipelines_with_params(
-            pipelines, {"Select Columns Transformer": {"columns": self._selected_cols}}
-        )
+        if self._split:
+            # rename all pipeline params with correct component names
```
freddyaboulton, line 239: I think this is worth it's own method, especially if we'll eventually have to do the same for hyperparams.

freddyaboulton (approved):
@jeremyliweishih Looks good to me! Thank you for making the changes. I think this unblocks further development by getting rid of the pesky bugs from categorical features but I think we can improve the implementation regarding the two samplers and splitting numeric vs categorical further so I'm looking forward to the follow-up issues!

Also, I'm wondering if this code will be substantially simpler if we had a first-class api for combining pipelines. Maybe we need to prioritize that? Seems like we'll have to do it for data check actions? FYI @angela97lin @dsherry @chukarsten

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -179,65 +184,105 @@ def _create_naive_pipelines(self, use_features=False):
...
-                parameters=parameters,
-                extra_components=[SelectColumns],
             )
             for estimator in estimators
         ]
 
-        pipelines = self._create_pipelines_with_params(
-            pipelines, {"Select Columns Transformer": {"columns": self._selected_cols}}
-        )
+        if self._split:
```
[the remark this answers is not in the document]
jeremyliweishih: Yes, this should be updated! Thanks for catching this!

evalml/pipelines/utils.py
jeremyliweishih: Good call!

evalml/automl/automl_algorithm/default_algorithm.py
jeremyliweishih: ah you're correct!

evalml/pipelines/utils.py
```diff
@@ -369,6 +387,120 @@ def _make_new_component_name(model_type, component_name, idx=None):
...
+                    if is_classification(problem_type):
+                        new_component_list.append("Label Encoder.y")
+                    else:
+                        new_component_list.append("y")
+                else:
+                    new_component_list.append(item)
+                if i != 0 and item.endswith(".y"):
+                    final_y = _make_new_component_name(
+                        component_pipeline_name, item, name_idx, sub_pipeline_name
+                    )
```
bchen1116, line 479: is it possible to move this check

```python
if i != 0 and item.endswith(".y"):
...
```
to fall under
```python
elif isinstance(item, str) and item not in ["X", "y"]:
    comp = _make_new_component_name(
                      component_pipeline_name, item, name_idx, sub_pipeline_name
                  )
    new_component_list.append(comp)
    if i != 0 and item.endswith(".y"):
        final_y = comp
elif ...
```
 This just reduces extra calls to `_make_new_component_name`, which could save time and memory for larger runs

bchen1116: Another optimization we could do is enumerate over the `input_pipelines`, and only when we are on the last pipeline, check for and set `final_y`. This way, if we have a larger number of pipelines in `input_pipelines`, we won't need to keep setting `final_y` and overwriting it. 
Not a needed change, but would be nice to be more efficient with memory usage if we can!

bchen1116 (changes requested):
Haven't gone through the tests yet, but this is looking good so far! Left many nits and some comments on things that I think could be changed to make this more robust and efficient. 

Hopefully can finish looking through the tests tomorrow.

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -264,3 +308,76 @@ def test_default_algo_drop_columns(mock_get_names, columns, X_y_binary):
...
+        "Embarked": ["Embarked_S"],
+    }
+
+    algo = DefaultAlgorithm(X, y, ProblemTypes.BINARY, None)
+
+    batch = algo.next_batch()
+    add_result(algo, batch)
+
+    batch = algo.next_batch()
+    add_result(algo, batch)
```
bchen1116, line 369: do we need to do 2 calls for this?

bchen1116 (approved):
Thanks for making the changes! Looks good to me, left a question for my own understanding. 

I remember there was some messy stuff with how the pipeline names are displayed in these longer split pipelines. Was there an issue filed to track fixing that for a future PR?

jeremyliweishih:
@bchen1116 will file that issue, along side the sampler issue and the cat-num issue!

evalml/tests/automl_tests/test_default_algorithm.py
jeremyliweishih: yes, its to get it to the correct batch!

jeremyliweishih:
waiting on latest perf tests before merging.

evalml/tests/automl_tests/test_default_algorithm.py
bchen1116: Can you add a comment on that for future reference? I'm sure i'll forget lol

Issue #2936, opened 2021-10-18
Add more doctests to class modules and methods


angela97lin:
While addressing https://github.com/alteryx/evalml/issues/2935, I noticed that we don't have many docstring tests. Most of our tests are in the data checks module. I think it could be good to add more, since they also serve as helpful examples on how to use a certain method!

(This might be better converted into an epic with each module as its own issue if we want to add extensive docstring coverage.)

Current docstring tests:
```
evalml/data_checks/class_imbalance_data_check.py .                                                                                   [  5%]
evalml/data_checks/datetime_format_data_check.py F                                                                                   [ 11%]
evalml/data_checks/datetime_nan_data_check.py .                                                                                      [ 16%]
evalml/data_checks/highly_null_data_check.py .                                                                                       [ 22%]
evalml/data_checks/id_columns_data_check.py .                                                                                        [ 27%]
evalml/data_checks/invalid_targets_data_check.py F                                                                                   [ 33%]
evalml/data_checks/multicollinearity_data_check.py .                                                                                 [ 38%]
evalml/data_checks/natural_language_nan_data_check.py .                                                                              [ 44%]
evalml/data_checks/outliers_data_check.py F                                                                                          [ 50%]
evalml/data_checks/sparsity_data_check.py .                                                                                          [ 55%]
evalml/data_checks/target_distribution_data_check.py F                                                                               [ 61%]
evalml/data_checks/target_leakage_data_check.py .                                                                                    [ 66%]
evalml/data_checks/uniqueness_data_check.py .                                                                                        [ 72%]
evalml/model_understanding/prediction_explanations/_algorithms.py .                                                                  [ 77%]
evalml/pipelines/component_graph.py F                                                                                                [ 83%]
evalml/problem_types/utils.py .                                                                                                      [ 88%]
evalml/tuners/grid_search_tuner.py .                                                                                                 [ 94%]
evalml/tuners/random_search_tuner.py .
```

PR #3002, opened 2021-11-02
Add more doctests

ParthivNaresh:
Completes #2936


evalml/data_checks/datetime_format_data_check.py
```diff
@@ -25,12 +25,12 @@ def validate(self, X, y):
         Returns:
             dict (DataCheckError): List with DataCheckErrors if unequal intervals are found in the datetime column.
 
-        Example:
+        Examples:
```
angela97lin, line 28: Could just be personal opinion, but I think if we add this many examples (which is great!), it gets a little harder to read. Maybe what we need here is an extra line above each example to better understand what each example is trying to convey? From a quick glance, it's a bit hard to gauge the purpose of each example 😬

ParthivNaresh: Sure thing, so maybe a comment above a test briefly explaining what it does?

evalml/data_checks/outliers_data_check.py
```diff
@@ -44,6 +44,19 @@ def validate(self, X, y=None):
             ...                   "details": {"columns": ["z"], "rows": [3], "column_indices": {"z": [3]}}}],
             ...     "errors": [],
             ...     "actions": [{"code": "DROP_ROWS", "metadata": {"rows": [3], "columns": None}}]}
+            ...
+            >>> assert OutliersDataCheck.get_boxplot_data(df['z']) == {
+            ...     'score': 0.888652177485398,
```
freddyaboulton, line 49: I wonder if there's a way we can round so that we don't check so many significant figures. If gamma.cdf is off in the ten-thousands place in a future release we're cooked.

evalml/objectives/standard_metrics.py
```diff
@@ -63,7 +81,13 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
 
 
 class BalancedAccuracyMulticlass(MulticlassClassificationObjective):
-    """Balanced accuracy score for multiclass classification."""
+    """Balanced accuracy score for multiclass classification.
+
+    Example:
+        >>> y_true = pd.Series([0, 1, 0, 2, 0, 1, 2, 1, 2, 0, 2])
+        >>> y_pred = pd.Series([0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2])
+        >>> assert BalancedAccuracyMulticlass().objective_function(y_true, y_pred) == 0.5555555555555555
```
freddyaboulton, line 89: Here I'm less worried about the significant figures because I think these objective definitions coming from sklearn are slow to change but our objective unit tests use `np.assert_almost_equal` so maybe we should too?

evalml/objectives/utils.py
```diff
@@ -125,6 +125,27 @@ def get_core_objectives(problem_type):
 
     Returns:
         List of ObjectiveBase instances
+
+    Examples:
+        >>> for objective in get_core_objectives("regression"):
```
freddyaboulton, line 130: Cool!

evalml/pipelines/components/utils.py
```diff
@@ -274,6 +281,13 @@ def generate_component_code(element):
 
     Raises:
         ValueError: If the input element is not a component instance.
+
+    Examples:
+        >>> from evalml.pipelines.components.estimators.regressors.decision_tree_regressor import DecisionTreeRegressor
+        >>> assert generate_component_code(DecisionTreeRegressor()) == "from evalml.pipelines.components.estimators.regressors.decision_tree_regressor import DecisionTreeRegressor\n\ndecisionTreeRegressor = DecisionTreeRegressor(**{'criterion': 'mse', 'max_features': 'auto', 'max_depth': 6, 'min_samples_split': 2, 'min_weight_fraction_leaf': 0.0})"
+        ...
```
freddyaboulton, line 288: Docs don't build without the `r` prefix? Is it cause of these three dots here? The example renders fine in the doc so I'm just wondering what happened but no need to change anything.

evalml/preprocessing/utils.py
```diff
@@ -62,6 +62,31 @@ def split_data(
 
     Returns:
         pd.DataFrame, pd.DataFrame, pd.Series, pd.Series: Feature and target data each split into train and test sets.
+
+    Examples:
+        >>> X = pd.DataFrame([1, 2, 3, 4, 5, 6], columns=["First"])
+        >>> y = pd.Series([8, 9, 10, 11, 12, 13])
+        >>> splits = split_data(X, y, "regression", random_seed=42)
```
freddyaboulton, line 69: Might be clearer if we unpack the tuple as `X_train, X_validation, y_train, y_validation = split_data(...)`

evalml/data_checks/highly_null_data_check.py
```diff
@@ -84,7 +110,7 @@ def validate(self, X, y=None):
             percent_null_rows >= self.pct_null_row_threshold
         ]
         if len(highly_null_rows) > 0:
-            warning_msg = f"{len(highly_null_rows)} out of {len(X)} rows are more than {self.pct_null_row_threshold*100}% null"
+            warning_msg = f"{len(highly_null_rows)} out of {len(X)} rows are {self.pct_null_row_threshold*100}% or more null"
```
freddyaboulton, line 113: Same comment as in the sparsity data check: Just wondering we were modifying this.

freddyaboulton (approved):
@ParthivNaresh Thank you for this! This looks good to me. I agree with @angela97lin that since some docstrings test different things, it might be helpful to add some in-line comments above each one to document what's happening. 

Other than that, I think it would be helpful if some of the `assert float1 == float2` checks used rounding so that these don't start to flake if we're off by negligible amount in a future release.

eccabay (approved):
What a beast of a change! Thanks for doing this, just left a few tiny tiny semantic nitpicks

docs/source/release_notes.rst
```diff
@@ -12,6 +12,7 @@ Release Notes
     * Documentation Changes
         * Fixed cost benefit matrix demo formatting :pr:`2990`
         * Update ReadMe.md with new badge links and updated installation instructions for conda :pr:`2998`
+        * Adding more comprehensive doctests :pr:`3002`
```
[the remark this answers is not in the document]
ParthivNaresh: I'm so bad with this, thanks 😄

evalml/data_checks/highly_null_data_check.py
ParthivNaresh: Because the threshold check is `percent_null_rows >= self.pct_null_row_threshold`, so the phrasing `or more` seemed more appropriate than `more than`

evalml/data_checks/sparsity_data_check.py
```diff
@@ -72,26 +79,31 @@ def validate(self, X, y=None):
...
-                },
-            ).to_dict()
-        )
-        results["actions"].append(
-            DataCheckAction(
-                action_code=DataCheckActionCode.DROP_COL,
-                metadata={"columns": too_sparse_cols},
-            ).to_dict()
-        )
+        if too_sparse_cols:
```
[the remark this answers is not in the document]
ParthivNaresh: If none of the columns are returned from the `sparsity_score`, results` still appends a `warning` and `actions`. The warning message ends up being "Input columns ({}) for multiclass problem type are too sparse."`

evalml/pipelines/components/utils.py
ParthivNaresh: The lint check doesn't pass, I think due to the `\n` characters in the output

evalml/data_checks/datetime_format_data_check.py
angela97lin: Exactly!

PR #3011, opened 2021-11-04
Added AutoML function to access ensemble pipeline's input pipelines IDs

### LINKED ISSUE #3008: AutoMLSearch: build API to access IDs of ensemble pipeline's input pipelines
**Background**
Ensemble models compute predictions from a group of input models, then apply a learning algorithm to combine those predictions into a prediction which is more accurate overall.

Each of the stacked ensembler pipelines built by our automl is constructed by grabbing a few pipelines from the automl leaderboard, and then building a graph where each of the input pipelines' predictions are provided as inputs to the stacked ensembler component. The resulting graph contains a full copy of each of the input pipelines.

**Proposal**
Add an API to our automl search which allows users to look up the IDs of the pipelines used by a particular ensembler pipeline.

This would allow users to dig further into the details of each input pipeline if they want to understand the dynamics of their stacked ensembler better.

An initial thought at the API design is below
```python
automl = AutoMLSearch(...)
automl.search(X, y)
# let's say we look at the rankings and see that an ensembler has ID 42
ensembler = automl.get_pipeline(42)
# we can now use this pipeline to compute predictions, scores, stats, model understanding, etc.
ensembler.fit(X, y)
ensembler.predict(X)
...

# and in addition to grabbing the ensembler pipeline in full, we can grab the list of IDs of the input pipelines
input_pipeline_ids = automl.get_ensembler_input_pipelines(42)
for pipeline_id in input_pipeline_ids:
    pipeline = automl.get_pipeline(pipeline_id)
    pipeline.fit(X, y)
    print(pipeline.feature_importance)
    ...

with pytest.raises(Exception):
    automl.get_ensembler_input_pipelines(7) # raise exception if pipeline isn't an ensembler
```

@christopherbunn FYI

christopherbunn:
Resolves #3008

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1590,3 +1590,28 @@ class AutoMLSearch:
     @property
     def _sleep_time(self):
         return self._SLEEP_TIME
+
+    def get_ensembler_input_pipelines(self, ensemble_pipeline_id):
+        """Score a list of pipelines on the given holdout data.
+
+        Args:
+            ensemble_pipeline_id (id): Ensemble pipeline ID to get input pipeline IDs from.
+
+        Returns:
+            list[int]: A list of ensemble input pipeline IDs.
+
+        Raises:
+            ValueError: If `ensemble_pipeline_id` does not correspond to a valid ensemble pipeline ID.
+        """
+        pipeline_results = self._results["pipeline_results"]
+        if (
+            ensemble_pipeline_id not in pipeline_results.keys()
+            or "input_pipeline_ids" not in pipeline_results[ensemble_pipeline_id].keys()
+        ):
+            raise ValueError(
+                f"Pipeline ID {ensemble_pipeline_id} is not a valid ensemble pipeline"
+            )
+
+        return self._results["pipeline_results"][ensemble_pipeline_id][
+            "input_pipeline_ids"
+        ]
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1592,7 +1592,7 @@ class AutoMLSearch:
         return self._SLEEP_TIME
 
     def get_ensembler_input_pipelines(self, ensemble_pipeline_id):
-        """Score a list of pipelines on the given holdout data.
+        """Returns a list of input pipeline IDs given an ensembler pipeline ID.
 
         Args:
             ensemble_pipeline_id (id): Ensemble pipeline ID to get input pipeline IDs from.
@@ -1605,8 +1605,8 @@ class AutoMLSearch:
         """
         pipeline_results = self._results["pipeline_results"]
         if (
-            ensemble_pipeline_id not in pipeline_results.keys()
-            or "input_pipeline_ids" not in pipeline_results[ensemble_pipeline_id].keys()
+            ensemble_pipeline_id not in pipeline_results
+            or "input_pipeline_ids" not in pipeline_results[ensemble_pipeline_id]
         ):
             raise ValueError(
                 f"Pipeline ID {ensemble_pipeline_id} is not a valid ensemble pipeline"
--- a/evalml/tests/automl_tests/test_automl.py
+++ b/evalml/tests/automl_tests/test_automl.py
@@ -5363,3 +5363,47 @@ def test_automl_respects_iterative_pipeline_order(X_y_binary, AutoMLTestEnv):
         searched_model_families,
         key=lambda model_family: _ESTIMATOR_FAMILY_ORDER.index(model_family),
     )
+
+
+def test_get_ensembler_input_pipelines(X_y_binary, AutoMLTestEnv):
+    X, y = X_y_binary
+    automl = AutoMLSearch(
+        X_train=X,
+        y_train=y,
+        problem_type="binary",
+        max_iterations=_get_first_stacked_classifier_no(),
+        objective="Log Loss Binary",
+        ensembling=True,
+        optimize_thresholds=False,
+        error_callback=raise_error_callback,
+    )
+
+    score_side_effect = [
+        {"Log Loss Binary": score}
+        for score in np.arange(
+            0, -1 * automl.max_iterations * automl.data_splitter.get_n_splits(), -0.1
+        )
+    ]  # Decreases with each call
+
+    test_env = AutoMLTestEnv("binary")
+    with test_env.test_context(mock_score_side_effect=score_side_effect):
+        automl.search()
+
+    best_pipeline_ids = [
+        pipeline["id"]
+        for pipeline in list(automl._automl_algorithm._best_pipeline_info.values())
+    ]
+    assert (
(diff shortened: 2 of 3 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -5363,3 +5363,83 @@ def test_automl_respects_iterative_pipeline_order(X_y_binary, AutoMLTestEnv):
...
+        pipeline["id"]
+        for pipeline in list(automl._automl_algorithm._best_pipeline_info.values())
+    ]
+    final_best_pipeline_ids.sort()
+
+    input_pipeline_0_ids = automl.get_ensembler_input_pipelines(ensemble_ids[0])
+    input_pipeline_0_ids.sort()
+
+    input_pipeline_1_ids = automl.get_ensembler_input_pipelines(ensemble_ids[1])
+    input_pipeline_1_ids.sort()
```
christopherbunn, line 5434: Here, we're rerunning AutoMLSearch so that we can get two ensemble pipelines (and to make sure that the IDs for best pipeline are updated). I did this since there isn't a super easy way to verify that the first ensemble pipeline generated is correct. Def open to ideas/suggestions!

evalml/automl/automl_search.py
```diff
@@ -1590,3 +1590,28 @@ def plot(self):
...
+        Returns:
+            list[int]: A list of ensemble input pipeline IDs.
+
+        Raises:
+            ValueError: If `ensemble_pipeline_id` does not correspond to a valid ensemble pipeline ID.
+        """
+        pipeline_results = self._results["pipeline_results"]
+        if (
+            ensemble_pipeline_id not in pipeline_results.keys()
+            or "input_pipeline_ids" not in pipeline_results[ensemble_pipeline_id].keys()
```
freddyaboulton, line 1609: nit: We can get kid of `.keys()` here.

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Thoughts on rewriting [test_describe_pipeline_with_ensembling](https://github.com/alteryx/evalml/blob/main/evalml/tests/automl_tests/test_automl.py#L1469) to also call this new method. There what we did was check that the input pipeline ids for the second ensemble are all greater than the input ids of the first ensemble since the scores is always decreasing.

I'm ok with whatever you decide. This looks great to me as is!

evalml/automl/automl_search.py
```diff
@@ -1590,3 +1590,28 @@ def plot(self):
     @property
     def _sleep_time(self):
         return self._SLEEP_TIME
+
+    def get_ensembler_input_pipelines(self, ensemble_pipeline_id):
+        """Score a list of pipelines on the given holdout data.
```
freddyaboulton, line 1595: Docstring first sentence doesn't match what the method does.

freddyaboulton (approved):
Thank you @christopherbunn !!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5363,3 +5363,83 @@ def test_automl_respects_iterative_pipeline_order(X_y_binary, AutoMLTestEnv):
...
+        max_batches=two_stacking_batches,
+        objective="Log Loss Binary",
+        ensembling=True,
+    )
+
+    test_env = AutoMLTestEnv("binary")
+    with test_env.test_context(mock_score_side_effect=score_side_effect):
+        automl.search()
+    pipeline_names = automl.rankings["pipeline_name"]
+    assert pipeline_names.str.contains("Ensemble").any()
```
angela97lin, line 5417: Is this assert carried over from a different ensemble test? I noticed that we don't test it above, and not exactly what we're testing here so we could remove it?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5363,3 +5363,83 @@ def test_automl_respects_iterative_pipeline_order(X_y_binary, AutoMLTestEnv):
...
+
+    test_env = AutoMLTestEnv("binary")
+    with test_env.test_context(mock_score_side_effect=score_side_effect):
+        automl.search()
+
+    best_pipeline_ids = [
+        pipeline["id"]
+        for pipeline in list(automl._automl_algorithm._best_pipeline_info.values())
+    ]
+    best_pipeline_ids.sort()
```
angela97lin, line 5394: Q: What do you think about just sorting the output for `get_ensembler_input_pipelines`? Is there ever a case where we wouldn't want to?

angela97lin (approved):
Looks good! I wonder if it's worth sorting the output of `get_ensembler_input_pipelines`, but otherwise 🚢

evalml/tests/automl_tests/test_automl.py
christopherbunn: Yep, removed this line.

evalml/tests/automl_tests/test_automl.py
christopherbunn: @angela97lin I think the pipeline IDs come out in model family order. In the case of the IterativeAlgorithm, it comes out in the order of the first batch. It's a minor detail but this is the same order that is fed into `_create_ensemble`, so all of the input pipelines in the ensemble pipeline's graph are in this order from top-down.

evalml/tests/automl_tests/test_automl.py
christopherbunn: I am open to either, but I'm leaning slightly towards keeping it as-is. If we need to update `get_ensembler_input_pipelines()` in the future there's a single point to update + it keeps this particular edge case together. The runtime difference for another AutoMLSearch is relatively minimal for this mocked example.

evalml/tests/automl_tests/test_automl.py
angela97lin: @christopherbunn Isn't `_best_pipeline_info` a dictionary? So the order isn't guaranteed, unless we're assuming python order from dict --> list is the order in which items are added to the dict?

christopherbunn: @angela97lin From Python 3.7 onwards it's guaranteed to be insertion order: https://stackoverflow.com/questions/60848/how-do-you-retrieve-items-from-a-dictionary-in-the-order-that-theyre-inserted

angela97lin: Ah right! 😂 
Okay back to the original question then: is it better to sort or not? Sounds like there's no benefit to doing so LGTM! My original question was because I noticed we were sorting the outputs in the tests to compare and wondered if there was any value to changing the method to sort. It's not too difficult to add later if we change our minds :)

christopherbunn: Ah gotcha! Yeah to be clear I think it would be best to retain insertion order for the output rather than sort it, but it's not necessary for now. I have it sorted for this test case so it's easier to compare values.

PR #3017, opened 2021-11-07
Fix permutation importance failing when target is categorical

### LINKED ISSUE #3012: Permutation importance fails on some objectives when target is categorical
Repro:
```
import pandas as pd
from evalml.model_understanding import calculate_permutation_importance

X = pd.DataFrame({'a': range(100)})
y = pd.Series(["Yes" if i%2==0 else "No" for i in range(100)])

component_graph = [
        "Imputer",
        "One Hot Encoder",
        "Elastic Net Classifier",
    ]
pipeline = BinaryClassificationPipeline(component_graph)
pipeline.fit(X, y)

permutation_importance = calculate_permutation_importance(
        pipeline, X, y, 'precision'
)
```

Also fails for accuracy, F1, and others. Does not fail for log loss, I'm not sure why. Discovered while trying to generate `readable_explanation` examples.

angela97lin:
Closes #3012

### DIFF
--- a/evalml/model_understanding/permutation_importance.py
+++ b/evalml/model_understanding/permutation_importance.py
@@ -5,6 +5,7 @@ from joblib import Parallel, delayed
 
 from evalml.objectives.utils import get_objective
 from evalml.problem_types import is_classification
+from evalml.problem_types.utils import is_regression
 from evalml.utils import infer_feature_types
 
 
@@ -324,6 +325,7 @@ def _fast_scorer(pipeline, features, X, y, objective):
         preds = pipeline.estimator.predict_proba(features)
     else:
         preds = pipeline.estimator.predict(features)
-        preds = pipeline.inverse_transform(preds)
+        if is_regression(pipeline.problem_type):
+            preds = pipeline.inverse_transform(preds)
     score = pipeline._score(X, y, preds, objective)
     return score if objective.greater_is_better else -score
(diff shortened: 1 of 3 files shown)


jeremyliweishih (approved):
nice!

eccabay (approved):
LGTM, thanks for picking this up! Super useful in enabling human readable pipeline explanations :)

chukarsten (commented):
Do we need to address the classification aspect that freddy mentioned in the issue?

evalml/model_understanding/permutation_importance.py
```diff
@@ -324,6 +325,7 @@ def _fast_scorer(pipeline, features, X, y, objective):
         preds = pipeline.estimator.predict_proba(features)
     else:
         preds = pipeline.estimator.predict(features)
-        preds = pipeline.inverse_transform(preds)
+        if is_regression(pipeline.problem_type):
```
freddyaboulton, line 328: I think we're ok for classification here. In `_fast_permutation_importance` we encode the target with the pipeline._encode_targets method. So the estimator predictions will also be encoded.

Of course, if the pipeline does not have an encoder but has string-valued targets this would fail but I would say it's a pipeline definition bug as opposed to a permutation importance bug.

The fact that our classification objectives only supports integer-valued targets and that to score a pipeline it should have an encoder may not be super clear though.

freddyaboulton (approved):

evalml/model_understanding/permutation_importance.py
angela97lin: Agreed! I wish there were a way to make this more clear. Pipelines require a label encoder for not only scoring but also fit so it still feel consistent, but I wonder if moving forward we could make these types of error messages more clear.

PR #3018, opened 2021-11-08
Data check Update

ParthivNaresh:
A data check update and suspension of the nullable types check until the Nullable work is merged.

-----


evalml/data_checks/outliers_data_check.py
```diff
@@ -128,8 +128,7 @@ def get_boxplot_data(data_):
         Returns:
             dict: A payload of box plot statistics.
         """
-        if not data_.ww._schema:
-            data_.ww.init()
+        data_ = infer_feature_types(data_)
```
chukarsten, line 131: I thought that we had come down on having like an arg for infer_feature_types like `infer_feature_types(ignore_nullable_types=True)` where the default value was false?  I thought that way we could leave up the testing and general protection while the nullable types work was planned out.  Either we agreed on that or, chances are incredibly high, I completely misremember.

ParthivNaresh: Ah gotcha, I must have misremembered that, I'll revert it to an arg being passed in

chukarsten (approved):
Cool man!  I'll add this to the nullable types doc and we can get you unblocked.

Issue #2874, opened 2021-10-05
Add clustering as a problem type


eccabay:
We should also make sure users get useful errors if they try to use AutoMLSearch or other functions irrelevant to clustering problem types.

PR #3022, opened 2021-11-08
Add clustering as a problem type

### LINKED ISSUE #2872: Allow no y input in `AutoMLSearch`

eccabay:
Closes #2872 and #2874


evalml/pipelines/utils.py
```diff
@@ -213,6 +214,10 @@ def make_pipeline(
     Raises:
         ValueError: If estimator is not valid for the given problem type, or sampling is not supported for the given problem type.
     """
+    problem_type = handle_problem_types(problem_type)
+    if is_clustering(problem_type):
+        raise ValueError("Pipelines for clustering problems are not supported yet")
+
     X = infer_feature_types(X)
     y = infer_feature_types(y)
```
bchen1116, line 223: You should be able to remove the call to `problem_type = handle_problem_types(problem_type)` below in the `if estimator` call since you now define it up top.

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -488,6 +488,8 @@ def __eq__(self, series_2):
 def test_default_data_checks_across_problem_types(problem_type):
     default_data_check_list = DefaultDataChecks._DEFAULT_DATA_CHECK_CLASSES
 
+    if is_clustering(problem_type):
+        return
```
bchen1116, line 492: Curious if we should do a `pytest.skip()` here rather than an empty return, the primary reason being so that it'd be easier to find and remember this is something we need to return to to fix in the future once we do get clustering incorporated.

bchen1116 (approved):
Looks solid! Left a comment on potentially using `pytest.skip` rather than return for some tests, and a small nit. Is there work underway to add clustering estimators? Not sure if it's a bit early to add clustering logic to AutoMLSearch. 

The impl looks good to me, but curious what others think about adding clustering logic to AutoMLSearch!

PR #3027, opened 2021-11-09
Add test to ensure binary pipeline maintains threshold after pickling

bchen1116:
Testing thresholds


evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -2877,3 +2878,31 @@ def test_component_graph_pipeline_initialized():
         ]
         == "median"
     )
+
+
+def test_pickled_pipeline_preserves_threshold(
```
freddyaboulton, line 2883: Can we add some of this logic to `test_serialization_protocol` or `test_load_pickled_pipeline_with_custom_objective` instead?

bchen1116: @freddyaboulton both of those tests seem to use `cloudpickle`, but I wanted to use `pickle`. Would it be better to edit those tests to include my pickle test, or leave the test as is?

freddyaboulton: Got it, I see! I wonder if it's possible to merge the tests. We could have a different test fixture than `pickled_pipeline_path` that uses pickle instead of cloudpickle and move some of the logic from this test in to `test_load_pickled_pipeline_with_custom_objective`. 

Up to you though. If it's better to keep the tests separate, I'm down for it but maybe let's move this test next to `test_load_pickled_pipeline_with_custom_objective` so that the pickling tests are at least grouped.

freddyaboulton (approved):
Looks good to me @bchen1116 ! Left a non-blocking suggestion for how to organize these tests.

PR #3033, opened 2021-11-10
Add ability for users to define which class is "positive" for label encoder in binary classification case

### LINKED ISSUE #2926: Binary classification: ability for users to define which class is "positive"
**Background**
For classification problems, the pipelines created by evalml's automl search will use a label encoder as the first step. This component maps each unique value in the target to an integer value ranging from 0 to `n-1`, where `n` in this case is the total number of unique values in the target.

Currently, the label encoder will define this mapping using [lexicographic](https://en.wikipedia.org/wiki/Lexicographic_order) (aka alphabetical) order. As an example of how that works, let's say I had a few binary classification datasets with the following two unique values in the target:
* "True" and "False" will be mapped to 1 and 0 respectively.
* "Correct" and "Incorrect" will be mapped to 0 and 1 respectively.
* "Positive" and "Negative" will be mapped to 1 and 0 respectively.
* "Zero" and "One" will be mapped to 1 and 0 respectively.
* The integers 0 and 1 will not be mapped and will stay as 0 and 1.
You get the idea.

**Problem**
What's good about this? We have deterministic behavior which is easy to understand: which class is considered positive is determined by alphabetical ordering.

What's bad about this? When we compute things like F1, precision, recall, precision-recall curve, AUC, ROC curve, confusion matrix, etc., those values all depend on the definition of which class is considered to be the positive class and which class is considered to be the negative class.

**Workaround**
If users want to change which class is considered positive, they can modify their target column to label the positive class as integer 1 and negative class as integer 0. This will produce the desired behavior throughout evalml's feature set.

**Feature Request**
For binary classification problems, let's define a way for users to tell us which class is positive and which is negative, rather than always only relying on the label encoder's default behavior.

**Notes**
My recommendation: allow the label encoder to accept the name of the positive class as an optional parameter, so that every pipeline is properly configured with which class value should be positive.

Also as an aside, I suggest we use the `problem_config` to represent this kind of info in the automl search API, following the pattern for how we represent time series parameters like gap and max delay.

angela97lin:
Closes #2926.

General question: since we have specific behavior in the label encoder... should we just move away from using sklearn's and completely write our own? There isn't too much additional functionality we get from using sklearn's impl but we could end up with different stack traces depending on whether or not positive_label is defined (if defined, uses our stack traces and errors, if not, uses sklearn's).

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/label_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/label_encoder.py
@@ -86,7 +86,7 @@ class LabelEncoder(Transformer):
                 f"y contains previously unseen labels: {y_unique_values.difference(self.mapping.keys())}"
             )
         y_t = y_ww.map(self.mapping)
-        return X, ww.init_series(y_t)
+        return X, ww.init_series(y_t, logical_type="integer")
 
     def fit_transform(self, X, y):
         """Fit and transform data using the label encoder.
--- a/evalml/pipelines/components/transformers/encoders/label_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/label_encoder.py
@@ -12,6 +12,7 @@ class LabelEncoder(Transformer):
     """A transformer that encodes target labels using values between 0 and num_classes - 1.
 
     Args:
+        positive_label (int, str): The label for the class that should be treated as positive (1) for binary classification problems. Ignored for multiclass problems. Defaults to None.
         random_seed (int): Seed for the random number generator. Defaults to 0. Ignored.
     """
 
@@ -22,8 +23,8 @@ class LabelEncoder(Transformer):
     modifies_features = False
     modifies_target = True
 
-    def __init__(self, random_seed=0, **kwargs):
-        parameters = {}
+    def __init__(self, positive_label=None, random_seed=0, **kwargs):
+        parameters = {"positive_label": positive_label}
         parameters.update(kwargs)
 
         label_encoder_obj = SKLabelEncoder()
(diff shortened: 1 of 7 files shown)


freddyaboulton (commented):
@angela97lin Thank you for picking up this issue! I think we can simplify the implementation if we just forgo the sklearn implementation and roll our own. 

I think this will work:

```python
    def fit(self, X, y):
        """Fits the label encoder.

        Args:
            X (pd.DataFrame): The input training data of shape [n_samples, n_features]. Ignored.
            y (pd.Series): The target training data of length [n_samples].

        Returns:
            self

        Raises:
            ValueError: If input `y` is None.
        """
        if y is None:
            raise ValueError("y cannot be None!")
        y = infer_feature_types(y)
        sorted_uniques = sorted(y.unique())
        self.mapping = {val:i for i, val in enumerate(sorted_uniques)}
        if self.parameters["positive_label"] is not None:
            if len(self.mapping) != 2:
                raise ValueError(
                    "positive_label should only be set for binary classification targets. Otherwise, positive_label should be None."
                )
            self.mapping = {val: int(val == self.parameters["positive_label"]) for val in sorted_uniques}
        self.inverse_mapping = {val: i for i, val in self.mapping.items()}

        return self
```

the in transform and inverse_transform we just call `map` on the series with the right mapping. 

Maybe there's a performance hit but we'll have to check.

evalml/pipelines/components/transformers/encoders/label_encoder.py
```diff
@@ -67,9 +80,13 @@ def transform(self, X, y=None):
...
-        y_t = self._component_obj.transform(y_ww)
-        y_t = pd.Series(y_t, index=y_ww.index)
-        return X, ww.init_series(y_t)
+        y_unique_values = set(y_ww.unique())
+        if y_unique_values.difference(self.mapping.keys()):
+            raise ValueError(
+                f"y contains previously unseen labels: {y_unique_values.difference(self.mapping.keys())}"
+            )
+        y_t = y_ww.map(self.mapping)
+        return X, ww.init_series(y_t, logical_type="integer")
```
angela97lin, line 89: Without setting the logical type, we end up with a categorical column which defeats the purpose 😂

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -261,6 +260,7 @@ def predict_in_sample(self, X, y, X_train, y_train, objective=None):
             proba = proba.iloc[:, 1]
             if objective is None:
                 predictions = proba > self.threshold
+                predictions = predictions.astype(int)
```
angela97lin, line 263: This is necessary because it returns bools, which we don't know how to map back (since we expect ints!). Although we treated bools and 0/1 as interchangeable, it actually matters now that we're storing a mapping.

Previously with sklearn this wasn't an issue because they used `np.searchsorted` which would treat ints and bools the same. This is no longer true. :)

bchen1116 (approved):
Nice! This looks good to me, solid test coverage!

freddyaboulton (approved):
Looks good to me @angela97lin ! Thank you

PR #3034, opened 2021-11-10
Add `data_check_name` attribute to each data check action

### LINKED ISSUE #2179: DataChecks: ability to tell which actions came from which data checks
Request from @dancuarini, we should have a code that maps data check actions to natural language strings. For example, we will want distinguish between a no variance col and a potential ID col action. Right now, both return DROP_COL.

angela97lin:
Closes #2179.


bchen1116 (approved):
LGTM!

PR #3044, opened 2021-11-15
Fix bug on force plots to show feature values for specific rows in training data

peterataylor:
…Fixes bug where only first row feature values are show in all row plots.

### Pull Request Description

Fixes bug as described in https://github.com/alteryx/evalml/issues/3045.

-----


freddyaboulton:
@peterataylor Do you need help getting all the checks to pass?

peterataylor:
> @peterataylor Do you need help getting all the checks to pass?

Thanks for the offer @freddyaboulton ! Having a quick look I should be able to do the release notes. But any guidance on the lint and dependency changes would be great. But I will look over the contrib guide again and check the approach.

freddyaboulton:
@peterataylor The "Detect dependency changes" should be fixed as soon as you merge in the latest `main` branch.

If you have the development dependencies installed, you should be able to do the following from the top level of the repo:
```
make lint-fix
make lint
```

If `make lint` still fails, the errors should hopefully be clear. 

If you do not have make installed you can look at the Makefile to see what those commands mean and run then yourself. 

Let me know if you have any other questions!

evalml/model_understanding/force_plots.py
```diff
@@ -45,12 +40,13 @@ def gen_force_plot(shap_values, training_data, expected_value, matplotlib):
         initjs()
 
     shap_plots = force_plot(pipeline, rows_to_explain, training_data, y)
-    for row in shap_plots:
+    for ix, row in enumerate(shap_plots):
```
freddyaboulton, line 43: @peterataylor This looks good to me! Thanks for the contribution. I think the one thing that's missing is maybe a unit test to make sure we don't commit the same mistake again in the future. Maybe do one with and without a custom index on `training_data`?

peterataylor: Thanks @freddyaboulton . I've added some tests to test_force_plots that use the existing parameter sets (rows 0 and 0-4). It required inspection of the returned plot object to ensure the feature values matched those that correspond to the rows_to_explain. I confirmed the tests failed with previous version and pass now. You will probably want to review though. Thanks!

freddyaboulton (approved):
@peterataylor This looks good to me! Thank you for the contribution.

PR #3047, opened 2021-11-15
Update integration tests to use `make_pipeline_from_actions` instead of private method

### LINKED ISSUE #3003: Update data checks integration tests to use new `make_pipeline_from_actions` instead of private method
We should update data checks integration tests to use new `make_pipeline_from_actions` instead of private method (`_make_component_list_from_actions`).

angela97lin:
Closes #3003.


freddyaboulton (approved):
Thanks @angela97lin !

PR #3049, opened 2021-11-15
Simplify `graph_json` output and fix bugs

### LINKED ISSUE #3038: Simplify output of `Pipeline.graph_json()` function
We should reduce down the amount of non-essential information returned by `graph_json` to get down to the core (nodes, edges) structure of DAGs.

eccabay:
Closes #3038 

Also fixes two bugs that prevented some pipelines from getting JSON serialized:
- Ensemble components store an entire estimator object as the "final_estimator" parameter. This code saves that estimator as "final_estimator_name" and "final_estimator_parameters" instead.
- JSON cannot serialize the `np.int64` datatype, which is the type returned by `Integer` ranges for hyperparameter searches. This can be changed within the `Integer()` call, but that was too many places across the code base for this PR, so I left the fix within `graph_json` itself. In the future, however, it might be better practice to move over to having `Integer` generate `int` types instead of `np.int64`.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,7 +4,10 @@ Release Notes
     * Enhancements
         * Added ``NumWords`` and ``NumCharacters`` primitives to ``TextFeaturizer`` and renamed ``TextFeaturizer` to ``NaturalLanguageFeaturizer`` :pr:`3030`
     * Fixes
+        * Fixed bug where ensembling components could not get converted to JSON format :pr:`3049`
+        * Fixed bug where components with tuned integer hyperparameters could not get converted to JSON format :pr:`3049`
     * Changes
+        * Updated the ``Pipeline.graph_json`` function to return a dictionary of "from" and "to" edges instead of tuples :pr:`3049`
     * Documentation Changes
     * Testing Changes
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,6 @@ Release Notes
         * Fixed bug where ensembling components could not get converted to JSON format :pr:`3049`
         * Fixed bug where components with tuned integer hyperparameters could not get converted to JSON format :pr:`3049`
     * Changes
-        * Updated the ``Pipeline.graph_json`` function to return a dictionary of "from" and "to" edges instead of tuples :pr:`3049`
     * Documentation Changes
     * Testing Changes
 
@@ -15,6 +14,7 @@ Release Notes
 
     **Breaking Changes**
         * Renamed ``TextFeaturizer` to ``NaturalLanguageFeaturizer`` :pr:`3030`
+        * Updated the ``Pipeline.graph_json`` function to return a dictionary of "from" and "to" edges instead of tuples :pr:`3049`
 
 
 **v0.37.0 Nov. 9, 2021**
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -450,8 +450,8 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
                     param_dict[f"{param}_name"] = val.name
                     param_dict[f"{param}_parameters"] = val.parameters
                 else:
-                    if np.issubdtype(type(val), np.number):
-                        val = val.item()
+                    if isinstance(val, np.int64):
+                        val = int(val)
                     param_dict[param] = val
             nodes[comp_] = {"Parameters": param_dict, "Name": att_.name}
 
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -450,7 +450,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
                     param_dict[f"{param}_name"] = val.name
                     param_dict[f"{param}_parameters"] = val.parameters
                 else:
-                    if isinstance(val, np.int64):
+                    if np.issubdtype(type(val), np.number):
                         val = int(val)
                     param_dict[param] = val
             nodes[comp_] = {"Parameters": param_dict, "Name": att_.name}
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -10,11 +10,12 @@ from abc import ABC, abstractmethod
 from collections import OrderedDict
 
 import cloudpickle
+import numpy as np
 import pandas as pd
 
 from .components import (
-    ComponentBase,
     PCA,
+    ComponentBase,
     DFSTransformer,
     Estimator,
     LinearDiscriminantAnalysis,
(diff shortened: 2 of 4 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -439,31 +441,44 @@ def graph_json(self):
...
+        for comp_, att_ in self.component_graph.component_instances.items():
+            param_dict = {}
+            for param, val in att_.parameters.items():
+                # Can't JSON serialize components directly, have to split them into name and parameters
+                if isinstance(val, ComponentBase):
+                    param_dict[f"{param}_name"] = val.name
+                    param_dict[f"{param}_parameters"] = val.parameters
+                else:
+                    if isinstance(val, np.int64):
+                        val = int(val)
```
eccabay, line 454: This is the fix for the `int64` bug. There is still room for a problem to occur here, if there is a tuned integer hyperparameter within an ensembling component. However, the default final estimator at this point is `ElasticNet`, which doesn't have any integer hyperparameters, so this won't be an issue in the default case.

I think we should leave this for another issue, where we move all the `Integer` hyperparameter ranges to return `int`s instead of `np.int64`, but if people prefer I could add another catch in here for ensemble parameters.

ParthivNaresh: Thanks for catching this, I say go for it. If users want to change the final estimator in ensembling this should at least give them an idea of why they can't get the `graph` for their component graph just yet.

ParthivNaresh (approved):
Looks great thanks for addressing this!

evalml/pipelines/pipeline_base.py
```diff
@@ -432,38 +434,51 @@ def graph_json(self):
...
+        nodes = {}
+        for comp_, att_ in self.component_graph.component_instances.items():
+            param_dict = {}
+            for param, val in att_.parameters.items():
+                # Can't JSON serialize components directly, have to split them into name and parameters
+                if isinstance(val, ComponentBase):
+                    param_dict[f"{param}_name"] = val.name
+                    param_dict[f"{param}_parameters"] = val.parameters
+                else:
+                    if isinstance(val, np.int64):
```
freddyaboulton, line 453: @eccabay I think we'll also have a problem with np.bool or any non-json-serializable parameter. Users can do pretty pathological things, like the following:

```python
import pandas as pd
import numpy as np
import json
import pytest

X = pd.DataFrame({"a": [1, 2, 2, 1],
                  "b": ["2021-01-01", "2021-01-02", "2021-01-03", "2021-01-04"]})
y = pd.Series([1, 2, 3])

pl = RegressionPipeline(["Delayed Feature Transformer"],
                        parameters={"Delayed Feature Transformer": {"delay_features": np.bool_(False)}})
pl.fit(X, y)
with pytest.raises(TypeError, match="Object of type bool_ is not JSON serializable"):
    pl.graph_json()
```

or 

```python
import pandas as pd
import json
import pytest

X = pd.DataFrame({"a": [1, 2, 3, 3, 4, 4, None],
                  "b": range(7)})
y = pd.Series([1, 2, 3])

pl = RegressionPipeline(["Imputer"],
                        parameters={"Imputer": {"numeric_fill_value": np.float32(2),
                                                "numeric_impute_strategy": "constant"}})
pl.fit(X, y)
with pytest.raises(TypeError, match="Object of type float32 is not JSON serializable"):
    pl.graph_json()
```

Users could also define their own components that take in pd.Series or np arrays which are also not json serializable. 

Throwing out an alternative, we could just return the python dict and let the user serialize it if they want to serialize it?

eccabay: You raise a very interesting question. 

You're completely correct in that users can set parameters that will break this code. I can pretty easily update this code so it catches any `numpy.number` dtypes, but there's no good way to safely convert and serialize all non-serializable cases. However, with these changes we finally support serializing output from any *default* AutoML search, and that should be guaranteed. 

We have a specific use case driving the need for this function, I'd hate to take away its functionality purely because it may not be able to JSON serialize non-JSON serializable types. I would rather add a bit of hardening here to throw our own error if a pipeline is not serializable, rather than change the underlying goal of this method.

evalml/pipelines/pipeline_base.py
```diff
@@ -432,38 +434,51 @@ def graph_json(self):
...
             dag_json (str): A serialized JSON representation of a DAG structure.
         """
-        nodes = {
-            comp_: {"Attributes": att_.parameters, "Name": att_.name}
-            for comp_, att_ in self.component_graph.component_instances.items()
-        }
-
-        x_edges = self.component_graph._get_edges(
+        nodes = {}
+        for comp_, att_ in self.component_graph.component_instances.items():
```
angela97lin, line 445: Just a thought but since we do something like this in LG, would overwriting the default encoder help us here? It's very similar to the impl you have here where we check the value of the attribute but it keeps all of the checking of attribute isolated in an object:
```
def dict_to_json_str(dict_in, indent=4):
    """Given a dictionary of values, convert to a JSON string. Safely handles np types as well.

    Args:
        dict_in (dict): the dictionary to be serialized.
        indent (int): the string indent to use on each newline.

    Returns:
        str: the dict as JSON.
    """

    class NpEncoder(json.JSONEncoder):
        def default(self, obj):
            if isinstance(obj, np.integer):
                return int(obj)
            elif isinstance(obj, np.floating):
                return float(obj)
            elif isinstance(obj, np.ndarray):
                return obj.tolist()
            elif isinstance(obj, PipelineBase) or isinstance(obj, ComponentBase):
                return obj.parameters
            try:
                return super(NpEncoder, self).default(obj)
            except TypeError as e:
                logger.warn(f"TypeError using the default encoder: {str(e)}; Will use string representation instead.")
                return str(obj)
    return json.dumps(dict_in, indent=indent, cls=NpEncoder)
```

docs/source/release_notes.rst
```diff
@@ -4,7 +4,10 @@ Release Notes
     * Enhancements
         * Added ``NumWords`` and ``NumCharacters`` primitives to ``TextFeaturizer`` and renamed ``TextFeaturizer` to ``NaturalLanguageFeaturizer`` :pr:`3030`
     * Fixes
+        * Fixed bug where ensembling components could not get converted to JSON format :pr:`3049`
+        * Fixed bug where components with tuned integer hyperparameters could not get converted to JSON format :pr:`3049`
     * Changes
+        * Updated the ``Pipeline.graph_json`` function to return a dictionary of "from" and "to" edges instead of tuples :pr:`3049`
```
freddyaboulton, line 10: This might technically be a breaking change?

freddyaboulton (approved):
@eccabay Thanks for adding this! 

I still prefer letting users encode the output themselves as opposed to erroring out on some pipelines but I'm ok with the implementation you have now. If we decide to always try to encode to JSON, then it may be good to incorporate @angela97lin 's custom encoder so that we can catch tricky bools or other data types. That can happen in a follow up PR though.

evalml/pipelines/pipeline_base.py
eccabay: This is very interesting, I didn't know it existed! This also could tie in with what @freddyaboulton was talking about as well. It feels out of scope for this specific PR, but I'll file a new issue to track discussion around hardening this method and/or returning a dict instead.

Issue #3000, opened 2021-11-02
DataCheckActions user guide formatting


chukarsten:
Attached is a picture of https://evalml.alteryx.com/en/latest/user_guide/data_actions.html.  We should probably either fix the formatting here by figuring out how to maybe get the offending index list to ellipsis (which seems difficult given that I think it's an underlying python native list in a python dictionary), or figure out a new way to deal with this.  It makes the user guide a little unwieldy!

![image](https://user-images.githubusercontent.com/64713315/139883615-20a3b72d-968f-4669-89f5-0b7649fa4462.png)

angela97lin:
@chukarsten There are at least two things we could do:
1. Remove "null_row_indices" from the output of the data check.
2. Find a way to format this properly. We could simply not print out the output that's producing this long string. Calling print(...) for the statement rather than just having jupyter run it produces a more condensed output but still one that's unwieldy:
![image](https://user-images.githubusercontent.com/5422235/139923601-9e6806b7-52c6-4b21-8911-10cd1cd7ef6f.png)


Thoughts? I think I like option 1 better, unless we feel a need to store this info somewhere, for cases just like this!

If we have a large dataset and the highly null check triggers, it'll always print out a long list of indices!

freddyaboulton:
I vote 1 too but I think we added it for data health? @ParthivNaresh might now. I don't think the data health _should_ need all these indices though so maybe we can refactor?

ParthivNaresh:
Data health does need this for the final payload. Maybe we could add this into a static method which could be called separately by data health but won't be necessary for the data checks here?

PR #3050, opened 2021-11-15
Update docs to use data check action methods rather than manually cleaning data

### LINKED ISSUE #3004: Update docs to use `make_pipeline_from_actions` to address data checks
Right now, in our docs, [we manually address the data check warnings and errors](https://evalml.alteryx.com/en/latest/user_guide/data_actions.html#Addressing-all-warnings-and-errors). Now that we have a public method to do so (`make_pipeline_from_actions`), we should update our docs to use it!

angela97lin:
Closes #3004. Docs here: https://feature-labs-inc-evalml--3050.com.readthedocs.build/en/3050/user_guide/data_actions.html

Main docs here: https://evalml.alteryx.com/en/stable/user_guide/data_actions.html

Q: In the documentation, there's a section where we manually address errors and later show that not addressing warnings leads to worse performance. However, data check actions don't differentiate between warnings/errors and severity of the action.

We could either:
- Remove this section. Reasoning being that we're showcasing *actions*, and this is manual cleaning
- Keep as is. It's lame that the error-cleaning section is manual, but there's still a point we get across that data check warnings are important and useful to address to increase model performance
- Add functionality to actions to only address / return components if we'll error out in search. I'm not fully convinced about the usefulness of this method outside of this case.

### DIFF
--- a/docs/source/user_guide/data_actions.ipynb
+++ b/docs/source/user_guide/data_actions.ipynb
@@ -205,7 +205,7 @@
     "actions_pipeline = make_pipeline_from_actions(\"binary\", actions)\n",
     "actions_pipeline.fit(X_train, y_train)\n",
     "X_train_cleaned, y_train_cleaned = actions_pipeline.transform(X_train, y_train)\n",
-    "print(\"The new length of X_train is {} and y_train is {}\".format(len(X_train),len(y_train)))"
+    "print(\"The new length of X_train is {} and y_train is {}\".format(len(X_train_cleaned),len(X_train_cleaned)))"
    ]
   },
   {
--- a/docs/source/user_guide/data_actions.ipynb
+++ b/docs/source/user_guide/data_actions.ipynb
@@ -380,7 +380,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.9"
+   "version": "3.8.6"
   }
  },
  "nbformat": 4,
--- a/docs/source/user_guide/data_actions.ipynb
+++ b/docs/source/user_guide/data_actions.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Understanding Data Check Actions\n",
     "\n",
-    "EvalML streamlines the creation and implementation of machine learning models for tabular data. One of the many features it offers is [data checks](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html), which help determine the health of the our data before we train a model on it. These data checks have associated actions with them and will be shown in this notebook. In our default data checks, we have the following checks:\n",
+    "EvalML streamlines the creation and implementation of machine learning models for tabular data. One of the many features it offers is [data checks](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html), which help determine the health of our data before we train a model on it. These data checks have associated actions with them and will be shown in this notebook. In our default data checks, we have the following checks:\n",
     "\n",
     "- `HighlyNullDataCheck`: Checks whether the rows or columns are highly null\n",
     "\n",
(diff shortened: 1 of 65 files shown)


jeremyliweishih:
haven't taken a look at the PR itself but the string for `null_row_indices` is ginormous, could we alter the dataset to prodouce less null rows or truncate the list somehow?

angela97lin:
@jeremyliweishih Yah, that's what https://github.com/alteryx/evalml/issues/3000 addresses! I'll probably pick that up this sprint too :)

chukarsten (approved):
I agree with your option of removing the manual data cleaning.  That doesn't really highlight what EvalML brings to the table.  DataChecks/Actions and utilize it to do the cleaning in a quick and convenient way, does.  I filed [this issue](https://github.com/alteryx/evalml/issues/3072) to follow up on this work, should we decide to do so.  I think this a good move, though.

bchen1116 (approved):
LGTM! I think either option1 or 2 would work. I do see value in showing that addressing all warnings/errors would be the best option to have better search results, but it also makes sense to showcase what EvalML can do versus manual cleaning.

docs/source/user_guide/data_actions.ipynb
```diff
@@ -242,27 +243,28 @@
...
-    "# The first action states to drop the row given by the action code\n",
-    "X_train.drop(1477, axis=0, inplace=True)\n",
-    "# we must also drop this for y since we are removing its associated feature input\n",
-    "y_train.drop(index=1477, inplace=True)\n",
-    "\n",
+    "from evalml.pipelines.utils import make_pipeline_from_actions\n",
+    "from evalml.data_checks import DataCheckAction\n",
+    "\n",
+    "# Convert dictionary form of actions returned from data check output dictionary as DataCheckAction objects\n",
+    "actions = [\n",
```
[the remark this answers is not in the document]
angela97lin: I love, love this idea and very much agree--thank you for filing! 🙏

docs/source/user_guide/data_actions.ipynb
```diff
@@ -4,9 +4,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Understanding Data Actions\n",
+    "# Understanding Data Check Actions\n",
     "\n",
-    "EvalML streamlines the creation and implementation of machine learning models for tabular data. One of the many features it offers is [data checks](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html), which are geared towards determining the health of the data before we train a model on it. These data checks have associated actions with them and will be shown in this notebook. In our default data checks, we have the following checks:\n",
+    "EvalML streamlines the creation and implementation of machine learning models for tabular data. One of the many features it offers is [data checks](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html), which help determine the health of the our data before we train a model on it. These data checks have associated actions with them and will be shown in this notebook. In our default data checks, we have the following checks:\n",
```
freddyaboulton, line 9: typo: "of the our data"

docs/source/user_guide/data_actions.ipynb
```diff
@@ -242,27 +243,28 @@
...
    "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
-    "# The first action states to drop the row given by the action code\n",
-    "X_train.drop(1477, axis=0, inplace=True)\n",
-    "# we must also drop this for y since we are removing its associated feature input\n",
-    "y_train.drop(index=1477, inplace=True)\n",
-    "\n",
+    "from evalml.pipelines.utils import make_pipeline_from_actions\n",
```
freddyaboulton, line 256: I think we can now get rid of the `In the future, we aim to provide a helper function to allow users to quickly clean the data by taking in the list of actions and creating an appropriate pipeline of transformers to alter the data` line below?

freddyaboulton (approved):
@angela97lin I think it's fine to keep option 2 you presented. I think it's pretty clear what's happening and it presents users with two different ways to clean their data prior to search. I don't think it's lame that one section is manual while the other isn't. Some users may prefer to do manual cleaning anyways.

I guess we can consider making highly null columns an error instead of a warning to side-step this point?

Other than that, just two minor nits. Looks good to me!

angela97lin:
@freddyaboulton @chukarsten @bchen1116 It sounds like there's no clear consensus on what's the better option here. Here are my thoughts after reading your comments:

I agree with @chukarsten's comment that we want to highlight what EvalML can bring to the table. I think adding the section about manual cleaning detracts from this since it doesn't get straight to the point of what we can provide. I'm going to *move the section about addressing via `make_pipeline_from_actions` above the manual cleaning section*.

However, we can still keep the manual cleaning section, since it could provide users with an idea of how they could address comments by looking at the output of data check actions.

LMK if you have any objections :)

PR #3051, opened 2021-11-15
Add support for scikit-learn>=1.0

angela97lin:
Opening up a branch, enabling `scikit-learn` v1.0, and seeing what breaks :)

### DIFF
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -2,7 +2,7 @@ numpy>=1.20.0
 numba==0.53
 pandas>=1.3.0
 scipy>=1.5.0
-scikit-learn>=0.24.0,<1.0
+scikit-learn>=0.24.0
 scikit-optimize>=0.8.1
 pyzmq>=20.0.0
 colorama>=0.4.4
--- a/evalml/tests/model_understanding_tests/test_graphs.py
+++ b/evalml/tests/model_understanding_tests/test_graphs.py
@@ -1403,7 +1403,7 @@ def test_t_sne_errors_marker_size(marker_size, has_minimal_dependencies):
 
 @pytest.mark.parametrize("data_type", ["np", "pd", "ww"])
 @pytest.mark.parametrize("perplexity", [0, 4.6, 100])
-@pytest.mark.parametrize("learning_rate", [100.0, -15, 0])
+@pytest.mark.parametrize("learning_rate", [100.0, 0.1])
 def test_graph_t_sne(data_type, perplexity, learning_rate):
     go = pytest.importorskip(
         "plotly.graph_objects",
--- a/evalml/tests/objective_tests/test_standard_metrics.py
+++ b/evalml/tests/objective_tests/test_standard_metrics.py
@@ -3,7 +3,6 @@ from itertools import product
 import numpy as np
 import pandas as pd
 import pytest
-from sklearn.metrics import matthews_corrcoef as sk_matthews_corrcoef
 
 from evalml.objectives import (
     F1,
@@ -662,8 +661,6 @@ def test_mse_linear_model():
 def test_mcc_catches_warnings():
     y_true = [1, 0, 1, 1]
     y_predicted = [0, 0, 0, 0]
-    sk_matthews_corrcoef(y_true, y_predicted)
-    # assert "invalid value" in str(record[-1].message)
     with pytest.warns(None) as record:
         MCCBinary().objective_function(y_true, y_predicted)
         MCCMulticlass().objective_function(y_true, y_predicted)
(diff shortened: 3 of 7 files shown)


evalml/tests/objective_tests/test_standard_metrics.py
```diff
@@ -662,9 +662,8 @@ def test_mse_linear_model():
 def test_mcc_catches_warnings():
     y_true = [1, 0, 1, 1]
     y_predicted = [0, 0, 0, 0]
-    with pytest.warns(RuntimeWarning) as record:
```
angela97lin, line 665: Between [0.24.0](https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/metrics/_classification.py#L875) and [1.0.0+](https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/metrics/_classification.py#L906), the implementation was changed to just return a matthews_corrcoef score of 0.0 if the denominator was 0 (previously caused invalid divide by 0)

core-requirements.txt
```diff
@@ -2,7 +2,7 @@ numpy>=1.20.0
 numba==0.53
 pandas>=1.3.0
 scipy>=1.5.0
-scikit-learn>=0.24.0,<1.0
+scikit-learn>=0.24.0
```
angela97lin, line 5: Note:
scikit-learn >1.0 will require scikit-optimize>=0.9.0 because modules / objects get shuffled around, causing the following:
```
evalml/pipelines/components/estimators/classifiers/logistic_regression_classifier.py:4: in <module>
    from skopt.space import Real
../evalml_venv/lib/python3.7/site-packages/skopt/__init__.py:55: in <module>
    from .searchcv import BayesSearchCV
../evalml_venv/lib/python3.7/site-packages/skopt/searchcv.py:16: in <module>
    from sklearn.utils.fixes import MaskedArray
E   ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes' (/Users/angela.lin/Desktop/evalml_venv/lib/python3.7/site-packages/sklearn/utils/fixes.py)```

Our tests, which automatically install the latest version of scikit-optimize (0.9.0) will thus not break, so maybe we don't need to set our requirement of `scikit-optimize>=0.9.0` but open to discussion if it feels like we should.

File: evalml/tests/model_understanding_tests/test_graphs.py [hunk 3]
```diff
@@ -1403,7 +1403,7 @@ def test_t_sne_errors_marker_size(marker_size, has_minimal_dependencies):
 
 @pytest.mark.parametrize("data_type", ["np", "pd", "ww"])
 @pytest.mark.parametrize("perplexity", [0, 4.6, 100])
-@pytest.mark.parametrize("learning_rate", [100.0, -15, 0])
```
angela97lin, line 1406: Negative and zero values are no longer accepted: https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/manifold/_t_sne.py#L816

freddyaboulton (approved):
@angela97lin This looks good to me!

core-requirements.txt
freddyaboulton: How did you run into this?

angela97lin: My local environment uses `scikit-optimize==0.8.1` so as soon as I tried using scikit-learn 1.0, things broke 🥲

freddyaboulton: I see. I'm wondering if this will affect users in some way. Did you just install scikit-learn or did you do `pip install -r requirements.txt` ?

angela97lin: I just installed `scikit-learn`! Our tests didn't break because they call `pip install -r requirements.txt` and in a clean environment, 0.9.0 will be installed.

This is a known issue in the scikit-optimize repo where it seems like the community solution is to either use scikit-learn<=0.24.2 or use scikit-optimize>=0.9.0: https://github.com/scikit-optimize/scikit-optimize/issues/569

freddyaboulton: Cool, makes sense to me!

bchen1116: Would it be simpler to make the core requirements use `scikit-learn >= 1.0.0` versus having support for both but test coverage for the later version?

evalml/tests/objective_tests/test_standard_metrics.py
```diff
@@ -662,9 +662,8 @@ def test_mse_linear_model():
...
-        sk_matthews_corrcoef(y_true, y_predicted)
-        assert "invalid value" in str(record[-1].message)
+    sk_matthews_corrcoef(y_true, y_predicted)
+    # assert "invalid value" in str(record[-1].message)
```
bchen1116, line 666: Should we remove this comment?

bchen1116 (approved):
LGTM! Is there a reason we chose to support `>=0.24` versus `>=1.0.0`?

core-requirements.txt
angela97lin: Hmm, I can be persuaded otherwise but I think this aligns with what we've currently been doing aka every time our dependencies come out with a new version, we pin until we can fix the tests to support the latest version--and thats the version that we run on!

PR #3071, opened 2021-11-17
Relax contraints on test set for time series

### LINKED ISSUE #3070: Relax constraints on test set for time series
Currently, the test set for time series must meet two criteria

* Length must match `forecast_horizon`
* The index must start gap + 1 rows from the test set. 

These are encoded in [_validate_holdout_set](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/time_series_pipeline_base.py#L96) as well as our [time series user guide.](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/time_series_pipeline_base.py#L96).

We can relax this as follows:
* Length <= `forecast_horizon`
* We can set the index on the test set ourselves. This is better UX, because the test set is usually not a part of the training + validation data, so the index will likely start at 0.

freddyaboulton:
### Pull Request Description
Fixes #3070 

-----

### DIFF
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -106,7 +106,7 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
             ValueError: If holdout data does not have forecast_horizon entries or if datasets
                 are not separated by gap.
         """
-        right_length = len(X) == self.forecast_horizon
+        right_length = len(X) <= self.forecast_horizon
         X_separated_by_gap = self._are_datasets_separated_by_gap(
             X_train.index, X.index, self.gap
         )
@@ -116,7 +116,7 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
                 "and its index needs to "
                 f"start {self.gap + 1} values ahead of the training index. "
                 f"Data received - Length X: {len(X)}, "
-                f"X index start: {X.index[0]}, X_train index end {X.index[-1]}."
+                f"X index start: {X.index[0]}, X_train index end {X_train.index[-1]}."
             )
 
     def _add_training_data_to_X_Y(self, X, y, X_train, y_train):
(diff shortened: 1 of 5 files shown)


evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -247,8 +247,11 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
                 "Cannot call predict() on a component graph because the final component is not an Estimator."
             )
         X = infer_feature_types(X)
+        X.index = self._move_index_forward(
+            X_train.index[-X.shape[0] :], self.gap + X.shape[0]
+        )
```
ParthivNaresh, line 252: I like this implementation but I had a question about it. Previously if a user were to set a forecast for indices that were more than `gap` units ahead of the training data, the second part of the warning above would be relevant: `its index needs to start {self.gap + 1} values ahead of the training index`.

Screen Shot 2021-11-17 at 5 58 04 PM

Now the same forecast attempt doesn't raise the error, but still sets the forecasts to the end of the training set.

Screen Shot 2021-11-17 at 5 42 34 PM

This is really useful for when the test set starts at 0 but if it continues the index from the end of the training set, this might give a false impression of future predictions. Maybe we should add a check here for test indices that begin more than `gap` observations away from the training set so the original error can still be raised.

ParthivNaresh (changes requested):
Solid work mate, just left a question about our implementation of `_move_index_forward`. I think it might be difficult for OS users to understand the limitations of what they can do with forecasting time series when the test set is more than `gap` observations beyond the training set.

evalml/pipelines/time_series_pipeline_base.py
freddyaboulton: Great question @ParthivNaresh ! I think the reason why I didn't want the check to be based on the index is that the index will probably not start at `gap` + 1 units away from the training/validation index. 

My reasoning is that the kind of examples we've been working off, where we manually split the same dataset into separate chunks are not reflective of what happens in "real life". Users will give one dataset to AutoMLSearch (that will have index from 0 to X.shape[0] or whatever) and then the test set will come from a different csv or file. When you read that in with pandas the index will start at 0. I feel like it's awkward for users to have to re-index when we can do that calculation for them.

That being said, it does assume the user is uploading data that's temporally right after the dataset given to AutoMLSearch. We can check that via the new `date_index` parameter that we're now requiring! 

I was actually thinking of doing that in a separate PR but the implementation should be small enough to do in this one maybe?

ParthivNaresh: I absolutely agree with the intent behind this. Users are almost guaranteed in the vast majority of cases to pass in a separate dataset for testing and forecasting from index 0, so this works great for that.

In the case that a user passes in a test set separately but tries to do something similar to above, where they might try and forecast values starting from `[50 : ]` as opposed to starting with the first observation, they'll run into a similar issue. But if that's going in as part of a separate PR I'm totally good with leaving it out of this one!

ParthivNaresh (approved):

PR #3080, opened 2021-11-18
Add pipeline threshold to confusion matrix returns

### LINKED ISSUE #3079: Have find_confusion_matrix_per_thresholds return the confusion matrix at the pipeline threshold as well
Through discussion with @dsherry @freddyaboulton, it would be ideal if the `find_confusion_matrix_per_thresholds` method also returns the confusion matrix at the set pipeline threshold.

bchen1116:
fix #3079 

There are 2 potential issues that I want to raise, although they are not severe enough to block this PR from being merged. 
First is the aesthetic of the index when it is very granular, as can be seen here:
image
Note that the threshold is `0.99999...`, but looking at the dataframe, it just shows as `1.0`, which could be confusing for OS users. By grabbing the index itself,, we can see the actual value. We could stringify the index to have them appear, but this might be annoying for our internal use:
image

The other is when our optimized pipeline threshold doesn't match the best threshold chosen through this method. This can be seen in the problem above, but also can be seen here:
image

Above, the pipeline is optimized using `accuracy binary`, and we see the pipeline threshold `0.360321` actually has a worse performance value compared to `0.5` with accuracy, which is the ideal value `find_confusion_matrix_per_thresholds` finds. The differences here are likely how we're finding the optimal threshold, with our `optimize_thresholds` using gradient descent, and our current method using a simple linear scan. Is this disparity an issue for our users? Discussing with @freddyaboulton, it could be confusing when our optimal thresholds don't match up. However, what would be the best approach to fix this, if necessary?

Again, these two issues shouldn't block the merge of this PR, but are both things I wanted to bring up for discussion.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -11,7 +11,7 @@ Release Notes
         * Added in error message when fit and predict/predict_proba data types are different :pr:`3036`
         * Fixed bug where ensembling components could not get converted to JSON format :pr:`3049`
         * Fixed bug where components with tuned integer hyperparameters could not get converted to JSON format :pr:`3049`
-        * Included confusion matrix at the pipeline threshold for ``find_confusion_matrix_per_threshold`` :pr:``
+        * Included confusion matrix at the pipeline threshold for ``find_confusion_matrix_per_threshold`` :pr:`3080`
     * Changes
         * Delete ``predict_uses_y`` estimator attribute :pr:`3069`
     * Documentation Changes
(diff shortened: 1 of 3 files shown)


chukarsten (approved):
This looks good to me!

angela97lin (commented):
Interesting, thanks for outlining these two concerns! Is the first issue because of the way pandas internally handles things? aka it just rounds up rather than printing the actual value? If so, I'd say it's not a big deal given the actual index is still what we expect it to be.

bchen1116:
@angela97lin yep, that was the first concern! Just an aesthetic issue. Filed an issue to address the second [here](https://github.com/alteryx/evalml/issues/3086)

ParthivNaresh (approved):
Looks good man, thanks for filing an issue for the second point. I think we're fine with the aesthetic issue for now.

PR #3085, opened 2021-11-19
Return details from `TargetDistributionDataCheck` as floats rather string

### LINKED ISSUE #2943: Return details from TargetDistributionDataCheck as floats rather than / in addition to string
Right now, we calculate whether or not a target has a lognormal distribution and add details if so like this:
```
details = {
  f"{normalization_test_string}-statistic/pvalue": f"{round(norm_test_og.statistic, 1)}/{round(norm_test_og.pvalue, 3)}"
            }
```
It might be useful to return the values from the strings (`round(norm_test_og.statistic, 1)`, `round(norm_test_og.pvalue, 3)`, `round(norm_test_og.statistic, 1)/round(norm_test_og.pvalue, 3)`), separately, rather than or in addition in a formatted string? That way, if the user needs the value or wants to use it elsewhere, they can easily without having to parse the given string.

angela97lin:
Closes #2943. I chose not to keep the string representation and moved the calculations into a separate helper method, but open to discussion if we feel like the string representation is helpful in addition to the float values.

### DIFF
--- a/evalml/data_checks/target_distribution_data_check.py
+++ b/evalml/data_checks/target_distribution_data_check.py
@@ -38,7 +38,7 @@ class TargetDistributionDataCheck(DataCheck):
             ...                   "data_check_name": "TargetDistributionDataCheck",
             ...                   "level": "warning",
             ...                   "code": "TARGET_LOGNORMAL_DISTRIBUTION",
-            ...                   "details": {"shapiro-statistic/pvalue": '0.8/0.045', "columns": None, "rows": None}}],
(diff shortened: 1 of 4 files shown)


evalml/data_checks/target_distribution_data_check.py
```diff
@@ -142,3 +126,26 @@ def validate(self, X, y):
             )
 
         return results
+
+
+def _detect_log_distribution_helper(y):
```
angela97lin, line 131: Moving mathy-implementation chunk into helper method to make `validate` easier to read!

jeremyliweishih (approved):
very nice!

Issue #3057, opened 2021-11-16
Nullable Types: Build Preprocessing Component


chukarsten:
As an EvalML developer, I want a `ReplaceNullableTypes` component so that I can handle the presence of `NullableBoolean` and `NullableInteger` in data fed to `AutoMLSearch()` via the `make_pipeline()` function in our preprocessing pipeline creation.

PR #3090, opened 2021-11-22
ReplaceNullableTypes Component

chukarsten:
Addresses #3057


evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,117 @@
...
+            y (pd.Series, optional): Ignored.
+
+        Returns:
+            pd.DataFrame: Transformed X
+        """
+        X_t = infer_feature_types(X, ignore_nullable_types=True)
+        for col in self._nullable_int_cols:
+            X_t[col] = replace_nullable_int(X_t[col])
+        for col in self._nullable_bool_cols:
+            X_t[col] = replace_nullable_bool(X_t[col])
```
freddyaboulton, line 116: I think we need to go through ww or else the schema will be invalidated. Here is an example from the unit tests where the logical and physical types don't match after transform and downstream components fail: 

image

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,117 @@
...
+    """Function to determine whether a column in a dataframe is a pandas NullableBoolean type.
+
+    Args:
+        df: Input dataframe to check.
+        col: Column of the input dataframe to check.
+
+    Returns:
+        boolean: Whether the column was of NullableBoolean type.
+    """
+    is_pandas_nullable = isinstance(df.dtypes[col], pca.boolean.BooleanDtype)
```
freddyaboulton, line 47: Given that we're calling `infer_feature_types` in `transfrom`, we know woodwork schema will be set. Can we get rid of the dtypes checks?

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,152 @@
...
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features].
+            y (pd.Series, optional): The target training data of length [n_samples].
+
+        Returns:
+            self
+        """
+        X_t = infer_feature_types(X, ignore_nullable_types=True)
+        for col in X_t.columns:
+            if is_nullable_int(X_t[col]):
```
ParthivNaresh, line 106: I think adding `X_t[col].ww.init()` before checking nullable types will allow us to remove the `is_pandas_nullable` checks

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,152 @@
...
+    return series.astype("category")
+
+
+class ReplaceNullableTypes(Transformer):
+    """Transformer to replace features with the new nullable dtypes with a dtype that is compatible in EvalML."""
+
+    name = "Replace Nullable Types Transformer"
+    hyperparameter_ranges = {}
+    """{}"""
+
```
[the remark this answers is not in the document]
chukarsten: Good call, I found this the hard way. 😂

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
chukarsten: I chatted with Tamar about the best way to do this and I think we should be good now.

evalml/tests/component_tests/test_nullable_types_replacer.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        # If the datatypes are not set properly, it does nothing.
+        assert nullable_types_replacer._nullable_int_cols == []
+        assert nullable_types_replacer._nullable_bool_cols == []
+
+    X_t, y_t = nullable_types_replacer.transform(X)
+    assert set(X_t.columns) == set(X.columns)
+    assert X_t.shape == X.shape
+
+    if nullable_types_properly_set:
```
freddyaboulton, line 71: nit-pick: Can we also check the ww logical types are properly set (`Double`, `Categorical`).

Makefile
```diff
@@ -85,6 +85,14 @@ installdeps:
 	pip install --upgrade pip -q
 	pip install -e . -q
 
+.PHONY: installdeps-min
```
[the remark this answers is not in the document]
freddyaboulton: We should use this in the CI job? Maybe in a follow-up pr.

freddyaboulton (changes requested):
@chukarsten Thank you for this! I think the behavior and testing is solid. I left some minor comments on how to stream-line the implementation.

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,171 @@
...
+    Raises:
+        AttributeError: If s
+    """
+    try:
+        is_pandas_nullable = isinstance(series.dtype, pca.integer.Int64Dtype)
+    except AttributeError:
+        is_pandas_nullable = False
+
+    try:
+        is_ww_nullable = isinstance(series.ww.logical_type, IntegerNullable)
```
tamargrey, line 34: Woodwork has an `AgeNullable` logical type that will have the nullable integer pandas dtype, so it'd have `is_pandas_nullable == True` from line 29 but then get `is_ww_nullable` as False here. Would you want to include `AgeNullable` here as well, or should this just be for logical types that can be inferred by Woodwork?

(I recognize that in cases where the woodwork typing info matches the sereis, this won't be a problem, but if the typing info is no longer valid for the series, you may still want an `AgeNullable` logical type to be indicated here)

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,171 @@
...
+        """Fits component to data.
+
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features].
+            y (pd.Series, optional): The target training data of length [n_samples].
+
+        Returns:
+            self
+        """
+        X_t = infer_feature_types(X, ignore_nullable_types=True)
```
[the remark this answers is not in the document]
tamargrey: I like this simplification! My question about whether to include `AgeNullable` still stands, though.

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,152 @@
...
+            y (pd.Series, optional): Ignored.
+
+        Returns:
+            pd.DataFrame: Transformed X
+        """
+        X_t = infer_feature_types(X, ignore_nullable_types=True)
+        for col in self._nullable_int_cols:
+            new_col = replace_nullable_int(X_t[col])
+            X_t.ww.pop(col)
+            X_t.ww[col] = new_col
```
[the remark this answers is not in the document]
thehomebrewnerd: I might not be following the flow of everything exactly, but I think `new_col` will not have Woodwork initialized, in which case Woodwork will infer the logical type of this column. That's probably safe and fast based on the dtype of the column, but if you want to be safer you could initialize WW on the new column before assigning here.

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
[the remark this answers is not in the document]
freddyaboulton: Yea I think we should include `AgeNullable`. After this component, all nullable types should be convert to non-nullable equivalent. Great point!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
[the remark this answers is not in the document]
chukarsten: Done!

Makefile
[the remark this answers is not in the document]
chukarsten: This is already done in the github workflows, specifically the min deps CI workflows.  I just put it in here so people can build a test environment to run them.  If there's a better way that people do it, I can take this out!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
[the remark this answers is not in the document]
chukarsten: Donezo

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
chukarsten: Included!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,171 @@
...
+    Categorical,
+    Double,
+    IntegerNullable,
+)
+
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils import infer_feature_types
+
+
+def is_nullable_int(series):
```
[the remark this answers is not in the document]
chukarsten: Changed :)

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,171 @@
...
+    """Function to determine whether a column in a dataframe is a pandas NullableInteger type.
+
+    Args:
+        series: The series to detect NullableIntegers in.
+
+    Returns:
+        boolean: Whether the column was of NullableInteger type.
+
+    Raises:
+        AttributeError: If s
```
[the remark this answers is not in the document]
chukarsten: Yup, thanks!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,171 @@
...
+        series: The column to change the type of.
+
+    Returns:
+        pandas.Series: The Dataframe column with the type changed to float64.
+
+    """
+    return series.astype("float64")
+
+
+def is_nullable_bool(series):
```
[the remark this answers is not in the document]
chukarsten: I think this is largely re-written!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -72,13 +74,9 @@ def transform(self, X, y=None):
         """
         X_t = infer_feature_types(X, ignore_nullable_types=True)
         for col in self._nullable_int_cols:
-            new_col = X_t[col].astype("float64")
-            X_t.ww.pop(col)
-            X_t.ww[col] = new_col
+            X_t.ww[col] = X_t[col].astype("float64")
```
tamargrey, line 78: Having looked at the conversation [here](https://github.com/alteryx/evalml/pull/3090#discussion_r759685131), I think it would be preferable to go with what @thehomebrewnerd suggested and explicitly initialize Woodwork on the `new_col` with the logical type that you're trying to have for the column.

Setting the dtype like this means that you're relying on Woodwork's inference to not then choose to infer the column as an unexpected logical type (for example, what if woodwork checked whether all values were actually integers and then initialized it as `IntegerNullable` again). It could also end up being a bit faster to just do the type conversion within Woodwork and not also have to do type inference.

chukarsten: Done!

evalml/tests/component_tests/test_nullable_types_replacer.py
```diff
@@ -0,0 +1,136 @@
...
+        assert isinstance(X_t.ww.logical_types["non_nullable_age"], Age)
+    else:
+        assert isinstance(X_t.ww.logical_types["non_nullable_age"], Integer)
+    assert isinstance(X_t.ww.logical_types["nullable_age"], Double)
+    assert isinstance(X_t.ww.logical_types["non_nullable_boolean"], Boolean)
+    assert isinstance(X_t.ww.logical_types["nullable_boolean"], Categorical)
+
+
+@pytest.mark.parametrize("input_type", ["ww", "pandas"])
+def test_replace_nullable_types_boolean_target(nullable_data, input_type):
```
freddyaboulton, line 92: nit-pick: Verify that y_t has a woodwork schema

evalml/pipelines/components/transformers/transformer.py
```diff
@@ -55,9 +55,9 @@ def fit_transform(self, X, y=None):
         Raises:
             MethodPropertyNotFoundError: If transformer does not have a transform method or a component_obj that implements transform.
         """
-        X_ww = infer_feature_types(X)
+        X_ww = infer_feature_types(X, ignore_nullable_types=True)
```
freddyaboulton, line 58: I think I prefer we implement fit_transform in ReplaceNullableTypes component until we're done with this epic as opposed to modifying the parent class code? This assumes the nullable types have been replaced prior to being passed to fit_transform which isn't true yet.

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,92 @@
...
+        for col in self._nullable_int_cols:
+            X_t.ww[col] = init_series(X_t[col], logical_type="double")
+        for col in self._nullable_bool_cols:
+            X_t.ww[col] = init_series(X_t[col], logical_type="categorical")
+
+        if y is not None:
+            y_t = infer_feature_types(y, ignore_nullable_types=True)
+            if self._nullable_target is not None:
+                if self._nullable_target == "nullable_int":
+                    y_t = init_series(y_t.astype("float64"), logical_type=Double)
```
freddyaboulton, line 86: nit: Is there a reason the target is init with `Double` as opposed to `"double"`. If possible I'd prefer if it was consistent.

freddyaboulton (approved):
Thanks for making the changes @chukarsten ! This looks good to me!

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
chukarsten: Changed!  There wasn't I just was learning what types that argument accepted.

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
```diff
@@ -0,0 +1,92 @@
...
+                elif self._nullable_target == "nullable_bool":
+                    y_t = init_series(y_t.astype("category"), logical_type=Categorical)
```
tamargrey, line 88: similar to the init series calls above, you shouldn't need the `astype` call here and on line 86; `init_series` will take care of the dtype conversion!

evalml/pipelines/components/transformers/transformer.py
chukarsten: Done!

evalml/tests/component_tests/test_nullable_types_replacer.py
chukarsten: Gucci

evalml/pipelines/components/transformers/preprocessing/replace_nullable_types.py
tamargrey: Awesome--music to my ears!

PR #3093, opened 2021-11-23
Remove dependency on `psutil` package

### LINKED ISSUE #2993: Deprecate the CLI
We've seen an example of someone uninstalling featuretools because of it's psutil dependency: https://github.com/intel-analytics/analytics-zoo/issues/3051#issuecomment-724528724

We currently also use psutil only for our cli. I wonder if we can get rid of the cli? How much value does it add? Maybe we can slim down our dependencies in doing so.

Just wanted to start the conversation!

angela97lin:
Closes #2993. Rather than deprecating the utils, I just removed the dependency on `psutils` which we used to show CPU metrics (# cores, memory).

Happy to discuss idea of removing the CLI entirely too!

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -35,7 +35,6 @@ outputs:
         - colorama >=0.4.4
         - cloudpickle >=1.5.0
         - click >=7.1.2
-        - psutil >=5.6.6
         - requirements-parser >=0.2.0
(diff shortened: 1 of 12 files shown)


freddyaboulton (approved):

chukarsten (approved):

Issue #3091, opened 2021-11-22
Do not set date index on data/target for Arima component


freddyaboulton:
I was chatting about this with @ParthivNaresh earlier today.

In our ARIMA component, we set the pandas index of the features and target to be the `date_index` parameter. If that `date_index` does not have a frequency that can be inferred by `pd.infer_freq`, ARIMA will error out. If we don't set the index to be the `date_index`, the component does not error out.

```python
import pandas as pd
import numpy as np
from sktime.forecasting.arima import AutoARIMA


oct_dates = pd.Series(pd.date_range("2021-10-01", "2021-10-31"))
nov_dates = pd.Series(pd.date_range("2021-11-03", "2021-11-15"))
dates = pd.concat([oct_dates, nov_dates]).reset_index(drop=True)
df = pd.DataFrame({"date": dates,
                   "feature": range(len(dates))})
y = pd.Series(range(len(dates))) * 2 + 4

assert pd.infer_freq(df['date']) is None

arima = AutoARIMA(start_P=1, start_Q=1, max_P=15, max_D=15, max_Q=15, sp=4, stationary=True)
arima.fit(y=y)
arima_preds = arima.predict(fh=[i for i in range(52)])
```

However, if you do `y.index = df['date']` before calling fit on ARIMA, the component will error out.

### Why consider doing this? 
We've noticed that `pd.infer_freq` is finicky and ARIMA may not support all frequencies inferred by `pd.infer_freq` (@ParthivNaresh  has done work on identifying which frequencies those are). By not setting the index manually ourselves, we may be able to avoid a class of bugs related to only a few missing dates (which is primarily why `pd.infer_freq` fails).

Issue #3067, opened 2021-11-16
Arima Errors out on Peyton Manning Dataset during AutoMLSearch


freddyaboulton:
Repro

```python
from evalml.automl import AutoMLSearch
import pandas as pd
from evalml.automl.callbacks import raise_error_callback
import pytest

X = pd.read_csv("/Users/freddy.boulton/Downloads/peyton_manning.csv")
y = X.pop("y")

automl = AutoMLSearch(X, y, "time series regression",
                     problem_configuration={"gap": 0, "max_delay": 14, "forecast_horizon": 14,
                                            "date_index": "ds"},
                      error_callback=raise_error_callback)

with pytest.raises(ValueError, match="Must supply freq for datetime value"):
    automl.search()
```

[peyton_manning.csv.zip](https://github.com/alteryx/evalml/files/7548128/peyton_manning.csv.zip)

Issue #3046, opened 2021-11-15
Refactor ARIMA/Prophet to use only use date_index parameter


freddyaboulton:
Our ARIMA/Prophet components were originally designed to work with an optional `date_index` parameter. Now that the date_index is required in #3041, we should refactor our components in light of this.

For example, we can simplify the logic in `_get_dates`:

```python
    def _get_dates(self, X, y):
        date_col = None
        if y is not None:
            y_index_type = infer_feature_types(
                pd.Series(y.index)
            ).ww.logical_type.type_string
            if y_index_type == "datetime":
                date_col = y.index
        if X is not None:
            X_index_type = infer_feature_types(
                pd.Series(X.index)
            ).ww.logical_type.type_string
            if self.parameters["date_index"] in X.columns:
                date_col = X.pop(self.parameters["date_index"])
            elif X_index_type == "datetime":
                date_col = X.index
        if date_col is None:
            msg = (
                "ARIMA regressor requires input data X to have a datetime column specified by the 'date_index' parameter. "
                "If not it will look for the datetime column in the index of X or y."
            )
            raise ValueError(msg)
        return date_col, X
```

ParthivNaresh:
`date_index` being a required parameter changes how we approach this for AutoML, but for OS users who want to call ARIMA/Prophet directly, we still have to take into consideration that they might choose to pass data that has datetime information in the index instead of as a column.

I'd say these are the two options:
1. Not support datetime information in the index for OS users who want to access the estimators directly.
2. Continue to support users who might pass in datetime information in the index, but make the necessary refactor to support the fact that `date_index` is a mandatory when using AutoML.

freddyaboulton:
Can we require users pass in a non-None `date_index` parameter that corresponds to a column in `X` ? I think it's ok to ask users to change their data to meet our requirements if they're trying to use our components directly, like how prophet requires `ds` and `y` columns.

Issue #2960, opened 2021-10-25
Which features should we pass to ARIMA? 


freddyaboulton:
Currently, if the time series problem is multivariate, ARIMA will receive an exact copy of the input data to to fit on.

Reproducer,

```python
from evalml.automl import AutoMLSearch
from unittest.mock import patch
from evalml.demos import load_weather
import pandas as pd

ARIMA_PATH = "sktime.forecasting.arima.AutoARIMA"

SIZE_OF_FIRST_SPLIT = 914

X, y = load_weather()
X.ww["other_feature"] = y - 1

automl = AutoMLSearch(X, y, "time series regression", verbose=True,
                      problem_configuration={"date_index": "Date",
                                             "forecast_horizon": 1,
                                             "gap": 0,
                                             "max_delay": 10})

with patch(f"{ARIMA_PATH}.fit") as mock_fit, \
    patch(f"{ARIMA_PATH}.predict", side_effect=lambda fh, X: pd.Series(range(X.shape[0]))):
    automl.search()

expected_answer = X.ww.copy()
expected_answer = X.ww.iloc[:SIZE_OF_FIRST_SPLIT, :]
expected_answer.index = pd.DatetimeIndex(expected_answer['Date'], "D")
expected_answer = expected_answer.ww.drop(["Date"])

pd.testing.assert_frame_equal(mock_fit.call_args_list[0].kwargs['X'], expected_answer)
```

I think this is problematic because the `other_feature` in this case may not be known in advanced and this could be a case of data leakage - training a model on data that would not be available at predict time.

The other pipelines receive the input features after feature-engineering, namely lagging and soon-to-be rolling windows, so the `other_feature` would not be given as-is to the pipeline. For that reason, the other pipelines do not suffer from the same problem as ARIMA.

We need to decide:

- Should we even pass features to ARIMA? Why not just pass in the target values
- If we want to pass in some features, we need to make sure there is no leakage.

Step 1 of this issue is reaching consensus on the above decision, then we can decide how to best implement that.

ParthivNaresh:
This was fixed in #3313

PR #3104, opened 2021-11-30
Refactor ARIMA/Prophet

ParthivNaresh:
Fixes: #3091 #3067 #3046 

Perf tests here: https://alteryx.atlassian.net/wiki/spaces/PS/pages/1152516105/ARIMA+-+No+longer+relying+on+date+index

I've also added 4 new datasets to the S3 bucket:
- Christmas_Daily_PageViews
- Summer_Daily_PageViews
- OlympicGames_Daily_PageViews
- Pandemic_Daily_PageViews

I got these by downloading the Wikipedia page views from [here](https://pageviews.toolforge.org/pageviews/?project=en.wikipedia.org&platform=all-access&agent=user&redirects=0&range=latest-20&pages=Cat|Dog) for different articles over various time horizons.

-----


evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -170,14 +144,22 @@ def fit(self, X, y=None):
         Raises:
             ValueError: If X was passed to `fit` but not passed in `predict`.
         """
+        X, y = self._manage_woodwork(X, y)
         if y is None:
             raise ValueError("ARIMA Regressor requires y as input.")
+        if X is not None:
```
ParthivNaresh, line 150: This addresses #2960. The issue is that since `TimeSeriesFeaturizer` isn't included in ARIMA-based pipelines, the potential for target leakage can occur.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -197,23 +179,16 @@ def predict(self, X, y=None):
             ValueError: If X was passed to `fit` but not passed in `predict`.
         """
         X, y = self._manage_woodwork(X, y)
-        dates, X = self._get_dates(X, y)
-        X, y, fh_ = self._format_dates(dates, X, y, predict=True)
-        if X is not None and not X.empty:
-            X = X.select_dtypes(exclude=["datetime64"])
+        fh_ = self._set_forecast(X)
+        X = X[self.cols_to_keep]
+        X = X.select_dtypes(exclude=["datetime64"])
```
ParthivNaresh, line 184: `_remove_datetime` isn't called here because ARIMA doesn't have a problem predicting on data that has an arbitrary temporal index, even if it was fit on data that doesn't have a temporal index (because the forecast horizon is relative).

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -187,21 +186,6 @@ def _create_pipelines(self):
                 self.problem_type, self.allowed_model_families
             )
             allowed_estimators = self._filter_estimators(allowed_estimators)
-            if (
```
ParthivNaresh, line 190: Now that the datetime information is not being set as an index for ARIMA, the time interval of `MS - Month Start` is no longer a concern.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -34,348 +34,247 @@ def test_model_instance(ts_data):
...
+            y_train.index = dates[:40]
+        if test_features_index_dt:
+            X_test.index = dates[40:]
+        if not no_features:
+            X_train["Feature"] = feature[:40]
+            X_test["Feature"] = feature[40:]
+            if datetime_feature:
+                X_train["Dates"] = dates[:40]
+                X_test["Dates"] = dates[40:]
+        if train_none:
```
ParthivNaresh, line 76: The difference between `train_none` and `no_features` is that `train_none` passes in None as X, while `no_features` passes in a dataframe with only an index set.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -170,14 +144,22 @@ def fit(self, X, y=None):
...
             ValueError: If X was passed to `fit` but not passed in `predict`.
         """
+        X, y = self._manage_woodwork(X, y)
         if y is None:
             raise ValueError("ARIMA Regressor requires y as input.")
+        if X is not None:
+            target_leakage = TargetLeakageDataCheck().validate(X, y)["actions"]
+            leaked_columns = (
+                target_leakage[0]["metadata"]["columns"] if target_leakage else []
+            )
```
chukarsten, line 154: These nested list/dict structures :(. I think we might want to think again about the output of `.validate()` to make accessing things like the target leaked columns clearer.  I'll file an issue for this.

evalml/pipelines/components/estimators/regressors/prophet_regressor.py
```diff
@@ -78,30 +79,19 @@ def __init__(
...
     def build_prophet_df(X, y=None, date_column="ds"):
         """Build the Prophet data to pass fit and predict on."""
-        if X is not None:
-            X = copy.deepcopy(X)
-        if y is not None:
-            y = copy.deepcopy(y)
+        X = copy.deepcopy(X)
+        y = copy.deepcopy(y)
+        if date_column is None:
+            raise ValueError("date_index cannot be None!")
```
chukarsten, line 85: Maybe a slight mismatch here between the param name and the str?

evalml/tests/component_tests/test_prophet_regressor.py
```diff
@@ -48,9 +48,11 @@ def test_init_with_other_params():
 
 def test_feature_importance(ts_data):
     X, y = ts_data
-    clf = ProphetRegressor(uncertainty_samples=False, changepoint_prior_scale=2.0)
+    clf = ProphetRegressor(
+        date_index="date", uncertainty_samples=False, changepoint_prior_scale=2.0
+    )
     clf.fit(X, y)
-    clf.feature_importance == np.zeros(1)
+    assert clf.feature_importance == np.zeros(1)
```
chukarsten, line 55: Good catch, lol.

chukarsten (approved):
This looks good to me, Parthiv!  I think just a few minor string related stuff.

chukarsten:
Just for completeness, I filed this : https://github.com/alteryx/evalml/issues/3121

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -37,12 +37,12 @@ def __init__(
...
         pipeline_params = parameters["pipeline"]
-        date_index = pipeline_params["date_index"]
-        if date_index is None:
-            raise ValueError("Parameter date_index cannot be None!")
         self.gap = pipeline_params["gap"]
         self.max_delay = pipeline_params["max_delay"]
         self.forecast_horizon = pipeline_params["forecast_horizon"]
+        self.date_index = pipeline_params["date_index"]
+        if self.date_index is None:
+            raise ValueError("Parameter date_index cannot be None!")
```
eccabay, line 45: Any particular reason for this change?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -170,14 +144,22 @@ def fit(self, X, y=None):
...
+            self.cols_to_keep = list(set(X.columns) - set(leaked_columns))
+            X = X[self.cols_to_keep]
```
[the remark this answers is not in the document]
ParthivNaresh: So the previous implementation dropped `datetime` columns because ARIMA doesn't accept those, so this is an extension of that. I totally understand the trepidation over this, but since this isn't a `Transformer` we luckily won't be passing the features onto anything else.

That being said, I do think that maybe a warning should be added to the logs, is that along the lines of what you were thinking?

evalml/pipelines/time_series_pipeline_base.py
ParthivNaresh: Since `date_index` is a required parameter in time series configurations, I figured it would be good to have it as an instance variable alongside `gap`, `max_delay`, and `forecast_horizon`. And the position change just because I felt it's cleaner to have the instance variables together

evalml/pipelines/components/estimators/regressors/arima_regressor.py
freddyaboulton: I think we should back this out of this PR and tackle it in it's own issue because I don't think this is the best fix for the problem.

I think the root cause of the problem is including features that will not be known during prediction time during fit. And whether or not a feature is known-in-advanced is independent of it's correlation to the target (which is what TargetLeakeage is checking). It's possible that there is a feature that will not be known in advanced that is not picked up by TargetLeakage and it's also possible that a feature that will be known in advanced will be dropped because it's correlated with the target. 

I think we gotta think this through a bit more it's possible #2960 is blocked by #2511

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -34,348 +34,247 @@ def test_model_instance(ts_data):
...
+        test_features_index_dt=False,
     )
 
-    a_clf = sktime_arima.AutoARIMA()
-    clf = a_clf.fit(X=X, y=y)
-    y_pred_sk = clf.predict(fh=fh_, X=X_test)
+    clf = ARIMARegressor()
+    fh_ = clf._set_forecast(X_test)
+    assert isinstance(fh_, forecasting.ForecastingHorizon)
+    assert len(fh_) == len(X_test)
```
freddyaboulton, line 170: nit: check that `is_relative` is true?

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -34,348 +34,247 @@ def test_model_instance(ts_data):
...
+        (False, True, False, False, False, True),
+        (False, False, True, False, True, False),
+        (False, False, True, True, False, False),
+        (False, True, True, True, False, False),
+        (False, True, True, False, True, False),
+    ],
 )
-def test_fit_predict_ts_with_y_not_X_index(
-    mock_get_dates, mock_format_dates, ts_data_seasonal_train
+def test_fit_predict_sk_failure(
```
freddyaboulton, line 238: nit: This is basically testing that without `remove_datetime`, our ARIMA component would fail right? Can we rename the test to reflect that and add a comment to the `pytest.raises` to explain why/what is happening?

freddyaboulton (approved):
@ParthivNaresh Thanks for this PR I think it looks great! I think it's awesome you added some new seasonal datasets from wikipedia for us to test on. Looking forward to the LG PR to add those so I can use them too. Can you upload the html report to your results page? I'm curious to hover over those datasets/look at the rankings table. 

The one thing is that I think we should back out the changes related to #2960 and tackle that problem in its own PR.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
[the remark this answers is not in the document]
eccabay: I would support adding something to logs! Even if it's not at the `WARNING` level, just a debug message could be better than nothing.

evalml/tests/component_tests/test_arima_regressor.py
ParthivNaresh: Actually this is testing a few different use cases in which the original `sktime.AutoARIMA` fails but our implementation succeeds.
- When the X index is of datetime type and the y index is not (and vice versa)
- When X has no features but still has an index
- When X has a datetime feature

So I named it `test_fit_predict_sk_failure` to represent the difference in behaviour from `sktime` and `evalml`

Issue #2919, opened 2021-10-14
Add `DFSTransformer` as part of AutoML when `features` are accepted as input


jeremyliweishih:
Add the DFSTransformer component to the beginning of EvalML pipelines by adding features as input for  AutoMLSearch() and will then be added to pipeline creation via automl algorithms

```
class AutoMLSearch:
    def __init__(
        self,
        X_train=None,
        y_train=None,
        ...,
        features=None
    )
    
    self.features = features
    if _automl_algorithm == "iterative":
            self._automl_algorithm = IterativeAlgorithm(
                max_iterations=self.max_iterations,
                allowed_pipelines=self.allowed_pipelines,
                tuner_class=self.tuner_class,
                random_seed=self.random_seed,
                n_jobs=self.n_jobs,
                number_features=self.X_train.shape[1],
                pipelines_per_batch=self._pipelines_per_batch,
                ensembling=run_ensembling,
                text_in_ensembling=text_in_ensembling,
                pipeline_params=parameters,
                custom_hyperparameters=custom_hyperparameters,
                features=self.features,
            )
        elif _automl_algorithm == "default":
            self._automl_algorithm = DefaultAlgorithm(
                X=self.X_train,
                y=self.y_train,
                problem_type=self.problem_type,
                sampler_name=self._sampler_name,
                tuner_class=self.tuner_class,
                random_seed=self.random_seed,
                pipeline_params=parameters,
                custom_hyperparameters=self.custom_hyperparameters,
                text_in_ensembling=text_in_ensembling,
                features=self.features
            )
```

chukarsten:
@jeremyliweishih since you're the epic owner, feel free to assign sprints to these as makes sense to you.

PR #3106, opened 2021-11-30
Accept serialized features into `DFSTransformer`

### LINKED ISSUE #2917: Accept `features` for `DFSTransformer` and add logic to skip computation if features are already present
Add features as input to DFSTransformer that takes in the output of featuretools.save_features() and follows the steps outlined [here](https://featuretools.alteryx.com/en/stable/guides/deployment.html) to load in new features and use these features to calculate a feature matrix for new data. Also add logic to determine if incoming data to DFSTransformer contains features that have already been computed by the features in DFSTransformer and skip fit and transform when being passed such data.

```
import featuretools as ft

class DFSTransformer(Transformer):
    """Featuretools DFS component that generates features for the input features.

    Arguments:
        index (string): The name of the column that contains the indices. If no column with this name exists,
            then featuretools.EntitySet() creates a column with this name to serve as the index column. Defaults to 'index'.
        random_seed (int): Seed for the random number generator. Defaults to 0.
    """

    name = "DFS Transformer"
    hyperparameter_ranges = {}
    """{}"""

    def __init__(self, index="index", random_seed=0, features=None, **kwargs):
        parameters = {"index": index}
        if not isinstance(index, str):
            raise TypeError(f"Index provided must be string, got {type(index)}")

        self.index = index
        self.features = None
        parameters.update(kwargs)
        super().__init__(parameters=parameters, random_seed=random_seed)
    
    def _should_skip_fit_transform(self, X):
      # check if column names matches Feature.get_feature_names
      # https://github.com/alteryx/featuretools/issues/1696 could make this logic easier
      for feature in self.features:
        feature_names = Feature.get_feature_names()
        if feature_names is not a subset of X.column_names:
          return False
      return True
    
    def fit(self, X, y=None):
        """Fits the DFSTransformer Transformer component.

        Arguments:
            X (pd.DataFrame, np.array): The input data to transform, of shape [n_samples, n_features]
            y (pd.Series, np.ndarray, optional): The target training data of length [n_samples]

        Returns:
            self
        """
        if not self.features:
          X_ww = infer_feature_types(X)
          X_ww = X_ww.ww.rename({col: str(col) for col in X_ww.columns})
          es = self._make_entity_set(X_ww)
          self.features = dfs(
              entityset=es, target_entity="X", features_only=True, max_depth=1
          )
        return self   
```

jeremyliweishih:
Fixes #2917. This PR allows `DFSTransformer` to accept features and skips computation when needed. This should work in tandem with #2919.


evalml/tests/component_tests/test_featuretools.py
```diff
@@ -56,6 +56,7 @@ def test_featuretools_index(mock_calculate_feature_matrix, mock_dfs, X_y_multi):
     assert arg_tr.to_list() == new_index
 
     # check if _make_entity_set fills in the proper index values
+    feature = DFSTransformer()
```
jeremyliweishih, line 59: These changes make it so that you cannot take one `DFSTransformer` and fit (generate new features) on new data. I don't believe that we need this use case for the `DFSTransformer` but I'm happy to discuss!

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +164,55 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
     feature.fit(X_pd, y)
     _, kwargs = mock_dfs.call_args
     assert kwargs["max_depth"] == 1
+
+
+@patch("evalml.pipelines.components.transformers.preprocessing.featuretools.dfs")
+def test_dfs_with_serialized_features(mock_dfs, X_y_binary):
```
jeremyliweishih, line 170: this test checks that DFS is not called when features are serialized and that the feature matrix is computed correctly using serialized features.

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +164,55 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
...
+    X_t = feature.transform(X_transform)
+    assert_frame_equal(feature_matrix, X_t)
+    assert features == feature.features
+
+
+@patch("evalml.pipelines.components.transformers.preprocessing.featuretools.dfs")
+@patch(
+    "evalml.pipelines.components.transformers.preprocessing.featuretools.calculate_feature_matrix"
+)
+def test_dfs_skip_transform(mock_calculate_feature_matrix, mock_dfs, X_y_binary):
```
jeremyliweishih, line 198: Give serialized features and input that contains the transformed columns, this test checks that both DFS and calculate feature matrix is not called.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -52,12 +61,13 @@ def fit(self, X, y=None):
         Returns:
             self
         """
-        X_ww = infer_feature_types(X)
-        X_ww = X_ww.ww.rename({col: str(col) for col in X_ww.columns})
-        es = self._make_entity_set(X_ww)
-        self.features = dfs(
-            entityset=es, target_dataframe_name="X", features_only=True, max_depth=1
-        )
+        if not self.features:
```
bchen1116, line 64: nit, but why not use `if not self._passed_in_features`?

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -42,6 +44,13 @@ def _make_entity_set(self, X):
             es = ft_es.add_dataframe(dataframe=X, dataframe_name="X", index=self.index)
         return es
 
+    def _should_skip_transform(self, X):
+        for feature in self.features:
+            feature_names = feature.get_feature_names()
+            if not set(feature_names).issubset(set(X.columns)):
```
bchen1116, line 50: Should we be giving the user a warning with this? What are scenarios where this would be false?

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +164,55 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
...
+    feature_matrix, features = ft.dfs(entityset=es, target_dataframe_name="X")
+
+    feature = DFSTransformer(features=features)
+    feature.fit(X_fit)  # no-op
+    X_t = feature.transform(feature_matrix)  # no-op as well
+    assert not mock_dfs.called
+    assert not mock_calculate_feature_matrix.called
+
+    assert_frame_equal(feature_matrix, X_t)
+    assert features == feature.features
```
bchen1116, line 218: Can we add a test for:
- Passing in features such that feature.get_feature_names() doesn't exist in the dataframe

bchen1116 (changes requested):
Looking good! I think it would be nice to add a test for features where the associated feature names aren't in the dataframe. I also left a question for my own understanding.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
freddyaboulton: Should this be `set(feature_names).intersection(set(X.columns))` ? Is there a case where users dropped some  features in between running dfs and modeling?

freddyaboulton: Cause right now we skip only if all the `feature_names` are in `X`. But if 3 out of the 4 features are present, we'll get a duplicate column name error right?

jeremyliweishih: @freddyaboulton thanks for the example, I'll rework the logic here to take that case into account.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
jeremyliweishih: can change! its because I wrote that logic before I added `self._passed_infeatures` 😄

chukarsten (approved):

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -12,19 +12,21 @@ class DFSTransformer(Transformer):
         index (string): The name of the column that contains the indices. If no column with this name exists,
             then featuretools.EntitySet() creates a column with this name to serve as the index column. Defaults to 'index'.
         random_seed (int): Seed for the random number generator. Defaults to 0.
+        features (list)[FeatureBase]: list of features to run DFS on.
```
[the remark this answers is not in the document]
jeremyliweishih: @chukarsten thats what I thought but it didn't catch it here.

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +164,55 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
     feature.fit(X_pd, y)
     _, kwargs = mock_dfs.call_args
     assert kwargs["max_depth"] == 1
+
+
+@patch("evalml.pipelines.components.transformers.preprocessing.featuretools.dfs")
+def test_dfs_with_serialized_features(mock_dfs, X_y_binary):
+    X, y = X_y_binary
+    X_pd = pd.DataFrame(X)
+    X_pd.columns = X_pd.columns.astype(str)
```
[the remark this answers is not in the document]
jeremyliweishih: Featuretools requires column names to be a string:

```
# Only allow string column names
        non_string_names = [name for name in dataframe.columns if not isinstance(name, str)]
        if non_string_names:
>           raise ValueError("All column names must be strings (Columns {} "
                             "are not strings)".format(non_string_names))
```

jeremyliweishih:
@freddyaboulton @chukarsten @dsherry re our discussion:

Something I didn’t realize was that `calculate_feature_matrix` would error out if the original columns used by the features didn’t exist. In the case of this notebook,  I cannot recalculate a feature matrix unless  “Date” exists (which makes sense since its needs a “Date” column to calculate year etc.). In this case I can check two things:
1. Check if the columns used by features exist
2. If they don’t, dont use these features in DFSTransformer
3. If they do, check if feature names exists and if it does exist, skip these features as well
4. Lastly, if the feature names don’t exist (and original columns exist), use these features for transform

I’ll attach the notebook here as well.

[skip_computation_experiments.ipynb.zip](https://github.com/alteryx/evalml/files/7652162/skip_computation_experiments.ipynb.zip)

Some other use cases that @thehomebrewnerd brought up to me:
- Some primitives use multiple columns as inputs, so for features generated from those primitives you will need to confirm all the input columns are present.
- Things will get tricky in a multi-table setup when aggregations are present because you would need to start checking for columns in multiple dataframes - but maybe that is out of scope here for now.

I will definitely address the first point in unit testing as well as file an issue tracking multi-table changes when we want to support multi-table as well. 

Let me know what you guys think!

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -42,6 +45,18 @@ def _make_entity_set(self, X):
             es = ft_es.add_dataframe(dataframe=X, dataframe_name="X", index=self.index)
         return es
 
+    def _filter_features(self, X, es):
+        features_to_use = []
+        for feature in self.features:
+            if not feature.entityset == es:
```
jeremyliweishih, line 51: @thehomebrewnerd here I'm using comparison of the feature's ES and the ES created using the input to judge if the columns needed by the feature exists. Is there a more direct method of looking at what columns a feature needs to simply this?

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -42,6 +45,18 @@ def _make_entity_set(self, X):
...
+                continue
+            if not isinstance(feature, IdentityFeature) and set(
```
jeremyliweishih, line 53: `not isinstance(feature, IdentityFeature)` is needed because by definition, if an `IdentityFeature` is passed in it will be part of `X`.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
thehomebrewnerd: @jeremyliweishih If you have a feature, you can access the other features it was built from by accessing `feature.base_features`. This will be a list - in a single table setup, I think this list will contain only `IdentifyFeatures`. From those base identify features you can access the dataframe name with `feature.dataframe_name`(if you need it) and the column name with `feature.column_name`.

So, putting this all together, to get a list of the columns used to create a feature, `f`,  in a single table entityset, you can do this:
```
inputs_cols = [f.column_name for f in f.base_features]
```

Things get a fair amount more complicated when you get to multi-table entitysets with an arbitrary feature depth, but that might be out of scope for what you are doing now.

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +165,86 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
...
+    )  # no-op as well, feature_matrix contains features already
+    assert not mock_dfs.called
+    assert not mock_calculate_feature_matrix.called
+
+    assert_frame_equal(feature_matrix, X_t)
+    assert features == dfs.features
+
+
+@patch("evalml.pipelines.components.transformers.preprocessing.featuretools.dfs")
+def test_dfs_does_not_skip_transform_with_non_identity_feature(mock_dfs, X_y_binary):
```
jeremyliweishih, line 225: Added an additional test since previous tests only had identity features.

evalml/tests/component_tests/test_featuretools.py
freddyaboulton: Can you explain? Without this line the test still passes

evalml/tests/component_tests/test_featuretools.py
freddyaboulton: I think this is a typo. Both tests use:

```python
    feature_matrix, features = ft.dfs(
        entityset=es, target_dataframe_name="X", trans_primitives=["absolute"]
    )
    features = list(filter(lambda f: not isinstance(f, IdentityFeature), features))
```

And both create these features:

```python
['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12',
       '13', '14', '15', '16', '17', '18', '19', 'ABSOLUTE(0)', 'ABSOLUTE(1)',
       'ABSOLUTE(10)', 'ABSOLUTE(11)', 'ABSOLUTE(12)', 'ABSOLUTE(13)',
       'ABSOLUTE(14)', 'ABSOLUTE(15)', 'ABSOLUTE(16)', 'ABSOLUTE(17)',
       'ABSOLUTE(18)', 'ABSOLUTE(19)', 'ABSOLUTE(2)', 'ABSOLUTE(3)',
       'ABSOLUTE(4)', 'ABSOLUTE(5)', 'ABSOLUTE(6)', 'ABSOLUTE(7)',
       'ABSOLUTE(8)', 'ABSOLUTE(9)']
```

evalml/tests/component_tests/test_featuretools.py
jeremyliweishih: @freddyaboulton with the new logic this isn't necessary anymore, will fix!

evalml/tests/component_tests/test_featuretools.py
freddyaboulton: I'm confused by this 😬  Both this test and `test_dfs_skip_transform` use :

```python
    feature_matrix, features = ft.dfs(
        entityset=es, target_dataframe_name="X", trans_primitives=["absolute"]
    )
    features = list(filter(lambda f: not isinstance(f, IdentityFeature), features))
```

Why does this test use `X_transform` to create the entity set as opposed to `X_pd` ?

freddyaboulton: @jeremyliweishih and I spoke about this. My confusion was because I thought `non_identity_feature` referred to the `features` passed in init, but it refers to the features passed to `transform`. In this test, only the original features are passed to `transform` (X_pd) but in the previous case, the original + absolute features are passed to `transform`.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -42,6 +45,24 @@ def _make_entity_set(self, X):
             es = ft_es.add_dataframe(dataframe=X, dataframe_name="X", index=self.index)
         return es
 
+    def _filter_features(self, X):
+        features_to_use = []
+        for feature in self.features:
+            input_cols = [f.column_name for f in feature.base_features]
+            if (
```
freddyaboulton, line 52: Should this be `and feature.columns_name in X.columns`? I think this is causing a bug where only the identify features are returned if you pass in data with identity and non-identity features?

```python
import featuretools as ft
from evalml.pipelines.components import DFSTransformer
import pandas as pd
import pytest


X, y = load_diabetes()
del X.ww

X_fit = X.iloc[:X.shape[0]//2]
X_transform = X.iloc[X.shape[0]//2:]


es = ft.EntitySet()
es = es.add_dataframe(
    dataframe_name="X", dataframe=X_fit, index="index", make_index=True
)
feature_matrix, features = ft.dfs(
    entityset=es, target_dataframe_name="X", trans_primitives=["absolute"]
)
#features = list(filter(lambda f: not isinstance(f, IdentityFeature), features))

dfs = DFSTransformer(features=features)
dfs.fit(X_fit)
X_t = dfs.transform(feature_matrix)

with pytest.raises(AssertionError):
    pd.testing.assert_frame_equal(X_t, feature_matrix)
```

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -163,3 +164,140 @@ def test_dfs_sets_max_depth_1(mock_dfs, X_y_multi):
...
+def test_dfs_with_serialized_features(mock_dfs, X_y_binary):
+    X, y = X_y_binary
+    X_pd = pd.DataFrame(X)
+    X_pd.columns = X_pd.columns.astype(str)
+
+    es = ft.EntitySet()
+    es = es.add_dataframe(
+        dataframe_name="X", dataframe=X_pd, index="index", make_index=True
+    )
+    feature_matrix, features = ft.dfs(
```
freddyaboulton, line 179: I'm not seeing the same behavior with the `year` primitive on the fraud dataset:

```python
import featuretools as ft
from evalml.pipelines.components import DFSTransformer
from featuretools.feature_base import IdentityFeature
import pandas as pd
import pytest


X, y = load_fraud(1000)
del X.ww

X_fit = X.iloc[:X.shape[0]//2]
X_transform = X.iloc[X.shape[0]//2:]


es = ft.EntitySet()
es = es.add_dataframe(
    dataframe_name="X", dataframe=X_fit, index="index", make_index=True
)
feature_matrix, features = ft.dfs(
    entityset=es, target_dataframe_name="X", trans_primitives=["year"]
)
features = list(filter(lambda f: not isinstance(f, IdentityFeature), features))

dfs = DFSTransformer(features=features)
dfs.fit(X_fit)
X_t = dfs.transform(X_transform)

with pytest.raises(AssertionError):
    pd.testing.assert_frame_equal(X_t, feature_matrix)
    
print(X_t.columns)
```

I would expect the same behavior in this test for this repro: X_t has all the original features plus YEAR (and is therefore equal to feature_matrix).

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -42,6 +45,24 @@ def _make_entity_set(self, X):
...
+                isinstance(feature, IdentityFeature)
+                and feature.column_name not in X.columns
+            ):
+                continue
+            if not set(input_cols).issubset(set(X.columns)):
```
thehomebrewnerd, line 57: Minor point, but you are converting `X.columns` to a set in multiple places in this loop each pass through. You could do the set conversion once outside of the `for` loop, which should make this faster.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
thehomebrewnerd, line 52: Might be helpful to add a brief comments about what exactly you are checking in each of the `if` conditions so people revisiting this in the future will be able to follow the logic a little easier.

For example: `# If input columns are not present in X, skip feature`

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -72,9 +94,18 @@ def transform(self, X, y=None):
         """
         X_ww = infer_feature_types(X)
         X_ww = X_ww.ww.rename({col: str(col) for col in X_ww.columns})
-        es = self._make_entity_set(X_ww)
 
-        feature_matrix = calculate_feature_matrix(features=self.features, entityset=es)
+        features_to_use = (
```
thehomebrewnerd, line 98: Not sure about this, but would it make sense to do the filtering during `fit` instead of here, so that  after calling `fit`, `self.features` would always contain the features that were actually used by the transformer? You could change `self._passed_in_features` to retain the original list of features that were provided if you need to hold on to those, and if the list is present or contains values you know the user passed in features, otherwise you know they were generated from DFS.

evalml/tests/component_tests/test_featuretools.py
freddyaboulton: This was my mistake heh. We need to pass the identity features in this case for X_t to match feature_matrix.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
jeremyliweishih: @thehomebrewnerd thought about this a little more and decided to keep the filtering on the transform end. Main reasoning being that ultimately we need to look at the columns that are passed into `transform` to decide on which features to use. I could add logic verifying that the columns between the dataframe for `fit` and `transform` are the same but I'd rather keep it simple and keep `fit` a no-op and do the filtering in `transform`. Let me know if that makes sense or if you other suggestions!

thehomebrewnerd: Ah, I guess I wasn't thinking about the set of columns used for `fit` being different from the set of columns passed to `transform`.  I don't have any strong feelings one way or the other here, so I'm fine if you keep the filtering in `transform`.

thehomebrewnerd (commented):
This looks pretty good to me, but you might want to consider adding a test case for a multi-input transform primitive (such `divide_numeric`) - both to confirm that the feature is calculated when both inputs are present and to confirm that no errors occur when one of the two input features is missing.

freddyaboulton (commented):
Looks good to me @jeremyliweishih ! Thank you for making the changes.

PR #3110, opened 2021-12-01
Change default parameters for feature selectors

jeremyliweishih:
This PR proposes changing the default parameters for `RFRegressorSelectFromModel` and `RFClassifierSelectFromModel`. The current, incorrect, behavior of these components is as follows:

- `number_features=None`, `percent_features=0.5`, and `threshold=-np.inf`
- max features is then calculated as: 
```
max_features = (
            max(1, int(percent_features * number_features)) if number_features else None
        )
```
- therefore, `max_features = None`
- with `max_features == None and `threshold=-np.inf`, the component will select every feature with importance above `-np.inf` which is every feature available.

Performance tests using default algorithm:
[fs_parameters_tests.zip](https://github.com/alteryx/evalml/files/7637092/fs_parameters_tests.zip)

### DIFF
--- a/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
+++ b/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
@@ -28,11 +28,11 @@ class RFClassifierSelectFromModel(FeatureSelector):
     name = "RF Classifier Select From Model"
     hyperparameter_ranges = {
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean"],
+        "threshold": ["mean", "median"],
     }
     """{
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean"],
+        "threshold": ["mean", "median"],
     }"""
 
     def __init__(
@@ -41,7 +41,7 @@ class RFClassifierSelectFromModel(FeatureSelector):
         n_estimators=10,
         max_depth=None,
         percent_features=0.5,
-        threshold="mean",
+        threshold="median",
         n_jobs=-1,
         random_seed=0,
         **kwargs,
--- a/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
+++ b/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
@@ -29,11 +29,11 @@ class RFClassifierSelectFromModel(FeatureSelector):
     name = "RF Classifier Select From Model"
     hyperparameter_ranges = {
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean", -np.inf],
+        "threshold": ["mean"],
     }
     """{
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean", -np.inf],
+        "threshold": ["mean"],
     }"""
 
     def __init__(
@@ -42,7 +42,7 @@ class RFClassifierSelectFromModel(FeatureSelector):
         n_estimators=10,
         max_depth=None,
         percent_features=0.5,
-        threshold=-np.inf,
+        threshold="mean",
         n_jobs=-1,
         random_seed=0,
         **kwargs,
--- a/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
+++ b/evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
@@ -1,5 +1,4 @@
 """Component that selects top features based on importance weights using a Random Forest classifier."""
-import numpy as np
 from sklearn.ensemble import RandomForestClassifier as SKRandomForestClassifier
 from sklearn.feature_selection import SelectFromModel as SkSelect
 from skopt.space import Real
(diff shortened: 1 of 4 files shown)


evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
```diff
@@ -29,11 +29,11 @@ class RFClassifierSelectFromModel(FeatureSelector):
     name = "RF Classifier Select From Model"
     hyperparameter_ranges = {
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean", -np.inf],
+        "threshold": ["mean"],
```
jeremyliweishih, line 32: This PR proposes using the `mean` as the threshold but experiments show that `median` performs similarly as well. `median` will select exactly half of the features and `mean` depending on the distribution. Happy to discuss which one to choose but I chose `mean` in the... meantime..

jeremyliweishih: On second thought: using `median` might be the move as the `mean` of feature importances will bound to be dragged down by low signal features (which are inevitable). However, performance results show similar model quality between `median` and `mean` but `median` having slightly longer fit times due to `median` selecting more features.

bchen1116: Why not allow both 'mean' and 'median' to be in the hyperparameter ranges? We can default to one, but allowing both in the ranges seems to be ideal for our automlsearch algo

evalml/pipelines/components/transformers/feature_selection/rf_regressor_feature_selector.py
```diff
@@ -29,11 +28,11 @@ class RFRegressorSelectFromModel(FeatureSelector):
     name = "RF Regressor Select From Model"
     hyperparameter_ranges = {
         "percent_features": Real(0.01, 1),
-        "threshold": ["mean", -np.inf],
+        "threshold": ["mean"],
```
bchen1116, line 31: Also here, seems to be ideal to allow `median` as a possible value in the hyperparams.

evalml/tests/component_tests/test_components.py
```diff
@@ -888,6 +888,10 @@ def test_transformer_transform_output_type(X_y_binary):
                 component, SelectByType
             ):
                 assert transform_output.shape == (X.shape[0], 0)
+            elif isinstance(component, RFRegressorSelectFromModel):
+                assert transform_output.shape == (X.shape[0], 2)
+            elif isinstance(component, RFClassifierSelectFromModel):
+                assert transform_output.shape == (X.shape[0], 5)
```
bchen1116, line 894: What's the `2` or `5` values coming from here?

bchen1116 (commented):
Looks great! So this change will then allow the `RF...FeatureSelector` to now choose the top features that fall above the mean importance weight?

Looking at the perf tests, the fit times and performance changes seem reasonable. In all three tests, though, LightGBM seems to drop very significantly in performance:
image

Do you know why this is? This seems to be something that we should figure out and resolve before moving ahead with the change, especially if it's a potential bug somewhere in the code.

evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
freddyaboulton: I like that suggestion. Right now the DefaultAlgorithm only runs one feature selector, so I'm down to set median as the default. Yesterday we discussed letting the algorithm tune the parameters of the selector in which case broadening the threshold search space (like parametrizing it as a quantile of the observed feature importance distribution) will be in play.

freddyaboulton (approved):
Looks good to me @jeremyliweishih ! Agree with @bchen1116 that setting median and mean as hyperparameter values makes sense.

evalml/pipelines/components/transformers/feature_selection/rf_classifier_feature_selector.py
jeremyliweishih: @bchen1116 @freddyaboulton 

sounds good, I'll add them both as hyperparameter ranges! My main concern is what Freddy brought up about the default parameter and having only 1 FS batch but I'm still on the fence on chosing `mean` or `median`. Either way it'll be a quick fix so I'm not too worried!

evalml/tests/component_tests/test_components.py
jeremyliweishih: it's the number of columns selected by the FS component using the default parameters on `X_y_binary` and `X_y_regression`.

jeremyliweishih:
@bchen1116 

LightGBM drops significantly in terms of percentage change (since the log loss is < 0.1) but it's not that big in absolute terms and likewise with the best pipeline validation score for that dataset so I'm not too worried about it. I guess the columns selected by the FS doesn't play nicely with LightGBM but I don't have enough knowledge about how LightGBM works to make any concrete statements about the change.

bchen1116 (approved):
Changes look good to me!

PR #3122, opened 2021-12-03
Clean up and refactor ``InvalidTargetDataCheck``

### LINKED ISSUE #1741: InvalidTargetDataCheck Cleanup
EvalML/evalml/data_checks/invalid_data_targets_data_check.py needs the following updates:

1. Refactor either the source python file to match the class name to be consistent or refactor the class name to match the source file.
2. Update the doc strings to reflect that this class validation does more than just check for highly null target series.
3. Perhaps either add more inline comments, add private functions to modularize the individual checks that are performed, or separate this into individual data checks.

angela97lin:
Closes #1741

This PR adds a more in-depth docstring detailing what `InvalidTargetDataCheck` checks, and breaks up each of the different checks into their own helper methods.


evalml/data_checks/invalid_target_data_check.py
```diff
@@ -44,10 +52,18 @@ def __init__(self, problem_type, objective, n_unique=100):
...
+            - Target is None.
+            - Target has NaN or None values.
+            - Target is of an unsupported Woodwork logical type.
+            - Target and features have different lengths or indices.
+            - Target does not have enough instances of a class in a classification problem.
+            - Target does not contain numeric data for regression problems.
 
         Args:
-            X (pd.DataFrame, np.ndarray): Features. Ignored.
+            X (pd.DataFrame, np.ndarray): Features. If not None, will be used to check that the target and features have the same dimensions and indices.
```
chukarsten, line 66: Maybe it's redundant, but perhaps some mention of this might also be in the main docstring?  This is a weird data check as it deals mostly with the target but its first arg is X, which is ignored, lol.  Sometimes.

freddyaboulton (approved):
Thanks @angela97lin ! Looks good to me!

evalml/data_checks/invalid_target_data_check.py
```diff
@@ -209,7 +246,99 @@ def validate(self, X, y):
...
+                            message="Input target and features have mismatched indices order",
+                            data_check_name=self.name,
+                            message_code=DataCheckMessageCode.MISMATCHED_INDICES_ORDER,
+                            details={},
+                        ),
+                        results,
+                    )
+                else:
+                    index_diff_not_in_X = list(set(y_index) - set(X_index))[:10]
+                    index_diff_not_in_y = list(set(X_index) - set(y_index))[:10]
```
chukarsten, line 285: I like that you preempt the huge warning messages.  Are we able to have the message somehow communicate that these are only up to 10 of the mismatched indices?

chukarsten (approved):
Looks good to me.  Left some minor comments but feel free to ignore.

evalml/data_checks/__init__.py
```diff
@@ -7,7 +7,7 @@
 from .data_check_message import DataCheckMessage, DataCheckWarning, DataCheckError
 from .data_check_message_type import DataCheckMessageType
 from .default_data_checks import DefaultDataChecks
-from .invalid_targets_data_check import InvalidTargetDataCheck
+from .invalid_target_data_check import InvalidTargetDataCheck
```
eccabay, line 10: I'm not sure, but would this count as a breaking change?

evalml/data_checks/invalid_target_data_check.py
angela97lin: I'll add it to the message! :)

evalml/data_checks/__init__.py
angela97lin: Oo good call! I'd say it's safer to add it as a breaking change since we're changing the module name. Adding to release notes, thanks!

PR #3123, opened 2021-12-03
Add ability to only impute specific columns

### LINKED ISSUE #3039: Add ability to only impute specific columns
Right now, our [PerColumnImputer](https://evalml.alteryx.com/en/stable/autoapi/evalml/pipelines/components/transformers/imputers/index.html?highlight=targetImputer#evalml.pipelines.components.transformers.imputers.PerColumnImputer) allows us to change the impute strategy for each specific column in the input features. However, it still imputes the entire dataframe! It just gives us control over what imputation strategy we want to use for particular columns.

If we want to enable null column imputation, this "works" in that it imputes the columns with null columns, and other columns should be left untouched. But if a user specifies that they *don't* want a column to be imputed, this isn't enough!

We could either create a new component or modify the PerColumnImputer. I suggest adding a new parameter such as `impute_other` that specifies whether or not the columns not specified with impute strategies should be imputed as well. If `impute_other` is True, we have our current behavior. If `impute_other` is False, we only impute the columns whose impute strategies are specified.

angela97lin:
Closes #3039

### DIFF
--- a/evalml/pipelines/components/transformers/imputers/per_column_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/per_column_imputer.py
@@ -18,6 +18,10 @@ class PerColumnImputer(Transformer):
         default_impute_strategy (str): Impute strategy to fall back on when none is provided for a certain column.
             Valid values include "mean", "median", "most_frequent", "constant" for numerical data,
             and "most_frequent", "constant" for object data types. Defaults to "most_frequent".
+        impute_all (bool): Whether or not to impute all columns or just the columns that are specified in `impute_strategies`. If True,
+            all columns will be imputed either using the strategy specified in `impute_strategies` or using the `default_impute_strategy`.
+            If False, only columns specified as keys in `default_impute_strategy` are imputed. If False and `impute_strategies` is None,
+            no columns will be imputed. Defaults to True.
         random_seed (int): Seed for the random number generator. Defaults to 0.
     """
 
@@ -29,22 +33,23 @@ class PerColumnImputer(Transformer):
         self,
         impute_strategies=None,
         default_impute_strategy="most_frequent",
+        impute_all=True,
         random_seed=0,
         **kwargs,
     ):
         parameters = {
             "impute_strategies": impute_strategies,
             "default_impute_strategy": default_impute_strategy,
+            "impute_all": impute_all
         }
         self.imputers = None
-        self.default_impute_strategy = default_impute_strategy
         self.impute_strategies = impute_strategies or dict()
-
         if not isinstance(self.impute_strategies, dict):
             raise ValueError(
                 "`impute_strategies` is not a dictionary. Please provide in Column and {`impute_strategy`: strategy, `fill_value`:value} pairs. "
             )
-
+        self.default_impute_strategy = default_impute_strategy
+        self.impute_all = impute_all
         super().__init__(
             parameters=parameters, component_obj=None, random_seed=random_seed
         )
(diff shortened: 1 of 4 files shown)


jeremyliweishih (approved):
looks great!

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
```diff
@@ -61,7 +66,9 @@ def fit(self, X, y=None):
         """
         X = infer_feature_types(X)
         self.imputers = dict()
-        for column in X.columns:
+
+        columns_to_impute = X.columns if self.impute_all else self.impute_strategies
```
eccabay, line 70: Does this work correctly even though `X.columns` is a list and `self.impute_strategies` is a dict? Why not use `self.impute_strategies.keys()`?

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
```diff
@@ -18,6 +18,10 @@ class PerColumnImputer(Transformer):
         default_impute_strategy (str): Impute strategy to fall back on when none is provided for a certain column.
             Valid values include "mean", "median", "most_frequent", "constant" for numerical data,
             and "most_frequent", "constant" for object data types. Defaults to "most_frequent".
+        impute_all (bool): Whether or not to impute all columns or just the columns that are specified in `impute_strategies`. If True,
+            all columns will be imputed either using the strategy specified in `impute_strategies` or using the `default_impute_strategy`.
+            If False, only columns specified as keys in the `impute_strategies` dictionary are imputed. If False and `impute_strategies` is None,
+            no columns will be imputed. Defaults to True.
```
eccabay, line 24: Would it be worth it to warn the user in the case that the imputer does not impute anything?

bchen1116 (approved):
LGTM!

eccabay (approved):
Love it, such a simple yet elegant change! Just left a few small questions.

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
```diff
@@ -18,6 +18,10 @@ class PerColumnImputer(Transformer):
         default_impute_strategy (str): Impute strategy to fall back on when none is provided for a certain column.
             Valid values include "mean", "median", "most_frequent", "constant" for numerical data,
             and "most_frequent", "constant" for object data types. Defaults to "most_frequent".
+        impute_all (bool): Whether or not to impute all columns or just the columns that are specified in `impute_strategies`. If True,
+            all columns will be imputed either using the strategy specified in `impute_strategies` or using the `default_impute_strategy`.
```
chukarsten, line 22: I am not sure what you mean by "If True, all columns will be imputed either using the strategy specified in `impute_strategies`".  Aren't the `impute_strategies` defined on a per-column basis?

angela97lin: @chukarsten Yup, `impute_strategies` is a dictionary that looks like `{ "col_name": {"impute_strategy": "mean", "fill_value": 0}, "col_name2": {"impute_strategy": "mean", "fill_value": 0} ... }`, but if they're not specified in the dictionary then we use the `default_impute_strategy` for that column. Does that make sense / do you think this could be clarified more?

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
angela97lin: Heh, I think here it works because we just use it to iterate, and iterating over a dictionary will use its keys! But happy to update this to be more clear / in case we update this logic :)

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
angela97lin: Updated the wording a bit!

evalml/pipelines/components/transformers/imputers/per_column_imputer.py
angela97lin: Added a warning during fit time for this case :)

PR #3126, opened 2021-12-06
Refactor tests to avoid using importorskip

### LINKED ISSUE #2922: Refactor tests so that they fail when a dependency that should be installed is not installed
This came out of out Sept B sprint retro. Our CI silently failed to install Prophet and we didn't notice because the Prophet tests didn't fail! 

The reason they didn't fail is because we have `pytest.import_or_skip` at the top of the prophet test file. This means that the prophet tests were being skipped in the `git-test-prophet` job!

We should refactor our tests so that they fail when a dependency that should be installed is missing.

One way to do this is to move away from `pytest.import_or_skip` and instead leverage the `--has-minimal-dependencies` custom pytest option in our [conftest.py](https://github.com/alteryx/evalml/blob/main/evalml/tests/conftest.py#L246) to skip tests.

This will make it so that tests that need our optional dependencies are skipped when we run our core dependencies tests. When we run our core + optional dependencies tests, no tests should be skipped from missing dependencies.

We can follow this [guide](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option) from the pytest documentation so that we can skip tests based on command line values.

freddyaboulton:
### Pull Request Description
Fixes #2922 

-----

### DIFF
--- a/evalml/tests/component_tests/test_oversampler.py
+++ b/evalml/tests/component_tests/test_oversampler.py
@@ -253,19 +253,17 @@ def test_samplers_perform_equally(
     np.testing.assert_equal(sorted(y_im), expected_y)
 
 
-def test_smoten_categorical_boolean(X_y_binary):
-    from imblearn import over_sampling as im
+def test_smoten_categorical_boolean(X_y_binary, im):
 
     X, y = X_y_binary
     X_ww = infer_feature_types(X, feature_types={0: "Categorical", 1: "Boolean"})
     X_ww = X_ww.drop(range(2, len(X_ww.columns)), axis=1)
     sn = Oversampler()
-    X_out, y_out = sn.fit_transform(X_ww, y)
+    _ = sn.fit_transform(X_ww, y)
     assert sn.sampler == im.SMOTEN
 
 
-def test_smotenc_boolean_numeric(X_y_binary):
-    from imblearn import over_sampling as im
+def test_smotenc_boolean_numeric(X_y_binary, im):
 
     X, y = X_y_binary
     X_ww = infer_feature_types(X, feature_types={5: "Boolean", 12: "Boolean"})
--- a/evalml/tests/component_tests/test_oversampler.py
+++ b/evalml/tests/component_tests/test_oversampler.py
@@ -9,10 +9,7 @@ from evalml.exceptions import ComponentNotYetFittedError
 from evalml.pipelines.components import Oversampler
 from evalml.utils.woodwork_utils import infer_feature_types
 
-im = pytest.importorskip(
-    "imblearn.over_sampling",
-    reason="Skipping test because imbalanced-learn not installed",
-)
+pytestmark = pytest.mark.noncore_dependency
 
 
 def test_init():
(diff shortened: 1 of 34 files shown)


import_or_skip.sh
```diff
@@ -0,0 +1,8 @@
+# if there are more than 0 lines with importorskip, go into the if branch
```
freddyaboulton, line 1: Will prevent us from using importorskip in the future:

image

evalml/tests/conftest.py
```diff
@@ -48,6 +48,7 @@ def pytest_configure(config):
         "markers",
         "skip_offline: mark test to be skipped if offline (https://api.featurelabs.com cannot be reached)",
     )
+    config.addinivalue_line("markers", "noncore_dependency: mark test as slow to run")
```
angela97lin, line 51: Nice! Is the reason why we mark these tests because they're slow to run? Do we want to mention instead that they contain dependencies that are not part of the core requirements instead?

import_or_skip.sh
angela97lin: Really cool idea, thank you for adding this as well as fixing all of the current references to `importorskip`! 😁

evalml/tests/component_tests/test_oversampler.py
```diff
@@ -272,7 +278,7 @@ def test_smotenc_categorical_features(X_y_binary):
     X, y = X_y_binary
     X_ww = infer_feature_types(X, feature_types={0: "Categorical", 1: "Categorical"})
     snc = Oversampler()
-    X_out, y_out = snc.fit_transform(X_ww, y)
+    _ = snc.fit_transform(X_ww, y)
     assert snc.categorical_features == [0, 1]
```
angela97lin, line 282: It's interesting that we don't test anything with the outputs 😅

evalml/tests/component_tests/test_oversampler.py
```diff
@@ -46,6 +43,8 @@ def test_sampler_selection(
     categorical_columns,
     mock_imbalanced_data_X_y,
 ):
+    from imblearn import over_sampling as im
```
angela97lin, line 46: Does this mean that if we want to use a non-core dependency in our tests, we now have to import it in every test that we want to use it in? 😮 

I wonder if this will cause the tests to run longer, since we have to import the same modules over and over. Probably not a noticeable difference given our tests but it makes me wonder if there's a way to create a fixture or something the way that we do for our estimators, and then call that instead of importing: `catboost = import_or_raise("catboost", error_msg=cb_error_msg)`

angela97lin (approved):
Really awesome work @freddyaboulton! LGTM, I just had some nit-picky comments about:
1. The message for our noncore_dependency marker
2. Having to import in each test we want to use the module. I wonder if there's a way to create some fixture for the file to get around that, but I'm not sure?

Thanks for doing this and adding a lint script to prevent us from falling back to this pattern! 🙏

evalml/tests/conftest.py
freddyaboulton: This is what happens when you copy-paste from the pytest docs 😂 Will update the message.

PR #3129, opened 2021-12-07
Nullable Types: Make Pipeline (Reloaded)

chukarsten:
Addresses: #3124 

This PR should update `make_pipeline()` such that it now appropriately addresses the presence of AgeNullable, BooleanNullable, and IntegerNullable dtypes in Woodwork dataframes and adds them into the preprocessing pipelines of these.  Added a new integration test file with the intent of adding more integration testing in addition to what's in here.

### DIFF
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -94,7 +94,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
         )
 
         self.X = infer_feature_types(X, ignore_nullable_types=True)
-        self.y = infer_feature_types(y)
+        self.y = infer_feature_types(y, ignore_nullable_types=True)
         self.problem_type = problem_type
         self.sampler_name = sampler_name
 
@@ -186,13 +186,14 @@ class DefaultAlgorithm(AutoMLAlgorithm):
         estimators = self._naive_estimators()
         pipelines = [
             make_pipeline(
-                self.X,
-                self.y,
-                estimator,
-                self.problem_type,
+                X = self.X,
+                y = self.y,
+                estimator = estimator,
+                problem_type = self.problem_type,
                 sampler_name=self.sampler_name,
                 extra_components=feature_selector,
                 extra_components_position="after_preprocessing",
+                parameters=self._pipeline_params
             )
             for estimator in estimators
         ]
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -93,7 +93,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             random_seed=random_seed,
         )
 
-        self.X = infer_feature_types(X)
+        self.X = infer_feature_types(X, ignore_nullable_types=True)
         self.y = infer_feature_types(y)
         self.problem_type = problem_type
         self.sampler_name = sampler_name
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -186,14 +186,14 @@ class DefaultAlgorithm(AutoMLAlgorithm):
         estimators = self._naive_estimators()
         pipelines = [
             make_pipeline(
-                X = self.X,
-                y = self.y,
-                estimator = estimator,
-                problem_type = self.problem_type,
+                X=self.X,
+                y=self.y,
+                estimator=estimator,
+                problem_type=self.problem_type,
                 sampler_name=self.sampler_name,
                 extra_components=feature_selector,
                 extra_components_position="after_preprocessing",
-                parameters=self._pipeline_params
+                parameters=self._pipeline_params,
             )
             for estimator in estimators
         ]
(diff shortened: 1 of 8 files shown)


evalml/pipelines/utils.py
```diff
@@ -74,6 +75,16 @@ def _get_drop_all_null(X, y, problem_type, estimator_class, sampler_name=None):
     return component
 
 
+def _get_replace_null(X, y, problem_type, estimator_class, sampler_name=None):
```
freddyaboulton, line 78: I think we need to modify `_get_ohe` to add a ohe not encoder when there are boolean nullable present. A good test case would be a dataset with only Double and BooleanNullable features.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         allow_long_running_models=False,
         verbose=False,
     ):
-        self.X = infer_feature_types(X)
+        self.X = infer_feature_types(X, ignore_nullable_types=True)
```
jeremyliweishih, line 92: should this be added to `DefaultAlgorithm` as well?

tamargrey (approved):
This looks good to me from a Woodwork perspective

evalml/tests/conftest.py
```diff
@@ -92,6 +92,94 @@ def graphviz():
...
+            if "categorical" in column_names:
+                logical_types.update({"categorical": "Categorical"})
+            if "url" in column_names:
+                logical_types.update({"url": "URL"})
+            if "email" in column_names:
+                logical_types.update({"email": "EmailAddress"})
+            if "int_null" in column_names:
+                logical_types.update({"int_null": "integer_nullable"})
+            if "age_null" in column_names:
+                logical_types.update({"age_null": "integer_nullable"})
```
thehomebrewnerd, line 170: Should this be `"age_nullable"`?

evalml/tests/conftest.py
```diff
@@ -92,6 +92,94 @@ def graphviz():
...
+@pytest.fixture
+def get_test_data_from_configuration():
+    def _get_test_data_from_configuration(input_type, problem_type, column_names=None):
+        X_all = pd.DataFrame(
+            {
+                "all_null": [np.nan, np.nan, np.nan, np.nan, np.nan, np.nan, np.nan]
+                * 2,
+                "int_null": [0, 1, 2, np.nan, 4, np.nan, 6] * 2,
+                "age_null": [0, 1, 2, np.nan, 4, np.nan, 6] * 2,
+                "bool_null": [True, None, False, True, False, None, True] * 2,
```
thehomebrewnerd, line 104: I think these will have dtypes of either `float64` for the integer columns or `object` for the boolean column when you create this dataframe.

Do you also need to make sure to handle the case where these columns might have the `Int64` or `boolean` dtype, particularly when testing with the pandas version of the input?

```
"int_null": pd.Series([0, 1, 2, np.nan, 4, np.nan, 6] * 2, dtype="Int64"),
"age_null": pd.Series([0, 1, 2, np.nan, 4, np.nan, 6] * 2, dtype="Int64"),
"bool_null": pd.Series([True, None, False, True, False, None, True] * 2, dtype="boolean"),
```

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -186,13 +186,14 @@ def _create_naive_pipelines(self, use_features=False):
...
-                estimator,
-                self.problem_type,
+                X=self.X,
+                y=self.y,
+                estimator=estimator,
+                problem_type=self.problem_type,
                 sampler_name=self.sampler_name,
                 extra_components=feature_selector,
                 extra_components_position="after_preprocessing",
+                parameters=self._pipeline_params,
```
freddyaboulton, line 196: Was this a previous bug?

freddyaboulton (approved):
Thank you for this @chukarsten ! Looks good to me. I think the integration test you have in place is easy to extend as we progress through the epic, which is neat.

evalml/automl/automl_algorithm/default_algorithm.py
jeremyliweishih: @freddyaboulton  the pipeline parameters should have been passed in through [line 201](https://github.com/alteryx/evalml/pull/3129/files#diff-3968539fa92e26ee6c6a8ae303303ee0aed6cae4a3afa3d1007d26b4b7b0c09fR201) where it's added in `_transform_parameters()`. 

@chukarsten did you notice that the pipeline parameters were not passing in?

evalml/tests/integration_tests/test_nullable_types.py
```diff
@@ -0,0 +1,78 @@
...
+        X_train=X,
+        y_train=y,
+        problem_type=problem_type,
+        problem_configuration=parameters,
+        _automl_algorithm=automl_algorithm,
+    )
+    if automl_algorithm == "iterative":
+        pipelines = [pl.name for pl in automl.allowed_pipelines]
+    elif automl_algorithm == "default":
+        pipelines = [pl.name for pl in automl._automl_algorithm.next_batch()]
```
jeremyliweishih, line 59: Line 59 would only test the for the first batch of pipelines in `DefaultAlgorithm` and not all the pipelines that would be generated. To get the full list of pipelines you would need to follow the pattern [here](https://github.com/alteryx/evalml/blob/main/evalml/tests/automl_tests/test_default_algorithm.py#L221).  However, I don't see why behavior would change from the first batch to the next re adding the component since everything is built from `make_pipelines()`.

Issue #3116, opened 2021-12-02
Update data checks' `validate` method to return actions under new API


angela97lin:
We want to support the idea of returning action options from `validate()`, rather than the `actions` themselves.

Right now, we return:
```
data_check = InvalidTargetDataCheck("binary", "Log Loss Binary")
data_checks_output = data_check.validate(None, y)
>>> [{"code": "IMPUTE_COL", 
      "metadata": {
          "impute_strategy": "most_frequent",
          "is_target": True,
          "rows": None,
          "columns": None
      }
  }
]
```

Completion of this issue would update `validate` to look something like this instead: 
```
data_check = InvalidTargetDataCheck("binary", "Log Loss Binary")
data_checks_output = data_check.validate(None, y)
>>>   
{
  "warnings": [...],
  "errors": [...],
  "actions": {
    "actions_list": [
{
  "data_check_name"
  "action_code": IMPUTE_COLS,
  "parameters": {
    "impute_strategies": {
      "parameter_type": "column",
      "columns": {
        "a": {
          "impute_strategy”: {"categories": ["mean", "mode"], "type": "category", "default_value": "mean"},
          "constant_fill_value": {"type": "float", "default_value": 0}
        }
        "b": {
          "impute_strategy”: {"categories": ["mean", "mode"], "type": "category", "default_value": "mean"},
          "constant_fill_value": {"type": "float", "default_value": 0}
        }
        "c": {
          "impute_strategy”: {"categories": ["mean", "mode"], "type": "category", "default_value": "mean"},
          "constant_fill_value": {"type": "float", "default_value": 0}
        }
        ...
      }
    }
    "constant_fill_value": {
        "type": "float"
    }
  }
  "is_target": False,
  "rows": None,
  "columns": ["a", "b", "c", "d", "e"]
}
    ],
    "default_action": 0 # index to action that is considered the default action to take
}
```

chukarsten:
@angela97lin I'm connecting this to the issue we chatted about with respect to generating the helper function (I think) to streamline the OS usage of data checks and actions.  Let me know if you think this is not the blocker for that.

PR #3134, opened 2021-12-08
Add ``DataCheckActionOption`` class

angela97lin:
Part of #3116


evalml/data_checks/data_check_action.py
```diff
@@ -31,7 +31,11 @@ def __eq__(self, other):
         Returns:
             bool: True if the other object is considered an equivalent data check action, False otherwise.
         """
-        return self.action_code == other.action_code and self.metadata == other.metadata
+        attributes_to_check = ["action_code", "data_check_name", "metadata"]
```
jeremyliweishih, line 34: any new cases we should add to `test_data_check_action_equality` or `test_data_check_action_inequality`?

evalml/data_checks/data_check_action_option.py
```diff
@@ -0,0 +1,129 @@
...
+        action_code (DataCheckActionCode): Action code associated with the action option.
+        data_check_name (str): Name of the data check that produced this option.
+        parameters (dict): Parameters associated with the action option. Defaults to None.
+        metadata (dict, optional): Additional useful information associated with the action option. Defaults to None.
+    """
+
+    def __init__(self, action_code, data_check_name, parameters=None, metadata=None):
+        self.action_code = action_code
+        self.data_check_name = data_check_name
+        self.parameters = parameters
```
jeremyliweishih, line 23: Might be useful to leave the structure required for what a parameter looks like (esp since there's so much validation going on)

jeremyliweishih (approved):
LGTM, just some comments

evalml/tests/data_checks_tests/test_data_check_action_option.py
```diff
@@ -0,0 +1,378 @@
...
+        parameters=parameters,
+    )
+    assert data_check_action_option.action_code == DataCheckActionCode.DROP_COL
+    assert data_check_action_option.data_check_name == dummy_data_check_name
+    assert data_check_action_option.metadata == {"columns": [1, 2], "rows": None}
+    assert data_check_action_option.parameters == parameters
+
+
+def test_data_check_action_option_equality(dummy_data_check_name):
+    data_check_action_option = DataCheckActionOption(
```
jeremyliweishih, line 63: I think the equality or inequality tests cover every case but it would be nice to double check since code coverage won't catch the cases because of how the loop is written!

evalml/data_checks/data_check_action_option.py
bchen1116: Agreed, I think it'd be useful to show an example of an initialized `DataCheckActionOption` object with valid  parameters

evalml/data_checks/data_check_action_option.py
```diff
@@ -0,0 +1,129 @@
...
+            ValueError: If input dictionary does not have keys `code` and `metadata` and if the `metadata` dictionary does not have keys `columns` and `rows`.
+
+        Returns:
+            DataCheckActionOption object from the input dictionary.
+        """
+        if "code" not in action_dict or "metadata" not in action_dict:
+            raise ValueError(
+                "The input dictionary should have the keys `code` and `metadata`."
+            )
+        if (
```
bchen1116, line 78: So we're enforcing that the metadata needs to have values for both `columns` and `rows`? I'm curious if it'd be better to enforce that we see at least one of those values, since for some actions, it might not make sense to include the other value (ie Dropping Rows, to include `columns` as a key). A nit, but figured it might simplify things for open source users

evalml/data_checks/data_check_action_option.py
```diff
@@ -0,0 +1,129 @@
...
+                if "columns" not in parameter_value:
+                    raise ValueError(
+                        "Each `column` parameter type must also have a `columns` key indicating which columns the parameter should address."
+                    )
+                columns = parameter_value["columns"]
+                if not isinstance(columns, dict):
+                    raise ValueError(
+                        "`columns` must be a dictionary, where each key is the name of a column and the associated value is a dictionary of parameters for that column."
+                    )
+                for _, column_parameters in columns.items():
```
bchen1116, line 117: nit for this and the following for loop, since you don't use the first element of the item, can you instead iterate through `.values()` instead of `.items()`?

evalml/tests/data_checks_tests/test_data_check_action.py
```diff
@@ -5,7 +5,7 @@
 
 @pytest.fixture
 def dummy_data_check_name():
```
bchen1116, line 7: nit: If you use this fixture in multiple files, would it be beneficial to move to the `conftest.py` file?

bchen1116 (commented):
Great test coverage! Left a few questions for myself and nitpicky suggestions that might be helpful for this

PR #3139, opened 2021-12-09
Add TimeSeriesParametersDataCheck to DefaultDataChecks

### LINKED ISSUE #3125: Add TimeSeriesParametersDataCheck to DefaultDataChecks
In https://github.com/alteryx/evalml/pull/3111, we added a data check to check whether the selected time series parameters are compatible with the data splitting strategy. We should add this data check to the set of default data checks and make sure it only runs for time series problems.

ParthivNaresh:
Fixes #3125 

-----


evalml/automl/automl_search.py
```diff
@@ -150,21 +150,6 @@ def search(
     y_train = infer_feature_types(y_train)
     problem_type = handle_problem_types(problem_type)
 
-    datetime_column = None
```
ParthivNaresh, line 153: I incorporated this `time_index` check into `contains_all_ts_parameters` to reduce repeat code

evalml/tests/data_checks_tests/test_ts_parameters_data_check.py
```diff
@@ -7,12 +7,33 @@
 )
 
 
-def test_time_series_param_data_check_raises_value_error():
+@pytest.mark.parametrize(
```
chukarsten, line 10: You could parametrize all 4 variable individually between the values [1, None, "missing"], if you wanted complete coverage!

chukarsten (approved):
Looks good, Parthiv.  Thanks for doing this.  Left a minor comment that you can take or leave!

evalml/automl/automl_search.py
freddyaboulton: Cause with this change, if user passes in a dict without `time_index` for example, they get a KeyError as opposed to us telling them the problem_configuration is not valid.

freddyaboulton (approved):
Thank you @ParthivNaresh ! This looks good to me. The only thing I think we need to address before merge is validating the problem config in the lower-case `search` methods like we currently do on main. I think that will mean we don't have to delete the tests @chukarsten asked about.

PR #3135, opened 2021-12-09
Changes and fixes for time series classification

ParthivNaresh:
Fixes #1681 #3095 #3125 

I'd recommend starting with `evalml/utils/gen_utils.py`

PR #3141, opened 2021-12-09
Add Time Series Splitting Data Check

### LINKED ISSUE #1681: Changes and fixes for time series classification problems
Comment with reproducer thanks to @bchen1116 :

https://github.com/alteryx/evalml/pull/1666#pullrequestreview-566400601

ParthivNaresh:
Fixes #1681


evalml/automl/automl_search.py
```diff
@@ -394,6 +406,7 @@ def __init__(
         max_time=None,
         patience=None,
         tolerance=None,
+        n_splits=3,
```
freddyaboulton, line 409: @ParthivNaresh I don't think we need to add this parameter to `AutoMLSearch`. Isn't it implicit in the `data_splitter`?

evalml/automl/automl_search.py
```diff
@@ -180,11 +182,14 @@ def search(
         "patience": patience,
         "tolerance": tolerance,
         "verbose": verbose,
+        "problem_configuration": problem_configuration,
+        "n_splits": n_splits,
```
freddyaboulton, line 186: Rather than threading this parameter to down to `AutoMLSearch` we should just call `make_data_splitter` and pass in as the `data_splitter` parameter?

evalml/data_checks/ts_splitting_data_check.py
```diff
@@ -0,0 +1,93 @@
...
+
+        y = infer_feature_types(y)
+
+        invalid_splits = {}
+        if y is not None:
+            for split_num, (train, val) in enumerate(self._splitter.split(X=y)):
+                invalid_dict = {}
+                train_targets = y[train]
+                val_targets = y[val]
+                if train_targets.nunique() < y.nunique():
```
freddyaboulton, line 73: I think we can call this once? I believe pandas doesn't cache the result so we're redoing all the work every time.

evalml/data_checks/ts_splitting_data_check.py
```diff
@@ -0,0 +1,93 @@
...
+            ...     "warnings": [],
+            ...     "actions": []}
+        """
+        results = {"warnings": [], "errors": [], "actions": []}
+
+        y = infer_feature_types(y)
+
+        invalid_splits = {}
+        if y is not None:
+            for split_num, (train, val) in enumerate(self._splitter.split(X=y)):
```
freddyaboulton, line 69: I left a [comment](https://github.com/alteryx/evalml/pull/3135/files#r766101417) about this in the original PR, but don't we need to only check the first training fold. The size of the first fold is n_obs // (n_splits + 1), maybe that can help us avoid instantiating a splitter and iterating over all splits?

freddyaboulton (commented):
@ParthivNaresh Looks good! I think the main thing is that we should be creating a `data_splitter` in `search` and `search_iterative` rather than adding a new parameter to `AutoMLSearch` init. 

Secondly, I think we only need to care about the first training split? As long as the pipeline is trained with all targets, the pipeline will be able to predict on the validation set, regardless of the number of unique target values in the validation set. Since the splits are sequentially constructed, if the first split has all the target values, all the subsequent ones will too.

docs/source/user_guide/data_checks.ipynb
```diff
@@ -541,6 +542,40 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Time Series Splitting\n",
+    "\n",
+    "Due to the nature of time series data, splitting cannot involve shuffling and has to be done in a sequential manner. This means splitting the data into `n_splits` + 1 different sections and increasing the size of the training data by the split size every iteration while keeping the test size equal to the split size.\n",
+    "\n",
+    "For every split in the data, the training and validation segments must contain target data that has an example of every class found in the entire target set for time series binary and time series multiclass problems. The reason for this is that many classification machine learning models run into issues if they're trained on data that doesn't contain an instance of a class but then the model is expected to be able to predict for it. For example, with 3 splits and a split size of 25, this means that every training/validation split: (0:25)/(25:50), (0:50)/(50:75), (0:75)/(75:100) must contain at least one instance of all unique target classes in the training and validation set.\n",
```
freddyaboulton, line 553: 👍

evalml/automl/automl_search.py
ParthivNaresh: I was thinking about the use case where a user starts with `search` and decides that while they don't want to pass in their own data splitter, they might want to determine the number of splits. It would be passed into the `make_data_splitter` [below](https://github.com/alteryx/evalml/pull/3141/files#diff-8b4df814812e6f5df24c600bc7f429b036e0e2f78613bfbdcdf0ed3ab86bffd6R613)

evalml/data_checks/ts_splitting_data_check.py
ParthivNaresh: So the issue is that it's not only the training but also validation split that incurs an error. It occurs [here](https://github.com/alteryx/evalml/blob/main/evalml/automl/engine/engine_base.py#L205) in `train_and_score_pipeline ` of `engine_base`. While the first training/validation split might be valid, the following one which includes the next `split_size` # of observations for validation might error out.

evalml/data_checks/ts_splitting_data_check.py
ParthivNaresh: good call!

evalml/data_checks/ts_splitting_data_check.py
freddyaboulton: Ah makes sense, thanks for explaining!

evalml/automl/automl_search.py
freddyaboulton: Makes sense! I'm not opposed to that but maybe that should be its own issue. This API change is not related to the scope of the original issue.

freddyaboulton (approved):
Thanks @ParthivNaresh ! Looks good to me.

In my opinion, it's best to leave the `n_splits` addition to the `AutoMLSearch` api for its own issue. In `search` and `search_iterative` we can create the data splitter with `make_data_splitter` and pass to `AutoMLSearch` as the `data_splitter` param. 

Curious what other people on the team think about that, but changing the AutoMLSearch api feels like it's own issue since it's not 100% needed to close this issue out.

evalml/automl/automl_search.py
ParthivNaresh: Gotcha, understandable. I'm fine with making a data split in `search` and `search_iterative` and leaving `n_splits` alone for another issue

evalml/tests/data_checks_tests/test_ts_splitting_data_check.py
```diff
@@ -0,0 +1,65 @@
...
+    [
+        ["time series binary", True],
+        ["time series binary", False],
+        ["time series multiclass", True],
+        ["time series multiclass", False],
+    ],
+)
+def test_time_series_param_data_check(problem_type, is_valid):
+    X = None
+    y = pd.Series([0 if i < 10 else 1 for i in range(100)])
```
bchen1116, line 29: This y isn't used, is it?

bchen1116 (approved):
LGTM! Left some minor comments, but looks great!

evalml/tests/data_checks_tests/test_ts_splitting_data_check.py
ParthivNaresh: Good catch!

Issue #3076, opened 2021-11-18
Use one sampler for split preprocessing pipeline


jeremyliweishih:
In the split preprocessing pipelines introduced by #2986 each preprocessing chain contains a sampler as each chain is generated by `make_pipeline()` and currently the y output to the final estimator is taken from the categorical chain's sampler. This might be an issue as oversampling is handled differently for categorical and numeric data as detailed in this conversation. This issue tracks investigating the effects of choosing the categorical sampler and ensuring there is no performance lost from doing so.

jeremyliweishih:
@freddyaboulton this tracked the double sampler issue but not sure why I closed this. I can open this up to track double sampler but a more generalized issue might be better as its touching multiple projects.

freddyaboulton:
@jeremyliweishih Thank you for reopening ! I've renamed it to be more general and also linking with #2997 which I think is the root cause of these kinds of issues.

freddyaboulton:
We should prioritize this as I think there may be some performance implications to this.

Repro

1. Comment out this line in `_create_fast_fina`
```python
    def _create_fast_final(self):
        estimators = [
            estimator
            for estimator in get_estimators(self.problem_type)
            #if estimator not in self._naive_estimators()
        ]
```
2. Run Automl search on the fraud dataset
```python
from evalml.automl import AutoMLSearch
from evalml.preprocessing import split_data
import pytest
from evalml.demos import load_fraud

X, y = load_fraud(1000)


X_train, _, y_train, _ = split_data(X, y, "binary", test_size=.5)

automl = AutoMLSearch(X, y, "binary", verbose=True, automl_algorithm="default")
automl.search()
```

You'll notice that the non-split RF pipeline does better than the split RF pipeline

### Non split
```
Random Forest Classifier w/ Label Encoder + DateTime Featurizer + Imputer + One Hot Encoder + Oversampler + RF Classifier Select From Model:
	Starting cross validation
	Finished cross validation - mean Log Loss Binary: 0.253
```

### Split
```
Random Forest Classifier w/ Label Encoder + Select Columns By Type Transformer + Label Encoder + DateTime Featurizer + Imputer + Oversampler + Select Columns Transformer + Select Columns Transformer + Label Encoder + Imputer + One Hot Encoder + Oversampler:
	Starting cross validation
	Finished cross validation - mean Log Loss Binary: 0.277
```

[...]

Issue #2987, opened 2021-10-28
Label Encoder appears twice for AutoML-generated stacked ensemble pipeline


christopherbunn:
Currently, for our stacked ensemble pipelines, we are adding a label encoder at the very beginning of all stacked ensemble pipelines for every single classification problem. While this ensures that we have a label encoder for each stacked ensemble pipeline, it also results in a duplicate label encoder if the input pipeline already has its own label encoder. Because these input pipelines are all AutoML generated, they will all have a label encoder in the beginning. An example of this is below:

![image](https://user-images.githubusercontent.com/8752455/139311098-92b82e86-5f0a-4a59-a58a-01227582e0dd.png)

We have two possible paths of fixing this:
- Remove the lines of code that adds a new Label Encoder to every single pipeline
   - Pros:
      - If (somehow) we have custom label encoders in the future and each input pipeline has it's own specific version, this will allow that to exist with fewer modifications 
   - Cons: 
     - No single source of truth for how label encoding works across pipeline. This opens up the (admittedly very small) possibility that each input pipeline has a different encoded target y as an output.
     - Adds additional clutter to the output for what is likely the same component in >95% of cases.
- Remove the label encoders from a pipeline and consolidate with a single new label encoder
  - Pros:
      - Single source of truth for the label encoder 
      - Having one consolidated component looks cleaner in the graph
   - Cons: 
      - Does not allow for different types of LabelEncoders in the future to exist in the same grap

christopherbunn:
Hmm, I guess it's less of a bug since it's intended behavior at this point. I think it's pretty reasonable from an end user perspective to expect that we only have one label encoder (whether that's one per input pipeline or one for the entire pipeline). If we do decide to add a label encoder to all classification ensembles, we should definitely have some sort of check so we aren't duplicating label encoders.

dsherry:
Which option is simplest?

@chukarsten and I like option 1 from your list, which is:
* In our code for constructing ensembler pipeline, don't add a label encoder at the beginning. In other words, assume that each pipeline will handle its own encoding

When we get around to #2911 , we may need to add a label encoder as part of the preprocessing pipeline to provide original input features to the ensembler alongside the input pipelines' predictions. We can handle that separately in #2911 .

chukarsten:
Let's make sure that when we tackle this, the code changed in #3149 is refactored.  The test should fail, but it's at least one known place where we will need to go back.

jeremyliweishih:
[...]

Issue #2868, opened 2021-10-01
Remove algorithm logic out of AutoML tests


jeremyliweishih:
Related to #2656.

This issue tracks moving algorithm specific tests out of test_automl.py, test_automl_search_classification.py and test_automl_search_regression.py by mocking out next_batch and algorithm specific methods. This will improve the process of adding new algorithms by removing connecting logic between AutoMLSearch and each algorithm.

jeremyliweishih:
After going through all the tests in #3261, I've decided not to remove logic out in #3261 but will do the following in #3261: 

1. Tag any `IterativeAlgorithm` specific tests to end with `_iterative` and organize such tests
2. Generalize any tests I can to be algorithm agnostic (i.e removing strict number of pipeline assertions)
3. Compile a list of automl tests that can benefit from having a `MockAlgorithm` (similar to the `AutomlTestEnv`) such that we can edit these tests to test only `AutoMLSearch`

Then this issue will implement `MockAlgorithm` (a generic testing algorithm) and edit the tests above. When #3261 is finished, I will update this issue with the list.

jeremyliweishih:
List of tests that we can use a mock algorithm with:
```
1. test_automl_algorithm
2. test_time_series_regression_with_parameters
3. test_max_batches_works
4. test_callback (there’s multiple)
5. test_automl_tuner_exception
6. test_results_getter
7. test_automl_pipeline_params_kwargs
8. test_search_with_text_and_ensembling_iterative
9. test_pipelines_per_batch_iterative
10. test_automl_respects_random_seed_iterative
11. test_automl_respects_pipeline_parameters_with_duplicate_components_iterative
12. test_automl_respects_pipeline_custom_hyperparameters_with_duplicate_components_iterative
13. test_automl_search_sampler_ratio
14. test_automl_search_sampler_method
15. test_time_series_pipeline_parameter_warnings
16. test_pipeline_count
```
List of tests we can maybe generalize with much larger lift:
- any tests with component graphs or iterative specific parameters

bchen1116:
Make sure to address [this comment](https://github.com/alteryx/evalml/pull/3373/#discussion_r832499447) as well when closing out this issue

Issue #2691, opened 2021-08-24
Support time series in new automl algo


jeremyliweishih:
The new automl algo has specific batches where ensembling is run but our implementation of _make_stacked_ensemble_pipeline() does not support time series as a problem type. This issue tracks either:

- support time series problems in _make_stacked_ensemble_pipeline()
or
- disabling ensembling for time series problems

chukarsten:
Per our discussion, I ran into this lack of support in the resolution of #3124 .  The PR creates integration tests in `integration_tests/test_nullable_types.py` and uses some datasets that have dates in them, prompting `AutoMLSearch()` to create time series pipelines.  The `DefaultAlgorithm` was unable to pass through the time series parameters into the generated pipelines.

#3124 makes a basic fix to allow the first 3 batches of `DefaultAlgorithm` to generate the proper pre-processing pipelines for time series, but we should add to the acceptance criteria of this issue the increasing of the number of batches  in the `test_nullable_types` integration test to the maximum to ensure all the proper time series pipelines are created with the transformer.

Additionally, this suggests we need more testing around time series/ensembling.  We can actually write these tests and use `pytest.xfail` since we know that this support doesn't exist yet.

```
../../automl/automl_algorithm/default_algorithm.py:322: in next_batch
    next_batch = self._create_fast_final()
../../automl/automl_algorithm/default_algorithm.py:260: in _create_fast_final
    pipelines = [
../../automl/automl_algorithm/default_algorithm.py:261: in <listcomp>
    self._make_split_pipeline(
../../automl/automl_algorithm/default_algorithm.py:432: in _make_split_pipeline
    numeric_pipeline = make_pipeline(
../../pipelines/utils.py:355: in make_pipeline
    return base_class(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'TimeSeriesMulticlassClassificationPipeline' object has no attribute 'component_graph'") raised in repr()] TimeSeriesMulticlassClassificationPipeline object at 0x1515dd160>
component_graph = {'DateTime Featurization Component': [<class 'evalml.pipelines.components.transformers.preprocessing.datetime_featuriz...pelines.components.transformers.imputers.imputer.Imputer'>, 'Drop Null Columns Transformer.x', 'Label Encoder.y'], ...}
[...]

PR #3149, opened 2021-12-13
Add support for known-in-advanced features for time series

### LINKED ISSUE #2511: Spike - Handle known-in-advance features
We have received feedback that we want our time series pipelines to be aware of features that are "known in advance". 

The outcome of this spike is a design document detailing know to do that. I think there are two important pieces

- [ ] What are "known in advance" features?
- [ ] I think we'll want to definitely cover well-known holidays with the ability to add custom holidays. Can we add transformers that compute indicator for `is_holiday`? Or should we be more granular and add `is_july_4th`, `is_christmas` holidays?

freddyaboulton:
### Pull Request Description
Fixes #2511 

Design doc [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1107230831/2511+-+Handle+Known+In+Advanced+Features)

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,7 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
         * Added the ability to accept serialized features and skip computation in ``DFSTransformer`` :pr:`3106`
-        * Added support for known-if-advance features :pr:`3149`
+        * Added support for known-in-advance features :pr:`3149`
     * Fixes
     * Changes
         * ``TimeSeriesParametersDataCheck`` was added to ``DefaultDataChecks`` for time series problems :pr:`3139`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,6 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
         * Added the ability to accept serialized features and skip computation in ``DFSTransformer`` :pr:`3106`
+        * Added support for known-if-advance features :pr:`3149`
     * Fixes
     * Changes
         * ``TimeSeriesParametersDataCheck`` was added to ``DefaultDataChecks`` for time series problems :pr:`3139`
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -152,3 +152,12 @@ class AutoMLAlgorithm(ABC):
                 self.logger.info(
                     f"Removing columns {unknown_columns} because they are of 'Unknown' type"
                 )
+        kin_columns = self._pipeline_params.get("pipeline", {}).get(
+            "known_in_advance", []
+        )
+        if kin_columns:
+            no_kin_columns = [c for c in self.X.columns if c not in kin_columns]
+            kin_name = "Known In Advance Pipeline - Select Columns Transformer"
+            no_kin_name = "Not Known In Advance Pipeline - Select Columns Transformer"
+            self._pipeline_params[kin_name] = {"columns": kin_columns}
+            self._pipeline_params[no_kin_name] = {"columns": no_kin_columns}
--- a/evalml/automl/automl_algorithm/automl_algorithm.py
+++ b/evalml/automl/automl_algorithm/automl_algorithm.py
@@ -152,12 +152,12 @@ class AutoMLAlgorithm(ABC):
                 self.logger.info(
                     f"Removing columns {unknown_columns} because they are of 'Unknown' type"
                 )
-        kin_columns = self._pipeline_params.get("pipeline", {}).get(
+        kina_columns = self._pipeline_params.get("pipeline", {}).get(
             "known_in_advance", []
         )
-        if kin_columns:
-            no_kin_columns = [c for c in self.X.columns if c not in kin_columns]
+        if kina_columns:
+            no_kin_columns = [c for c in self.X.columns if c not in kina_columns]
             kin_name = "Known In Advance Pipeline - Select Columns Transformer"
             no_kin_name = "Not Known In Advance Pipeline - Select Columns Transformer"
-            self._pipeline_params[kin_name] = {"columns": kin_columns}
+            self._pipeline_params[kin_name] = {"columns": kina_columns}
             self._pipeline_params[no_kin_name] = {"columns": no_kin_columns}
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -199,6 +199,9 @@ class IterativeAlgorithm(AutoMLAlgorithm):
                         self.problem_type,
                         parameters=self._pipeline_params,
                         sampler_name=self.sampler_name,
+                        known_in_advance=self._pipeline_params.get("pipeline", {}).get(
+                            "known_in_advance", None
+                        ),
                     )
                     for estimator in allowed_estimators
                 ]
@@ -432,9 +435,14 @@ class IterativeAlgorithm(AutoMLAlgorithm):
                 component_parameters["n_jobs"] = self.n_jobs
             if "number_features" in init_params:
                 component_parameters["number_features"] = self.number_features
+            names_to_check = [
+                "Drop Columns Transformer",
+                "Known In Advance Pipeline - Select Columns Transformer",
+                "Not Known In Advance Pipeline - Select Columns Transformer",
+            ]
             if (
                 name in self._pipeline_params
-                and name == "Drop Columns Transformer"
+                and name in names_to_check
                 and self._batch_number > 0
             ):
                 component_parameters["columns"] = self._pipeline_params[name]["columns"]
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -313,6 +313,8 @@ def _make_pipeline_time_series(
 
     if known_in_advance:
         preprocessing_components = [SelectColumns] + preprocessing_components
+    else:
+        preprocessing_components += [estimator]
 
     component_graph = PipelineBase._make_component_dict_from_component_list(
         preprocessing_components
@@ -337,7 +339,6 @@ def _make_pipeline_time_series(
                 pipeline.name: "Not Known In Advance",
             },
         )
-        pipeline = pipeline.new(parameters)
     return pipeline
 
 
@@ -475,8 +476,8 @@ def _make_stacked_ensemble_pipeline(
         else {}
     )
     final_components = []
-    # used_model_families = []
-    # parameters = {}
+    used_model_families = []
+    parameters = {}
 
     if is_classification(problem_type):
         parameters = {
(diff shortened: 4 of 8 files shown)


evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -197,6 +198,114 @@ def test_make_pipeline(
...
+        assert expected_known_in_advance_components == known_in_advance_components
+        for k in [
+            "pipeline",
+            "Known In Advance Pipeline - Select Columns Transformer",
+            "Not Known In Advance Pipeline - Select Columns Transformer",
+        ]:
+            assert pipeline.parameters[k] == parameters[k]
+        if is_classification(problem_type):
+            assert (
+                len([c for c in pipeline.component_graph if "Label Encoder" in c.name])
```
freddyaboulton, line 300: This is #2987

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -197,6 +198,114 @@ def test_make_pipeline(
...
+                == 2
+            )
+            assert (
+                len([c for c in pipeline.component_graph if "Undersampler" in c.name])
```
freddyaboulton, line 304: This is #3076

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -432,9 +435,14 @@ def _transform_parameters(self, pipeline, proposed_parameters):
                 component_parameters["n_jobs"] = self.n_jobs
             if "number_features" in init_params:
                 component_parameters["number_features"] = self.number_features
+            names_to_check = [
```
freddyaboulton, line 438: This is https://github.com/alteryx/evalml/issues/3150

evalml/tests/integration_tests/test_time_series_integration.py
```diff
@@ -61,3 +61,70 @@ def test_can_run_automl_for_time_series_with_categorical_and_boolean_features(
...
+        problem_type=problem_type,
+        problem_configuration={
+            "max_delay": 5,
+            "gap": 3,
+            "forecast_horizon": 2,
+            "time_index": "date",
+            "known_in_advance": ["bool_feature", "cat_feature"],
+        },
+        optimize_thresholds=False,
+        sampler_method="Undersampler",
```
freddyaboulton, line 116: Using undersampler because https://github.com/alteryx/evalml/issues/3151

evalml/tests/integration_tests/test_time_series_integration.py
```diff
@@ -61,3 +61,70 @@ def test_can_run_automl_for_time_series_with_categorical_and_boolean_features(
...
+    ],
+)
+def test_can_run_automl_for_time_series_known_in_advance(
+    problem_type,
+):
+
+    X = pd.DataFrame(
+        {"features": range(101, 601), "date": pd.date_range("2010-10-01", periods=500)}
+    )
+    y = pd.Series(range(500))
```
chukarsten, line 81: super-nit: perhaps consolidate the periods into a var and refactor the range calls  and even the generation of the target to use that var?

evalml/tests/integration_tests/test_time_series_integration.py
```diff
@@ -61,3 +61,70 @@ def test_can_run_automl_for_time_series_with_categorical_and_boolean_features(
...
+            "max_delay": 5,
+            "gap": 3,
+            "forecast_horizon": 2,
+            "time_index": "date",
+            "known_in_advance": ["bool_feature", "cat_feature"],
+        },
+        optimize_thresholds=False,
+        sampler_method="Undersampler",
+    )
+    automl.search()
```
chukarsten, line 118: hahaha, I was going to ask "shouldn't this be an integration test?" Then I saw that it was.  Nice.

chukarsten (approved):
This looks great, Freddy.  Thanks for putting this together and thanks especially for the linking of all the new and existing issues.  I find this *extremely* helpful for people that pick up refactoring issues to find all the places where resolution of their underlying issue should lead to a refactor.

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -152,3 +152,12 @@ def _set_additional_pipeline_params(self):
                 self.logger.info(
                     f"Removing columns {unknown_columns} because they are of 'Unknown' type"
                 )
+        kin_columns = self._pipeline_params.get("pipeline", {}).get(
```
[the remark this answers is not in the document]
freddyaboulton: I'll use kina and add a comment? 😆

evalml/pipelines/utils.py
```diff
@@ -287,6 +289,85 @@ def _get_pipeline_base_class(problem_type):
...
+        # Since we specify the correct problem type for the not known-in-advance pipeline
+        # the label encoder and time series featurizer will be correctly added
+        kin_preprocessing = [SelectColumns] + _get_preprocessing_components(
+            X_known_in_advance, y, ProblemTypes.REGRESSION, estimator, sampler_name
+        )
+        kin_component_graph = PipelineBase._make_component_dict_from_component_list(
+            kin_preprocessing
+        )
+        kin_pipeline = base_class(
+            kin_component_graph, parameters=parameters, custom_name="Pipeline"
```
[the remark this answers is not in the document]
freddyaboulton: I'll add a comment for this! This is just so that both pipelines don't have the same name. Otherwise, there could be clashes in `make_pipeline_from_multiple_graphs` for some time series estimators like ARIMA. We could name it anything so `Pipeline` seemed like a good choice.

evalml/automl/automl_algorithm/iterative_algorithm.py
jeremyliweishih: question for my understanding: why do we need to check only specific components for `_pipeline_params` and not every component that shows up? [This is how it looks like in `DefaultAlgorithm`](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/default_algorithm.py#L419-L421).

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5454,3 +5454,60 @@ def test_get_ensembler_input_pipelines(X_y_binary, AutoMLTestEnv):
...
+
+@pytest.mark.parametrize(
+    "problem_type",
+    [
+        ProblemTypes.TIME_SERIES_REGRESSION,
+        ProblemTypes.TIME_SERIES_MULTICLASS,
+        ProblemTypes.TIME_SERIES_BINARY,
+    ],
+)
+def test_automl_passes_known_in_advance_pipeline_parameters_to_all_pipelines(
```
jeremyliweishih, line 5467: One tech debt issue with the algorithm upgrade is separating pipeline building logic out of search and removing pipeline building out of tests as well (#2868 tracks this). Do you think this test better belongs in `test_iterative_algorithm.py`? You can just test against `allowed_pipelines` to see if the pipelines generated match what you would expect.  Once I get to #2691 I can add a test in `DefaultAlgorithm` as well!

jeremyliweishih:
@freddyaboulton overall looks good to me! I'm adding a comment to test support for this in #2691 (my catch-all for TS issues and Default Algorithm). Will post updates there once I get to it.

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Thanks for bringing this up! The known-in-advance implementation requires us to set the parameters for SelectColumns correctly for all pipelines. That will have to hold for all automl algorithms - iterative, default, whatever we come up with in the future. Because of that, I thought it'd be cleaner to have a test in `test_automl` and then parametrize over automl algorithms (when default algorithm supports time series) as opposed to a test in every automl test file.

What do you think?

jeremyliweishih: Gotcha, that make a whole lot of sense to me. Thanks for explaining, the current test is great!

evalml/automl/automl_algorithm/iterative_algorithm.py
freddyaboulton: I think this is precisely the tech debt around the #3150. Our current approach for specifying pipeline parameters through the search is really confusing.

My guess of what happened:
* In https://github.com/alteryx/evalml/pull/1676, we added the ability to freeze parameters/hyperparameters via the `pipeline_parameters` component. This had a [check for batch 0](https://github.com/alteryx/evalml/commit/51cba3fa041f8c94b60261e6c2ceb0f944b87ea5), to make sure we initialized the pipelines with values that are in the hyperparameter ranges. This could only apply for batch 0 otherwise we would always override the values proposed by the tuner with random values. 
* Then in https://github.com/alteryx/evalml/pull/2138, we added support to drop index columns throughout search. Since we can't specify the parameters for DropColumns as a hyperparameter, we had to manually add logic to add them.
* In https://github.com/alteryx/evalml/pull/2317, we separated the custom hyperparameters from pipeline parameters. The design choice was for pipeline parameters to only apply for the first batch and for custom hyperparemeters to apply for all subsequent batches. Therefore we're in a position were the only components whose parameters should be overwritten with what's in `names_to_check` are the select/drop columns transformers.

Regarding the difference in behavior with `DefaultAlgorithm`, that may be a bug? I think pipeline parameters should only apply in the first batch - at least that's what the design stipulated when we separated parameters from custom hyperparameters.

jeremyliweishih: Thanks for the context @freddyaboulton, really helpful in catching up on this logic. I made the explicit choice to pass down `pipeline_parameters` into each `DefaultAlgorithm` pipeline because my thought was that `pipeline_params` represented static parameters that a user wanted in each pipeline. But I think you're right that the design in #2317 [made it only apply in the first batch.](https://github.com/alteryx/evalml/pull/2317/files#r645831953). I filed #3153 to track this divergent behavior for further discussion!

evalml/automl/automl_algorithm/automl_algorithm.py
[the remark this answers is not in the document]
chukarsten: It's OK, I just wanted to make sure that it wasn't something else that I just wasn't tracking.

evalml/pipelines/utils.py
```diff
@@ -287,6 +289,88 @@ def _get_pipeline_base_class(problem_type):
...
+    component_graph = PipelineBase._make_component_dict_from_component_list(
+        preprocessing_components
+    )
+    base_class = _get_pipeline_base_class(problem_type)
+    pipeline = base_class(component_graph, parameters=parameters)
+    if X_known_in_advance is not None:
+        # Pre-processing components do not depend on problem type so we
+        # are ok by specifying regression for the known-in-advance sub pipeline
+        # Since we specify the correct problem type for the not known-in-advance pipeline
+        # the label encoder and time series featurizer will be correctly added
```
ParthivNaresh, line 347: I still don't think I fully understand why `problem_type` can't be passed here as well

evalml/pipelines/utils.py
```diff
@@ -287,6 +289,88 @@ def _get_pipeline_base_class(problem_type):
...
+             Defaults to None
+         known_in_advance (list[str], None): List of features that are known in advance.
+
+    Returns:
+        PipelineBase: TimeSeriesPipeline''
+    """
+    if known_in_advance:
+        not_known_in_advance = [c for c in X.columns if c not in known_in_advance]
+        X_not_known_in_advance = X.ww[not_known_in_advance]
+        X_known_in_advance = X.ww[known_in_advance]
```
ParthivNaresh, line 324: Nice and clean implementation

ParthivNaresh (approved):
Looks great, the splitting into subgraphs is done very neatly and I love it. Also great test coverage.

docs/source/release_notes.rst
```diff
@@ -3,6 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
         * Added the ability to accept serialized features and skip computation in ``DFSTransformer`` :pr:`3106`
+        * Added support for known-if-advance features :pr:`3149`
```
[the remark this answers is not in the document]
freddyaboulton: Fixing now 😆

evalml/pipelines/utils.py
freddyaboulton: Great question. We can't specify a time series problem type because then the pipeline for the known-in-advance features will have a time series featurizer which is not what we want. We could convert a time series problem type to a non-time-series problem type (time series binary -> binary) but that's the same as just using regression. Since the not-known-in-advance pipeline will have the "right" components for the problem type, e.g. label encoder, samplers, etc, the overall pipeline will have them too.

freddyaboulton: Elaborating on this in the comment!

Issue #3061, opened 2021-11-16
Nullable Types: Verify DataChecks Work


chukarsten:
As an EvalML user, I would like to verify that the Data Checks still function after the modification to `infer_feature_types` so that I can be confident in my changes.

Issue #3060, opened 2021-11-16
Nullable Types: Remove NT Check from infer_feature_types()


chukarsten:
As an EvalML user, I no longer want to raise an exception on detection of the nullable types, so that I can handle them using both our DataChecks and preprocessing components.

PR #3156, opened 2021-12-15
Nullable Types: infer_feature_types modifications

chukarsten:
Addresses: #3060 and #3061

Removing the ignore_nullable_types arg, the nullable types check from infer_feature_types and testing for infer_feature_types to raise an exception when they are detected.


evalml/data_checks/highly_null_data_check.py
```diff
@@ -105,7 +105,7 @@ def validate(self, X, y=None):
         """
         results = {"warnings": [], "errors": [], "actions": []}
 
-        X = infer_feature_types(X, ignore_nullable_types=True)
+        X = infer_feature_types(X)
```
freddyaboulton, line 108: Does this mean that #3061 is also done?

bchen1116 (commented):
Agreed with Freddy! I think this should be done in conjunction with #3061 so that we can ensure this change doesn't break anything for data checks

chukarsten:
@bchen1116 @freddyaboulton I've added additional testing for the DefaultDataChecks suite as well as some of the data checks that explicitly deal with nulls to check that they jibe with the new nullable types.  Let me know what you think!

evalml/utils/woodwork_utils.py
```diff
@@ -10,6 +10,8 @@
     ww.logical_types.Integer.type_string,
     ww.logical_types.Double.type_string,
     ww.logical_types.Boolean.type_string,
+    ww.logical_types.IntegerNullable.type_string,
```
freddyaboulton, line 13: We should probably add Age/AgeNullable here? I wonder if we can get rid of this list though.

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -216,7 +218,13 @@ def test_default_data_checks_classification(input_type):
 
     y = pd.Series([0, 1, np.nan, 1, 0])
     y_multiclass = pd.Series([0, 1, np.nan, 2, 0])
-    X.ww.init(logical_types={"natural_language_nan": "NaturalLanguage"})
+    X.ww.init(
+        logical_types={
```
freddyaboulton, line 222: Nice, I think this covering the default data checks gives us coverage for a lot of the data checks.

freddyaboulton (approved):
@chukarsten This looks good to me! Thanks for adding the coverage for the data checks + nullable types.

PR #3160, opened 2021-12-20
Gap separated training and test

### LINKED ISSUE #3078: Make sure date_index of test set is separated by gap from the training set
Came up during the review of #3071 (comment [here](https://github.com/alteryx/evalml/pull/3071#discussion_r751813099) by @ParthivNaresh ).

We used to validate the holdout set for time series by checking if the pandas indices of the dataframes were separated by  `gap + 1` units. However, this doesn't work for all possible index types like str and the woodwork index type. We should check that the `date_index` is `gap` + 1 units away before generating features and predicting. That will always be a datetime type and so the arithmetic is well defined.

ParthivNaresh:
Fixes #3078 

-----


evalml/demos/weather.py
```diff
@@ -15,4 +17,18 @@ def load_weather():
         + evalml.__version__
     )
     X, y = load_data(filename, index=None, target="Temp")
```
ParthivNaresh, line 19: There were two dates missing from this dataset which prevented this new implementation from working properly. Because we no longer rely on the index to determine validity of the training/test separation, the `time_index` column in X has to be as we expect it to be (ordered, interval has to be detected, datetime based) for the new tools in woodwork 0.10.0 to be usable since the `DateTimeFormatDataCheck` isn't run in most use cases.

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -80,20 +80,37 @@ def _move_index_forward(index, gap):
...
+        train_copy = train.copy()
+        test_copy = test.copy()
+        train_copy.ww.init(time_index=self.time_index)
+        test_copy.ww.init(time_index=self.time_index)
+
+        X_frequency_dict = train_copy.ww.infer_temporal_frequencies(
+            temporal_columns=[train_copy.ww.time_index]
+        )
+        freq = X_frequency_dict[test_copy.ww.time_index]
+        if freq is None:
```
ParthivNaresh, line 100: This was added regarding a conversation with @freddyaboulton. He raised the valid point that we don't want AutoMLSearch to pass only to fail when a user tries and uses `predict`. In this case, AutoMLSearch will raise this error if the `time_index` does not have values that can be inferred.
I'm up for a discussion regarding this since I know it touches a lot of time series and is essentially the first time that our whole algorithm will be directly exposed to a proper interval frequency check on time series (outside of DateTimeFormatDataCheck).

evalml/tests/conftest.py
```diff
@@ -860,11 +860,11 @@ def __init__(
 
 
 @pytest.fixture
-def dummy_ts_binary_linear_classifier_pipeline_class():
-    log_reg_classifier = LogisticRegressionClassifier
+def dummy_ts_binary_tree_classifier_pipeline_class():
```
bchen1116, line 863: Why did we need to change the estimator used here?

evalml/tests/demo_tests/test_datasets.py
```diff
@@ -86,5 +86,25 @@ def test_datasets(dataset_name, expected_shape, local_datasets):
...
     X, y = demo_method
     X_local, y_local = local_datasets[dataset_name]
+
+    if dataset_name == "daily_temp":
+        missing_date_1 = pd.DataFrame([pd.to_datetime("1984-12-31")], columns=["Date"])
+        missing_date_2 = pd.DataFrame([pd.to_datetime("1988-12-31")], columns=["Date"])
+        missing_y_1 = pd.Series([14.5], name="Temp")
+        missing_y_2 = pd.Series([14.5], name="Temp")
+
+        X_local = pd.concat(
```
bchen1116, line 96: Just a suggestion, but what do you think about doing the `X_local` and `y_local` concats in 2 operations rather than 4? For instnace:
```python
X_local = pd.concat(
     [X_local.iloc[:1460], missing_date_1, X_local.iloc[1460: 2920], missing_date_2, X_local.iloc[2920:]]
).reset_index(drop=True)
```

Not a necessary change, but could be a little neater?

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1428,3 +1428,59 @@ def test_time_index_cannot_be_none(time_series_regression_pipeline_class):
...
+)
+@pytest.mark.parametrize(
+    "pipeline_class,estimator_name",
+    [
+        (TimeSeriesRegressionPipeline, "Random Forest Regressor"),
+        (TimeSeriesBinaryClassificationPipeline, "Logistic Regression Classifier"),
+        (TimeSeriesMulticlassClassificationPipeline, "Logistic Regression Classifier"),
+    ],
+)
+def test_noninferrable_data(pipeline_class, estimator_name, gap, reset_index, freq):
```
bchen1116, line 1453: is there test coverage for when the gaps should raise an error/return false?

bchen1116 (changes requested):
Great work here! Left some questions about test coverage that I think need to be addressed first

evalml/tests/conftest.py
ParthivNaresh: The linear regressor doesn't support datetime columns and throws `TypeError: The DType <class 'numpy.dtype[datetime64]'> could not be promoted by <class 'numpy.dtype[float64]'>.`

With `DecisionTreeClassifier`, we can support datetime columns while still choosing an estimator supported with minimum requirements.

evalml/tests/demo_tests/test_datasets.py
ParthivNaresh: Love it

evalml/tests/pipeline_tests/test_time_series_pipeline.py
ParthivNaresh: Are you referring to line [1479](https://github.com/alteryx/evalml/pull/3160/files#diff-1b99a3819e82323aca1e02c75772961e13269542eccb070128773fb5bcc5f024R1479) below? It covers uninferrable holdout data.
The other `ValueError` in `_are_datasets_separated_by_gap_time_index` is covered by frequencies of 1 in this test, where the output frequency is `D` or `M` instead of `1D` or `1M`.
I also went ahead and added to `test_time_series_pipeline_validates_holdout_data` to cover the frequency aspect of the error!

evalml/tests/conftest.py
bchen1116: Thanks for explaining!

PR #3161, opened 2021-12-20
Remove usage of scikit-learn's `LabelEncoder`

### LINKED ISSUE #3132: Remove usage of scikit-learn's `LabelEncoder`
Now that we've implemented our own LabelEncoder, it could be useful to replace usages of scikit-learn LabelEncoder classes used in other parts of our code (ex: CatBoost / XGBoost classifiers) and replace them with our own LabelEncoder instead!

angela97lin:
Closes #3132


evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -167,7 +168,7 @@ def _encode_labels(self, y):
         if not is_integer_dtype(y_encoded):
             self._label_encoder = LabelEncoder()
             y_encoded = pd.Series(
-                self._label_encoder.fit_transform(y_encoded), dtype="int64"
+                self._label_encoder.fit_transform(None, y_encoded)[1], dtype="int64"
```
jeremyliweishih, line 171: do we need the `pd.Series` call here when we don't have it in catboost?

evalml/pipelines/components/estimators/classifiers/xgboost_classifier.py
```diff
@@ -103,7 +103,7 @@ def _convert_bool_to_int(X):
     def _label_encode(self, y):
         if not is_integer_dtype(y):
             self._label_encoder = LabelEncoder()
-            y = pd.Series(self._label_encoder.fit_transform(y), dtype="int64")
+            y = pd.Series(self._label_encoder.fit_transform(None, y)[1], dtype="int64")
```
jeremyliweishih, line 106: Same question here!

jeremyliweishih (approved):
Impeccable work!

bchen1116 (approved):
LGTM! Agreed with @jeremyliweishih, the label encoder component already returns the `y` value as an int-encoded series, so I don't think we need to do the extra initialization and type-casting in lightgbm and xgboost

PR #3162, opened 2021-12-20
Change validation_score and mean_cv_score to properly reflect cv or non-cv cases

### LINKED ISSUE #3097: Rankings: when CV is used, have "validation" score be the CV score
**Background**
We have two scoring columns in our automl rankings, `mean_cv_score` and `validation_score`. Currently, `mean_cv_score` is only populated if CV was run. And `validation_score` is always the validation score from the 1st training-validation split -- if CV was used this is the first CV fold, and if a training-validation split was used this is the validation split from that partitioning.

From [our automl user guide](https://evalml.alteryx.com/en/stable/user_guide/automl.html#View-Rankings):
> A summary of all the pipelines built can be returned as a pandas DataFrame which is sorted by score. The score column contains the average score across all cross-validation folds while the validation_score column is computed from the first cross-validation fold.

**Problem**
If CV was used, it is misleading to report the "validation" score as coming from only one of the CV folds. If that fold for some reason had better or worse performance than the mean, and also different performance vs the same model trained on the entire training data, we'd be giving users a pessimistic or optimistic view of model generalizability.

This came up in discussion with @rpeck this week. 

**Proposed Change**
If the data splitting strategy used is a cross-validation strategy, fill the `validation_score` column with the value from `mean_cv_score`.

Also in this case, use the value from `mean_cv_score` to compute the value for `percent_better_than_baseline`.
Users can still access detailed per-CV-fold info via the `automl.results` structure.

And if the data splitting strategy used is not a cross-validation strategy, the `mean_cv_score` column should be unpopulated. I believe this is our current behavior.

I believe all of the above can be applied for time series

**Impl notes**
I think the class hierarchy of our data splitters can be used to determine whether or not a data splitter counts as a CV splitter, e.g. `issubclass(data_splitter, OurMainCVClass)`. But I'm not sure if that will work for time series without some changes. We could change to use a mixin pattern instead of standard inheritance. We could also flatten the inheritance hierarchy completely and store metadata as class attributes set per-class, similar to how we organize our objectives and the behavior flags supported by each specific objective.

bchen1116:
fix #3097 

USing regular AutoMLSearch on a binary problem (CV folds by default):
image

Using a training-validation split rather than CV folds:
image

Time Series confirmation:
image


evalml/automl/automl_search.py
```diff
@@ -955,7 +955,7 @@ def search(self, show_iteration_plot=True):
             full_rankings = self.full_rankings
             current_batch_idx = full_rankings["id"].isin(new_pipeline_ids)
             current_batch_pipeline_scores = full_rankings[current_batch_idx][
-                "mean_cv_score"
+                "validation_score"
```
bchen1116, line 958: Since `validation_score` now will either hold the validation score of a training-validation split or the mean cv score of the CV folds, it will be the better measure to use for sorting model performance

evalml/automl/automl_search.py
```diff
@@ -1145,7 +1145,11 @@ def _post_evaluation_callback(self, pipeline, evaluation_results, job_log):
         cv_data = evaluation_results["cv_data"]
         cv_scores = evaluation_results["cv_scores"]
         is_baseline = pipeline.model_family == ModelFamily.BASELINE
-        cv_score = cv_scores.mean()
+        if len(cv_scores) == 1:
+            validation_score = cv_scores[0]
+            mean_cv_score = np.nan
```
bchen1116, line 1150: If there is only 1 score, it was a TV split. We then set the `mean_cv_score~ to NaN

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5511,3 +5515,69 @@ def test_automl_passes_known_in_advance_pipeline_parameters_to_all_pipelines(
...
+        "gap": 0,
+        "max_delay": 0,
+        "forecast_horizon": 1,
+    }
+    automl = AutoMLSearch(
+        X_train=X,
+        y_train=y,
+        problem_type="time series binary",
+        max_iterations=3,
+        data_splitter=TimeSeriesSplit(n_splits=3),
```
bchen1116, line 5572: TimeSeriesSplit is based on KFold splitter, which means we cannot do 1 split for this scenario. For time series problems, we currently only have support for CV folds

evalml/automl/automl_search.py
angela97lin: Curious... are there any performance implications to this? Are different pipelines now ranked differently, potentially resulting in different scores? Might be over the top but could be good to have perf test results just to see changes to rankings, if any :)

bchen1116: @angela97lin Pipelines shouldn't be ranked differently! `validation_score` now will hold what `mean_cv_score` previously held, which would be the the mean score in CV folds and the validation score in TV splits. The main difference, now, is that `mean_cv_score` will be `NaN` when there are no CV folds, which means that we need to use `validation_score` to sort the pipelines

angela97lin: I see, thank you for explaining, I understand better now 👍

evalml/tests/automl_tests/test_automl.py
ParthivNaresh: I just wanted to verify what this meant, we can still pass in just a `TrainingValidationSplit` for time series can't we?

Screen Shot 2021-12-22 at 4 41 01 PM

bchen1116: Yep! I just wanted to check whether or not the `TimeSeriesSplit` would be able to do a single TV split. Coverage for the `TrainingValidationSplit` is already ensured through the previous test!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -157,7 +157,7 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type, obje
         assert automl.get_pipeline(pipeline_id).parameters == results["parameters"]
         assert (
             results["validation_score"]
-            == pd.Series([fold["mean_cv_score"] for fold in results["cv_data"]])[0]
+            == pd.Series([fold["mean_cv_score"] for fold in results["cv_data"]]).mean()
```
angela97lin, line 160: 👍

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1165,7 +1168,8 @@ def test_add_to_rankings_regression_large(
         automl.add_to_rankings(dummy_regression_pipeline_class({}))
         assert isinstance(automl.data_splitter, TrainingValidationSplit)
         assert len(automl.rankings) == 1
-        assert 0.1234 in automl.rankings["mean_cv_score"].values
+        assert 0.1234 not in automl.rankings["mean_cv_score"].values
+        assert np.isnan(automl.rankings["mean_cv_score"].values[0])
```
angela97lin, line 1172: This suffices right? Do we still need the line above?

angela97lin (approved):
LGTM, thank you @bchen1116!

PR #3166, opened 2021-12-21
Update `nlp-primitives` version to 2.1.0

angela97lin:
Need to wait for `nlp-primitives` to upload to conda but requesting review so we can press go when it's out: https://anaconda.org/conda-forge/nlp-primitives

Up for debate: do we *need* to update our min version of `nlp-primitives` to 2.1.0? Technically, our CI will pass even if we don't (since it'll install the latest). Doing it here because I think it'd be safer / in case users get unexpected behavior with different versions of nlp-primitives and nltk.

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -40,7 +40,7 @@ outputs:
         - texttable >=1.6.2
         - woodwork >=0.8.2
         - featuretools>=1.2.0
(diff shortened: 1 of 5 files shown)


freddyaboulton (approved):
Thank you @angela97lin !!

Issue #2867, opened 2021-10-01
Make DefaultAlgorithm the default AutoML Algorithm


jeremyliweishih:
Flip the switch!

- check AutoMLSearch output
- check test coverage

Some options on logic tying `AutoMLSearch` to `IterativeAlgorithm`

- rip out old logic pertaining to `IterativeAlgorithm`
- standardize API between `IterativeAlgorithm` and `DefaultAlgorithm`

jeremyliweishih:
The switch to make `DefaultAlgorithm` the default is hinging on many automl tests that assume logic from `IterativeAlgorithm`. After discussion with the team, the plan is to remove such logic from these tests and I will be working on #2868 concurrently as well!

PR #3177, opened 2022-01-03
Support time series in `DefaultAlgorithm`

### LINKED ISSUE #2691: Support time series in new automl algo

jeremyliweishih:
Fixes #2691.

Performance test results:
[ts_comp.html.zip](https://github.com/alteryx/evalml/files/7830748/ts_comp.html.zip)

Results look good and matches up with past Default tests. This case is even simpler in that the pipelines are exactly the same, but just in a different search order. You can see that the only difference is very minor (<5%) variations in estimator performance in change in fit time due to having 2 additional pipelines searched in Default.


evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -381,3 +382,106 @@ def test_select_cat_cols(
...
+    algo = DefaultAlgorithm(
+        X, y, problem_type, sampler_name, pipeline_params=pipeline_params
+    )
+    naive_model_families = set([ModelFamily.LINEAR_MODEL, ModelFamily.RANDOM_FOREST])
+
+    first_batch = algo.next_batch()
+    assert len(first_batch) == 2
+    assert {p.model_family for p in first_batch} == naive_model_families
+    for pipeline in first_batch:
+        assert pipeline.parameters["pipeline"] == pipeline_params["pipeline"]
```
jeremyliweishih, line 428: add asserts for KIA as well.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -314,22 +356,36 @@ def next_batch(self):
...
+                next_batch = self._create_ensemble()
+            else:
+                next_batch = self._create_n_pipelines(
+                    self._top_n_pipelines, self.num_long_pipelines_per_batch
+                )
         else:
-            next_batch = self._create_n_pipelines(
-                self._top_n_pipelines, self.num_long_pipelines_per_batch
-            )
+            if self._batch_number == 0:
```
jeremyliweishih, line 377: This means `DefaultAlgorithm` will have different batches for TS and non-TS problems as ensembling is turned off for TS.  Moreover, the pipeline structures are different for TS problems as well (we defer pipeline creation to `_make_pipeline_time_series` which incorporates the KIA work @freddyaboulton did). For the future but this begs the question: should we have a separate automl algorithm implementation for TS problems vs non-TS problems? I can envision the batch structure being the same (once ensembling for TS is fixed) but with a different implementation (pipelines are different etc.).

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -381,3 +383,193 @@ def test_select_cat_cols(
...
+@pytest.mark.parametrize(
+    "automl_type",
+    [
+        ProblemTypes.TIME_SERIES_BINARY,
+        ProblemTypes.TIME_SERIES_MULTICLASS,
+        ProblemTypes.TIME_SERIES_REGRESSION,
+    ],
+)
+@patch("evalml.pipelines.components.FeatureSelector.get_names")
+def test_default_algorithm_time_series(
```
jeremyliweishih, line 397: Once #2867 goes in, any AutoML test that used `IterativeAlgorithm` will use `DefaultAlgorithm` instead. So I won't duplicate coverage for those tests in this PR.

evalml/tests/integration_tests/test_nullable_types.py
```diff
@@ -56,7 +56,7 @@ def test_nullable_types_builds_pipelines(
     if automl_algorithm == "iterative":
         pipelines = [pl.name for pl in automl.allowed_pipelines]
     elif automl_algorithm == "default":
-        # TODO: Upon resolution of GH Issue #2691, increase the num of batches.
+        # TODO: Upon resolution of GH Issue #3186, increase the num of batches.
```
jeremyliweishih, line 59: Changed to the new issue I set up to track adding ensembling support for TS problems.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         self.verbose = verbose
         self._selected_cat_cols = []
         self._split = False
+        self._ensembling = True if not is_time_series(self.problem_type) else False
```
jeremyliweishih, line 113: I can add ensembling as a parameter to `DefaultAlgorithm` as well but currently I don't believe we have a use case for it.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -257,12 +265,28 @@ def _create_fast_final(self):
             if estimator not in self._naive_estimators()
         ]
 
-        pipelines = [
-            self._make_split_pipeline(
-                estimator,
-            )
-            for estimator in estimators
-        ]
+        if is_time_series(self.problem_type):
```
jeremyliweishih, line 268: Theres a couple `is_time_series` scattered in this PR and I like how explicit it is (although it does make this code more verbose). An alternative would be to move `is_time_series` into `_make_split_pipeline`.

evalml/automl/automl_algorithm/default_algorithm.py
freddyaboulton: If the mechanics of this algorithm for non-time series pipelines yield good results when applied to time series pipelines without violating any "rules of machine learning", e.g. data-leakage etc, then we should use the same algorithm.

I suspect that will be the case when we get ensembling to work for time series.

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -381,3 +383,193 @@ def test_select_cat_cols(
...
+@pytest.mark.parametrize(
+    "automl_type",
+    [
+        ProblemTypes.TIME_SERIES_BINARY,
+        ProblemTypes.TIME_SERIES_MULTICLASS,
+        ProblemTypes.TIME_SERIES_REGRESSION,
+    ],
+)
+@patch("evalml.pipelines.components.FeatureSelector.get_names")
+def test_default_algorithm_time_series_known_in_advance(
```
chukarsten, line 478: Perhaps these two tests could be combined by parameterizing over KIA?

chukarsten (approved):
Nice work, Jeremy.  Nothing blocking from me, just some refactor suggestions.  I think we should try to remove some of the duplicate code if the complexity isn't unmanageable!  I'll leave it up to you which suggestions to take and which to pass on!

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -304,7 +328,25 @@ def _create_long_exploration(self, n):
...
+                    known_in_advance=self._pipeline_params.get("pipeline", {}).get(
+                        "known_in_advance", None
+                    ),
+                )
+                for estimator in estimators
+            ]
+        else:
+            pipelines = [
+                self._make_split_pipeline(estimator) for estimator in estimators
+            ]
```
bchen1116, line 349: Noticed that these two chunks of code are the same (for `_create_fast_final` and `_create_long_exploration`) beyond the estimators used. Wondering if it'd be useful to take this chunk and make it a helper function that we can then call and pass in the estimators to.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -314,22 +356,36 @@ def next_batch(self):
...
+                next_batch = self._create_naive_pipelines()
+            elif self._batch_number == 1:
+                next_batch = self._create_naive_pipelines(use_features=True)
+            elif self._batch_number == 2:
+                next_batch = self._create_fast_final()
+            elif self.batch_number == 3:
+                next_batch = self._create_ensemble()
+            elif self.batch_number == 4:
+                next_batch = self._create_long_exploration(n=self.top_n)
+            elif self.batch_number % 2 != 0:
```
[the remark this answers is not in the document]
bchen1116: The 3rd batch is also ensembling. The current logic looks fine to me!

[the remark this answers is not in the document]
jeremyliweishih: there's a ensembling batch in batch 4 as well but I'll try to think of a way to clean this up as well.

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -381,3 +383,193 @@ def test_select_cat_cols(
...
+        X, y, problem_type, sampler_name, pipeline_params=pipeline_params
+    )
+    naive_model_families = set([ModelFamily.LINEAR_MODEL, ModelFamily.RANDOM_FOREST])
+
+    first_batch = algo.next_batch()
+    assert len(first_batch) == 2
+    assert {p.model_family for p in first_batch} == naive_model_families
+    for pipeline in first_batch:
+        assert pipeline.parameters["pipeline"] == pipeline_params["pipeline"]
+        assert pipeline.parameters["DateTime Featurization Component"]["time_index"]
```
bchen1116, line 430: As a heads up, since [3192](https://github.com/alteryx/evalml/pull/3192) is in, this component is now called `DateTime Featurizer`.

bchen1116 (approved):
LGTM! I left a few suggestions on ways to clean up code, but nothing blocking!

evalml/tests/automl_tests/test_default_algorithm.py
jeremyliweishih: ah thank you, will update!

Issue #968, opened 2020-07-23
Pipelines: run data checks on fit data


freddyaboulton:
As a user of evalml *who is training pipelines outside of automl*, I would like pipelines to catch any data errors before I call `fit`.

I think this can be accomplished by letting `PipelineBase` accept a list of `DataCheck`s on `__init__`. It would be cool if pipelines returned by `AutoMLSearch.get_pipeline` already have data checks passed in to them. This would also make sure all our validation logic lives in the same place, which would be easy to maintain.

dsherry:
@freddyaboulton is this still relevant?

ParthivNaresh:
Based on our conversation in OH, is this issue still relevant or does a broader discussion on the philosophy/intention of Data Checks need to take place?

freddyaboulton:
@ParthivNaresh I think we can close it out! We'll only throw exceptions for egregious problems (which is a subset of what our data checks catch) which is incompatible with this feature request. I think there's a challenge in making sure there isn't a lot of code duplication between data checks and pipeline code but that's not suited for this issue.

PR #3185, opened 2022-01-05
Limit Classification problems to appropriate target data

### LINKED ISSUE #2975: Limit BinaryClassificationPipeline to binary problems only.
Repro:
```python

X, y = load_diabetes()
bcp = BinaryClassificationPipeline(["Simple Imputer", "Logistic Regression Classifier"])

bcp.fit(X, y)
```
This also works with `load_wine`, a multiclass dataset.

Should this be allowed? This could introduce wild bugs with `optimize_thresholds`, and it seems like we should enforce the data to be binary for the binary pipelines.

ParthivNaresh:
Fixes #2975 

The scope of this issue was expanded to ensure that the appropriate number of classes exist for binary and multiclass pipelines.


evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -67,8 +68,18 @@ def fit(self, X, y):
 
         Returns:
             self
+
+        Raises:
+            ValueError: If the target is not numeric.
```
eccabay, line 73: This should be more specific - the target doesn't need to be numeric in all cases, just regression cases, right?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1981,12 +1981,15 @@ def test_percent_better_than_baseline_in_rankings(
     dummy_regression_pipeline_class,
     dummy_time_series_regression_pipeline_class,
     ts_data_binary,
+    X_y_multi,
 ):
     if not objective.is_defined_for_problem_type(problem_type_value):
         pytest.skip("Skipping because objective is not defined for problem type")
 
     # Ok to only use binary labels since score and fit methods are mocked
```
eccabay, line 1989: This comment no longer applies

evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -1639,8 +1641,13 @@ def __init__(self, parameters, random_seed=0):
...
 
     # Test fitted equality
-    X = pd.DataFrame({})
-    y = pd.Series([])
+    if pipeline_class == BinaryClassificationPipeline:
+        X, y = X_y_binary
+    elif pipeline_class == MulticlassClassificationPipeline:
+        X, y = X_y_multi
+    else:
+        X, y = X_y_regression
```
eccabay, line 1649: With the number of times you've had to add this logic to tests, it might be worth it to add a util function somewhere to grab the test data based on the problem type.

eccabay (commented):
Good stuff! Just left a few cleanup notes

evalml/pipelines/classification_pipeline.py
```diff
@@ -50,9 +51,20 @@ def fit(self, X, y):
...
         """
         X = infer_feature_types(X)
         y = infer_feature_types(y)
+
+        if self.problem_type == ProblemTypes.BINARY and y.nunique() != 2:
+            raise ValueError(f"Binary pipelines require y to have 2 unique classes!")
+        elif self.problem_type == ProblemTypes.MULTICLASS and y.nunique() in [1, 2]:
+            raise ValueError(
+                "Multiclass pipelines require y to have 3 or more unique classes!"
+            )
```
chukarsten, line 66: I think I asked this on your previous PR, but is there anyway we can tie this to the logic in `invalid_target_data_check`?  That data check, albeit somewhat complex, does check for this.  But at the same time, this is starting to feel weird, adding bits and pieces of certain data checks back into the actual pipelines.  @angela97lin , any thoughts?  Are we moving in a direction contrary to the philosophy we had in mind for data checks?

evalml/pipelines/time_series_classification_pipelines.py
```diff
@@ -40,8 +40,24 @@ def fit(self, X, y):
 
         Returns:
             self
+
+        Raises:
+            ValueError: If the number of unique classes in y are not appropriate for the type of pipeline.
         """
         X, y = self._convert_to_woodwork(X, y)
+
```
chukarsten, line 48: Isn't this exactly the same as `ClassificationPipeline.fit()`?  I feel like we can actually just get rid of the time series fit, unless there's currently a compelling reason to override it with the same thing? @freddyaboulton , did you have something in mind?

evalml/tests/pipeline_tests/test_pipelines.py
chukarsten: yes

chukarsten (changes requested):
Thanks for tackling this one, Parthiv!  I am gonna block based on the duplication of the data checks/actions logic.  I am not sure what we want to do with this.  The situation is kind of the same for the now reverted #3160 .  What do we want to do when we're testing for something that a data check already tests for?  I suspect we're going to want to refactor the datachecks in such a way that we can use the same logic.  Let's talk about it in OH Monday?

evalml/pipelines/classification_pipeline.py
angela97lin: I don't think that's a bad idea at all @chukarsten, especially since we already have a check for whether or not a target has enough instances of a class in a classification problem--it seems to fit the philosophy for data checks quite well.

Pretty interesting problem, maybe this is worth a discussion with the larger group? I wonder if we need to better draw the line between when we want to throw an exception to strongly disallow something vs use data checks.

evalml/pipelines/time_series_pipeline_base.py
ParthivNaresh: Yes specifically just time series regression in this function

evalml/pipelines/classification_pipeline.py
freddyaboulton: Calling attention to this issue a filed forever ago that I think tackles the same problem https://github.com/alteryx/evalml/issues/968

Basically, let pipelines accept data checks on init so that we can run data checks prior to fit. 

I think that's just one particular solution to the problem at hand but I think it's worth considering.

bchen1116 (approved):
LGTM! Left a nit. Agreed with Karsten, I think it would be useful to see if there was a way to solve this without duplicating code that already exists in our data check

chukarsten (approved):
LGTM (Looks Gucci to me)

PR #3187, opened 2022-01-05
Add Support for Bool Features for ARIMA

### LINKED ISSUE #3181: ARIMA cannot handle boolean features
Repro

```python
from evalml.demos import load_weather
import pandas as pd
import pytest

X, y = load_weather()

X.ww['bool_1'] = pd.Series([True, False]).sample(n=X.shape[0], replace=True, random_state=0).reset_index(drop=True)
X.ww['bool_2'] = pd.Series([True, False]).sample(n=X.shape[0], replace=True, random_state=1).reset_index(drop=True)

ar = ARIMARegressor(trace=1, error_action='trace')

with pytest.raises(TypeError, match="numpy boolean subtract, the `-` operator, is not supported"):
    ar.fit(X, y)

X.ww['float_1'] = pd.Series([1.2, 2.3, 3.14]).sample(n=X.shape[0], replace=True, random_state=0).reset_index(drop=True)

with pytest.raises(ValueError, match="Could not successfully fit a viable ARIMA model to input data."):
    ar.fit(X, y)
    
X.ww.set_types({"bool_1": "Double", "bool_2": "Double"})
ar.fit(X, y)
```

We should convert boolean to double in our arima component.

freddyaboulton:
### Pull Request Description
Fixes #3181

Perf tests: 
[report.html.zip](https://github.com/alteryx/evalml/files/7830288/report.html.zip)

-----

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -99,13 +99,13 @@ class ARIMARegressor(Estimator):
     def _remove_datetime(self, data, features=False):
         if data is None:
             return None
-        data_no_dt = data.copy()
+        data_no_dt = data.ww.copy()
         if isinstance(
             data_no_dt.index, (pd.DatetimeIndex, pd.PeriodIndex, pd.IntervalIndex)
         ):
-            data_no_dt = data_no_dt.reset_index(drop=True)
+            data_no_dt = data_no_dt.ww.reset_index(drop=True)
         if features:
-            data_no_dt = data_no_dt.select_dtypes(exclude=["datetime64"])
+            data_no_dt = data_no_dt.ww.select(exclude=["Datetime"])
 
         return data_no_dt
 
@@ -141,6 +141,13 @@ class ARIMARegressor(Estimator):
             raise ValueError("ARIMA Regressor requires y as input.")
 
         X = self._remove_datetime(X, features=True)
+        if X is not None:
+            X.ww.set_types(
+                {
+                    col: "Double"
+                    for col in X.ww.select(["Boolean"], return_schema=True).columns
+                }
+            )
         y = self._remove_datetime(y)
         X, y = self._match_indices(X, y)
 
@@ -165,7 +172,13 @@ class ARIMARegressor(Estimator):
         """
         X, y = self._manage_woodwork(X, y)
         fh_ = self._set_forecast(X)
-        X = X.select_dtypes(exclude=["datetime64"])
+        X = X.ww.select(exclude=["Datetime"])
+        X.ww.set_types(
+            {
+                col: "Double"
+                for col in X.ww.select(["Boolean"], return_schema=True).columns
+            }
+        )
 
         if not X.empty:
             y_pred = self._component_obj.predict(fh=fh_, X=X)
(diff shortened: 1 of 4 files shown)


evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -165,7 +172,13 @@ def predict(self, X, y=None):
         """
         X, y = self._manage_woodwork(X, y)
         fh_ = self._set_forecast(X)
-        X = X.select_dtypes(exclude=["datetime64"])
+        X = X.ww.select(exclude=["Datetime"])
```
jeremyliweishih, line 175: should we call `self._remove_datetime` here as well? Unsure if we need the other functionaility in `self._remove_datetime`

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -269,3 +269,35 @@ def test_different_time_units_out_of_sample(
...
+        .reset_index(drop=True)
+    )
+    y = pd.Series(range(10))
+
+    ar = ARIMARegressor(time_index="dates")
+
+    ar._component_obj = MagicMock()
+    ar.fit(X, y)
+
+    pd.testing.assert_series_equal(
```
jeremyliweishih, line 294: very cool!

jeremyliweishih (approved):
Looks great to me! Just one question.

bchen1116 (approved):
LGTM!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
freddyaboulton: Great question! I think we could but that would cause an extra copy so I'd rather keep as is?

jeremyliweishih: sounds good to me!

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -269,3 +269,35 @@ def test_different_time_units_out_of_sample(
...
+        ar._component_obj.fit.call_args[1]["X"]["bool_1"], X["bool_1"].astype(float)
+    )
+    pd.testing.assert_series_equal(
+        ar._component_obj.fit.call_args[1]["X"]["bool_2"], X["bool_2"].astype(float)
+    )
+
+    ar = ARIMARegressor(time_index="dates")
```
angela97lin, line 301: just wondering-- `ar._component_obj = MagicMock()` no longer applies to this line, right? Is that intentional? Are we just checking that we can actually fit/predict?

If so, should we also check that predict doesn't return nan values?

angela97lin (approved):
LGTM, thank you @freddyaboulton ! Just left some comments on testing I was curious about :)

evalml/tests/component_tests/test_arima_regressor.py
freddyaboulton: Yea the intention is to check that we can call non-mocked fit/predict on data with bools and not error out. I will add a comment and check for non-nan!

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -269,3 +269,35 @@ def test_different_time_units_out_of_sample(
...
     assert (y_pred_sk.values == y_pred.values).all()
     assert y_pred.index.equals(X[15:].index)
+
+
+def test_arima_supports_boolean_features():
+    X = pd.DataFrame({"dates": pd.date_range("2021-01-01", periods=10)})
+    X.ww.init()
+    X.ww["bool_1"] = (
+        pd.Series([True, False])
+        .sample(n=10, replace=True, random_state=0)
```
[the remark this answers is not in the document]
freddyaboulton: We technically do not 😅

PR #3189, opened 2022-01-06
Fix nightlies

freddyaboulton:
### Pull Request Description
Fix broken nightlies

-----

### DIFF
--- a/.github/workflows/linux_nightlies.yml
+++ b/.github/workflows/linux_nightlies.yml
@@ -3,6 +3,8 @@ name: Nightly unit tests, linux
 on:
   schedule:
       - cron: '0 7 * * *'
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   unit_tests:
@@ -22,7 +24,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: ${{ github.repository }}
-          ref: main
+          ref: fix-windows-3.9-nightlies-ets
       - name: Update apt and install Graphviz
         run: sudo apt update && sudo apt install -y graphviz
       - name: Create virtual environment, upgrade pip
--- a/.github/workflows/linux_nightlies.yml
+++ b/.github/workflows/linux_nightlies.yml
@@ -56,15 +56,15 @@ jobs:
         run: |
           source test_python/bin/activate
           make ${{matrix.command}}
-      - name: Notify on Slack
-        uses: 8398a7/action-slack@v3
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-        with:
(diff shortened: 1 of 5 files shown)


freddyaboulton:
Might be good to get this in before release/EOD so that the nightlies don't fail again tonight.

.github/workflows/linux_nightlies.yml
```diff
@@ -3,6 +3,8 @@ name: Nightly unit tests, linux
 on:
   schedule:
       - cron: '0 7 * * *'
+  pull_request:
```
freddyaboulton, line 6: Will revert changes in this file prior to merge.

.github/workflows/linux_nightlies.yml
```diff
@@ -54,15 +56,15 @@ jobs:
...
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-        with:
-          job_name: Nightly ${{ matrix.python_version }} ${{matrix.command}}
-          status: ${{ job.status }}
-          fields: workflow,job,took
-          mention: channel
-          if_mention: failure,cancelled
-          text: ':elmofire:'
-        if: failure()
+#      - name: Notify on Slack
```
angela97lin, line 59: I'm guessing this file will be reverted? 👀

evalml/tests/component_tests/test_utils.py
```diff
@@ -115,10 +115,7 @@
     ]
 )
 not_supported_in_linux_py39 = set(
-    [
-        "ARIMA Regressor",
-        "Polynomial Detrender",
-    ]
+    ["ARIMA Regressor", "Polynomial Detrender", "Exponential Smoothing Regressor"]
```
angela97lin, line 118: Hmmm, just curious on your thoughts @freddyaboulton: is there something we could have done to better catch this without using the nightlies? I guess not since we don't run 3.9 on each commit...

angela97lin: I am happy that the error was pretty clear in the failure though: https://github.com/alteryx/evalml/runs/4728893144

angela97lin (approved):
Thank you @freddyaboulton 🙏

jeremyliweishih (approved):

evalml/tests/component_tests/test_utils.py
freddyaboulton: Yea maybe we should have expected the nightlies/conda to fail when we add an sktime component but ultimately I think it's ok that this was caught after merge during the nightlies. We decided to only test 3.9 during the nightlies to save on developer time for 95% of prs that don't exhibit any behavior between python < 3.9 and python 3.9 so fixing stuff is just part of the trade-off.

Issue #2055, opened 2021-03-29
Imputer modifies user data when user passes in a DataTable


freddyaboulton:
```python
import woodwork as ww
import pandas as pd
import numpy as np
from evalml.pipelines.components import Imputer
df = ww.DataTable(pd.DataFrame({
        "all nan": [np.nan, np.nan, np.nan, np.nan, np.nan],
        "all nan cat": pd.Series([np.nan, np.nan, np.nan, np.nan, np.nan], dtype='category')
    }))
X = Imputer().fit_transform(df)
assert df.to_dataframe().empty
```

This came up during #2018 . The imputer is expected to drop all null columns but, as a user, I wouldn't expect the `Imputer` to modify the data pass in.

The underlying issue is that `infer_feature_types` does not copy the data when users pass in a data table.

gsheni:
I would caution about copy the user's data. If the user has a large data set, the copying might be expensive. 

For Featuretools, we modify the user dataframe when inputted to an Entity.

dsherry:
Let's do the copy for now. I agree this has performance implications, but its important to keep our API contract clear.

gsheni:
Since https://github.com/alteryx/evalml/issues/2751 was merged in, can we close out this issue?

PR #3196, opened 2022-01-07
Remove all nan to double conversion in infer_feature_types

### LINKED ISSUE #3194: Cant predict on single-row dataframes with Unknown features
Repro

```python
import pandas as pd
import woodwork as ww
import pytest 

X = pd.DataFrame({"a": [1, 2, 3],
                  "b": [4, 5, None],
                   "c": [7, 8, 9]})
y = pd.Series([1, 2, 3])
X.ww.init(logical_types={"b": "Unknown"})

pipeline = RegressionPipeline(component_graph=["Drop Columns Transformer", "Random Forest Regressor"],
                              parameters={"Drop Columns Transformer": {"columns": ['b']}})
pipeline.fit(X, y)

with pytest.raises(ValueError,
                   match="Input X data types are different from the input types the pipeline was fitted on."):
    pipeline.predict(X.ww.loc[[2]])
```

This is because infer_feature_types changes the logical type of all-nan Unknown columns to Double.

freddyaboulton:
### Pull Request Description
Fixes #3194

Perf tests:
[nan-to-double-reports.zip](https://github.com/alteryx/evalml/files/7831693/nan-to-double-reports.zip)
New branch seems to be faster on large datasets for search and model understanding (nyc-taxi, regress, kddcup)

-----

### DIFF
--- a/evalml/tests/component_tests/test_per_column_imputer.py
+++ b/evalml/tests/component_tests/test_per_column_imputer.py
@@ -213,7 +213,7 @@ def test_fit_transform_drop_all_nan_columns():
         pd.DataFrame(
             {
                 "all_nan": [np.nan, np.nan, np.nan],
-                "some_nan": [0., 1., 0.],
+                "some_nan": [0.0, 1.0, 0.0],
                 "another_col": [0, 1, 2],
             }
         ),
@@ -245,7 +245,7 @@ def test_transform_drop_all_nan_columns():
         pd.DataFrame(
             {
                 "all_nan": [np.nan, np.nan, np.nan],
-                "some_nan": [0., 1., 0.],
+                "some_nan": [0.0, 1.0, 0.0],
                 "another_col": [0, 1, 2],
             }
         ),
@@ -338,8 +338,9 @@ def test_per_column_imputer_impute_all_is_false():
             "column_with_nan_included": "double",
         }
     )
-    X.ww.init(logical_types={"all_nan_included": "Double",
-                             "all_nan_not_included": "Double"})
+    X.ww.init(
+        logical_types={"all_nan_included": "Double", "all_nan_not_included": "Double"}
+    )
     X_t = transformer.fit_transform(X)
     assert_frame_equal(X_expected, X_t)
     assert_frame_equal(
@@ -350,7 +351,7 @@ def test_per_column_imputer_impute_all_is_false():
                 "all_nan_included": [np.nan, np.nan, np.nan],
                 "column_with_nan_not_included": [np.nan, 1, 0],
                 # Because of https://github.com/alteryx/evalml/issues/2055
-                "column_with_nan_included": [0., 1., 0.],
+                "column_with_nan_included": [0.0, 1.0, 0.0],
             }
         ),
     )
--- a/evalml/tests/component_tests/test_per_column_imputer.py
+++ b/evalml/tests/component_tests/test_per_column_imputer.py
@@ -198,6 +198,7 @@ def test_fit_transform_drop_all_nan_columns():
             "another_col": [0, 1, 2],
         }
     )
+    X.ww.init(logical_types={"all_nan": "Double"})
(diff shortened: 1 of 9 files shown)


evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -212,7 +213,7 @@ def test_fit_transform_drop_all_nan_columns():
         pd.DataFrame(
             {
                 "all_nan": [np.nan, np.nan, np.nan],
-                "some_nan": [np.nan, 1, 0],
+                "some_nan": [0.0, 1.0, 0.0],
```
bchen1116, line 216: Why did this need to change? Also, we should rename this column since there are no longer nans

evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -243,7 +245,7 @@ def test_transform_drop_all_nan_columns():
...
```
bchen1116, line 248: same here

evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -344,7 +350,8 @@ def test_per_column_imputer_impute_all_is_false():
                 "all_nan_not_included": [np.nan, np.nan, np.nan],
                 "all_nan_included": [np.nan, np.nan, np.nan],
                 "column_with_nan_not_included": [np.nan, 1, 0],
-                "column_with_nan_included": [0, 1, np.nan],
+                # Because of https://github.com/alteryx/evalml/issues/2055
+                "column_with_nan_included": [0.0, 1.0, 0.0],
```
bchen1116, line 354: rename

bchen1116 (approved):
This looks good to me! I left a few comments on updating column names. 
As for the perf tests, do you have a hypothesis on why some model understanding tests increased a fair bit in time? 
image

Not blocking, but curious as to these slight regressions

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -131,7 +130,7 @@ def test_make_pipeline(
                 if estimator_class.model_family == ModelFamily.LINEAR_MODEL
                 else []
             )
-            drop_null = [DropNullColumns] if "all_null" in column_names else []
+            drop_null = [DropColumns] if "all_null" in column_names else []
```
angela97lin, line 133: This change is because all_null will be considered an unknown column now, rather than double right?

angela97lin (approved):
Looks good, thank you @freddyaboulton!

evalml/tests/pipeline_tests/test_pipeline_utils.py
freddyaboulton: Yes!

evalml/tests/component_tests/test_per_column_imputer.py
freddyaboulton: Sorry I thought I had answered this yesterday.

The only way to pass an all-NaN column to the imputer is if it's not typed as `Unknown`, which is what the default woodwork inference returns. So we need to init woodwork on the dataframe. However, because of https://github.com/alteryx/evalml/issues/2055, the imputer will modify the input data. 

So the `some_nan` column will have the NaNs imputed. We can't change the name of the column because the intention is to compare the input data before and after running through the imputer and the column name doesn't change after going through the imputer.

evalml/tests/component_tests/test_per_column_imputer.py
freddyaboulton: Same comment as above ^

freddyaboulton:
@bchen1116 I actually think the performance on model understanding is faster on this feature branch! See datasets like nyc taxi and regress.csv. The particular dataset you've pointed out is slower but the difference is 0.1 seconds, which I don't think is meaningful.

PR #3197, opened 2022-01-07
Rename `HighlyNullDataCheck` to `NullDataCheck` and update data check to return impute action for non-highly null columns.

### LINKED ISSUE #3144: Update `HighlyNullDataCheck` to detect when columns have any null values
Add new `NullDataCheck` data check that will warn users of any null values in their columns and return an action to impute columns with null values.

angela97lin:
Closes #3144.

- Renames `HighlyNullDataCheck` to `NullDataCheck`
- Previously, `HighlyNullDataCheck` only detected highly null columns (columns that have more than a threshold percentage of null values) and returned a suggested action to drop highly null columns. This PR expands on this data check to detect columns with any null values. For any columns with null values but are *not* considered highly null (to prevent overlap), this data check will now have a warning and suggest an imputation action for those columns.

Note: this might get unwieldy for wide datasets that may just have one or two null values?--More UX to think about?


evalml/data_checks/null_data_check.py
```diff
@@ -180,6 +243,54 @@ def validate(self, X, y=None):
                 ).to_dict()
             )
 
+        if below_highly_null_cols:
+            results["warnings"].append(
+                DataCheckWarning(
+                    message="Columns {} have null values".format(
```
eccabay, line 249: Nit: would it make more sense to say "Column(s)" instead of "Columns"?

evalml/data_checks/null_data_check.py
```diff
@@ -180,6 +243,54 @@ def validate(self, X, y=None):
...
+                col_in_df = X.ww[col]
+                impute_strategies_dict[col] = {
+                    "impute_strategy": {
+                        "categories": ["mean", "mode"]
+                        if col_in_df.ww.schema.is_numeric
+                        else ["mode"],
+                        "type": "category",
+                        "default_value": "mean"
+                        if col_in_df.ww.schema.is_numeric
+                        else "mode",
```
eccabay, line 273: Our linter definitely made this formatting, but it's kind of hard to read. Does it look cleaner/clearer if you pull the if/else logic out of this block?

evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
```diff
@@ -51,7 +51,7 @@ def test_data_checks_suggests_drop_cols():
         }
     )
     y = pd.Series([1, 0, 0, 1, 1])
-    data_check = HighlyNullDataCheck()
+    data_check = NullDataCheck()
```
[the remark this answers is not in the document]
angela97lin: For more clarity: 

If we were to put a debugger in, `actions` would look something like `[{'code': 'DROP_COL', 'data_check_name': 'NullDataCheck', 'metadata': {'columns': ['all_null'], 'rows': None, 'parameters': {}}}, {'code': 'IMPUTE_COL', 'data_check_name': 'NullDataCheck', 'metadata': {'columns': ['lots_of_null'], 'rows': None, 'is_target': False, 'parameters': {'impute_strategies': {'lots_of_null': {'impute_strategy': 'mean'}}}}}]` which has the imputation information we want!

It's `make_pipeline_from_actions` which won't actually convert that action to anything useful in the pipeline right now, since it's not sure what to do with a `IMPUTE_COL` action where `is_target` is False (we currently support just target imputation). It's coming up though!

evalml/data_checks/null_data_check.py
angela97lin: Ah I was trying to make this consistent with the highly null message, but I'll just change both to use "Column(s)" instead 😂

evalml/data_checks/null_data_check.py
angela97lin: Yes, I think so! 😁 
image

evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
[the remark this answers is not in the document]
freddyaboulton: @angela97lin Thanks for the context! Yep I agree the actions are returning the right thing (I'm still not sure about `mode` though) I just thought it was in-scope to modify `make_pipeline_from_actions` in this PR. I'll wait for #3237 !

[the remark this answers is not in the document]
angela97lin: @freddyaboulton Oops totally missed the mode comment, you're right 😝 Let me update that to `most_frequent`!

ParthivNaresh (approved):
Looks good to me!

PR #3208, opened 2022-01-10
Gap separation

ParthivNaresh:
Fixes: #3078


evalml/pipelines/utils.py
```diff
@@ -815,6 +816,75 @@ def make_timeseries_baseline_pipeline(problem_type, gap, forecast_horizon, time_
...
+    test_copy = test.copy()
+    train_copy.ww.init(time_index=pipeline_params["time_index"])
+    test_copy.ww.init(time_index=pipeline_params["time_index"])
+
+    X_frequency_dict = train_copy.ww.infer_temporal_frequencies(
+        temporal_columns=[train_copy.ww.time_index]
+    )
+    freq = X_frequency_dict[test_copy.ww.time_index]
+    if freq is None:
+        return True
```
ParthivNaresh, line 846: Based on the third point of [this](https://github.com/alteryx/evalml/issues/3078#issuecomment-1010307947): If the training data does not have an inferable frequency, let's assume the datasets are correctly separated by the gap for now.

evalml/pipelines/utils.py
```diff
@@ -815,6 +816,75 @@ def make_timeseries_baseline_pipeline(problem_type, gap, forecast_horizon, time_
...
+    """
+    forecast_horizon = pipeline_params["forecast_horizon"]
+    gap = pipeline_params["gap"]
+    time_index = pipeline_params["time_index"]
+    right_length = len(X) <= forecast_horizon
+    X_separated_by_gap = are_datasets_separated_by_gap_time_index(
+        X_train, X, pipeline_params
+    )
+    if not (right_length and X_separated_by_gap):
+        raise PartialDependenceError(
```
ParthivNaresh, line 878: Based on the second point of [this](https://github.com/alteryx/evalml/issues/3078#issuecomment-1010307947): This helper function will return "something" if the test data violates our constraints. Users can then use this "something" to display warning messages prior to calling predict.

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -235,7 +200,6 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
         X.index = self._move_index_forward(
             X_train.index[-X.shape[0] :], self.gap + X.shape[0]
         )
-        self._validate_holdout_datasets(X, X_train)
```
ParthivNaresh, line 238: Based on the third point of [this](https://github.com/alteryx/evalml/issues/3078#issuecomment-1010307947): Rather than raising a ValueError in predict, let's refactor that logic into a helper function that could be used before calling predict. predict should no longer raise exceptions if the data violates our constraints.

freddyaboulton (commented):
@ParthivNaresh Thank you for this! I think this is looking good. I made a suggestion to move away from `PartialDependenceError` and `PartialDependenceErrorCode` since this isn't being used in partial dependence. 

Pinging @fjlanasa for visibility as well.

evalml/pipelines/utils.py
fjlanasa: i think that makes sense

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -66,55 +66,20 @@ def _move_index_forward(index, gap):
...
-                f"Data received - Length X: {len(X)}, "
-                f"X index start: {X.index[0]}, X_train index end {X_train.index[-1]}."
-            )
-
     def _add_training_data_to_X_Y(self, X, y, X_train, y_train):
         """Append the training data to the holdout data.
 
         Need to do this so that we have all the data we need to compute lagged features on the holdout set.
         """
+        from evalml.pipelines.utils import (
```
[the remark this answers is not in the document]
ParthivNaresh: We end up running into a circular dependency issue unfortunately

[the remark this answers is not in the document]
freddyaboulton: Gotcha. Let's move it to `gen_utils` then? That's where `are_ts_parameters_valid_for_split` so I think its sensible to include it there.

evalml/tests/demo_tests/test_datasets.py
```diff
@@ -86,5 +86,31 @@ def test_datasets(dataset_name, expected_shape, local_datasets):
...
+        ).reset_index(drop=True)
+        y_local = pd.concat(
+            [
+                y_local.iloc[:1460],
+                missing_y_1,
+                y_local.iloc[1460:2920],
+                missing_y_2,
+                y_local.iloc[2920:],
+            ]
+        ).reset_index(drop=True)
```
chukarsten, line 113: Shouldn't this already be changed with this PR's `.load_weather()` function modifications?

evalml/utils/gen_utils.py
```diff
@@ -578,3 +578,90 @@ def are_ts_parameters_valid_for_split(
...
+
+
+_holdout_validation_result = namedtuple(
+    "TSHoldoutValidationResult",
+    ("is_valid", "error_messages", "error_codes"),
+)
+
+
+def validate_holdout_datasets(X, X_train, pipeline_params):
+    """Validate the holdout datasets match out expectations.
```
chukarsten, line 630: hyper-nit: should out be our?

chukarsten (approved):
This looks good! Just have a question about why we seem to have duplicate modifications of the weather data set and a hypernitpick!

evalml/tests/demo_tests/test_datasets.py
ParthivNaresh: This is changing the `X_local` and `y_local` variables which are read in from the `local_datasets`. They come from `conftest` via `load_data` instead of `load_weather`, so they have to be changed again. I decided to move this modification into that `conftest` function instead to make it look cleaner!

PR #3209, opened 2022-01-10
Fully separate numeric and categorical data in `DefaultAlgorithm`

### LINKED ISSUE #3020: Default Algorithm: fully split numeric and categorical columns in preprocessing split
#2986 introduces a split preprocessing pipeline where categorical columns are treated in a separate chain and numeric columns in another. However, the current behavior is that categorical columns selected by the feature selector is treated by one chain and the numerical columns + non-FS selected categorical columns are treated in another. The non-FS selected categorical columns do not make it to the final estimator as the numeric chain selects the FS selected numeric columns at the end.  This issue tracks only selecting numeric columns in the numeric chain so that we do not introduce unnecessary preprocessing components in the other chain. 

The most logical solution would be to move the numeric `SelectColumns` component to the front but doing so would void the exact reasoning we introduced the split preprocessing pipeline. The main reasoning was that if we move the numeric `SelectColumns` component to the front, like the categorical preprocessing pipeline, we would select all original text/date time columns if their derivatives are chosen by the feature selector. Thus we could do the following:

1. Select numerical columns at the beginning of the numeric chain and verify performance difference in selecting original numerical columns instead of their derivatives.
2.  Add another `SelectColumns` component at the beginning of the numeric chain. However, it is currently difficult to pass parameters to  two of same components using our current pipelines API. Another approach would be to add a type exclusion component (the opposite of `SelectByType`) and exclude all categorical columns.
  

Closing criteria: 
- all categorical columns go into categorical preprocessing chain and all numeric columns go into numeric preprocessing chain

------------------------------------------------------------------------------------
Full context:

1.   We wanted to add FS to pipelines
2.   FS with select columns transformer would error out on categorical columns because it was placed last and not every “category” was passed in (https://github.com/alteryx/evalml/issues/2904)
3.   Tried re-creating the empty columns but scratched that in favor of only selecting available columns. (https://github.com/alteryx/evalml/pull/2912 and https://github.com/alteryx/evalml/pull/2944)
4.   This still left the issue of selecting a couple encoded columns but not the original column
5.   Tried to fix this by moving column selector first but ran into the issue of the column selector not selecting anything as the “selected columns” are post-encoding columns 
6.   Solution to this was to find the original columns and then move column selector to the front (https://github.com/alteryx/evalml/pull/2966)
7.   Couldn’t move selector to the front because it'll drop features created by TextFeaturizer/DateTimeFeaturizer (I think the "select the original feature if any of the sub-features are selected" should only apply to categorical. For time series problems, it probably makes sense to ignore the sub-features created by DateTimeFeaturizer that are NOT selected by FS. For example, if the data is all from the same year, you probably want to drop YEAR(date) or if HOUR(date) turns out to not be informative. By moving FS to the beginning, you'll always select date so all the sub features will always be generated)
8.   Therefore we need the split, and cannot move numeric FS to front for exact same reasoning in the previous point (https://github.com/alteryx/evalml/pull/2986)

jeremyliweishih:
Fixes #3020

Example:
split

You can see that a one hot encoder does not exist in the numeric pipeline. 

Perf test results:
[split_fixed_no_change.html.zip](https://github.com/alteryx/evalml/files/7926857/split_fixed_no_change.html.zip)


evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -152,8 +152,7 @@ class SelectByType(ColumnSelector):
     def __init__(self, column_types=None, random_seed=0, **kwargs):
         parameters = {"column_types": column_types}
         parameters.update(kwargs)
-        Transformer.__init__(
```
jeremyliweishih, line 155: This was causing parameters to disappear. Still slightly unsure of the exact mechanism of how the parameters disappear but my hypothesis is that the `parameters` field in `SelectByType` you can see here is different from the `parameters` field created by this init logic.

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -136,7 +136,7 @@ def _modify_columns(self, cols, X, y=None):
         return X.ww[column_intersection]
 
 
-class SelectByType(ColumnSelector):
+class SelectByType(Transformer):
```
jeremyliweishih, line 139: Decided to just inherit from `Transformer` and override `fit` as well.

evalml/pipelines/utils.py
```diff
@@ -397,8 +397,8 @@ def make_pipeline(
              An empty dictionary or None implies using all default values for component parameters.
          sampler_name (str): The name of the sampler component to add to the pipeline. Only used in classification problems.
              Defaults to None
-         extra_components (list[ComponentBase]): List of extra components to be added after preprocessing components. Defaults to None.
-         extra_components_position (str): Where to put extra components. Defaults to "before_preprocessing" and any other value will put components after preprocessing components.
+         extra_components_before (list[ComponentBase]): List of extra components to be added before preprocessing components. Defaults to None.
```
jeremyliweishih, line 400: Revised this API here to just have two lists for components in both locations. I was thinking about changing it to a component to index mapping but thought it would be too complicated (I would need to figure out index locations and how they would change as we insert each component).

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -427,17 +430,18 @@ def _transform_parameters(self, pipeline, proposed_parameters):
 
     def _make_split_pipeline(self, estimator, pipeline_name=None):
         numeric_pipeline_parameters = {
-            "Select Columns Transformer": {"columns": self._selected_cols}
+            "Select Columns Transformer": {"columns": self._selected_cols},
+            "Select Columns By Type Transformer": {"column_types": ["numeric"]},
         }
         numeric_pipeline = make_pipeline(
-            self.X,
+            self.X.ww.select(exclude="category"),
```
jeremyliweishih, line 437: here I exclude all `category` semantic tag columns so there won't be preprocessing components for those components.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -447,14 +451,13 @@ def _make_split_pipeline(self, estimator, pipeline_name=None):
                 "Select Columns Transformer": {"columns": self._selected_cat_cols}
             }
             categorical_pipeline = make_pipeline(
-                self.X,
+                self.X.ww.select(include=["category"]),
```
jeremyliweishih, line 454: Likewise I do the opposite here.

evalml/pipelines/utils.py
```diff
@@ -397,8 +397,8 @@ def make_pipeline(
...
+         extra_components_after (list[ComponentBase]): List of extra components to be added before preprocessing components. Defaults to None.
```
chukarsten, line 401: Copy pasta?  Should be after pp_components?

evalml/pipelines/utils.py
chukarsten: I think this is a solid change. I think you should let the use cases drive the API.  Do we currently have (or foresee) a reason to be able to inject components at different locations?

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -167,6 +166,20 @@ def _check_input_for_columns(self, X):
     def _modify_columns(self, cols, X, y=None):
         return X.ww.select(cols)
 
+    def fit(self, X, y=None):
+        """Fits the transformer by checking if column names are present in the dataset.
+
+        Args:
+            X (pd.DataFrame): Data to check.
+            y (pd.Series, optional): Targets.
```
chukarsten, line 174: Hyper-nit: are we indicating y as "ignored" here?

chukarsten (approved):
Looks good, Jeremy!  Just a copy pasta and a nit.  Do with them what you want!

evalml/pipelines/utils.py
jeremyliweishih: I don't see any short term usage other than in `DefaultAlgorithm` but I can imagine sometime in the future we would want more control over where exact components go in our pipelines. I agree that we should let the use cases drive the API so let's leave this for now!

evalml/pipelines/components/transformers/column_selectors.py
jeremyliweishih: Will edit here and above as well.

evalml/pipelines/components/transformers/column_selectors.py
freddyaboulton: It's not that they're disappearing, it's that they're put in the `parameters` dictionary because that's that the `Transformer` init does. 

```
'Numeric Pipeline - Select Columns By Type Transformer': {'columns': None, 'parameters': {'column_types': ['numeric']}, 'component_obj': None}
```

I agree with the redesign. `ColumnSelectors` are meant to take a list of columns so we need to either need to rename `column_types` to `columns` or not make this component a subclass. What we here doing before of trying to skip the parent init and get to the grandparent init is probably not the right thing to do.

Interesting find!

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -484,35 +489,40 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
-            parameters=numeric_pipeline_parameters,
-            extra_components=[SelectColumns],
-            extra_components_position="before_estimator",
-            use_estimator=False if self._selected_cat_cols else True,
-        )
-
         if self._selected_cat_cols:
             self._split = True
+
+            if self._X_with_cat_cols is None or self._X_with_cat_cols is None:
```
freddyaboulton, line 495: @jeremyliweishih I think this is a typo? `self._X_with_cat_cols is None` is repeated twice?

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -484,35 +489,40 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
-            extra_components=[SelectColumns],
-            extra_components_position="before_estimator",
-            use_estimator=False if self._selected_cat_cols else True,
-        )
-
         if self._selected_cat_cols:
             self._split = True
+
+            if self._X_with_cat_cols is None or self._X_without_cat_cols is None:
+                self._X_without_cat_cols = self.X.drop(self._selected_cat_cols, axis=1)
```
freddyaboulton, line 496: @jeremyliweishih I think we should use ww here to select the columns. I think this is causing `make_pipeline` to reinfer the types which could cause problems

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -484,35 +489,40 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
+                self._X_with_cat_cols = self.X[self._selected_cat_cols]
+
+            numeric_pipeline_parameters = {
```
freddyaboulton, line 499: @jeremyliweishih I think numeric vs categorical is not the right distinction. It's categorical vs non-categorical. I think right now we're dropping boolean features (and anything that is not numeric or categorical, e.g. natural language)

```python
import pandas as pd
import woodwork as ww

X, y = load_fraud(1000)

automl = AutoMLSearch(X, y, "binary", verbose=True, _automl_algorithm="default",
                      max_batches=3)
automl.search()

assert X.ww['customer_present'].ww.logical_type.type_string == 'boolean' 
assert 'customer_present' in automl._automl_algorithm._selected_cols
assert not 'custsomer_present' in automl.best_pipeline.transform_all_but_final(X).columns
```

jeremyliweishih: ah good catch, will fix!

jeremyliweishih: @freddyaboulton will rerun perf tests as well!

evalml/pipelines/components/transformers/column_selectors.py
```diff
@@ -149,23 +150,36 @@ class SelectByType(ColumnSelector):
...
             parameters=parameters,
             component_obj=None,
             random_seed=random_seed,
         )
 
     def _check_input_for_columns(self, X):
-        col_types = self.parameters.get("column_types")
-        if col_types and X.ww.select(col_types).empty:
-            raise ValueError("Columns of type {column_types} not found in input data.")
+        pass
```
eccabay, line 163: If this function is not used, why is it necessary? Can we just remove any calls to it?

eccabay (commented):
This is really cool work! Just left a few small nits

ParthivNaresh (approved):
Excellent work, the perf tests look great!

evalml/pipelines/components/transformers/column_selectors.py
jeremyliweishih: Good catch, this was added before inheriting from `Transformer` instead. Will fix!

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -484,37 +495,47 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
+            numeric_pipeline = make_pipeline(
+                self._X_without_cat_cols,
+                self.y,
+                estimator,
+                self.problem_type,
+                sampler_name=self.sampler_name,
+                parameters=numeric_pipeline_parameters,
+                extra_components_before=[SelectByType],
+                extra_components_after=[SelectColumns],
+                use_estimator=False if self._selected_cat_cols else True,
```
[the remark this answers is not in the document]
jeremyliweishih: good catch!

PR #3211, opened 2022-01-11

machineFL:


angela97lin (approved):

angela97lin:
It says I merged the commit because I enabled auto-merge, but it did so when only 27/28 checks passed--I think this is likely because `build_conda_pkg` isn't a requirement to merge? Interesting to note, could lead to cases where we merge in a request that fails `build_conda_pkg` 😬

PR #3212, opened 2022-01-11
Update latest dependency yaml to remove comment about keeping branch and set delete-branch to true

angela97lin:
- Removes the line "Please do not delete the `latest-dep-update` branch because it's needed by the auto-dependency bot." since it doesn't seem to be true. We don't even have that branch available: https://github.com/alteryx/evalml/branches/all?query=latest-dep-update
- Adds a new configuration that auto-deletes the branch when merging or closing the PR so we don't end up with too many dependency update branches that are stale.
- Updates auto approving to run when unit tests. Here's an example in the featuretools repo: https://github.com/alteryx/featuretools/pull/1830

Sample PR looks like: https://github.com/alteryx/evalml/pull/3213. I closed the PR but don't see the branch auto-delete. :( According to [this issue](https://github.com/peter-evans/create-pull-request/issues/936):

```
When delete-branch is set to true the action will delete the branch in two situations:

If the action is closing the pull request itself (because there is no longer any diff)
If the branch remains in the remote, but there are no further changes. In this case, most likely it has been left undeleted after merging.
Your expectation seems to be that the branch should be deleted immediately after merge, but that's not how it works. The action will find and delete it next time it runs.
```


.github/workflows/latest_dependency_checker.yml
```diff
@@ -25,9 +25,9 @@ jobs:
           commit-message: Update latest dependencies
           title: Automated Latest Dependency Updates
           author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
-          body: "This is an auto-generated PR with **latest** dependency updates.
-                 Please do not delete the `latest-dep-update` branch because it's needed by the auto-dependency bot."
+          body: "This is an auto-generated PR with **latest** dependency updates."
           branch: latest-dep-update
           branch-suffix: short-commit-hash
+          delete-branch: true
           base: main
```
gsheni, line 32: can we add assignees?
```suggestion
          base: main
          assignees: machineFL
```

.github/workflows/latest_dependency_checker.yml
```diff
@@ -25,9 +25,9 @@ jobs:
...
           reviewers: angela97lin, dsherry, jeremyliweishih, freddyaboulton, bchen1116, chukarsten, ParthivNaresh
```
gsheni, line 33:
```suggestion
          reviewers: machineAYX
```

PR #3221, opened 2022-01-11
Fix SHAP prediction explanations bug

### LINKED ISSUE #3220: SHAP prediction explanations fail when classification predicts probability of 1
Repro: run `explain_predictions_best_worst` on an estimator that predicts a class with 100% certainty for at least one data point. 

This comes from a SHAP library bug, [issue #2000](https://github.com/slundberg/shap/issues/2000). The logit link function fails because said function is infinite at 0 and 1.

eccabay:
Closes #3220 

Proposed solution is to remove the usage of the logit link function in our explanations altogether. The only difference between the function is what space the final SHAP values are reported in. With the logit function they sit in log-odds space, but between -1 and 1 with the identity function. Due to the lack of significant difference between the two methods, I figure there’s no significant impact in making this change.


freddyaboulton (approved):
Looks good to me @eccabay ! Thank you for the fix. Can you please run the model understanding perf tests to be sure we're not accidentally introducing a regression?

angela97lin (approved):
LGTM, thank you @eccabay! Agreed with Freddy's suggestion about running perf tests, and I'm almost tempted to say that perhaps the identity method could be easier to understand than the logit link function :)

bchen1116 (approved):
Agreed with both above! Nice fix!

chukarsten (approved):
Thank you!

eccabay:
@freddyaboulton great suggestion for perf testing! Results look pretty consistent to me, lmk what you think!
[report.html.zip](https://github.com/alteryx/evalml/files/7857648/report.html.zip)

freddyaboulton:
@eccabay Thanks for running the tests! Send it!

PR #3237, opened 2022-01-13
Update `make_pipeline_from_actions` to handle null column imputation

### LINKED ISSUE #3013: Support per-column imputation via data checks / data check actions
Right now, our data checks will suggest a IMPUTE_COLS action which is only used to impute the target column. We should also support an action that suggests per-column imputation for features with missing values. This will likely require a change in the actions API to store imputation strategy, fill value, and possibly other parameters associated with each column.

### LINKED ISSUE #3144: Update `HighlyNullDataCheck` to detect when columns have any null values

angela97lin:
Closes #3013.


evalml/data_checks/null_data_check.py
```diff
@@ -178,9 +178,19 @@ def validate(self, X, y=None):
         }
 
         X = infer_feature_types(X)
+        X_to_validate = X.ww.select(
```
angela97lin, line 181: Updating this data check to only check categorical + numerical cols (+ unknown which might just be fully null cols).

Reason for this is because two-fold:
1. We have specific data checks such as the `DateTimeNaNDataCheck` data check and the `NaturalLanguageNaNDataCheck` data check for specific types.
2. https://github.com/alteryx/evalml/issues/3241: PerColumnImputer does not currently support datetime columns, so if we have a nan value that gets passed along as a parameter, then we'll error out 😬

angela97lin: Open to discussion though! Seems like we assume natural language columns will be included, given `test_default_data_checks_classification`

docs/source/user_guide.ipynb
```diff
@@ -94,7 +94,7 @@
     }
   },
    "source": [
-    "[Data Check Actions](user_guide/data_actions)"
+    "[Data Check Actions](user_guide/data_check_actions)"
```
angela97lin, line 97: Unrelated, just renaming for clarity :)

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -160,7 +160,7 @@ def validate(self, X, y):
         },
     ).to_dict(),
     DataCheckWarning(
-        message="Column(s) 'lots_of_null', 'nullable_integer', 'nullable_bool', 'natural_language_nan', 'nan_dt_col' have null values",
```
angela97lin, line 163: Checking only a subset of columns means no more detection of datetime cols with nan values, see comment in `evalml/data_checks/null_data_check.py`

evalml/tests/data_checks_tests/test_invalid_target_data_check.py
```diff
@@ -758,7 +758,7 @@ def test_invalid_target_data_check_numeric_binary_does_not_return_warnings():
 
 @pytest.mark.parametrize("use_nullable_types", [True, False])
 @pytest.mark.parametrize("problem_type", ProblemTypes.all_problem_types)
-def test_invalid_target_data_action_for_data_with_null(
+def test_invalid_target_data_check_action_for_data_with_null(
```
angela97lin, line 761: More renaming, unrelated :)

evalml/data_checks/null_data_check.py
freddyaboulton: I thought it was confusing that `NaturalLanguage` is included in `X_to_validate` given your comment that this data check would only check "categorical + numerical cols". 

My thought is that if we allow NaturalLanguage & DateTime in this check and update the action for those columns to only be "drop column" if they are highly null, then we can close out #3240 and #3242 with minimal changes to what you have here. 

Since this is going into the feature branch, I think that change can be made in a follow-up pr though.

evalml/tests/integration_tests/test_data_checks_and_actions_integration.py
```diff
@@ -60,25 +65,37 @@ def test_data_checks_suggests_drop_cols():
...
+                "y",
+            ],
+        },
+        parameters={
+            "Per Column Imputer": {
+                "impute_strategies": {
+                    "null_with_categorical": {"impute_strategy": "most_frequent"},
+                    "lots_of_null": {"impute_strategy": "mean"},
+                },
+                "default_impute_strategy": "most_frequent",
```
freddyaboulton, line 82: Thanks for making the change

freddyaboulton (approved):
@angela97lin Looks good to me! I left a comment about what I think the best way to handle nulls in natural language + datetime columns is. I think that can be done in a follow-up. I will add that comment to #3242

PR #3257, opened 2022-01-19
Bc ensembling caching

### LINKED ISSUE #2835: Speed up ensembler component fit time
This issue tracks finding ways to speed up the time it takes to do the following operations
1. Train an ensembler pipeline during automl
2. Generate predictions from an ensembler pipeline

I think we should focus on the first point first, since it probably has higher time impact. An outcome of this spike could be to file speeding up ensembler predict-time as future work.

Our ideas so far for how to achieve faster ensembler fit-time during automl search:
* Add caching of fitted components to our component graph. This is tracked by #2348 
    * If the input pipelines each use a particular component with the same parameters, fit-caching would mean that component is only fit once with those parameters across the entire ensemble pipeline training step (per training dataset of course).
    * If we implement and calibrate this well, this actually gets us another benefit which is even better: ensembler training during automl would only need to fit the metalearner! This is because the fitted estimators sitting at the end of each input pipeline would all be cached, and could be loaded to generate predictions during TV or CV split 😎 
    * One implication (or future feature request) for #2348 : when ensembling is enabled in automl, that should increment the priority of keeping the pipeline estimators in the fit-cache. Depending on the base caching strategy we choose, this could help us avoid completely missing the benefit of caching! Lots of tricky questions to unpack here, in #2348 :)
* Add caching of component predictions to our component graph. Currently untracked.
    * Similarly to the above, if we implement and calibrate this well, as we train and score the components from each input pipeline on each CV or TV fold, the predictions will be saved in the cache and we'll only have to generate predictions from the metalearner itself!

These optimizations will become important once we update the automl algo to run ensembling more frequently as part of our "long" mode.

chukarsten:
fixes #2835


evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -104,6 +104,9 @@ def _create_ensemble(self):
         problem_type = best_pipelines[0]["pipeline"].problem_type
         n_jobs_ensemble = 1 if self.text_in_ensembling else self.n_jobs
         input_pipelines = []
+        cached_data = {
+            mf: x["cached_data"] for mf, x in self._best_pipeline_info.items()
```
chukarsten, line 108: What are mf and x here?

bchen1116: mf is model family and x is the value associated. The `_best_pipeline_info` dictionary is structured as 
```python
_bpi = {ModelFamily.XGBOOST: {"pipeline": ..., other_info_here}}

```

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -169,7 +168,7 @@ def fit_resample(self, X, y):
             )
 
         y = infer_feature_types(y)
-
+        self.random_state = np.random.RandomState(self.random_seed)
```
bchen1116, line 171: Moved this to make the undersampler deterministic!

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -105,41 +105,20 @@ def _create_ensemble(self):
...
-                    pipeline_params.update(
-                        {
-                            "Select Columns Transformer": {
-                                "columns": self._selected_cat_cols
-                            }
-                        }
-                    )
-            input_pipelines.append(
-                pipeline.new(parameters=pipeline_params, random_seed=self.random_seed)
-            )
```
bchen1116, line 136: Since we are using the already-trained components for ensembling, we no longer need to pass the parameters through. Additionally, if we did, we would lose the trained components when we instantiate a new pipeline.

evalml/automl/engine/engine_base.py
```diff
@@ -157,7 +158,11 @@ def train_pipeline(pipeline, X, y, automl_config, schema=True):
         X,
         y,
     )
-    return cv_pipeline
+    if not get_hashes:
+        return cv_pipeline
+
+    X_hash = hash(tuple(X.index))
```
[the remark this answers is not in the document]
bchen1116: Right, the hash here is now mainly to reduce the memory consumption. If we are passing in data with lots of rows, keeping references to all of the rows for each cv split could end up consuming a fair bit of memory. We can simplify this by hashing, and this hashing call itself is fairly quick, which allows us to easily and conveniently use this.

Will update the docstring though!

evalml/pipelines/component_graph.py
```diff
@@ -414,7 +420,10 @@ def _transform_features(
             self.input_feature_names.update({component_name: list(x_inputs.columns)})
             if isinstance(component_instance, Transformer):
                 if fit:
-                    output = component_instance.fit_transform(x_inputs, y_input)
+                    if not component_instance._is_fitted or self.cached_data is None:
```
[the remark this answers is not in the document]
bchen1116: @chukarsten before introducing a cache, if we call `fit`, we would fit/refit all components in the component graph. There were no scenarios here where calling fit could allow us to skip fitting certain components. However, now, there are things we want to keep in mind:
1) If we call fit and the `component_instance` in question is already fitted but there is no cached data, then we want to refit. This can happen in cases where we pull an already-fitted pipeline and call fit on some `X, y` data. Without knowing whether or not we've fitted originally on this `X,  y` data, we should opt to re-fit this, even if the component instance was already fitted.
2) If the `component_instance` is already fitted and we have cached data, we can instead just rely on the cached data, as long as the hashes match. We only pass in cached data during creation of the ensembler, so this should only trigger when we are using ensembling inside of `AutoMLSearch`. 

The logic above handles these scenarios to ensure that only when we have cached data and have an already-fitted component, can we skip the `fit` call. Otherwise, it's best to call `fit` again to ensure the pipelines can operate as expected.

evalml/pipelines/component_graph.py
```diff
@@ -455,8 +465,20 @@ def _transform_features(
                 else:
                     output = component_instance.predict(x_inputs)
                 output_cache[f"{component_name}.x"] = output
+            if self.cached_data is not None and fit:
+                self.component_instances[component_name] = component_instance
+
         return output_cache
 
+    def _get_stacked_ensemble_component(self, hashes, component_name, fit):
```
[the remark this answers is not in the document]
bchen1116: Currently, I didn't want to emit any warnings since it doesn't seem to be useful to the user. If users got warnings, what would they do with the information? Our component graph would continue to train the components as usual if it wasn't able to find them in the cache. However, it could be useful to log it in the debugger (logger.DEBUG?).

As for `fit` being true for cache retrieval, the logic I've put goes as follows:
- If we are fitting, we try to find the appropriately-trained component in the cache. If it exists, we pull that and insert it into our `component_graph.component_instances` dictionary, which we then use for `predict` and other future calls.
- If we are predicting, we expect the fit call to already have been made. Rather than looking at the cache, the properly-fitted components should then already be in the `component_instances` dictionary, and we can operate through that directly.

evalml/automl/engine/engine_base.py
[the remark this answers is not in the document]
chukarsten: Will it consume a lot of memory?  The index is basically just an array of integers.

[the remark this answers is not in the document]
bchen1116: Mmm, I think it could be substantial if the dataframe passed in is large, ie 100k+ rows.

Python's hashing algo is quite fast, so I feel comfortable relying on this to control our memory usage.

evalml/automl/engine/engine_base.py
```diff
@@ -102,7 +102,23 @@ def submit_scoring_job(
         """Submit job for pipeline scoring."""
 
 
-def train_pipeline(pipeline, X, y, automl_config, schema=True):
+# global hash dictionary to allow minimal calculation of hashes on data splits
```
angela97lin, line 105: Just some food for thought but is there a way to make this not global via refactoring some methods? Seems like we want it mostly for `train_pipeline` and `train_and_score_pipeline` which are primarily called via the Engine classes... could we somehow encapsulate these functions in the Engine classes and then have the hash_dict be a class attribute?

jeremyliweishih: another option is having `AutoMLSearch` hold the cache and pass it as part of `automl_config` or another parameter.

evalml/automl/engine/engine_base.py
```diff
@@ -244,6 +285,7 @@ def train_and_score_pipeline(
                 f"\t\t\tFold {i}: {automl_config.objective.name} score: {scores[automl_config.objective.name]:.3f}"
             )
             score = scores[automl_config.objective.name]
+            pipeline_cache[hashes] = cv_pipeline.component_graph.component_instances
```
angela97lin, line 288: Omega nit-pick since `hashes` is defined within the if/else loop. Can we just call pipeline_cache[hash_dict[i]]? That way, we can remove line 256 too. (we could take this a step further and just pass in get_hashes = i not in hash_dict in the parameters for `train_pipeline` and then set `hash_dict[i] = hashes` when we need to

evalml/pipelines/component_graph.py
jeremyliweishih, line 473: nit: might be misunderstanding here but I think we should rename this method if it's also used to get components other than the stacked ensemble component

evalml/pipelines/component_graph.py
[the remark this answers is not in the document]
angela97lin: So basically we only choose to fit skipping if the instance is fitted AND we have cached data right? Maybe it could be helpful to swap the if/else here to make it more clear what case we're not fitting?

angela97lin (commented):
The performance tests results are super exciting, great work on this @bchen1116! 👏 Just left a couple of smaller questions/comments for now, hoping to take a deeper dive later!

evalml/automl/engine/engine_base.py
bchen1116: I did try that initially! I basically hit a point where I wasn't sure the best way to approach this without completely refactoring the methods and changing the args that we pass to and from the engines to the `train_and_score_pipelines` method. Using global made the code a lot more manageable, which was why I went with this approach. If anyone has suggestions on how I can refactor instead, would love to hear it!

evalml/automl/engine/engine_base.py
```diff
@@ -111,9 +127,11 @@ def train_pipeline(pipeline, X, y, automl_config, schema=True):
         y (pd.Series): Target to train on.
         automl_config (AutoMLSearch): The AutoMLSearch object, used to access config and the error callback.
         schema (bool): Whether to use the schemas for X and y. Defaults to True.
+        get_hashes (bool): Whether to return the hashes of the data used to train (and potentially threshold). Defaults to False
 
     Returns:
         pipeline (PipelineBase): A trained pipeline instance.
+        hash (optional): The hash of the input data indices, only returned when get_hashes is True.
```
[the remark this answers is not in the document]
bchen1116: I did try this and want to say that there was another reason beyond "I broke a lot of tests" that made me revert this change. I don't remember why though, so I will retry this again!

evalml/pipelines/components/transformers/samplers/undersampler.py
bchen1116: We don't need to store it, so I could replace this with 
```python
random_state = np.random.RandomState(...)
```
We use RandomState here to select a random subset of indexes to drop though! Although there are probably other methods of doing the same thing, I found this to be the most convenient method

evalml/automl/engine/engine_base.py
freddyaboulton: @bchen1116 The `hash_dict` is only used to keep track of whether or not we've hashed the data previously right? Is that still a concern now that we're only hashing the index as opposed to the entire data?

I did some back-of-the-envelope calculations and hashing an index of 10 million entries, 3 times for every pipeline, and for 50 pipelines would take 15 seconds. Since this assumes a 3-fold cross-val, this means the total search time increase for 30 million rows is 15 seconds. I think that is a small fraction of time compared to the total search time. The speed up for KDDCup (less than 30 million rows) is much bigger than 15 seconds.

I vote we get rid of this optimization if possible. If we want to keep it for max performance, I think we should brainstorm ways to refactor this. Can we store the `hash_dict` in the engine instance as opposed to AutoMLSearch?

chukarsten (commented):
Great work Bryan, thoroughly tested.  I suspect some of the testing falls more into the integration rather than unit test regime, which is a universal problem for ALL of our testing, not just this.  But it will be hard to address that here without holding things up for a while, and I don't want to do that.  

Just left a few questions/comments here and there.  Nothing really blocking, just a few possible light refactor suggestions for readability as we get fairly deep into if/else branches.

As a side note, I'll have to merge this one since I'm the author 😂

evalml/pipelines/components/transformers/samplers/undersampler.py
```diff
@@ -179,9 +178,7 @@ def fit_resample(self, X, y):
             # iterate through the classes we need to undersample and remove the number of samples we need to remove
             for key, value in result.items():
                 indices = y.index[y == key].values
-                indices_to_remove = self.random_state.choice(
-                    indices, value, replace=False
-                )
+                indices_to_remove = random_state.choice(indices, value, replace=False)
```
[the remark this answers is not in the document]
bchen1116: We don't initialize the `np.random.RandomState` in the initlaization of this component in order to make it deterministic. Since we now initialize it during `fit`, we don't need to save the `np.randomstate` value, so this was for simplification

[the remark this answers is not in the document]
freddyaboulton: Good catch @bchen1116

evalml/pipelines/component_graph.py
```diff
@@ -414,7 +421,10 @@ def _transform_features(
             self.input_feature_names.update({component_name: list(x_inputs.columns)})
             if isinstance(component_instance, Transformer):
                 if fit:
-                    output = component_instance.fit_transform(x_inputs, y_input)
+                    if component_instance._is_fitted and self.cached_data is not None:
```
freddyaboulton, line 424: In what case is the component_instance fitted but cached_data is None? I guess I'm wondering why this isn't just checking `_is_fitted`

evalml/pipelines/component_graph.py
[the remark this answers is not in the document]
freddyaboulton: @bchen1116 

I have a question about 

> This can happen in cases where we pull an already-fitted pipeline and call fit on some X, y data. Without knowing whether or not we've fitted originally on this X, y data

If `get_relevant_component` gets the components from the cache corresponding to the hash of the data, and the `X`, and `y` in question are not in the cache, how will the component instance be fitted?

evalml/pipelines/component_graph.py
```diff
@@ -431,9 +441,10 @@ def _transform_features(
                 output_cache[f"{component_name}.x"] = output_x
                 output_cache[f"{component_name}.y"] = output_y
             else:
-                if fit:
+                if fit and (
+                    not component_instance._is_fitted or self.cached_data is None
```
freddyaboulton, line 445: Same question here about why we need the `self.cached_data is None` check.

I just realized there was some previous discussion [here](https://github.com/alteryx/evalml/pull/3257#discussion_r817108930) so I commented there as well.

freddyaboulton (commented):
@bchen1116 Thanks for doing this great work. Going to take a look at the tests next.

evalml/automl/engine/engine_base.py
```diff
@@ -303,6 +346,7 @@ def train_and_score_pipeline(
             "training_time": training_time,
             "cv_scores": cv_scores,
             "cv_score_mean": cv_score_mean,
+            "cached_data": pipeline_cache,
```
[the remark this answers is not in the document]
bchen1116: Fixing!

evalml/tests/automl_tests/test_automl_algorithm.py
```diff
@@ -48,3 +54,38 @@ def test_automl_algorithm_invalid_pipeline_add(dummy_regression_pipeline):
...
+    pipelines = algo._create_ensemble()[0]
+
+    # check component graph expected cache
+    expected_comp_graph = {
+        "hash1": {
+            "Linear Pipeline - Logistic Regression Classifier": lrc,
+            "Decision Tree Pipeline - Decision Tree Classifier": dtc,
+        }
+    }
+    assert pipelines.component_graph.cached_data == expected_comp_graph
```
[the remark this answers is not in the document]
bchen1116: Yep! This test is to ensure that we are passing the cache properly to the `_make_stacked_ensemble_pipeline` method. There are tests in both the `test_iterative_algorithm` and `test_default_algorithm` that make sure we are caching and passing this cache properly

evalml/pipelines/component_graph.py
bchen1116: Good point! I think I was worried since I no longer pass in newly-initialized pipelines for the stacked ensemble pipeline, but since I'm grabbing the component from the cache only after checking if the cache exists, then this check is repetitive. Removing!

evalml/pipelines/component_graph.py
bchen1116: Addressed in another comment!

evalml/pipelines/component_graph.py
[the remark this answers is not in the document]
bchen1116: Good point! A smooth-brained moment. Had made some code changes that makes this double-check irrelevant

evalml/pipelines/utils.py
```diff
@@ -526,6 +535,21 @@ def _make_new_component_name(model_type, component_name, idx=None):
...
+        # sets the new cached component dictionary using the cached data and model family information
+        if len(cached_data) and model_family in list(cached_data.keys()):
+            for hashes, component_instances in cached_data[model_family].items():
+                if hashes not in list(cached_component_instances.keys()):
+                    cached_component_instances[hashes] = {}
+                try:
+                    cached_component_instances[hashes][
+                        new_component_name
+                    ] = cached_data[model_family][hashes][name]
+                except KeyError:
```
freddyaboulton, line 550: When will there be a key error? That shouldn't happen because the ensemble is made from the pipelines during search so all the hashes should match?

freddyaboulton (commented):
@bchen1116 Tests look good! I have a question on one of the ones that use mocking

evalml/pipelines/utils.py
bchen1116: Yep! there shouldn't be any keyerrors if the search proceeds successfully! I kept this in the off-chance that a pipeline maybe failed during search, but can remove.

evalml/tests/component_tests/test_stacked_ensemble_classifier.py
```diff
@@ -369,3 +369,179 @@ def test_ensembler_use_component_preds_multi(
...
+    pl_no_cache = _make_stacked_ensemble_pipeline(
+        input_pipelines=input_pipelines,
+        problem_type=ProblemTypes.BINARY,
+        cached_data=None,
+    )
+
+    # ensure we are not calling transform or fit for the components we have in our cache
+    pl_cache.fit(X, y)
+    mock_transform_fit.assert_not_called()
+    mock_estimator_fit.assert_called_once()
```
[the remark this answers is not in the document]
bchen1116: @freddyaboulton this fit call is the call we'd expect for the final metalearner estimator fit, which we will have to fit regardless of cache or not. Will add a comment to clarify this!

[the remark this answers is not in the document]
freddyaboulton: Makes sense!

evalml/pipelines/utils.py
freddyaboulton: What's the current behavior if we run ensembling and one of the pipelines fails? My guess is that we'd use another pipeline from that model family that didn't fail in the ensemble so we would never try to read from the cache of a failed pipeline. 

I think we should preserve the current behavior on `main` in this case. If we need the KeyError for that let's keep it but add a comment and test for that case!

freddyaboulton (approved):
@bchen1116 Looks good to me! Left a comment about adding a test for the `KeyError` case if it helps us preserve the current behavior of running ensembling when a pipeline fails.

evalml/pipelines/utils.py
bchen1116: @freddyaboulton You're right! It should only be getting the pipelines from the saved `_best_pipeline_info`, which would only be updated upon successful completion of pipelines. I think we should be able to remove it without issue!

PR #3267, opened 2022-01-20
Remove impute_all parameter from `PerColumnImputer`

### LINKED ISSUE #3241: ``PerColumnImputer`` fails if the input dataframe has datetime columns
``PerColumnImputer`` fails if the input dataframe has datetime columns.

I think similar to our `Imputer`, we should allow datetime columns to be part of the input, even if we take no actions on them.

```python
from evalml.pipelines.components import PerColumnImputer

X, y = load_fraud(n_rows=1500)
X.head()

pci = PerColumnImputer()
pci.fit(X, y)
```

gives us this stack trace:
```
<ipython-input-2-7f4dae7e2db2> in <module>
      1 pci = PerColumnImputer()
----> 2 pci.fit(X, y)

     17         @wraps(method)
     18         def _set_fit(self, X, y=None):
---> 19             return_value = method(self, X, y)
     20             self._is_fitted = True
     21             return return_value

~/Desktop/evalml/evalml/pipelines/components/transformers/imputers/per_column_imputer.py in fit(self, X, y)
     89 
     90         for column, imputer in self.imputers.items():
---> 91             imputer.fit(X.ww[[column]])
     92 
     93         return self

     17         @wraps(method)
     18         def _set_fit(self, X, y=None):
---> 19             return_value = method(self, X, y)
     20             self._is_fitted = True
     21             return return_value

~/Desktop/evalml/evalml/pipelines/components/transformers/imputers/simple_imputer.py in fit(self, X, y)
     64             X = X.astype("category")
     65 
---> 66         self._component_obj.fit(X, y)
     67         return self
     68 

~/Desktop/evalml_venv/lib/python3.7/site-packages/sklearn/impute/_base.py in fit(self, X, y)
    317             Fitted estimator.
    318         """
--> 319         X = self._validate_input(X, in_fit=True)
    320 
    321         # default fill_value is 0 for numerical input and "missing_value"

~/Desktop/evalml_venv/lib/python3.7/site-packages/sklearn/impute/_base.py in _validate_input(self, X, in_fit)
    295                 "categorical data represented either as an array "
    296                 "with integer dtype or an array of string values "
--> 297                 "with an object dtype.".format(X.dtype)
    298             )
    299 

ValueError: SimpleImputer does not support data with dtype datetime64[ns]. Please provide either a numeric array (with a floating point or integer dtype) or categorical data represented either as an array with integer dtype or an array of string values with an object dtype.
```

eccabay:
Closes #3241

### DIFF
--- a/evalml/tests/component_tests/test_per_column_imputer.py
+++ b/evalml/tests/component_tests/test_per_column_imputer.py
@@ -141,7 +141,12 @@ def test_non_numeric_valid(non_numeric_df):
         }
     )
     # most frequent with all strings
-    strategies = {"C": {"impute_strategy": "most_frequent"}}
+    strategies = {
+        "A": {"impute_strategy": "most_frequent"},
+        "B": {"impute_strategy": "most_frequent"},
+        "C": {"impute_strategy": "most_frequent"},
+        "D": {"impute_strategy": "most_frequent"},
+    }
     transformer = PerColumnImputer(impute_strategies=strategies)
 
     X_expected = pd.DataFrame(
@@ -158,7 +163,11 @@ def test_non_numeric_valid(non_numeric_df):
 
     X = non_numeric_df.copy()
     # constant with all strings
-    strategies = {"D": {"impute_strategy": "constant", "fill_value": 100}}
+    strategies = {
+        "B": {"impute_strategy": "most_frequent"},
+        "C": {"impute_strategy": "most_frequent"},
+        "D": {"impute_strategy": "constant", "fill_value": 100},
+    }
     transformer = PerColumnImputer(impute_strategies=strategies)
 
     X.ww.init(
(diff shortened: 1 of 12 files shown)


angela97lin (commented):
Blocking on https://github.com/alteryx/evalml/issues/3241#issuecomment-1018126554! Let's discuss, I think there's some confusion here :)

evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -190,6 +190,15 @@ def test_non_numeric_valid(non_numeric_df):
     assert_frame_equal(X_expected, X_t)
 
 
+def test_datetime_doesnt_error(fraud_100):
```
freddyaboulton, line 193: nit: I would also maybe specify a column to impute such as `amount` or `provider`

freddyaboulton (approved):
Looks good to me! Thanks @eccabay

evalml/tests/component_tests/test_per_column_imputer.py
```diff
@@ -324,7 +331,7 @@ def test_per_column_imputer_impute_all_is_false():
         "all_nan_included": {"impute_strategy": "most_frequent"},
         "column_with_nan_included": {"impute_strategy": "most_frequent"},
     }
-    transformer = PerColumnImputer(impute_strategies=strategies, impute_all=False)
+    transformer = PerColumnImputer(impute_strategies=strategies)
```
jeremyliweishih, line 334: question here: why does this test return three columns? I guess the overall question I have is, what is the behavior of this component for columns not specified in `impute_strategies`?

angela97lin (approved):
LGTM, thanks @eccabay! Agreed that we should add a comment that the behavior of this component has changed but otherwise 🚢

evalml/tests/component_tests/test_per_column_imputer.py
eccabay: The new behavior for  columns not specified in `impute_strategies` is to completely leave them alone. Any `NaN` values will remain, and unspecified columns will not be dropped even if they are fully nan. 

In this test, you can see that of the two `NaN` columns, the one included in the impute strategies was dropped but the one that wasn't, remains. Same with the some-nan columns, the one that was included was imputed but the one that wasn't still has nans after transformation.

PR #3268, opened 2022-01-21
Update `pmdarima` version to fix nightlies

angela97lin:
### Pull Request Description

-----

### DIFF
--- a/.github/workflows/linux_nightlies.yml
+++ b/.github/workflows/linux_nightlies.yml
@@ -1,9 +1,6 @@
 name: Nightly unit tests, linux
 
 on:
-  schedule:
-      - cron: '0 7 * * *'
-on:
   pull_request:
     types: [opened, synchronize]
   push:
@@ -60,15 +57,3 @@ jobs:
         run: |
           source test_python/bin/activate
           make ${{matrix.command}}
-      - name: Notify on Slack
-        uses: 8398a7/action-slack@v3
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-        with:
-          job_name: Nightly ${{ matrix.python_version }} ${{matrix.command}}
-          status: ${{ job.status }}
-          fields: workflow,job,took
-          mention: channel
-          if_mention: failure,cancelled
-          text: ':elmofire:'
-        if: failure()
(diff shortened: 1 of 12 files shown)


.github/workflows/linux_nightlies.yml
```diff
@@ -1,8 +1,11 @@
 name: Nightly unit tests, linux
 
 on:
-  schedule:
-      - cron: '0 7 * * *'
+  pull_request:
```
angela97lin, line 4: Will revert this file before merging

Makefile
```diff
@@ -84,7 +84,7 @@ git-test-integration:
 .PHONY: installdeps
 installdeps:
 	pip install --upgrade pip -q
-	pip install -e . -q
+	pip install -e . -v
```
angela97lin, line 87: Will also revert this file

gsheni (approved):
looks good. assuming `on-schedule` is reverted

setup.py
```diff
@@ -13,7 +13,7 @@
     name='evalml',
     version='0.42.0',
     author='Alteryx, Inc.',
-    author_email='support@featurelabs.com',
```
jeremyliweishih, line 16: Good change!

jeremyliweishih (approved):
Thank you!

angela97lin:
@freddyaboulton Never mind, looks like just using a newer version of `pmdarima` resolved the problems 😅

requirements.txt
```diff
@@ -10,7 +10,7 @@ graphviz>=0.13;platform_system!='Windows'
 seaborn>=0.11.1
 category_encoders>=2.2.2
 imbalanced-learn>=0.8.0
-pmdarima==1.8.0
+pmdarima>=1.8.1
```
freddyaboulton, line 13: Does 1.8.1 actually work in 3.9? Let’s merge this in to fix the nightlies but wondering if we need to use >=1.8.4

freddyaboulton (approved):
Thank you so much! Glad it was an easy fix

requirements.txt
angela97lin: @freddyaboulton I think since I just set a lower bound, it ended up installing `1.8.4` anyways (https://github.com/alteryx/evalml/runs/4902413563?check_suite_focus=true); I can check 1.8.1 by hardcoding and see what happens? 👀

angela97lin: Looks good in 3.9 even with 1.8.1!! https://github.com/alteryx/evalml/pull/3271

PR #3272, opened 2022-01-22

machineFL:
This is an auto-generated PR with **latest** dependency updates.


freddyaboulton:
Looks like the root cause may be that converting a nullable type to `Categorical` preserves the nullable types in the pandas Categorical categories. On the top is the latest pandas and on the bottom is 1.3.4 pandas.

image

This causes np.asarray calls in our imputers to introduce the new pandas null value place-holder, `<NA>` which throws off sklearn:

image

PR #3280, opened 2022-01-25
Fix binary classification pipeline thresholding

### LINKED ISSUE #3086: Optimize_threshold not working as expected
Repro (in test file):
```python
def test_find_confusion_matrix_pipeline_threshold_tester(
    logistic_regression_binary_pipeline_class, X_y_binary
):
    bcp = logistic_regression_binary_pipeline_class({})
    X, y = X_y_binary
    y = pd.Series(y)
    bcp.fit(X, y)
    preds = bcp.predict_proba(X).iloc[:, 1]
    bcp.optimize_threshold(X, y, preds, AccuracyBinary())
    print(bcp.threshold)
    print(bcp.score(X, y, [AccuracyBinary()]))
    bcp.threshold = 0.5
    print(bcp.score(X, y, [AccuracyBinary()]))
```
Note that here, the accuracy is higher when we use a threshold of 0.5 versus the threshold we optimized with. A summary of this finding can be seen [here](https://github.com/alteryx/evalml/pull/3080). 

We see that our optimize threshold method isn't able to find this better threshold, which we can note as a performance bug. We expect that we should be able to find the 0.5 threshold as the optimal, but currently do not. This issue tracks resolving this performance discrepancy and fixing our `optimize_threshold` method.

bchen1116:
fix #3086 

Perf tests here
[report_threshold.html.zip](https://github.com/alteryx/evalml/files/7952032/report_threshold.html.zip)

Tests for two other methods here:
[report_threshold_binexp.html.zip](https://github.com/alteryx/evalml/files/7961737/report_threshold_binexp.html.zip)
[report_threshold_randexp.html.zip](https://github.com/alteryx/evalml/files/7961738/report_threshold_randexp.html.zip)

Current defaults are faster, so we will continue with that.

Writeup under name "Global Binary Pipeline Thresholding Perf Tests"


docs/source/demos/lead_scoring.ipynb
```diff
@@ -41,7 +41,7 @@
    "outputs": [],
    "source": [
     "lead_scoring_objective = LeadScoring(\n",
-    "    true_positives=100,\n",
+    "    true_positives=50,\n",
```
bchen1116, line 44: Change needed to get the assert to pass in this file. Likely due to the new thresholding method

evalml/objectives/binary_classification_objective.py
```diff
@@ -50,16 +50,14 @@ def optimize_threshold(self, ypred_proba, y_true, X=None):
...
-                ypred_proba=ypred_proba, threshold=threshold, X=X
+                ypred_proba=ypred_proba, threshold=threshold[0], X=X
             )
             cost = self.objective_function(y_true, y_predicted, X=X)
             return -cost if self.greater_is_better else cost
 
-        optimal = minimize_scalar(
-            cost, bracket=(0, 1), method="Golden", options={"maxiter": 250}
-        )
+        optimal = differential_evolution(cost, bounds=[(0, 1)], seed=0)
```
bchen1116, line 58: this method is a global minimum optimizer, whereas `minimize_scalar` was a local minimizer. This will take a little longer, but should allow us to avoid most issues where the optimized threshold doesn't align with our expectations

docs-requirements.txt
```diff
@@ -1,5 +1,5 @@
 docutils>=0.15.2,<0.17
-pydata-sphinx-theme>=0.3.1
+pydata-sphinx-theme>=0.7.2
```
bchen1116, line 2: Required to fix this issue: 
https://github.com/numpy/numpy/issues/20690

Otherwise doc-builds continued to fail

evalml/objectives/binary_classification_objective.py
jeremyliweishih, line 58: might be interesting to play with some of the parameters to see if we can get the same performance but quicker. Def not blocking though!

jeremyliweishih (approved):
Looks great to me!

docs/source/release_notes.rst
```diff
@@ -6,10 +6,12 @@
         * Updated ``DefaultAlgorithm`` to also limit estimator usage for long-running multiclass problems :pr:`3099`
         * Added ``make_pipeline_from_data_check_output()`` utility method :pr:`3277`
     * Fixes
+        * Updated the binary classification pipeline's ``optimize_thresholds`` method to use Nelder-Mead :pr:`3280`
     * Changes
         * Removed ``DateTimeNaNDataCheck`` and ``NaturalLanguageNaNDataCheck`` in favor of ``NullDataCheck`` :pr:`3260`
     * Documentation Changes
         * Update documentation and docstring for `validate_holdout_datasets` for time series problems :pr:`3278`
+        * Updated version of ``pydata-sphinx-theme`` :pr:`3280`
```
freddyaboulton, line 14: Did this actually happen in this PR?

evalml/objectives/binary_classification_objective.py
bchen1116: From my impression, the default values should perform fairly similar in speed to the other optimization methods available to this. I can try running a few other perf tests to double check though!

docs/source/demos/lead_scoring.ipynb
freddyaboulton: I'm confused by this. If this method finds the global max, and the other found a local max, shouldn't this method give the same score or higher with the same parameters?

freddyaboulton (commented):
@bchen1116 I'm confused as to why we need to change the parameters in the lead scoring doc. Doesn't that mean this method found a threshold that's less optimas than the previous method?

docs/source/release_notes.rst
bchen1116: Ah, an old attempt at a fix!

bchen1116:
@freddyaboulton, took me a while to track down why the scores changed.
First, the thresholding data was optimized on a subset of the training data, while the holdout data is used to score. This can lead to deviations in holdout data performance. Along with that, here are two screenshots of the thresholding parameter and associated objective score:

Threhsolding param and Lead scoring value [This branch] (I lightly highlighted the threshold and score of the AUC/F1 oriented run):
image

Thresholding param and F1 value [This branch]:
image


Thresholding param and Lead scoring  [Main branch]:
image

I've highlighted the optimal objectives here for the optimal thresholds on this new branch. Using the current parameters, the F1 score is highest at a threshold of ~0.44, while the lead scoring is highest at a threshold of ~0.40. Because of this slight deviation, we see a drop in the actual lead scoring value on the holdout data, which is why I needed to change the lead scoring objective parameters.

We see that on the main branch, we don't even search for objective scores when the threshold is below 0.42. This would explain why we see such a difference, since the local optimization method on main doesn't search as thoroughly as the global method here. While it does find a local minima, we don't find as optimal of a minima as with this new branch. Although this leads to a lower holdout score on this new branch, that is likely due to the way we structured the Lead scoring objective cost function itself. 

I believe it's fine to move forward with this PR

freddyaboulton:
Thanks for the explanation @bchen1116 !

freddyaboulton (approved):

chukarsten (approved):
LGTM!

PR #3285, opened 2022-01-26
Fix feature importance for time series baseline

### LINKED ISSUE #3282: Baseline time series regression pipeline doesn't show target delay feature
When running AutoMLSearch with the Time Series Regression problem type, the feature importance dataframe for the baseline pipeline only shows the time index column with a feature importance of `0.0`. 

image

As an evalML user, it'd be nice to also see the target delay feature with its feature importance of 1.0, since it's useful for comparing between pipelines.

### LINKED ISSUE #3284: Time Series documentation shows negative percent better than baseline
image

The issue is that in #3205 , we changed the default objective for time series to MedianAE so we're computing percent better than baseline for R2 scores as if they were MedianAE scores.

freddyaboulton:
### Pull Request Description

Fixes #3284 , Fixes #3282

-----

### DIFF
--- a/docs/source/user_guide/timeseries.ipynb
+++ b/docs/source/user_guide/timeseries.ipynb
@@ -217,7 +217,7 @@
     "\n",
     "pl.fit(X_train, y_train)\n",
     "\n",
-    "best_pipeline_score = pl.score(X_test, y_test, ['R2'], X_train, y_train)['R2']"
+    "best_pipeline_score = pl.score(X_test, y_test, ['MedianAE'], X_train, y_train)['MedianAE']"
    ]
   },
   {
@@ -237,7 +237,7 @@
    "source": [
     "baseline = automl.get_pipeline(0)\n",
     "baseline.fit(X_train, y_train)\n",
-    "naive_baseline_score = baseline.score(X_test, y_test, ['R2'], X_train, y_train)['R2']"
+    "naive_baseline_score = baseline.score(X_test, y_test, ['MedianAE'], X_train, y_train)['MedianAE']"
    ]
   },
   {
@@ -253,7 +253,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The pipeline found by AutoMLSearch has a 268% improvement over the naive forecast!"
+    "The pipeline found by AutoMLSearch has a 31% improvement over the naive forecast!"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


evalml/pipelines/components/estimators/regressors/time_series_baseline_estimator.py
```diff
@@ -91,6 +95,8 @@ def predict(self, X):
                 "Time Series Baseline Estimator is meant to be used in a pipeline with "
                 "a Time Series Featurizer"
             )
+        self._num_features = X.shape[1]
```
freddyaboulton, line 98: We typically don't set fields like this in `predict` but since the `feature_name` is not used until predict, it seemed better to keep this logic together.

docs/source/user_guide/timeseries.ipynb
```diff
@@ -253,7 +253,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The pipeline found by AutoMLSearch has a 268% improvement over the naive forecast!"
+    "The pipeline found by AutoMLSearch has a 31% improvement over the naive forecast!"
```
bchen1116, line 256: Personally, I'm hesitant on putting numeric values for these since if we change/add components or change the way things are calculated, these numbers will likely change. Not blocking, but something to keep in mind!

bchen1116 (approved):
LGTM! Left one comment but nothing blocking.

tamargrey:
Ran this branch with my time series demo, and I'm seeing these feature importances for the baseline
image
Should `target_rolling_mean` be there? And if so, how does it have 0 importance when, in the other pipelines, it's one of the most important features?

ParthivNaresh:
> Ran this branch with my time series demo, and I'm seeing these feature importances for the baseline image Should `target_rolling_mean` be there? And if so, how does it have 0 importance when, in the other pipelines, it's one of the most important features?

We [include](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py#L200) the `target_rolling_mean` from the `TimeSeriesFeaturizer` if it's numeric.

Also with the baseline pipeline, predictions are based entirely off of the delayed feature, so we'd have to set that to 1 I believe.

ParthivNaresh (approved):
Looks solid!

tamargrey:
> We [include](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py#L200) the `target_rolling_mean` from the `TimeSeriesFeaturizer` if the target is numeric.
> 
> Also with the baseline pipeline, predictions are based entirely off of the delayed feature, so we'd have to set that to 1 I believe.

So If I'm understanding correctly: The target_rolling_mean feature shows up in the baseline pipeline importances because its added via the TimeSeriesFetureizer even though it's not actually included in the model and should always have zero importance?

freddyaboulton:
@tamargrey Exactly. The estimator technically "sees" the rolling mean feature, but it doesn't use it, so it gets an importance of 0.

PR #3260, opened 2022-01-19
Removes ``DateTimeNaNDataCheck`` and ``NaturalLanguageNaNDataCheck``

angela97lin:
Closes #3242. Currently pointing to my branch right now but will wait to merge into main after #3182 is merged!

PR #3288, opened 2022-01-27
Add more granularity to `DateTimeFormatDataCheck`

### LINKED ISSUE #3258: Improvements to DateTimeFormatDataCheck
As a user of evalml, I wish the date time format data check would raise errors for the following cases:

- Verify that primary datetime feature has only one row per datetime
- Verify that primary datetime feature has no dates missing between start and end date

eccabay:
Closes #3258


evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +151,52 @@ def validate(self, X, y):
...
+            frequencies = []
+            missing = []
+            for i in range(len(datetime_values) - 2):
+                freq = pd.infer_freq(datetime_values[i : i + 3])
+                if freq is None and i not in missing:
+                    missing.append(i + 1)
+                frequencies.append(freq)
+
+            # Check for only one row per datetime
+            if len(np.unique(datetime_values)) < len(datetime_values):
```
freddyaboulton, line 170: Since more these two errors can apply to the same data, might make sense to return all that apply as opposed to one or the other?

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +151,52 @@ def validate(self, X, y):
...
-                    data_check_name=self.name,
-                    message_code=DataCheckMessageCode.DATETIME_HAS_UNEVEN_INTERVALS,
-                ).to_dict()
-            )
-
-        if not (pd.DatetimeIndex(datetime_values).is_monotonic_increasing):
+            frequencies = []
+            missing = []
+            for i in range(len(datetime_values) - 2):
+                freq = pd.infer_freq(datetime_values[i : i + 3])
```
freddyaboulton, line 164: The missing dates part is really tricky. I like this but I saw that it couldn't catch this one case:

Screen Shot 2022-01-27 at 4 59 34 PM

If we change line 180 to read `len(set(frequencies)) == 2`, then it does catch this case as having missing dates.

Screen Shot 2022-01-27 at 5 01 46 PM

Would this algorithm work?

```
missing_dates = datetime_values.diff(1).isna().nunique() > 1
```

If there are no missing dates, the first order differences should be constant.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +151,52 @@ def validate(self, X, y):
...
+                if freq is None and i not in missing:
+                    missing.append(i + 1)
```
freddyaboulton, line 166: @fjlanasa Do we need to return the missing indices? It could make the response very big since there could potentially be many missing dates.

freddyaboulton (commented):
Looks good @eccabay. Left some suggestions for minor tweaks.

evalml/data_checks/datetime_format_data_check.py
freddyaboulton: Now that I think about it, we can also use `datetime_values.diff(1)` to determine if there are any duplicate dates. 

```
duplicate_dates = (datetime_values.diff(1) == pd.Timedelta(0)).any()
```

And we should probably filter out the rows equal to Timedelta(0) prior to computing the uniques.

evalml/data_checks/datetime_format_data_check.py
fjlanasa: I think we can do without

evalml/data_checks/datetime_format_data_check.py
eccabay: @freddyaboulton the case you highlighted is an interesting one, and plays into my uncertainty about where the line of "missing dates" vs "uneven interval" is. IMO, it makes sense for your example to have uneven intervals rather than missing dates, since there is that segment in the middle where there is a noticeable two day interval. 

Of course, I can see an argument for the other way, so I'm curious to hear what you think!

eccabay: Just discussed this with @dsherry during office hours. Looks like the best way forward may be to raise both errors in this case, instead of just one or the other. I'll update this PR to reflect that shortly!

ParthivNaresh: @eccabay That's a good point, I would say that:

- if more than one frequency (aside from None) is identified in the dates column, then it would be uneven interval
- if one frequency (aside from None) is identified, and then None has also been identified **once**, then it would indicate missing dates
- if one frequency (aside from None) is identified, and then None has also been identified **more than once non-consecutively**, then it would indicate uneven interval.

But I think raising both errors might be the more straightforward path as you said!

freddyaboulton (approved):
@eccabay Thank you for making the changes! This looks great.

I apologize for the scope creep but I think we should check if the column has any NaNs in DateTimeFormatDataCheck. We can't allow users to build time series models if the time index column has NaNs. This was previously handled by `DateTimeNaNDataCheck` but we deleted it in #3260, (which was the right call for non-timeseries uses but we still need the check for time series).

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
+        )
+
+        is_increasing = pd.DatetimeIndex(datetime_values).is_monotonic_increasing
+
+        nan_columns = datetime_values.isna().any()
+        if nan_columns:
             messages.append(
                 DataCheckError(
-                    message=f"No frequency could be detected in {col_name}, possibly due to uneven intervals.",
+                    message=f"Input datetime column ({col_name}) contains NaN values. Please impute NaN values or drop these rows.",
```
freddyaboulton, line 192: 👍

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
+
+        if not inferred_freq:
 
-        if not (pd.DatetimeIndex(datetime_values).is_monotonic_increasing):
+            # Check for only one row per datetime
+            duplicate_dates = pd.Series(datetime_values).diff(1) == pd.Timedelta(0)
+            if duplicate_dates.any():
+                messages.append(
+                    DataCheckError(
+                        message=f"{col_name} has more than one row with the same datetime value.",
```
chukarsten, line 207: I think the only thing these messages need is maybe a bit more clarity like "Column '{col_name}' has...".  I think, without the single quotes, it's not entirely clear that it's a column name and not some var.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
+                        data_check_name=self.name,
+                        message_code=DataCheckMessageCode.DATETIME_HAS_REDUNDANT_ROW,
+                    ).to_dict()
+                )
+                # drop any duplicates before continuing
+                datetime_values = datetime_values[~duplicate_dates]
+
+            frequencies = []
+            for i in range(len(datetime_values) - 2):
+                freq = pd.infer_freq(datetime_values[i : i + 3])
```
chukarsten, line 217: hypernit: I think 3 should be assigned to some variable `interval_size` or something with the previous line referencing that `interval_size`, e.g. `- (interval_size-1)`

Curious, though, why not go with the `num_disruptions = len(set(datetime_values.diff(1)))`?  I'll have to check your testing, but I'm not sure how "fuzzy" the frequency inference is, either, like if the given datetimes are separated by basically a day but with a few seconds difference.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
+                frequencies.append(freq)
+            num_disruptions = len(set(frequencies))
+
+            # Check for no date missing in ordered dates
+            if num_disruptions > 1 and is_increasing:
```
chukarsten, line 222: What does `is_increasing` give us here?  It would seem this DataCheckError should be generated whether the dataset is ordered properly or not.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
+                        data_check_name=self.name,
+                        message_code=DataCheckMessageCode.DATETIME_IS_MISSING_VALUES,
+                    ).to_dict()
+                )
+
+            # Give a generic uneven interval error if two or more frequencies are detected
+            if num_disruptions > 2 or (num_disruptions > 0 and not is_increasing):
+                messages.append(
+                    DataCheckError(
+                        message=f"No frequency could be detected in {col_name}, possibly due to uneven intervals.",
```
chukarsten, line 235: If we're going for more information, maybe it might be worth it give the user a place to look.  I know when I was looking at the bike share data, it was irritating having to go through and ID the little "steps" that were occuring over the weekends.  We could also file a follow on issue if you think it's worth doing and worth doing later.

evalml/tests/data_checks_tests/test_datetime_format_data_check.py
```diff
@@ -12,11 +13,18 @@
 
 @pytest.mark.parametrize("input_type", ["pd", "ww"])
 @pytest.mark.parametrize(
-    "uneven,type_errors", [(True, False), (False, True), (False, False)]
+    "redundant,missing,uneven,type_errors",
```
chukarsten, line 16: Any reason to not just parametrize over each individually?  I think it makes the failures look a little more obvious.

evalml/tests/data_checks_tests/test_datetime_format_data_check.py
```diff
@@ -25,11 +33,21 @@ def test_datetime_format_data_check_typeerror_uneven_intervals(
...
         dates = pd.date_range("2021-01-01", periods=30)
 
-    if uneven:
+    if missing:
         # Skips 2021-01-30 and appends 2021-01-31, skipping a date and triggering the error
         dates = pd.date_range("2021-01-01", periods=29).append(
             pd.date_range("2021-01-31", periods=1)
         )
+    if uneven:
+        dates = pd.DatetimeIndex(
```
chukarsten, line 42: Do we know what will happen if there is a non-standard frequency?  Like a 3-day frequency or a 30 minute frequency?

chukarsten (changes requested):
Thanks for getting this out there Becca!  I think the only things I'll block on are the single quotes around the column name and perhaps the testing of additional intervals/non-standard periods.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -118,21 +179,66 @@ def validate(self, X, y):
...
                 ).to_dict()
             )
+            # NaN values are a more significant error than missing values and will break the data check logic, so we return here
+            return messages
+
+        if not inferred_freq:
 
-        if not (pd.DatetimeIndex(datetime_values).is_monotonic_increasing):
+            # Check for only one row per datetime
+            duplicate_dates = pd.Series(datetime_values).diff(1) == pd.Timedelta(0)
```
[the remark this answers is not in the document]
eccabay: Yep, I think that's a fair assumption. The data being ordered is a more pressing issue than missing values, and should take priority over this check.

evalml/data_checks/datetime_format_data_check.py
eccabay: I like these suggestions! This method of checking for disruptions stems from before I knew about the `diff` function, so I'll update this if it work.

eccabay: After much experimentation, it seems like `num_disruptions = len(set(datetime_values.diff(1)))` changes the distinction between missing vs uneven errors in a way we probably don't want. I'm going to add the `interval_size`, but leave out the other change, at least for now.

evalml/data_checks/datetime_format_data_check.py
eccabay: This error is specifically for missing values. Without the `is_increasing` check, anything that's out of order will say there are values missing (even if the dates are only scrambled but otherwise ok), which wouldn't be accurate.

evalml/data_checks/datetime_format_data_check.py
eccabay: The way the logic works out right now, this error will never be the only one raised by this data check - either the `missing_values` or `is_monotonic` check will be there as well. Ideally, these will give more useful information while the uneven error gives the indication that something's significantly wrong with the datetime values. I'm happy to update that in a later issue though.

evalml/tests/data_checks_tests/test_datetime_format_data_check.py
eccabay: It seems to work the same way, but I added another test to be sure!

chukarsten (approved):
Love it!  Thank you so much for the changes!

ParthivNaresh (approved):
Excellent work with this and thank you for incorporating all the changes!

PR #3289, opened 2022-01-27
Automated Latest Dependency Updates

machineFL:
This is an auto-generated PR with **latest** dependency updates.

PR #3290, opened 2022-01-28
Upgrade to WW version 0.12.0

jeremyliweishih:
Fixes docstring failing for WW version 0.12.0 brought up in #3289. Sneaking in a vowpalwabbit cap and some other dependency changes to unblock CI.

To recap:

1. We first upgraded woodwork and fixed the docstring failure
2. However, shap tests were breaking on [min-deps tests ](https://github.com/alteryx/evalml/runs/4985911311?check_suite_focus=true) which could be related to [this](https://github.com/slundberg/shap/issues/1895) and [this](https://github.com/slundberg/shap/pull/1802). So we bumped up the SHAP version.
3. vowpalwabbit also released during this period of time and [broke our windows tests](https://github.com/alteryx/evalml/runs/4986435992?check_suite_focus=true). To fix this we capped VPW for now
4. Lastly added the lastest dask release to `latest_dependency_versions.txt` as dask released as well.
5. However, [VPW wheel got picked up on PyPI](https://github.com/alteryx/evalml/runs/4986677302?check_suite_focus=true) so reverting the cap.


evalml/data_checks/outliers_data_check.py
```diff
@@ -140,7 +140,7 @@ def get_boxplot_data(data_):
             ...                "median": -3.0,
             ...                "q3": -2.0,
             ...                "low_bound": -7.0,
```
jeremyliweishih, line 142: @tamargrey with the changes should this low bound be `-4.0` instead of `-7.0`?

freddyaboulton: It should be the minimum if low bound is less than the minimum but since the min is -1201 I think it's expected for the low bound to stay at -7?

tamargrey: that's correct, thank's Freddy!

tamargrey (approved):
lgtm!

PR #3297, opened 2022-02-01
Add easier way to determine whether data splitter is CV

### LINKED ISSUE #3098: Easy way to determine whether or not a data splitter counts as a CV splitter
**Background**
We support [the following data splitters](https://github.com/alteryx/evalml/tree/main/evalml/preprocessing/data_splitters): training-validation, rolling origin in-order time series CV, no-split, and all sklearn splitters inheriting from sklearn's `BaseCrossValidator`. The confusing thing is, `BaseCrossValidator` defines a `split` method and `n_splits` attribute, but the way we've set up our inheritance, the subclasses of it which we define could have `n_splits=1` and therefore not be a CV splitting method. Example of that is our training-validation split.

**Proposed feature**
As a dev working with evalml I'd like to quickly know whether my data splitter counts as CV or not, so that I can inform my users about what data splitting method automl selected for me.

Something like this would do it at the API level:
```python
automl = AutoMLSearch(...)
data_splitter = automl.data_splitter
# option one: helper method
print(evalml.preprocessing.is_splitter_cv(data_splitter))
# option two: attribute. could be a class property
print(data_splitter.is_cv)
```

**Workaround**
A workaround is to simply check whether or not `mean_cv_score` was populated in the rankings list. If its not populated, automl didn't consider the split method to be a CV method. This works but is too indirect.

**Impl Notes**
Let's think through what structure would make the most sense here. We could update the inheritance structure and have two base classes. But will that work well for current and future time series splitters we'd like to define?

bchen1116:
fix #3098

### DIFF
--- a/evalml/automl/utils.py
+++ b/evalml/automl/utils.py
@@ -2,7 +2,6 @@
 from collections import namedtuple
 
 import pandas as pd
-from sklearn.model_selection import KFold, StratifiedKFold
 
 from evalml.objectives import get_objective
 from evalml.pipelines import (
@@ -14,6 +13,8 @@ from evalml.pipelines import (
     TimeSeriesRegressionPipeline,
 )
 from evalml.preprocessing.data_splitters import (
+    KFold,
+    StratifiedKFold,
     TimeSeriesSplit,
     TrainingValidationSplit,
 )
@@ -97,16 +98,11 @@ def make_data_splitter(
             test_size=_LARGE_DATA_PERCENT_VALIDATION, shuffle=shuffle
         )
     if problem_type == ProblemTypes.REGRESSION:
-        kfold = KFold(n_splits=n_splits, random_state=random_seed, shuffle=shuffle)
-        # can set this to true directly since k-fold requires >1 splits
-        kfold.is_cv = True
-        return kfold
+        return KFold(n_splits=n_splits, random_state=random_seed, shuffle=shuffle)
     elif problem_type in [ProblemTypes.BINARY, ProblemTypes.MULTICLASS]:
-        skfold = StratifiedKFold(
+        return StratifiedKFold(
             n_splits=n_splits, random_state=random_seed, shuffle=shuffle
         )
-        skfold.is_cv = True
-        return skfold
 
 
 def tune_binary_threshold(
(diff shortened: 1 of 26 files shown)


evalml/automl/utils.py
```diff
@@ -97,11 +97,16 @@ def make_data_splitter(
             test_size=_LARGE_DATA_PERCENT_VALIDATION, shuffle=shuffle
         )
     if problem_type == ProblemTypes.REGRESSION:
-        return KFold(n_splits=n_splits, random_state=random_seed, shuffle=shuffle)
+        kfold = KFold(n_splits=n_splits, random_state=random_seed, shuffle=shuffle)
+        # can set this to true directly since k-fold requires >1 splits
+        kfold.is_cv = True
```
chukarsten, line 102: This is kind of worrisome.  The KFold class is an sklearn object.  There's not really much reason for contributors or other devs to expect this attribute added to the standard sklearn object if they don't know about this code segment that modifies it.  Maybe we should consider a simple class wrapper with the same name the inherits from `KFold` and `StratifiedKFold` but defines the property as the other splitters do.  Curious what others think...

chukarsten (changes requested):
Thanks for tackling, Bryan.  I think we might need to rethink dynamically adding the attribute to the sklearn object, though.  Perhaps a subclass of KFold/StratifiedKFold?

chukarsten (approved):
Thanks so much @bchen1116 !! Good to go.

evalml/preprocessing/data_splitters/no_split.py
```diff
@@ -29,6 +29,15 @@ def get_n_splits():
         """
         return 0
 
+    @property
```
angela97lin, line 32: Omega nit-pick but can we make this abstract on `BaseCrossValidator` to so that we don't forget to implement it for any future classes we might add?

angela97lin (approved):
LGTM, left a comment about making is_cv abstract for our base class but not blocking

evalml/preprocessing/data_splitters/no_split.py
bchen1116: @angela97lin `BaseCrossValidator` is a class we take from sklearn. Not sure if we can add an abstract to this unless we create our own `BaseCrossValidator` class with this attribute?

PR #3261, opened 2022-01-19
Make `DefaultAlgorithm` the default AutoML Algorithm

jeremyliweishih:
Fixes #2867. Since there is a lot in this PR, I will use the PR for #2693 to fix docstrings and think about changes to `AutoMLSearch` API.

I will be doing the following in this PR:
- justify why a test deserves a `_iterative` tag
- moving such `_iterative` tests to their own files
- generalize any tests I can to be algorithm agnostic

I will also create a list of tests that we can improve in #2868 with a `MockAlgorithm` with these following criteria:
- test logic hinges on algorithm implementation (i.e number of pipelines and such)
- I cannot generalize the test to work with both current algorithm implementations

List of tests that we can use a mock algorithm with:
```
1. test_automl_algorithm
2. test_time_series_regression_with_parameters
3. test_max_batches_works
4. test_callback (there’s multiple)
5. test_automl_tuner_exception
6. test_results_getter
7. test_automl_pipeline_params_kwargs
8. test_search_with_text_and_ensembling_iterative
9. test_pipelines_per_batch_iterative
10. test_automl_respects_random_seed_iterative
11. test_automl_respects_pipeline_parameters_with_duplicate_components_iterative
12. test_automl_respects_pipeline_custom_hyperparameters_with_duplicate_components_iterative
13. test_automl_search_sampler_ratio
14. test_automl_search_sampler_method
15. test_time_series_pipeline_parameter_warnings
16. test_pipeline_count
```
List of tests we can maybe generalize with much larger lift:
- any tests with component graphs or iterative specific parameters

Performance tests vs. using top level search method (results are identical):
[default_swap_com.html.zip](https://github.com/alteryx/evalml/files/8063770/default_swap_com.html.zip)

PR #3304, opened 2022-02-03
Change `AutoMLSearch` parameters for AutoML Algorithm Upgrade

jeremyliweishih:
Adds on to #3261.

Changes:

- Set default max batches in search to 4 batches (3 for time series), which corresponds to fast mode
- Make `_automl_algorithm` parameter public

### DIFF
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -25,7 +25,7 @@ from evalml.pipelines.utils import (
     make_pipeline,
 )
 from evalml.problem_types import is_regression, is_time_series
-from evalml.utils import classproperty, infer_feature_types
+from evalml.utils import infer_feature_types
 from evalml.utils.logger import get_logger
 
 
@@ -144,10 +144,10 @@ class DefaultAlgorithm(AutoMLAlgorithm):
                         " and Real!"
                     )
 
-    @classproperty
+    @property
     def default_max_batches(self):
         """Returns the number of max batches AutoMLSearch should run by default."""
-        return 4
+        return 4 if not is_time_series(self.problem_type) else 3
 
     def _naive_estimators(self):
         if is_regression(self.problem_type):
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -25,7 +25,7 @@ from evalml.pipelines.utils import (
     make_pipeline,
 )
 from evalml.problem_types import is_regression, is_time_series
-from evalml.utils import infer_feature_types
+from evalml.utils import classproperty, infer_feature_types
 from evalml.utils.logger import get_logger
 
 
@@ -97,7 +97,6 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             tuner_class=None,
             random_seed=random_seed,
         )
-
         self.X = infer_feature_types(X)
         self.y = infer_feature_types(y)
         self.problem_type = problem_type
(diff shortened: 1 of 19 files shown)


evalml/automl/automl_search.py
```diff
@@ -559,7 +559,7 @@ def __init__(
         self.max_batches = max_batches
         self._pipelines_per_batch = _pipelines_per_batch
         if not self.max_iterations and not self.max_time and not self.max_batches:
-            self.max_batches = 1
+            self.max_batches = 4 if automl_algorithm == "default" else 1
```
dsherry, line 562: @jeremyliweishih could this be an attribute of the automl algo?
```python
self.max_batches = automl_algorithm.default_max_batches
```

evalml/tests/automl_tests/test_automl_iterative_algorithm.py
```diff
@@ -425,8 +426,13 @@ def test_max_batches_num_pipelines(
             + num_ensemble_batches
         )
         n_automl_pipelines = n_results
-    assert automl._automl_algorithm.batch_number == max_batches
-    assert automl._automl_algorithm.pipeline_number + 1 == n_automl_pipelines
+    if max_batches is None:
```
jeremyliweishih, line 429: This is a weird one and let me try to explain! In this PR, I make[ these changes ](https://github.com/alteryx/evalml/pull/3304/files#diff-8b4df814812e6f5df24c600bc7f429b036e0e2f78613bfbdcdf0ed3ab86bffd6R732-R736) which not only set the default behavior for max batches but more importantly moved this logic after the automl algorithm is instantiated. So, on main, the default max batches is set before the automl algorithm is set. This meant that if `max_batches`, `max_iterations` etc. were all `None` on main, `max_batches` would be set to 1 and would be passed into `IterativeAlgorithm`. However, `IterativeAlgorithm` has [this logic ](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_algorithm/iterative_algorithm.py#L227-L254) which sets `max_iterations` if `ensembling = True` and if `max_batches is not None`. This meant that on `main`  , turning `ensembling=True` would mean that `AutoMLSearch` would use `max_iterations` as stopping criteria. When I made the changes to search in this PR, it meant that search was correctly using `max_batches` as stopping criteria which meant that search would stop when `batch_number > max_batches`. These changes to this test reflect that and with an added complication of `_pipelines_per_batch`.

jeremyliweishih: Think the above makes sense but still have some details to iron out.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -145,6 +144,11 @@ def __init__(
                         " and Real!"
                     )
 
+    @property
+    def default_max_batches(self):
+        """Returns the number of max batches AutoMLSearch should run by default."""
+        return 4 if not is_time_series(self.problem_type) else 3
```
jeremyliweishih, line 150: Will need to update after ##3191.

chukarsten (approved):
This LGTM, Jeremy!  I think as a suggestion to anyone reading this comment, we should try to keep small and less significant changes (like publicizing automl_algorithm) in separate PRs to maximize views on perhaps the more impactful parts of a PR.  Thanks for doing this, though!  Just trying to increase visibility on the more important stuff rather than the tedious!

evalml/tests/automl_tests/test_automl.py
```diff
@@ -2600,7 +2598,7 @@ def test_search_with_text(AutoMLTestEnv):
     env = AutoMLTestEnv("binary")
     with env.test_context(score_return_value={"Log Loss Binary": 0.30}):
         automl.search()
-    assert automl.rankings["pipeline_name"][1:].str.contains("Natural Language").all()
+    assert automl.rankings["pipeline_name"][1:-1].str.contains("Natural Language").all()
```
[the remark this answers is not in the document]
jeremyliweishih: Its the ensembling pipeline so it won't contain `Natural Language` in the pipeline name.

PR #3305, opened 2022-02-04
Standardize feature importance for estimators

### LINKED ISSUE #3183: Standardize Feature Importance Return
For estimators like the recently added Exponential Smoothing Regressor, feature importance is not defined.  We've been returning a numpy array of zeros for the last few estimators we've added that are similar.  It would be nice to perhaps define a fixture in conftest.py with that array of zeroes and refactor all the `test_feature_importance` unit tests to use this fixture.

This is currently done in KNN, ARIMA, prophet, Vowpal (different return).

angela97lin:
Closes #3183

### DIFF
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -1526,7 +1526,7 @@ def get_linear_coefficients(estimator, features=None):
             "before using this estimator."
         )
     coef_ = estimator.feature_importance
-    coef_ = pd.Series(coef_, name="Coefficients")
+    coef_.name = "Coefficients"
     coef_.index = features
     coef_ = coef_.sort_values()
     coef_ = pd.Series(estimator._component_obj.intercept_, index=["Intercept"]).append(
--- a/evalml/model_understanding/graphs.py
+++ b/evalml/model_understanding/graphs.py
@@ -1526,7 +1526,8 @@ def get_linear_coefficients(estimator, features=None):
             "before using this estimator."
         )
     coef_ = estimator.feature_importance
-    coef_ = pd.Series(coef_, name="Coefficients", index=features)
+    coef_ = pd.Series(coef_, name="Coefficients")
+    coef_.index = features
     coef_ = coef_.sort_values()
     coef_ = pd.Series(estimator._component_obj.intercept_, index=["Intercept"]).append(
         coef_
--- a/evalml/tests/component_tests/test_estimators.py
+++ b/evalml/tests/component_tests/test_estimators.py
@@ -359,3 +359,28 @@ def test_estimator_fit_predict_and_predict_proba_respect_custom_indices(
         )
     X_pred = estimator.predict(X)
     pd.testing.assert_index_equal(X_original_index, X_pred.index, check_names=True)
+
+
+@pytest.mark.parametrize("estimator_class", _all_estimators())
+@pytest.mark.parametrize(
+    "problem_type",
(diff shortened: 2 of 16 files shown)


evalml/model_understanding/graphs.py
```diff
@@ -1526,7 +1526,8 @@ def get_linear_coefficients(estimator, features=None):
             "before using this estimator."
         )
     coef_ = estimator.feature_importance
-    coef_ = pd.Series(coef_, name="Coefficients", index=features)
+    coef_ = pd.Series(coef_, name="Coefficients")
```
jeremyliweishih, line 1529: Do we need all the `pd.Series` calls with this call here?

evalml/tests/component_tests/test_estimators.py
```diff
@@ -359,3 +359,30 @@ def test_estimator_fit_predict_and_predict_proba_respect_custom_indices(
         )
     X_pred = estimator.predict(X)
     pd.testing.assert_index_equal(X_original_index, X_pred.index, check_names=True)
+
+
+@pytest.mark.parametrize("estimator_class", _all_estimators())
```
jeremyliweishih, line 364: Should we delete any of the previous tests? I think we could remove all the tests don't check equality (i.e the estimators that don't support feature importance) since there's some redundancy between these tests.

jeremyliweishih (approved):
Some food for thought but not blocking!

evalml/tests/component_tests/test_estimators.py
chukarsten: Indeed, I can see this going both ways.  I can see tests here in `test_estimators` existing for the purpose of determining that no `feature_importance()` sends back NaN.  If we wanted to do any assertion on the shapes of feature importances relative to the input datasets, now and here would be the place to do it.  I think if we wanted, for some reason, to test the specific `feature_importance()` of an individual estimator and make sure that it calculates the right values, we'd want to do it in that specific module.

chukarsten (approved):
Thanks for standardizing and reducing my OCD :)

evalml/tests/component_tests/test_exponential_smoothing_regressor.py
```diff
@@ -119,7 +119,9 @@ def test_feature_importance(ts_data):
     regressor = ExponentialSmoothingRegressor()
     with patch.object(regressor, "_component_obj"):
         regressor.fit(X, y)
-        assert regressor.feature_importance == np.zeros(1)
+        pd.testing.assert_series_equal(
+            regressor.feature_importance, pd.Series(np.zeros(1))
```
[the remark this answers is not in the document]
angela97lin: Filed https://github.com/alteryx/evalml/issues/3330 that we can use to track this! :)

evalml/model_understanding/graphs.py
angela97lin: Nope since `coef_` should be a `pd.Series` object now. Though we still want the name, so I can update this line to set the name instead.

evalml/model_understanding/graphs.py
```diff
@@ -1526,7 +1526,8 @@ def get_linear_coefficients(estimator, features=None):
...
+    coef_.index = features
```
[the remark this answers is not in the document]
angela97lin: I couldn't get it to work with them together, because of pandas behavior so it reindexes after and returns all NaNs if put on the same line (or that's my rough understanding)?https://pandas.pydata.org/docs/reference/api/pandas.Series.html

Issue #2548, opened 2021-07-23
ReadtheDocs failure 


angela97lin:
I noticed a weird error in https://github.com/alteryx/evalml/pull/2546, a small PR which moved `get_hyperparameter_ranges` to `PipelineBase`.

The failed ReadtheDocs build is here: https://readthedocs.com/projects/feature-labs-inc-evalml/builds/683782/, with the following error:
```

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/application.py", line 352, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 296, in build_update
    self.build(to_build,
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 310, in build
    updated_docnames = set(self.read())
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 417, in read
    self._read_serial(docnames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 438, in _read_serial
    self.read_doc(docname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 478, in read_doc
    doctree = read_doc(self.app, self.env, self.env.doc2path(docname))
  File "/home/docs/checkouts/readthedocs.org/user_builds/feature-labs-inc-evalml/envs/2546/lib/python3.8/site-packages/sphinx/io.py", line 221, in read_doc
    pub.publish()
[...]

PR #3307, opened 2022-02-04
Add a transformer to drop outliers

### LINKED ISSUE #2704: Add `DropOutlierRowsTransformer`
https://github.com/alteryx/evalml/pull/2692 introduced a generic `DropRowsTransformer` but based on the thread [here](https://github.com/alteryx/evalml/issues/2100#issuecomment-905875874), it'd be a good idea to introduce a data-agnostic component which detects and drops outliers.

This could probably be done by subclassing `DropRowsTransformer` and adding outlier detection logic!

angela97lin:
Closes #2704


docs/source/api_index.rst
```diff
@@ -410,13 +411,13 @@ Data Check Classes
...
     evalml.data_checks.IDColumnsDataCheck
     evalml.data_checks.TargetLeakageDataCheck
-    evalml.data_checks.OutliersDataCheck
     evalml.data_checks.NoVarianceDataCheck
     evalml.data_checks.ClassImbalanceDataCheck
     evalml.data_checks.MulticollinearityDataCheck
     evalml.data_checks.DateTimeFormatDataCheck
     evalml.data_checks.TimeSeriesParametersDataCheck
     evalml.data_checks.TimeSeriesSplittingDataCheck
+    evalml.data_checks.outliers_data_check.OutliersDataCheck
```
angela97lin, line 420: Another example of https://github.com/alteryx/evalml/issues/2548

angela97lin: (Failed build: https://readthedocs.com/projects/feature-labs-inc-evalml/builds/852750/)

evalml/pipelines/components/transformers/preprocessing/drop_outliers_transformer.py
```diff
@@ -0,0 +1,70 @@
+"""Transformer to drop rows specified by row indices."""
+from evalml.data_checks.outliers_data_check import OutliersDataCheck
+from evalml.pipelines.components.transformers import Transformer
+from evalml.utils import infer_feature_types
+
+
+class DropOutliersTransformer(Transformer):
```
freddyaboulton, line 7: I don't think it is a good idea to re-run the outlier calculation in this component. There is no guarantee that the same rows will be dropped when users go to re-train the dataset. This is the scenario I'm thinking about:

1. Users create a training/validation split
2. Users create pipelines with DropOutliersTransformers and pass those pipelines to AutoMLSearch
3. Outliers are calculated on each fold of the training set, those are dropped, and scores are calculated on the validation set.
4. Scores look good, user gets a pipeline with DropOutliersTransformer and calls fit on the entire dataset (training + validation concatted together).

In steps 3 and 4, different rows could identified as having outliers because the distribution of the data is slightly changing between the folds. I think it's problematic that when users fit on their entire data set, new rows could be dropped from the training set without their knowledge. 

So I think this component should not determine what an outlier is for users and then automatically drop. We should let users tell us what they think outliers are based on their knowledge of their own data. A way to do this is to accept an `outlier_ranges` parameter:

```python
outlier_ranges: {"col_name" {"low_bound": float, "high_bound": float}
```

Anything lower than `low_bound` or higher than `high_bound` is an outlier and that will be true for all datasets users train the pipeline on.

evalml/pipelines/components/transformers/preprocessing/drop_outliers_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+        all_rows_with_indices_set = set()
+        for row_indices in self.outlier_rows_dict.values():
+            all_rows_with_indices_set.update(row_indices)
+
+        all_rows_with_indices = list(all_rows_with_indices_set)
+        all_rows_with_indices.sort()
+        self.outlier_indices = all_rows_with_indices
+
+        X_t = X_t.drop(self.outlier_indices, axis=0)
+        X_t.ww.init()
```
freddyaboulton, line 65: This will reset the woodwork schema

jeremyliweishih, line 65: Would using `X_t.ww.drop()` instead make it so we don't need to re-init?

jeremyliweishih (approved):
looks great!

evalml/pipelines/components/transformers/preprocessing/drop_outliers_transformer.py
angela97lin: @freddyaboulton Interesting thoughts on this! I see the concern you're raising. I still wonder if there's a more "automatic" approach to dropping outliers (or if that's even desirable). Having an upper and lower bound requires the user to pass in the information to AutoMLSearch but I wish there were a way to automatically detect what good values for that is.

Is the scenario we outlined not the same for other components such as an Imputer with an imputation strategy of "most_frequent" for example? Where when we train, we could impute with what is considered the most frequent value found during training but once we concatenate training + validation, we end up with a different most_frequent value and end up imputing using a different value?

freddyaboulton: Yea that's a great point. It is similar to the Imputer case but I think there's still a difference - the concept of an outlier is always relative to the other data you have collected. What you consider an outlier based on 1000 row sample could change if you collect another 1000 rows. The rows that are missing on the first 1000 row sample will still be missing if you add the next 1000 rows to the original sample. Of course, the value you use to fill in those missing values could change depending on the data you call `fit` on but the subjectiveness of outliers makes me wary of automatically dropping rows in pipeline fitting.

angela97lin:
Closing with respect to https://github.com/alteryx/evalml/issues/2704#issuecomment-1040677347

Issue #2997, opened 2021-10-31
Add helper method to combine preprocessing pipeline + another pipeline


angela97lin:
Separating out work from https://github.com/alteryx/evalml/issues/2058, https://github.com/alteryx/evalml/pull/2968 tackled the first half of creating a preprocessing pipeline that will encompass all of the components created from data check actions.

This issue will tackle the second half: adding a utility method to combine a preprocessing pipeline with another pipeline.

freddyaboulton:
I think this shouldn't be iced too much! 

If I'm correct, @jeremyliweishih is working on a way to combine separate pipelines to put categorical feature selection "first" in the default algorithm.

In the design of #2511, I'm proposing we similarly combine two different pipelines into one pipeline.

@christopherbunn Already wrote a util method to combine a list of pipelines into one pipeline but tailoring the implementation specifically for stacked ensembles.

Seems like we need to have general way to combine component graphs! Let's do it! Otherwise we'll have four different util methods doing basically the same thing.

angela97lin:
@freddyaboulton since I just saw your comment: just chatted with @chukarsten, agreed that perhaps we should prioritize this higher, since this has shown up in multiple different different issues.

@chukarsten @dsherry Can we prioritize this for the upcoming sprint?

chukarsten:
@angela97lin @freddyaboulton @dsherry I'm taking the liberty and putting this one in our current sprint!

freddyaboulton:
Marking this as tech debt because there are two related issues that stem from not having an api for combining pipelines:

* #3076
* #2987 

I think we'll continue to have issues with our combined pipelines if we have different util methods for each case when a combined pipeline is desired.

For context, we currently have two ways of combining pipelines: 
* [make_stacked_ensemble_pipeline](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L391)
* [make_pipeline_from_multiple_graphs](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L495)

dsherry:
@angela97lin what's the status of this?

angela97lin:
There's no current work being done on this, as it wasn't necessary for the data check actions work. Still a valid tech debt issue but until we have plans to combine the preprocessing pipeline for applying actions with that generated with AutoMLSearch, this isn't relevant to data check actions specifically.

PR #3310, opened 2022-02-08
Added drop NaN component to some time series pipelines

### LINKED ISSUE #3259: Create drop NaN rows component and use for time series pipelines
As part of #3186, we will need to move out the drop NaN rows operation out of `TimeSeriesPipelineBase` into its own component. In addition, the time series pipeline construction needs to be updated so that this component is included for datasets with estimators that cannot handle NaN values.

- [ ] Create drop NaN component
   - [ ] Add parameter to drop only first `n` number of rows
- [ ] Update time series component graph construction to include new component before relevant estimators

christopherbunn:
Resolves #3259


evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -500,6 +500,18 @@ def _transform_parameters(self, pipeline, proposed_parameters):
...
             ):
                 component_parameters["columns"] = self._pipeline_params[name]["columns"]
+            # Check that Drop Row Transformer is before estimator
+            if (
+                is_time_series(self.problem_type)
+                and name == "Drop Rows Transformer"
+                and pipeline.component_graph.compute_order[-2] == name
+            ):
+                n_rows_to_drop = (
+                    self._pipeline_params["pipeline"]["max_delay"]
```
freddyaboulton, line 510: @christopherbunn not all of the pipelines have the same number of NaN rows. After the first batch, the number of delays, rolling window size will change as the search progresses. 

```python
from evalml.demos import load_weather

X, y = load_weather()

automl = AutoMLSearch(X, y, "time series regression",
                      max_batches=4, _automl_algorithm="iterative",
                      problem_configuration={"max_delay": 30, "forecast_horizon": 7, "gap": 1,
                                             "time_index": "Date"},
                      verbose=True)

automl.search()

light_gbm_pipelines = [automl.get_pipeline(i) for 
                       i in automl.full_rankings.loc[automl.full_rankings.pipeline_name.str.contains("LightGBM")].id]

[pl.parameters['Time Series Featurizer']['conf_level'] for pl in light_gbm_pipelines]


[pl.parameters['Time Series Featurizer']['rolling_window_size'] for pl in light_gbm_pipelines]

fitted_light_gbm_pipelines = [pl.fit(X, y) for pl in light_gbm_pipelines]

[pl.transform_all_but_final(X, y).isna().sum(axis=0).max() for pl in fitted_light_gbm_pipelines]
```

image

image

Note that the number of NaN rows is never `max_delay` + `forecast_horizon` + `gap`. I worry we may be throwing out too much data with the approach. I like the idea of only dropping NaNs for some estimators. Can we test with DropNaN component for the estimators that need it for more than two batches?

christopherbunn: Sure! Here's the results I got for first N vs. drop NaN only for select estimators. It looks like across the board we're getting better results with the drop NaN component than the drop rows component. As you pointed out, it's most likely because we're dropping rows that do not have NaN values. I'll adjust the preprocessing chain to use the drop NaN component for time series pipelines when needed. 

[drop_n_vs_drop_nan_necessary.html.zip](https://github.com/alteryx/evalml/files/8062520/drop_n_vs_drop_nan_necessary.html.zip)

evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
```diff
@@ -0,0 +1,50 @@
...
+            y (pd.Series, optional): Target data.
+
+        Returns:
+            (pd.DataFrame, pd.Series): Data with NaN rows dropped.
+        """
+        X_t = infer_feature_types(X)
+        y_t = infer_feature_types(y) if y is not None else None
+
+        X_t, y_t = drop_rows_with_nans(X_t, y_t)
+        X_t.ww.init()
```
freddyaboulton, line 47: @christopherbunn I think we need to store the schema prior to drop and then call init with the old schema so that we make sure we don't reset the schema.

freddyaboulton (approved):
@christopherbunn Looks good to me. I think we just need to make sure the `DropNaNRows...` doesn't reset the woodwork schema. Let's add a test for that too.

evalml/pipelines/utils.py
```diff
@@ -364,14 +389,21 @@ def _make_pipeline_time_series(
...
                 pipeline.name: "Not Known In Advance",
             },
         )
+        if need_drop_nan:
+            last_component_name = pipeline.component_graph.get_last_component().name
+            pipeline.component_graph.component_dict[estimator.name] = [
+                estimator,
+                last_component_name + ".x",
+                last_component_name + ".y",
+            ]
```
christopherbunn, line 406: I feel like this and lines 381-383 + 392 are a bit hacky, but I couldn't find a better function to do this.

christopherbunn (commented):
@freddyaboulton do you mind taking another look? the `DefaultAlgorithm` as the default got merged to main after your review and I ended up having to make some changes...

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    X_expected = pd.DataFrame({"a column": [3], "another col": ["c"]}, index=[2])
+    X_expected = X_expected.astype({"a column": "float", "another col": "string"})
+    X_expected_schema = X.ww.schema
+
+    y = pd.Series([3, 2, 1, np.NaN])
+    y = init_series(y, logical_type="IntegerNullable", semantic_tags="y_custom_tag")
+
+    y_expected = pd.Series([True], index=[2])
```
chukarsten, line 46: OK, so why does this need to be `True`?  I assume Woodwork is helping us out and casting this to an integer, but can't y_expected be 1?

And a check on my understanding of the drop nan rows transformer is that we want to drop rows where the target is NaN for sure and, I guess, rows where there are NaN values in the columns.  I would expect then that we lose the 1st row due to "a column", the 2nd row due to "another col" and the last row due to the target.  That leaves the 3rd row, index 2.  Did I get that right?

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -150,6 +151,18 @@ def test_make_pipeline(
...
+            drop_nan_rows_transformer = (
+                [DropNaNRowsTransformer]
+                if is_time_series(problem_type)
+                and estimator_class.model_family
+                in [
+                    ModelFamily.EXTRA_TREES,
+                    ModelFamily.RANDOM_FOREST,
+                    ModelFamily.LINEAR_MODEL,
+                    ModelFamily.DECISION_TREE,
+                ]
```
chukarsten, line 163: I think we should strongly consider what I mentioned before, about making this a function, so that we don't accidentally diverge in who needs this transformer.

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
freddyaboulton: I think the conversion to IntegerNullable in `init_series` will convert it to the correct value (1) but I agree it's confusing to have ti say `True` here.

evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+
+        X_t_schema = X_t.ww.schema
+        if y_t is not None:
+            y_t_logical = y_t.ww.logical_type
+            y_t_semantic = y_t.ww.semantic_tags
+
+        X_t, y_t = drop_rows_with_nans(X_t, y_t)
+        X_t.ww.init_with_full_schema(X_t_schema)
+        if y_t is not None:
+            y_t = init_series(y_t, logical_type=y_t_logical, semantic_tags=y_t_semantic)
```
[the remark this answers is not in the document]
freddyaboulton: Yep and thanks for adding the test!

evalml/pipelines/utils.py
```diff
@@ -356,6 +378,9 @@ def _make_pipeline_time_series(
         kina_component_graph = PipelineBase._make_component_dict_from_component_list(
             kina_preprocessing
         )
+        need_drop_nan = (
```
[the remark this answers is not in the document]
christopherbunn: Agreed, I separated this logic out into `estimator_unable_to_handle_nans()` in pipeline utils/

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
christopherbunn: Yeah, this was left over from experimenting with different data types, I intended this to be 1.

@chukarsten your explanation is correct!

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+    drop_rows_transformer.fit(X)
+    transformed_X, _ = drop_rows_transformer.transform(X)
+    assert_frame_equal(transformed_X, X_expected)
+
+    drop_rows_transformer = DropNaNRowsTransformer()
+    fit_transformed_X, _ = drop_rows_transformer.fit_transform(X)
+    assert_frame_equal(fit_transformed_X, X_expected)
+
+
+def test_drop_rows_transformer_retain_ww_schema():
```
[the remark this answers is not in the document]
christopherbunn: I agree, I adapted this from `DropRowsTransformer` tests . I think it would be good to streamline this into a single component test.

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+
+from evalml.pipelines.components.transformers.preprocessing import (
+    DropNaNRowsTransformer,
+)
+from evalml.utils.woodwork_utils import _schema_is_equal
+
+
+def test_drop_rows_transformer():
+    # Expecting float because of np.NaN values
+    X = pd.DataFrame({"a column": [np.NaN, 2, 3], "another col": [4, np.NaN, 6]})
```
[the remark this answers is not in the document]
christopherbunn: Added coverage for `pd.NA` and `np.NaN`.

evalml/pipelines/utils.py
freddyaboulton: I think this is another example of why we need this: https://github.com/alteryx/evalml/issues/2997

evalml/pipelines/components/utils.py
```diff
@@ -98,6 +98,28 @@ def get_estimators(problem_type, model_families=None):
     return estimator_classes
 
 
+def estimator_unable_to_handle_nans(estimator_class):
```
freddyaboulton, line 101: It may be safer to implement this as the opposite, e.g. "able_to_handle_nans". The concern I have with doing it like this is that if we add a new estimator, this function will return `False` if we forget to update this file and a false negative in this case is costly because it will mean AutoMLSearch will error out.

Not blocking - let's handle it in a follow-up.

freddyaboulton (approved):
Thank you @christopherbunn !

evalml/tests/component_tests/test_drop_nan_rows_transformer.py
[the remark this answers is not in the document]
chukarsten: Cool, quickly wrote up this issue : #3338

chukarsten (approved):
Looks good, thanks for making the changes and making them so quickly!

PR #3312, opened 2022-02-08
Add auto approve dependency workflow schedule

gsheni:
- For some reason, the auto approve workflow does not get kickoff (with latest dependency PRs)
- This PR makes it so that it runs every 30 mins (which we knows works)


freddyaboulton (approved):
Auto approve

Issue #3275, opened 2022-01-24
Add support for pandas 1.4.0


freddyaboulton:
Looks like the root cause may be that converting a nullable type to `Categorical` preserves the nullable types in the pandas Categorical categories. On the top is the latest pandas and on the bottom is 1.3.4 pandas.

![image](https://user-images.githubusercontent.com/41651716/150818958-adb14e58-1d51-4632-baad-ff938d238618.png)

This causes np.asarray calls in our imputers to introduce the new pandas null value place-holder, `<NA>` which throws off sklearn:

![image](https://user-images.githubusercontent.com/41651716/150819366-a6246c7a-2ac3-4e67-b31c-6df3c7cae7fb.png)

_Originally posted by @freddyaboulton in https://github.com/alteryx/evalml/issues/3272#issuecomment-1020261986_

freddyaboulton:
Xgboost issue may be fixed here: https://github.com/dmlc/xgboost/issues/7593

chukarsten:
I'm going to have to pick this up to satisfy support for nullable types, @dsherry !  Just FYI, moving to this sprint.

thehomebrewnerd:
@chukarsten Just a heads up, Featuretools doesn't yet fully support `pandas==1.4.0` either. It *might* work for you though, but no guarantees at this point.

https://github.com/alteryx/featuretools/issues/1865

chukarsten:
Ultimately, the problem here is that sklearn can't recognize pd.NA as a type of missing value.  When the imputer looks for a missing value, it generates a mask of the dataframe, where it can recognize None and np.nan, but running the new pandas.NA through that same function returns a simple boolean instead of a similarly shaped array with boolean values in it, with True in the cells for nan values.

I think in the long term, we should submit an issue for sklearn and perhaps a PR to fix this in sklearn.  The function that they have in place for trying to root out NaN values in an array with a dtype of "object" is pretty simple and could use some love/tests. 

```
def _object_dtype_isnan(X):
    return X !=  X
```

Still working on a short term fix that makes sense.

chukarsten:
Looks like the XGBoost warnings are remedied, per https://github.com/dmlc/xgboost/pull/7595!  Just need to wait for a release from XGBoost.

PR #3324, opened 2022-02-14
Pandas 1.4.0 Support

chukarsten:
Fixes: #3275 

Like the issue says, there are basically three areas to address:

1. **XGBoost future warning:** This is handled by expecting this test to fail on this PR and then, when XGBoost does their next release, fixing the auto dependency update as the test will no longer fail as there will be no more warning.
2. **Results of email featurization:**  The underlying problem is that featuretools is returning a CategoricalDtype with the dtype of "string" when we, in our test, build an expected dataframe with CategoricalDtype with dtype of "object."  I'm not entirely sure why this is the case as I've stepped through both the woodwork accessor and featuretools trying to find out how this happens.  I've also tried to reproduce making a CategoricalDtype with categories that have dtype "string" but to no avail.  I've rewritten the equality check here to check the values, column names, indices and logical types separately as a stop-gap until this behavior is understood.
3. **Imputer/SimpleImputer failures:** We decided that the NaturalLanguage should not be imputed, so it is imputed, no longer.

### DIFF
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -77,13 +77,20 @@ class SimpleImputer(Transformer):
         nan_ratio = X.ww.describe().loc["nan_count"] / X.shape[0]
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
 
-        X, _  = self._drop_natural_language_columns(X)
-        X= self._set_boolean_columns_to_categorical(X)
+        X, _ = self._drop_natural_language_columns(X)
+        X = self._set_boolean_columns_to_categorical(X)
 
-        unique_logical_types = set([x.type_string for x in list(X.ww.logical_types.values())])
-        if len(unique_logical_types) > 1 and self.impute_strategy == "constant":
+        unique_logical_types = set(
+            [x.type_string for x in list(X.ww.logical_types.values())]
+        )
+        if (
+            len(unique_logical_types) > 1
+            and self.impute_strategy == "constant"
+            and "naturallanguage" in unique_logical_types
+        ):
             raise ValueError(
-                "SimpleImputer received data with multiple logical types, but 'constant' imputation strategy.")
+                "SimpleImputer received data with multiple logical types, but 'constant' imputation strategy."
+            )
 
         # If the Dataframe only had natural language columns, do nothing.
         if X.shape[1] == 0:
@@ -116,14 +123,18 @@ class SimpleImputer(Transformer):
         X_t, natural_language_cols = self._drop_natural_language_columns(X)
         if X_t.shape[-1] == 0:
             return X
-        not_all_null_or_natural_language_cols = [col for col in not_all_null_cols if col not in natural_language_cols]
+        not_all_null_or_natural_language_cols = [
+            col for col in not_all_null_cols if col not in natural_language_cols
+        ]
 
         X_t = self._component_obj.transform(X_t)
         X_t = pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)
 
-        #Add back in natural language columns, unchanged
+        # Add back in natural language columns, unchanged
         if len(natural_language_cols) > 0:
-            X_t = pd.merge(X_t, X[natural_language_cols], left_index=True, right_index=True)
+            X_t = pd.merge(
+                X_t, X[natural_language_cols], left_index=True, right_index=True
+            )
 
         if not_all_null_or_natural_language_cols:
             X_t.index = original_index
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -1,7 +1,7 @@
 """Component that imputes missing data according to a specified imputation strategy."""
 import pandas as pd
 from sklearn.impute import SimpleImputer as SkImputer
-from woodwork.logical_types import NaturalLanguage
+from woodwork.logical_types import NaturalLanguage, BooleanNullable, Boolean
 
 from evalml.pipelines.components.transformers import Transformer
 from evalml.utils import infer_feature_types
@@ -27,6 +27,7 @@ class SimpleImputer(Transformer):
     def __init__(
         self, impute_strategy="most_frequent", fill_value=None, random_seed=0, **kwargs
     ):
+        self.impute_strategy=impute_strategy
         parameters = {"impute_strategy": impute_strategy, "fill_value": fill_value}
         parameters.update(kwargs)
         imputer = SkImputer(strategy=impute_strategy, fill_value=fill_value, **kwargs)
@@ -34,6 +35,31 @@ class SimpleImputer(Transformer):
         super().__init__(
             parameters=parameters, component_obj=imputer, random_seed=random_seed
         )
+
+    def _get_columns_of_type(self, X, ww_dtype):
+        return  [
+            col
+            for col, ltype in X.ww.logical_types.items()
+            if type(ltype) == ww_dtype
+        ]
+
+    def _drop_natural_language_columns(self, X):
+        # Not using select because we just need column names, not a new dataframe
+        natural_language_columns = self._get_columns_of_type(X, NaturalLanguage)
+        if natural_language_columns:
+            X = X.ww.copy()
+            X = X.ww.drop(columns=natural_language_columns)
+        return X
+
+    def _set_boolean_columns_to_categorical(self, X):
+        boolean_null_columns = self._get_columns_of_type(X, BooleanNullable)
+        boolean_columns = self._get_columns_of_type(X, Boolean)
+        boolean_columns += boolean_null_columns
+        if boolean_columns:
+            X = X.ww.copy()
+            X.ww.set_types({col: "Categorical" for col in boolean_columns})
+        return X
+
 
     def fit(self, X, y=None):
         """Fits imputer to data. 'None' values are converted to np.nan before imputation and are treated as the same.
@@ -46,22 +72,25 @@ class SimpleImputer(Transformer):
             self
         """
         X = infer_feature_types(X)
+
         nan_ratio = X.ww.describe().loc["nan_count"] / X.shape[0]
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
 
-        # Not using select because we just need column names, not a new dataframe
-        natural_language_columns = [
-            col
-            for col, ltype in X.ww.logical_types.items()
-            if type(ltype) == NaturalLanguage
-        ]
-        if natural_language_columns:
-            X = X.ww.copy()
-            X.ww.set_types({col: "Categorical" for col in natural_language_columns})
+        # # Convert all bool dtypes to category for fitting
+        # if (X.dtypes == bool).all():
+        #     X = X.astype("category")
 
-        # Convert all bool dtypes to category for fitting
-        if (X.dtypes == bool).all():
-            X = X.astype("category")
+        # Determine if imputer is being used with incompatible imputation strategies
+        boolean_columns = self._get_columns_of_type(X, BooleanNullable)
+        if self.impute_strategy in ["median", "mean"] and len(boolean_columns) > 0:
+            raise ValueError(f"Cannot use {self.impute_strategy} strategy with non-numeric data: {boolean_columns} contain boolean values and cannot be imputed with the 'median' or 'mode' strategy.")
+
+        X = self._drop_natural_language_columns(X)
+        X = self._set_boolean_columns_to_categorical(X)
+
+        # If the Dataframe only had one natural language column, do nothing.
+        if X.shape[1] == 0:
+            return self
 
         self._component_obj.fit(X, y)
         return self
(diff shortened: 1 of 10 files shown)


evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -353,86 +362,93 @@ def test_simple_imputer_with_none():
     assert_frame_equal(expected, transformed, check_dtype=False)
 
 
-def test_simple_imputer_supports_natural_language_constant():
+@pytest.mark.parametrize("na_type", ["python_none", "numpy_nan", "pandas_na"])
+@pytest.mark.parametrize("data_type", ["Categorical", "NaturalLanguage"])
+def test_simple_imputer_supports_natural_language_and_categorical_constant(
```
chukarsten, line 367: Since these tests failed, I just refactored them to be more descriptive per the parameters in case they should fail again.

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -353,86 +362,93 @@ def test_simple_imputer_with_none():
...
     expected = pd.DataFrame(
         {
-            "cat with None": pd.Series(
-                ["a", "b", "a", "placeholder"], dtype="category"
-            ),
-            "natural language col": pd.Series(
-                ["free-form text", "will", "be imputed", "placeholder"], dtype="string"
+            "Categorical": pd.Series(["a", "b", "a", "placeholder"], dtype="category"),
+            "NaturalLanguage": pd.Series(
+                ["free-form text", "will", "be imputed", pd.NA], dtype="string"
```
chukarsten, line 390: Might be a little redundant, but the imputer now ignores the natural language columns.

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -353,86 +362,93 @@ def test_simple_imputer_with_none():
...
+    logical_type = {
+        "Integer": Integer,
+        "Double": Double,
+        "Categorical": Categorical,
+        "NaturalLanguage": NaturalLanguage,
+        "Boolean": Boolean,
+    }[logical_type]
     y = pd.Series([1, 2, 1])
-    override_types = [Integer, Double, Categorical, NaturalLanguage, Boolean]
-    for logical_type in override_types:
```
chukarsten, line 408: Refactored this test so that all the parameters are listed in the test failure sections and it's easier to understand what's failing, specifically which override type.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -86,17 +103,13 @@ def transform(self, X, y=None):
...
 
-        # Not using select because we just need column names, not a new dataframe
-        X.ww.set_types(
-            {
-                col: "Categorical"
-                for col, ltype in X.ww.logical_types.items()
-                if isinstance(ltype, NaturalLanguage)
-            }
-        )
+        X_t, natural_language_columns = self._drop_natural_language_columns(X)
```
freddyaboulton, line 106: `natural_language_columns` is unused. Can we drop it from the return type?

evalml/tests/component_tests/test_simple_imputer.py
freddyaboulton: I think it only ignores natural language columns if the only column in the dataframe is a natural language column?

```python
from evalml.pipelines.components import SimpleImputer
import pandas as pd
import woodwork as ww
import pytest

X = pd.DataFrame({"a": ["foo", "bar", "baz"], "b": [1, 2, 3]})
X.ww.init(logical_types={"a": "NaturalLanguage"})
si = SimpleImputer()

with pytest.raises(ValueError):
    si.fit_transform(X)
```

I think it might be better to raise an error if non-imputable types are present?

evalml/tests/component_tests/test_imputer.py
```diff
@@ -521,53 +521,87 @@ def test_imputer_does_not_erase_ww_info():
...
-            )
-        ),
+        "integer_data",
+        "float_data",
+        "categorical_data",
+        "boolean_data",
     ],
 )
-@pytest.mark.parametrize("has_nan", [True, False])
+@pytest.mark.parametrize(
```
freddyaboulton, line 558: Thanks for reorganizing this

evalml/tests/component_tests/test_ft_transform_primitive_components.py
```diff
@@ -225,10 +225,23 @@ def test_component_fit_transform(
     component.fit(data)
     new_X = component.transform(data)
 
-    pd.testing.assert_frame_equal(new_X, expected)
+    # TODO: Understand why the "EMAIL_ADDRESS_TO_DOMAIN(email)" engineered feature
```
freddyaboulton, line 228: I think the new change in behavior with pandas 1.4.0 is that the `categories` attribute of the categorical dtype now includes "original" feature values. Since the physical type of `URL`/`EmailAddress` logical types is "string" (nullable), that's why the categories  have nullable values.

I think we can get around this behavior if we update the `make_answer` functions but I think what we have here is fine too.

freddyaboulton (commented):
@chukarsten Looks good to me! I just want to synch up on the expected behavior of natural language in the simple imputer prior to merge.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -86,17 +103,13 @@ def transform(self, X, y=None):
...
+        if X_t.shape[-1] == 0:
+            return X
 
-        X_t = self._component_obj.transform(X)
+        X_t = self._component_obj.transform(X_t)
```
jeremyliweishih, line 110: think we need to add back `natural_language_columns` after this step.

evalml/tests/component_tests/test_simple_imputer.py
jeremyliweishih: @freddyaboulton I'm more of a fan of ignoring the columns if our imputer explicitly (we should not this in the docstring if we pursue this path) does not support natural language columns and that's the design decision I made with all of the select column transformers. If we raise an error it would mean that we would error out during search if the input contains a natural language column with nulls (but I'm unsure how the `NaturalLanguageFeaturizer` deals with nulls. Maybe raising a warning is a happy medium? 

@chukarsten if we want to keep with the ignoring path - we would need to add the natural language columns back in after `transform`-ing as well.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -44,24 +66,19 @@ def fit(self, X, y=None):
...
-        if natural_language_columns:
-            X = X.ww.copy()
-            X.ww.set_types({col: "Categorical" for col in natural_language_columns})
+        X, _ = self._drop_natural_language_columns(X)
+        X = self._set_boolean_columns_to_categorical(X)
 
-        # Convert all bool dtypes to category for fitting
-        if (X.dtypes == bool).all():
-            X = X.astype("category")
+        # If the Dataframe only had one natural language column, do nothing.
```
angela97lin, line 79: Nit pick, could have been more than one natural language column right? This happens whenever our input data *only* has natural language columns

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -44,24 +66,19 @@ def fit(self, X, y=None):
...
-        # Not using select because we just need column names, not a new dataframe
-        natural_language_columns = [
-            col
-            for col, ltype in X.ww.logical_types.items()
-            if type(ltype) == NaturalLanguage
-        ]
-        if natural_language_columns:
-            X = X.ww.copy()
-            X.ww.set_types({col: "Categorical" for col in natural_language_columns})
+        X, _ = self._drop_natural_language_columns(X)
```
angela97lin, line 76: Similar to Freddy's comment, do we use the second output ever or can we remove it altogether?

evalml/tests/component_tests/test_imputer.py
```diff
@@ -521,53 +521,87 @@ def test_imputer_does_not_erase_ww_info():
     mock_transform.call_args[0][0].ww.schema == df_holdout.ww[["b"]].ww.schema
 
 
+def get_int_df():
```
angela97lin, line 524: omega nit-pick / general curiosity but should these be fixtures? (not really sure if there's a clear yay or nay here, just wanted to raise awareness since that's how we've organized our other dummy datasets but it doesn't matter to me either way)

evalml/tests/component_tests/test_ft_transform_primitive_components.py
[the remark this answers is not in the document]
chukarsten: https://github.com/alteryx/evalml/issues/3331 beat ya to it ;)

evalml/tests/component_tests/test_imputer.py
```diff
@@ -521,53 +521,87 @@ def test_imputer_does_not_erase_ww_info():
...
 def test_imputer_woodwork_custom_overrides_returned_by_components(
-    X_df, has_nan, numeric_impute_strategy
+    data, logical_type, has_nan, numeric_impute_strategy
 ):
+    X_df = {
+        "integer_data": get_int_df,
+        "float_data": get_float_df,
+        "categorical_data": get_category_df,
+        "boolean_data": get_bool_df,
+    }[data]()
```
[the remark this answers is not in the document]
chukarsten: It...does things!

evalml/pipelines/components/transformers/imputers/simple_imputer.py
chukarsten: Thank you :)

evalml/pipelines/components/transformers/imputers/simple_imputer.py
chukarsten: We can, it was the result of my over-engineering and failure to scale all the way back *heavy breathing*

bchen1116 (approved):
These changes are looking good! Agreed about moving some of the dfs to the `conftest.py` file, as well as left a few questions. 

As per Freddy's comment [here](https://github.com/alteryx/evalml/pull/3324/files#r807340893), it could be useful to add a non-categorical/NL column to the `test_simple_imputer_ignores_natural_language` in `test_simple_imputer` to ensure that this continues to work as we want.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
chukarsten: Donezo

evalml/tests/component_tests/test_ft_transform_primitive_components.py
chukarsten: Do you know how to do that?  I pinged the woodwork team for how we can add in that specific column with the "string" type, but I can't seem to do it.

evalml/tests/component_tests/test_imputer.py
chukarsten: Cool, so I just moved up `test_imputer`'s imputer_test_data to `conftest.py` and I use that instead.  Sniped these garbage data frames as I figured I should drink my own koolaid and practice what I preach 😂

evalml/tests/component_tests/test_imputer.py
```diff
@@ -521,53 +521,87 @@ def test_imputer_does_not_erase_ww_info():
...
+        "float_data",
+        "categorical_data",
+        "boolean_data",
     ],
 )
-@pytest.mark.parametrize("has_nan", [True, False])
+@pytest.mark.parametrize(
+    "logical_type", ["Integer", "Double", "Categorical", "NaturalLanguage", "Boolean"]
+)
+@pytest.mark.parametrize("has_nan", ["has_nan", "no_nans"])
```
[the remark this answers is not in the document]
chukarsten: I did it so that the output in either pytest terminal or PyCharm looks like this:

image

The parameters used to just have like "-True-" or "-False-" which made the debugger wonder: "what parameter is failing?"  I refactored them this way so that you could more easily see that data that "has_nan" is buggy rather than seeing that data with "True" was buggy then having to constantly remember that the parameter in that position is "has_nan".

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -11,6 +12,14 @@
     NaturalLanguage,
 )
 
+from .test_imputer import (
```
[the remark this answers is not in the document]
chukarsten: I've wiped these dfs from existence.  It's like they've never happened.

evalml/tests/component_tests/test_ft_transform_primitive_components.py
freddyaboulton: I think like this:

image

freddyaboulton: Not blocking though. The current equality logic in the test is sufficient.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
chukarsten: Donezo

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -35,6 +35,28 @@ def __init__(
...
         )
 
+    def _get_columns_of_type(self, X, ww_dtype):
+        return [
+            col for col, ltype in X.ww.logical_types.items() if type(ltype) == ww_dtype
+        ]
+
+    def _drop_natural_language_columns(self, X):
+        # Not using select because we just need column names, not a new dataframe
+        natural_language_columns = self._get_columns_of_type(X, NaturalLanguage)
```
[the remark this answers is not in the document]
chukarsten: Donezo!  Thanks for the catch.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -86,18 +99,24 @@ def transform(self, X, y=None):
...
+        not_all_null_or_natural_language_cols = [
+            col for col in not_all_null_cols if col not in natural_language_cols
+        ]
+
+        X_t = self._component_obj.transform(X_t)
+        X_t = pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)
+
+        # Add back in natural language columns, unchanged
+        if len(natural_language_cols) > 0:
+            X_t = pd.merge(
```
freddyaboulton, line 115: I think `merge` and `get_subset_schema` is forcing us to re-infer the natural language columns. I think we can move `X_t.ww.init` above this line and use `ww.concat_columns`?

PR #3328, opened 2022-02-14
Shorten results output in docs

### LINKED ISSUE #3032: Access Raw Results Output is too long
Just reviewing the docs and found this under the AutoML User Guide:

image

We should figure out a way to deal with this kind of thing.  I think a couple of options here are:

1. Modifying the cell to only show the first few keys or so of the output.
2. Modifying the output cell so that it's a scrollable window that's limited in size.

bchen1116:
fix #3032 

Rather than editing the output of the results themselves, I figured the better approach would be to truncate the cell output for this.

image

https://feature-labs-inc-evalml--3328.com.readthedocs.build/en/3328/user_guide/automl.html#Access-raw-results

### DIFF
--- a/docs/source/user_guide/automl.ipynb
+++ b/docs/source/user_guide/automl.ipynb
@@ -653,7 +653,11 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "output_scroll"
+    ]     
+   },
    "outputs": [],
    "source": [
     "automl.results"
--- a/docs/source/user_guide/automl.ipynb
+++ b/docs/source/user_guide/automl.ipynb
@@ -653,14 +653,13 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "tags": [
-     "output_scroll"
-    ]     
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
-    "automl.results"
+    "import pprint\n",
+    "pp = pprint.PrettyPrinter(indent=0, width=100, depth=3, compact=True, sort_dicts=False)\n",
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Thanks @bchen1116 ! This looks good to me. If we decide we need a pprint call in more places in the docs we can invest in trying to find a way to apply the pretty-printing automatically.

docs/source/user_guide/automl.ipynb
```diff
@@ -657,7 +657,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "automl.results"
+    "import pprint\n",
+    "pp = pprint.PrettyPrinter(indent=0, width=100, depth=3, compact=True, sort_dicts=False)\n",
```
jeremyliweishih, line 661: could we set `depth=2` and then showcase just one pipelines results with:
`automl.results['pipeline_results'][0]`?  I still think the output is rather long and takes multiple scrolls to get through.

bchen1116: My main worry is that `depth=2` gives users very little information:
image

I can definitely try to combine it with your latter suggestion and see how it works!

jeremyliweishih: ah true that doesn't look the best.. lmk if adding more information will make it better but def not blocking!

angela97lin (approved):
Thanks @bchen1116! Agreed with @freddyaboulton that it could be a good idea to do this automagically if we run into it more in the future but this is great 😁

docs/source/user_guide/automl.ipynb
bchen1116: image

I like the previous better. will change back!

PR #3339, opened 2022-02-17
Make target optional for NoVarianceDataCheck

### LINKED ISSUE #3337: NoVarianceDataCheck should allow the target to be optional

```python
from evalml.data_checks import NoVarianceDataCheck
import pytest

X, _ = load_diabetes()

with pytest.raises(AttributeError, match="'NoneType' object has no attribute 'ww'"):
    NoVarianceDataCheck().validate(X, None)
```

Calculating whether a column is "no variance" does not depend on a target variable being defined. This is a break from out DataChecks api of having a default value of `None` for the target variable.

freddyaboulton:
### Pull Request Description
Fixes #3337 

-----


evalml/data_checks/no_variance_data_check.py
```diff
@@ -153,7 +155,10 @@ def validate(self, X, y):
         messages = []
 
         X = infer_feature_types(X)
-        y = infer_feature_types(y)
+        if y is not None:
+            y = infer_feature_types(y)
+        else:
+            y = infer_feature_types(pd.Series([1, 2]))
```
chukarsten, line 161: Is this intentional?

freddyaboulton: Rather than adding an `if` below and having to indent all the target-related logic (or returning early), I figured I would just use a dummy target that is not "no variance"

angela97lin: Hmm, I don't think checking below and then returning early for the target-related logic is too bad of an idea but I'll leave it up to you guys! 

I personally would be more okay hardcoding if this were in the tests and not in the implementation :)

angela97lin (approved):
Left my thoughts on the hardcoding for impl that @chukarsten raised but whatever you guys decide SGTM. :shipit: 🚢 !

PR #3341, opened 2022-02-17
Add test_size parameter to ClassImbalanceDataCheck

### LINKED ISSUE #3334: Add optional test_size parameter to ClassImbalanceDataCheck
- As a user of EvalML, I wish the ClassImbalanceDataCheck would use the holdout percentage to see if class imbalance would exist after splitting the data into training and validation sets.

Consider the following case on the fish dataset. Using the entire dataset would not be a problem but if you split the dataset into two halves, one of the classes would have less than the number of cv folds! 

image

Here's what I propose:

We add an optional holdout_pct/test_size parameter to the `ClassImbalanceDataCheck`. If None, the `ClassImbalanceDataCheck` behaves as it currently does. If not, we use the test_size to modify the counts of the classes in the class imbalance calculations.

freddyaboulton:
### Pull Request Description
Fixes #3334

-----

### DIFF
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -26,6 +26,12 @@ class ClassImbalanceDataCheck(DataCheck):
         num_cv_folds (int): The number of cross-validation folds. Must be positive. Choose 0 to ignore this warning. Defaults to 3.
         test_size (None, float, int): Percentage of test set size. Used to calculate class imbalance prior to splitting the
             data into training and validation/test sets.
+
+    Raises:
+        ValueError: If threshold is not within 0 and 0.5
+        ValueError: If min_samples is not greater than 0
+        ValueError: If number of cv folds is negative
+        ValueError: If test_size is not between 0 and 1
     """
 
     def __init__(self, threshold=0.1, min_samples=100, num_cv_folds=3, test_size=None):
--- a/evalml/data_checks/class_imbalance_data_check.py
+++ b/evalml/data_checks/class_imbalance_data_check.py
@@ -2,6 +2,8 @@
 
 Use for classification problems.
 """
+import numpy as np
+
 from evalml.data_checks import (
     DataCheck,
     DataCheckError,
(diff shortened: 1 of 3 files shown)


evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -22,9 +24,11 @@ class ClassImbalanceDataCheck(DataCheck):
         min_samples (int): The minimum number of samples per accepted class. If the minority class is both below the threshold and min_samples,
             then we consider this severely imbalanced. Must be greater than 0. Defaults to 100.
         num_cv_folds (int): The number of cross-validation folds. Must be positive. Choose 0 to ignore this warning. Defaults to 3.
+        test_size (None, float, int): Percentage of test set size. Used to calculate class imbalance prior to splitting the
+            data into training and validation/test sets.
```
chukarsten, line 28: Hyper nit - won't you need a `Raises:` block for the new ValueError?  I guess not given the existing ValueError isn't complained about.

chukarsten (approved):
All this increasing of the targets make me wonder if there's any consolidation that can be done in the `evalml/tests/data_checks_tests/test_class_imbalance_data_check.py` module to perhaps push some of those synthetic targets out of the individual tests and either to the module level or to a `conftest.py` further up.   But other than that, looks gucci.

evalml/data_checks/class_imbalance_data_check.py
angela97lin: +1 to ValueError, lame that it's not caught by our linter because we move our docstring from __init__ to our class docstring (IIRC for docs) so it doesn't automatically pick it up 😢

angela97lin (approved):
Clever use of numpy array multiplication to update the tests, LGTM! Thanks @freddyaboulton 😁

evalml/data_checks/class_imbalance_data_check.py
freddyaboulton: My guess is that darglint doesn't check for raises blocks in constructors.  @angela97lin do you know?

angela97lin: Yeah @freddyaboulton, I don't think it does a good job for class docstrings in general (I think I recently came across a case where we added a new parameter but it didn't complain about the missing docstring).

If we had this docstring in the constructor (__init__) it might actually pick it up, but IIRC, then our docs won't look like this: https://evalml.alteryx.com/en/stable/autoapi/evalml/data_checks/class_imbalance_data_check/index.html#evalml.data_checks.class_imbalance_data_check.ClassImbalanceDataCheck

We'd have to add back __init__ and click on the __init__ method link specifically to see the parameters, rather than just at the class page.

PR #3348, opened 2022-02-19
Shuffling preprocessing functions

chukarsten:
Refactoring some preprocessing components to be a little more intuitive.  To me, at least.


chukarsten:
This is probably not worth the effort

PR #3354, opened 2022-03-02
Update DateTimeFormatDataCheck to flag all errors

eccabay:
Instead of returning just an error if NaNs are found, `DateTimeFormatDataCheck` will now continue on and raise any other errors as well


freddyaboulton (approved):
Thank you for making the change @eccabay !!

angela97lin (approved):
LGTM, cool that we can do this 😁 Thanks @eccabay!

PR #3360, opened 2022-03-04
Fix threshold return in AutoMLSearch

### LINKED ISSUE #3295: AutoMLSearch uses optimal threshold of last fold in get_pipeline for binary classification
`AutoMLSearch.get_pipeline` will set the optimal threshold as the threshold found in the last cv fold. However, we store the optimal threshold for all cv folds. It's not clear that the threshold on the last fold is the best one or the one that should be used in "production".

```python

X, y = load_fraud(1000)

automl = AutoMLSearch(X, y, "binary", verbose=True)
automl.search()

pipeline = automl.get_pipeline(1)
threshold_last_fold = automl.results['pipeline_results'][1]['cv_data'][-1]['binary_classification_threshold']

assert pipeline == threshold_last_fold
```

```python
print([i['binary_classification_threshold'] for i in automl.results['pipeline_results'][1]['cv_data']])
```

```
[0.2739189310625345, 0.3644718206383961, 0.2954551911484392]
```

bchen1116:
fix #3295 

Don't use the threshold of the last cv fold. Instead, we set to none.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -1276,7 +1276,7 @@ class AutoMLSearch:
             )
         new_pipeline = pipeline.new(parameters, random_seed=self.random_seed)
         if is_binary(self.problem_type):
-            new_pipeline.threshold = pipeline.threshold
+            new_pipeline.threshold = None
         return new_pipeline
 
     def describe_pipeline(self, pipeline_id, return_dict=False):
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,7 @@
     * Enhancements
         * Added caching capability for ensemble training during ``AutoMLSearch`` :pr:`3257`
     * Fixes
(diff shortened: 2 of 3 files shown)


jeremyliweishih (approved):
LGTM, anywhere we should add more info or change in our docs? Not sure if we have anything stating we recommend retraining pipelines after search.

evalml/automl/automl_search.py
```diff
@@ -1276,7 +1276,7 @@ def get_pipeline(self, pipeline_id):
             )
         new_pipeline = pipeline.new(parameters, random_seed=self.random_seed)
         if is_binary(self.problem_type):
-            new_pipeline.threshold = pipeline.threshold
+            new_pipeline.threshold = None
```
freddyaboulton, line 1279: I think it might be better to take the average threshold over none? My concern with setting it to None is that it won’t reflect the threshold used to compute the scores in the leaderboard. In my opinion, taking the average threshold is closer to matching the mean_cv_score we use to sort the leaderboard.

PR #3374, opened 2022-03-14
Add a TimeSeriesImputer component

### LINKED ISSUE #3370: Add a TimeSeriesImputer Component

eccabay:
Closes #3370 

Goes hand in hand with #3376. The time series regularizer will create a perfectly-spaced time column within a dataframe, this will fill in any gaps that were created by the regularizer or were pre-existing,


evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,478 @@
...
+        "categorical_impute_strategy": ["backwards_fill", "forwards_fill"],
+        "numeric_impute_strategy": ["backwards_fill", "forwards_fill", "interpolate"],
+        "target_impute_strategy": ["backwards_fill", "forwards_fill", "interpolate"],
+    }
+    assert imputer.name == "Time Series Imputer"
+    assert imputer.parameters == expected_parameters
+    assert imputer.hyperparameter_ranges == expected_hyperparameters
+
+
+def test_numeric_only_input(imputer_test_data):
```
freddyaboulton, line 69: I think we should come up with a test case where forwards_fill, backwards_fill, interpolate yield different answers. Otherwise we could easily introduce a bug and not notice.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,199 @@
...
+            df = pd.DataFrame(index=X.index)
+            df.ww.init()
+            return df, y
+
+        X_no_all_null = X.ww.drop(self._all_null_cols)
+
+        if self._forwards_cols is not None:
+            X_forward = X.ww[self._forwards_cols.tolist()]
+            imputed = X_forward.pad()
+            imputed = imputed.bfill()  # Fill in the first value, if missing
```
freddyaboulton, line 173: Why isn't this `imputed.ffill()` ?

freddyaboulton: Ah, I think I misunderstood what `pad` was doing!

freddyaboulton: Rather than calling `bfill` or `pad` on the entire dataset just to fill in the first/last value, I wonder if we could do something more targeted to prevent having to create a whole new copy of the data.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Args:
+            X (pd.DataFrame, np.ndarray): The input training data of shape [n_samples, n_features]
+            y (pd.Series, optional): The target training data of length [n_samples]
+
+        Returns:
+            self
+        """
+        forwards_fill_cats = (
+            ["category", "boolean"]
```
freddyaboulton, line 98: I think this has to be anything that's not numeric. Theoretically, someone could pass in a dirty dataset with email features to the regularizer since this will happen early in the process, probably prior to modeling. I think those cases are unlikely but it should be easy to extend what you have now to handle them.

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,478 @@
...
+    ],
+)
+@pytest.mark.parametrize(
+    "logical_type", ["Integer", "Double", "Categorical", "NaturalLanguage", "Boolean"]
+)
+@pytest.mark.parametrize("has_nan", ["has_nan", "no_nans"])
+@pytest.mark.parametrize(
+    "numeric_impute_strategy", ["forwards_fill", "backwards_fill", "interpolate"]
+)
+def test_imputer_woodwork_custom_overrides_returned_by_components(
```
freddyaboulton, line 438: I know we have this kind of test scattered throughout the components but honestly it's kind of a pain to reason through. I like the idea of parametrizing over the impute strategy but I would be happy if we took `imputer_test_data`, set the logical types to something where they don't match the default inference and then just check the logical types at the end are what they were at the beginning.

No need to parametrize over logical type and then try to catch all the possible incompatible conversions.

freddyaboulton (commented):
@eccabay Looks good! I left some comments on the implementation/tests. I like the use of `imputer_test_data` but it has the unfortunate property where forwards, backwards, and interpolation all yield the same answer.

evalml/tests/component_tests/test_time_series_imputer.py
eccabay: This test does cover that scenario - the third value in the "float with nan" column is -1, 1, or 0 depending on whether the imputation strategy is backwards fill, forwards fill, or interpolate, respectively. It can be hard to notice though, so I am happy to add an explicit test if that makes more sense!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
eccabay: I played around with some implementation options to just target the first/last row, but everything I tried took significantly longer than running `bfill` or `pad` on the whole dataset. If we want to avoid creating a copy of the data, we could use `inplace=True` instead.

evalml/tests/component_tests/test_time_series_imputer.py
freddyaboulton: 🤦‍♂️

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
freddyaboulton: Awesome, let's use `inplace`!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,175 @@
...
+        Returns:
+            pd.DataFrame: Transformed X
+        """
+        X = infer_feature_types(X)
+        if len(self._all_null_cols) == X.shape[1]:
+            df = pd.DataFrame(index=X.index)
+            df.ww.init()
+            return df, y
+
+        X_no_all_null = X.ww.drop(self._all_null_cols)
```
chukarsten, line 144: omega level hyper nit : `X_not_all_null`.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,175 @@
...
+        if self._interpolate_cols is not None:
+            X_interpolate = X.ww[self._interpolate_cols]
+            imputed = X_interpolate.interpolate()
+            imputed.bfill(inplace=True)  # Fill in the first value, if missing
+            X_no_all_null[X_interpolate.columns] = imputed
+
+        y_imputed = pd.Series(y)
+        if self._impute_target == "forwards_fill":
+            y_imputed = y.pad()
+            y_imputed = y_imputed.bfill()
```
chukarsten, line 167: For my own learning, is there a reason you do the X data imputation in place vs. the target?

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,506 @@
...
+    expected = pd.Series([0, 1, 0, 1, 0] * 3 + [0, 1, 0, 1, 1])
+    assert_series_equal(y_t, expected, check_dtype=False)
+
+    imputer = TimeSeriesImputer(target_impute_strategy="interpolate")
+    _, y_t = imputer.fit_transform(X, y)
+    expected = pd.Series([0, 1, 0, 1, 0.5] * 3 + [0, 1, 0, 1, 1])
+    assert_series_equal(y_t, expected, check_dtype=False)
+
+
+def test_drop_all_columns(imputer_test_data):
```
chukarsten, line 212: OK, I think this probably addresses my previous comment about explicitly asserting that these columns were not in the results.  I think this is fine then, maybe just hyper nit on the name `test_imputer_drops_columns_with_all_nan`.  I know it's a bit wordier but I think it's a little clearer.

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,506 @@
...
+            ),
+            "object col": pd.Series(["b", "b", "a", "c", "d"] * 4, dtype="category"),
+            "bool col": [True, False, False, True, True] * 4,
+        }
+    )
+    assert_frame_equal(X_t, X_expected, check_dtype=False)
+    assert_series_equal(y_t, y, check_dtype=False)
+
+
+def test_imputer_with_none():
```
chukarsten, line 317: It's probably not worth the effort, but any thoughts on modifying the `imputer_data` fixture we're using so that it can return different types of null values?

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,506 @@
...
+    X = make_data_type(data_type, X)
+    y = make_data_type(data_type, y)
+    imputer = TimeSeriesImputer()
+    imputer.fit(X, y)
+    X_t, _ = imputer.transform(X)
+    assert_frame_equal(X_expected_arr, X_t)
+
+
+def test_imputer_unusual_types_as_categorical():
+    X = pd.DataFrame(
```
chukarsten, line 384: Is it worth just adding these to the `imputer_test_data` fixture?

chukarsten (approved):
Great work Becca.  Been really looking forward to this one.  I left a few questions/nits on some of the test names and some take-them-or-leave-them suggestions for additions/refactors to the test fixture.  Nothing blocking though!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
eccabay: Nope, total oversight on my part. I'll update this to be inplace as well.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,199 @@
...
+
+class TimeSeriesImputer(Transformer):
+    """Imputes missing data according to a specified timeseries-specific imputation strategy.
+
+    This Transformer should be used after the `TimeSeriesRegularizer` in order to impute the missing values that were
+    added to X and y (if passed).
+
+    Args:
+        categorical_impute_strategy (string): Impute strategy to use for string, object, boolean, categorical dtypes.
+                                              Valid values include "backwards_fill" and "forwards_fill". Defaults to "forwards_fill".
```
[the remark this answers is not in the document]
eccabay: Docstring linting doesn't seem to run on class-level docstrings, I did this formatting myself. 4 spaces would be the correct thing to do though, so I'll update this.

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -0,0 +1,506 @@
...
+            {
+                "input_val": [1.0, 2, 3, 4, 5, 6, 7, 8, 9],
+                "input_cat": pd.Categorical(["a"] * 6 + ["b"] * 3),
+            },
+            index=list(range(1, 10)),
+        ),
+    )
+
+
+def test_imputer_no_nans(imputer_test_data):
```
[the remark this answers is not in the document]
eccabay: Great point, I overlooked that! Will add numeric columns as well.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+    _valid_target_impute_strategies = set(
+        ["backwards_fill", "forwards_fill", "interpolate"]
+    )
+
+    def __init__(
+        self,
+        categorical_impute_strategy="forwards_fill",
+        numeric_impute_strategy="interpolate",
+        target_impute_strategy="forwards_fill",
```
ParthivNaresh, line 51: Out of curiosity, why wasn't `interpolate` chosen for the default strategy for the target if the target is a regression type?

ParthivNaresh (approved):
I think this looks great, nice work!

freddyaboulton (approved):
Looks good @eccabay Thanks for making the changes!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
eccabay: Great question! That hadn't occurred to me. Do we have a precedent for doing that with other components?

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -0,0 +1,175 @@
...
+
+        Returns:
+            self
+        """
+        X = infer_feature_types(X)
+
+        nan_ratio = X.ww.describe().loc["nan_count"] / X.shape[0]
+        self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
+
+        def _filter_cols(impute_strat, X):
```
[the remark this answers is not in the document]
eccabay: Any particular reason behind this thought? There are a number of class attributes we'd have to pass through as arguments in the case of a static method.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
eccabay: Totally, makes sense. We could also consider this when we add the `TimeSeriesImputer` to `make_pipeline`, adding `interpolation` in as the target impute strategy parameter if not already explicitly set.

PR #3379, opened 2022-03-15
Add clustering models

### LINKED ISSUE #3202: Implement unsupervised learning models
Part of the work to support unsupervised learning in AutoML. Models for V1:
- KMeans (along with KModes/KPrototype for categorical features)
- DBSCAN

eccabay:
Closes #3202


evalml/utils/gen_utils.py
```diff
@@ -242,7 +242,20 @@ def get_importable_subclasses(base_class, used_in_automl=True):
...
+        from evalml.pipelines import Transformer
+        from evalml.problem_types import ProblemTypes
+
+        classes = [
+            cls
+            for cls in classes
+            if (
+                cls.__name__ not in _not_used_in_automl
+                and (
+                    issubclass(cls, Transformer)
```
ParthivNaresh, line 254: What's the purpose of `issubclass(cls, Transformer)`?

bchen1116 (approved):
Great work on this! Left many nit-picks, but none are blocking merge. Agreed with the question about the check for `Transformer` in `gen_utils`.

evalml/pipelines/components/estimators/clusterers/dbscan.py
```diff
@@ -0,0 +1,54 @@
...
+    hyperparameter_ranges = {}
+    """{}"""
+    model_family = ModelFamily.DENSITY
+    """ModelFamily.DENSITY"""
+    supported_problem_types = [ProblemTypes.CLUSTERING]
+    """[
+        ProblemTypes.CLUSTERING
+    ]"""
+
+    def __init__(self, eps=0.5, min_samples=5, n_jobs=-1, random_seed=0, **kwargs):
```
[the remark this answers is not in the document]
eccabay: Good point! Epsilon and min_samples control the actual behavior of the algorithm, but adding leaf_size to include control over fit time seems like a great plan to me.

evalml/pipelines/components/estimators/clusterers/kmeans.py
```diff
@@ -0,0 +1,54 @@
...
+        parameters.update(kwargs)
+        kmeans_clusterer = SKKMeans(**parameters, random_state=random_seed)
+        parameters["n_jobs"] = n_jobs
+        super().__init__(
+            parameters=parameters,
+            component_obj=kmeans_clusterer,
+            random_seed=random_seed,
+        )
+
+    def predict(self, X=None):
```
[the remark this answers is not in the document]
eccabay: I was thinking the same thing 🤔 glad I'm not alone in that thought! I'll give it a shot, see how it shakes out.

evalml/utils/gen_utils.py
eccabay: The classes iterated through here can be both estimators and transformers. Estimators have a `supported_problem_types` attribute, but Transformers don't. If we try to check a transformer's supported_problem_types, we just get an error, so we have to check if it's a Transformer first. (I learned this the hard way 😅

evalml/pipelines/components/estimators/clusterers/dbscan.py
```diff
@@ -0,0 +1,54 @@
+"""DBSCAN Clusterer."""
+from sklearn.cluster import DBSCAN as SKDBSCAN
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+from evalml.utils import infer_feature_types
+
+
+class DBSCANClusterer(Estimator):
```
freddyaboulton, line 10: Just FYI, dbscan only finds one cluster for the breast cancer dataset with and without scaling:

image

evalml/utils/gen_utils.py
freddyaboulton: I think we can just add the clusterers to `_not_used_in_automl` ?

evalml/pipelines/components/estimators/clusterers/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""EvalML clustering components."""
```
freddyaboulton, line 1: Thoughts on renaming this module to `unsupervised`? Easier to say than `clusterer` imo.

freddyaboulton (approved):
@eccabay Nice work! Looks good to me. The only thing that stood out to me was the change in `get_importable_subclasses`. I think we can make that cleaner by just modifying `_not_used_in_automl`.

evalml/pipelines/components/estimators/clusterers/dbscan.py
eccabay: I'm not surprised. Of all these initial models, DBSCAN is most sensitive to hyperparameter values. Once we have a good way to performance test the models I'd love to adjust the default values, but for now I think following the sklearn default values makes the most sense.

evalml/utils/gen_utils.py
eccabay: I started out adding each clusterer to `_not_used_in_automl` as I added it, but it became a pain to remember to add, so I changed it. My thinking here is that this way we ensure no clusterers, present or future, are included in search. I'm happy to change it back, but personally I'm a fan of the less hard-coded approach.

freddyaboulton: Good point! But I think there are better ways of accomplishing that - maybe we can add an `used_in_search` attribute to component classes and then update this function to respect that value. Then you could just set `used_in_search` in the clusterer base class to `False`.

I think we should stick with the current pattern of modifying `_not_used_in_automl` and tackle ways to improve this method in a separate issue.

evalml/pipelines/components/estimators/clusterers/__init__.py
eccabay: I think it's important to keep the component names as clusterers, since there are more types of unsupervised learning than just clustering. I'm happy to rename the higher-level module to unsupervised though!

chukarsten (approved):
This looks good to me, great work getting through all these models.  I wonder at what point we update the documentation and make a user's guide.  Probably later as all we're really doing is wrapping existing stuff.

PR #3384, opened 2022-03-16
Small fix to make time series featurizer deterministic

bchen1116:
After this fix, we see this in the perf tests:
image
Ultimately, we note that the before and after scores for both of these pipelines are the same, so there could be an issue with how we're storing the full rankings or finding the order in LG. However, this PR is focused on making TS deterministic.

The full, fixed report is below:
[report_ts.html.zip](https://github.com/alteryx/evalml/files/8289934/report_ts.html.zip)

The perf tests look good! Seems like the behavior is now the same between runs


evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1584,3 +1584,49 @@ def test_time_index_cannot_be_none(time_series_regression_pipeline_class):
                 }
             }
         )
+
+
+def test_time_series_random_forest_transform_all_but_final(
+    ts_data, time_series_regression_pipeline_class
+):
+    X, y = ts_data
```
jeremyliweishih, line 1592: can we drop rows to make this test faster? Think we could just use the minimum amount of rows we need for the forecast horizon.

jeremyliweishih (approved):
Left a comment but looks great!

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -190,9 +190,11 @@ def _compute_rolling_transforms(self, X, y, original_features):
             min_periods=size + 1,
         )
         rolling_mean = rolling_mean.get_function()
-        numerics = set(
-            X.ww.select(["numeric"], return_schema=True).columns
-        ).intersection(original_features)
+        numerics = sorted(
```
freddyaboulton, line 193: I believe the test passes without this change. @bchen1116 May need to add some more numeric columns to the test case to test the intended behavior.

freddyaboulton (approved):
Thank you @bchen1116 !!

PR #3395, opened 2022-03-21
Conditionally run scheduled workflows

### LINKED ISSUE #3393: Prevent workflows from running on forked repos
Our scheduled workflows such as Auto Approve Dependency PRs and Latest Dependency Checker should not run on forked repos. 
We should make it so the workflow only runs if the repo owner `(if: github.owner == alteryx)`.

thehomebrewnerd:
### Conditionally run scheduled workflows

Closes #3393 

Update scheduled workflows to run only if repository_owner is alteryx. This should prevent these scheduled workflows from running on repo forks and failing repeatedly.

### DIFF
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -12,6 +12,7 @@ on:
       - completed
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


gsheni (approved):
lgtm

PR #3396, opened 2022-03-21
Add clustering objectives

### LINKED ISSUE #3200: Add clustering objective(s)
Part of the work supporting unsupervised learning in AutoML. Potential options for objectives: https://scikit-learn.org/stable/modules/clustering.html#clustering-performance-evaluation

eccabay:
Closes #3200

Originally I was just going to add the [Silhouette Coefficient](https://scikit-learn.org/stable/modules/clustering.html#rand-index), but while working on this I figured it would be good to have an unsupervised learning objective that took ground truth data into account as well, so the [Adjusted Rand Index](https://scikit-learn.org/stable/modules/clustering.html#rand-index) is also included in this PR.

### DIFF
--- a/evalml/objectives/standard_metrics.py
+++ b/evalml/objectives/standard_metrics.py
@@ -13,6 +13,7 @@ from .multiclass_classification_objective import (
 )
 from .regression_objective import RegressionObjective
 from .time_series_regression_objective import TimeSeriesRegressionObjective
+from .unsupervised_objective import UnsupervisedLearningObjective
 
 
 class AccuracyBinary(BinaryClassificationObjective):
@@ -899,6 +900,40 @@ class ExpVariance(RegressionObjective):
         return metrics.explained_variance_score(
             y_true, y_predicted, sample_weight=sample_weight
         )
+
+
+class SilhouetteCoefficient(UnsupervisedLearningObjective):
+    """Silhouette Coefficient for clustering. Does not require ground truth values."""
+
+    name = "Silhouette Coefficient"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = False
+    expected_range = [-1, 1]
+    requires_y_true = False
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
+        """Objective function for silhouette coefficient for clustering."""
+        try:
+            return metrics.silhouette_score(X, y_predicted, metric="euclidean")
+        except ValueError:
+            return 0
+
+
+class AdjustedRandScore(UnsupervisedLearningObjective):
+    """Rand Score for clustering, adjusted for chance. Does require ground truth values."""
+
+    name = "Adjusted Rand Score"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = True
+    expected_range = [0, 1]
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
+        """Objective function for rand score for clustering."""
+        return metrics.adjusted_rand_score(y_true, y_predicted)
 
 
 def _handle_predictions(y_true, y_pred):
--- a/evalml/objectives/standard_metrics.py
+++ b/evalml/objectives/standard_metrics.py
@@ -915,14 +915,13 @@ class SilhouetteCoefficient(UnsupervisedLearningObjective):
 
     def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
         """Objective function for silhouette coefficient for clustering."""
-        try:
-            return metrics.silhouette_score(X, y_predicted, metric="euclidean")
-        except ValueError:
+        if len(set(y_predicted)) == 1:
             return 0
+        return metrics.silhouette_score(X, y_predicted, metric="euclidean")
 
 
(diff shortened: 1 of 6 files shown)


evalml/objectives/standard_metrics.py
```diff
@@ -901,6 +902,40 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+
+    name = "Silhouette Coefficient"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = False
+    expected_range = [-1, 1]
+    requires_y_true = False
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
```
bchen1116, line 916: Is it possible to make `y_true` default to None here, or are we going to expect users to pass in `None` as the `y_true` value whenever using this?

evalml/objectives/standard_metrics.py
```diff
@@ -901,6 +902,40 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+    """Rand Score for clustering, adjusted for chance. Does require ground truth values."""
+
+    name = "Adjusted Rand Score"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = True
+    expected_range = [0, 1]
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
```
freddyaboulton, line 934: Where do the ground truth labels come from? At this point is this a classification objective?

evalml/objectives/standard_metrics.py
```diff
@@ -901,6 +902,40 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+    perfect_score = 1.0
+    is_bounded_like_percentage = False
+    expected_range = [-1, 1]
+    requires_y_true = False
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
+        """Objective function for silhouette coefficient for clustering."""
+        try:
+            return metrics.silhouette_score(X, y_predicted, metric="euclidean")
+        except ValueError:
```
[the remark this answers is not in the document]
freddyaboulton: Good point!

evalml/objectives/standard_metrics.py
freddyaboulton: The problem is that we can't change the default value of `y_true` without breaking the api signature we inherit from ObjectiveBase. If `UnSupervisedLearningObjective`s never needed `y_true` maybe we can modify the implementation of `score` in `UnsupervisedLearningObjective` to always pass in None for `y_true` but I think right now we have to force users to pass in `y_true=None` .

evalml/tests/objective_tests/test_standard_metrics.py
```diff
@@ -698,6 +704,39 @@ def test_mape_time_series_model():
...
+    obj = SilhouetteCoefficient()
+
+    X = pd.DataFrame({"f1": [1, 1, 1, 3, 3, 3], "f2": [1, 1, 1, 4, 4, 4]})
+    pred_good = np.array([0, 0, 0, 1, 1, 1])
+    pred_wrong = np.array([0, 1, 0, 1, 0, 1])
+    pred_bad = np.array([0, 0, 0, 0, 0, 0])
+
+    assert obj.score(y_true=None, y_predicted=pred_good, X=X) == 1.0
+    np.testing.assert_almost_equal(
+        obj.score(y_true=None, y_predicted=pred_wrong, X=X), -0.0555555
```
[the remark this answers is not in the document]
freddyaboulton: I'm ok with the hard-coded result on toy data - maybe we can add some comments explaining the why the answers are correct, e.g. "X has two clusters, first three rows vs second three rows ... "

evalml/objectives/standard_metrics.py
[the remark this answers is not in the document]
eccabay: The silhouette coefficient throws a `ValueError` in the case where there is only one cluster detected. Since other metrics (such as the rand score) simply return a score of 0 in this case, that's what this catch is trying to emulate.

It occurs to me now, I could implement this in a slightly safer and cleaner way:
```
if len(y_predicted) == 0:
    return 0
return metrics.silhouette_score(...)
```
This way any other ValueErrors will raise as necessary. Will update!

evalml/objectives/standard_metrics.py
eccabay: Yeah, I was on the fence about this. It does feel a little unwieldy to ask users to pass in `None` as the `y_true` every time, but I wanted to keep the `objective_function` signature intact. Really though, it does make more sense to set defaults in this case. I'll change this.

freddyaboulton (commented):
@eccabay Thank you for this! Agree with the comments @bchen1116 left

evalml/objectives/standard_metrics.py
eccabay: While less common, there is a chance a user may have ground-truth information about their data. Alternatively, having a metric that takes true labels will be useful for our performance tests in the future.

The distinction between this metric and a standard classification metric is that this metric is permutation-independent. You can see an example in the tests, where the cluster numbers in `y_true` and `y_predicted` are swapped, but the score is 1.0 regardless.

bchen1116 (approved):
Thanks for the quick fix! Looks great

PR #3399, opened 2022-03-23
Add MinMaxScaler component for normalization

### LINKED ISSUE #3355: Add a normalization component
Part of the work for supporting unsupervised learning in evalml, we need a normalization component to ensure features on different scales don't negatively affect the performance of models.

eccabay:
Closes #3355

### DIFF
--- a/evalml/tests/component_tests/test_minmax_scaler.py
+++ b/evalml/tests/component_tests/test_minmax_scaler.py
@@ -1,4 +1,6 @@
 import pandas as pd
+import pytest
+import woodwork as ww
 
 from evalml.pipelines.components import MinMaxScaler
 
@@ -53,4 +55,19 @@ def test_minmax_scaler_categorical_only():
     scaler.fit(X)
     X_t = scaler.transform(X)
 
+    assert scaler._is_fitted
     pd.testing.assert_frame_equal(X, X_t, check_dtype=False)
+
+
+def test_minmax_scaler_maintains_categorical_ww_schema():
+    X = pd.DataFrame(
+        {
+            "col_1": [1, -2, 5, 8, 0],
+            "col_2": ["a", "b", "a", "c", "b"],
+            "col_3": [0.2, 3.0, 4.7, 2.2, 0.9],
+        }
+    )
+    X.ww.init(logical_types={"col_2": "natural_language"})
+    scaler = MinMaxScaler()
+    X_t = scaler.fit_transform(X)
+    assert X_t.ww.logical_types["col_2"].type_string == "natural_language"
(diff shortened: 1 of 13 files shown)


chukarsten (approved):
LGTM Becca!  I think just need to add those asserts showing that the dataset values already falls outside the MinMaxScaler's default range.

evalml/tests/component_tests/test_minmax_scaler.py
```diff
@@ -0,0 +1,56 @@
...
+        assert max(X_t[col]) <= 1.0
+
+    scaler = MinMaxScaler(feature_range=(-1, 3))
+    X_t = scaler.fit_transform(X, y)
+    for col in X_t.columns:
+        assert -1 <= min(X_t[col]) < 0
+        assert 1 < max(X_t[col]) <= 3.1
+
+
+def test_minmax_scaler_numeric_and_categorical():
```
jeremyliweishih, line 25: might be a good idea to check that woodwork inference isn't lost in this test.

jeremyliweishih (approved):
Left one testing comment, looks great!

evalml/tests/component_tests/test_minmax_scaler.py
```diff
@@ -0,0 +1,56 @@
...
+    X_t = scaler.transform(X)
+    for col in X_t.columns:
+        assert min(X_t[col]) >= 0.0
+        assert max(X_t[col]) <= 1.0
+
+    scaler = MinMaxScaler(feature_range=(-1, 3))
+    X_t = scaler.fit_transform(X, y)
+    for col in X_t.columns:
+        assert -1 <= min(X_t[col]) < 0
+        assert 1 < max(X_t[col]) <= 3.1
```
bchen1116, line 22: Why is the range here 3.1 instead of 3?

evalml/tests/component_tests/test_minmax_scaler.py
```diff
@@ -0,0 +1,56 @@
...
+            "col_1": ["a", "a", "b", "b", "a"],
+            "col_2": ["a", "b", "a", "c", "b"],
+            "col_3": [False, True, True, False, True],
+        }
+    )
+    scaler = MinMaxScaler()
+    scaler.fit(X)
+    X_t = scaler.transform(X)
+
+    pd.testing.assert_frame_equal(X, X_t, check_dtype=False)
```
bchen1116, line 56: just a nit, but can we do `assert scaler._is_fitted` to ensurre that the scaler is fitted?

bchen1116 (approved):
Left a question on a test and a suggestion, but LGTM! Can you add this component, and the other Unsupervised components if applicable, to the `api_reference.rst` file?

evalml/tests/component_tests/test_minmax_scaler.py
eccabay: There's a small rounding error, so the max value in this case is actually 3.00004. Close enough to not matter, but far enough to break the test if left at 3.

bchen1116: Wow, that's interesting that the rounding error would pass sklearn's internal tests. 
This seems fine to me then!

PR #3404, opened 2022-03-24
Refactor partial dependence functions

eccabay:
Moves all partial dependence functions to two files: a private file for all the implementation and utility functions and a public file for the API. Part of the work to refactor `graph.py` for ease of use.

This PR looks big but none of the code is actually different, just in different places.


jeremyliweishih (approved):
LGTM!

chukarsten (approved):
LGTM!  Thank you, as always, for taking the initiative.

PR #3425, opened 2022-03-30
Update `DatetimeFormatDataCheck` to use woodwork

### LINKED ISSUE #3385: Update `DateTimeFormatDataCheck` to use woodwork frequency inference
Our current implementation uses pandas and then checks for different irregularities ourselves. Once [woodwork's implementation](https://github.com/alteryx/woodwork/pull/1281) is merged and released, we should use their function instead.

eccabay:
Closes #3385


evalml/data_checks/datetime_format_data_check.py
```diff
@@ -179,71 +187,78 @@ def validate(self, X, y):
...
         col_name = (
             self.datetime_column if self.datetime_column != "index" else "either index"
         )
 
-        nan_columns = datetime_values.isna().any()
-        if nan_columns:
+        ww_payload = infer_frequency(
+            pd.Series(datetime_values),
+            debug=True,
+            window_length=3,
```
ParthivNaresh, line 208: We default to a window length of 5 in the regularizer, should we maintain parity?

eccabay: Yep, great idea!

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -36,14 +37,6 @@ def validate(self, X, y):
             >>> datetime_format_dc = DateTimeFormatDataCheck(datetime_column="dates")
             >>> assert datetime_format_dc.validate(X, y) == [
             ...     {
-            ...         "message": "Column 'dates' has datetime values missing between start and end date.",
```
jeremyliweishih, line 39: Why is this being removed?

jeremyliweishih (approved):
This LGTM! Just some clarifying questions but might be prudent to wait for a review from someone with more context than me 😄  good work!

evalml/data_checks/datetime_format_data_check.py
eccabay: It's no longer detected by woodwork - it purely errors out unable to detect a frequency, so no more missing values. IMO, this is the more accurate output!

evalml/tests/data_checks_tests/test_datetime_format_data_check.py
```diff
@@ -28,8 +28,8 @@ def test_datetime_format_data_check_typeerror_uneven_intervals(
 
     if issue == "missing":
         # Skips 2021-01-30 and appends 2021-01-31, skipping a date and triggering the error
```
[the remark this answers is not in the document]
eccabay: Good catch!

evalml/data_checks/data_check_message_code.py
```diff
@@ -113,10 +113,13 @@ class DataCheckMessageCode(Enum):
     """Message code for when the datetime values have uneven intervals."""
 
     DATETIME_HAS_REDUNDANT_ROW = "datetime_has_redundant_row"
-    """Message code for when datetime information has more than one row per datetime"""
+    """Message code for when datetime information has more than one row per datetime."""
```
chukarsten, line 116: Angela would be proud.

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -179,71 +195,79 @@ def validate(self, X, y):
...
-        if nan_columns:
+        ww_payload = infer_frequency(
+            pd.Series(datetime_values),
+            debug=True,
+            window_length=5,
+            threshold=0.8,
+        )
+        inferred_freq = ww_payload[0]
+        debug_object = ww_payload[1]
+        if inferred_freq is not None:
```
chukarsten, line 221: A little confused why you'd want to return here if there's a detected inferable frequency.  At this point you've determined whether it's monotonic or not and that it has datetime information...

evalml/data_checks/datetime_format_data_check.py
```diff
@@ -36,14 +37,6 @@ def validate(self, X, y):
             >>> datetime_format_dc = DateTimeFormatDataCheck(datetime_column="dates")
```
chukarsten, line 37: Do we need to update `.validate()`'s docstring a little bit more outside of the pydocstring?

chukarsten (approved):
Awesome, looks great.  I think just a question about maybe updating the docstring summary to make it a bit more comprehensive.  Other than that, super excited here!

evalml/data_checks/datetime_format_data_check.py
eccabay: If woodwork returns an inferred frequency, it's the same as if the `pd.infer_freq` function returned a frequency. In that case, the datetime column is perfect and we don't need to run the rest of the checks as we know they don't hold.

Now that I talk that through though, I realize we can move that check up higher. We do run `inferred_freq = pd.infer_freq(...)`, so it would make more sense to check if the frequency is perfect up there instead.

evalml/data_checks/datetime_format_data_check.py
eccabay: I can add another more descriptive line!

evalml/data_checks/datetime_format_data_check.py
chukarsten: Thanks for explaining that one to me!

ParthivNaresh (approved):
Looks good thanks for getting this in!

PR #3427, opened 2022-03-30
Update Pipeline parameters pt 2

### LINKED ISSUE #3414: Fix performance bug in search_parameters PR
We reverted this commit because the implementation of `get_starting_parameters` was overriding the default estimator parameters. This was causing search time to increase.

https://github.com/alteryx/evalml/commit/b44245377bb325876cf54439e22d7037686c2c29

bchen1116:
Fix #3414


evalml/tuners/tuner.py
```diff
@@ -89,6 +89,35 @@ def _convert_to_pipeline_parameters(self, flat_parameters):
             pipeline_parameters[component_name][parameter_name] = parameter_value
         return pipeline_parameters
 
+    def get_starting_parameters(self, hyperparameter_ranges):
```
bchen1116, line 92: This is the main change. Rather than using the hyperparameter ranges attached to the tuner, we can just pass in the hyperparameter ranges that the user specified

freddyaboulton: I think this fixes the problem with iterative algorithm but not the default algorithm. When we go to the "fast final" batch, we'll take some parameters from the tuner and not use the default ones.

This is the logic now for creating the pipelines in the fast final batch, note that we just call `propose`. 
```python
    def _create_n_pipelines(self, pipelines, n):
        next_batch = []
        for _ in range(n):
            for pipeline in pipelines:
                if pipeline.name not in self._tuners:
                    self._create_tuner(pipeline)

                select_parameters = self._create_select_parameters()
                proposed_parameters = self._tuners[pipeline.name].propose()
                parameters = self._transform_parameters(pipeline, proposed_parameters)
                parameters.update(select_parameters)
                next_batch.append(
                    pipeline.new(parameters=parameters, random_seed=self.random_seed)
                )
        return next_batch
```

In main, we don't take a sample from the tuner:
```python
    def _create_fast_final(self):
        estimators = [
            estimator
            for estimator in get_estimators(self.problem_type)
            if estimator not in self._naive_estimators()
        ]
        estimators = self._filter_estimators(
            estimators,
            self.problem_type,
            self.allow_long_running_models,
            None,
            self.y.nunique(),
            self.logger,
        )
        pipelines = self._make_pipelines_helper(estimators)

        if self._split:
            self._rename_pipeline_parameters_custom_hyperparameters(pipelines)

        next_batch = []
        for pipeline in pipelines:
            parameters = self._create_select_parameters()
            pipeline = pipeline.new(
                parameters=self._transform_parameters(pipeline, parameters),
                random_seed=self.random_seed,
            )
            next_batch.append(pipeline)

        for pipeline in next_batch:
            self._create_tuner(pipeline)
        return next_batch
```

And then in `_transform_parameters` we make sure we we take a sample if custom hyperparameters are present - 
```python
    def _create_fast_final(self):
        estimators = [
            estimator
            for estimator in get_estimators(self.problem_type)
            if estimator not in self._naive_estimators()
        ]
        estimators = self._filter_estimators(
            estimators,
            self.problem_type,
            self.allow_long_running_models,
            None,
            self.y.nunique(),
            self.logger,
        )
        pipelines = self._make_pipelines_helper(estimators)

        if self._split:
            self._rename_pipeline_parameters_custom_hyperparameters(pipelines)

        next_batch = []
        for pipeline in pipelines:
            parameters = self._create_select_parameters()
            pipeline = pipeline.new(
                parameters=self._transform_parameters(pipeline, parameters),
                random_seed=self.random_seed,
            )
            next_batch.append(pipeline)

        for pipeline in next_batch:
            self._create_tuner(pipeline)
        return next_batch
```

On this branch, nyc taxi takes 10 minutes on the default algoirhtm
image

But one minute on main:
image

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -300,16 +273,9 @@ def _create_fast_final(self):
...
-        for pipeline in pipelines:
-            parameters = self._create_select_parameters()
-            pipeline = pipeline.new(
-                parameters=self._transform_parameters(pipeline, parameters),
-                random_seed=self.random_seed,
-            )
-            next_batch.append(pipeline)
+        next_batch = self._create_n_pipelines(pipelines, 1)
 
         for pipeline in next_batch:
```
freddyaboulton, line 280: We're creating the tuners twice here, once in `_create_n_pipelines` and then here in the `for pipeline in next_batch` for loop.

freddyaboulton (changes requested):
@bchen1116 Almost there! Let's make sure to run perf tests prior to merging so we can be extra confident as well.

evalml/automl/automl_algorithm/default_algorithm.py
bchen1116: Great catch!

evalml/tuners/tuner.py
bchen1116: Great catch @freddyaboulton! I will fix this and run tests locally to make sure it works as expected.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -326,8 +289,14 @@ def _create_n_pipelines(self, pipelines, n):
                     self._create_tuner(pipeline)
 
                 select_parameters = self._create_select_parameters()
-                proposed_parameters = self._tuners[pipeline.name].propose()
-                parameters = self._transform_parameters(pipeline, proposed_parameters)
+                parameters = (
```
bchen1116, line 292: Fixed the parameters we use when we call this through `_create_fast_final` versus `_create_n_pipelines` with a larger size of pipelines

evalml/tuners/tuner.py
```diff
@@ -89,6 +89,36 @@ def _convert_to_pipeline_parameters(self, flat_parameters):
             pipeline_parameters[component_name][parameter_name] = parameter_value
         return pipeline_parameters
 
+    def get_starting_parameters(self, hyperparameter_ranges, random_seed=0):
```
bchen1116, line 92: Add the `hyperparam_ranges` and `random_seed` args to limit the starting parameters to only custom hyperparameters that users pass in.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -326,8 +289,14 @@ def _create_n_pipelines(self, pipelines, n):
...
+                    self._tuners[pipeline.name].get_starting_parameters(
+                        self._hyperparameters, self.random_seed
+                    )
+                    if n == 1
```
freddyaboulton, line 296: I think we want `_batch_number <= 2` here right? I think this works right now because we only call `_create_n_pipelines` with `n=1` in the create_fast_final batch but would not work if the code was refactored I think.

evalml/tests/automl_tests/parallel_tests/test_automl_dask.py
```diff
@@ -205,10 +205,9 @@ def test_automl_immediate_quit(
         pipelines_per_batch=5,
         ensembling=False,
         text_in_ensembling=False,
-        pipeline_params={},
-        custom_hyperparameters=None,
+        search_parameters={},
     )
-    automl.automl_algorithm.allowed_pipelines = pipelines
+    automl.automl_algorithm._set_allowed_pipelines(pipelines)
```
freddyaboulton, line 210: Is the tuner not created otherwise? Wondering why we're making this change

evalml/automl/automl_algorithm/automl_algorithm.py
```diff
@@ -64,14 +65,72 @@ def next_batch(self):
...
+            # Only overwrite the parameters that were passed in on pipeline parameters
+            # if they don't exist in the propsed parameters
+            if name in self._pipeline_parameters and name not in component_parameters:
+                for param_name, value in self._pipeline_parameters[name].items():
+                    component_parameters[param_name] = value
+            # Inspects each component and adds the following parameters when needed
+            if "n_jobs" in init_params:
+                component_parameters["n_jobs"] = self.n_jobs
+            try:
+                if "number_features" in init_params:
```
freddyaboulton, line 124: Will file an issue to delete this. Don't know what it's used for and why it's only for iterative algorithm.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -4488,6 +4460,87 @@ def test_cv_validation_scores_time_series(
...
+@pytest.mark.parametrize("algorithm,batches", [("iterative", 2), ("default", 3)])
+@pytest.mark.parametrize(
+    "parameter,expected",
+    [
+        ("mean", ["mean", "median", "most_frequent"]),
+        (Categorical(["mean"]), Categorical(["mean"])),
+    ],
+)
+@pytest.mark.parametrize("problem_type", ["binary", "time series binary"])
+def test_search_parameters_held_automl(
```
freddyaboulton, line 4472: Thoughts on adding a test to check whether the estimator parameters used match the default values? I think e can use `inspect` to make sure the test does not get stale. We chose the default values precisely cause they're fast/get good enough performance so might be good to add that coverage. And I think that unit test would have caught the original problem with the first version of this pr right?

freddyaboulton (approved):
@bchen1116 Looks good to me. Thank you for sticking with this! The perf tests look good and I was able to verify the previous bug in how the default algorithm created the starting batch is no longer there. 

I feel confident about the merge but I left a suggestion for a unit test that I think will nail it home for us now and in the future.

Would be good to get two approvals on this one since it's a big refactor. Maybe wait for @jeremyliweishih !

evalml/tests/automl_tests/test_automl.py
jeremyliweishih: I like this idea!

evalml/tests/automl_tests/parallel_tests/test_automl_dask.py
bchen1116: Yep! With the changes that we made to iterative, just setting `allowed_pipelines` doesn't automatically create tuners for them.

evalml/tests/automl_tests/test_automl.py
bchen1116: yep, can do!

evalml/automl/automl_algorithm/default_algorithm.py
jeremyliweishih: a good alternative here would be to refactor this logic to be under a flag instead of basing it off of the number of pipelines generated

bchen1116: @jeremyliweishih great suggestion! I'll change that.

PR #3434, opened 2022-03-31
3370 ts imputer proposed

chukarsten:
Just a proposed change to how the imputer does fit to cut some repetitive code down.


chukarsten:
@eccabay before you pushed the most recent changes, I was going to suggest this refactor....do you think it will be helpful?  Seems like with the updates to your branch, it might be more useful and cut out some more code.

eccabay:
@chukarsten we're really good at pushing things at the same time 😂  I do like this helper function though, it achieves exactly what I want to do but couldn't figure out. I'll see if I can apply this based on the new changes!

eccabay:
@chukarsten ok to close this?

chukarsten:
Closing as it was used

PR #3419, opened 2022-03-29
Fix feature provenance for URL and Email features in `DefaultAlgorithm`

jeremyliweishih:
Fixes #3392.

PR #3435, opened 2022-03-31
Enable ensembling as a parameter for `DefaultAlgorithm`

jeremyliweishih:


docs/source/release_notes.rst
```diff
@@ -3,7 +3,9 @@
 
 **Future Releases**
     * Enhancements
+        * Enabled ensembling as a parameter for ``DefaultAlgorithm`` :pr:`3435`
     * Fixes
+        * Fix ``DefaultAlgorithm`` not handling Email and URL features :pr:`3419`
```
jeremyliweishih, line 8: Didn't move this out in #3419.

evalml/automl/automl_search.py
```diff
@@ -208,7 +208,7 @@ def search(
         if data_check_result["level"] == DataCheckMessageType.ERROR.value:
             return None, data_check_results
 
-    automl = AutoMLSearch(automl_algorithm="default", **automl_config)
+    automl = AutoMLSearch(automl_algorithm="default", ensembling=True, **automl_config)
```
jeremyliweishih, line 211: Keeping the same behavior for the top level search method.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -300,11 +302,11 @@ def test_pipeline_limits(
         automl.search()
     out = caplog.text
     if verbose:
-        assert "Using default limit of max_batches=4." in out
-        assert "Searching up to 4 batches for a total of" in out
+        assert "Using default limit of max_batches=3." in out
```
jeremyliweishih, line 305: Had to change because ensembling is turned off by default in `AutoMLSearch`.

chukarsten:
One thing I forgot to check - do we need to update the documentation since turning ensembling on/off for iterative algorithm is called out?

docs/source/user_guide/automl.ipynb
```diff
@@ -605,7 +605,7 @@
...
+    "AutoML enables stacking using the `ensembling` flag during initalization; this is set to `False` by default. How ensembling runs is defined by the AutoML algorithm you choose. In the `IterativeAlgorithm`, the stacking ensemble pipeline runs in its own batch after a whole cycle of training has occurred (each allowed pipeline trains for one batch). Note that this means __a large number of iterations may need to run before the stacking ensemble runs__. It is also important to note that __only the first CV fold is calculated for stacking ensembles__ because the model internally uses CV folds. See below in the AutoML Algorithms section to see how ensembling is run for `DefaultAlgorithm`."
```
jeremyliweishih, line 608: Made some doc changes here @chukarsten.

chukarsten: gucci

docs/source/user_guide/automl.ipynb
```diff
@@ -735,7 +735,7 @@
...
+    "To this end, it is recommended to use the top level `search()` method to run `DefaultAlgorithm`. This allows users to specify running search with just the `mode` parameter, where `fast` is recommended for users who want a fast scan at how EvalML pipelines will perform on their problem and where `long` is reserved for a deeper dive into high performing pipelines. If one needs finer control over AutoML parameters, one can also specify `automl_algorithm='default'` using `AutoMLSearch` and it will default to using `fast` mode. However, in this case ensembling will be defined by the `ensembling` flag (if `ensembling=False` the abovementioned ensembling batches will be skipped). Users are welcome to select `max_batches` according to the algorithm above (or other stopping criteria) but should be aware that results may not be optimal if the algorithm does not run for the full length of `fast` mode."
```
freddyaboulton, line 738: Let's say ensembling for time series is not enabled

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         n_jobs=-1,
         text_in_ensembling=False,
         top_n=3,
+        ensembling=True,
```
freddyaboulton, line 93: Should we set this false to match the behavior in AutoMLSearch? 

What I don't like about this is that we change the user parameter value silently for time series if it's not `False`.  Perhaps it would be better to raise an exception if ensembling is set to true for time series?

evalml/tests/automl_tests/test_automl.py
freddyaboulton: Let's remember to update our perf test code! I think we still want the ability to test the ensembler pipeline in batch 4 when we kick off a job?

freddyaboulton (commented):
Looks good @jeremyliweishih ! Saw you already merged but was half way through the review so figured I'd finish it. 

Nothing blocking and if we decide address these comments we can do so in a follow up.

evalml/automl/automl_algorithm/default_algorithm.py
jeremyliweishih: will do!

PR #3446, opened 2022-04-05
Increase coverage of model understanding functions in docs

eccabay:
We were missing descriptions of a few functions, this adds those.

### DIFF
--- a/evalml/tests/dependency_update_check/latest_dependency_versions.txt
+++ b/evalml/tests/dependency_update_check/latest_dependency_versions.txt
@@ -25,7 +25,7 @@ scikit-optimize==0.9.0
 scipy==1.7.3
 seaborn==0.11.2
 shap==0.40.0
-sktime==0.11.0
+sktime==0.11.1
 statsmodels==0.13.2
 texttable==1.6.4
 vowpalwabbit==9.1.0


freddyaboulton (approved):
Thank you @eccabay !

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -91,6 +91,35 @@
...
+    "            \"Imputer\": [\"Imputer\", \"X\", \"Label Encoder.y\"],\n",
+    "            \"Logistic Regression Classifier\": [\n",
+    "                \"Logistic Regression Classifier\",\n",
+    "                \"Imputer.x\",\n",
+    "                \"Label Encoder.y\",\n",
+    "            ],\n",
+    "        })\n",
+    "pipeline_linear.fit(X_train, y_train)\n",
+    "\n",
+    "get_linear_coefficients(pipeline_linear.estimator, features=X.columns)"
```
bchen1116, line 120: nit: Not sure if this would be good as a change to the function output or just a small change to the cell, but I think it'd look cleaner and make more sense for OS users if we showed this as a dataframe with column names labeled:
image

Similar to 
image

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -763,6 +792,36 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from evalml.model_understanding import graph_t_sne\n",
+    "\n",
+    "graph_t_sne(X)"
```
bchen1116, line 823: There's no output here:
image

Filed an issue [here](https://github.com/alteryx/evalml/issues/3447) but likely need to address that first before we can move in with this.

docs/source/user_guide/model_understanding.ipynb
```diff
@@ -763,6 +792,36 @@
...
     "The returned `objective_thresholds` dictionary has the objective measure as the key, and the dictionary value associated contains both the best objective score and the threshold that results in the associated score."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Visualize high dimensional data in lower space\n",
+    "\n",
+    "We can use T-SNE to visualize data with many features on a 2D plot, making it easier to see relationships in your data."
```
bchen1116, line 802: Might be nice to include a link to the documentation for this function

bchen1116 (changes requested):
Great work adding these in! There is a graphing issue that I think blocks this, but left some nits as well!

docs/source/user_guide/model_understanding.ipynb
freddyaboulton: So weird and good catch @bchen1116 ! Maybe we can try to build the docs locally to see if they show up in the generated html

docs/source/user_guide/model_understanding.ipynb
bchen1116: I think just a link to [this](https://evalml.alteryx.com/en/stable/autoapi/evalml/model_understanding/index.html#evalml.model_understanding.graph_t_sne) would help!

docs/source/user_guide/model_understanding.ipynb
eccabay: @bchen1116 since this graphing issue is independent of these changes, not sure it's worth it to block this PR on them getting fixed

bchen1116 (approved):
Good point @eccabay! The PR looks good to me! Hope we can get graphing fixed in the docs soon, but let's get this in first!

PR #3463, opened 2022-04-13
Change `PipelineBase.graph_json()` to return python dictionary

### LINKED ISSUE #3073: Pipeline graphs: refactor to use data generation vs viz methods
As discussed in #3049, [returning a JSON representation of a graph may not be the best choice moving forward](https://github.com/alteryx/evalml/pull/3049#discussion_r750585608), especially since we can't guarantee this method will always succeed. Things to consider:
- Should we return a python dictionary instead, and leave it to the user to handle JSON serialization?
- We should have a function that converts our component graph representation into one that's easy to pass to graphing libraries. Right now, that task is handled by `graph_json`. However, we use a lot of the same logic within `graph` as well. Should we refactor the `graph_json` and `graph` functions so that the latter calls the former?

jeremyliweishih:
Fixes #3073. Will file another issue to keep track of refactoring for `graph` and `graph_dict`.

### DIFF
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -439,7 +439,7 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         "y_edges": [[from_component_name, to_component_name], [from_component_name, to_component_name], ...]}
 
         Returns:
-            dag_dict (dict): A serialized JSON representation of a DAG structure.
+            dag_dict (dict): A dictionary of the DAG structure.
         """
         nodes = {}
         for comp_, att_ in self.component_graph.component_instances.items():
--- a/evalml/pipelines/pipeline_base.py
+++ b/evalml/pipelines/pipeline_base.py
@@ -428,8 +428,8 @@ class PipelineBase(ABC, metaclass=PipelineBaseMeta):
         df = pd.DataFrame(importance, columns=["feature", "importance"])
         return df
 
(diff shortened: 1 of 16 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -428,8 +427,8 @@ def feature_importance(self):
         df = pd.DataFrame(importance, columns=["feature", "importance"])
         return df
 
-    def graph_json(self):
-        """Generates a JSON with nodes consisting of the component names and parameters, and edges detailing component relationships.
+    def graph_dict(self):
```
freddyaboulton, line 430: I think this is ok for backwards compatibility

freddyaboulton (approved):
Looks good to me @jeremyliweishih !

evalml/tests/automl_tests/test_automl_iterative_algorithm.py
```diff
@@ -1008,12 +1007,11 @@ def test_graph_automl(X_y_multi):
         automl_algorithm="iterative",
     )
 
-    dag_str = automl.allowed_pipelines[0].graph_json()
-    dag_json = json.loads(dag_str)
+    dag_dict = automl.allowed_pipelines[0].graph_dict()
```
eccabay, line 1010: So glad this can be simplified!

eccabay (approved):
Looks great, thanks for taking care of this Jeremy!

PR #3486, opened 2022-04-28
Fix some unit tests after vowpal wabbit got added to conda

freddyaboulton:
### Pull Request Description

The build_conda_pkg job doesn't run when you change the meta.yaml so we didn't catch some very minor unit test failures once we merged to main:

https://github.com/alteryx/evalml/runs/6202590714?check_suite_focus=true

-----

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: acb758f2234a60dd606a4aa7dcde385e159c95a08470b0c5815d085db48aa985
 
 build:
-  number: 1
+  number: 2
 
 outputs:
   - name: evalml-core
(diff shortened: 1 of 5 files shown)


.github/workflows/build_conda_pkg.yml
```diff
@@ -10,6 +10,7 @@ on:
       - 'requirements.txt'
       - 'core-requirements.txt'
       - 'evalml/tests/dependency_update_check/*.txt'
+      - '.github/meta.yaml'
```
freddyaboulton, line 13: So that it runs when we update the recipe but not the requirement files.

.github/meta.yaml
```diff
@@ -9,7 +9,7 @@ source:
   sha256: acb758f2234a60dd606a4aa7dcde385e159c95a08470b0c5815d085db48aa985
 
 build:
-  number: 1
+  number: 2
```
jeremyliweishih, line 12: why is this changing?

jeremyliweishih (approved):
thank you!

.github/meta.yaml
freddyaboulton: Just so that I could run `build_conda_pkg`. In the real recipe it will start at 0.

PR #3494, opened 2022-05-02
Transition to pyproject.toml and setup.cfg

### LINKED ISSUE #3412: Transition EvalML to use pyproject.toml and setup.cfg (away from setup.py)
> The future of Python packaging is pyproject.toml, and (for now) setup.cfg, based on [PEP 518](https://snarky.ca/clarifying-pep-518/) and (soon) [PEP 621](https://www.python.org/dev/peps/pep-0621/).

- http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html

- How we did this in Woodwork:
  - https://github.com/alteryx/woodwork/pull/1346
  -  https://github.com/alteryx/woodwork/pull/1361

eccabay:
Closes #3412


setup.cfg
```diff
@@ -1,9 +1,142 @@
+[metadata]
+name = evalml
+version = 0.51.0
```
gsheni, line 3:
```suggestion
version = attr: evalml.version.__version__```

File: setup.cfg [hunk 2]
```diff
@@ -1,9 +1,142 @@
...
+author = Alteryx, Inc.
+license = BSD 3-clause
+long_description = file: READEME.md
+url = https://github.com/alteryx/evalml/
```
gsheni, line 7:
```suggestion
url = https://github.com/alteryx/evalml/
download_url = https://pypi.org/project/evalml/
```

setup.cfg
```diff
@@ -1,9 +1,142 @@
...
+    black == 22.3.0
+    isort == 5.0.0
+    pydocstyle == 6.1.1
+    darglint == 1.8.0
+    %(test)s
+    %(docs)s
+
+test = 
+    wheel >= 0.33.1
+    setuptools >= 50.0.1
```
gsheni, line 89: You might not need these
```suggestion
```

setup.cfg
```diff
@@ -12,16 +145,26 @@ ignore = E501,W504,W503
...
 test=pytest
+
 [isort]
 profile=black
 forced_separate=evalml
+known_first_party=evalml
 multi_line_output=3
 skip=__init__.py
+
 [darglint]
```
gsheni, line 159: this can be moved to pyproject.toml

setup.py
```diff
@@ -1,37 +1,4 @@
-from setuptools import find_packages, setup
+from setuptools import setup
```
gsheni, line 1: You can delete this file

Makefile
```diff
@@ -91,37 +74,36 @@ installdeps-min:
...
+	python -m pip install --upgrade build
 
 .PHONY: package_evalml
-package_evalml:
-	python setup.py sdist
-	$(eval EM_VERSION=$(shell python setup.py --version))
+package_evalml: upgradepip upgradebuild
+	python -m build
+	$(eval EM_VERSION := $(shell grep '__version__\s=' evalml/__init__.py | grep -o '[^ ]*$'))
 	tar -zxvf "dist/evalml-${EM_VERSION}.tar.gz"
```
gsheni, line 108:
```suggestion
	$(eval PACKAGE=$(shell python -c "from pep517.meta import load; metadata = load('.'); print(metadata.version)"))
	tar -zxvf "dist/evalml-${PACKAGE}.tar.gz"
```

docs-requirements.txt
```diff
@@ -1,10 +0,0 @@
-docutils>=0.15.2,<0.17
```
freddyaboulton, line 1: Makes sense to delete this since we now install the `dev` target in the read the docs yaml

tools/format_release_notes.sh
```diff
@@ -9,7 +9,7 @@ Usage: tools/format_release_notes.sh
 Note: this was written to work properly on mac. It should work on linux if -r is used instead of -E for sed.'
     exit 0
 fi
-evalml_version=$(sed -n -E "s/.*version=\'([0-9A-Za-z\.\_]+)\'.*/\1/p" setup.py)
+evalml_version=$(python -c "import evalml; print(evalml.__version__)")
```
freddyaboulton, line 12: This assumes EvalML is installed where this is run. I think that's ok since this script is run by the devs but maybe it's better to use the `pep517.meta` trick Gaurav suggested here too.

Makefile
```diff
@@ -91,37 +74,36 @@ installdeps-min:
...
+
+.PHONY: upgradebuild
+upgradebuild:
+	python -m pip install --upgrade build
 
 .PHONY: package_evalml
-package_evalml:
-	python setup.py sdist
-	$(eval EM_VERSION=$(shell python setup.py --version))
+package_evalml: upgradepip upgradebuild
```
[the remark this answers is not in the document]
gsheni: We were upgrading pip before so I kept it in the transition.
The `build` package helps us make sdist and the wheel files.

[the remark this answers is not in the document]
freddyaboulton: Cool, let's keep it. I don't think it will cause problems.

setup.cfg
```diff
@@ -1,9 +1,143 @@
...
+long_description = file: READEME.md
+url = https://github.com/alteryx/evalml/
+download_url = https://pypi.org/project/evalml/
+long_description_content_type = text/markdown
+author_email = open_source_support@alteryx.com
+keywords = data science machine learning optimization automl
+description = EvalML is an AutoML library that builds, optimizes, and evaluates machine learning pipelines using domain-specific objective functions.
+
+classifiers =
+    Development Status :: 3 - Alpha
```
freddyaboulton, line 17: I think it's more appropriate to mark it at least [beta](https://pypi.org/classifiers/) (we don't have a 1.0 release so maybe not "stable" yet).

setup.cfg
```diff
@@ -1,9 +1,143 @@
...
+    Intended Audience :: Developers
+    Topic :: Software Development
+    Topic :: Scientific/Engineering
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
freddyaboulton, line 24: We support 3.9 and 3.10 in testing, so I would delete 3.7 and 3.10 but keep `:: 3`

freddyaboulton (changes requested):
Thank you @eccabay for sticking with this and thank you @gsheni for helping out for the review. Left some small comments. Should be good to merge after that.

.github/workflows/install_test.yaml
```diff
@@ -10,12 +10,13 @@ on:
 name: Install Test
 jobs:
   install_evalml:
-    name: ${{ matrix.python_version }} install EvalML
-    runs-on: ubuntu-latest
+    name: ${{ matrix.os }} - ${{ matrix.python_version }} install EvalML
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest]
```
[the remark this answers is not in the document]
gsheni: Yes, it's good to check that your package installs on ARM chips

evalml/automl/automl_search.py
```diff
@@ -585,10 +585,6 @@ def __init__(
         self.random_seed = random_seed
         self.n_jobs = n_jobs
 
-        if not self.plot and self.verbose:
```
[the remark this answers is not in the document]
eccabay: Plotly moved from being a core-requirement to one of the regular requirements. `self.plot` is only None if plotly isn't installed. Since we no longer test this case, I removed this to make codecov happy (and users that pip install us shouldn't have an issue regardless)

freddyaboulton (approved):
Looks good to me @eccabay !

PR #3496, opened 2022-05-03
Preserve schema during column re-naming in xgboost and lightgbm

freddyaboulton:
### Pull Request Description

The fact that we don't preserve the schema during column re-naming at the last step of xgboost + lightgbm pipelines makes it so that this unrelated ww bug (https://github.com/alteryx/woodwork/issues/1411) causes some of our pipelines to fail in corner cases.

This is a fix to prevent pipelines from crashing. We should have been preserving the schema anyways.

### Perf tests
[report.html.zip](https://github.com/alteryx/evalml/files/8614188/report.html.zip)

-----

### DIFF
--- a/evalml/utils/gen_utils.py
+++ b/evalml/utils/gen_utils.py
@@ -261,9 +261,11 @@ def _rename_column_names_to_numeric(X, flatten_tuples=True):
         return pd.DataFrame(X)
 
     X_renamed = X.copy()
+    logical_types = X.ww.logical_types
     if flatten_tuples and (len(X.columns) > 0 and isinstance(X.columns, pd.MultiIndex)):
         flat_col_names = list(map(str, X_renamed.columns))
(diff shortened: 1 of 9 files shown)


evalml/utils/gen_utils.py
```diff
@@ -261,9 +261,11 @@ def _rename_column_names_to_numeric(X, flatten_tuples=True):
         return pd.DataFrame(X)
 
     X_renamed = X.copy()
+    logical_types = X.ww.logical_types
     if flatten_tuples and (len(X.columns) > 0 and isinstance(X.columns, pd.MultiIndex)):
```
eccabay, line 265: If I'm reading this correctly, it seems like we no longer ever set `flatten_tuples` to False. In that case, are we ok to just remove the argument entirely?

eccabay (approved):
LGTM! Just left a few small comments.

Issue #2237, opened 2021-05-06
Pmdarima can't be installed via conda alongside sktime


ParthivNaresh:
Currently `pmdarima` and `sktime` are incompatible when installing through conda. Therefore `pmdarima` can't be included in the feedstock as it will fail to install all dependencies due to conflicts.

This issue tracks [that conflict](https://github.com/alan-turing-institute/sktime/issues/860).

dsherry:
Yep. Thanks for filing. Tracks known limitation. Hopefully the community can resolve this. Let's check back on this soon.

freddyaboulton:
May be fixed. Marking as tech debt. 

https://github.com/alkaline-ml/pmdarima/issues/429

Closing criteria is to add pmdarima to our feedstock recipe?

PR #3505, opened 2022-05-08
Added pmdarima to conda

chukarsten:
Fixes: #2237


evalml/tests/pipeline_tests/test_pipelines.py
```diff
@@ -132,7 +132,7 @@ def test_all_estimators(
         assert len((_all_estimators_used_in_search())) == 9
     else:
         if is_using_conda:
-            n_estimators = 16
```
chukarsten, line 135: Was talking to @jeremyliweishih about this one.  Apparently this line isn't covered in the global checks, but my fiddling with it caused the patch diff to fail.

jeremyliweishih (approved):
LGTM 💪

PR #3506, opened 2022-05-08
Updated NoVarianceDataCheck to return only Warnings

ParthivNaresh:


docs/source/user_guide/data_check_actions.ipynb
```diff
@@ -99,9 +99,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# let's copy the datetime at row 1 for future use\n",
```
ParthivNaresh, line 102: `DateTimeFormatDataCheck` isn't run for a binary problem type so all datetime error references are obsolete

docs/source/user_guide/data_check_actions.ipynb
```diff
@@ -301,47 +292,6 @@
     "results_no_errors = search_iterative(X_train_no_errors, y_train_no_errors, problem_type='binary')\n",
     "results_no_errors"
    ]
-  },
```
ParthivNaresh, line 304: After these changes, the only error left is that the target has null. That leaves little reason to keep this in, but I think a separate issue should be filed to create a more robust example using more data checks with errors.

chukarsten: Sure, please feel free to file and complete!

chukarsten (approved):
I think once I understand why we're doing this, this looks good to go.

PR #3507, opened 2022-05-09
Added threshold for NullDataCheck

### LINKED ISSUE #3286: Add thresholding for warning and suggesting `IMPUTE_COL` action in `NullDataCheck`
Brief discussion with @asniyaz: it could be a good idea to include a threshold (% of nulls) before issuing a warning that a column contains null values.

Right now, we issue a DROP_COL action if the column is highly null, and a IMPUTE_COL action if the column has any null values.

However, if a user has 500 columns and each column has just one or two null values, should we really raise a warning about every single column? Especially if we automatically impute columns in AutoMLSearch anyways? Or should we follow our approach with highly nulls and add a threshold, and only if a column is a certain % null (but not highly null) do we issue a warning?

ParthivNaresh:
Fixes #3286

### DIFF
--- a/evalml/data_checks/null_data_check.py
+++ b/evalml/data_checks/null_data_check.py
@@ -28,19 +28,15 @@ class NullDataCheck(DataCheck):
         pct_moderately_null_col_threshold=0.20,
         pct_null_row_threshold=0.95,
     ):
-        if pct_null_col_threshold < 0 or pct_null_col_threshold > 1:
+        if not 0 <= pct_null_col_threshold <= 1:
             raise ValueError(
                 "`pct_null_col_threshold` must be a float between 0 and 1, inclusive."
             )
-        if (
-            pct_moderately_null_col_threshold < 0
-            or pct_moderately_null_col_threshold > 1
-            or pct_moderately_null_col_threshold > pct_null_col_threshold
-        ):
+        if not 0 <= pct_moderately_null_col_threshold <= pct_null_col_threshold <= 1:
             raise ValueError(
                 "`pct_moderately_null_col_threshold` must be a float between 0 and 1, inclusive, and must be less than or equal to `pct_null_col_threshold`."
             )
-        if pct_null_row_threshold < 0 or pct_null_row_threshold > 1:
+        if not 0 <= pct_null_row_threshold <= 1:
             raise ValueError(
                 "`pct_null_row_threshold` must be a float between 0 and 1, inclusive."
             )
--- a/evalml/data_checks/null_data_check.py
+++ b/evalml/data_checks/null_data_check.py
@@ -101,7 +101,7 @@ class NullDataCheck(DataCheck):
             ...         ]
             ...     },
             ...     {
-            ...         "message": "Column(s) 'few_null' have between 20% and 50% null values",
+            ...         "message": "Column(s) 'few_null' have between 20.0% and 50.0% null values",
             ...         "data_check_name": "NullDataCheck",
             ...         "level": "warning",
             ...         "details": {"columns": ["few_null"], "rows": None},
@@ -177,7 +177,7 @@ class NullDataCheck(DataCheck):
             ...         ]
             ...     },
             ...     {
-            ...         "message": "Column(s) 'lots_of_null', 'few_null' have between 20% and 95% null values",
+            ...         "message": "Column(s) 'lots_of_null', 'few_null' have between 20.0% and 95.0% null values",
             ...         "data_check_name": "NullDataCheck",
             ...         "level": "warning",
             ...         "details": {"columns": ["lots_of_null", "few_null"], "rows": None},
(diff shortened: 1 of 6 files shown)


evalml/data_checks/null_data_check.py
```diff
@@ -15,21 +15,38 @@ class NullDataCheck(DataCheck):
...
-                "pct null column threshold must be a float between 0 and 1, inclusive."
+                "`pct_null_col_threshold` must be a float between 0 and 1, inclusive."
+            )
+        if (
+            pct_moderately_null_col_threshold < 0
+            or pct_moderately_null_col_threshold > 1
+            or pct_moderately_null_col_threshold > pct_null_col_threshold
+        ):
+            raise ValueError(
+                "`pct_moderately_null_col_threshold` must be a float between 0 and 1, inclusive, and must be less than or equal to `pct_null_col_threshold`."
```
ParthivNaresh, line 41: I made `pct_moderately_null_col_threshold` less than or equal to `pct_null_col_threshold` instead of just less than so if a user wanted to set `pct_null_col_threshold` to 0.0 so that any column with null values returned a highly null warning, they can do so by setting `pct_moderately_null_col_threshold` as 0.0.

evalml/pipelines/components/transformers/preprocessing/time_series_regularizer.py
```diff
@@ -117,7 +117,7 @@ def fit(self, X, y=None):
             ww_payload = self.frequency_payload
         else:
             ww_payload = infer_frequency(
-                X[self.time_index],
+                X_ww[self.time_index],
```
ParthivNaresh, line 120: Sneaking this in here

evalml/data_checks/null_data_check.py
chukarsten: Is it at all useful to return something like "0 < pct_moderately_null_col_threshold < pct_null_col_threshold < 1"

evalml/data_checks/null_data_check.py
```diff
@@ -15,21 +15,38 @@ class NullDataCheck(DataCheck):
...
+             but is less than the percentage specified in pct_null_col_threshold, that column will be considered
+             moderately-null. Defaults to 0.20.
         pct_null_row_threshold(float): If the percentage of NaN values in an input row exceeds this amount,
             that row will be considered highly-null. Defaults to 0.95.
     """
 
-    def __init__(self, pct_null_col_threshold=0.95, pct_null_row_threshold=0.95):
+    def __init__(
+        self,
+        pct_null_col_threshold=0.95,
```
chukarsten, line 27: Do we want to change the name of this to reflect that this is "more null" than "moderately null"?  Like "heavily null" or something.  I dunno, I'm not a thesaurus.

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -146,7 +146,7 @@ def get_expected_messages(problem_type):
             ],
         ).to_dict(),
         DataCheckWarning(
-            message="Column(s) 'lots_of_null', 'nullable_integer', 'nullable_bool' have null values",
+            message="Column(s) 'lots_of_null', 'nullable_integer', 'nullable_bool' have between 20.0% and 95.0% null values",
```
chukarsten, line 149: This is a nit, but is there any value between making these values like a module level variable for the data check and import it here to build this string?  Or do we just want to accept that we might change the datacheck's default values and then have to update this test?

evalml/tests/data_checks_tests/test_null_data_check.py
```diff
@@ -61,22 +70,39 @@ def test_highly_null_data_check_init():
...
     with pytest.raises(
         ValueError,
-        match="pct null column threshold must be a float between 0 and 1, inclusive.",
+        match="`pct_null_col_threshold` must be a float between 0 and 1, inclusive.",
     ):
         NullDataCheck(pct_null_col_threshold=-0.1)
     with pytest.raises(
         ValueError,
-        match="pct null column threshold must be a float between 0 and 1, inclusive.",
+        match="`pct_null_col_threshold` must be a float between 0 and 1, inclusive.",
```
chukarsten, line 78: nice change, thank you

chukarsten (approved):
This looks solid to me.  Left a few nits and some things that aren't blocking that you can feel free to reject.

evalml/data_checks/null_data_check.py
ParthivNaresh: The only reason I chose not to was to prevent this from becoming a breaking change!

evalml/data_checks/null_data_check.py
ParthivNaresh: I like it!

evalml/tests/data_checks_tests/test_data_checks.py
ParthivNaresh: Hmm I think we could try something like what woodwork [has](https://github.com/alteryx/woodwork/blob/main/woodwork/config.py#L3). Since we have enough instances throughout our data checks that if we go down the path of creating a config dc defaults file with all default parameters, we can call upon that in all our tests if we aren't using the default parameters. Should I file an issue for this?

PR #3517, opened 2022-05-12
Update featuretools main workflow

rwedge:
PR to iron out issues with the featuretools main CI workflow


rwedge:
@freddyaboulton perhaps could we use `make` commands to make the install commands less dependent on setup.py / setup.cfg configuration

freddyaboulton:
That's a great idea @rwedge ! I wonder if we can use the existing `make installdeps-test` command or if we need to add a new one to install from the same package. Both sound equivalent to me.

.github/workflows/unit_tests_with_featuretools_main_branch.yml
```diff
@@ -18,15 +18,15 @@ jobs:
...
+      - name: Update apt and install Graphviz
+        run: sudo apt update && sudo apt install -y graphviz
       - name: Set up pip
         run: |
           pip config --site set global.progress_bar off
           python -m pip install -U pip
       - name: Install EvalML & Featuretools - test requirements
         run: |
-          python -m pip install -e unpacked_sdist/[test]
+          make installdeps-test
```
gsheni, line 29: Wouldn't it be best to package EvalML and then install?

rwedge: I was worried that `make installdeps-test` wouldn't work nicely with packaged EvalML but thinking about it more it should probably work

Worth noting that this workflow seems to be the only version of the unit tests that packages EvalML

gsheni: I disagree, we package Featuretools `main` with Woodwork `main`:
https://github.com/alteryx/featuretools/blob/main/.github/workflows/unit_tests_with_woodwork_main_branch.yml#L25

rwedge: What do you disagree with?

gsheni: nvm. I get your point. Up to you if we should package

jeremyliweishih (approved):
🤌

.github/workflows/unit_tests_with_featuretools_main_branch.yml
```diff
@@ -18,22 +18,26 @@ jobs:
...
         run: |
           pip config --site set global.progress_bar off
           python -m pip install -U pip
+      - name: Build EvalMl package
+        run: make package_evalml
       - name: Install EvalML & Featuretools - test requirements
         run: |
-          python -m pip install -e unpacked_sdist/[test]
+          cd unpacked_sdist
+          make -f ../Makefile installdeps-prophet
```
gsheni, line 32: There is no point in running `make package_evalml` of you are just going to run `installdeps` which just installs from the requirement text files.

gsheni (approved):

PR #3521, opened 2022-05-12
Update the Time Series User Guide page

### LINKED ISSUE #3335: Improve time series documentation
The current section explaining time series in our docs is missing some info and has some typos.

In the attached screenshot, "Here is an example of these two methods in action" only has one example, and there is a typo with "sunseen" instead of "unseen". There is also no section explaining known_in_advanced features, here or anywhere in the documentation.

The section seems more like a tutorial than a user guide page, so I would suggest we move most of the current content to a time series tutorial and update the user guide page to focus more on the specific functionality evalml provides.

eccabay:
Closes #3335


chukarsten (approved):
Looks great! Thank you!

freddyaboulton (approved):
Awesome! Thanks @eccabay !

PR #3540, opened 2022-06-03
Surpress warnings in target encoder

freddyaboulton:
### Pull Request Description
Fix nightlies. Surpress warnings in target encoder due to https://github.com/scikit-learn-contrib/category_encoders/issues/327

-----

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/target_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/target_encoder.py
@@ -1,4 +1,6 @@
 """A transformer that encodes categorical features into target encodings."""
+import warnings
+
 import pandas as pd
 
 from ..transformer import Transformer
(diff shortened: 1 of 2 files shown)


evalml/pipelines/components/transformers/encoders/target_encoder.py
```diff
@@ -65,9 +67,14 @@ def __init__(
             "category_encoders",
             error_msg="category_encoders not installed. Please install using `pip install category_encoders`",
         )
+        # Supress warnings for now due to problems discussion here:
+        # https://github.com/scikit-learn-contrib/category_encoders/issues/327
+        with warnings.catch_warnings():
```
freddyaboulton, line 72: We could also change the default value of `smoothing` and add `min_samples_leaf` parameter but there is still some discussion as to what the new defaults should be so figured it was better to suppress the warnings

jeremyliweishih: agreed!

jeremyliweishih (approved):
thank you!

PR #3559, opened 2022-06-10

machineFL:
This is an auto-generated PR with **latest** dependency updates.


evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -3,7 +3,7 @@ click==8.1.3
 cloudpickle==2.1.0
 colorama==0.4.4
 dask==2022.5.2
-featuretools==1.9.1
+featuretools==1.9.2
```
freddyaboulton, line 6: This is probably a bug in the diff but won't make a difference in the end

freddyaboulton (approved):

PR #3564, opened 2022-06-13
Changes for latest sktime

freddyaboulton:
### Pull Request Description

-----


evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -95,6 +94,7 @@ def __init__(
         )
         arima_model = sktime_arima.AutoARIMA(**parameters)
         parameters["use_covariates"] = use_covariates
+        parameters["time_index"] = time_index
```
jeremyliweishih, line 97: would this interfere with the previous versions? I guess I'm just wondering why we used kwargs in the first place?

jeremyliweishih (approved):
Just a clarifying comment! LGTM otherwise.

PR #3569, opened 2022-06-15
2734 Component Graph Instantiation Error

### LINKED ISSUE #2734: Issue calling ComponentGraph's `describe()` method if ComponentGraph is not instantiated
If a user does not instantiate a ComponentGraph before calling `describe()`, we error out with a cryptic error.

```
component_dict = {
        "Imputer": ["Imputer", "X", "y"],
        "Target Imputer": ["Target Imputer", "X", "y"],
        "OneHot_RandomForest": ["One Hot Encoder", "Imputer.x", "Target Imputer.y"],
        "OneHot_ElasticNet": ["One Hot Encoder", "Imputer.x", "y"],
        "Random Forest": ["Random Forest Classifier", "OneHot_RandomForest.x", "y"],
        "Elastic Net": ["Elastic Net Classifier", "OneHot_ElasticNet.x", "Target Imputer.y"],
        "Logistic Regression": [
            "Logistic Regression Classifier",
            "Random Forest.x",
            "Elastic Net.x",
            "y",
        ],
}
cg_with_estimators = ComponentGraph(component_dict)
cg_with_estimators.describe()
```

```
<ipython-input-5-3005b62da7ab> in <module>
     14 }
     15 cg_with_estimators = ComponentGraph(component_dict)
---> 16 cg_with_estimators.describe()

~/Desktop/evalml/evalml/pipelines/component_graph.py in describe(self, return_dict)
    514             logger.info(component_string)
    515             import pdb; pdb.set_trace()
--> 516             components.update(
    517                 {
    518                     component.name: component.describe(

TypeError: describe() missing 1 required positional argument: 'self'
```

We should raise specific error if calling `describe` on graph that is not instantiated, since there will be no parameters available.

MichaelFu512:
Closes #2734.

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,8 +2,8 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
-        * Using the ``describe`` method on an un-instantiated ComponentGraph now gives a more helpful error message :pr:`3569`
     * Fixes
+        * Added a clearer error message when ``describe`` is called on an un-instantiated ComponentGraph :pr:`3569`
     * Changes
         * Don't pass ``time_index`` as kwargs to sktime ARIMA implementation for compatibility with latest version :pr:`3564`
     * Documentation Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,6 +2,7 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
+        * Using the ``describe`` method on an un-instantiated ComponentGraph now gives a more helpful error message :pr:`3569`
     * Fixes
     * Changes
         * Don't pass ``time_index`` as kwargs to sktime ARIMA implementation for compatibility with latest version :pr:`3564`
--- a/evalml/pipelines/component_graph.py
+++ b/evalml/pipelines/component_graph.py
@@ -24,9 +24,6 @@ from evalml.utils import (
     infer_feature_types,
 )
 
-# from tkinter import W
-
-
 logger = get_logger(__file__)
 
 
@@ -686,11 +683,11 @@ class ComponentGraph:
         Args:
             return_dict (bool): If True, return dictionary of information about component graph. Defaults to False.
 
-        Raises:
-            TypeError: If ComponentGraph is not instantiated
-
         Returns:
             dict: Dictionary of all component parameters if return_dict is True, else None
+
+        Raises:
+            ValueError: If the componentgraph is not instantiated
         """
         logger = get_logger(f"{__name__}.describe")
         components = {}
@@ -706,9 +703,9 @@ class ComponentGraph:
                     }
                 )
             except TypeError:
-                raise TypeError(
+                raise ValueError(
                     "You need to instantiate ComponentGraph before calling describe()"
-                ) from None
+                )
         if return_dict:
             return components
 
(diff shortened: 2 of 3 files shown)


docs/source/release_notes.rst
```diff
@@ -2,6 +2,7 @@ Release Notes
 -------------
 **Future Releases**
     * Enhancements
+        * Using the ``describe`` method on an un-instantiated ComponentGraph now gives a more helpful error message :pr:`3569`
```
eccabay, line 5: Thanks for adding this! Personally, I would consider this change a "fix" instead of an "enhancement", since it's patching an existing bug.

Also, [this is a super nitpick and we haven't been doing a great job of maintaining the pattern recently but] generally we try to keep our changes in past tense (v.0.52.0 is a good example of that). So instead, this commend might be something closer to "Added a clearer error message when ``describe`` is called on an un-instantiated ComponentGraph"

evalml/pipelines/component_graph.py
```diff
@@ -24,6 +24,9 @@
     infer_feature_types,
 )
 
+# from tkinter import W
+
+
```
eccabay, line 29: Can you remove these lines completely?

evalml/pipelines/component_graph.py
```diff
@@ -691,13 +697,18 @@ def describe(self, return_dict=False):
...
+            try:
+                components.update(
+                    {
+                        component.name: component.describe(
+                            print_name=False, return_dict=return_dict
+                        )
+                    }
+                )
+            except TypeError:
+                raise TypeError(
```
eccabay, line 709: I think a different error would be more useful here, a TypeError doesn't necessarily make sense for illegally calling a method. Personally, I always default to a ValueError.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -1685,6 +1685,30 @@ def test_describe_component_graph(return_dict, example_graph, caplog):
         assert component.name in out
 
 
+def test_describe_component_graph_TypeError():
```
eccabay, line 1688: Nitpick: function names should be exclusively snake case - `typeerror` or `type_error` is just fine in this case.

eccabay (changes requested):
Fantastic job getting this up, Michael! I've left a few suggestions for changes, mostly just to follow our conventions.

evalml/pipelines/component_graph.py
```diff
@@ -691,13 +697,18 @@ def describe(self, return_dict=False):
...
+                    "You need to instantiate ComponentGraph before calling describe()"
+                ) from None
```
jeremyliweishih, line 711: do we need this `None` here? If not we can remove it!

MichaelFu512: If I don't include `from None` it spits out 2 error messages rather than one weirdly enough (I'm not entirely sure why it does that).

jeremyliweishih: can you show me the stack trace if you have it?

eccabay (approved):
Fantastic!

jeremyliweishih (approved):
LGTM - 🚢

evalml/pipelines/component_graph.py
chukarsten: Agreed, let's try to refrain from committing any unused, commented code.

PR #3577, opened 2022-06-21
2470 report batch times automlsearch

### LINKED ISSUE #2470: Report batch times in AutoMLSearch
Right now, we only report the amount of time search takes to entirely complete. I wonder if there's a way to report the amount of time a specific batch takes to complete. Even more interesting could be the time each pipeline takes.

I noticed this while debugging for #2410 via LG, because I wanted to confirm that an increase in runtime was indeed related to XGBoost / Catboost pipelines in particular but realized we only know the complete automl fit time.

MichaelFu512:
### Pull Request Description
Added an option to record batch and pipeline search times from `automl.search()` by using an optional parameter called "timing". Using timing = True will output the batch timings to stdout. `automl.search()` now also returns a dictionary that holds individual batches and pipelines times.

There's also a value in the inner dictionary called "Total time of batch" which records how long the batch took in total.

Closes #2470

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -857,7 +857,10 @@ class AutoMLSearch:
             else:
                 leading_char = ""
 
-    def search(self, show_iteration_plot=True):
+    def search(self, show_iteration_plot=True, timing=None):
+        batch_times = {}
+        print(timing, show_iteration_plot)
+        count = 0
         """Find the best pipeline for the data set.
 
         Args:
@@ -924,7 +927,10 @@ class AutoMLSearch:
         current_batch_pipeline_scores = []
         new_pipeline_ids = []
         loop_interrupted = False
+
         while self._should_continue():
+            pipeline_time = {}
+            total_batch_time = time.time()
             computations = []
             try:
                 if not loop_interrupted:
@@ -933,12 +939,14 @@ class AutoMLSearch:
                 self.logger.info("AutoML Algorithm out of recommendations, ending")
                 break
             try:
+                
                 if self._should_continue():
                     new_pipeline_ids = []
                     log_title(
                         self.logger,
                         f"Evaluating Batch Number {self._get_batch_number()}",
                     )
+                    batch_start_time = time.time()
                     for pipeline in current_batch_pipelines:
                         self._pre_evaluation_callback(pipeline)
                         computation = self._engine.submit_evaluation_job(
@@ -952,6 +960,7 @@ class AutoMLSearch:
                         current_computation_index
                     ]
                     if computation.done() and not has_been_processed:
+                        start_pipeline_time = time.time()
                         evaluation = computation.get_result()
                         data, cached_data, pipeline, job_log = (
                             evaluation.get("scores"),
@@ -962,6 +971,8 @@ class AutoMLSearch:
                         pipeline_id = self._post_evaluation_callback(
                             pipeline, data, cached_data, job_log
                         )
+                        pipeline_time[pipeline.name] = time_elapsed(start_pipeline_time)
+                        print(pipeline_time[pipeline.name])
                         new_pipeline_ids.append(pipeline_id)
                         computations[current_computation_index] = (computation, True)
                         computations_left_to_process -= 1
@@ -983,6 +994,7 @@ class AutoMLSearch:
             current_batch_pipeline_scores = full_rankings[current_batch_idx][
                 "validation_score"
             ]
+            
             if (
                 len(current_batch_pipeline_scores)
                 and current_batch_pipeline_scores.isna().all()
@@ -990,12 +1002,22 @@ class AutoMLSearch:
                 raise AutoMLSearchException(
                     f"All pipelines in the current AutoML batch produced a score of np.nan on the primary objective {self.objective}."
                 )
+            pipeline_time["Total time of batch " + str(self._get_batch_number())] = time_elapsed(total_batch_time)
+            batch_times[self._get_batch_number()] = pipeline_time
 
         self.search_duration = time.time() - self._start
         elapsed_time = time_elapsed(self._start)
         desc = f"\nSearch finished after {elapsed_time}"
         desc = desc.ljust(self._MAX_NAME_LEN)
         self.logger.info(desc)
+
+        if timing == "log":
+            #TODO: Integrate into logger.py, get rid of prints / use logs
+            for batch in batch_times:
+                print("Batch",batch,"time stats:")
+                for key in batch_times[batch]:
+                    print(key+":",(batch_times[batch])[key])
+                    print()
 
         self._find_best_pipeline()
         if self._best_pipeline is not None:
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -865,17 +865,29 @@ class AutoMLSearch:
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
             timing (str, None): Shows timing of the batches and the individual timings of each pipeline.
-                Doesn't show by default.
+                Default: None
+                log=prints out batch/pipeline timing to console.
+                return=returns a dict with all the batch/pipeline timings
+                both=both prints out the timing to the console and returns a dict with the timings
 
         Raises:
             AutoMLSearchException: If all pipelines in the current AutoML batch produced a score of np.nan on the primary objective.
+            ValueError: If timing is not set correctly
 
         Returns:
-            Optinal Dict[int, Dict[str, str]]: Returns dict if timing is set to "return" or "both".
-                Key=batch #, value=dictionary[key=pipeline name, value=time of pipeline].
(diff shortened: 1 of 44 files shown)


evalml/automl/automl_search.py
```diff
@@ -857,16 +858,24 @@ def _handle_keyboard_interrupt(self):
...
                 leading_char = ""
 
-    def search(self, show_iteration_plot=True):
+    def search(self, show_iteration_plot=True, timing=None):
         """Find the best pipeline for the data set.
 
         Args:
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
+            timing (str, None): Shows timing of the batches and the individual timings of each pipeline.
```
jeremyliweishih, line 867: Nit: "Default: None".

Can you also show the options? Explain what "log", "return" and "both" does here.

evalml/automl/automl_search.py
```diff
@@ -857,16 +858,24 @@ def _handle_keyboard_interrupt(self):
...
+                Doesn't show by default.
 
         Raises:
             AutoMLSearchException: If all pipelines in the current AutoML batch produced a score of np.nan on the primary objective.
+
+        Returns:
+            Optinal Dict[int, Dict[str, str]]: Returns dict if timing is set to "return" or "both".
```
jeremyliweishih, line 874: "Optional" in line 873. I have a spell checker in VSCode that helps with this 😄

jeremyliweishih: Also, this should be `Dict[int, Dict[str, Timestamp]]`

evalml/automl/automl_search.py
```diff
@@ -1656,3 +1684,16 @@ def get_ensembler_input_pipelines(self, ensemble_pipeline_id):
         return self._results["pipeline_results"][ensemble_pipeline_id][
             "input_pipeline_ids"
         ]
+
+    def is_timing_input_correct(self, timing):
```
jeremyliweishih, line 1688: we should raise a `ValueError` or at least a warning if the user provides an input we don't accept. I think I would go for a `ValueError` though! Be sure to add a test case too.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -216,6 +223,61 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type, obje
     )
 
 
+def test_search_batch_times(caplog):
+    caplog.clear()
+    X, y = load_data(
```
jeremyliweishih, line 228: We can use a smaller dataset here: check out the `X_y_binary` fixture in `contest.py` and just search for tests that use it as an example.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -216,6 +223,61 @@ def test_search_results(X_y_regression, X_y_binary, X_y_multi, automl_type, obje
...
+        X,
+        feature_types={
+            "store_id": "categorical",
+            "expiration_date": "datetime",
+            "lat": "categorical",
+            "lng": "categorical",
+            "provider": "categorical",
+        },
+    )
+    X_train, _, y_train, _ = evalml.preprocessing.split_data(
```
jeremyliweishih, line 251: We can skip out on all this logic about splitting and datachecks as well. Look at something like `test_pipeline_score_raises` for an example.

jeremyliweishih (changes requested):
Just some initial comments - lmk if they make sense! I'll give it another final review after your edits. Thanks!

chukarsten (changes requested):
Thanks for tackling this and for doing it so quickly!

evalml/automl/automl_search.py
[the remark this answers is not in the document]
MichaelFu512: Do you think I should get rid of "return" and "both" as options for timing if we want it to always return this dictionary (meaning we only keep "log" as an option for timing)?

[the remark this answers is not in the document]
chukarsten: I'm sorry this took me so long, but yea, I think yes, it's fine to always return the dictionary with all the things in it.

evalml/automl/automl_search.py
```diff
@@ -857,16 +858,34 @@ def _handle_keyboard_interrupt(self):
...
-    def search(self, show_iteration_plot=True):
+    def search(self, show_iteration_plot=True, timing=None):
         """Find the best pipeline for the data set.
 
         Args:
             show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
+            timing (str, None): Shows timing of the batches and the individual timings of each pipeline.
+                Default: None
+                log=prints out batch/pipeline timing to console.
```
eccabay, line 869: Since the only options are "log" or None, I vote that we switch to a boolean flag for this, something like `log_timing` which defaults to false. It'd make both our lives as the implementers easier (checking a boolean instead of string equality, not having to validate the input) and make it clearer for users to boot.

evalml/automl/automl_search.py
```diff
@@ -857,16 +858,34 @@ def _handle_keyboard_interrupt(self):
...
 
         Raises:
             AutoMLSearchException: If all pipelines in the current AutoML batch produced a score of np.nan on the primary objective.
+            ValueError: If timing is not set to a correct value
+
+        Returns:
+                Dict[int, Dict[str, Timestamp]]: Returns dict.
+                Key=batch #, value=Dict[key=pipeline name, value=timestamp of pipeline].
+                Inner dict has key called "Total time of batch" with value=total time of batch.
```
eccabay, line 878: This is really hard to understand without reading closely. I'd refactor it, something more like:
```
Dict[int, Dict[str, Timestamp]]: Dictionary keyed by batch number that maps to the timings for pipelines run in that batch, 
as well as the total time for each batch. Pipelines within a batch are labeled by pipeline name. 
```

As a side note, `=` in docstrings really throws me off. It'd be better to stick to using colons, which maintain consistency with the rest of our docs!

evalml/automl/automl_search.py
```diff
@@ -983,20 +1009,28 @@ def search(self, show_iteration_plot=True):
...
 
         self.search_duration = time.time() - self._start
         elapsed_time = time_elapsed(self._start)
         desc = f"\nSearch finished after {elapsed_time}"
         desc = desc.ljust(self._MAX_NAME_LEN)
         self.logger.info(desc)
 
+        if timing == "log":
+            log_title(self.logger, "Batch Time Stats")
+            log_batch_times(self.logger, batch_times)
```
eccabay, line 1032: I would move the call to `log_title` into `log_batch_times` itself, since we don't need to call `log_batch_times` without setting the title as well.

evalml/automl/automl_search.py
```diff
@@ -857,16 +858,34 @@ def _handle_keyboard_interrupt(self):
             else:
                 leading_char = ""
 
-    def search(self, show_iteration_plot=True):
+    def search(self, show_iteration_plot=True, timing=None):
```
eccabay, line 861: I think we should move this to be an argument in `AutoMLSearch.__init__` instead of `AutoMLSearch.search`. Reason being, we have two ways for users to run search. This is one of them, but we're trying to move more over to running the top level `search` method instead of manually instantiating `AutoMLSearch` first. With the argument living here, users have no access to the argument. 

If we move the arg to `AutoMLSearch.__init__` and add it to the top level search methods as well, that will ensure users have full access to controlling this.

eccabay (changes requested):
Huge fan of this, but I think we can make things a little clearer/more accessible for our users, most importantly where and what the argument that controls the behavior lives.

jeremyliweishih (commented):
Agreed with @eccabay's comments. @MichaelFu512 can you request a re-review once those changes are in? Thanks!

evalml/automl/automl_search.py
```diff
@@ -143,6 +145,7 @@ def search(
             in time series problems, values should be passed in for the time_index, gap, forecast_horizon, and max_delay variables.
         n_splits (int): Number of splits to use with the default data splitter.
         verbose (boolean): Whether or not to display semi-real-time updates to stdout while search is running. Defaults to False.
+        timing (boolean): Whether or not to display pipeline search times to stdout. Defaults to False.
```
eccabay, line 148: Nitpicky clarification: logging info is not guaranteed to display that information in stdout, that will only happen if the logging level is set low enough to expose it. It'd be more accurate to say:
```
Whether or not to write pipeline search times to the logger. Defaults to False.
```

By default, if `timing` is set to True, the user would still not see the timings being logged in stdout since the default logging behavior is at the `warning` level (will not show info/debug level logs). To see this, they would either need to configure the logger themselves, or set `verbose=True`. Alternatively, they can choose to dump the log into a file or otherwise redirect that info, in which case it would be logged somewhere but not appearing in stdout.

Sorry for the info dump, I'm just intimately familiar with our logging behavior 😅

MichaelFu512: It's always good for me to learn more so "info dump(s)" are always great.

chukarsten (approved):

PR #3579, opened 2022-06-21
Better error msg for time series predict + tests

### LINKED ISSUE #3443: Provide a useful error when there is no y value passed to time series predict
Something along the lines of "`predict` requires `y_train` in order to move indices for prediction"!

#### Repro

```python
import evalml

X_ts_regress, y_ts_regress = evalml.demos.load_weather()
X_train_ts_reg, X_test_ts_reg, y_train_ts_reg, y_test_ts_reg = evalml.preprocessing.split_data(
    X_ts_regress, 
    y_ts_regress, 
    problem_type='time series regression'
)

problem_config = {"gap": 0, "max_delay": 7, "forecast_horizon": 7, "time_index": "Date"}
pipeline_ts_regress = TimeSeriesRegressionPipeline(
    component_graph=['One Hot Encoder', 'Linear Regressor'], 
    parameters={"pipeline": problem_config}
)
pipeline_ts_regress.fit(X_train_ts_reg, y_train_ts_reg)
y_pred = pipeline_ts_regress.predict(X_test_ts_reg)
```
Raises this error:
Screen Shot 2022-04-05 at 10 53 38 AM

MichaelFu512:
### Pull Request Description
Closes #3443 

-----

### DIFF
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -190,11 +190,11 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
         """
         if X_train is None:
             raise ValueError(
-                "Make sure to have a non None value for X_train when calling time series' predict"
+                "Make sure to include an input for X_train when calling time series' predict"
             )
         elif y_train is None:
             raise ValueError(
-                "Make sure to have a non None value for y_train when calling time series' predict"
+                "Make sure to include an input for y_train when calling time series' predict"
             )
         X_train, y_train = self._convert_to_woodwork(X_train, y_train)
         if self.estimator is None:
--- a/evalml/tests/pipeline_tests/test_time_series_pipeline.py
+++ b/evalml/tests/pipeline_tests/test_time_series_pipeline.py
@@ -1088,7 +1088,7 @@ def time_series_default_pipeline_classification_parameters():
         ProblemTypes.TIME_SERIES_REGRESSION,
     ],
 )
-def test_time_series_pipeline_predict_valueerror(
+def test_time_series_pipeline_predict_none_parameter_valueerror(
     problem_type,
     ts_data_binary,
     ts_data_multi,
@@ -1134,12 +1134,12 @@ def test_time_series_pipeline_predict_valueerror(
     clf.fit(X_train, y_train)
     with pytest.raises(
         ValueError,
-        match="Make sure to have a non None value for y_train when calling time series' predict",
+        match="Make sure to include an input for y_train when calling time series' predict",
     ):
         clf.predict(X=X, X_train=X_train)
     with pytest.raises(
         ValueError,
-        match="Make sure to have a non None value for X_train when calling time series' predict",
+        match="Make sure to include an input for X_train when calling time series' predict",
     ):
         clf.predict(X=X)
 


docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
         * Updated the Imputer and SimpleImputer to work with scikit-learn 1.1.1. :pr:`3525`
         * Bumped the minimum versions of scikit-learn to 1.1.1 and imbalanced-learn to 0.9.1. :pr:`3525`
         * Added a clearer error message when ``describe`` is called on an un-instantiated ComponentGraph :pr:`3569`
+        * Added a clear error message when ``predict`` is called without a y_train parameter :pr:`3579`
```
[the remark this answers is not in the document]
eccabay: Also, this should mention the error message for `X_train` as well, since we now have a clearer error message in that case too!

eccabay (changes requested):
This is looking great! I just left a few suggestions on making this code as robust as it can be.

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -183,12 +183,18 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
...
         Raises:
-            ValueError: If final component is not an Estimator.
+            ValueError: If X_train and/or y_train has no value or if final component is not an Estimator.
 
         Returns:
             Predictions.
         """
-        X_train, y_train = self._convert_to_woodwork(X_train, y_train)
+        try:
+            print("HELLO")
```
jeremyliweishih, line 192: need to remove this print

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -183,12 +183,18 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
             y_train (pd.Series or None): Training labels.
```
jeremyliweishih, line 183: @ParthivNaresh if we need `X_train` and `y_train` should we still accept `None` as input? I don't have the context on why we allowed this before!

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -183,12 +183,18 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
...
+            ValueError: If X_train and/or y_train has no value or if final component is not an Estimator.
 
         Returns:
             Predictions.
         """
-        X_train, y_train = self._convert_to_woodwork(X_train, y_train)
+        try:
+            print("HELLO")
+            X_train, y_train = self._convert_to_woodwork(X_train, y_train)
+        except AttributeError:
```
jeremyliweishih, line 194: Think this is good but we may need to reconsider allow `X_train` or `y_train` as optional arguments. We can also consider just validating that `X_train` and `y_train` exists. Let's keep this for now and see what @ParthivNaresh has to say!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1060,6 +1060,92 @@ def test_binary_predict_pipeline_objective_mismatch(
...
+    ts_data,
+    time_series_binary_classification_pipeline_class,
+    time_series_multiclass_classification_pipeline_class,
+    time_series_regression_pipeline_class,
+):
+    if problem_type == ProblemTypes.TIME_SERIES_BINARY:
+        X, y = ts_data_binary
+        clf = time_series_binary_classification_pipeline_class(
+            parameters={
+                "Logistic Regression Classifier": {"n_jobs": 1},
```
jeremyliweishih, line 1084: if all the parameters are the same for each of these pipelines, we can save the parameters up top and then call it to save some space.

evalml/pipelines/time_series_pipeline_base.py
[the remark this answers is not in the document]
MichaelFu512: oh shoot sorry about that!

chukarsten: I am glad to see that someone else does this as well

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1060,6 +1060,92 @@ def test_binary_predict_pipeline_objective_mismatch(
...
+        )
+
+    X, y = pd.DataFrame(X), pd.Series(y)
+    X_train, y_train = X.iloc[:21], y.iloc[:21]
+    clf.fit(X_train, y_train)
+    with pytest.raises(
+        ValueError,
+        match="Make sure to have a value for both X_train and y_train when calling predict",
+    ):
+        clf.predict(X_train)
```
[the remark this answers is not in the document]
MichaelFu512: It looks like if X_train is None, the `convert_to_woodwork()` function makes X_train a `pd.DataFrame()` and infers its type from there. It doesn't do the same for y_train which is why this initial error occurs. 

So probably, instead of checking both X_train and y_train, I just need to check y_train since if X_train is null the code still runs (though it does make X_train a blank dataframe).

evalml/pipelines/time_series_pipeline_base.py
ParthivNaresh: Yeah I'm not sure why `X_train` and `y_train` are optional here, they're very much required

ParthivNaresh: Oh I think it had to do with the fact that the parent `PipelineBase` has them as optional because regular regression and classification problems don't need them

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1060,6 +1060,67 @@ def test_binary_predict_pipeline_objective_mismatch(
         binary_pipeline.predict(X[30:32], "precision micro", X[:30], y[:30])
 
 
+@pytest.fixture
+def my_parameters():
```
eccabay, line 1064: Generally, if we're going to go through the effort of creating a pytest fixture, we try to use it in more than one test. I'd either put this at the top of the function instead of a fixture, or update the other tests that have the same parameters in this file to use the same fixture.

Also, `my_parameters` isn't a very useful name 😂 could you name it something more problem-specific (and not first person!)

eccabay (commented):
This is looking better! Since we've determined that both X_train and y_train should be required parameters (but we should probably keep the function signature the same as other predict functions), we should shift this back to checking for both the parameters, but separately this time!

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -183,11 +183,19 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
...
         Raises:
-            ValueError: If final component is not an Estimator.
+            ValueError: If X_train and/or y_train are None or if final component is not an Estimator.
 
         Returns:
             Predictions.
         """
+        if X_train is None:
+            raise ValueError(
+                "Make sure to have a non None value for X_train when calling time series' predict and that in the predict method you set X_train = X_train"
```
eccabay, line 193: I don't think you need the "and that in the predict method you set X_train = X_train" here - mirroring the y_train error message is perfect!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1068,20 +1088,32 @@ def test_binary_predict_pipeline_objective_mismatch(
         ProblemTypes.TIME_SERIES_REGRESSION,
     ],
 )
-def test_time_series_pipeline_not_fitted_error(
+def test_time_series_pipeline_no_ytrain_predict_error(
```
eccabay, line 1091: If I'm reading these correctly in that all the pipeline setup is identical, I don't think we need y_train and X_train to be separate. We can just run one raises test after the other to cut down on repeated code.

eccabay (approved):
Awesome, Michael, thanks for making these updates! I just left a final few comments, but other than that this is good to go! 🚢

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -183,11 +183,19 @@ def predict(self, X, objective=None, X_train=None, y_train=None):
...
         Raises:
-            ValueError: If final component is not an Estimator.
+            ValueError: If X_train and/or y_train are None or if final component is not an Estimator.
 
         Returns:
             Predictions.
         """
+        if X_train is None:
+            raise ValueError(
+                "Make sure to have a non None value for X_train when calling time series' predict"
```
jeremyliweishih, line 193: nit: instead of "non Non value" maybe just "to include input for X-train". Double negatives can get a little confusing. just a style thing so up to you!

jeremyliweishih (approved):
Sweet LGTM - just a style comment 😄

evalml/pipelines/time_series_pipeline_base.py
MichaelFu512: For sure!

PR #3592, opened 2022-06-28
1692 iterative graph not showing in docs

### LINKED ISSUE #1692: AutoMLSearch plots not appearing in Docs
The plots generated in AutoMLSearch are not appearing in the [documentation](https://evalml.alteryx.com/en/stable/user_guide/automl.html#Limiting-the-AutoML-Search-Space)

image

MichaelFu512:
### Pull Request Description
Fixes the iterative graph not showing up in docs. This is a placeholder fix because the docs do not actually load the dynamic iterative graph, so what this pr does is, after searching, it graphs all the points and displays that. 

I changed all the docs that search to not show the iterative graph since it will have just shown the user empty graphs.

Closes #1692 

-----

### DIFF
--- a/docs/source/demos/fraud.ipynb
+++ b/docs/source/demos/fraud.ipynb
@@ -115,7 +115,7 @@
     "                      optimize_thresholds=True,\n",
     "                      verbose=True)\n",
     "\n",
-    "automl.search()"
+    "automl.search(interactive_plot=False)"
    ]
   },
   {
@@ -219,7 +219,7 @@
     "                          optimize_thresholds=True,\n",
     "                          verbose=True)\n",
     "\n",
-    "automl_auc.search()"
+    "automl_auc.search(interactive_plot=False)"
    ]
   },
   {
@@ -281,7 +281,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
@@ -296,6 +296,11 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.8.6"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "425958b484b95e5a5ab632cc76b33302e47efa72ef6819296d6f51644ebb328b"
+   }
   }
  },
  "nbformat": 4,
--- a/docs/source/demos/fraud.ipynb
+++ b/docs/source/demos/fraud.ipynb
@@ -281,7 +281,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
   },
@@ -296,11 +296,6 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.8.6"
-  },
-  "vscode": {
-   "interpreter": {
-    "hash": "425958b484b95e5a5ab632cc76b33302e47efa72ef6819296d6f51644ebb328b"
-   }
   }
  },
  "nbformat": 4,
--- a/docs/source/user_guide/automl.ipynb
+++ b/docs/source/user_guide/automl.ipynb
@@ -993,7 +993,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.13"
+   "version": "3.8.6"
   },
   "vscode": {
    "interpreter": {
--- a/docs/source/user_guide/automl.ipynb
+++ b/docs/source/user_guide/automl.ipynb
@@ -858,7 +858,7 @@
     "                                  problem_type=\"binary\",\n",
     "                                  allowed_model_families=[ModelFamily.LINEAR_MODEL],\n",
     "                                  engine=\"cf_threaded\")\n",
-    "automl_cf_threaded.search(show_iteration_plot = False)\n",
+    "automl_cf_threaded.search(interactive_plot = False)\n",
     "automl_cf_threaded.close_engine()"
    ]
   },
@@ -888,7 +888,7 @@
     "                                  problem_type=\"binary\",\n",
     "                                  allowed_model_families=[ModelFamily.LINEAR_MODEL],\n",
     "                                  engine=cf_engine)\n",
-    "automl_cf_threaded.search(show_iteration_plot = False)\n",
+    "automl_cf_threaded.search(interactive_plot = False)\n",
     "automl_cf_threaded.close_engine()"
    ]
   },
@@ -906,7 +906,7 @@
     "automl_cf_process = AutoMLSearch(X_train=X, y_train=y,\n",
     "                                 problem_type=\"binary\",\n",
     "                                 engine=\"cf_process\")\n",
-    "automl_cf_process.search(show_iteration_plot = False)\n",
+    "automl_cf_process.search(interactive_plot = False)\n",
     "automl_cf_process.close_engine()\n",
     "```"
    ]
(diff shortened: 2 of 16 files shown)


evalml/automl/automl_search.py
```diff
@@ -1006,6 +1009,24 @@ def search(self, show_iteration_plot=True):
...
+            # layout = {
+            #     "title": title,
+            #     "xaxis": {"title": "Iteration", "rangemode": "tozero"},
+            #     "yaxis": {"title": "Score"},
+            # }
+            if not show_iteration_plot:
+                self.search_iteration_plot = self.plot.search_iteration_plot(
+                    interactive_plot=show_iteration_plot
+                )
+                return self.search_iteration_plot
```
eccabay, line 1029: Do we need to return this? Does it not render in the docs without it?

evalml/automl/automl_search.py
```diff
@@ -1006,6 +1009,24 @@ def search(self, show_iteration_plot=True):
...
+            # )
+            # layout = {
+            #     "title": title,
+            #     "xaxis": {"title": "Iteration", "rangemode": "tozero"},
+            #     "yaxis": {"title": "Score"},
+            # }
+            if not show_iteration_plot:
+                self.search_iteration_plot = self.plot.search_iteration_plot(
+                    interactive_plot=show_iteration_plot
+                )
```
eccabay, line 1028: I had a really hard time understanding what this does. After falling down an iteration_plot rabbit hole, I think we should rename the `show_iteration_plot` argument to `interactive_plot`, to match the arg that's passed into `search_iteration_plot`. 

In my mind, that's how this logic makes sense. If `interactive_plot` is True, we call `self.plot.search_iteration_plot` up around line 910, which will create the interactive plot as we know and love. If `interactive_plot` is False and `verbose` is True (the other scenario where we'd want a plot), we run this code at the end to generate that final post-search plot that the docs require.

Could you refactor this section and the section around line 910 to follow that logic? It'd keep things much simpler and easier to follow.

evalml/automl/automl_search.py
```diff
@@ -1006,6 +1009,24 @@ def search(self, show_iteration_plot=True):
...
+            #     error_msg="Cannot find dependency plotly.graph_objects",
+            # )
+            # title = "Pipeline Search: Iteration vs. {}<br><sub>Gray marker indicates the score at current iteration</sub>".format(
+            #     self.objective.name
+            # )
+            # layout = {
+            #     "title": title,
+            #     "xaxis": {"title": "Iteration", "rangemode": "tozero"},
+            #     "yaxis": {"title": "Score"},
+            # }
```
eccabay, line 1024: We don't need commented out code!

eccabay (changes requested):
Ok, I fell down into a rabbit hole of iteration plots while reviewing this, and I don't actually think this is a much of a band-aid solution as I originally thought. I left a comment about renaming the `show_iteration_plot` argument, and I think if we do the refactoring I mentioned there this will actually make a lot of sense. 

Also, I don't think we should change the default behavior. It makes it easier to run the docs, but harder on our users to get that interactive plot. I would rather we update the docs to set `show_iteration_plot` (`interactive_plot` after renaming) to be False and add a note there mentioning this is a documentation workaround and you'll see an interactive plot by default.

evalml/automl/automl_search.py
eccabay: To be a little clearer here, I think this section could look like 
```
if self.verbose and not self.interactive_plot:
    self.search_iteration_plot = self.plot.search_iteration_plot(interactive_plot=interactive_plot)
```

chukarsten (commented):
I think Becca's left some good comments for addressing.  Please keep an eye out for commiting a commented code.  We find that a good way to prevent this is, while your PR is in draft status, doing a PR review on yourself and look over the diff.  Review your PR as if you were an approver before changing status to ready!

evalml/automl/automl_search.py
MichaelFu512: From what I remember, without returning the `self.search_iteration_plot`, the docs don't render.

I could use figure.show() but that causes tests to take a really long time. (https://github.com/alteryx/evalml/runs/7100924419?check_suite_focus=true)

TL;DR: the automl tests was taking over an hour when it usually takes like 20-25 minutes when I used figure.show(). When I ran the test locally, figure.show() caused every test from automl to open a web browser tab and show me the graph, multiple times (before I stopped it manually safari had like 30+ tabs of just graphs). I suspect that's the reason why it takes so long (for whatever reason). 

I tried to mess around with figure.show() parameters, but couldn't figure out a way to render the graph so that:
1. It's interactive (and not like a jpg or png of the graph)
2. It doesn't make the local test open up a browser every time it runs `automl.search()`
3. Do both of the above and actually render when I build the doc using sphinx (sometimes it loads locally, then I check and it gives an error message when I upload it to github).

Of course there's a high chance I'm missing something pretty obvious (maybe I could try making it so that it tries to use your web-browser to render it, then it doesn't?).

docs/source/user_guide/automl.ipynb
```diff
@@ -994,6 +994,11 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.8.6"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "425958b484b95e5a5ab632cc76b33302e47efa72ef6819296d6f51644ebb328b"
```
MichaelFu512, line 1000: I don't recall adding any of this, I wonder where it came from?

evalml/automl/automl_search.py
```diff
@@ -1006,6 +1010,14 @@ def search(self, show_iteration_plot=True):
                 f"Best pipeline {self.objective.name}: {best_pipeline['validation_score']:3f}"
             )
         self._searched = True
+        if self.search_iteration_plot is not None:
+            if self.verbose and not interactive_plot:
+                self.search_iteration_plot = self.plot.search_iteration_plot(
+                    interactive_plot=interactive_plot
+                )
+                if pio.renderers.default != "browser":
```
MichaelFu512, line 1018: This stops the automl tests from mass opening graphs in the browser, which was causing the tests to be really slow.

docs/source/demos/fraud.ipynb
```diff
@@ -281,7 +281,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
```
MichaelFu512, line 284: Not sure where these metadata changes are coming from.

docs/source/user_guide/automl.ipynb
jeremyliweishih: did you open the notebook using vscode? Not sure if `make lint-fix` covers this part of the notebook but you could always manually remove it by opening the notebook in a text editor.

evalml/automl/automl_search.py
```diff
@@ -1006,6 +1010,14 @@ def search(self, show_iteration_plot=True):
...
+                    self.search_iteration_plot.show()
+                # return self.search_iteration_plot
```
eccabay, line 1020: This comment should be removed!

evalml/automl/automl_search.py
```diff
@@ -857,15 +858,18 @@ def _handle_keyboard_interrupt(self):
...
         Args:
-            show_iteration_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
+            interactive_plot (boolean, True): Shows an iteration vs. score plot in Jupyter notebook.
                 Disabled by default in non-Jupyter enviroments.
 
         Raises:
             AutoMLSearchException: If all pipelines in the current AutoML batch produced a score of np.nan on the primary objective.
+
+        Returns:
+            A plotly figure of the completed search
```
eccabay, line 872: With using the browser check instead, we no longer need to have this in the docstring!

eccabay (approved):
Looks good, thanks for making these changes!

docs/source/user_guide/automl.ipynb
MichaelFu512: I did open the notebook using vscode. I'll remove these from now on.

chukarsten (approved):
Looks good, thanks Michael!

PR #3597, opened 2022-06-30
Add the `sp` parameter to ARIMA models

eccabay:
A bit of information on setting this parameter: http://alkaline-ml.com/pmdarima/tips_and_tricks.html#period 

I tried to offer the flexibility for the user to set it if they know it, but also add some quick default values for common time series frequencies.

Perf tests are complete.

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -27,7 +27,7 @@ class ARIMARegressor(Estimator):
         max_q (int): Maximum Moving Average order. Defaults to 5.
         seasonal (boolean): Whether to fit a seasonal model to ARIMA. Defaults to True.
         sp (int): Period for seasonal differencing, specifically the number of periods in each season. If "detect" and the time series is one of
-                  a few standard frequencies, this model will automatically detect this parameter, defaulting to 1 (no seasonality). Defaults to "detecct".
+                  a few standard frequencies, this model will automatically detect this parameter, defaulting to 1 (no seasonality). Defaults to "detect".
         n_jobs (int or None): Non-negative integer describing level of parallelism used for pipelines. Defaults to -1.
         random_seed (int): Seed for the random number generator. Defaults to 0.
     """
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -26,6 +26,8 @@ class ARIMARegressor(Estimator):
         max_d (int): Maximum Differencing degree. Defaults to 2.
         max_q (int): Maximum Moving Average order. Defaults to 5.
         seasonal (boolean): Whether to fit a seasonal model to ARIMA. Defaults to True.
+        sp (int): Period for seasonal differencing, specifically the number of periods in each season. If "detect" and the time series is one of
+                  a few standard frequencies, this model will automatically detect this parameter, defaulting to 1 (no seasonality). Defaults to "detecct".
         n_jobs (int or None): Non-negative integer describing level of parallelism used for pipelines. Defaults to -1.
         random_seed (int): Seed for the random number generator. Defaults to 0.
     """
@@ -65,13 +67,14 @@ class ARIMARegressor(Estimator):
         max_d=2,
         max_q=5,
         seasonal=True,
+        sp="detect",
         n_jobs=-1,
         random_seed=0,
         maxiter=10,
         use_covariates=True,
         **kwargs,
     ):
-        parameters = {
+        self.arima_parameters = {
             "trend": trend,
             "start_p": start_p,
             "d": d,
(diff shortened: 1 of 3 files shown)


evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -26,6 +26,8 @@ class ARIMARegressor(Estimator):
         max_d (int): Maximum Differencing degree. Defaults to 2.
         max_q (int): Maximum Moving Average order. Defaults to 5.
         seasonal (boolean): Whether to fit a seasonal model to ARIMA. Defaults to True.
+        sp (int): Period for seasonal differencing, specifically the number of periods in each season. If "detect" and the time series is one of
+                  a few standard frequencies, this model will automatically detect this parameter, defaulting to 1 (no seasonality). Defaults to "detect".
```
jeremyliweishih, line 30: I find `....defaulting to 1 (no seasonality). Defaults to "detect".` a little confusing. Maybe something along the liens of: 

`If "detect", this model will automatically detect this parameter (given the time series is a standard frequency) and will fall back to 1 (no seasonality) if it cannot be detected. Defaults to "detect".`

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -129,6 +132,23 @@ def _set_forecast(self, X):
...
+            "W": 52,
+            "M": 12,
+            "Q": 4,
+        }
+        time_index = self._parameters.get("time_index", None)
+        sp = self.arima_parameters["sp"]
+        if sp == "detect":
+            inferred_freqs = X.ww.infer_temporal_frequencies()
+            freq = inferred_freqs.get(time_index, None)
+            sp = freq_mappings.get(freq, 1)
```
jeremyliweishih, line 149: this handles the case where the freq is `None` right? Just for my own clarity 😄

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -145,6 +146,26 @@ def test_set_forecast(get_ts_X_y):
     assert fh_.is_relative
 
 
+def test_get_sp():
```
jeremyliweishih, line 149: again for my clarity - does the last case here cover the `None` case? Might be helpful to add a comment here.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -243,6 +264,25 @@ def test_fit_predict_sk_failure(
     assert y_pred.index.equals(X_test.index)
 
 
+def test_arima_sp_changes_result():
+    y = pd.Series([math.sin(i) for i in range(200)])
+
+    X = pd.DataFrame({"dates": pd.date_range("2021-01-01", periods=200, freq="D")})
```
jeremyliweishih, line 270: can we check that `sp` is set correctly for the two examples here? I.e `sp == 7` and `sp==4`?

jeremyliweishih (approved):
great work on this @eccabay! Just some additional testing suggestions but lets wait to review the perf test results before merging!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -129,6 +132,23 @@ def _set_forecast(self, X):
...
+        freq_mappings = {
+            "D": 7,
+            "W": 52,
+            "M": 12,
+            "Q": 4,
+        }
+        time_index = self._parameters.get("time_index", None)
+        sp = self.arima_parameters["sp"]
+        if sp == "detect":
+            inferred_freqs = X.ww.infer_temporal_frequencies()
```
chukarsten, line 147: So, dumb question here: if X already has its freqs inferred, will this reinfer them?  Inference of data types is expensive, I'd assume freq inference is, too.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -140,14 +160,20 @@ def fit(self, X, y=None):
...
             X = X.fillna(X.mean())
         X, y = self._manage_woodwork(X, y)
         if y is None:
             raise ValueError("ARIMA Regressor requires y as input.")
 
+        parameters = self.arima_parameters
+        parameters["sp"] = self._get_sp(X)
+
+        arima_model = self.sktime_arima.AutoARIMA(**parameters)
+        self._component_obj = arima_model
```
chukarsten, line 175: So, is the reason we deviated from letting the parent class set the `_component_obj` was because we needed to see the X data in fit and use the frequency of the data to set `sp`?  

I'm a little shaky on that approach.  After we instantiate an `ARIMARegressor`, it, unlike other regressors/estimators, will have no `_component_obj`.  Perhaps we could try keeping the passing of the model, with most of the parameters set in `__init__()` to the parent's constructor to set the `_component_obj` but then call `set_params()` [docs](https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.arima.AutoARIMA.html#pmdarima.arima.AutoARIMA.set_params) after using `.clone()` or `.get_params` on the AutoARIMA forecaster in `.fit()`

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -145,6 +146,26 @@ def test_set_forecast(get_ts_X_y):
...
+    sp_ = clf_day._get_sp(X)
+    assert sp_ == 7
+
+    X = pd.DataFrame({"dates": pd.date_range("2021-01-01", periods=500, freq="M")})
+    X.ww.init()
+    clf_month = ARIMARegressor(time_index="dates")
+    sp_ = clf_month._get_sp(X)
+    assert sp_ == 12
+
+    X = pd.DataFrame({"dates": pd.date_range("2021-01-01", periods=500, freq="2D")})
```
chukarsten, line 162: I think going back to one of my earlier comments, if X already knows the frequency, we might not want to have the freq re-infered.  Perhaps we want to add a test counting the number of times `infer_temporal_freqs()` is called in the case of a known frequency provided and limit it to 1?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
eccabay: Yep, thanks to the `get`, a frequency of None will default to 1 (no seasonality)!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
eccabay: Unfortunately, from what I can tell, neither woodwork nor pandas saves the frequency information - you can run inference in either library to get the frequency, but it's not saved anywhere between runs.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
eccabay: Gotcha - I was a little shaky as well, but I was worried setting a parameter after instantiation would affect the results. I've switched it back to instantiating the `component_obj` in init and everything still seems to work fine, so I'll leave it there.

PR #3602, opened 2022-07-05
All tests don't skip python 3.9

### LINKED ISSUE #3600: Undo Skipping of Tests for Python 3.9
In [PR #3133](https://github.com/alteryx/evalml/pull/3133), we marked tests to skip if the environment was a Python 3.9 environment.  I don't think all the tests that are being skipped need to be skipped anymore.  In working through the PolynomialDetrender tests, it was noted that Python 3.9 environments were skipping these tests, probably due to sktime not being compatible with that version of Python at the time. 

We should: 

1. Search for `pytest.mark.skip_if_39` in the evalml codebase
2. Generate an EvalML environment with Python 3.9.
3. Attempt to run any testing modules using this pytest mark.
4. Remove the mark from modules that are now successful with Python 3.9.
5. Generate some feedback as for any modules that still don't support Python 3.9.

MichaelFu512:
### Pull Request Description
Removed `pytest.mark.skip_if_39` from all tests as testing them in python 3.9, none of them failed.

Did `conda create -n evalml_py_39 python=3.9` then did `make installdeps-dev`.

After that, did a search using ctrl-f and looked for `pytest.mark.skip_if_39` from all files, removed them from the files that had them and ran each tests. None of the tests that previous were skipped failed.

Only issue was two of the tests complained about not having prophet and cmdstanpy installed, but those were easily rectified by doing `pip install prophet` and `pip install cmdstanpy`.

Closes #3600 

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,11 +3,12 @@ Release Notes
 **Future Releases**
     * Enhancements
     * Fixes
-        * Updated the `load_diabetes()` method to account for scikit-learn 1.1.1 changes to the dataset :pr:`3591`
+        * Updated the ``load_diabetes()`` method to account for scikit-learn 1.1.1 changes to the dataset :pr:`3591`
     * Changes
     * Documentation Changes
     * Testing Changes
         * Pinned GraphViz version for Windows CI Test :pr:`3596`
+        * Removed ``pytest.mark.skip_if_39`` pytest marker :pr:`3602`
 
 .. warning::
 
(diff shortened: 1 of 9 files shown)


eccabay (commented):
Thanks for tackling this, it's encouraging that no tests needed to be updated! I'm wondering if we should add prophet to the dev requirements - although at the moment I don't think it's necessary.

It looks like there are a few other files that have references to skipping with python 3.9:
- `evalml/tests/conftest.py`
- `evalml/tests/component_tests/test_utils.py`

Once those are updated as well, this should be good to go!

chukarsten (approved):
Thanks Michael!  Nice job!

PR #3607, opened 2022-07-11
Fix conda test failing

MichaelFu512:
### Pull Request Description
I changed some things in component_tests/test_utils.py and it made the conda build fail when pushed to main so hopefully this fixes it.

-----

### DIFF
--- a/evalml/tests/component_tests/test_utils.py
+++ b/evalml/tests/component_tests/test_utils.py
@@ -102,13 +102,10 @@ not_supported_in_windows = set(
         "Prophet Regressor",
     ]
 )
-not_supported_in_windows_py39 = set(
-    [
-        "Prophet Regressor",
-    ]
-)
+
 # Keeping here in case we need to add to it when a new component is added
 not_supported_in_linux_py39 = set()
+not_supported_in_windows_py39 = set()
 
 
 def test_all_components(is_using_conda, is_using_windows):
--- a/evalml/tests/component_tests/test_utils.py
+++ b/evalml/tests/component_tests/test_utils.py
@@ -92,50 +92,14 @@ all_requirements_set = set(
         "XGBoost Regressor",
     ]
 )
-not_supported_in_conda = set(
-    [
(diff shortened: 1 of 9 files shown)


evalml/tests/component_tests/test_utils.py
```diff
@@ -92,15 +92,51 @@
...
+def test_all_components(
+    is_running_py_39_or_above,
+    is_using_conda,
+    is_using_windows,
+):
+    if is_using_conda:
+        # No prophet, ARIMA, and vowpalwabbit
+        expected_components = all_requirements_set.difference(not_supported_in_conda)
+
+    elif is_using_windows and not is_running_py_39_or_above:
```
jeremyliweishih, line 124: if the issue was conda - can we test dropping these windows + py 39 cases?

MichaelFu512: Probably! I did it earlier but wasn't sure if that was the right approach. I'll change it.

jeremyliweishih: if the purpose of your previous PR was to increase coverage for py39 (and these test cases still pass under those scenarios) I would say to drop these cases but keep the conda case because that is still needed. Thank you!

evalml/tests/conftest.py
```diff
@@ -55,6 +55,10 @@ def pytest_configure(config):
         "markers",
         "skip_during_conda: mark test to be skipped if running during conda build",
     )
+    config.addinivalue_line(
```
jeremyliweishih, line 58: do we need this still?

jeremyliweishih (approved):
LGTM, thanks for the quick fix! @chukarsten will need you to merge due to coverage.

evalml/tests/conftest.py
MichaelFu512: oops I should've caught that, thanks!

eccabay (approved):
Thanks for jumping on this!

PR #3608, opened 2022-07-12
Add pre-commit hooks for linting 

gsheni:
- This makes it so that linting is automatically run right before you commit code:
- We use this https://pre-commit.com/
- Example of this:
Screen Shot 2022-07-11 at 5 30 10 PM
- More linting checks were added with the pre-commit-config
  - Strip out Jupyter notebook outputs
  - End of file fixer
  - removing trailing whitespaces
  - absolute important everywhere for EvalML
  - Add trailing commas
  - Run black[juypter] on py and jupyter files


pyproject.toml
```diff
@@ -3,7 +3,27 @@ requires = [
     "setuptools >= 47",
     "wheel"
 ]
-build-backend = "setuptools.build_meta"
```
gsheni, line 6: @jeremyliweishih here is the moved linting standards.

jeremyliweishih (approved):
awesome - thanks so much for doing this!

eccabay (approved):

PR #3630, opened 2022-07-27
Updated data check code for regression

simha104:
### Updated data check code for regression
This is to avoid duplicate errors produced by non numerical regression and unsupported data types

### DIFF
--- a/evalml/data_checks/data_check_message_code.py
+++ b/evalml/data_checks/data_check_message_code.py
@@ -33,7 +33,7 @@ class DataCheckMessageCode(Enum):
     """Message code for target data that is of an unsupported type."""
 
     TARGET_UNSUPPORTED_TYPE_REGRESSION = "target_unsupported_type_regression"
-    """Message code for target data that is incompatible for regression"""
+    """Message code for target data that is incompatible with regression"""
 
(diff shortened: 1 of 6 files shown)


evalml/data_checks/data_check_message_code.py
```diff
@@ -32,6 +32,9 @@ class DataCheckMessageCode(Enum):
     TARGET_UNSUPPORTED_TYPE = "target_unsupported_type"
     """Message code for target data that is of an unsupported type."""
 
+    TARGET_REGRESSION_NON_NUMERIC = "target_regression_non_numeric"
```
chukarsten, line 35: Maybe we just rename this to `TARGET_UNSUPPORTED_TYPE_REGRESSION`?  As named, it doesn't seem to get the point across to me.  Also a nit: "incompatible with regression"

chukarsten (changes requested):
Thanks for making the adjustment!  I think we should still adjust the name, though.

bchen1116 (approved):
Agreed with Karsten here, but looks good otherwise.

chukarsten (approved):
Thanks a bunch!

PR #3631, opened 2022-07-28
Add option to AutoMLSearch to exclude featurizers from pipelines

### LINKED ISSUE #3590: Implement an option to prevent feature generation on time index column for time series problems
If a user has previously used Featuretools to generate feature values for a time series problem, including features created from the time index column (year, month, day, etc), EvalML should provide a mechanism to prevent additional datetime features from being generated by the `DatetimeFeaturizer` component.

Since the time index is required as part of the data input, if EvalML uses this time index to create new features, there can be duplication of features, with the some of the same features generated by both Featuretools and EvalML. There should be some mechanism for users to instruct EvalML to skip generating new features on datetime columns in this scenario.

### LINKED ISSUE #3619: Allow users to optionally omit TimeSeriesFeaturizer from pipeline for time series problems
In some applications, users may have already generated time series features and included them in the input data to EvalML. In those scenarios, users want to skip having EvalML generate additional time series features from the input data. An update should be made to allow users to optionally omit the `TimeSeriesFeaturizer` component from the pipeline for cases where the input data already contains the desired features.

thehomebrewnerd:
### Add option to AutoMLSearch to exclude featurizers from pipelines

Adds an optional parameter to `AutoMLSearch` to allow users to optionally exclude the EvalML `*Featurizer` components from pipelines in situations where feature engineering has been performed ahead of executing the search.

Closes #3619 
Closes #3590

### DIFF
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -429,7 +429,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             )
         if (
             list(self.X.ww.select("URL", return_schema=True).columns)
-            and not "URLFeaturizer" in self.exclude_featurizers
+            and "URLFeaturizer" not in self.exclude_featurizers
         ):
             self._get_feature_provenance_and_remove_engineered_features(
                 pipeline,
@@ -439,7 +439,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             )
         if (
             list(self.X.ww.select("EmailAddress", return_schema=True).columns)
-            and not "EmailFeaturizer" in self.exclude_featurizers
+            and "EmailFeaturizer" not in self.exclude_featurizers
         ):
             self._get_feature_provenance_and_remove_engineered_features(
                 pipeline,
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -117,7 +117,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
         self._X_without_cat_cols = None
         self.features = features
         self.ensembling = ensembling
-        self.exclude_featurizers = exclude_featurizers
+        self.exclude_featurizers = exclude_featurizers or []
 
         # TODO remove on resolution of 3186
         if is_time_series(self.problem_type) and self.ensembling:
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -66,7 +66,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             AutoMLSearch will not use Elastic Net or XGBoost when there are more than 75 multiclass targets and will not use CatBoost when there are more than 150 multiclass targets. Defaults to False.
         features (list)[FeatureBase]: List of features to run DFS on in AutoML pipelines. Defaults to None. Features will only be computed if the columns used by the feature exist in the input and if the feature has not been computed yet.
         verbose (boolean): Whether or not to display logging information regarding pipeline building. Defaults to False.
-        exclude_featurizers (list -> str): A list of featurizer components to exclude from the components.
+        exclude_featurizers (list[str]): A list of featurizer components to exclude from the components.
             Valid options are "DatetimeFeaturizer", "EmailFeaturizer", "URLFeaturizer", "NaturalLanguageFeaturizer", "TimeSeriesFeaturizer"
     """
 
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -66,7 +66,7 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             AutoMLSearch will not use Elastic Net or XGBoost when there are more than 75 multiclass targets and will not use CatBoost when there are more than 150 multiclass targets. Defaults to False.
         features (list)[FeatureBase]: List of features to run DFS on in AutoML pipelines. Defaults to None. Features will only be computed if the columns used by the feature exist in the input and if the feature has not been computed yet.
         verbose (boolean): Whether or not to display logging information regarding pipeline building. Defaults to False.
-        exclude_featurizers (list[str]): A list of featurizer components to exclude from the components.
+        exclude_featurizers (list[str]): A list of featurizer components to exclude from the pipelines built by DefaultAlgorithm.
             Valid options are "DatetimeFeaturizer", "EmailFeaturizer", "URLFeaturizer", "NaturalLanguageFeaturizer", "TimeSeriesFeaturizer"
     """
 
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -427,14 +427,20 @@ class DefaultAlgorithm(AutoMLAlgorithm):
                 self._selected_cols,
                 self._selected_cat_cols,
             )
-        if list(self.X.ww.select("URL", return_schema=True).columns):
+        if (
+            list(self.X.ww.select("URL", return_schema=True).columns)
+            and not "URLFeaturizer" in self.exclude_featurizers
+        ):
             self._get_feature_provenance_and_remove_engineered_features(
                 pipeline,
                 URLFeaturizer.name,
                 self._selected_cat_cols,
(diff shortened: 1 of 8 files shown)


evalml/tests/automl_tests/test_automl.py
```diff
@@ -5047,6 +5054,60 @@ def test_default_algorithm_uses_n_jobs(X_y_binary, AutoMLTestEnv):
     assert n_checked and n_feature_selector_checked
 
 
+@pytest.mark.parametrize("input_type", ["pd", "ww"])
+@pytest.mark.parametrize("automl_algorithm", ["default", "iterative"])
+@pytest.mark.parametrize("problem_type", ProblemTypes.all_problem_types)
+def test_exclude_featurizers(
```
jeremyliweishih, line 5060: could you repurpose this test to run search and then check the pipelines generated by search with `get_pipeline()`? I think this current test would be better suited in both `test_default_algorithm` and `test_iterative_algorithm` separately where we check pipeline generation with each algorithm.  I general I try to separate testing top level search behavior in this file and then algorithm specific pipeline generation logic in each algorithm's test file. You can use our `AutoMLTestEnv` fixture to skip computation here. Here's a good [example](https://github.com/alteryx/evalml/blob/main/evalml/tests/automl_tests/test_automl.py#L235-L237).

jeremyliweishih (changes requested):
This looks great to me but just have some docstring suggestions and a testing suggestion!

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -66,6 +66,8 @@ class DefaultAlgorithm(AutoMLAlgorithm):
             AutoMLSearch will not use Elastic Net or XGBoost when there are more than 75 multiclass targets and will not use CatBoost when there are more than 150 multiclass targets. Defaults to False.
         features (list)[FeatureBase]: List of features to run DFS on in AutoML pipelines. Defaults to None. Features will only be computed if the columns used by the feature exist in the input and if the feature has not been computed yet.
         verbose (boolean): Whether or not to display logging information regarding pipeline building. Defaults to False.
+        exclude_featurizers (list[str]): A list of featurizer components to exclude from the components.
```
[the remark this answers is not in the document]
thehomebrewnerd: Yeah, not sure what I was thinking when I wrote the original. I'll fix!

evalml/tests/automl_tests/test_automl.py
thehomebrewnerd: @jeremyliweishih Just to be clear on what you are proposing.

You think I should split the existing test into two, and move the first case for default algorithm into `test_default_algorithm.py` and the second iterative case into `test_iterative_algorithm.py`. Then, create a new test in `test_automl.py` that uses the `AutoMLTestEnv` fixture like the example you linked? Is that correct?

jeremyliweishih: @thehomebrewnerd yes! how does that sound?

thehomebrewnerd: Works for me! I'll work on that update this afternoon.

thehomebrewnerd: @jeremyliweishih I'm having a little trouble with using `AutoMLTestEnv`, so might need some pointers next week on how to resolve. This commit contains my attempt at updating the test in `test_automl.py`: [c9c3377](https://github.com/alteryx/evalml/pull/3631/commits/c9c3377dfe15103c1e722cbb8f6dfabb7d4a674d)

Is that what you were envisioning?

jeremyliweishih: yup! it looks basically there. Please ping me next week!

thehomebrewnerd: Thanks - will do. This is the error I am getting, but haven't really dug too deep to understand what is happening.

```
evalml/automl/automl_search.py:1064: in search
    pipeline_id = self._post_evaluation_callback(
evalml/automl/automl_search.py:1388: in _post_evaluation_callback
    self.automl_algorithm.add_result(
evalml/automl/automl_algorithm/default_algorithm.py:485: in add_result
    self._parse_selected_categorical_features(pipeline)
evalml/automl/automl_algorithm/default_algorithm.py:431: in _parse_selected_categorical_features
    self._get_feature_provenance_and_remove_engineered_features(
evalml/automl/automl_algorithm/default_algorithm.py:411: in _get_feature_provenance_and_remove_engineered_features
    component = pipeline.get_component(component_name)
evalml/pipelines/pipeline_base.py:224: in get_component
    return self.component_graph.get_component(name)

FAILED evalml/tests/automl_tests/test_automl.py::test_exclude_featurizers[binary-default-pd] - ValueError: Component URL Featurizer is not in the graph
```

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5047,6 +5054,62 @@ def test_default_algorithm_uses_n_jobs(X_y_binary, AutoMLTestEnv):
...
+        automl.search()
+
+    pipelines = [
+        automl.get_pipeline(i) for i in range(len(automl.results["pipeline_results"]))
+    ]
+
+    # A check to make sure we actually retrieve constructed pipelines from the algo.
+    assert len(pipelines) > 0
+
+    assert not any([DateTimeFeaturizer.name in pl for pl in pipelines])
```
jeremyliweishih, line 5106: `get_pipeline()` returns a pipeline instance so you would need to check for `DateTimeFeaturizer.name in pl.component_graph.compute_order for pl in pipelines`. You could also do `DateTimeFeaturizer.name in pl.name for pl in pipelines` but the compute order check is more explicit as `compute_order` is the list of components in text.

jeremyliweishih (changes requested):
This looks basically there! Just one small thing to fix re checking if a component is in a pipeline.

evalml/tests/automl_tests/test_automl.py
thehomebrewnerd: Thanks for the clarification.

Updated: [f111676](https://github.com/alteryx/evalml/pull/3631/commits/f111676fb3aae55b9fb37fde1a05660a22ba8e3d)

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -922,3 +927,52 @@ def test_default_algorithm_accepts_URL_email_features(
...
+    )
+
+    pipelines = []
+    for _ in range(4):
+        pipelines.extend([pl.name for pl in algo.next_batch()])
+
+    # A check to make sure we actually retrieve constructed pipelines from the algo.
+    assert len(pipelines) > 0
+
+    assert not any([DateTimeFeaturizer.name in pl for pl in pipelines])
```
jeremyliweishih, line 974: could you repeat the changes you've made to `test_exclude_featurizers` here and in `test_exclude_featurizers_iterative_algorithm` as well? Thanks!

jeremyliweishih (approved):
This LGTM! Great work and this brings our two libraries even closer 😄 just left a comment about updating the algo tests with the same component checking logic!

chukarsten (approved):
This looks good @thehomebrewnerd .  I think the only thing to perhaps consider is : what happens if the user inputs "TmeSeriesFeaturizer" expecting, but not checking for the featurizer to be removed?  I get that this is a flag added mostly for internal stakeholder use, but it might be nice to add a quick name check by pulling all the featurizer components and raising a ValueError if the provided name is not contained.

evalml/tests/automl_tests/test_default_algorithm.py
thehomebrewnerd: Updated here: [f115377](https://github.com/alteryx/evalml/pull/3631/commits/f115377ae5e23dbe4789bc33f5e03456c5ae6f27)

thehomebrewnerd:
> This looks good @thehomebrewnerd . I think the only thing to perhaps consider is : what happens if the user inputs "TmeSeriesFeaturizer" expecting, but not checking for the featurizer to be removed? I get that this is a flag added mostly for internal stakeholder use, but it might be nice to add a quick name check by pulling all the featurizer components and raising a ValueError if the provided name is not contained.

I added a check for this in AutoMLSearch: [9d79cf5](https://github.com/alteryx/evalml/pull/3631/commits/9d79cf574775287f06ca5ee08976b60570135fec)

thehomebrewnerd:
@chukarsten @jeremyliweishih Added the check for invalid parameter inputs as requested, just wanted to confirm you are good with that change (and the new test) before I merge.

jeremyliweishih (approved):
LGTM - thanks!

PR #3632, opened 2022-07-29
Add `Progress` abstraction

jeremyliweishih:


evalml/automl/progress.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _build_progress_dict(self, stopping_criteria, current_state, end_state, unit):
+        progress_dict = {}
+        progress_dict["stopping_criteria"] = stopping_criteria
+        progress_dict["current_state"] = current_state
+        progress_dict["end_state"] = end_state
+        progress_dict["unit"] = unit
+        return progress_dict
+
+    def return_progress(self):
```
jeremyliweishih, line 128: most of this PR is refactoring to use `should_continue` in this class but this is the important method. Key questions being: is this sufficient information and is the formatting easy to work with to gather information on search progress?

evalml/automl/progress.py
```diff
@@ -0,0 +1,172 @@
...
+        self.tolerance = tolerance
+        self.automl_algorithm = automl_algorithm
+        self.objective = objective
+
+        if verbose:
+            self.logger = get_logger(f"{__name__}.verbose")
+        else:
+            self.logger = logging.getLogger(__name__)
+
+    def should_continue(self, results, interrupted=False, mid_batch=False):
```
jeremyliweishih, line 58: this `mid_batch` flag is actually quite important. If you look at `search()` we call `should_continue` multiple times. In the past, `should_continue` would return false only if the current batch number was greater than `max_batches`. With this flag, we can properly return false when we hit the right batch number.

evalml/tests/automl_tests/test_automl_iterative_algorithm.py
```diff
@@ -415,9 +415,7 @@ def test_max_batches_num_pipelines(
         # automl_algorithm will include all allowed_pipelines in the first batch even
         # if they are not searched over. That is why n_automl_pipelines does not equal
         # n_results when max_iterations and max_batches are None
-        n_automl_pipelines = (
-            1 + len(automl.allowed_pipelines) + automl._pipelines_per_batch
-        )
+        n_automl_pipelines = 1 + len(automl.allowed_pipelines)
```
jeremyliweishih, line 418: first batch should only be baseline + a full round of possible estimators. What we had previously was erroneous due to the `should_continue()` and `max_batches` logic I mentioned above.

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1267,16 +1256,19 @@ def test_automl_passes_allow_long_running_models(
...
-        max_batches=2,
+        max_batches=3,
         verbose=True,
     )
     assert (
         automl.automl_algorithm.allow_long_running_models == allow_long_running_models
     )
     if algo == "default":
-        automl.search()
+        env = AutoMLTestEnv("multiclass")
```
jeremyliweishih, line 1269: Just used this to speed up this test.

evalml/tests/automl_tests/test_progress.py
```diff
@@ -0,0 +1,69 @@
...
+    for id in mock_results["search_order"]:
+        mock_results["pipeline_results"][id] = {}
+        mock_results["pipeline_results"][id]["mean_cv_score"] = scores[id]
+        mock_results["pipeline_results"][id][
+            "pipeline_class"
+        ] = logistic_regression_binary_pipeline.__class__
+
+    assert p.should_continue(mock_results)
+
+    progress_dict = p.return_progress()
```
jeremyliweishih, line 57: this test is mainly just checking if the stopping criteria is updated correctly in the right form. I thought this was sufficient but lmk if you have any more ideas how to properly test this!

evalml/automl/progress.py
```diff
@@ -0,0 +1,172 @@
...
+            max_iterations = (
+                self.max_iterations
+                if self.max_iterations
+                else sum(
+                    [
+                        self.automl_algorithm.num_pipelines_per_batch(n)
+                        for n in range(self.max_batches)
+                    ],
+                )
+            )
```
eccabay, line 154: I'm a little confused by the logic here - if `max_batches` is the stopping criteria, we add stopping conditions for both the number of batches and the number of iterations, but we calculate the number of iterations based on the number of batches. Isn't that redundant/ is there a reason we need both in this case?

I see in the tests that you have an example of default algorithm setting both `max_iterations` and `max_batches`, but is that possible when using AutoMLSearch outside of testing?

evalml/tests/automl_tests/test_automl_search_classification.py
```diff
@@ -1267,16 +1256,19 @@ def test_automl_passes_allow_long_running_models(
         X_train=X,
         y_train=y,
         problem_type="multiclass",
+        objective="Log Loss Multiclass",
         allow_long_running_models=allow_long_running_models,
         automl_algorithm=algo,
-        max_batches=2,
+        max_batches=3,
```
eccabay, line 1262: What was the reason for this change?

evalml/tests/automl_tests/test_default_algorithm.py
```diff
@@ -929,6 +929,46 @@ def test_default_algorithm_accepts_URL_email_features(
...
+        problem_type=automl_type,
+        sampler_name=None,
+        ensembling=ensembling,
+        top_n=2,
+        num_long_explore_pipelines=10,
+        num_long_pipelines_per_batch=5,
+    )
+    for i in range(7):
+        batch = algo.next_batch()
+        add_result(algo, batch)
```
eccabay, line 968: Why is the `add_result` call necessary here?

eccabay (commented):
This is super interesting! Looks pretty good to me, just left some impl notes and clarification questions. I'm holding off on approving for now so I can come back and re-review the tests more thoroughly.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -142,6 +142,51 @@ def default_max_batches(self):
...
+        """
+        if self.ensembling:
+            if batch_number == 0 or batch_number == 1:
+                return len(self._naive_estimators())
+            elif batch_number == 2:
+                return len(
+                    [
+                        estimator
+                        for estimator in get_estimators(self.problem_type)
+                        if estimator not in self._naive_estimators()
```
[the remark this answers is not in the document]
jeremyliweishih: good call! I'll update the other similar calls.

evalml/automl/automl_search.py
```diff
@@ -1060,7 +1076,14 @@ def search(self, interactive_plot=True):
                         computations.append((computation, False))
                     current_computation_index = 0
                     computations_left_to_process = len(computations)
-                while self._should_continue() and computations_left_to_process > 0:
+                while (
+                    self.progress.should_continue(
+                        self._results,
+                        self._interrupted,
+                        True,
```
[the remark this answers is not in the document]
jeremyliweishih: that I can do!

evalml/automl/progress.py
```diff
@@ -0,0 +1,172 @@
...
+            return False
+
+        # check for early stopping
+        if self.patience is None or self.tolerance is None:
+            return True
+
+        first_id = results["search_order"][0]
+        best_score = results["pipeline_results"][first_id]["mean_cv_score"]
+        num_without_improvement = 0
+        for id in results["search_order"][1:]:
```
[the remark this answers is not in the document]
jeremyliweishih: great idea!

evalml/automl/progress.py
```diff
@@ -0,0 +1,172 @@
...
+                )
+                return False
+        return True
+
+    def _build_progress_dict(self, stopping_criteria, current_state, end_state, unit):
+        progress_dict = {}
+        progress_dict["stopping_criteria"] = stopping_criteria
+        progress_dict["current_state"] = current_state
+        progress_dict["end_state"] = end_state
+        progress_dict["unit"] = unit
```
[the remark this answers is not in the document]
jeremyliweishih: I originally thought that it would read better but I agree with your points here 😄

evalml/automl/progress.py
jeremyliweishih: I don't believe we add `max_iterations` as stopping criteria in search (I might be confusing your point here). The reason for this conversion is to allow users to ping the number of pipelines progress/left if they only select `max_batches` as stopping criteria. I can see how that is confusing though. I wanted to keep the structure the same if a user did select both `max_iterations` and `max_batches` (`max_iterations` would take precedence) but maybe I could add the pipeline information to the `max_batches` dictionary instead?

evalml/tests/automl_tests/test_automl_search_classification.py
jeremyliweishih: `DefaultAlgorithm` doesn't call `filter_estimators()` and drops the long running estimators until the final non-ensembling batch (batch 2 if starting from batch 0) so this previous test case was not actually covering the `DefaultAlgorithm` case. IIRC it was obfuscated by the patches in the previous implementation of this test. With `AutoMLTestEnv` adding a batch shouldn't cause much overhead if we add a batch as well!

evalml/tests/automl_tests/test_default_algorithm.py
jeremyliweishih: we save the feature selector information in `add_result` so this allows us to mock out `@patch("evalml.pipelines.components.FeatureSelector.get_names")`

evalml/tests/automl_tests/test_progress.py
```diff
@@ -0,0 +1,135 @@
...
+    assert p.automl_algorithm == DefaultAlgorithm
+    assert p.max_batches == 3
+    assert p.current_batches == 0
+    assert p.max_iterations == 2
+    assert p.current_iterations == 0
+    assert p.max_time == 1000
+    assert p.current_time is None
+
+
+def test_progress_should_continue(X_y_binary, logistic_regression_binary_pipeline):
```
chukarsten, line 33: Is it possible to refactor some of this repeated code into a parametrized test? I like tests parametrized with `@pytest.mark.parametrize("test_type", ["descriptive title 1", "descriptive title 2"....]` with the stuff that's changing in the test switched using if/else.

evalml/tests/automl_tests/test_progress.py
```diff
@@ -0,0 +1,135 @@
...
+    for id in mock_results["search_order"]:
+        mock_results["pipeline_results"][id] = {}
+        mock_results["pipeline_results"][id]["mean_cv_score"] = scores[id]
+        mock_results["pipeline_results"][id][
+            "pipeline_class"
+        ] = logistic_regression_binary_pipeline.__class__
+
+    p.should_continue(mock_results)
+    progress_dict = p.return_progress()
+    print(progress_dict)
```
chukarsten, line 123: Should probably nuke the print

chukarsten (approved):
Just a couple of mostly non-blocking suggestions (should certainly kill the print statement!). Sorry this took so long.

evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -142,6 +142,36 @@ def default_max_batches(self):
...
+                return self.num_long_pipelines_per_batch * self.top_n
+        else:
+            if batch_number == 0 or batch_number == 1:
+                return len(self._naive_estimators())
+            elif batch_number == 2:
+                return len(self._non_naive_estimators())
+            elif batch_number == 3:
+                return self.num_long_explore_pipelines * self.top_n
+            else:
+                return self.num_long_pipelines_per_batch * self.top_n
```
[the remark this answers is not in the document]
jeremyliweishih: thank you!

evalml/automl/progress.py
christopherbunn: Unless I'm missing something more specific, I think we should return progress as a single dictionary rather than as a list of dictionaries. This is mainly due to the large amount of overlap exists between these conditions. We see this with the max_batches and max_iterations logic below. Another example is that I would like to know the elasped time regardless of the stopping criteria.

As an example, I think this function should return the below:
```
{
   "stopping_criteria": "max_batches"
   "elasped_time": self.elasped(),
   "elasped_time_unit": "seconds"
   "max_time": self.max_time, # So None in this case
    "max_batches": self.max_batches, 
   ...
}
```

christopherbunn (changes requested):
Reviewed the implementation; will come back to the tests once comments are addressed!

christopherbunn (approved):

PR #3633, opened 2022-08-01
exclude stackoverflow from broken link check

fjlanasa:
### Pull Request Description
We are receiving a 403 status code when following the link to https://stackoverflow.com/questions/tagged/evalml so just going to exclude that domain.

https://github.com/alteryx/evalml/runs/7626204325?check_suite_focus=true

-----


eccabay (approved):
🚢

PR #3640, opened 2022-08-04
Add Fit_transform to pipeline_base and component_graph

### LINKED ISSUE #2665: Add `fit_transform()` to pipelines and component graphs
With https://github.com/alteryx/evalml/pull/2625 merged, we can also add a `fit_transform()` method for pipelines / component graphs--probably sufficient to just call .fit().transform()!

deanpatel2:
### Pull Request Description

Fixes #2665

-----

### DIFF
--- a/evalml/tests/pipeline_tests/test_component_graph.py
+++ b/evalml/tests/pipeline_tests/test_component_graph.py
@@ -490,6 +490,47 @@ def test_fit_correct_inputs(
     assert_frame_equal(expected_x, mock_ohe_fit_transform.call_args[0][0])
     assert_series_equal(expected_y, mock_ohe_fit_transform.call_args[0][1])
 
+@patch("evalml.pipelines.components.Transformer.fit_transform", autospec=True)
+@patch("evalml.pipelines.components.OneHotEncoder.transform", autospec=True)
+@patch("evalml.pipelines.components.Imputer.transform", autospec=True)
+@patch("evalml.pipelines.components.LabelEncoder.transform", autospec=True)
+def test_component_graph_fit_transform(
+    mock_label_encoder_transform,
+    mock_inputer_transform,
+    mock_ohe_transform,
+    mock_fit_transform,
+    example_graph,
+    example_graph_with_transformer_last_component,
+    X_y_binary,
+):
+    X, y = X_y_binary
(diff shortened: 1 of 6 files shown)


evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -491,6 +491,51 @@ def test_fit_correct_inputs(
...
+    mock_imputer_transform.return_value = ones_df
+    mock_ohe_transform.return_value = ones_df
+
+    component_graph.fit_transform(X, y)
+
+    assert mock_label_encoder_transform.call_count == 2
+    assert mock_imputer_transform.call_count == 1
+    assert mock_ohe_transform.call_count == 1
+    assert mock_fit_transform.call_count == 2
+
```
christopherbunn, line 537: Did you ever figure out why we had to mock the encoder call twice and the other transformer components once? Not blocking but just curious to see if you found out.

christopherbunn (approved):
Excellent, thanks for doing this!

evalml/tests/pipeline_tests/test_component_graph.py
deanpatel2: haven't got a chance to dig deeper there unfortunately

chukarsten: Would definitely like to know the answer here before moving on!

chukarsten (changes requested):
Think I would just like an answer to Chris' question about the multiple calls being required.

evalml/tests/pipeline_tests/test_component_graph.py
deanpatel2: Found out the `LabelEncoder` class overrides its `fit_transform` method with `self.fit().transform()`, so when the LabelEncoder transform mock is called, it actually triggers 2 calls to its `transform()` unlike the OHE and Imputer

deanpatel2:
Note: Found out the LabelEncoder class overrides its `fit_transform` method with its own `self.fit().transform()`. Thus, when its mock is called, it actually triggers 2 calls to its `transform()` unlike the OHE and Imputer

jeremyliweishih (approved):
LGTM! great work @deanpatel2

PR #3650, opened 2022-08-09
Update split_data with new default for timeseries

eccabay:
Reduces the default `test_size` in `split_data` to be 0.1 instead of 0.2, as long as the new value is greater than the passed in forecast horizon.

### DIFF
--- a/evalml/tests/preprocessing_tests/test_split_data.py
+++ b/evalml/tests/preprocessing_tests/test_split_data.py
@@ -63,28 +63,17 @@ def test_split_data(
 
 @pytest.mark.parametrize("problem_type", ProblemTypes.all_problem_types)
 @pytest.mark.parametrize("data_type", ["np", "pd", "ww"])
-def test_split_data_defaults(
-    problem_type,
-    data_type,
-    X_y_binary,
-    X_y_multi,
-    X_y_regression,
-    make_data_type,
-):
-    if is_binary(problem_type):
-        X, y = X_y_binary
-    if is_multiclass(problem_type):
-        X, y = X_y_multi
-    if is_regression(problem_type):
-        X, y = X_y_regression
+def test_split_data_defaults(problem_type, data_type, get_test_data_from_configuration):
+    X, y = get_test_data_from_configuration(
+        data_type,
+        problem_type,
+        column_names=["numerical"],
+        scale=10,
+    )
+
     problem_configuration = None
     if is_time_series(problem_type):
         problem_configuration = {"gap": 1, "max_delay": 7, "time_index": "ts_data"}
-
-    X = make_data_type(data_type, X)
-    y = make_data_type(data_type, y)
-
-    if is_time_series(problem_type):
         test_pct = 0.1
     else:
         test_pct = 0.2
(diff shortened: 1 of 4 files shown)


fjlanasa (approved):

evalml/tests/preprocessing_tests/test_split_data.py
```diff
@@ -59,3 +59,97 @@ def test_split_data(
...
+    if is_multiclass(problem_type):
+        X, y = X_y_multi
+    if is_regression(problem_type):
+        X, y = X_y_regression
+    problem_configuration = None
+    if is_time_series(problem_type):
+        problem_configuration = {"gap": 1, "max_delay": 7, "time_index": "ts_data"}
+
+    X = make_data_type(data_type, X)
+    y = make_data_type(data_type, y)
```
chukarsten, line 85: I suspect this is boiler plate not just for these two tests but also across many of the tests.  Did you want to refactor it into something that returns `problem_configuration`, `X`, and `y` given `data_type`, `problem_type`, etc?  I am pretty sure I've seen this a few other places too.  I think it's also fine to file an issue to call out the refactor with the modules we should refactor this code into.

evalml/tests/preprocessing_tests/test_split_data.py
```diff
@@ -59,3 +59,97 @@ def test_split_data(
...
+        y,
+        problem_type="time series regression",
+        problem_configuration=problem_configuration,
+    )
+    test_size = fh
+    train_size = len(X) - fh
+    assert len(X_train) == train_size
+    assert len(X_test) == test_size
+    assert len(y_train) == train_size
+    assert len(y_test) == test_size
```
chukarsten, line 155:
```suggestion
@pytest.mark.parametrize("test", ["fh_limitation", "no_fh_limitation"])
def test_split_data_ts(X_y_regression, test):
    X, y = X_y_regression

    if test == "no_fh_limitation":
        test_pct = 0.1
        fh = 5
        test_size = len(X) * test_pct
        train_size = len(X) - test_size
	elif test == "fh_limitation":
		fh = 25    
		test_size = fh
        train_size = len(X) - fh
    problem_configuration = {
        "gap": 1,
        "max_delay": 7,
        "forecast_horizon": fh,
        "time_index": "ts_data",
    }
    X_train, X_test, y_train, y_test = split_data(
        X,
        y,
        problem_type="time series regression",
        problem_configuration=problem_configuration,
    )
    assert len(X_train) == train_size
    assert len(X_test) == test_size
    assert len(y_train) == train_size
    assert len(y_test) == test_size
```

chukarsten: I'm a little bigger of a fan using parametrization to drop in a nice string that replaces your in-line comments, so when the test case fails, it leaves a nice message.  It also allows both test cases to run if the first half test case fails!  I am not sure the suggestion runs though, my IDE was acting weird.

chukarsten (approved):
LGTM, just added a few suggestions, take them or leave them.  Nothing blocking!

evalml/tests/preprocessing_tests/test_split_data.py
eccabay: I like this. A slight refactor to `get_test_data_from_configuration` works wonderfully to factor this out! We can go through and replace other instances in a future story.

PR #3657, opened 2022-08-11
Conditionally include the imputer in pipelines

### LINKED ISSUE #3603: Conditionally include the imputer in pipelines
Right now, we add the imputer to every pipeline that includes datatypes that the imputer supports. However, the imputer is only useful when there is missing data for the component to impute. On first sight this doesn't seem to be a huge issue, but recent profiling has highlighted this as a performance issue. Consider this snakeviz profile, where the input data has no missing values:
Screen Shot 2022-07-07 at 11 39 25 AM
Here, despite having no need for the imputer, the one component takes up a full quarter of the time it takes to call predict on the pipeline.

It is possible that there may be null values in the holdout data, which `make_pipeline` doesn't see, in which case we may want the imputer regardless. We should discuss this, but my initial thought is that it's too much of an edge case for us to take the performance hit for.

eccabay:
Closes #3603


jeremyliweishih (approved):
LGTM - great work!

chukarsten (approved):
Great to see a performance improvement for a change!  But the pessimist in me filed https://github.com/alteryx/evalml/issues/3668 to investigate whether this is reverting a previous regression.

PR #3659, opened 2022-08-11
Disable holdout set in AutoML search by default

christopherbunn:
Unblocking next EvalML release, will need to renable once we investigate why it is erroring out with the oversampler.

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -474,7 +474,7 @@ class AutoMLSearch:
         verbose=False,
         timing=False,
         exclude_featurizers=None,
-        holdout_set_size=0.1,
+        holdout_set_size=0,
     ):
         self.verbose = verbose
         if verbose:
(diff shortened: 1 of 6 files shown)


evalml/automl/automl_search.py
```diff
@@ -1724,6 +1724,7 @@ def train_pipelines(self, pipelines):
             if computation.done():
                 try:
                     fitted_pipeline = computation.get_result()[0]
+                    self.logger.info(fitted_pipeline.name)
```
jeremyliweishih, line 1727: Let's remove this

chukarsten (approved):

jeremyliweishih (approved):
LGTM once tests pass. We should cleanup that debugging log though. @christopherbunn just to make sure - is this the only logic we need to change? It looks like it to me. Let's check the docs as well to see if our examples still make sense with this change.

christopherbunn:
@jeremyliweishih I'll update the docs to reflect our new default behavior

christopherbunn:
https://github.com/alteryx/evalml/pull/3659/commits/a652d2dd006748e1bcd6f35fb0b410ea395f6e6e has the tests and docs prior to the adjustments made to pass the tests

PR #3662, opened 2022-08-12
Knn imputer implementation

MichaelFu512:
### Pull Request Description

This is the KNNImputer. I've added it to the hyper-parameter range in `Imputer.py`.
-----


evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -75,16 +75,28 @@ def __init__(
...
-            **kwargs,
-        )
-        self._numeric_imputer = SimpleImputer(
-            impute_strategy=numeric_impute_strategy,
-            fill_value=numeric_fill_value,
-            **kwargs,
-        )
+        if boolean_impute_strategy == "knn":
+            self._boolean_imputer = KNNImputer(
+                number_neighbors=1,
```
jeremyliweishih, line 80: why is `number_neighbors=1` here but `number_neighbors=3` for numeric columns?

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -0,0 +1,158 @@
...
+            X = X.ww.copy()
+            X = X.ww.drop(columns=natural_language_columns)
+        return X, natural_language_columns
+
+    def _set_boolean_columns_to_categorical(self, X):
+        boolean_columns = list(
+            X.ww.select(["Boolean"], return_schema=True).columns.keys(),
+        )
+        if boolean_columns:
+            X = X.ww.copy()
```
jeremyliweishih, line 53: why do we need this copy here?

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -0,0 +1,158 @@
...
+            "boolean",
+            "categorical",
+        } and not all(
+            [
+                is_categorical_actually_boolean(X, col)
+                for col in X.ww.select("Categorical")
+            ],
+        ):
+            raise ValueError(
+                "KNNImputer cannot handle dataframes with both boolean and categorical features.  Use Imputer instead.",
```
jeremyliweishih, line 83: so can `KNNImputer` handle the case where there are only boolean and only categorical features? Why is this the case if we set boolean columns to categorical below?

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -0,0 +1,158 @@
...
+        Args:
+            X (pd.DataFrame): Data to transform.
+            y (pd.Series, optional): Ignored.
+
+        Returns:
+            pd.DataFrame: Transformed X
+        """
+        X = infer_feature_types(X)
+
+        # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
```
jeremyliweishih, line 111: if the columns are boolean they would be of dtype `boolean` instead of `bool` as that is what pandas uses for boolean nullables so it doesn't seem like we would hit this case. Can you explain why sklearn errors if all cols are bool? Related to my comment above, shouldn't we be transforming these columns to categorical?

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -0,0 +1,158 @@
...
+            col for col in not_all_null_cols if col not in natural_language_cols
+        ]
+        X_t = self._component_obj.transform(X_t)
+        X_t = pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)
+
+        # Add back in natural language columns, unchanged
+        if len(natural_language_cols) > 0:
+            X_t = woodwork.concat_columns([X_t, X[natural_language_cols]])
+
+        booleanNullable_columns = list(
```
jeremyliweishih, line 132: if we're imputing these `BooleanNullable` cols we can cast them to regular `Boolean` columns since they won't have null values anymore. I have a [PR](https://github.com/alteryx/evalml/pull/3678/files#diff-6daa63a0b019305d4ef1d44d1be7ac73769d1912ff22486bd551854d723d4393R176) up that does it for `Imputer.py`

jeremyliweishih (changes requested):
Left a couple suggestions but we should discuss how we're handling booleans here!

evalml/pipelines/components/transformers/imputers/imputer.py
MichaelFu512: If we're dealing with booleans, if I do any number of neighbors above 1, KNNImputer imputes the boolean as numbers such as 0.3333 or 0.5555.


I could make it so that it rounds to 0 (False) if the imputed number < 0.5 and round to 1 (True) if the imputed number >= 0.5

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+    X_multi = make_data_type(data_type, X_multi)
+
+    imputer = KNNImputer(number_neighbors=1)
+    imputer.fit(X_multi, y)
+    X_multi_t = imputer.transform(X_multi)
+    assert_frame_equal(X_multi_expected_arr, X_multi_t)
+
+
+def test_knn_imputer_all_bool():
```
eccabay, line 107: This test is almost functionally identical to `test_knn_imputer_all_bool_return_original`, we don't need both

eccabay (changes requested):
Looking pretty good! Agreed with Jeremy's comments though. I'd also love for you to take a closer look at the tests again - make sure we're testing the new behavior and the weird edge cases, but not throwing too much compute power at unnecessary tests. In general as well, I'd recommend leveraging our pytest fixtures for data creation whenever possible.

evalml/pipelines/components/utils.py
```diff
@@ -163,6 +163,45 @@ def handle_component_class(component_class):
     return component_class
 
 
+def drop_natural_language_columns(X):
```
eccabay, line 166: Thanks for doing this! It'd be great if you could add test cases for this function as well as `set_boolean_columns_to_categorical`.

eccabay (commented):
Looking really good! Just a few small things to fix before this goes in

evalml/pipelines/components/transformers/imputers/imputer.py
jeremyliweishih: I think thats a good idea - we should be able to up the number of neighbors. Can you file an issue for this?

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+def test_knn_imputer_all_bool_return_original(data_type, make_data_type):
+    X = pd.DataFrame([True, True, False, True, True], dtype=bool)
+    y = pd.Series([1, 0, 0, 1, 0])
+    X = make_data_type(data_type, X)
+    y = make_data_type(data_type, y)
+    X_expected_arr = pd.DataFrame([True, True, False, True, True], dtype=bool)
+    imputer = KNNImputer()
+    imputer.fit(X, y)
+    X_t = imputer.transform(X)
+    assert_frame_equal(X_expected_arr, X_t)
```
jeremyliweishih, line 40: let's also check the woodwork logical types are the same here.

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@pytest.mark.parametrize("data_type", ["pd", "ww"])
+def test_knn_imputer_boolean_dtype(data_type, make_data_type):
+    X = pd.DataFrame(
+        {
+            "some_nan": pd.Series([True, np.nan, False, np.nan, True], dtype="boolean"),
+        },
+    )
+    X.ww.init(logical_types={"some_nan": "BooleanNullable"})
```
jeremyliweishih, line 50: should we only init here based off if `data_type == ww`? we should either init or run `make_data_type` later.

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+    X_expected_arr = pd.DataFrame(
+        {
+            "some_nan": pd.Series([True, True, False, True, True]),
+        },
+    )
+    X = make_data_type(data_type, X)
+    imputer = KNNImputer(number_neighbors=1)
+    X_t = imputer.fit_transform(X, y)
+    X_ww_expected = "<Series: some_nan (Physical Type = bool) (Logical Type = Boolean) (Semantic Tags = set())>"
+    assert str(X_t.ww["some_nan"].ww) == X_ww_expected
```
jeremyliweishih, line 61: you can just assert that the logical type is Boolean by indexing into `X_t.ww`

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+
+@pytest.mark.parametrize("data_type", ["pd", "ww"])
+def test_knn_imputer_multitype_with_one_bool(data_type, make_data_type):
+    X_multi = pd.DataFrame(
+        {
+            "bool with nan": pd.Series([True, np.nan, False, np.nan, False]),
+            "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
+        },
+    )
+    X_multi.ww.init(logical_types={"bool with nan": "BooleanNullable"})
```
jeremyliweishih, line 73: same question about `init` and `make_data_type` here.

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+            ),
+            "bool no nan": pd.Series([False, False, False, False, True], dtype=bool),
+        },
+    )
+    X_multi = make_data_type(data_type, X_multi)
+
+    imputer = KNNImputer(number_neighbors=1)
+    imputer.fit(X_multi, y)
+    X_multi_t = imputer.transform(X_multi)
+    assert_frame_equal(X_multi_expected_arr, X_multi_t)
```
jeremyliweishih, line 89: let's assert that the logical types make sense here too.

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def test_knn_imputer_revert_categorical_to_boolean():
```
jeremyliweishih, line 92: is this test to check if `BooleanNullables` will be converted to `Boolean`? If so let's parameterize this method to use pandas and ww as well and then check the logical types in the WW case.

evalml/tests/component_tests/test_knn_imputer.py
```diff
@@ -0,0 +1,162 @@
...
+        X_df.ww.init()
+    y = pd.Series([x for x in range(X_df.shape[1])])
+
+    imputer = KNNImputer(number_neighbors=3)
+
+    imputer.fit(X_df, y)
+
+    result = imputer.transform(X_df, y)
+
+    # raise Exception
```
jeremyliweishih, line 150: is this comment necessary?

jeremyliweishih (approved):
Good work! Should be good to go after some more testing changes as well as Becca's comments.

evalml/tests/component_tests/test_knn_imputer.py
MichaelFu512: Line 92 was to test if I could revert categorical to boolean. However KNNImputer doesn't have the same issues that SimpleImputer has if you pass in just bools (no nulls), therefore I can get rid of line 92's test (since I don't have to actually change bool columns that are all bools into categorical).

evalml/tests/component_tests/test_knn_imputer.py
MichaelFu512: No oops

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -0,0 +1,128 @@
...
+        X_t = pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)
+        X_schema = X.ww.schema
+
+        original_X_schema = X_schema.get_subset_schema(
+            subset_cols=X_schema._filter_cols(exclude=["BooleanNullable"]),
+        )
+        X_bool_nullable_cols = X_schema._filter_cols(include=["Categorical"])
+        new_ltypes_for_bool_nullable_cols = {
+            col: "Boolean" for col in X_bool_nullable_cols
+        }
```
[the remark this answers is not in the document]
MichaelFu512: I rewrote the code above since honestly it confused even me reading it now. `downcast_nullable_types` turns booleans into doubles which I would prefer not to have.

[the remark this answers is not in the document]
eccabay: Hm, `downcast_nullable_types` appears to turn `BooleanNullable` into boolean, if you were getting doubles that seems like a problem we should address. It's not relevant to this issue though, and the cleaned up code is a lot easier to read so I'm fine with it!

eccabay (approved):
Awesome, thanks for powering through this! Let's get this in 😎

PR #3683, opened 2022-08-24
Tml 2022 backend 1st column id mark as primary key

simha104:
### Pull Request Description
IDColumnCheck now handles primary key columns containing "integer" values that are typed as doubles


evalml/data_checks/id_columns_data_check.py
```diff
@@ -175,9 +175,27 @@ def validate(self, X, y=None):
         ]  # columns whose name is "id"
         id_cols = {col: 0.95 for col in cols_named_id}
 
-        X = X.ww.select(include=["Integer", "Categorical"])
+        X_double = X.ww.select(include=["Double"])
 
-        check_all_unique = X.nunique() == len(X)
+        # Temporary solution for baton logical types mapping integers to doubles in woodwork logical types.
+        # Will be removed when resolved.
+        check_all_unique = X_double.nunique() == len(X_double)
```
bchen1116, line 182: To reduce the re-used code, I wonder if we can do either a for loop:
```python
for dtypes in [['Double'], [Integer', 'Categorical']]:
   # logic here
```
or to separate out this into a helper method, where we can pass in the `Double` and `Integer/Categorical` arguments. I know this is just a stopgap fix for now as we wait, but might be nice to not have repetitive code and reuse parts when we can.

evalml/data_checks/id_columns_data_check.py
```diff
@@ -175,9 +175,27 @@ def validate(self, X, y=None):
...
+        cols_with_all_unique = check_all_unique[
+            check_all_unique
+        ].index.tolist()  # columns whose values are all unique and doubles
+        cols_with_all_unique_integers = [
+            col for col in cols_with_all_unique if all(X_double[col].mod(1).eq(0))
```
bchen1116, line 187: I wonder if we can use
```python
col for col in cols_with_all_unique if all(X_double[col].is_integer())
```
to capture this logic

bchen1116 (commented):
LEft some comments on some suggestions to clean/improve code! looking good otherwise

evalml/data_checks/id_columns_data_check.py
simha104: This doesn't work unfortunately because `.is_integer` doesn't work for series types

evalml/data_checks/id_columns_data_check.py
```diff
@@ -1,4 +1,6 @@
 """Data check that checks if any of the features are likely to be ID columns."""
+from xml.etree.ElementInclude import include
```
bchen1116, line 2: What is this line for?

bchen1116 (approved):
Thanks for making the changes! This looks good to me

ParthivNaresh (approved):
Looks good to me!

PR #3686, opened 2022-08-30
Apply standard scalar only to numeric columns

jeremyliweishih:


evalml/pipelines/components/transformers/scalers/standard_scaler.py
```diff
@@ -21,14 +21,31 @@ class StandardScaler(Transformer):
     def __init__(self, random_seed=0, **kwargs):
         parameters = {}
         parameters.update(kwargs)
-
+        self._supported_types = [
```
jeremyliweishih, line 24: I think this covers all the types we want to manage here.

evalml/utils/woodwork_utils.py
```diff
@@ -147,5 +147,6 @@ def downcast_nullable_types(X, ignore_null_cols=True):
     new_ltypes.update(
         {col: "Double" for col in X_int_nullable_cols if col in non_null_columns},
     )
-    X.ww.init(schema=original_X_schema, logical_types=new_ltypes)
+    if new_ltypes:
```
jeremyliweishih, line 150: This was causing some weirdness where even though `new_ltypes` was empty - `BooleanNullable` columns would be re-inferred as a uknown string. @ParthivNaresh any idea why this might come up?

evalml/pipelines/components/transformers/scalers/standard_scaler.py
```diff
@@ -41,15 +68,20 @@ def transform(self, X, y=None):
         """
         X = infer_feature_types(X)
         X = X.ww.select(exclude=["datetime"])
```
jeremyliweishih, line 70: I think dropping datetime columns like this is weird but we're doing this to keep this compatible with partial dependence. Just going to leave this here for now but I do the same [here](https://github.com/alteryx/evalml/pull/3686/files#diff-da2d6fc3e28ef107b8015b493e115ed8b6ed81d796782a174111d222b9d82ba7R50) and removed the separate dropping in `fit_transform()`

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -516,17 +516,17 @@ def round_dict_keys(dictionary, places=6):
...
-        0.6353030372157324: 1,
-        0.031171284274810262: 1,
-        0.009093086236362272: 1,
-        0.33547688991040336: 1,
-        0.01746660818843149: 1,
-        0.018205973481273202: 1,
-        0.2876661156872482: 1,
-        0.015320197702897345: 1,
-        0.2821023107719306: 1,
+        0.08359496149480808: 63,
```
jeremyliweishih, line 519: this should have changed due to the the types we're looking at for the scalar.

evalml/utils/woodwork_utils.py
jeremyliweishih: The string uknown fix is actually in lines [147](https://github.com/alteryx/evalml/pull/3686/files#diff-19da4cc2ff7fdbf98a854e6d8a0ce26496df464ff0f2f0d8b4d470d1f54ace31R147-R153). I'm going to leave this if statement in as I don't think we need to go through an reinit if there's nothing to change!

evalml/utils/woodwork_utils.py
```diff
@@ -144,8 +144,16 @@ def downcast_nullable_types(X, ignore_null_cols=True):
...
+            for col in X_bool_nullable_cols
+            if col not in new_ltypes
+        },
+    )
     new_ltypes.update(
         {col: "Double" for col in X_int_nullable_cols if col in non_null_columns},
     )
-    X.ww.init(schema=original_X_schema, logical_types=new_ltypes)
+    if new_ltypes:
+        X.ww.init(schema=original_X_schema, logical_types=new_ltypes)
```
jeremyliweishih, line 158: @thehomebrewnerd

evalml/utils/woodwork_utils.py
thehomebrewnerd: @jeremyliweishih I think you should be good to use `.ww.set_types` here instead of reinitializing Woodwork. Furthermore, I think you might be able to simplify this code some by doing so. You should be able to drop all the code related to saving the original schema at the start, and use the `ww.select` method to get the boolean nullable and integer nullable columns directly.

evalml/utils/woodwork_utils.py
```diff
@@ -144,6 +144,13 @@ def downcast_nullable_types(X, ignore_null_cols=True):
     new_ltypes = {
         col: "Boolean" for col in X_bool_nullable_cols if col in non_null_columns
     }
+    new_ltypes.update(
+        {
+            col: "BooleanNullable"
+            for col in X_bool_nullable_cols
+            if col not in new_ltypes
+        },
+    )
```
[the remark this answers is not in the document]
jeremyliweishih: great idea! My thought here was that it would save time if it didn't have to set the logical type for every column but if we use 'init' I don't think we would be saving time. I'm discussing with @thehomebrewnerd and @ParthivNaresh about using 'set_types' here and it could be different with that.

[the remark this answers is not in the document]
fjlanasa: OK, that would be fine. But yeah I think not setting the type with `init` will actually take more time as woodwork would then try to infer it.

evalml/pipelines/components/transformers/scalers/standard_scaler.py
```diff
@@ -40,16 +65,18 @@ def transform(self, X, y=None):
...
-            return_schema=True,
+        X = X.ww.select(exclude=["datetime"])
+        if not self.scaled_columns:
+            return X
+        X_scaled_columns = X.ww[self.scaled_columns]
+        scaled = self._component_obj.transform(X_scaled_columns)
+        X[self.scaled_columns] = scaled
+        X.ww.init(
+            schema=X.ww.schema,
+            logical_types={col: "Double" for col in self.scaled_columns},
```
fjlanasa, line 77: should we also use `set_types` here?

fjlanasa (approved):

evalml/tests/component_tests/test_standard_scaler.py
```diff
@@ -1,32 +1,51 @@
...
-            continue
-
-        std_scaler = StandardScaler()
-        std_scaler.fit(X_df, y)
-        transformed = std_scaler.transform(X_df, y)
-        assert isinstance(transformed, pd.DataFrame)
-        assert {k: type(v) for k, v in transformed.ww.logical_types.items()} == {
-            0: Double,
-        }
+def test_standard_scaler_applies_to_numeric_columns_only():
```
eccabay, line 7: This is only a nit as it relates to my current work with hardening our tests against woodwork changes - using pytest fixtures for tests means we have all our woodwork typing consolidated and have to change fewer things should we need to tweak types for compatibility reasons. Plus, it should reduce the number of lines of code here.
I think either the `get_test_data_from_configuration` or the `imputer_test_data` fixture should have the same column types you're looking for.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -516,17 +516,17 @@ def round_dict_keys(dictionary, places=6):
...
-        0.6353030372157324: 1,
-        0.031171284274810262: 1,
-        0.009093086236362272: 1,
-        0.33547688991040336: 1,
-        0.01746660818843149: 1,
-        0.018205973481273202: 1,
-        0.2876661156872482: 1,
-        0.015320197702897345: 1,
-        0.2821023107719306: 1,
+        0.08407392065748104: 63,
```
eccabay, line 519: Why did this change?

docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
         * ``IDColumnsDataCheck`` now only returns an action code to set the first column as the primary key if it contains unique values :pr:`3639`
         * ``IDColumnsDataCheck`` now can handle primary key columns containing "integer" values that are of the double type :pr:`3683` 
         * Added support for BooleanNullable columns in EvalML pipelines and imputer :pr:`3678`
+        * Only apply ``StandardScaler`` to numeric columns :pr:`3686`
```
eccabay, line 9: Mega tiny itsy bitsy nit: `Only apply StandardScaler to numeric columns` -> `Updated StandardScaler to only apply to numeric columns`

I'm way too pedantic about the past tense 😁

eccabay (approved):
Looks solid! I just left a few nitpicks and conciseness comments, but nothing blocking.

evalml/pipelines/components/transformers/scalers/standard_scaler.py
```diff
@@ -21,14 +19,41 @@ class StandardScaler(Transformer):
...
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features].
+            y (pd.Series, optional): The target training data of length [n_samples].
+
+        Returns:
+            self
+        """
+        X = infer_feature_types(X)
+        X_can_scale_columns = X.ww.select(self._supported_types)
+        X_can_scale_columns = X_can_scale_columns.ww.select(exclude=["datetime"])
```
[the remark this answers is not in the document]
jeremyliweishih: I'm using this to keep a reference of the columns that scaling is applied to in the training data!

evalml/tests/model_understanding_tests/test_partial_dependence.py
jeremyliweishih: the fraud test data set contains `Categoricals` that are being ignored now.

evalml/pipelines/components/transformers/scalers/standard_scaler.py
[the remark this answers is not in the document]
jeremyliweishih: ah nvm I got what you mean!

PR #3690, opened 2022-09-01
add github action to run looking glass perf tests on merge to main

fjlanasa:
### Pull Request Description

This action will:
* run perf tests for the current and previous commit on main using both iterative and default algos
* generate reports and upload to s3
* send a message in slack with links to the reports

-----

### DIFF
--- a/.github/workflows/evalml_main_performance_tests.yaml
+++ b/.github/workflows/evalml_main_performance_tests.yaml
@@ -1,0 +1,168 @@
+name: Run looking glass performance tests
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  performance_tests:
+    name: Run looking glass performance tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_SECRET_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      AWS_DEFAULT_REGION: 'us-east-1'
+      AWS_SESSION_TOKEN: ""
+    steps:
+      - name: Checkout evalml
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Get commit hashes
+        id: get_hashes
+        run: |
+          current_hash=$(git rev-parse --short HEAD)
+          echo "Latest commit hash: $current_hash"
+          echo "::set-output name=current_hash::$current_hash"
+          previous_hash=$(git rev-parse --short HEAD~1)
+          echo "Previous commit hash: $previous_hash"
+          echo "::set-output name=previous_hash::$previous_hash"
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Checkout looking glass
+        uses: actions/checkout@v2
+        with:
+          repository: alteryx/looking-glass
+          ref: master
+          token: ${{ secrets.LOOKING_GLASS_TOKEN }}
+      - name: Installing Dependencies
+        run: |
+          make installdeps
+          make installdeps-test
+      - name: Run default algorithm performance test (current)
+        id: current_default
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo default --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.current_hash }} --job-desc ${{ steps.get_hashes.outputs.current_hash }}_default )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run iterative algorithm performance test (current)
+        id: current_iterative
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo iterative --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.current_hash }} --job-desc ${{ steps.get_hashes.outputs.current_hash }}_iterative )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run default algorithm performance test (previous)
+        id: previous_default
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo default --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_default )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run iterative algorithm performance test (previous)
+        id: previous_iterative
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo iterative --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_iterative )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Await results
+        id: results
+        run: |
+          for id in ${{ steps.current_default.outputs.job_id }} ${{ steps.current_iterative.outputs.job_id }} ${{ steps.previous_default.outputs.job_id }} ${{ steps.previous_iterative.outputs.job_id }}; do
+            echo "Waiting for job id: $id"
+            result=''
+            sleep_time=0
+            while [ -z "$result"  ]
+            do
+              sleep $sleep_time
+              result=$(looking-glass get-job --job-id $id | grep -n 'COMPLETED' || :;)
+              sleep_time=60
+            done
+            looking-glass get-results --job-id $id --file-path $id
+          done
+        shell: bash
+      - name: Generate reports
+        run: |
+          filename_base=${{ steps.get_hashes.outputs.current_hash }}_v_${{ steps.get_hashes.outputs.previous_hash }}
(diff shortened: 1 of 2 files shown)


jeremyliweishih (approved):
LGTM

chukarsten (approved):
This looks solid to me!  We might need to update the ticket and keep it open or add a new ticket to integrate this further with the "run lg on server" story.  Our ultimate goal will be that this story will not just perf test main, but also generate the report to diff against the last commit.  As it is right now (which is great), we're just perf testing main and will need to manually diff against the previous commit.

- [ ] Filed this [ticket](https://alteryx.atlassian.net/browse/TML-4918) to update our slackbot.
- [ ] Filed this [ticket](https://alteryx.atlassian.net/browse/TML-4919) to generate a report against the previous commit and post it in the slackbot.

fjlanasa:
> This looks solid to me! We might need to update the ticket and keep it open or add a new ticket to integrate this further with the "run lg on server" story. Our ultimate goal will be that this story will not just perf test main, but also generate the report to diff against the last commit. As it is right now (which is great), we're just perf testing main and will need to manually diff against the previous commit.
> 
> * [x]  Filed this [ticket](https://alteryx.atlassian.net/browse/TML-4918) to update our slackbot.
> * [x]  Filed this [ticket](https://alteryx.atlassian.net/browse/TML-4919) to generate a report against the previous commit and post it in the slackbot.

Yeah, I was

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+        id: previous_default
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo default --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_default )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run iterative algorithm performance test (previous)
+        id: previous_iterative
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo iterative --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_iterative )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Await results
```
jeremyliweishih, line 67: should we add a timeout here or is there a default timeout for github actions? If the ECS machines fall over, the status will never hit 'COMPLETED'.

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "<!subteam^${{ secrets.LG_SLACK_NOTIFICATION_USER_GROUP }}>, performance tests for latest commit on evalml are complete"
+                }
+              },
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "Author: ${{ github.event.head_commit.author.name }}"
```
jeremyliweishih, line 119: anyway to tag the author here? Does adding `@` work?

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+          ref: master
+          token: ${{ secrets.LOOKING_GLASS_TOKEN }}
+      - name: Installing Dependencies
+        run: |
+          make installdeps
+          make installdeps-test
+      - name: Run default algorithm performance test (current)
+        id: current_default
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo default --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.current_hash }} --job-desc ${{ steps.get_hashes.outputs.current_hash }}_default )
```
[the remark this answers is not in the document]
fjlanasa: Yeah, the output from that command would be a pain to parse. not many more lines to just do it like this.

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: Tag the author in slack? No. We are not consistent in our usernames/emails in git, so we'd have to maintain a mapping somewhere and hope people don't change it.

jeremyliweishih: Gotcha! Maybe we can consider doing if we think its worth it.

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: i'm assuming the jobs don't get marked as 'FAILED', so there's no way to know they died.

jeremyliweishih: I think six works for what we have (we could even lower it to 3 hours) but it could be insufficient if we ever need to run really large datasets.

jeremyliweishih: yup - thats true currently.

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+              sleep_time=60
+            done
+            looking-glass get-results --job-id $id --file-path $id
+          done
+        shell: bash
+      - name: Generate reports
+        run: |
+          filename_base=${{ steps.get_hashes.outputs.current_hash }}_v_${{ steps.get_hashes.outputs.previous_hash }}
+          looking-glass get-report --previous-path ${{ steps.previous_default.outputs.job_id }} -n ${{ steps.current_default.outputs.job_id }} --output-name ${filename_base}_default --output-type html
+          looking-glass get-report --previous-path ${{ steps.previous_iterative.outputs.job_id }} -n ${{ steps.current_iterative.outputs.job_id }} --output-name ${filename_base}_iterative --output-type html
```
christopherbunn, line 87: This will have to be updated once the[ looking glass server report generation ](https://github.com/alteryx/looking-glass/pull/546)gets merged in.

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run default algorithm performance test (previous)
+        id: previous_default
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo default --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_default )
+          echo "::set-output name=job_id::$(echo $response | sed 's/.*Job ID: \([^ ]*\).*/\1/')"
+      - name: Run iterative algorithm performance test (previous)
+        id: previous_iterative
+        run: |
+          response=$( looking-glass run-evalml-automl --username machineFL --automl-algo iterative --datasets-yaml datasets_v2.yaml --evalml-branch ${{ steps.get_hashes.outputs.previous_hash }} --job-desc ${{ steps.get_hashes.outputs.previous_hash }}_iterative )
```
christopherbunn, line 65: Not necessarily related to this specific PR, but I wonder if we can have a flag in Looking Glass that disables the slack output. I think it would prevent the looking glass notifications channel from being to cluttered every time we push a new PR.

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "Commit: <${{ github.event.head_commit.url }}|${{ steps.get_hashes.outputs.current_hash }}> - ${{ github.event.head_commit.message }}"
+                }
+              },
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "<$presigned_default_url|Default Algorithm Report>\n<$presigned_iterative_url|Iterative Algorithm Report>"
```
christopherbunn, line 133: Can we also get the looking glass job numbers in the Slack message? It would help in the future if we need to re-run reports.

.github/workflows/evalml_main_performance_tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "Author: ${{ github.event.head_commit.author.name }}"
+                }
+              },
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "Commit: <${{ github.event.head_commit.url }}|${{ steps.get_hashes.outputs.current_hash }}> - ${{ github.event.head_commit.message }}"
```
christopherbunn, line 126: Can we have the previous commit hash that was used in the message? It would make it clear what the report is comparing to.

christopherbunn (approved):
A few Slack message suggestions and next step suggestions but otherwise looks good to merge.

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: Yep will do

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: sure

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: Yeah that sounds like a good idea

.github/workflows/evalml_main_performance_tests.yaml
fjlanasa: Will it? Is the `get-report` command going away?

christopherbunn: Not really, but it might need to get renamed to `get-local-report`. Alternatively, there would also be a new command that kicks off the reporting job and stores it as a looking glass job. There's some more detail in the [updated user guide](https://github.com/alteryx/looking-glass/pull/546/files#) in that PR.

fjlanasa: can we just not change the name so it is backwards compatible?

PR #3691, opened 2022-09-02
Handle time index with exclude featurizers

thehomebrewnerd:
Various updates to handle dropping time index properly when `DateTimeFeaturizer` and `TimeSeriesFeaturizer` are excluded for time series problems. These updates retain time index column for time series native estimators, but drop it otherwise.


jeremyliweishih (approved):
LGTM

evalml/tests/integration_tests/test_time_series_integration.py
```diff
@@ -146,3 +147,63 @@ def test_can_run_automl_for_time_series_known_in_advance(
...
+            "time_index": "date",
+        },
+        optimize_thresholds=False,
+        exclude_featurizers=["DatetimeFeaturizer", "TimeSeriesFeaturizer"],
+    )
+    automl.search()
+
+    rankings = automl.rankings
+    for score in rankings["validation_score"].values:
+        assert pd.notnull(score)
```
eccabay, line 209: Can you add a test, either here or as an independent test, that `should_drop_time_index` is getting set correctly for all pipelines?

thehomebrewnerd: @eccabay I added the explicit check for the value of `should_drop_time_index` to this test. Let me know if there is a better way to do this as I'm not super familiar with the best way to iterate overall the pipelines generated by `.search()`.

eccabay: Works for me! I don't think we have a super clean way to iterate through all the pipelines right now, so imo whatever works goes.

eccabay (approved):

PR #3696, opened 2022-09-06
Use one sampler in split `DefaultAlgorithm` pipelines

### LINKED ISSUE #3076: Use one sampler for split preprocessing pipeline

jeremyliweishih:
Fixes #3076.


evalml/pipelines/utils.py
```diff
@@ -742,6 +743,7 @@ def _make_pipeline_from_multiple_graphs(
         pipeline_name (str): Custom name for the final pipeline.
         sub_pipeline_names (Dict): Dictionary mapping original input pipeline names to new names. This will be used to rename components. Defaults to None.
         prior_components (Dict): Component graph of components preceding the split of multiple graphs. Must be in component graph format, {"Label Encoder": ["Label Encoder", "X", "y"]} and currently restricted to components that only alter X input.
+        pre_estimator_components (Dict): Component graph of components before the estimator after the split of multiple graphs. Must be in component graph format, {"Label Encoder": ["Label Encoder", "X", "y"]} and currently restricted to components that only alter X input.
```
eccabay, line 746: I like this implementation - it's a smooth way of hooking up these sorts of components into the component graph. Slightly overengineered for the problem at hand, but will make our lives easier if/when we need to add more pre-estimator components in the future. Do you have any sort of plan for how we'll do that if that becomes the case? Since right now, we just have the one sampler and we set the input X and y knowing they'll be reset as the `first_pre_estimator_component`, but what would this look like with more components?
I don't think this is a blocking question, just something to keep in mind and maybe file an issue to track.

eccabay (approved):

evalml/pipelines/utils.py
jeremyliweishih: this implementation should work with more components if `pre_estimator_components` is a a component graph that hooks all the `pre_estimator_components` together. For example I believe the impl still would work if we have:

```
{sampler.name: [sampler.name, "X", "y"], "StandardScaler": ["StandardScaler, "sampler.X", "Sampler.y"]}
```

christopherbunn: +1 for this implementation, I think this will be super useful in the future. I'm okay with merging it as-is, but we might want to consider updating the parameter names for `prior_components` and `pre_estimator_components` to be a bit more precise? Maybe something with a pre/post scheme like `pre_pipeline_components`/`post_pipelines_components`.

christopherbunn (approved):

evalml/pipelines/utils.py
jeremyliweishih: will update @christopherbunn

evalml/pipelines/utils.py
```diff
@@ -844,9 +848,34 @@ def _make_new_component_name(name, component_name, idx=None, pipeline_name=None)
...
     final_y = final_y_candidate if final_y_candidate else final_y
-    component_graph[estimator.name] = (
-        [estimator] + [comp + ".x" for comp in final_components] + [final_y]
-    )
+    last_pre_estimator_component = None
+    if post_pipelines_components:
+        first_pre_estimator_component = (
+            list(post_pipelines_components.keys())[0]
+            if post_pipelines_components
+            else None
```
chukarsten, line 856: Isn't this first post_pipelines_components.keys()[0] guaranteed to exist?  You're already inside a `if post_pipelines_components:` branch here.

evalml/pipelines/utils.py
```diff
@@ -844,9 +848,34 @@ def _make_new_component_name(name, component_name, idx=None, pipeline_name=None)
...
+        post_pipelines_components[first_pre_estimator_component] = (
+            [first_pre_estimator_component]
+            + [comp + ".x" for comp in final_components]
+            + [final_y]
+        )
+        component_graph.update(post_pipelines_components)
+        last_pre_estimator_component = (
+            list(post_pipelines_components.keys())[-1]
+            if post_pipelines_components
+            else None
```
chukarsten, line 867: Same comment here with respect to the branch.

chukarsten (changes requested):
I think we just need some minor cleanups with respect to logic, but otherwise, good to go.

chukarsten (approved):
LGTM thanks man.

PR #3702, opened 2022-09-09
Provide useful debugging info when all pipelines are NaN

eccabay:
Adds a new `errors` attribute to the `AutoMLSearch` object which provides information on all the errors that occurred while running search. The error thrown if all pipelines in a batch are None or NaN references this so devs have more useful information during debugging.


jeremyliweishih (approved):

chukarsten (approved):
LGTM!  Thanks so much for this.  I think we might need a follow up story maybe to pretty print  the dictionary or something, but this is a big step in the right direction.

PR #3703, opened 2022-09-09
Fix warnings in RTD

### LINKED ISSUE #3620: Fix the Time Series Docs
[They](https://evalml.alteryx.com/en/stable/user_guide/timeseries.html) look a hot mess.

We need to handle all these warnings and clean it up.

jeremyliweishih:
Fixes #3620.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -26,7 +26,6 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.8
   install:
     - requirements: docs/readthedocs-requirements.txt
     - method: pip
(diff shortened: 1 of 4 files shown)


.readthedocs.yaml
```diff
@@ -3,6 +3,13 @@
 
 # Required
 version: 2
+build:
+  os: "ubuntu-20.04"
+  tools:
+    python: "3.8"
```
gsheni, line 9:
```suggestion
  tools:
    python: "3.8"
  apt_packages:
    - graphviz
```

.readthedocs.yaml
```diff
@@ -19,7 +26,6 @@ formats:
```
gsheni, line 26: You could remove the pdf output to speed up the build

PR #3704, opened 2022-09-09
Fix search order changing due to using a set in `DefaultAlgorithm`

jeremyliweishih:
Fixes search order changing.

### DIFF
--- a/evalml/automl/automl_algorithm/default_algorithm.py
+++ b/evalml/automl/automl_algorithm/default_algorithm.py
@@ -182,9 +182,12 @@ class DefaultAlgorithm(AutoMLAlgorithm):
         return estimators
 
     def _non_naive_estimators(self):
-        return list(
-            set(get_estimators(self.problem_type)) - set(self._naive_estimators()),
-        )
+        print(get_estimators(self.problem_type))
+        return [
+            est
+            for est in get_estimators(self.problem_type)
+            if est not in self._naive_estimators()
+        ]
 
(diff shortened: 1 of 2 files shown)


evalml/automl/automl_algorithm/default_algorithm.py
```diff
@@ -182,9 +182,11 @@ def _naive_estimators(self):
         return estimators
 
     def _non_naive_estimators(self):
-        return list(
-            set(get_estimators(self.problem_type)) - set(self._naive_estimators()),
```
jeremyliweishih, line 186: Using a set could change the order. This was only seen recently and between branches as well 😮

eccabay: Can we add a test to make sure this doesn't somehow happen again?

jeremyliweishih: not sure how I can test this since it didn't appear to change within the same run. Any suggestions?

eccabay: Hm, not exactly sure. We could hard code the expected order, although we'd have to keep it up to date with every new added estimator. Up to you whether you think we should add something like that or not. Either way, the change looks good!

eccabay (approved):

christopherbunn (approved):
Interesting to see this pop up because we're using a set, but ty for fixing!

PR #3721, opened 2022-09-20
Introduce the concept of ranking metrics

### LINKED ISSUE #3710: Add concept of ranking objective
Right now, we define all objectives as being either "core" objectives or "non-core" objectives, and use this distinction to divide objectives into domain-specific objectives and those which should be used in AutoMLSearch. However, there are objectives (such as recall and specificity) that are not domain-specific but nonetheless should not be used in AutoMLSearch. It would be more precise for us to define "optimization" objectives and "ranking" objectives to enable this scenario and divide objectives appropriately.

eccabay:
Closes #3710 

Note: with this implementation, `get_optimization_objectives` is a roundabout way to get our current core metrics. We will want to move away from core for optimization purposes, but we can’t remove the non-optimization non-core (i.e. recall) metrics until our downstream users have made the switch.


docs/source/demos/text_input.ipynb
```diff
@@ -453,7 +453,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
```
chukarsten, line 456: Might need a `make lint-fix`?

evalml/objectives/utils.py
```diff
@@ -30,6 +30,65 @@ def get_non_core_objectives():
...
+    Returns:
+        List of ObjectiveBase instances
+    """
+    problem_type = handle_problem_types(problem_type)
+    all_objectives_dict = _all_objectives_dict()
+    objectives = [
+        obj()
+        for obj in all_objectives_dict.values()
+        if obj.is_defined_for_problem_type(problem_type)
+        and (obj not in get_non_core_objectives() or obj in ranking_only_objectives())
```
chukarsten, line 87: remind me what's behind the `obj not in get_non_core_objectives() `

chukarsten (approved):

evalml/objectives/utils.py
eccabay: There are some objectives in that list that require additional setup/configuration before they can be used during optimization. An example, the cost benefit matrix needs weights for each true positive/negative and false positive/negative

evalml/objectives/utils.py
```diff
@@ -30,6 +30,65 @@ def get_non_core_objectives():
...
+def ranking_only_objectives():
+    """Get ranking-only objective classes.
+
+    Ranking-only objectives are objectives that are useful for evaluating the performance of a model, but should not
+    be used as an optimization objective during AutoMLSearch for various reasons.
+
+    Returns:
+        List of ObjectiveBase classes
+    """
+    return [
```
jeremyliweishih, line 42: thoughts on adding `ranking_only` as a property instead of maintaining a list? I know this is what we do for `get_non_core_objectives` as well but maybe we could file this as a future enhancement.

eccabay: I'm not quite sure I understand your suggestion - do you mean a property of the objective classes themselves? I do actually like that better, I think.

eccabay: Quick investigation shows that making this switch is unfortunately nontrivial, so I think I'll leave it as is for now. It's a good future enhancement though!

PR #3723, opened 2022-09-20
Ignore sorting of time index via woodwork for DataChecks

ParthivNaresh:


bchen1116 (approved):
Nice! Good catch

bchen1116:
Might be good to add a test to ensure that this is upheld as expected

Cmancuso (commented):
Also looks good to me, but agree with @bchen1116 on testing. Could be an integration test or tempo test as well if that's easier.

evalml/tests/data_checks_tests/test_data_checks.py
```diff
@@ -828,3 +829,39 @@ def validate(self, X, y):
...
+        "1/1/21",
+        "1/7/21",
+        "1/7/21",
+        "1/7/21",
+        "1/3/21",
+        "1/10/21",
+        "1/10/21",
+        "1/4/21",
+    ]
+    y = pd.Series([i for i in range(9)])
```
bchen1116, line 847: could shorthand this to `y = pd.Series(range(9))`

bchen1116 (approved):
Thanks for adding the test!

Cmancuso (approved):

PR #3732, opened 2022-09-26
Fix oversampler data typing error

eccabay:
Provides a workaround for an issue where categorical datatypes were causing the SMOTENC oversampler to fail. The SMOTENC oversampler can only handle `object` datatypes and not `categorical`. This was manifesting itself in CatBoost pipelines, since we don't one-hot encode the categorical features for CatBoost.

### DIFF
--- a/evalml/pipelines/components/transformers/samplers/base_sampler.py
+++ b/evalml/pipelines/components/transformers/samplers/base_sampler.py
@@ -94,7 +94,17 @@ class BaseSampler(Transformer):
             pd.DataFrame, pd.Series: Transformed features and target.
         """
         X, y = self._prepare_data(X, y)
+
+        categorical_columns = X.ww.select("Categorical", return_schema=True).columns
+        if len(categorical_columns):
+            for col in categorical_columns:
+                X[col] = X[col].astype("object")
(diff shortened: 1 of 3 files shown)


evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -94,8 +94,18 @@ def transform(self, X, y=None):
             pd.DataFrame, pd.Series: Transformed features and target.
         """
         X, y = self._prepare_data(X, y)
+
+        categorical_columns = X.ww.select("Categorical", return_schema=True).columns
+        if len(categorical_columns):
```
chukarsten, line 99: Might be a bit of a nit, but can't we just iterate over `categorical_columns` without the if?  I assume the select returns an empty iterable vs. None.

chukarsten (approved):

evalml/pipelines/components/transformers/samplers/base_sampler.py
eccabay: Great point!

fjlanasa (approved):

Issue #1389, opened 2020-11-02
Add Ordinal encoder component


dsherry:
https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html

Will need to run perf tests. Ideally, we can come up with some key examples where ordinal encoding outperforms one-hot encoding.

dsherry:
Added "needs design" because we should write out how this would be used. Need user to be able to select which categorical features have an ordering vs those which don't and that may require woodwork support first.

thehomebrewnerd:
@dsherry This issue came up recently in some experiments I have been doing. In reviewing the results with @rpeck and @rwedge we noticed that several ordinal columns were getting encoded as regular categorical columns by the EvalML OneHotEncoder, so we would get a feature such as `MONTH(Created)_9` for the 9th week of the year. @rpeck suggested we should not be encoding the Ordinal columns in this manner.

Any Woodwork columns that are ordered should be specified with the `Ordinal` logical type. Setting a column as `Ordinal` in Woodwork requires the order values to be defined, and the pandas dtype is set as `CategoricalDtype` with the specification that the values are ordered.

As a concrete example of this, the Featuretools `Month` primitive outputs an `Ordinal` column in the feature matrix with the following dtype:
```
CategoricalDtype(categories=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], ordered=True)
```

Between the Woodwork logical type and the pandas dtype ordering, it seems like there should be enough information present to determine what columns should have Ordinal encoding applied.

@gsheni FYI

PR #3736, opened 2022-09-28
Add OrdinalEncoder component

tamargrey:
Adds the OrdinalEncoder component, implementing https://github.com/alteryx/evalml/issues/1389. 

The implementation is loosely based off of the OneHotEncoder, with a few key differences:
- We only need to make one encoded feature for each ordinal column, so the naming schema is simplified
- Because all values in a column must be encoded to _something_, we do not have the option to ignore unknown values. We can either specify a value to encode them as or raise an error.
- There's not the same need to drop categories, since we don't have the problem of duplicate features in binary columns that exists with one-hot encoding
- There's not the same need to have a `top_n` parameter to limit the number of features created
- Missing values cannot be treated as a category. This is because Woodwork's Ordinal logical type doesn't include nans, which means that we can't specify a null value's place in the order. Therefore, we either keep nulls as `np.nan` or convert them to a separate encoded value that users can specify. 

(Note, this PR does not integrate the new ordinal encoder into the EvalML pipeline)

### DIFF
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -110,7 +110,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
         top_n = self.parameters["top_n"]
         X = infer_feature_types(X)
         if self.features_to_encode is None:
-            # --> should update to not include ordinals? Maybe that's configurable based on whether ordinal encoder is used?
+            # --> should update to not include ordinals once the ord encoder is integragted? Maybe that's configurable based on whether ordinal encoder is used?
             self.features_to_encode = self._get_cat_cols(X)
 
         X_t = X
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -110,7 +110,6 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
         top_n = self.parameters["top_n"]
         X = infer_feature_types(X)
         if self.features_to_encode is None:
-            # --> should update to not include ordinals once the ord encoder is integragted? Maybe that's configurable based on whether ordinal encoder is used?
             self.features_to_encode = self._get_cat_cols(X)
 
         X_t = X
--- a/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/onehot_encoder.py
@@ -110,6 +110,7 @@ class OneHotEncoder(Transformer, metaclass=OneHotEncoderMeta):
         top_n = self.parameters["top_n"]
         X = infer_feature_types(X)
         if self.features_to_encode is None:
+            # --> should update to not include ordinals? Maybe that's configurable based on whether ordinal encoder is used?
             self.features_to_encode = self._get_cat_cols(X)
 
         X_t = X
--- a/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
@@ -16,7 +16,7 @@ from evalml.utils import infer_feature_types
 
 class OrdinalEncoderMeta(ComponentBaseMeta):
     METHODS_TO_CHECK = ComponentBaseMeta.METHODS_TO_CHECK + [
-        # --> probs need to add categories back
+        "categories",
         "get_feature_names",
     ]
 
@@ -224,19 +224,19 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         """Returns a list of the unique categories to be encoded for the particular feature, in order.
 
         Args:
-            feature_name (str): The name of any feature provided to one-hot encoder during fit.
+            feature_name (str): The name of any feature provided to ordinal encoder during fit.
 
         Returns:
             np.ndarray: The unique categories, in the same dtype as they were provided during fit.
 
         Raises:
-            ValueError: If feature was not provided to one-hot encoder as a training feature.
+            ValueError: If feature was not provided to ordinal encoder as a training feature.
         """
         try:
             index = self.features_to_encode.index(feature_name)
         except Exception:
             raise ValueError(
-                f'Feature "{feature_name}" was not provided to one-hot encoder as a training feature',
+                f'Feature "{feature_name}" was not provided to ordinal encoder as a training feature',
             )
         return self._encoder.categories_[index]
 
--- a/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
@@ -147,9 +147,10 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         X_t = X
 
         ww_logical_types = X.ww.logical_types
+        categories = []
         if len(self.features_to_encode) == 0:
-            # No ordinal features present - categories don't need to be specified
-            categories = "auto"
+            # No ordinal features present - no transformation can take place so return early
+            return self
         elif self.parameters["categories"] is not None:
             # Categories specified - make sure they match the ordinal columns
             input_categories = self.parameters["categories"]
@@ -163,7 +164,6 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
 
             # Categories, as they're passed into SKOrdinalEncoder should be in the same order
             # as the data's Ordinal.order categories even if it's a subset
-            categories = []
             for i, col_categories in enumerate(input_categories):
                 categories_order = ww_logical_types[self.features_to_encode[i]].order
 
@@ -173,7 +173,6 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
                 categories.append(ordered_categories)
         else:
             # Categories unspecified - use ordered categories from a columns' Ordinal logical type
-            categories = []
             for col in X_t[self.features_to_encode]:
                 ltype = ww_logical_types[col]
                 # Copy the order list, since we might mutate it later by adding nans
@@ -184,10 +183,9 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         # This is needed because Ordinal.order won't indicate if nulls are present, and SKOrdinalEncoder
         # requires any null values be present in the categories list if they are to be encoded as
         # missing values
-        if isinstance(categories, list):
-            for i, col in enumerate(X_t[self.features_to_encode]):
-                if X_t[col].isna().any():
-                    categories[i] += [np.nan]
+        for i, col in enumerate(X_t[self.features_to_encode]):
+            if X_t[col].isna().any():
+                categories[i] += [np.nan]
 
         encoded_missing_value = self.parameters["encoded_missing_value"]
         if encoded_missing_value is None:
@@ -238,7 +236,7 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         """Return feature names for the ordinal features after fitting.
 
         Since ordinal encoding creates one encoded feature per column in features_to_encode, feature
-        names are formatted as {column_name}_ordinally_encoded
+        names are formatted as {column_name}_ordinal_encoding
 
         Returns:
             np.ndarray: The feature names after encoding, provided in the same order as input_features.
@@ -247,7 +245,7 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         unique_names = []
         provenance = {}
         for col_name in self.features_to_encode:
-            encoded_name = f"{col_name}_ordinally_encoded"
+            encoded_name = f"{col_name}_ordinal_encoding"
             unique_names.append(encoded_name)
             provenance[col_name] = [encoded_name]
         self._provenance = provenance
@@ -276,7 +274,7 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
     def get_feature_names(self):
         """Return feature names for the ordinal features after fitting.
 
-        Feature names are formatted as {column name}_ordinally_encoded.
+        Feature names are formatted as {column name}_ordinal_encoding.
 
         Returns:
             np.ndarray: The feature names after encoding, provided in the same order as input_features.
--- a/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
@@ -63,7 +63,7 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
         categories=None,
         handle_unknown="error",
         unknown_value=None,
-        encoded_missing_value=np.nan,
+        encoded_missing_value=None,
         random_seed=0,
         **kwargs,
     ):
@@ -224,11 +224,15 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
                 if X_t[col].isna().any():
                     categories[i] += [np.nan]
 
+        encoded_missing_value = self.parameters["encoded_missing_value"]
+        if encoded_missing_value is None:
+            encoded_missing_value = np.nan
+
         self._encoder = SKOrdinalEncoder(
             categories=categories,
             handle_unknown=self.parameters["handle_unknown"],
             unknown_value=self.parameters["unknown_value"],
-            encoded_missing_value=self.parameters["encoded_missing_value"],
+            encoded_missing_value=encoded_missing_value,
         )
 
         self._encoder.fit(X_t[self.features_to_encode])
--- a/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
+++ b/evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
@@ -142,9 +142,13 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
             ]
 
         X_t = X
+
+        ww_logical_types = X.ww.logical_types
         if len(self.features_to_encode) == 0:
+            # No ordinal features present - categories don't need to be specified
             categories = "auto"
         elif self.parameters["categories"] is not None:
+            # Categories specified - make sure they match the ordinal columns
             input_categories = self.parameters["categories"]
             if len(input_categories) != len(self.features_to_encode) or not isinstance(
                 input_categories[0],
@@ -156,27 +160,29 @@ class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
 
             # Categories, as they're passed into SKOrdinalEncoder should be in the same order
             # as the data's Ordinal.order categories even if it's a subset
-            # --> refactor this to be nicer
             categories = []
             for i, col_categories in enumerate(input_categories):
-                categories_order = X.ww.logical_types[self.features_to_encode[i]].order
+                categories_order = ww_logical_types[self.features_to_encode[i]].order
 
                 ordered_categories = [
                     cat for cat in categories_order if cat in col_categories
                 ]
                 categories.append(ordered_categories)
         else:
+            # Categories unspecified - use ordered categories from a columns' Ordinal logical type
+            # and account for top_n
             categories = []
-            ww_logical_types = X.ww.logical_types
             for col in X_t[self.features_to_encode]:
                 ltype = ww_logical_types[col]
+                # Copy the order list, since we might mutate it later by adding nans
+                # and don't want to impact the Woodwork types
+                column_ordered_categories = ltype.order.copy()
 
-                column_categories = ltype.order
-
-                if top_n is None or len(column_categories) <= top_n:
-                    unique_values = column_categories
+                if top_n is None or len(column_ordered_categories) <= top_n:
+                    unique_values = column_ordered_categories
                 else:
-                    value_counts = X_t[col].value_counts(dropna=False).to_frame()
+                    # Don't count nans as a value
+                    value_counts = X_t[col].value_counts(dropna=True).to_frame()
                     value_counts = value_counts.sample(
                         frac=1,
                         random_state=self._initial_state,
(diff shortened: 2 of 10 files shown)


evalml/pipelines/components/transformers/encoders/onehot_encoder.py
```diff
@@ -110,6 +110,7 @@ def fit(self, X, y=None):
         top_n = self.parameters["top_n"]
         X = infer_feature_types(X)
         if self.features_to_encode is None:
+            # --> should update to not include ordinals once the ord encoder is integragted? Maybe that's configurable based on whether ordinal encoder is used?
```
tamargrey, line 113: A note to myself about integrating the new component. Do we need a separate issue for actually using the component? Or do we use the same issue?

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,322 @@
...
+        unknown_value=None,
+        encoded_missing_value=None,
+        random_seed=0,
+        **kwargs,
+    ):
+        parameters = {
+            "top_n": top_n,
+            "features_to_encode": features_to_encode,
+            "categories": categories,
+            "handle_unknown": handle_unknown,
```
tamargrey, line 74: I didn't include a `handle_missing` parameter, since "as_category" isn't an option here, but it means that we don't have the option to error when nans are seen. 

Do we want a `handle_missing` parameter that is either "use_encoded_value" or "error" and then pairs with the encoded_missing_value parameter?

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,322 @@
...
+
+        Feature names are formatted as {column name}_ordinally_encoded.
+
+        Returns:
+            np.ndarray: The feature names after encoding, provided in the same order as input_features.
+        """
+        return self._get_feature_names()
+
+    def _get_feature_provenance(self):
+        return self._provenance
```
tamargrey, line 322: This isn't yet covered by tests, I assume, because it's not used in the EvalML pipeline. I didn't see tests in the other components for this method, so are we okay leaving this uncovered for now?

evalml/tests/component_tests/test_ordinal_encoder.py
```diff
@@ -0,0 +1,1019 @@
...
+    - test_drop_parameter_is_array
+    - test_drop_binary_and_top_n_2
+    - test_ohe_column_names_unique
+Couldn't understand the reason for
+    - test_categorical_dtype
+    - test_all_numerical_dtype
+    - test_ordinal_encoder_woodwork_custom_overrides_returned_by_components
+Seemed redundant to other tests
+    - test_more_top_n_unique_values_large
+    - test_large_number_of_categories - kind of just another test of top_n arg
```
tamargrey, line 1018: I'll remove this comment, but I wanted to highlight the tests from the one hot encoder that I didn't include for the ordinal encoder. 

The top section here is just for tests that were for behaviors that don't exist on this encoder, but for the bottom two sections, I'm less confident in having not included them and just wanted to see if anyone had particularly strong feelings about me not including them.

evalml/tests/component_tests/test_ordinal_encoder.py
```diff
@@ -0,0 +1,1019 @@
...
+def test_data_types(data_type):
+    if data_type == "list":
+        X = [["a"], ["b"], ["c"]] * 5
+    elif data_type == "np":
+        X = np.array([["a"], ["b"], ["c"]] * 5)
+    elif data_type == "pd_no_index":
+        X = pd.DataFrame(["a", "b", "c"] * 5)
+    elif data_type == "pd_index":
+        # --> doing int 0 here might defeat the purpose of the no index one?
+        X = pd.DataFrame(["a", "b", "c"] * 5, columns=[0])
```
tamargrey, line 972: Not totally sure what this case was testing when `columns=["0"]`, but I think me setting the index to be an integer means it's essentially the same as the `pd_no_index` case, but given that this test can only actually perform ordinal encoding in the `ww` case (because Ordinal will never be inferred and so won't be used unless woodwork is initialized), I'm not sure this test is as useful as it is for the one-hot encoder. Maybe I should just remove it?

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+        # requires any null values be present in the categories list if they are to be encoded as
+        # missing values
+        if isinstance(categories, list):
+            for i, col in enumerate(X_t[self.features_to_encode]):
+                if X_t[col].isna().any():
+                    categories[i] += [np.nan]
+
+        encoded_missing_value = self.parameters["encoded_missing_value"]
+        if encoded_missing_value is None:
+            encoded_missing_value = np.nan
```
eccabay, line 227: I'm having trouble figuring out if this section is worth the effort or not. If I'm following correctly, we should only allow `np.nan` as a category if `handle_unknown` is `"use_encoded_value"`. In that case, we don't allow `encoded_missing_value` to be None (lines 86 and 87 in init), so the second half of this seems moot.

In that case (I've been thinking while typing here), I might replace this section (lines 220-227) with:
```
if self.parameters["handle_unknown"] == "use_encoded_value":
    for i, col in enumerate(X_t[self.features_to_encode]):
        if X_t[col].isna().any():
            categories[i].append(np.nan)
```
Then, when instantiating the encoder, I'd just do `encoded_missing_value=self.parameters["encoded_missing_value"] or np.nan`. 
I think that would be a clearer way to do it, as it's a little confusing as it is now.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+            y (pd.Series): Ignored.
+
+        Returns:
+            pd.DataFrame: Transformed data, where each ordinal feature has been encoded into
+            a numerical column using where of ordinal integers represent
+            the relative order of categories.
+        """
+        X = infer_feature_types(X)
+
+        X_copy = X.ww.copy()
```
eccabay, line 253: Nitpick about variable names: I would replace the X with X_t and X_copy with X. We should also avoid the need to copy in that case, since calling `X.ww.drop` should return a new DataFrame.
```
X_t = X.ww.drop(columns=self.features_to_encode)
```

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+
+            X = ww.utils.concat_columns([X, X_ord])
+
+        return X
+
+    def _get_feature_names(self):
+        """Return feature names for the ordinal features after fitting.
+
+        Since ordinal encoding creates one encoded feature per column in features_to_encode, feature
+        names are formatted as {column_name}_ordinally_encoded
```
eccabay, line 274: Biggest nitpick: can we name them `{column_name}_ordinal_encoding`? It feels neater idk

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
eccabay: I'm certainly not fussed about it, but if anyone disagrees speak now

evalml/tests/component_tests/test_ordinal_encoder.py
eccabay: We do still want to make sure none of our components break when passed any standard data type, so we should still have a test that throws them at the encoder making sure it doesn't fall over. I think you're right in that we don't need to differentiate about the pandas index, so we can probably condense this to just test list, np, pd, and ww instead, and make sure nothing happens in any case except woodwork.

(And in that case I think we can remove the numpy test on line 744 since we're doing it here instead)

evalml/tests/component_tests/test_ordinal_encoder.py
eccabay: IMO all your choices seem correct. I think the `test_categorical_dtype`/`test_all_numeric_dtype` is to ensure the component only works on the correct columns (and nothing more). As long as we have a test in this file where none of the columns in the (woodwork-typed) input data are Ordinal, that should be covered for this case.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+                    # Categories should be in the same order as the data's Ordinal.order categories
+                    # instead of sorted by most frequent
+                    unique_values = [
+                        cat for cat in column_ordered_categories if cat in unique_values
+                    ]
+
+                categories.append(unique_values)
+
+        # Add any null values into the categories lists so that they aren't treated as unknown values
+        # This is needed because Ordinal.order won't indicate if nulls are present, and SKOrdinalEncoder
```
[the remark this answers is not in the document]
tamargrey: It's not specifically mentioned in any of the docs, from what I can tell. But I do think it's intended and necessary for two reasons:
1. The Ordinal ltype sets the pandas `category` dtype's categories to be `order`, and the categories are not allowed to be null, or you get `ValueError: Categorical categories cannot be null` from pandas.
2. The implementation of Ordinal, when validating the `order`, checks for values in the series that are not in `order` by [explicitly excluding nans from that check](https://github.com/alteryx/woodwork/blob/main/woodwork/logical_types.py#L638). This allows nans to be in the series without being in the `order`, without which, you couldn't have nans in an Ordinal column at all (because of reason 1).

So maybe I just make a documentation issue asking them to add a note to the logical type docstring about how nans are handled

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: (I haven't totally thought through what you're actually talking about in this comment, but the use of `or` here brought up thoughts I didn't want to forget!)

Whenever I see the `or` operator used like this, I always worry about falsy values falling to the second option, and I think we might be in the clear here on a technicality, but I'm not sure we'd want to still go with it. 

The encoded missing value has to be numeric in nature, which takes away empty string, empty set, etc as issues. `0` is where I'm not sure; technically users aren't allowed to use `0` since it'll always be the first encoded value unless none of the known categories are seen in the column. But instead of ever getting an error for using `0`, we'd convert to np.nan and be able to proceed. So we'd be allowing something that's not allowed in most cases. And in the few cases where it's allowed, we wouldn't be doing what the user asked for.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Of course! I'm not confident that "ordinally" is even a word, so I'm happy to change

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+                ltype = logical_types[col]
+                if not isinstance(ltype, Ordinal):
+                    raise TypeError(
+                        f"Column {col} specified in features_to_encode is not Ordinal in nature",
+                    )
+
+            # Put features_to_encode in the same relative order as the columns in the dataframe
+            self.features_to_encode = [
+                col for col in X.columns if col in self.features_to_encode
+            ]
```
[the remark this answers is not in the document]
tamargrey: This is in place because we say that 
> `categories[i]` is a list of the categories for the column at index `i` in the dataframes passed in at fit and transform

That kind of breaks down once you're specifying `features_to_encode` or if not all of the dataframe's columns are ordinal in nature, so I'm making the assumption that the order of the sublists in `categories` is just the relative order from the original dataframe. 

I can make note of that assumption in the docstring. Alternatively, we could just make `categories` a dictionary mapping column name to the list of categories--was there a specific reason it's a list of lists other than that's what we pass into the SKLearn encoder? The SKLearn encoder requires that categories be the same length as the number of columns in the dataframe, and now that I really think about it, a dictionary probably fits our use case better.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: @eccabay So "missing" value handling is actually totally independent of "unknown" handling with the way we do things. (Lines 86 and 87 are referencing `unknown_value`)

We don't ever want missing values to get counted as unknown (because with Woodwork, they can't actually be categories and can only ever be missing values - the same thing as I mentioned in another comment - so I wouldn't consider them unknown if they're outside of the concept of categories to woodwork). If we want the option to error when nans are seen, I think that should be its own parameter. 

And I think my reservations about using the `or` syntax still hold, so I'm going to keep it as is for now. 

But overall, I agree with the sentiment that it feels like there should be a cleaner way to handle a lot of this logic. Maybe fresh eyes next week will see something I'm missing right now!

evalml/tests/component_tests/test_ordinal_encoder.py
tamargrey: Yep, we're covered with `test_ordinal_encoder_is_no_op_for_df_of_non_ordinal_features`

evalml/tests/component_tests/test_ordinal_encoder.py
tamargrey: Done!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+class OrdinalEncoder(Transformer, metaclass=OrdinalEncoderMeta):
+    """A transformer that encodes ordinal features as an array of ordinal integers representing the relative order of categories.
+
+    Args:
+        top_n (int): Number of categories per column to encode. If None, all categories will be encoded.
+            Otherwise, the `n` most frequent will be encoded and all others will be handled as unknown values.
+            To not have unknown values raise an error, set handle_unknown to "use_encoded_value".
+            Defaults to 10.
```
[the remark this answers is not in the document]
tamargrey: Removed: bf0fec73c

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+            self.features_to_encode = [
+                col for col in X.columns if col in self.features_to_encode
+            ]
+
+        X_t = X
+
+        ww_logical_types = X.ww.logical_types
+        if len(self.features_to_encode) == 0:
+            # No ordinal features present - categories don't need to be specified
+            categories = "auto"
```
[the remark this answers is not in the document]
tamargrey: Good call -- making this change - it also lets us simplify the check below of whether `categories` is a list

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Changed!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Implemented this in dd5c075ff - happy to change if we should be doing this differently!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+        elif self.parameters["categories"] is not None:
+            # Categories specified - make sure they match the ordinal columns
+            input_categories = self.parameters["categories"]
+            if len(input_categories) != len(self.features_to_encode) or not isinstance(
+                input_categories[0],
+                list,
```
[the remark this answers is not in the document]
tamargrey: Fair point--changed to now be two errors and to confirm all elements are lists: f791cfdcb

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def __init__(
+        self,
+        features_to_encode=None,
+        categories=None,
+        handle_unknown="error",
+        unknown_value=None,
+        encoded_missing_value=None,
+        random_seed=0,
+        **kwargs,  # --> why do we have kwargs if they arent used?? Maybe needed for larger evalml pipeline
```
eccabay, line 63: To answer your question here, kwargs is generally added across all components in case the user wants to set parameters in the component object that we don't include in our parameters. 

Nominally, with the ordinal encoder the sklearn implementation offers a `dtype` parameter that we don't expose, but the user could choose to set here. Since we don't instantiate the component until fit time this becomes a little moot. I'm not sure whether we should worry about passing them into the component object or not, but we should at least keep this function signature the same. The same holds for the `random_seed` parameter which I see you've commented on as well.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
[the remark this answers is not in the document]
eccabay: I'll defer to you on this one! If a dictionary makes more sense I support switching over, but this is just fine as is.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,289 @@
...
+                    raise TypeError(
+                        f"Column {col} specified in features_to_encode is not Ordinal in nature",
+                    )
+
+            # Put features_to_encode in the same relative order as the columns in the dataframe
+            self.features_to_encode = [
+                col for col in X.columns if col in self.features_to_encode
+            ]
+
+        X_t = X
```
eccabay, line 147: Why do this here? We shouldn't transform X at all during fit, and it seems like the rest of this just uses one or the other interchangeably.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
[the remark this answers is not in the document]
eccabay: Gotcha, a documentation update would definitely be helpful!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,289 @@
...
+        # missing values
+        for i, col in enumerate(X_t[self.features_to_encode]):
+            if X_t[col].isna().any():
+                categories[i] += [np.nan]
+
+        encoded_missing_value = self.parameters["encoded_missing_value"]
+        if encoded_missing_value is None:
+            encoded_missing_value = np.nan
+
+        self._encoder = SKOrdinalEncoder(
```
eccabay, line 196: This should be saved as the `_component_obj` to match the pattern of the rest of our components!

eccabay (approved):
Looks great! Just left some final cleanup comments but overall this is solid

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
eccabay: Ah, thanks for pointing out the missing vs unknown distinction, I totally combined them! Your point about `or` is also definitely important, you're right that we should avoid it!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
eccabay: I'm not sure what you mean here by "we don't want users messing around with the same object they passed in", could you explain that a little bit? If we're trying not to modify the user's original data, this wouldn't be the place to enforce it. I'm also worried that copying the data (especially when this component should be a no-op) uses unnecessary time and storage, especially with larger datasets.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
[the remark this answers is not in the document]
tamargrey: Created a woodwork issue: https://github.com/alteryx/woodwork/issues/1538

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Yeah, I was assuming that this was the place to be worrying about modifying the user's original data. My thought was that if you pass a dataframe into `transform`, we shouldn't ever give you back the exact same object. But if that's not a contract we need to uphold here, I'm more than happy to avoid the extra computation!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Awesome--I'll keep them in. Thanks for explaining!

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: Changing! 

Out of curiosity: Is this out of convention or is it needed to work properly in evalml? I ask, because this was another holdover from the onehot encoder, and I do see `_encoder` used elsewhere across the repo (though not nearly as much as `_component_obj`)

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
tamargrey: I think that was a holdover from the onehot encoder, but I can definitely remove

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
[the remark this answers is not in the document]
tamargrey: Okay, I went through the refactor to use a dict, and I continue to be a fan. It wasn't a small number of changes, though:  54386f0fa

It lets users avoid having to worry about the relative order of their inputs to the encoder and lets us remove this line.

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
eccabay: It's mostly a convention thing, afaik

evalml/pipelines/components/transformers/encoders/ordinal_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+                ltype = logical_types[col]
+                if not isinstance(ltype, Ordinal):
+                    raise TypeError(
+                        f"Column {col} specified in features_to_encode is not Ordinal in nature",
+                    )
+
+            # Put features_to_encode in the same relative order as the columns in the dataframe
+            # self.features_to_encode = [
+            #     col for col in X.columns if col in self.features_to_encode
+            # ]
```
eccabay, line 143: Commented code!

tamargrey: removed! 🤦‍♀️

jeremyliweishih (approved):
LGTM! well done 😄

evalml/pipelines/components/transformers/encoders/onehot_encoder.py
jeremyliweishih: We usually file another ticket but it's flexible!

tamargrey: Created a follow-up issue: https://github.com/alteryx/evalml/issues/3744

PR #3742, opened 2022-10-05
Add forecast functions to time series regression pipeline

christopherbunn:
Adds `get_forecast_periods()` and `get_forecast_predictions()` to time series regression pipelines.

### DIFF
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -107,7 +107,7 @@ class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
             pd.Series: Datetime periods out to `forecast_horizon + gap`.
 
         Example:
-            >>> X = X = pd.DataFrame({'date': pd.date_range(start='1-1-2022', periods=10, freq='D'), 'feature': range(10, 20)})
+            >>> X = pd.DataFrame({'date': pd.date_range(start='1-1-2022', periods=10, freq='D'), 'feature': range(10, 20)})
             >>> y = pd.Series(range(0, 10), name='target')
             >>> gap = 1
             >>> forecast_horizon = 2
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -1,6 +1,11 @@
 """Pipeline base class for time series regression problems."""
+import numpy as np
+import pandas as pd
+from woodwork.statistics_utils import infer_frequency
+
 from evalml.pipelines.time_series_pipeline_base import TimeSeriesPipelineBase
 from evalml.problem_types import ProblemTypes
+from evalml.utils.woodwork_utils import infer_feature_types
 
 
 class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
@@ -54,6 +59,7 @@ class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
             ValueError: If the target is not numeric.
         """
         X, y = self._convert_to_woodwork(X, y)
+        self.frequency = infer_frequency(X[self.time_index])
 
         if "numeric" not in y.ww.semantic_tags:
             raise ValueError(
@@ -88,3 +94,56 @@ class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
             y_pred_proba=None,
             objectives=objectives,
         )
+
+    def get_forecast_periods(self, X):
+        """Generates all possible forecasting periods based on last period of X.
+
+        Args:
+            X (pd.DataFrame, np.ndarray): Data the pipeline was trained on of shape [n_samples_train, n_feautures].
+
+        Raises:
+            ValueError: If pipeline is not trained.
+
+        Returns:
+            pd.Series: Datetime periods out to `forecast_horizon + gap`.
+
+        """
+        if not self._is_fitted:
+            raise ValueError("Pipeline must be fitted before getting forecast.")
+
+        X = infer_feature_types(X)
+        # Generate prediction periods
+        first_idx = X.index[-1]
+        first_date = X.iloc[-1][self.time_index]
+
+        # Add additional period to account for dropping first date row
+        predicted_date_range = pd.Series(
+            pd.date_range(
+                start=first_date,
+                periods=self.forecast_horizon + self.gap + 1,
+                freq=self.frequency,
+            ),
+        )
+
+        # Generate sequential index
+        num_idx = pd.Series(range(first_idx, first_idx + predicted_date_range.size))
+
+        predicted_date_range.index = num_idx
+        predicted_date_range = predicted_date_range.drop(predicted_date_range.index[0])
+        predicted_date_range.name = self.time_index
+        return predicted_date_range
+
+    def get_forecast_predictions(self, X, y):
+        """
+
+        Args:
+            X (pd.DataFrame, np.ndarray): Data the pipeline was trained on of shape [n_samples_train, n_feautures].
+            y (pd.Series, np.ndarray): Targets used to train the pipeline of shape [n_samples_train].
+
+        Returns:
+            Predictions.
+        """
+        X, y = self._convert_to_woodwork(X, y)
+        pred_dates = pd.DataFrame(self.get_forecast_periods(X), index=[self.time_index])
+        preds = self.predict(pred_dates, objective=None, X_train=X, y_train=y)
+        return preds
--- a/evalml/pipelines/time_series_regression_pipeline.py
+++ b/evalml/pipelines/time_series_regression_pipeline.py
@@ -111,7 +111,9 @@ class TimeSeriesRegressionPipeline(TimeSeriesPipelineBase):
         if not self._is_fitted:
             raise ValueError("Pipeline must be fitted before getting forecast.")
 
+        X = X.reset_index(drop=True)  # Create sequential numeric index
         X = infer_feature_types(X)
+
         # Generate prediction periods
         first_idx = X.index[-1]
         first_date = X.iloc[-1][self.time_index]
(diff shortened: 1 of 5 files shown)


evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -73,3 +74,41 @@ def test_time_series_baseline(
...
+    assert np.issubdtype(result.dtype, np.datetime64)
+    assert result.name == "date"
+
+
+@pytest.mark.parametrize("forecast_horizon,gap", [[3, 0], [10, 2], [2, 5]])
+def test_time_series_get_forecast_predictions(forecast_horizon, gap, ts_data):
+    X, _, y = ts_data(problem_type=ProblemTypes.TIME_SERIES_REGRESSION)
+
+    X_train, y_train = X.iloc[:15], y.iloc[:15]
+    X_validation = X.iloc[15 : (15 + gap + forecast_horizon)]
```
christopherbunn, line 113: This holdout dataset has a `feature` column that is obviously not generated inside of `get_forecast_predictions()` (i.e. it is multivariate).

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -88,3 +93,57 @@ def score(self, X, y, objectives, X_train=None, y_train=None):
...
+                start=first_date,
+                periods=self.forecast_horizon
+                + self.gap
+                + 1,  # Add additional period to account for dropping first date row
+                freq=self.frequency,
+            ),
+        )
+
+        # Generate numerical index
+        first_idx = len(X) - 1 if not isinstance(X.index.dtype, int) else X.index[-1]
```
christopherbunn, line 128: Using this to create a int index that is sequential after X_train

christopherbunn (commented):
Changes are going to be only for regression pipelines for now. Can move to base class when we revisit for time series classification problems.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -88,3 +93,57 @@ def score(self, X, y, objectives, X_train=None, y_train=None):
             y_pred_proba=None,
             objectives=objectives,
         )
+
+    def get_forecast_periods(self, X):
+        """Generates all possible forecasting periods based on last period of X.
```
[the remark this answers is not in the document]
eccabay: Also, a wording clarification/nitpick: how are you defining "period"? When I see period I think about a range of dates, in which case this function would return a single period. I know pandas calls them periods so that's how we're using in this case, but I'm not sure it's very clear with the wording in this function.

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -73,3 +74,52 @@ def test_time_series_baseline(
...
+    ):
+        clf.get_forecast_periods(X)
+
+    clf.fit(X, y)
+    result = clf.get_forecast_periods(X)
+
+    assert result.size == forecast_horizon + gap
+    assert all(result.index == range(len(X), len(X) + forecast_horizon + gap))
+    assert result.iloc[0] == X.iloc[-1]["date"] + np.timedelta64(1, "D")
+    assert np.issubdtype(result.dtype, np.datetime64)
```
eccabay, line 104: Out of curiosity, why `issubdtype` here? What sort of dtypes are we getting?

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -73,3 +74,52 @@ def test_time_series_baseline(
...
+        ProblemTypes.TIME_SERIES_REGRESSION,
+        gap,
+        forecast_horizon,
+        time_index="date",
+    )
+    clf.fit(X_train, y_train)
+    forecast_preds = clf.get_forecast_predictions(X=X_train, y=y_train)
+    X_val_preds = clf.predict(X_validation, X_train=X_train, y_train=y_train)
+
+    assert_series_equal(forecast_preds, X_val_preds)
```
eccabay, line 125: Check my understanding here: the only reason these two predictions are the same is because we're using a baseline pipeline, so the extra feature in X doesn't have an impact. For other models though, this isn't going to be the case.

Can we add info/a warning in the function (and docstring, as well as tests) to make users aware that the `get_forecast_predictions` function only works for univariate problems? 

Is that something we're ok with, since we're forecasting?

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
christopherbunn: Essentially I'm looking to make sure we're returning datetimes here. I'm using `issubtype` since this [generates `<M8[ns]` objects](https://stackoverflow.com/questions/29206612/difference-between-data-type-datetime64ns-and-m8ns) and per [this Stack Overflow](https://stackoverflow.com/questions/23063362/consistent-way-to-check-if-an-np-array-is-datetime-like) this is the best way to check for datetimes.

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
christopherbunn: These forecasting functions are meant to work for both univariate and multivariate problems. Because we have a forecast horizon and a gap with time series problems, we need to pull from past data. In the case of multivariate problems, we are pulling these features from past rows of `X_train` after it is transformed through the entire pipeline. For univariate problems, just the periods generated should be fine. 

I agree that we should probably use a more complex pipeline that utilizes features though. Updated to include this.

evalml/pipelines/time_series_regression_pipeline.py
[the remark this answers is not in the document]
christopherbunn: Good point. My intent was to have a period = a single point of time but I guess this isn't technically true. I can change the function name to `get_forecast_period()` (singular) and say "Generates all possible forecasting time points based on last data point in X" for the docstring.

Will update with docstring example.

jeremyliweishih (approved):
LGTM - good work!

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -88,3 +93,57 @@ def score(self, X, y, objectives, X_train=None, y_train=None):
...
+            X (pd.DataFrame, np.ndarray): Data the pipeline was trained on of shape [n_samples_train, n_feautures].
+
+        Raises:
+            ValueError: If pipeline is not trained.
+
+        Returns:
+            pd.Series: Datetime periods out to `forecast_horizon + gap`.
+
+        """
+        if not self._is_fitted:
```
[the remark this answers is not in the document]
christopherbunn: We need the pipeline to be fitted so we can grab the frequency that was found during the fit process.

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -88,3 +93,57 @@ def score(self, X, y, objectives, X_train=None, y_train=None):
...
+            raise ValueError("Pipeline must be fitted before getting forecast.")
+
+        X = infer_feature_types(X)
+
+        # Generate prediction periods
+        first_date = X.iloc[-1][self.time_index]
+        predicted_date_range = pd.Series(
+            pd.date_range(
+                start=first_date,
+                periods=self.forecast_horizon
```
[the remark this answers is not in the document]
christopherbunn: We would have to calculate out the end date, which would involve more code. IMO this looks a bit cleaner even if it isn't immediately obvious what periods means.

evalml/tests/pipeline_tests/test_time_series_baseline_pipeline.py
```diff
@@ -73,3 +74,52 @@ def test_time_series_baseline(
...
+        match="Pipeline must be fitted before getting forecast.",
+    ):
+        clf.get_forecast_periods(X)
+
+    clf.fit(X, y)
+    result = clf.get_forecast_periods(X)
+
+    assert result.size == forecast_horizon + gap
+    assert all(result.index == range(len(X), len(X) + forecast_horizon + gap))
+    assert result.iloc[0] == X.iloc[-1]["date"] + np.timedelta64(1, "D")
```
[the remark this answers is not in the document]
christopherbunn: Updated to use `clf.frequency`

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -88,3 +93,71 @@ def score(self, X, y, objectives, X_train=None, y_train=None):
...
+            >>> X = X = pd.DataFrame({'date': pd.date_range(start='1-1-2022', periods=10, freq='D'), 'feature': range(10, 20)})
+            >>> y = pd.Series(range(0, 10), name='target')
+            >>> gap = 1
+            >>> forecast_horizon = 2
+            >>> pipeline = TimeSeriesRegressionPipeline(component_graph=["Linear Regressor"],
+            ...                                         parameters={"Linear Regressor": {"normalize": True},
+            ...                                                     "pipeline": {"gap": gap, "max_delay": 1, "forecast_horizon": forecast_horizon, "time_index": "date"}},
+            ...                                        )
+            >>> pipeline.fit(X, y)
+            pipeline = TimeSeriesRegressionPipeline(component_graph={'Linear Regressor': ['Linear Regressor', 'X', 'y']}, parameters={'Linear Regressor':{'fit_intercept': True, 'normalize': True, 'n_jobs': -1}, 'pipeline':{'gap': 1, 'max_delay': 1, 'forecast_horizon': 2, 'time_index': 'date'}}, random_seed=0)
```
eccabay, line 119: I don't think we need this line

eccabay (approved):
Looks great, thanks Chris!

evalml/pipelines/time_series_regression_pipeline.py
chukarsten: Why is this so important to have an integer index?

chukarsten (approved):

PR #3753, opened 2022-10-13
Partial Dependence Fast Mode

tamargrey:
Optimize partial dependence by avoiding the repeated transformations that come from running `pipeline.predict(X)` on the entire dataset for every grid value. This implementation transforms only the features whose partial dependence we are calculating by running `pipeline.transform_all_but_final(X[[single_col]])`, updating the transformed full X, and running `estimator.predict` on the updated transformed X. 

Has the following limitations:
- Fast mode cannot be used if the pipeline contains components that rely on more than one column to fit and transform data correctly.
  - If there is a way of handling such components for fast mode, it will exist in a `_handle_partial_dependence_fast_mode` method on the component itself.
  - Known components with this issue that have not been handled are `Oversampler`, `StackedEnsembleRegressor`, and `StackedEnsembleClassifier`. They are blocked with a `_can_be_used_for_fast_partial_dependence = False` property.
  - Known components that have handed this in some way are `DFSTransformer`, `RFClassifierSelectFromModel`, and `RFRegressorSelectFromModel`. 
- Requires that training X and y are passed in in order to properly refit cloned pipelines with the correct data. Spoke to @eccabay about how `X_train` is known to be needed in partial dependence parameters for future time series work. `y_train` is known to be needed for the oversampler, which is currently blocked, so we could in theory remove it for now, but I think keeping it is a good failsafe for any components that might use for fit and transform that we missed here.

Note that listed above are only the _known_ components with problems in fast mode. This implementation provides some great performance improvements, but there are risks, as it is not hard to accidentally add or change a component in such a way as to make it not compatible with fast mode. Therefore, it is an opt-in API.

### DIFF
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -13,6 +13,15 @@ def _get_cloned_feature_pipelines(
     pipeline,
     variable_has_features_passed_to_estimator,
 ):
+    # --> long term: put flag on the object
+
+    for component in pipeline.component_graph.component_instances.values():
+        X, pipeline_params = component._handle_partial_dependence_fast_mode(
+            X,
+            pipeline.parameters,
+        )
+        pass
+
     # --> add docstring - and maybe typehinting?
     # mock out y for pipeline fitting
     len_X = len(X)
@@ -30,7 +39,9 @@ def _get_cloned_feature_pipelines(
     if dfs_transformer is not None:
         dfs_features = dfs_transformer["features"]
         X_cols = set(X.columns)
-        if features is None or any(f.get_name() not in X_cols for f in dfs_features):
+        if dfs_features is None or any(
+            f.get_name() not in X_cols for f in dfs_features
+        ):
             # set_trace()
             raise ValueError(
                 "Cannot use fast mode with DFS Transformer when features are unspecified or not all present in X.",
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -13,10 +13,10 @@ def _get_cloned_feature_pipelines(
     """Clones and fits pipelines for partial dependence fast mode.
 
     Args:
-        pipeline (PipelineBase or subclass): Fitted pipeline that will be cloned.
-        features (string, tuple[int or string]): The target feature for which to create the partial dependence plot for.
+        features (list(str)): The target feature for which to create the partial dependence plot for.
             If features is a string, it must be a valid column name in X.
             If features is a tuple of int/strings, it must contain valid column integers/names in X.
+        pipeline (PipelineBase or subclass): Fitted pipeline that will be cloned.
         variable_has_features_passed_to_estimator (dict[str, bool]): Dictionary mapping variable names to whether or not
             the variable has an impact on predictions. Variables that do not have features passed to the estimator
             will not have an impact on predictions and therefore do not need to have pipelines fitted for them.
@@ -26,7 +26,7 @@ def _get_cloned_feature_pipelines(
             be used to train the cloned pipelines.
 
     Returns:
-        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
+        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline
             fit for it.
     """
     if X_train is None or y_train is None:
@@ -68,15 +68,15 @@ def _transform_single_feature(
     """Transforms single column using cloned pipeline and column that has been updated for partial dependence calculations.
 
     Args:
-        X_t (pd.DataFrame, np.ndarray): The transformed data into which we insert the transformed single column.
         X (pd.DataFrame, np.ndarray): The original data used for calculating partial dependence.
+        X_t (pd.DataFrame, np.ndarray): The transformed data into which we insert the transformed single column.
         feature_provenance (dict[str, set[str]]): Dictionary mapping base feature names to engineered feature names.
         variable (str): The name of the single column we're transforming.
         part_dep_column (pd.Series): The updated column that will be transformed by the fit pipeline.
         cloned_pipeline (PipelineBase or subclass): Fitted pipeline that will be used to transform a single feature.
 
     Returns:
-        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
+        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline
             fit for it.
     """
     changed_col_df = pd.DataFrame({variable: part_dep_column})
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -29,9 +29,6 @@ def _get_cloned_feature_pipelines(
         dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline
             fit for it.
     """
-    if X_train is None or y_train is None:
-        raise ValueError("Training data is required for partial dependence fast mode.")
-
     X_train = infer_feature_types(X_train)
 
     # Make sure that only components that are capable of handling fast mode are in the pipeline
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -7,8 +7,8 @@ def _get_cloned_feature_pipelines(
     features,
     pipeline,
     variable_has_features_passed_to_estimator,
-    X_training,
-    y_training,
+    X_train,
+    y_train,
 ):
     """Clones and fits pipelines for partial dependence fast mode.
 
@@ -17,25 +17,25 @@ def _get_cloned_feature_pipelines(
         features (string, tuple[int or string]): The target feature for which to create the partial dependence plot for.
             If features is a string, it must be a valid column name in X.
             If features is a tuple of int/strings, it must contain valid column integers/names in X.
-        X_training (pd.DataFrame, np.ndarray): The data that was used to train the original pipeline. Will
+        X_train (pd.DataFrame, np.ndarray): The data that was used to train the original pipeline. Will
             be used to train the cloned pipelines.
-        y_training (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
+        y_train (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
             be used to train the cloned pipelines.
 
     Returns:
         dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
             fit for it.
     """
-    if X_training is None or y_training is None:
+    if X_train is None or y_train is None:
         raise ValueError("Training data is required for partial dependence fast mode.")
 
-    X_training = infer_feature_types(X_training)
+    X_train = infer_feature_types(X_train)
 
     # Make sure that only components that are capable of handling fast mode are in the pipeline
     new_parameters = pipeline.parameters
     for component in pipeline.component_graph.component_instances.values():
         new_parameters = component._handle_partial_dependence_fast_mode(
-            X_training,
+            X_train,
             new_parameters,
         )
 
@@ -48,7 +48,7 @@ def _get_cloned_feature_pipelines(
         pipeline_copy = pipeline.new(
             parameters=new_parameters,
         )
-        pipeline_copy.fit(X_training.ww[[variable]], y_training)
+        pipeline_copy.fit(X_train.ww[[variable]], y_train)
         cloned_feature_pipelines[variable] = pipeline_copy
 
     return cloned_feature_pipelines
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -13,6 +13,19 @@ def _get_cloned_feature_pipelines(
     pipeline,
     variable_has_features_passed_to_estimator,
 ):
+    """Clones and fits pipelines for partial dependence fast mode.
+
+    Args:
+        pipeline (PipelineBase or subclass): Fitted pipeline that will be cloned.
+        X (pd.DataFrame, np.ndarray): The input data used to fit the cloned pipelines.
+        features (string, tuple[int or string]): The target feature for which to create the partial dependence plot for.
+            If features is a string, it must be a valid column name in X.
+            If features is a tuple of int/strings, it must contain valid column integers/names in X.
+
+    Returns:
+        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
+            fit for it.
+    """
     # Make sure that only components that are capable of handling fast mode are in the pipeline
     new_parameters = pipeline.parameters
     for component in pipeline.component_graph.component_instances.values():
@@ -21,7 +34,6 @@ def _get_cloned_feature_pipelines(
             new_parameters,
         )
 
-    # --> add docstring - and maybe typehinting?
     # mock out y for pipeline fitting
     len_X = len(X)
     if is_regression(pipeline.problem_type):
@@ -56,6 +68,21 @@ def _transform_single_feature(
     part_dep_column,
     cloned_pipeline,
 ):
+    """Transforms single column using cloned pipeline and column that has been updated
+        for partial dependence calculations.
+
+    Args:
+        X_t (pd.DataFrame, np.ndarray): The transformed data into which we insert the transformed single column.
+        X (pd.DataFrame, np.ndarray): The original data used for calculating partial dependence.
+        feature_provenance (dict[str, set[str]]): Dictionary mapping base feature names to engineered feature names.
+        variable (str): The name of the single column we're transforming.
+        part_dep_column (pd.Series): The updated column that will be transformed by the fit pipeline.
+        cloned_pipeline (PipelineBase or subclass): Fitted pipeline that will be used to transform a single feature.
+
+    Returns:
+        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
+            fit for it.
+    """
     changed_col_df = pd.DataFrame({variable: part_dep_column})
     changed_col_df.ww.init(
         logical_types={variable: X.ww.logical_types[variable]},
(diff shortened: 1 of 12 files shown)


evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        cols_to_replace = [
+            col for col in X_t if col in feature_provenance.get(variable)
+        ]
+    # If some categories got dropped during transform of the original X_t,
+    # they won't be in X_t_single_col, so don't include them
+    if len(cols_to_replace) != len(X_t_single_col.columns):
+        X_t_single_col = X_t_single_col[cols_to_replace]
+
+    # --> not keeping in woodwork - problematic?
+    X_t[cols_to_replace] = X_t_single_col
```
tamargrey, line 104: I wasn't sure if I should be adding the columns into X_t though woodwork or not. This X_t gets passed to `estimator.predict`, so I would think that it'd be problematic if the woodwork schema was invalid. What I _think_ is happening here to allow this to work is that the `cols_to_replace` are the correct types to not break the woodwork schema. 

But maybe I should just go through woodwork to be safe. Open to hearing other peoples' thoughts!

evalml/pipelines/components/component_base.py
```diff
@@ -20,6 +20,10 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
     """
 
     _default_parameters = None
+    # --> this won't make custom components invalid for fast mode, but lets us opt into it.
+    # Given that fast mode isn't default behavior and we'll have proper warnings about not
+    # using it for custom components, (maybe we raise a warning if we find custom?), this makes the most sense
+    _can_be_used_for_fast_partial_dependence = True
```
tamargrey, line 26: I just wanted to highlight this behavior: Any component defaults to being valid for fast mode, so if a user created a custom component or a developer added a new component, it's possible they might get incorrect PD results for pipelines containing it. 

I added a test in `test_components.py` that will raise an error when a new component is added to evalml, forcing developers to update the test and hopefully spend a little time thinking about if their component should be used for fast mode. It should be very quick to update, and devs can just default to disallowing it if they want. If people think that's a pain in the butt, I can remove it. 

There's nothing currently in place to stop custom components, but I was considering adding a warning using the `all_components` util to let users know the results may be wrong if a custom component is detected. Does anyone feel strongly about whether this warning should exist or not?

evalml/pipelines/pipeline_base.py
```diff
@@ -667,6 +667,7 @@ def new(self, parameters, random_seed=0):
         Returns:
             A new instance of this pipeline with identical components.
         """
+        # --> why is there no setting of the threshold for binary like there is in clone?
         return self.__class__(
```
tamargrey, line 671: Not really relevant to the rest of the PR, but I noticed that the following check that's in `clone` isn't in this method and wanted to see if anyone knew why:

```python
        if is_binary(self.problem_type):
            clone.threshold = self.threshold
```

evalml/model_understanding/_partial_dependence_utils.py
```diff
@@ -221,45 +225,105 @@ def _cartesian(arrays):
...
 
     if is_regression(pipeline.problem_type):
-        prediction_method = pipeline.predict
+        prediction_method = prediction_object.predict
     else:
-        prediction_method = pipeline.predict_proba
+        prediction_method = prediction_object.predict_proba
+    if fast_mode and no_features_passed_to_estimator:
+        original_predictions = prediction_method(X_eval)
+        original_predictions_mean = np.mean(original_predictions, axis=0)
```
tamargrey, line 294: In theory we could do this for slow mode too-- it would require us to calculate `pipeline.transform_all_but_final(X_eval)`, though, and I didn't do a full set of analysis on whether that ends up saving time in many cases, so I decided to not add it in for now. I created [an issue](https://github.com/alteryx/evalml/issues/3807) to look into adding it at a separate time.

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -0,0 +1,100 @@
...
+    """Clones and fits pipelines for partial dependence fast mode.
+
+    Args:
+        pipeline (PipelineBase or subclass): Fitted pipeline that will be cloned.
+        features (string, tuple[int or string]): The target feature for which to create the partial dependence plot for.
+            If features is a string, it must be a valid column name in X.
+            If features is a tuple of int/strings, it must contain valid column integers/names in X.
+        X_training (pd.DataFrame, np.ndarray): The data that was used to train the original pipeline. Will
+            be used to train the cloned pipelines.
+        y_training (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
```
jeremyliweishih, line 22: missing the docstring for `variable_has_features_passed_to_estimator`

evalml/pipelines/components/component_base.py
```diff
@@ -102,6 +103,21 @@ def default_parameters(cls):
...
+    def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters):
+        """Determines whether or not a component can be used with partial dependence's fast mode.
+
+        Args:
+            X (pd.DataFrame): Holdout data being used for partial dependence calculations.
+            pipeline_parameters (dict): Pipeline parameters that will be used to create the pipelines
+                used in partial dependence fast mode.
+        """
+        if self._can_be_used_for_fast_partial_dependence:
+            return pipeline_parameters
```
jeremyliweishih, line 115: do we need `pipeline_parameters` here? It seems like we can just check `_can_be_used_for_fast_partial_dependence` in `_can_be_used_for_fast_partial_dependence` and raise there error there as well. 

```
for component in pipeline.component_graph.component_instances.values():
        if not component._can_be_used_for_fast_partial_dependence:
                   raise error
```

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        new_parameters = component._handle_partial_dependence_fast_mode(
+            X_training,
+            new_parameters,
+        )
+
+    # Create a fit pipeline for each feature
+    cloned_feature_pipelines = {}
+    for variable in features:
+        # Don't fit pipelines if the feature has no impact on predictions
+        if not variable_has_features_passed_to_estimator[variable]:
```
jeremyliweishih, line 46: can you give me an example of when this happens in practice? Can't think of one off the top of my head!

evalml/pipelines/components/component_base.py
jeremyliweishih: I think its ok for now since fast mode is off by default but we should file an issue to track it!

evalml/pipelines/components/component_base.py
jeremyliweishih: nvm I see what you're doing here!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
```diff
@@ -70,3 +70,23 @@ def fit_transform(self, X, y=None):
             pd.DataFrame: Transformed data.
         """
         return self.fit(X, y).transform(X, y)
+
+    def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters):
```
jeremyliweishih, line 74: if the feature selector decided to drop a feature - should we be calculating partial dependence for it? I'm not sure what the current behavior is but I feel like we shouldn't be if the feature selector deemed the feature unnecessary.  Don't think we need to solve in this PR but might be worth putting a issue up to discuss!

jeremyliweishih: I saw your test case! So does this mean features dropped by the feature selector show no impact on predictions? I'm a little confused how the code chunk below affects it.

jeremyliweishih (approved):
LGTM - just some nits and questions but overall looks great! Is there an issue add to our docs? If not we should create one so we can tell users to use fast mode.

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
tamargrey: If the `RFRegressorSelectFromModel` decides that the same feature you're trying to calculate PD for is unimportant and should be dropped prior to passing the data to the estimator. 

It's what lets us just use the original data's average predictions as PD values. We know we won't have to change that feature's values and re-transform, so there's no need to waste time fitting a cloned pipeline for that feature!

jeremyliweishih: gotcha - I think I understand how everything is linked now. Thanks!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
tamargrey: (forgive me if this is now unnecessary after my comment above!)

We can use the information from the original pipeline to determine if the feature gets dropped or not, and if it doesn't get dropped, this code lets us keep it. It basically turns off the Feature Selector for the single column-fitted pipelines. We say "take 100% of features" and "take features above the importance threshold of 0", so that we don't drop the single feature we're fitting that cloned pipeline with! 

It's definitely a bit confusing to look at completely out of context, though. Let me know if you think the docstring isn't descriptive enough.

As to your other point: I don't know that it should be up to the partial dependence function to say that it won't calculate partial dependence for you if it deems the feature to not have an impact on predictions. For all we know, the partial dependence plot could be the way a user discovers this, and I could see it being useful to emit a warning or add something to the plot that indicates why this feature's PD is static. 

If a user wants to not calculate PD at all in this case, I say it's up to them to decide that, but maybe we should be making that information (whether or not a feature gets passed to the estimator at all) more readily available. I can make an issue to think more about the concept of knowing that a feature won't have an impact on predictions and how that information can be accessible to users.

jeremyliweishih: no this is an amazing explanation! it might be good to provide a little more context in the docstring 😄

evalml/model_understanding/_partial_dependence_utils.py
```diff
@@ -291,6 +355,9 @@ def _partial_dependence(
     grid_resolution=100,
     kind="average",
     custom_range=None,
+    fast_mode=False,
+    X_training=None,
```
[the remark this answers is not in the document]
tamargrey: Now seeing that the only usage of `X_training` in the whole repo is from this PR 🤦‍♀️ Changing!

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
tamargrey: Thanks for catching--adding!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
tamargrey: Added a `Return` section to the docstring that hopefully explains this a little--let me know if it makes sense to you!

evalml/pipelines/components/component_base.py
tamargrey: Created an issue: https://github.com/alteryx/evalml/issues/3817

While thinking about this more, it occurred to me that we could add a check in fast mode itself as to whether the pipeline relies on multiple columns. I think that check will be a little tricky to implement correctly, and we should test the impact on fast mode's runtime, so we shouldn't do that here, but I outlined the idea in the issue!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
tamargrey: Created an issue for adding a feature to show whether a feature will be passed to the estimator: https://github.com/alteryx/evalml/issues/3818

evalml/model_understanding/_partial_dependence_utils.py
```diff
@@ -307,6 +374,16 @@ def _partial_dependence(
...
+        fast_mode (bool, optional): Whether or not performance optimizations should be
+            used. Defaults to False. When True, copies of pipelines will be used to transform just the
+            column(s) we're calculating partial dependence for. This means that any pipeline
+            containing a component that relies on multiple columns for fit and transform should
+            not be used. See the ``_can_be_used_for_fast_partial_dependence`` property on components
+            to determine which components cannot be used for fast mode.
+        X_train (pd.DataFrame, np.ndarray): The data that was used to train the original pipeline. Will
+            be used in fast mode to train the cloned pipelines.
+        y_train (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
+            be used in fast mode to train the cloned pipelines.
```
eccabay, line 386: Nit: another case where the docstring order doesn't match the argument order! It'd also be nice to add that X_train and y_train default to None here.

evalml/model_understanding/partial_dependence_functions.py
```diff
@@ -57,6 +60,15 @@ def partial_dependence(
...
             the grid but averaged over all of the samples in X.
+        fast_mode (bool, optional): Whether or not performance optimizations should be
+            used for partial dependence calculations. Defaults to False.
+            Note that user-specified components may not produce correct partial dependence results, so fast mode
+            should only be used with EvalML-native components. Additionally, some components are not compatible
+            with fast mode; in those cases, an error will be raised indicating that fast mode should not be used.
+        X_train (pd.DataFrame, np.ndarray): The data that was used to train the original pipeline. Will
+            be used in fast mode to train the cloned pipelines.
+        y_train (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
+            be used in fast mode to train the cloned pipelines.
```
eccabay, line 71: Same comment about defaulting to None here!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
```diff
@@ -70,3 +70,29 @@ def fit_transform(self, X, y=None):
...
+        """Updates pipeline parameters to not drop any features based off of feature importance.
+
+        Note:
```
eccabay, line 77: A style mega nit is that we don't need this in its own section. We can mirror other parts of the docstring here by just keeping it as its own unmarked separate paragraph.

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
```diff
@@ -70,3 +70,29 @@ def fit_transform(self, X, y=None):
...
+        Args:
+            X (pd.DataFrame): Holdout data being used for partial dependence calculations.
+            pipeline_parameters (dict): Pipeline parameters that will be used to create the pipelines
+                used in partial dependence fast mode.
+
+        Return:
+            pipeline_parameters (dict): Pipeline parameters updated to allow the FeatureSelector component
+                to not drop any features. This is needed because the cloned pipeline that uses these features
+                will be fit on a single column, and we do not want to drop that column. We can use the original
+                pipeline to determine if that feature gets dropped or not.
```
eccabay, line 91: It feels odd and/or slightly redundant to mention the "this is needed" within the return statement. It might fit better and be easier to reference in the note section instead?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -1938,3 +2457,344 @@ def test_partial_dependence_jupyter_graph_check(
...
+        ]
+    pipeline = _make_stacked_ensemble_pipeline(
+        input_pipelines=input_pipelines,
+        problem_type=problem_type,
+    )
+    pipeline.fit(X, y)
+
+    error = re.escape(
+        f"cannot run partial dependence fast mode",
+    )
```
eccabay, line 2721: This might be something w.r.t regex I'm unfamiliar with, but why do we need `re.escape` here? Does keeping this as a regular string fail?

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            be used to train the cloned pipelines.
+        y_train (pd.Series, np.ndarray): The target data that was used to train the original pipeline. Will
+            be used to train the cloned pipelines.
+
+    Returns:
+        dict[str, PipelineBase or subclass]: Dictionary mapping feature name to the pipeline pipeline
+            fit for it.
+    """
+    if X_train is None or y_train is None:
+        raise ValueError("Training data is required for partial dependence fast mode.")
```
eccabay, line 33: This is purely my opinion, but it may be neater to raise this error higher up in the call stack - potentially even in the highest level `partial_dependence` function. Just as fewer things to wade through when debugging an error.

eccabay (approved):
This looks solid to me! Most of my comments are nitpicks related to docstrings. Awesome work!

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
tamargrey: @eccabay Do you mean like this?
image

evalml/pipelines/components/transformers/feature_selection/feature_selector.py
tamargrey: Good call -- removed that section here and added another sentence above, so the note section now says:     

>  This is needed, because fast mode refits cloned pipelines on single columns,
>             so categorical columns that have one-hot encoding applied may lose some
>             of their encoded columns with the default parameters. Therefore, we update the
>             parameters here to not drop any columns, and use the original
>             pipeline to determine if that feature gets dropped or not.

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    new_parameters = pipeline.parameters
+    for component in pipeline.component_graph.component_instances.values():
+        new_parameters = component._handle_partial_dependence_fast_mode(
+            X_train,
+            new_parameters,
+        )
+
+    # Create a fit pipeline for each feature
+    cloned_feature_pipelines = {}
+    for variable in features:
```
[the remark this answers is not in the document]
tamargrey: Very good call. I also think it should just be `(list(str))`, because it's getting passed in from `_partial_dependence_calculation`, which already says it'll just be a list of strings at that point.

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: Nope, I think this was just an accidental holdover from when I tried to include the full error message with the component name and the extra quotation marks and whatnot weren't playing nice. I'll remove!

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
tamargrey: I think that makes sense--moved up to `partial_dependence` with all the other error logic

PR #3766, opened 2022-10-20
Prophet prediction intervals

ParthivNaresh:


ParthivNaresh:
> I suspect we will want to have a function perhaps at the `Estimator` level that defines a baseline behavior for `get_prediction_intervals()`

@chukarsten Sounds good, I was going to add that in a separate PR so I could put up all 3 PRs for review instead of adding it to each one, but we can cover Prophet first and move forward!

PR #3773, opened 2022-10-24
Only select lags within FE window dictated by FH and gap

### LINKED ISSUE #3770: TimeSeriesFeaturizer doesn't appear to use correct significant lags in most cases
The lagged features that get generated by the `TimeSeriesFeaturizer` do not use the significant lags found by `_find_significant_lags` in most cases. Changing the forecast horizon and/or the gap parameters change the lagged features that are created, but these parameters are not directly used when identifying which lags are significant, only the max_delay is used.

The current process is something like this, if I am following the code correctly:
- Find all significant lags in the range of (0, max_delay)
- Add the start delay (which is gap + forecast_horizon) to the lags found from the previous step

This means that the significant lags that were identified can be offset, making them insignificant.

I believe the proper approach should be more like this:
- Use the gap and forecast_horizon and max_delay to identify the rows that fall within the feature engineering window
- Identify significant lags that fall within the feature engineering window
- Create lagged features from those significant lags without further modification

The problem is illustrated in the example below. Changing the forecast horizon changes the lagged features that are created.

**Forecast Horizon of 1**

```python
import pandas as pd
import featuretools as ft

df = pd.read_csv("DailyDelhiClimateTrain.csv", nrows=200)

max_delay = 10
gap = 0
forecast_horizon = 1

component_graph_as_list = ["Time Series Featurizer"]
parameters = {
    "pipeline": {
        "time_index": "date",
        "gap": gap,
        "max_delay": max_delay,
        "forecast_horizon": forecast_horizon,
    },
    "Time Series Featurizer": {
        "time_index": "date",
        "gap": gap,
        "max_delay": max_delay,
        "forecast_horizon": forecast_horizon,
    }
}
pipeline = TimeSeriesRegressionPipeline(component_graph=component_graph_as_list, parameters=parameters)

y = df.pop("meantemp")
X = df[["date"]]

p = pipeline.fit(X, y)

# For simplicity assume the first and fourth lags were found to be significant (zero indexed)
p.component_graph["Time Series Featurizer"].statistically_significant_lags = [0, 3]

p.transform(X, y).head()
```

For this first case, we create target delay features of 1 period and 4 periods as we expect
```
        date  target_delay_1  target_delay_4  target_rolling_mean
0 2013-01-01             NaN             NaN                  NaN
1 2013-01-02       10.000000             NaN                  NaN
2 2013-01-03        7.400000             NaN                  NaN
3 2013-01-04        7.166667             NaN             8.188889
4 2013-01-05        8.666667            10.0             7.744444
```

If we change nothing but making `forecast_horizon = 2` and run this again, we get a different result. We now create delay features of 2 period and 5 periods, neither of which were identified as significant. It seems to me that changing the number of points we want to forecast in the future should have no bearing on which lag features are created, only on which lags we can use.

For this case, we could not use the lag of 1 feature, because we would have no data available for the second forecast point (the feature engineering window doesn't include the previous row in this case). We should still be able to use the lag of 4 feature, which was significant.

**Forecast Horizon of 2**
```
        date  target_delay_2  target_delay_5  target_rolling_mean
0 2013-01-01             NaN             NaN                  NaN
1 2013-01-02             NaN             NaN                  NaN
2 2013-01-03       10.000000             NaN                  NaN
3 2013-01-04        7.400000             NaN                  NaN
4 2013-01-05        7.166667             NaN             8.188889
```

jeremyliweishih:
Fixes #3770


evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -150,8 +151,8 @@ def _encode_X_while_preserving_index(X_categorical):
         )
 
     @staticmethod
-    def _find_significant_lags(y, conf_level, max_delay):
-        all_lags = np.arange(max_delay + 1)
+    def _find_significant_lags(y, conf_level, start_delay, max_delay):
+        all_lags = np.arange(start_delay, start_delay + max_delay + 1)
```
jeremyliweishih, line 155: only filter by lags that start in the feature engineering window.

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -173,7 +174,9 @@ def _find_significant_lags(y, conf_level, max_delay):
                 set(index[significant]).intersection(peaks).union(first_significant_10)
             )
             # If no lags are significant get the first lag
-            significant_lags = sorted(significant_lags.intersection(all_lags)) or [1]
+            significant_lags = sorted(significant_lags.intersection(all_lags)) or [
+                start_delay,
```
jeremyliweishih, line 178: default to using the start of the feature engineering window.

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -246,7 +249,7 @@ def _compute_delays(self, X_ww, y):
                     feature_name = f"{col_name}_delay_{self.start_delay + t}"
                     lagged_features[
                         f"{col_name}_delay_{self.start_delay + t}"
-                    ] = col.shift(self.start_delay + t)
+                    ] = col.shift(t)
```
jeremyliweishih, line 252: lags already have the start delay baked in since we're only selecting lags that start from the feature engineering window.

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -256,7 +259,7 @@ def _compute_delays(self, X_ww, y):
             for t in self.statistically_significant_lags:
                 lagged_features[
                     self.target_colname_prefix.format(t + self.start_delay)
-                ] = y.shift(self.start_delay + t)
+                ] = y.shift(t)
```
jeremyliweishih, line 262: same here.

ParthivNaresh: If the `start_delay` is 11 and the significant lag is 11, then this comes out to a format of 21 (because 11 + 11 - 1). So we'd shifting by 11 and the feature would be called `target_delay_21`? Shouldn't the `shift` and the column name suffix come out to the same number?

thehomebrewnerd (approved):
Did some testing with this new function and the results I got seemed to always fall within the bounds of the feature engineering window. Looks good to me!

evalml/tests/component_tests/test_time_series_featurizer.py
```diff
@@ -252,7 +252,7 @@ def test_delayed_feature_extractor_maxdelay5_forecasthorizon1_gap0(
     f"evalml.pipelines.TimeSeriesFeaturizer.{ROLLING_TRANSFORM_METHOD_NAME}",
     return_value=pd.DataFrame(),
 )
-def test_delayed_feature_extractor_maxdelay3_forecasthorizon7_gap1(
+def test_delayed_feature_extractor_maxdelay2_forecasthorizon7_gap1(
```
eccabay, line 255: What was the reasoning behind changing this test?

eccabay (approved):

evalml/tests/component_tests/test_time_series_featurizer.py
jeremyliweishih: IIRC just to match the lags on the dataset

PR #3774, opened 2022-10-24
Fix output in dependency check

jeremyliweishih:


chukarsten (approved):

eccabay (approved):

chukarsten (approved):

jeremyliweishih:
Issue here is potentially that by upgrading the min FT version to the latest our dask versions are incompatible because FT wants `dask[dataframe]==2022.2.0` and we try to install from `dask >= 2021.10.0`

.github/meta.yaml
```diff
@@ -38,7 +38,7 @@ outputs:
         - shap >=0.40.0
         - texttable >=1.6.2
         - woodwork >=0.19.0
-        - featuretools>=1.7.0
+        - featuretools>=1.16.0
```
thehomebrewnerd, line 41: @jeremyliweishih  I don't think you should have to make this change. I am looking into it in a separate PR to confirm.

jeremyliweishih:
Closing I favor of #3778

PR #3778, opened 2022-10-25
Fix nlp imports

thehomebrewnerd:
This PR makes a few updates for compatibility with the latest versions of Featuretools and nlp-primitives:
- Change nlp primitive import statements
- Remove upper bound version restriction on Featuretools and nlp-primitives
- Bump nlp-primitives lowest allowed version to allow all primitives to be imported using `from featuretools.primitives import ...` instead of `from nlp_primitives import ...`


gsheni (approved):

jeremyliweishih (approved):
Thanks!

PR #3781, opened 2022-10-25
Integrate Decomposition into AutoMLSearch

eccabay:
Includes some other small code changes to allow integration to be successful!


evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -116,6 +117,16 @@ def _add_training_data_to_X_Y(self, X, y, X_train, y_train):
                 self.gap,
             )
 
+            # Properly fill in the dates in the gap
```
eccabay, line 120: This was a pre-existing bug where we fill in the gap between the training and testing data with the last row of training data, but that means creating an irregularly spaced time index.

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -135,9 +146,15 @@ def _add_training_data_to_X_Y(self, X, y, X_train, y_train):
...
-    def _drop_time_index(self, X):
+    def _drop_time_index(self, X, y):
         """Helper method to drop the time index column from the data if DateTime Featurizer is not present."""
         if self.should_drop_time_index and self.time_index in X.columns:
+            X_schema = X.ww.schema
+            y_schema = y.ww.schema
+            X = X.set_index(X[self.time_index])
+            y = y.set_axis(X[self.time_index])
+            X.ww.init(schema=X_schema)
+            y.ww.init(schema=y_schema)
```
eccabay, line 157: Feels weird to be saving the time index in the data's indices when we're trying to drop it, but the decomposer needs the time index data to decompose successfully. If anyone has suggestions on a better or cleaner way to get around this I am very open to it.

jeremyliweishih: Can we drop after fit or transform in the decomposer then? The problem is non-time-series native estimators will fail with a datetime column. Maybe we can amend the logic here to save the time index only when having the decomposer in the pipeline and then dropping the time index later?

jeremyliweishih (changes requested):
I think we need to figure out when and where we're dropping the time index so that the pipelines don't fail for estimators that can't accept time indexes

jeremyliweishih:
can we also add `test_automl_supports_time_series_regression` to make sure that the decomposer and its parameters are correct?

jeremyliweishih (approved):
nvm - we're good but think we should add to `test_automl_supports_time_series_regression` still!

eccabay:
Closing in favor of #3785

PR #3785, opened 2022-10-26
Integration of STL into EvalML

chukarsten:
Rebasing of Beccas #3781

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import logging
 
 import pandas as pd
-from pandas.core.index import Int64Index
+from pandas.core.index import Int64Index, RangeIndex
 from statsmodels.tsa.arima.model import ARIMA
 from statsmodels.tsa.forecasting.stl import STLForecast
 from statsmodels.tsa.seasonal import STL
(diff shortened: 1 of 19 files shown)


evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -83,7 +83,7 @@ def _project_trend(self, y):
                 )
                 - 1
             )
-        elif isinstance(y.index, Int64Index):
+        elif isinstance(y.index, Int64Index) or isinstance(y.index, RangeIndex):
```
chukarsten, line 86:
```suggestion
        elif isinstance(y.index, (RangeIndex, Int64Index)):
```

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -185,6 +188,59 @@ def test_make_pipeline(
...
+    }
+    estimators = get_estimators(problem_type=problem_type)
+    pipeline_class = _get_pipeline_base_class(problem_type)
+    for estimator_class in estimators:
+        if problem_type in estimator_class.supported_problem_types:
+
+            pipeline = make_pipeline(X, y, estimator_class, problem_type, parameters)
+            assert isinstance(pipeline, pipeline_class)
+
+            if is_regression(problem_type) and frequency in ["D", "MS"]:
```
chukarsten, line 227: I thought we were catching "A" and "MS"?  If that's the case, I would think we might want to capture the list of frequencies as a module level list in the STLDecomposer (imported) and reference it here?  We can file that as a separate issue though.

eccabay: Annual and ten minute intervals are the frequencies we've found to fail at this point. MS seems to fail with the polynomial decomposer with the millisecond issue but does fine with STL. I do have the failing frequencies in a constant within utils, I'll reference that here!

eccabay (approved):
Feels silly to approve a PR for my own work but here we are

PR #3794, opened 2022-10-28
Update _get_decomposer to use stripped freq

eccabay:
Fixes bug where any invalid frequencies that were either offset (i.e. 10 minutes -> `10T`) or anchored (annually at the year end -> `A-DEC`) were not properly getting excluded from training with the decomposer.

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -231,10 +231,7 @@ def _get_decomposer(X, y, problem_type, estimator_class, sampler_name=None):
     components = []
     if is_time_series(problem_type) and is_regression(problem_type):
         time_index = get_time_index(X, y, None)
-        if time_index.freq.is_anchored:
-            freq = time_index.freq.name.split("-")[0]
-        else:
-            freq = time_index.freq.name
+        freq = time_index.freq.name.split("-")[0]
         if freq[-1] not in _UNSUPPORTED_FREQUENCIES_STL_DECOMPOSER:
             components.append(STLDecomposer)
     return components
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -231,8 +231,11 @@ def _get_decomposer(X, y, problem_type, estimator_class, sampler_name=None):
     components = []
     if is_time_series(problem_type) and is_regression(problem_type):
         time_index = get_time_index(X, y, None)
-        freq = time_index.freq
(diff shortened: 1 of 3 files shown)


jeremyliweishih (approved):
thanks!

evalml/pipelines/utils.py
```diff
@@ -65,7 +65,7 @@
 from evalml.utils import get_time_index, infer_feature_types
 from evalml.utils.gen_utils import contains_all_ts_parameters
 
-_UNSUPPORTED_FREQUENCIES_STL_DECOMPOSER = ["A", "T"]
+_UNSUPPORTED_FREQUENCIES_STL_DECOMPOSER = ["A", "Y", "T", "AS", "YS"]
```
eccabay, line 68: For reference: A and Y are both year end, AS and YS are year start, T is minute

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -197,10 +196,22 @@ def test_make_pipeline(
...
+    [
+        ("D", True),
+        ("MS", True),
+        ("A", False),
+        ("T", False),
+        ("10T", False),
+        ("AS-JAN", False),
+        ("YS", False),
+    ],
+)
```
chukarsten, line 210: I suspect this is going to be on a per-decomposer basis...so we might need to do this differently going forward, when we introduce new decomposers and also introduce new ways of trying more pipelines with different decomposers.  I think it's fine for now and we can kick the can to when we do X11 or X13.

Could we also add mention that this is a test for time series only in the test name?

evalml/pipelines/utils.py
chukarsten, line 68: we might want to move this either 1.) into the stl_decomposer module and/or 2.) make it returnable by the decomposer itself.  Then we can dynamically know what decomposers support or do not.  You could also build the parametrization list and parametrize across STL decomposers in your decomposer test that way.

Part of the motivation behind the move, in my mind, is the answer to "where should I go to find what frequencies are not supported by X Decomposer?"

chukarsten (approved):
No blocking comments, but looks good.  I do think each decomposer module should be cognizant of what frequencies are no bueno for it.  And I'll leave it up to you whether you think the child-decomposer class has a method to return that.

evalml/pipelines/utils.py
eccabay: A lovely idea - filed #3795 to track this!

PR #3797, opened 2022-10-31
Pin dask below 2022.10.1

eccabay:
The most recent dask version seems to break a bunch of our unit tests. Pinning the version until we figure out what's up.

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -26,7 +26,7 @@ outputs:
       run:
         - numpy >=1.21.0
         - pandas >=1.4.0
-        - dask >=2021.10.0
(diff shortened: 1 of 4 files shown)


jeremyliweishih (approved):

bchen1116:
[This PR](https://github.com/dask/distributed/pull/7230) on dask should fix the tests but is not yet released.

eccabay:
Thanks for pointing that out @bchen1116! I'll file an issue to unpin dask once that PR is in, since we don't need dependency updates blocking us until then.

chukarsten (approved):

PR #3811, opened 2022-11-02
Decomposer valid frequency logic

eccabay:
Consolidates all the logic for whether a frequency will work with a given decomposer to the decomposer classes themselves.


jeremyliweishih (approved):

jeremyliweishih (approved):
very nice

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -95,6 +96,21 @@ def fit_transform(
...
+
+        Args:
+            freq (str): A frequency to validate. See the pandas docs at https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases for options.
+
+        Returns:
+            boolean representing whether the frequency is valid or not.
+        """
+        freq = freq.split("-")[0]
+        return (
+            freq[-1] not in self.invalid_frequencies if freq[0].isdigit() else True
```
fjlanasa, line 111: What would happen if the frequency was `2BQ`?

fjlanasa: i think something like this would let you extract the entire frequency string to check whether it is valid or not
```
match = re.match(r"(^\d+)?([A-Z]+)-?([A-Z]+)?", freq)
count, freq, anchor = match.groups()
```

christopherbunn (approved):
LGTM once edge case mentioned in Frank's comment is addressed.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
eccabay: `2BQ` drops the `2` before getting passed to this function so the test case passes as expected, but I'm not exactly sure why, so I'll update to this logic to be safe!

fjlanasa: hmm, i would have thought that it checked if `Q` is valid (you grab `freq[-1]`, which is not listed in the list of invalid frequencies.

eccabay: I thought so as well, but in stepping through the `2` being dropped in the call to this function (somehow?? it was there before) meant the `isdigit()` check was false so we checked membership of `BQ`

eccabay: Yup exactly - the first half of the `or` returned True but failed in the second check. I should have clarified earlier, the `2` was dropped I think when we isolated the frequency string from the `pd.DatetimeIndex` before calling this function, so the string being evaluated is `BQ` for the whole time here. We get the False result through the second half of the `or` statement.

Still not exactly clear on why that happened though, and it's certainly safer and easier to understand with the regex instead!

fjlanasa: ah i missed the `and freq not in self.invalid_frequencies` lol. sorry.

fjlanasa (approved):

PR #3829, opened 2022-11-09
Clean up decomposer index logic

eccabay:
Ensures we set y's time index as well as X during `drop_time_index`, and saves the frequency in `STLDecomposer` so we don't mis-assume the frequency when determining the seasonality offset in `inverse_transform`

### DIFF
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -154,7 +154,7 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
             y.ww.init(schema=y_schema)
             X.ww.set_index(self.time_index)
             X = X.ww.drop(self.time_index)
-        return X
+        return X, y
 
     def transform_all_but_final(self, X, y=None, X_train=None, y_train=None):
         """Transforms the data by applying all pre-processing components.
@@ -207,8 +207,8 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
             raise ValueError(
                 "Cannot call predict_in_sample() on a component graph because the final component is not an Estimator.",
             )
-        X = self._drop_time_index(X, y)
-        X_train = self._drop_time_index(X_train, y_train)
+        X, y = self._drop_time_index(X, y)
+        X_train, y_train = self._drop_time_index(X_train, y_train)
         target = infer_feature_types(y)
         features = self.transform_all_but_final(X, target, X_train, y_train)
         predictions = self._estimator_predict(features)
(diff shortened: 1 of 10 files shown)


evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1072,7 +1072,7 @@ def test_binary_predict_pipeline_objective_mismatch(
         ValueError,
         match="Objective Precision Micro is not defined for time series binary classification.",
     ):
-        binary_pipeline.predict(X[30:32], "precision micro", X[:30], y[:30])
```
fjlanasa, line 1075: What's this change for? Is it because we can't infer frequencies with only two values? Should we handle this somewhere?

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -149,19 +149,25 @@ def _add_training_data_to_X_Y(self, X, y, X_train, y_train):
...
+            )
+
+            y_schema = y.ww.schema
+            y = y.set_axis(time_index)
             y.ww.init(schema=y_schema)
+
             if X.ww.schema is not None:
-                X = X.ww.drop([self.time_index])
+                X.ww.set_index(self.time_index)
+                X = X.ww.drop(self.time_index)
```
jeremyliweishih, line 164: I thought setting the index dropped the column in woodwork as well? I might be remembering incorrectly though!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
eccabay: That's correct - pandas needs 3 or more values to infer a frequency. I'm not sure where we could handle this, though.

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -327,6 +327,66 @@ def test_decomposer_build_seasonal_signal(
...
+    test_first_index = {
+        "on period": 3 * period,
+        "before period": 3 * period - 1,
+        "just after period": 3 * period + 1,
+        "mid period": 3 * period + 4,
+    }[test_first_index]
+
+    X, _, y = ts_data()
+    datetime_index = pd.date_range(start="01-01-2002", periods=len(X), freq="M")
+    datetime_index.freq = None
```
jeremyliweishih, line 348: should we check the case where there is a `freqstr` as well?

evalml/tests/pipeline_tests/test_time_series_pipeline.py
fjlanasa: OK fair enough. How does this affect our ability to forecast out 1 unit? is it still possible?

jeremyliweishih (approved):
LGTM just a little suggestion!

evalml/tests/pipeline_tests/test_time_series_pipeline.py
eccabay: @jeremyliweishih we infer from y_train in the decomposer's fit, so the pipeline has no frequency knowledge.

@fjlanasa that was a good catch - this change would have prevented us from forecasting out 1 unit. Fortunately, the frequency detection isn't actually required within drop_time_index thanks to the other changes here, so I've just reverted it.

evalml/pipelines/time_series_pipeline_base.py
christopherbunn: Looking at the `_set_underlying_index()` function [here](https://woodwork.alteryx.com/en/stable/_modules/woodwork/table_accessor.html#WoodworkTableAccessor) it looks like it doesn't drop it automatically

christopherbunn (approved):

PR #3830, opened 2022-11-10
Allow engineered features in pipeline with dfs transfomer to have pd calculations done

tamargrey:
There are several partial dependence fixes relating to the DFS Transformer that happen in this PR:

1. This PR updates partial dependence to keep the origin values the same. Closes https://github.com/alteryx/evalml/issues/3834
2.  This PR excludes the target from consideration if it's present in the list of `features`. Closes https://github.com/alteryx/evalml/issues/3833
3.  This PR uses `f.get_feature_names` instead of `f.get_name` to handle multi output features correctly. Closes https://github.com/alteryx/evalml/issues/3832

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,6 +10,9 @@ Release Notes
         * Updated demo dataset links to point to new endpoint :pr:`3826`
         * Updated ``STLDecomposer`` to infer the time index frequency if it's not present :pr:`3829`
         * Updated ``_drop_time_index`` to move the time index from X to both ``X.index`` and ``y.index`` :pr:`3829`
+        * Fixed bug where engineered features lost their origin attribute in partial dependence, causing it to fail :pr:`3830`
+        * Fixed bug partial dependence's DFS Transformer fast mode handling wouldn't work with multi output features :pr:`3830`
+        * Allowed target to be present and ignored in partial dependence's DFS Transformer fast mode handling  :pr:`3830`
     * Changes
         * Consolidated decomposition frequency validation logic to ``Decomposer`` class :pr:`3811`
         * Removed Featuretools version upper bound and prevent Woodwork 0.20.0 from being installed :pr:`3813`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -11,7 +11,7 @@ Release Notes
         * Updated ``STLDecomposer`` to infer the time index frequency if it's not present :pr:`3829`
         * Updated ``_drop_time_index`` to move the time index from X to both ``X.index`` and ``y.index`` :pr:`3829`
         * Fixed bug where engineered features lost their origin attribute in partial dependence, causing it to fail :pr:`3830`
-        * Fixed bug partial dependence's DFS Transformer fast mode handling wouldn't work with multi output features :pr:`3830`
+        * Fixed bug where partial dependence's fast mode handling for the DFS Transformer wouldn't work with multi output features :pr:`3830`
         * Allowed target to be present and ignored in partial dependence's DFS Transformer fast mode handling  :pr:`3830`
     * Changes
         * Consolidated decomposition frequency validation logic to ``Decomposer`` class :pr:`3811`
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -35,8 +35,8 @@ def _get_cloned_feature_pipelines(
     new_parameters = pipeline.parameters
     for component in pipeline.component_graph.component_instances.values():
         new_parameters = component._handle_partial_dependence_fast_mode(
-            X_train,
             new_parameters,
+            X=X_train,
             target=pipeline.input_target_name,
         )
 
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -37,6 +37,7 @@ def _get_cloned_feature_pipelines(
         new_parameters = component._handle_partial_dependence_fast_mode(
             X_train,
             new_parameters,
+            target=pipeline.input_target_name,
         )
 
     # Create a fit pipeline for each feature
--- a/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
+++ b/evalml/model_understanding/_partial_dependence_fast_mode_utils.py
@@ -77,9 +77,7 @@ def _transform_single_feature(
             fit for it.
     """
     changed_col_df = pd.DataFrame({variable: part_dep_column})
-    changed_col_df.ww.init(
-        logical_types={variable: X.ww.logical_types[variable]},
-    )
+    changed_col_df.ww.init(schema=X.ww.schema.get_subset_schema([variable]))
 
     # Take the changed column and send it through transform by itself
     X_t_single_col = cloned_pipeline.transform_all_but_final(changed_col_df)
--- a/evalml/model_understanding/_partial_dependence_utils.py
+++ b/evalml/model_understanding/_partial_dependence_utils.py
@@ -318,8 +318,8 @@ def _partial_dependence_calculation(
                     X_eval.ww[variable] = ww.init_series(
                         part_dep_column,
                         logical_type=X_eval.ww.logical_types[variable],
+                        origin=X_eval.ww.columns[variable].origin,
                     )
-
             pred = prediction_method(X_eval)
             predictions.append(pred)
             # average over samples
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -103,14 +103,19 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
     def _supported_by_list_API(cls):
         return not cls.modifies_target
 
-    def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters, target):
+    def _handle_partial_dependence_fast_mode(
+        self,
+        pipeline_parameters,
+        X=None,
+        target=None,
+    ):
         """Determines whether or not a component can be used with partial dependence's fast mode.
 
         Args:
-            X (pd.DataFrame): Holdout data being used for partial dependence calculations.
             pipeline_parameters (dict): Pipeline parameters that will be used to create the pipelines
                 used in partial dependence fast mode.
-            target (str): The target whose values we are trying to predict.
+            X (pd.DataFrame, optional): Holdout data being used for partial dependence calculations.
+            target (str, optional): The target whose values we are trying to predict.
         """
         if self._can_be_used_for_fast_partial_dependence:
             return pipeline_parameters
--- a/evalml/pipelines/components/component_base.py
+++ b/evalml/pipelines/components/component_base.py
@@ -103,13 +103,14 @@ class ComponentBase(ABC, metaclass=ComponentBaseMeta):
     def _supported_by_list_API(cls):
         return not cls.modifies_target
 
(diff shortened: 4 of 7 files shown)


evalml/model_understanding/_partial_dependence_utils.py
```diff
@@ -318,8 +318,8 @@ def _partial_dependence_calculation(
                     X_eval.ww[variable] = ww.init_series(
                         part_dep_column,
                         logical_type=X_eval.ww.logical_types[variable],
+                        origin=X_eval.ww.columns[variable].origin,
```
tamargrey, line 321: Ideally, we'd be able to init with the original column's schema (as we do in fast mode above), as that would make sure that we don't lose any woodwork typing info as we replace this column, but that's not something we can currently do with init_series, so for now we'll just explicitly set individual woodwork parameters, but I opened up a Woodwork issue to make that possible: https://github.com/alteryx/woodwork/issues/1573

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2794,3 +2794,51 @@ def test_partial_dependence_fast_mode_errors_if_train(
...
+    pipeline = BinaryClassificationPipeline(
+        [dfs_transformer, "Standard Scaler", "Random Forest Classifier"],
+    )
+
+    # Engineered features have the their origins specified as either "base" or "engineered"
+    # it has to remain set for partial dependence to be able to predict on the updated data
+    engineered_feature = "ABSOLUTE(1)"
+    assert X_fm.ww.columns[engineered_feature].origin == "engineered"
+
+    pipeline = pipeline.clone()
```
jeremyliweishih, line 2831: don't think you need this clone call!

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2794,3 +2794,51 @@ def test_partial_dependence_fast_mode_errors_if_train(
...
+    pipeline.fit(X_fm, y)
+    part_dep = partial_dependence(
+        pipeline,
+        X_fm,
+        features=engineered_feature,
+        grid_resolution=5,
```
jeremyliweishih, line 2837: we could also try turning down the grid resolution so the test runs faster!

evalml/model_understanding/_partial_dependence_fast_mode_utils.py
```diff
@@ -77,9 +77,7 @@ def _transform_single_feature(
             fit for it.
     """
     changed_col_df = pd.DataFrame({variable: part_dep_column})
-    changed_col_df.ww.init(
-        logical_types={variable: X.ww.logical_types[variable]},
-    )
+    changed_col_df.ww.init(schema=X.ww.schema.get_subset_schema([variable]))
```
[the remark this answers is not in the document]
tamargrey: By passing in the schema, we get any and all woodwork information specified for that column. By passing in parameters, we only get the parameters we specify, which means we need to update it if woodwork type info changes.

So this will help us avoid something like this in the future if we were to make partial dependence rely on some other woodwork property (for example, depend on a semantic tag that's not tied to the logical type - we'd miss it with the previous call, because we aren't also passing along the semantic tags).

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: lowered the grid resolution to 2!

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: removed!

tamargrey:
Converting to draft temporarily, as I may include other DFS Transformer-related bug fixes in this PR

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2794,3 +2794,141 @@ def test_partial_dependence_fast_mode_errors_if_train(
...
+    )
+    seed_features = [ft.Feature(es["X"].ww["target"])]
+    X_fm, features = ft.dfs(
+        entityset=es,
+        target_dataframe_name="X",
+        trans_primitives=["absolute"],
+        ignore_columns={"X": ["target"]},
+        seed_features=seed_features,
+    )
+    assert any(f.get_name() == "target" for f in features)
```
jeremyliweishih, line 2916: probably not necessary but would we want to use `get_feature_names()` here as well for consistency?

jeremyliweishih (approved):
New changes LGTM

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: there's not really a need, imo, bc `target` is a base feature, so it's not possible for it to be a multi output feature (which are definitionally engineered features built from multi output primitives). It would just complicate the assertion here

evalml/pipelines/components/component_base.py
```diff
@@ -103,13 +103,14 @@ def default_parameters(cls):
     def _supported_by_list_API(cls):
         return not cls.modifies_target
 
-    def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters):
+    def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters, target):
```
eccabay, line 106: Can we make `target` an optional parameter? I had assumed it was based on the usage in the fast mode utils, and I think it makes more sense since we don't need it in most cases.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -141,14 +141,24 @@ def _handle_partial_dependence_fast_mode(self, X, pipeline_parameters):
             X (pd.DataFrame): Holdout data being used for partial dependence calculations.
             pipeline_parameters (dict): Pipeline parameters that will be used to create the pipelines
                 used in partial dependence fast mode.
+            target (str): The target whose values we are trying to predict. May be present in the
+                list of features in the DFS Transformer's parameters, in which case we should ignore it.
```
eccabay, line 145: It may be helpful to rephrase the second sentence here to make why we need it more explicit - i.e. "This is used to know which column to ignore if the target column is present in the list of features in the DFS Transformer's parameters" or something slightly cleaner sounding.

eccabay (approved):
Looks good, just a couple nitpicks but nothing blocking!

evalml/pipelines/components/component_base.py
tamargrey: Yes, and I can do the same for `X` as well, since it's only needed for the dfs transformer as well,

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2794,3 +2794,141 @@ def test_partial_dependence_fast_mode_errors_if_train(
...
+        target_dataframe_name="X",
+        trans_primitives=["LSA"],
+    )
+
+    dfs_transformer = DFSTransformer(features=features)
+    pipeline = RegressionPipeline(
+        [dfs_transformer, "Standard Scaler", "Random Forest Regressor"],
+    )
+    # Confirm that the LSA primitive was actually used
+    assert any(len(f.get_feature_names()) > 1 for f in features)
```
rwedge, line 2875: I think 
```
any(isinstance(f.primitive, ft.primitives.LSA) for f in features)
``` 

would be a clearer check criteria for the presence of LSA

evalml/pipelines/components/component_base.py
tamargrey: as more explanation: the reason why I hadn't originally made the extra args be optional is because we only call `_handle_partial_dependence_fast_mode` in one place where we have to pass in all the optional args. 

There's something that felt off about making the args optional when they will always be passed in and not be null. And because they're always passed in, we have to have them present in the component args even when they aren't used. I'd almost rather just use `_` in those components for the unused args, but I get that having them be optional makes them more flexible if we did want to use them in some other way in the future and is more technically correct, so I'm happy to make the change!

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: Good catch--I probably should've update the comment to say "Confirm that a multi output feature is present."

My thought as to why check explicitly for the multi-output nature of the feature is that EvalML doesn't really care about which primitive was used -- just that it's multi output.

rwedge: Makes sense!  Changing the comment would be nice.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: I like this suggestion -- changing to it!

docs/source/release_notes.rst
```diff
@@ -10,6 +10,9 @@ Release Notes
         * Updated demo dataset links to point to new endpoint :pr:`3826`
         * Updated ``STLDecomposer`` to infer the time index frequency if it's not present :pr:`3829`
         * Updated ``_drop_time_index`` to move the time index from X to both ``X.index`` and ``y.index`` :pr:`3829`
+        * Fixed bug where engineered features lost their origin attribute in partial dependence, causing it to fail :pr:`3830`
+        * Fixed bug partial dependence's DFS Transformer fast mode handling wouldn't work with multi output features :pr:`3830`
```
thehomebrewnerd, line 14: nit: mabye clean up the wording here a little.

evalml/model_understanding/_partial_dependence_utils.py
thehomebrewnerd: Should we also create a corresponding EvalML issue to update this once the WW issue is resolved so we don't forget about it?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2794,3 +2794,141 @@ def test_partial_dependence_fast_mode_errors_if_train(
...
+        X_fm,
+        features=engineered_feature,
+        grid_resolution=2,
+        fast_mode=fast_mode,
+        X_train=X_fm,
+        y_train=y,
+    )
+
+    assert not part_dep.feature_values.isna().any()
+    assert not part_dep.partial_dependence.isna().any()
```
thehomebrewnerd, line 2843: I think these assertions might be clearer if rewritten, if I'm following right and assuming we are checking there are no null values present.

```
assert part_dep.feature_values.notnull().all()
assert part_dep.partial_dependence.notnull().all()
```

evalml/tests/model_understanding_tests/test_partial_dependence.py
thehomebrewnerd: Building off of Roy's comment, I think you could just check the `number_output_features` property of the Feature directly:
```
assert any(f.number_output_features > 1 for f in features)
```

evalml/model_understanding/_partial_dependence_utils.py
tamargrey: Created: https://github.com/alteryx/evalml/issues/3847

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: awesome, forgot that property exists!

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: made this change (and also updated other tests to use this pattern - they were doing the wrong check before, anyway)

thehomebrewnerd (approved):

PR #3836, opened 2022-11-15
Add DFSTransformer to time series pipelines

jeremyliweishih:
Fixes #3931


evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -170,14 +182,15 @@ def test_make_pipeline(
...
+                    dfs
+                    + label_encoder
                     + replace_null
                     + email_featurizer
                     + url_featurizer
                     + drop_null
                     + delayed_features
-                    + natural_language_featurizer
                     + datetime
+                    + natural_language_featurizer
```
jeremyliweishih, line 193: I'm not sure how this was passing before. For non-TS problems the `NaturalLanguageFeaturizer` should always go after the `DatetimeFeaturizer`

ozzieD (approved):

evalml/pipelines/utils.py
```diff
@@ -385,6 +386,7 @@ def _make_pipeline_time_series(
            Valid options are "DatetimeFeaturizer", "EmailFeaturizer", "URLFeaturizer", "NaturalLanguageFeaturizer", "TimeSeriesFeaturizer"
         include_decomposer (bool): For time series regression problems, whether or not to include a decomposer in the generated pipeline.
             Defaults to True.
+        features (bool): Whether to add a DFSTransformer component to this pipeline.
```
thehomebrewnerd, line 389: Just wondering if this param should be called something like `add_dfs_transformer` instead of `features` to make it a bit more clear? I guess this is consistent with what is used in `make_pipeline` though.

thehomebrewnerd (approved):
Just one non-blocking comment, but otherwise looks good.

PR #3849, opened 2022-11-17
Add pipeline.should_skip_featurization flag

eccabay:
Fixes an issue where time series native estimators were double-featurizing during predict when featurization happens before running evalml


evalml/tests/integration_tests/test_time_series_integration.py
```diff
@@ -224,3 +227,120 @@ def test_can_run_automl_for_time_series_with_exclude_featurizers(
             assert not pipeline.should_drop_time_index
         else:
             assert pipeline.should_drop_time_index
+        assert pipeline.should_skip_featurization
+
+
+@patch(
```
jeremyliweishih, line 233: do you think its cleaner to add the pipeline fitting logic to `test_can_run_automl_for_time_series_with_exclude_featurizers` and check after search is run? I think either works!

jeremyliweishih (approved):

evalml/tests/integration_tests/test_time_series_integration.py
eccabay: Great idea, done!

chukarsten (approved):

PR #3858, opened 2022-11-23
Make evalml compatible with `woodwork == 0.20.0`, upgrade to `0.21.0`

ParthivNaresh:


ParthivNaresh:
Windows tests are failing here due to the editable file installation of woodwork main I believe. This should resolve itself once Woodwork 0.20.0 is released.

ParthivNaresh:
@chukarsten I might need you to force merge this for me once approved. There's one line not covered which is the `TypeError` check for the inference function. It doesn't need to be tested since it's an edge case logic that's already tested in Woodwork.

.github/meta.yaml
```diff
@@ -37,7 +37,7 @@ outputs:
         - requirements-parser >=0.2.0
         - shap >=0.40.0
         - texttable >=1.6.2
-        - woodwork <=0.19.0
+        - woodwork >=0.21.0
```
eccabay, line 40: Are the changes in this PR incompatible with older versions of woodwork?

docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
         * Allowed the DFS Transformer to calculate feature values for Features with a ``dataframe_name`` that is not ``"X"`` :pr:`3873`
     * Changes
         * Reduced the threshold for setting ``use_covariates`` to False for ARIMA models in AutoMLSearch :pr:`3868`
+        * Changed ``BooleanNullable`` inference logic :pr:`3858`
```
eccabay, line 10: Can you add a line about upgrading the woodwork version as well? (only really useful since there's the note about pinning it in here)

evalml/utils/woodwork_utils.py
```diff
@@ -27,6 +28,32 @@ def _list_to_pandas(list):
...
+                return True
+        except TypeError:  # Necessary to check for non-hashable values because of object dtype consideration
+            return False
+    return False
+
+
+ww.type_system.update_inference_function(
+    "BooleanNullable",
+    inference_function=_new_infer_boolean_nullable_func,
+)
```
eccabay, line 54: This is wacky. If I understand correctly, we're directly changing woodwork inference behavior during init of the evalml package? Why is this necessary for evalml? It seems odd to fork off of standard woodwork behavior, it may cause headaches in the future.

evalml/utils/woodwork_utils.py
```diff
@@ -27,6 +28,32 @@ def _list_to_pandas(list):
     return _numpy_to_pandas(np.array(list))
 
 
+def _new_infer_boolean_nullable_func(series):
```
eccabay, line 31: Even though this function is private, context on what it's doing would be useful for future work. Could you add a docstring?

eccabay (commented):
Thanks for taking care of this Parthiv! I just have a few questions before we go ahead and merge this in

evalml/utils/woodwork_utils.py
ParthivNaresh: This was the reason that the `update_inference_function` was added in Woodwork, so that inference functions could be changed/added whenever necessary.

Unfortunately the alternative is to not only update every test that runs into this Boolean inference issue, but also to update several components like the `LabelEncoder` that run into reverse mapping issues. In order for the new Boolean inference to be supported here, there needs to be an underlying change in cases where previously `Categorical` variables will be treated as `Boolean` instead.

Luckily any downstream `Boolean` transformations have to go through this inference first, so this can act as a sort of "gatekeeper" for this behaviour, for now.

jeremyliweishih: @ParthivNaresh is what EvalML considers "boolean" different from what WW does now? and what is the difference? Just curious why inference shouldn't be standard between WW and EvalML.

.github/meta.yaml
ParthivNaresh: Confirmed that `0.19.0` works fine while `0.20.0` has dependency issues. I updated the dependencies appropriately!

jeremyliweishih (approved):
LGTM!

eccabay (approved):

chukarsten (changes requested):
I've got a bad feeling about this

PR #3861, opened 2022-11-28
Fix skip_featurization bug

eccabay:
Time series classification pipelines weren't dropping the time index. This wasn't an issue in the short term, but could become one in the long term.


eccabay:
After discussion with @jeremyliweishih, this change may not be the fix we're looking for. For now, it's better for us to revert the skip_featurization change.

PR #3868, opened 2022-11-30
Reduce use_covariates threshold

eccabay:
Moves to defaulting to `use_covariates=False` in more cases, since it produces significant slowdown without much benefit for larger datasets

### DIFF
--- a/evalml/automl/automl_search.py
+++ b/evalml/automl/automl_search.py
@@ -711,7 +711,7 @@ class AutoMLSearch:
         self.search_parameters = search_parameters or {}
         # Fitting takes a long time if the data is too wide or long.
         if is_time_series(problem_type) and (
-            self.X_train.shape[1] >= 10 or self.X_train.shape[0] >= 10000
+            self.X_train.shape[1] >= 7 or self.X_train.shape[0] >= 1000
         ):
             user_arima_hyperparams = ARIMARegressor.name in self.search_parameters
             if user_arima_hyperparams and not self.search_parameters[


evalml/automl/automl_search.py
```diff
@@ -711,7 +711,7 @@ def __init__(
         self.search_parameters = search_parameters or {}
         # Fitting takes a long time if the data is too wide or long.
         if is_time_series(problem_type) and (
-            self.X_train.shape[1] >= 10 or self.X_train.shape[0] >= 10000
+            self.X_train.shape[1] >= 7 or self.X_train.shape[0] >= 1000
```
jeremyliweishih, line 714: can we bake this in a constant somewhere and use this constant in the test as well?

jeremyliweishih (approved):
LGTM just a suggestion

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -61,6 +61,9 @@ class ARIMARegressor(Estimator):
     supported_problem_types = [ProblemTypes.TIME_SERIES_REGRESSION]
     """[ProblemTypes.TIME_SERIES_REGRESSION]"""
 
+    max_rows = 1000
+    max_cols = 7
```
christopherbunn, line 65: Not a blocker but is there a reason why this is 7 and not like 5 or 10? Just very curious haha

eccabay: Nope, no strong motivation behind it. It was just a bit of a bump down, and without a more comprehensive set of datasets there was no way to test where the best line in the sand would be. Since these results seemed ok, I just stuck with it.

Issue #3872, opened 2022-12-05
DFS Transformer will error if features passed in have DataFrame name that is not "X"


tamargrey:
In the dfs transformer, we create an entity set with a dataframe named `X`. In the cases where users pass in their own `features`, we may end up running `calculate_feature_matrix` using those features and the entityset created by the DFS Transformer. In that situation, if the features were not created from a DataFrame named `X`, we get an error in Featuretools that `DataFrame <different name> does not exist in entity set` at transform.

#### Code Sample, a copy-pastable example to reproduce your bug.

```python
    X, y = X_y_binary
    X_pd = pd.DataFrame(X)
    X_pd.columns = X_pd.columns.astype(str)

    es = ft.EntitySet()
    es = es.add_dataframe(
        dataframe_name="Xdiff",
        dataframe=X_pd,
        index="index",
        make_index=True,
    )
    feature_matrix, features = ft.dfs(
        entityset=es,
        target_dataframe_name="Xdiff",
        trans_primitives=["absolute"],
    )

    dfs = DFSTransformer(features=features)
    dfs.fit(X_pd)  # no-op
    X_t = dfs.transform(X_pd)
```
If this is a requirement, we should make that clear in the documentation, but if the name of the dataframe in the entityset being `"X"` is not a requirement, we should just set the name to be the same as the dataframe used to create the features, if features are passed in at init. We can do this by looking at `self.features[0].dataframe_name`.

PR #3873, opened 2022-12-05
Handle features with dataframe names that aren't 'X'

tamargrey:
closes https://github.com/alteryx/evalml/issues/3872

If features are passed in, uses the `dataframe_name` from the Feature object to set the `dastaframe_name` in `_make_entity_set` instead of always calling it `"X"`.

This allows us to run calculate feature matrix from the DFS transformer on Features with a dataframe name that is not `"X"`.


evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -28,6 +28,8 @@ def __init__(self, index="index", features=None, random_seed=0, **kwargs):
         self.index = index
         self.features = features
         self._passed_in_features = True if features else None
+        # If features are passed in, they'll have a dataframe_name we should utilize
+        self._dataframe_name = self.features[0].dataframe_name if self.features else "X"
```
tamargrey, line 32: The use of `self.features[0]` here assumes that all features were created from the same dataframe.

I didn't add a comment here explaining that, since it seems that assumption is true for the whole component (for example, in how we create the entityset in `_make_entity_set`), but I'd be happy to note that here if it would be useful.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -36,16 +38,21 @@ def _make_entity_set(self, X):
         ft_es = EntitySet()
         # TODO: This delete was introduced for compatibility with Featuretools 1.0.0.  This should
         # be removed after Featuretools handles unnamed dataframes being passed to this function.
+        # But even then, we should still use any dataframe name that might be available from input features.
```
tamargrey, line 41: The way I see it, if a user has passed `Feature` objects in, then we _do_ have a dataframe_name that we can use for `X` instead of `X.ww.name` (even if they're the same!), so this TODO explaining why we delete the woodwork schema does still apply.

I wanted to make sure that was clear, since I think it's a little less obvious now that we aren't fully hardcoding `"X"`.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
jeremyliweishih: let's add the note!

evalml/pipelines/components/transformers/preprocessing/featuretools.py
jeremyliweishih: love it

jeremyliweishih (approved):
amazing!

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: @jeremyliweishih added it 4c374086d - let me know if that makes sense to you!

chukarsten (approved):

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -36,16 +40,21 @@ def _make_entity_set(self, X):
...
                 index=self.index,
                 make_index=True,
             )
         else:
-            es = ft_es.add_dataframe(dataframe=X, dataframe_name="X", index=self.index)
+            es = ft_es.add_dataframe(
+                dataframe=X,
+                dataframe_name=self._dataframe_name,
+                index=self.index,
+            )
```
thehomebrewnerd, line 57: This is a bit of an unrelated nitpick, but I think this entire code block can be cleaned up to remove the if/else construct:

```
del X.ww
should_make_index = self.index not in X.columns
es = ft_es.add_dataframe(
    dataframe=X,
    dataframe_name=self._dataframe_name,
    index=self.index,
    make_index=should_make_index,
)
```

thehomebrewnerd (approved):
One no-blocking suggestion, but looks good to me overall.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: nice, changed

Issue #3874, opened 2022-12-05
DFS Transformer can produce incorrect results in partial dependence fast mode if engineered features are present


tamargrey:
The intent of passing `features` into the DFS transformer is that it allows us to avoid calling calculate feature matrix on `X` if we've already calculated the feature values for the inputted list of features. For partial dependence fast mode, we only allow pipelines that contain the DFS Transformer when this is the case (See for explanation why [here](https://github.com/alteryx/evalml/blob/main/evalml/tests/model_understanding_tests/test_partial_dependence.py#L2657-L2660)). 

The logic that determines whether or not calculate feature matrix is called cannot properly account for fast mode refitting a cloned pipeline on a single column, so if the single column had a feature engineered from it, we will attempt to calculate feature values for it, which does not match the behavior of slow mode.

Because of https://github.com/alteryx/evalml/issues/3872, this will currently raise errors when the features were not trained on a dataframe named `"X"`.  Once that is fixed, though, we can still see that the partial dependence values for fast mode in these cases are different from slow mode.

We should update `DFSTransformer._handle_partial_dependence_fast_mode` to update the `features` parameter such that we do not run `calcualte_feature_matrix` when we transform the single column pipeline. One way to do this is to set features to be an empty list, but that will require changing how we set `_passed_in_features`, which just looks at the truthy value of features.

PR #3875, opened 2022-12-05
Update dfs transformer features arg to be empty list for fast mode

tamargrey:
Closes https://github.com/alteryx/evalml/issues/3874

Avoids passing the DFS Transformer's full list of features into the cloned pipeline for partial dependence fast mode. To achieve this, updates the DFS Transformer to allow an empty list as input for the `features` arg. 

Note: This PR is currently set to merge into #3873 because it's dependent on changes made in that PR. We should merge that first and then merge this into main.


evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, index="index", features=None, random_seed=0, **kwargs):
 
         self.index = index
         self.features = features
-        self._passed_in_features = True if features else None
+        self._passed_in_features = True if features is not None else None
```
tamargrey, line 30: To make this fix, I needed the ability to set the parameters of the DFS Transformer so that a `transform` call would be no op. One option I considered was only passing in the `IdentityFeature` objects in at ` _handle_partial_dependence_fast_mode`, but I felt that left some ambiguity as to what the behavior would be. 

Therefore, I updated the behavior here to no longer treat an empty list passed in for `features` the same as `features` being unspecified. This change made sense to me, but I wanted to highlight it in case anyone disagrees or thinks this could be problematic.

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2934,3 +2935,69 @@ def test_partial_dependence_dfs_transformer_target_in_features(fast_mode, X_y_bi
...
+        max_iterations=2,
+        features=features,
+        automl_algorithm="default",
+    )
+
+    automl.search()
+
+    assert not mock_calculate_feature_matrix.called
+    # --> Going through automl, because we have to instantiate the DFSTransformer to make a pipeline directly,
+    # which means we don't get the updated parameters applied, causing this test to fail
```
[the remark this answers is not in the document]
tamargrey: If I do need to keep this, I'll plan to use the AutoMLTestEnv

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -166,4 +166,8 @@ def _handle_partial_dependence_fast_mode(self, pipeline_parameters, X, target):
                 raise ValueError(
                     "Cannot use fast mode with DFS Transformer when features are unspecified or not all present in X.",
                 )
+            # Pass in empty list of features so we don't run calculate feature matrix
+            # which would happen with the full set of features for a single column at refit
+            dfs_transformer["features"] = []
+            pipeline_parameters["DFS Transformer"] = dfs_transformer
```
eccabay, line 172: A clarification for my own understanding:  we only use this function with the single column pipeline, correct? This code would never run when we're generating the transformations from the full dataset?

evalml/tests/component_tests/test_featuretools.py
```diff
@@ -241,6 +241,31 @@ def test_dfs_with_serialized_features_dataframe_name(
...
+@patch(
+    "evalml.pipelines.components.transformers.preprocessing.featuretools.calculate_feature_matrix",
+)
+def test_dfs_with_empty_input_features(
+    mock_calculate_feature_matrix,
+    mock_dfs,
+    X_y_binary,
+):
+    """Confirms that the features arg being an empty list is not treated the same as
+    it being unspecified."""
```
eccabay, line 254: Can we be explicit about testing this? i.e. generate two DFS objects, one with `features=[]` and one with `features=None` and test the differences?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -2934,3 +2935,67 @@ def test_partial_dependence_dfs_transformer_target_in_features(fast_mode, X_y_bi
...
+        grid_resolution=2,
+        fast_mode=True,
+        X_train=X_fm,
+        y_train=y,
+    )
+    assert not mock_calculate_feature_matrix.called
+
+    assert part_dep.feature_values.notnull().all()
+    assert part_dep.partial_dependence.notnull().all()
+    pd.testing.assert_frame_equal(part_dep, fast_part_dep)
```
eccabay, line 3001: Another question just for my own understanding: does this test fail with our current main branch?

eccabay (approved):
Looks pretty good! Just left a few clarification questions.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: Correct - the parameters returned here will only be used to train a pipeline on a single column, and the check above this that confirms all features have corresponding columns in X confirms that calculate feature matrix is never called in the original pipeline; this parameter change makes that the case for the single column as well.

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: Yes, it does fail, as `calculate_feature_matrix` gets called in fast mode causing the mock checks above to fail

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: Maybe the warning could come from AutoMLSearch itself. If it sees that a user is trying to have a DFS Transformer in automl search with no features, it could raise that warning. If it's passed directly into a pipeline, maybe we're okay not warning users since it might be more clear to users at that point what the impact would be. 

Then again, my goal here was specifically to have no-op functionality in the DFS Transformer, but I definitely get the discomfort. At the very least, I need to update the dosctring making this explicit.

evalml/tests/model_understanding_tests/test_partial_dependence.py
eccabay: Ah gotcha. I guess I'm more concerned about the `pd.testing.assert_frame_equal(part_dep, fast_part_dep)` line specifically, since that's what this PR is aimed at combatting

evalml/pipelines/components/transformers/preprocessing/featuretools.py
eccabay: Ah, the warning within AutoMLSearch is a good idea. Having it in the docstring is also a super good callout, sounds like a good path forward to me

tamargrey: Turns out AutomlSearch doesn’t add the DFS Transformer to pipelines if features is an empty list.

We definitely add it to the internal search parameters [here](https://github.com/alteryx/evalml/blob/main/evalml/automl/automl_search.py#L736-L740) because we explicitly check if it’s None or not, but when we make_pipeline we exclude it if features is falsy: https://github.com/alteryx/evalml/blob/main/evalml/pipelines/utils.py#L412. We do the same for split pipelines. So currently in automl, an empty features list already is no-op.

My preferance here would be to not touch the automl code (though, maybe it’d be worth adding a note to the docstring there about the behavior when features is an empty list). And it makes me more confident that we probably want this behavior for the component itself as well.

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: yep totally doable

evalml/pipelines/components/transformers/preprocessing/featuretools.py
tamargrey: Yep, updated the automl docstring to make this clear there as well

PR #3878, opened 2022-12-08
Update leaderboard names to show `ranking_score` instead of `validation_score`

christopherbunn:
Change `validation_score` to `ranking_score`. This PR will also show the `holdout_score` if there is a holdout set in the leaderboard.


eccabay (approved):
LGTM!

jeremyliweishih (approved):
LGTM as well

PR #3883, opened 2022-12-10
Set threshold to account for too many duplicate or nan values in `DateTimeFormatDataCheck`

ParthivNaresh:
Currently if a `time_index` is chosen that has duplicate or nan values, these values are dropped prior to the creation of an `alias_dict` in Woodwork which iterates over all identified frequencies in the time series to determine the most likely one. 

The side effect of this is that in a dataset with a length of 20,000, if the `time_index` is just 50 consecutive daily datetime values repeated 400 times, then we currently return `DATETIME_HAS_UNEVEN_INTERVALS` and an action code to fix this data.

This `time_index` should not be considered for regularization as regularizing this data would reduce this dataset from 20,000 observations to 50, since all duplicates and nans would be dropped and the remaining datetime values would have a frequency of `1 day`. This is more of a multi-series concern which is outside the scope of this data check.

I've added a check where if 25% of the `time_index` consists of solely duplicate and nan values, then `DATETIME_NO_FREQUENCY_INFERRED` will be returned instead of `DATETIME_HAS_UNEVEN_INTERVALS`. I'm open to changing this threshold or making it a passable argument, it was more or less arbitrary.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+        * Added a threshold to ``DateTimeFormatDataCheck`` to account for too many duplicate or nan values :pr:`3883`
     * Documentation Changes
     * Testing Changes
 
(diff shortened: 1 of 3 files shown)


eccabay (approved):
Since the threshold was arbitrary, I think it certainly wouldn't hurt to make it a passable argument. Regardless, looks great!

christopherbunn (approved):
Makes sense to me! +1 to @eccabay's comment on making it a passable argument or even just an internal global variable.

jeremyliweishih (approved):

chukarsten (approved):

PR #3884, opened 2022-12-12
Fix ARIMA not accounting for gap in prediction from end of training data

### LINKED ISSUE #3853: `ARIMARegressor` does not support predictions of input with a gap off of training data
```
import pandas as pd

df = pd.read_csv("/Users/jeremy.shih/Downloads/souvenirs.csv")
df.ww.init()
y = df["Sales"]
X = df.drop('Sales', axis=1)

X_train, X_test, y_train, y_test = split_data(
        X, y, problem_type="time series regression", test_size=0.10, random_seed=0
    )

X_test_last_5 = X_test.tail(5)

arima = ARIMARegressor()
arima.fit(X_train, y_train)

assert (arima.predict(X_test).tail(5).tolist() == arima.predict(X_test_last_5).tolist())
```

Current `ARIMARegressor` creates a `ForecastingHorizon` with the `is_relative` flag turned on. This means that the ARIMA object will try to predict dates off of the last date known in the training data. Thus, it will provide the wrong predictions for dates with a gap off of the training data. 

A potential fix would be to use the `is_relative` flag but account for the difference in index between the last index in `X_train` and in `X_test`

```
def _set_forecast(self, X):
        from sktime.forecasting.base import ForecastingHorizon
        units_diff = X.index[0] - self.last_X_index
        fh_ = ForecastingHorizon([units_diff + i for i in range(len(X))], is_relative=True)
```

jeremyliweishih:
Fixes #3853.

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -142,13 +142,25 @@ class ARIMARegressor(Estimator):
     def _set_forecast(self, X):
         from sktime.forecasting.base import ForecastingHorizon
 
-        units_diff = X.index[0] - self.last_X_index
-        if isinstance(X.index, pd.DatetimeIndex):
-            units_diff = int(units_diff / np.timedelta64(1, X.index.freqstr))
-        fh_ = ForecastingHorizon(
-            [units_diff + i for i in range(len(X))],
-            is_relative=True,
-        )
+        # we can only calculate the difference if the indices are of the same type
+        if isinstance(X.index[0], type(self.last_X_index)):
+            units_diff = X.index[0] - self.last_X_index
+            if isinstance(X.index, pd.DatetimeIndex):
+                dates_diff = pd.date_range(
+                    start=self.last_X_index,
+                    end=X.index[0],
+                    freq=X.index.freq,
+                )
+                units_diff = len(dates_diff) - 1
+            fh_ = ForecastingHorizon(
+                [units_diff + i for i in range(len(X))],
+                is_relative=True,
+            )
+        else:
+            fh_ = ForecastingHorizon(
+                [i + 1 for i in range(len(X))],
+                is_relative=True,
+            )
         return fh_
 
     def _get_sp(self, X):
@@ -191,7 +203,7 @@ class ARIMARegressor(Estimator):
 
         sp = self._get_sp(X)
         self._component_obj.sp = sp
-        self.last_X_index = X.index[-1]
+        self.last_X_index = X.index[-1] if X is not None else y.index
 
         X = self._remove_datetime(X, features=True)
 
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -143,7 +143,8 @@ class ARIMARegressor(Estimator):
         from sktime.forecasting.base import ForecastingHorizon
 
         units_diff = X.index[0] - self.last_X_index
-        units_diff = int(units_diff / np.timedelta64(1, X.index.freqstr))
+        if isinstance(X.index, pd.DatetimeIndex):
+            units_diff = int(units_diff / np.timedelta64(1, X.index.freqstr))
         fh_ = ForecastingHorizon(
             [units_diff + i for i in range(len(X))],
             is_relative=True,
--- a/evalml/pipelines/components/estimators/regressors/arima_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/arima_regressor.py
@@ -142,7 +142,12 @@ class ARIMARegressor(Estimator):
     def _set_forecast(self, X):
         from sktime.forecasting.base import ForecastingHorizon
 
-        fh_ = ForecastingHorizon([i + 1 for i in range(len(X))], is_relative=True)
+        units_diff = X.index[0] - self.last_X_index
+        units_diff = int(units_diff / np.timedelta64(1, X.index.freqstr))
+        fh_ = ForecastingHorizon(
+            [units_diff + i for i in range(len(X))],
+            is_relative=True,
+        )
         return fh_
 
     def _get_sp(self, X):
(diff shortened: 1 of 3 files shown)


evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -142,7 +142,25 @@ def _match_indices(self, X, y):
     def _set_forecast(self, X):
         from sktime.forecasting.base import ForecastingHorizon
 
-        fh_ = ForecastingHorizon([i + 1 for i in range(len(X))], is_relative=True)
+        # we can only calculate the difference if the indices are of the same type
```
jeremyliweishih, line 145: set this as a fallback to previous behavior in case we ever receive inconsistent indices

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -142,7 +142,25 @@ def _match_indices(self, X, y):
...
+            units_diff = X.index[0] - self.last_X_index
+            if isinstance(X.index, pd.DatetimeIndex):
+                dates_diff = pd.date_range(
+                    start=self.last_X_index,
+                    end=X.index[0],
+                    freq=X.index.freq,
+                )
+                units_diff = len(dates_diff) - 1
+            fh_ = ForecastingHorizon(
+                [units_diff + i for i in range(len(X))],
```
eccabay, line 156: Seems like we needed an off-by-one offset in the non-gap case. Should that still be the case here as well?

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -139,6 +140,7 @@ def test_set_forecast(ts_data):
     )
 
     clf = ARIMARegressor()
+    clf.last_X_index = X.index[-1]
```
eccabay, line 143: Why was this necessary?

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -397,3 +399,25 @@ def test_arima_regressor_respects_use_covariates(mock_predict, mock_fit, ts_data
     assert "y" in mock_fit.call_args.kwargs
     mock_predict.assert_called_once()
     assert "X" not in mock_predict.call_args.kwargs
+
+
+@pytest.mark.parametrize("use_covariates", [True, False])
+def test_arima_regressor_can_forecast_arbitrary_dates(use_covariates, ts_data):
```
eccabay, line 405: It might be useful here to highlight how this is arbitrary dates in a comment or something - how long is X_test here, so how big of a gap is it between the training data and what we're asking to predict on?

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -397,3 +399,25 @@ def test_arima_regressor_respects_use_covariates(mock_predict, mock_fit, ts_data
...
+    )
+
+    X_test_last_5 = X_test.tail(5)
+
+    arima = ARIMARegressor(use_covariates=use_covariates)
+    arima.fit(X_train, y_train)
+
+    assert (
+        arima.predict(X_test).tail(5).tolist() == arima.predict(X_test_last_5).tolist()
+    )
```
eccabay, line 423: Does this fail when run on current main?

eccabay (approved):
Looks pretty good, just had a few questions! Thanks for taking care of this.

evalml/tests/component_tests/test_arima_regressor.py
jeremyliweishih: we never ran fit in this test so `last_X_index` wasn't being set!

evalml/tests/component_tests/test_arima_regressor.py
jeremyliweishih: will add!

evalml/tests/component_tests/test_arima_regressor.py
jeremyliweishih: yes!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
chukarsten: Is the index type being different indicative of something?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -185,8 +203,10 @@ def fit(self, X, y=None):
 
         sp = self._get_sp(X)
         self._component_obj.sp = sp
+        self.last_X_index = X.index[-1] if X is not None else y.index
```
chukarsten, line 206: This looks kind of weird....either the value is a value of an index...or it's an index?

evalml/pipelines/components/estimators/regressors/arima_regressor.py
jeremyliweishih: We need the + 1 since `range(len(X))` starts at 0. In the gap case `units_diff` should always be > 0 so we don't need it!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
```diff
@@ -226,7 +246,17 @@ def predict(self, X, y=None):
             },
         )
         if not X.empty and self.use_covariates:
-            y_pred = self._component_obj.predict(fh=fh_, X=X)
+            if fh_[0] != 1:
+                # pmdarima (which sktime uses under the hood) only forecasts off the training data
+                # but sktime circumvents this by predicting everything from the end of training data to the date / periods requested
+                # and only returning the values for dates / periods given to sktime. Because of this,
+                # pmdarima requires the number of covariate rows to equal the length of the total number of periods (X.shape[0] == fh_[-1]) if covariates are used.
+                # We circument this by adding arbitrary rows to the start of X since sktime discards these values when predicting.
```
chukarsten, line 254:
```suggestion
                # We circumvent this by adding arbitrary rows to the start of X since sktime discards these values when predicting.
```

chukarsten (approved):
Just some small things and questions.  Little nit to approve.

evalml/pipelines/components/estimators/regressors/arima_regressor.py
jeremyliweishih: it should only believe the result of inconsistent user behavior and this is a safeguard against that!

evalml/pipelines/components/estimators/regressors/arima_regressor.py
jeremyliweishih: oops - will fix!

PR #3890, opened 2022-12-14
Hid non-essential warning messages in time series docs

### LINKED ISSUE #3620: Fix the Time Series Docs

christopherbunn:
Fixes #3620 

A lot of the warning messages that we are seeing has to do with a lot of the internal solvers being unable to find convergence. A significant amount of these errors can be filtered out by filtering them in a hidden cell. 

In addition, for the known in advance section, a number of errors are `UserErrors` related to initial parameter inits in addition to the above convergence error messages. This second cell hiding UserErrors I am a bit more skeptical about as it's a bit broad and leaves room for actual errors to be hidden in the future. 

Current page: https://evalml.alteryx.com/en/stable/user_guide/timeseries.html#Known-in-advance-features
Updated page: https://feature-labs-inc-evalml--3890.com.readthedocs.build/en/3890/user_guide/timeseries.html#Known-in-advance-features

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,7 +8,7 @@ Release Notes
     * Changes
         * Added a threshold to ``DateTimeFormatDataCheck`` to account for too many duplicate or nan values :pr:`3883`
     * Documentation Changes
-        * Hide non-essential warning messages in time series docs :pr:`3890`
+        * Hid non-essential warning messages in time series docs :pr:`3890`
     * Testing Changes
 
 .. warning::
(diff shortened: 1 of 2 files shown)


eccabay (approved):
Looks good!

Should we file a follow up issue to track removing those parameter not used warnings? It seems like those shouldn't be raised in the first place.

Unrelated to this issue, I'm concerned by the fact that we're getting info-level logging information. We shouldn't be, since `verbose` is set to false. I think one of our dependencies may be mucking with the logging configuration, which is something for us to keep an eye on.

jeremyliweishih (approved):
@eccabay agreed - filed #3891!

PR #3892, opened 2022-12-15
Make EvalML compatible with the new Woodwork Boolean inference

ParthivNaresh:
The Boolean inference has been revised in Woodwork and has been broken into two stages:

1. The first stage consists of removing the new automatic Boolean inference but maintaining the ability to transform columns into `Boolean` using this new inference by specifying the `Boolean` or `BooleanNullable` logical types. What does this mean? A column of `1s and 0s` will not be inferred as `Boolean` through `df.ww.init()`, however specifying `df.ww.init(logical_types={"col": "Boolean"})` will transform that column into `True and False`. This PR is solely to get EvalML compatible with this stage in `woodwork==0.22.0`.
2. The second stage is longer term and will be a cross team effort involving deeper compatibility changes in EvalML components and tests to prevent inference changes from being a blocker to future releases.

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -37,7 +37,7 @@ outputs:
         - requirements-parser >=0.2.0
         - shap >=0.40.0
         - texttable >=1.6.2
-        - woodwork >=0.19.0,!=0.20.0
+        - woodwork >=0.19.0,!=0.20.0,!=0.21.0
         - featuretools>=1.16.0
         - nlp-primitives>=2.9.0
         - python >=3.8.*
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -37,7 +37,7 @@ outputs:
         - requirements-parser >=0.2.0
         - shap >=0.40.0
         - texttable >=1.6.2
-        - woodwork <=0.19.0
+        - woodwork >=0.19.0,!=0.20.0
         - featuretools>=1.16.0
         - nlp-primitives>=2.9.0
         - python >=3.8.*
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -37,7 +37,7 @@ outputs:
         - requirements-parser >=0.2.0
         - shap >=0.40.0
         - texttable >=1.6.2
-        - woodwork >=0.19.0,!=0.20.0,!=0.21.0
+        - woodwork >=0.21.1
         - featuretools>=1.16.0
         - nlp-primitives>=2.9.0
         - python >=3.8.*
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -11,7 +11,7 @@ requirements-parser>=0.2.0
 shap>=0.40.0
 statsmodels>=0.12.2
 texttable>=1.6.2
-woodwork>= 0.19.0, !=0.20.0
+woodwork>= 0.19.0, !=0.20.0, !=0.21.0
 dask>=2022.2.0, !=2022.10.1
 nlp-primitives>=2.9.0
 featuretools>=1.16.0
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -11,7 +11,7 @@ requirements-parser>=0.2.0
 shap>=0.40.0
 statsmodels>=0.12.2
 texttable>=1.6.2
-woodwork>= 0.19.0, !=0.20.0, !=0.21.0
+woodwork>= 0.21.1
 dask>=2022.2.0, !=2022.10.1
 nlp-primitives>=2.9.0
 featuretools>=1.16.0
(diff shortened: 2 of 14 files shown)


.github/meta.yaml
```diff
@@ -37,7 +37,7 @@ outputs:
         - requirements-parser >=0.2.0
         - shap >=0.40.0
         - texttable >=1.6.2
-        - woodwork <=0.19.0
+        - woodwork >=0.19.0,!=0.20.0
```
bchen1116, line 40: Should be `!=0.21.0`, right? Since that's the version that introduces the breaking changes?

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -146,7 +147,15 @@ def validate(self, X, y):
         """
         messages = []
 
+        original_vc = pd.Series(y).value_counts(sort=True)
         y = infer_feature_types(y)
+        new_vc = y.value_counts(sort=True)
+        reverse_mapping = {
```
bchen1116, line 153: Might be nice to name this and `normal_mapping` to something more easily understandable, like `ww_transformed_mapping` or something of the sort. Would make comprehending this code much easier.

.github/meta.yaml
ParthivNaresh: Ah good call, it'll actually be both `0.20.0` and `0.21.0` that get blocked

core-requirements.txt
```diff
@@ -11,7 +11,7 @@ requirements-parser>=0.2.0
 shap>=0.40.0
 statsmodels>=0.12.2
 texttable>=1.6.2
-woodwork<=0.19.0
+woodwork>= 0.19.0, !=0.20.0
```
bchen1116, line 14: Should update this as well!

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -146,7 +147,19 @@ def validate(self, X, y):
...
+        new_vc = y.value_counts(sort=True)
+        after_to_before_inference_mapping = {
+            new: old for old, new in zip(original_vc.keys(), new_vc.keys())
+        }
+        before_to_after_inference_mapping = {
+            old: new for new, old in after_to_before_inference_mapping.items()
+        }
+        if str(y.ww.logical_type) not in ["Boolean", "BooleanNullable"]:
+            after_to_before_inference_mapping = {
+                old: old for old, _ in after_to_before_inference_mapping.items()
```
bchen1116, line 161: nit: iterate through `.keys()` instead of `.items()`?

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -146,7 +147,19 @@ def validate(self, X, y):
...
+        after_to_before_inference_mapping = {
+            new: old for old, new in zip(original_vc.keys(), new_vc.keys())
+        }
+        before_to_after_inference_mapping = {
+            old: new for new, old in after_to_before_inference_mapping.items()
+        }
+        if str(y.ww.logical_type) not in ["Boolean", "BooleanNullable"]:
+            after_to_before_inference_mapping = {
+                old: old for old in after_to_before_inference_mapping.keys()
+            }
```
christopherbunn, line 162: This logic is a bit convoluted but I think I follow it. Can we move lines 159 - 162 up to before 153 and set it as an if/else for the `after_to_before_inference_mapping`?

christopherbunn (changes requested):
Mostly looks good! Have 1-2 q's but ready to approve once addressed.

I'm curious about is if we need to have a test case for the case you specified in the PR description? E.g. transforming a column of 0 and 1 into boolean values if explicitly set during `df.ww.init()`

evalml/data_checks/class_imbalance_data_check.py
ParthivNaresh: Sure thing, added some comments for clarity too

ParthivNaresh:
@christopherbunn Added a test to verify `Boolean` behaviour!

evalml/data_checks/class_imbalance_data_check.py
```diff
@@ -146,7 +147,23 @@ def validate(self, X, y):
...
+            # If the inferred logical type is not in Boolean/BooleanNullable, then a
+            # mapping to the original values is not necessary.
+            after_to_before_inference_mapping = {new: new for new in new_vc.keys()}
+        else:
+            # If the inferred logical type is in Boolean/BooleanNullable, then a
+            # mapping to the original values will be needed for the data check messages
+            after_to_before_inference_mapping = {
+                new: old for old, new in zip(original_vc.keys(), new_vc.keys())
+            }
+        # Needed for checking severe imbalance to verify values present below threshold
```
christopherbunn, line 163: Much more clear, thanks!

christopherbunn (approved):
LGTM

eccabay (approved):

PR #3896, opened 2022-12-19
Split STL `seasonal_period` into separate `seasonal_smoother` and `period`

eccabay:
Currently, in [evalml/stl_decomposer.py at main · alteryx/evalml](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py), the STLDecomposer mistakenly takes a value for seasonal and passes it to the seasonal parameters of the underlying STL algorithm.  It seems that seasonal is actually a value for a window that does smoothing (possibly for the trend decomposition).  This fixes that by splitting the `seasonal_period` parameter into separate `seasonal_smoother` (ideally a clearer name for this parameter) and `period`

### DIFF
--- a/evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
+++ b/evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
@@ -12,22 +12,23 @@ from evalml.tests.component_tests.decomposer_tests.test_decomposer import (
 
 
 def test_stl_decomposer_init():
-    delayed_features = STLDecomposer(degree=3, time_index="dates")
-    assert delayed_features.parameters == {
+    decomp = STLDecomposer(degree=3, time_index="dates")
+    assert decomp.parameters == {
         "degree": 3,
-        "seasonal_period": 7,
+        "period": None,
+        "seasonal_smoother": 7,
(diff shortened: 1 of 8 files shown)


christopherbunn (approved):

fjlanasa (approved):

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
```diff
@@ -12,22 +12,28 @@
...
-    stl = STLDecomposer(seasonal_period=3)
-    assert stl.seasonal_period == 3
 
-    stl = STLDecomposer(seasonal_period=4)
-    assert stl.seasonal_period == 5
+def test_stl_raises_warning_high_smoother(caplog, ts_data):
+    X, _, y = ts_data()
+    stl = STLDecomposer(seasonal_smoother=101)
+    stl.fit(X, y)
+    assert "STLDecomposer may perform poorly" in caplog.text
```
chukarsten, line 36: I suspect the behavior we want here is actually to warn off the `period` rather than the `seasonal_smoother` parameter here.

PR #3899, opened 2022-12-22
Fix broken links

eccabay:
Removes catboost from the links for our broken link checker since it's giving us 403 errors


fjlanasa (approved):
thanks!

christopherbunn (approved):

PR #3906, opened 2023-01-04
Add `dates_needed_for_prediction` to `TimeSeriesPipelineBase`

jeremyliweishih:
Adds `dates_needed_for_prediction` for time series pipelines. Erring on the side of caution and asking for a longer date period and will optimize further if necessary.

### DIFF
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -1,5 +1,4 @@
 """Pipeline base class for time-series problems."""
-import numpy as np
 import pandas as pd
 import woodwork as ww
 
@@ -296,19 +295,20 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
         """Return dates needed to forecast the given date in the future.
 
         Args:
-            date (np.datetime64,): Date to forecast in the future.
+            date (pd.Timestamp,): Date to forecast in the future.
 
         Returns:
-            dates_needed (tuple(np.datetime64)): Range of dates needed to forecast the given date.
+            dates_needed (tuple(pd.Timestamp)): Range of dates needed to forecast the given date.
         """
-        beginning_date = date - np.timedelta64(
+        print(self.frequency)
+        beginning_date = date - pd.Timedelta(
             self.forecast_horizon
             + self.max_delay  # include start delay for featurization
             + self.gap  # add first gap for the actual gap from the end date
             + self.gap,  # add another gap to ensure training data is greater than gap
             self.frequency,
         )
-        end_date = date - np.timedelta64(
+        end_date = date - pd.Timedelta(
             1 + self.gap,
             self.frequency,
         )
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -317,9 +317,9 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
         """
         beginning_date = date - np.timedelta64(
             self.forecast_horizon
-            + self.max_delay,  # include start delay for featurization
-            +self.gap,  # add first gap for the actual gap from the end date
-            +self.gap,  # add another gap to ensure training data is greater than gap
+            + self.max_delay  # include start delay for featurization
+            + self.gap  # add first gap for the actual gap from the end date
+            + self.gap,  # add another gap to ensure training data is greater than gap
             self.frequency,
         )
         end_date = date - np.timedelta64(
--- a/evalml/pipelines/time_series_pipeline_base.py
+++ b/evalml/pipelines/time_series_pipeline_base.py
@@ -98,7 +98,12 @@ class TimeSeriesPipelineBase(PipelineBase, metaclass=PipelineBaseMeta):
         gap_features = pd.DataFrame()
         gap_target = pd.Series()
         if (
-            are_datasets_separated_by_gap_time_index(X_train, X, self.pipeline_params)
+            are_datasets_separated_by_gap_time_index(
+                X_train,
+                X,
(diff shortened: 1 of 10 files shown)


evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -286,3 +306,25 @@ def _estimator_predict(self, features):
...
+        Args:
+            date (np.datetime64,): Date to forecast in the future.
+
+        Returns:
+            dates_needed (tuple(np.datetime64)): Range of dates needed to forecast the given date.
+        """
+        beginning_date = date - np.timedelta64(
+            self.forecast_horizon
+            + self.max_delay  # include start delay for featurization
+            + self.gap  # add first gap for the actual gap from the end date
```
jeremyliweishih, line 322: We have a limitation s.t we require the training data (using the dates from this method) to be larger than the gap. Thus, I'm directly adding the gap here so that we ensure our training data is always long enough. This can be optimized further by the dates needed to push the length of the training data over the gap but decided to err on the side of caution for the first pass. I believe its safer to ask the user for more data unless we know that it is hard for the user to provide us such data.

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -286,3 +306,25 @@ def _estimator_predict(self, features):
...
+            + self.gap,  # add another gap to ensure training data is greater than gap
```
jeremyliweishih, line 323: We have a limitation s.t we require the training data (using the dates from this method) to be larger than the gap. Thus, I'm directly adding the gap here so that we ensure our training data is always long enough. This can be optimized further by the dates needed to push the length of the training data over the gap but decided to err on the side of caution for the first pass. I believe its safer to ask the user for more data unless we know that it is hard for the user to provide us such data.

fjlanasa: A frequency of 'MS' (month start) here would be interpreted as milliseconds, right? https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1724,3 +1724,76 @@ def test_drop_time_index_woodwork(ts_data, time_series_regression_pipeline_class
...
+def test_dates_needed_for_prediction(
+    mock_predict,
+    mock_fit,
+    forecast_horizon,
+    gap,
+    max_delay,
+    ts_data,
+    time_series_regression_pipeline_class,
+):
+    X, X_t, y = ts_data()
```
eccabay, line 1745: Based on Frank's comment above, it'd probably be a good idea to test a few different frequencies here!

evalml/pipelines/time_series_pipeline_base.py
eccabay: Yeah, I think Frank is right here in that `MS` would be milliseconds. From my understanding, I think numpy defaults to milliseconds while pandas maps to month start for this.

jeremyliweishih: ah gotcha - I'll change to using a datetime timedelta!

evalml/pipelines/time_series_pipeline_base.py
```diff
@@ -286,3 +290,30 @@ def _estimator_predict(self, features):
         This helper passes y as an argument if needed by the estimator.
         """
         return self.estimator.predict(features)
+
+    def dates_needed_for_prediction(self, date):
+        """Return dates needed to forecast the given date in the future.
+
+        Args:
+            date (pd.Timestamp,): Date to forecast in the future.
```
eccabay, line 298: mega nit: unnecessary comma after Timestamp

eccabay (approved):

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1724,3 +1724,82 @@ def test_drop_time_index_woodwork(ts_data, time_series_regression_pipeline_class
...
+    pipeline.fit(X, y)
+
+    prediction_date = pd.Timestamp("2022-01-31")
+    beginning_date, end_date = pipeline.dates_needed_for_prediction(prediction_date)
+    assert beginning_date <= end_date
+    assert end_date < prediction_date
+    date_diff = pipeline.forecast_horizon + pipeline.max_delay + pipeline.gap
+    assert end_date == beginning_date + pd.tseries.frequencies.to_offset(
+        f"{date_diff}{pipeline.frequency}",
+    )
```
chukarsten, line 1777: Do we want to reparametrize this so that the `pytest` parameters contain the answers to the incoming inputs and just test them against each other?

jeremyliweishih: I think it'll get pretty messy including the answers with the different combinations of parameters and frequencies above. I could do so but I would need to narrow down the scope of combinations.

evalml/tests/pipeline_tests/test_time_series_pipeline.py
```diff
@@ -1724,3 +1724,82 @@ def test_drop_time_index_woodwork(ts_data, time_series_regression_pipeline_class
...
+    assert X_test.shape[0] == 1
+
+    X_train.ww.init()
+    X_test.ww.init()
+
+    X_train.ww.set_time_index("date")
+    X_test.ww.set_time_index("date")
+
+    _ = pipeline.predict(X_test, X_train=X_train, y_train=y_train).all()
+    assert not mock_predict.call_args[0][0].empty
```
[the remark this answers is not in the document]
jeremyliweishih: During implementation I created date ranges that were mathematically correct but didn't work entirely with our featurization so added this last part to ensure!

chukarsten (approved):

PR #3912, opened 2023-01-06
Improve automatic periodicity determination

eccabay:
Simplifies and improves performance of our `decomposer.determine_period` method by removing the need to pre-fit the model for detrending, using a moving average to detrend instead.


evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -126,75 +126,51 @@ def determine_periodicity(
...
+                may return either "7" or "365", depending on which seasonality is more strongly autocorrelated.  If no
+                period is detected, returns None.
 
         """
 
         def _get_rel_max_from_acf(y):
             """Determines the relative maxima of the target's autocorrelation."""
             acf = sm.tsa.acf(y, nlags=np.maximum(400, len(y)))
-            filter_acf = [acf[i] if (acf[i] > 0) else 0 for i in range(len(acf))]
+            filter_acf = [acf[i] if (acf[i] > 0.01) else 0 for i in range(len(acf))]
```
jeremyliweishih, line 149: curious why we need this change!

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -126,75 +126,51 @@ def determine_periodicity(
...
-
-        if method == "autocorrelation":
-            _get_rel_max = _get_rel_max_from_acf
-        elif method == "partial-autocorrelation":
-            self.logger.warning(
-                "Partial autocorrelations are not currently guaranteed to be accurate due to the need for continuing "
-                "algorithmic work and should not be used at this time.",
-            )
-            _get_rel_max = _get_rel_max_from_pacf
+            moving_avg = min(51, len(y) // 3)
```
jeremyliweishih, line 162: can you explain why `moving_avg` is calculated this way? comment might be useful here.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -126,75 +126,51 @@ def determine_periodicity(
...
-            return rel_max
-
-        def _get_rel_max_from_pacf(y):
-            """Determines the relative maxima of the target's partial autocorrelation."""
-            pacf = sm.tsa.pacf(y)
-            return argrelextrema(pacf, np.greater)[0]
+            return rel_max[np.argmax(max_acfs)]
 
         def _detrend_on_fly(X, y):
             """Uses the underlying decomposer to determine the target's trend and remove it."""
```
jeremyliweishih, line 161: this docstring should be updated since we're not using the decomposer anymore!

jeremyliweishih (approved):
Some clarification questions and quick fixes but otherwise looks good to me!

evalml/pipelines/components/transformers/preprocessing/decomposer.py
eccabay: Great question! We were seeing that datasets with no seasonality were still getting a result due to having some very small autocorrelations. This is to filter those out so we correctly say these datasets have no seasonality. Datasets with a seasonal period will be more on the scale of 0.1-0.6, so we aren't filtering out anything significant here.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
eccabay: Great catch!

evalml/pipelines/components/transformers/preprocessing/decomposer.py
chukarsten: We might want this 0.01 to be a variable that's defined in the decomposer.  I think that's probably a better practice.

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -444,27 +444,16 @@ def test_decomposer_set_period(decomposer_child_class, period, generate_seasonal
...
-        ),
-    ],
-)
 @pytest.mark.parametrize("decomposer_picked_correct_degree", [True, False])
 @pytest.mark.parametrize(
     "synthetic_data,trend_degree,period",
-    [*itertools.product(["synthetic"], [1, 2, 3], [7, 30, 365]), ("real", 1, 365)],
+    [*itertools.product(["synthetic"], [1, 2, 3], [1, 7, 30, 365]), ("real", 1, 365)],
 )
 def test_decomposer_determine_periodicity(
```
chukarsten, line 452: I hate to be a jerk, particularly because this is initially my failure to add - but can we include either enhancing this test or add a separate one calling out the no-seasonality results?

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -126,75 +126,54 @@ def determine_periodicity(
...
         def _get_rel_max_from_acf(y):
             """Determines the relative maxima of the target's autocorrelation."""
             acf = sm.tsa.acf(y, nlags=np.maximum(400, len(y)))
-            filter_acf = [acf[i] if (acf[i] > 0) else 0 for i in range(len(acf))]
+            # Filter out small values to avoid picking up noise
+            filter_acf = [acf[i] if (acf[i] > 0.01) else 0 for i in range(len(acf))]
             rel_max = argrelextrema(
                 np.array(filter_acf),
                 np.greater,
                 order=5,  # considers 5 points on either side to determine rel max
```
chukarsten, line 154: If you turn the ACF relative max threshold into a variable, can you pretty please clean up _my_ mess and do the same with order?

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -126,75 +126,54 @@ def determine_periodicity(
...
-        elif method == "partial-autocorrelation":
-            self.logger.warning(
-                "Partial autocorrelations are not currently guaranteed to be accurate due to the need for continuing "
-                "algorithmic work and should not be used at this time.",
-            )
-            _get_rel_max = _get_rel_max_from_pacf
+            """Uses a moving average to determine the target's trend and remove it."""
+            # A larger moving average will be less likely to remove the seasonal signal
+            # but we need to make sure we're not passing in a window that's larger than the data
+            moving_avg = min(51, len(y) // 3)
```
chukarsten, line 165: The 51 and 3 seem a little arbitrary and kind of trigger my "should this be hardcoded" sense, but I think in this case it's probably fine.

chukarsten (approved):
Just a few minor comments!  Nice work!!

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
eccabay: Ah, here I added the `1` to the period parametrization to represent that, since a seasonal period of 1 is identical to having no seasonal period. I'll switch this over to being explicitly `None` though.

PR #3914, opened 2023-01-10
Remove non-selected-categorical columns from dataframe used to create numeric pipelines in DefaultAlgo

jeremyliweishih:
If a dataframe contained a categorical column that was not selected by the feature selector, ``DefaultAlgorithm`` would send a dataframe with such categorical column to the make pipeline util for the numeric pipeline. This would cause a OHE to be inserted into the numeric pipeline. This change makes it so that a dataframe matching the parameters in the `SelectByType` is sent to the pipeline util instead.


eccabay (approved):
Looks good!

ParthivNaresh (approved):
Thank you!

PR #3932, opened 2023-01-17
Fixed `set_period()` not updating decomposer parameters

### LINKED ISSUE #3771: Decomposer does not update parameters dictionary when set_seasonal_period is called.
There's some weird gotcha with the parameters dictionary in the Decomposer class that needs to be fixed so that `set_seasonal_period` can set that parameter dictionary correctly.

```python
from evalml.pipelines.components.transformers.preprocessing.stl_decomposer import STLDecomposer

stl = STLDecomposer()
stl.fit(X,y)

stl.set_seasonal_period(X,y)

assert stl.parameters["seasonal_period"] != -1 # the default
```

christopherbunn:
Also introduces a way to update parameters in a component.

Resolves #3771

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/decomposer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/decomposer.py
@@ -205,7 +205,7 @@ class Decomposer(Transformer):
 
         """
         self.period = self.determine_periodicity(X, y)
-        self.parameters.update({"period": self.period})
+        self.update_parameters({"period": self.period})
 
     def _check_oos_past(self, y):
         """Function to check whether provided target data is out-of-sample and in the past."""
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,6 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
     * Fixes
+        * Fixed `set_period()` not updating decomposer parameters :pr:`3932`
     * Changes
         * Updated ``PolynomialDecomposer`` to work with sktime v0.15.1 :pr:`3930`
         * Pinned `category-encoders`` to 2.5.1.post0 :pr:`3933``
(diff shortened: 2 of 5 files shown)


evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -205,7 +205,7 @@ def set_period(self, X: pd.DataFrame, y: pd.Series):
 
         """
         self.period = self.determine_periodicity(X, y)
-        self.parameters.update({"period": self.period})
+        self.update_parameters({"period": self.period})
```
jeremyliweishih, line 208: can you add a test verifying the period updates?

jeremyliweishih (approved):
LGTM but would be nice to have an explicit test for the decomposer.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
eccabay: There's a line in `test_decomposer.py::test_decomposer_set_period` that's currently commented out testing this, you should be able to uncomment that!

eccabay (approved):
Looks good! I agree with Jeremy about the testing. Have you verified if there's anywhere else we update parameters that could benefit from this function?

evalml/pipelines/components/transformers/preprocessing/decomposer.py
christopherbunn: I uncommented that line, I think that should be sufficient? @jeremyliweishih I can also break it out into its own separate test if need be.

christopherbunn:
@eccabay I did a quick ctrl+f search for places that update parameters and this should be it. I think this is the only place where were updating the parameters from inside the component itself.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
jeremyliweishih: @christopherbunn thats great!

PR #3934, opened 2023-01-17
Add Recursive Feature Elimination Selector component

thehomebrewnerd:
### Add Recursive Feature Elimination Selector component

Adds a new feature selector component that uses scikit-learn's RFECV to perform the selection.


evalml/pipelines/components/transformers/feature_selection/recursive_feature_elimination_selector.py
```diff
@@ -0,0 +1,128 @@
...
+class RecursiveFeatureEliminationSelector(FeatureSelector):
+    """Selects relevant features using recursive feature elimination."""
+
+    hyperparameter_ranges = {
+        "perc": Integer(0, 100),
+    }
+    """{
+        "percent_features": Real(0.01, 1),
+        "threshold": ["mean", "median"],
+    }"""
```
rwedge, line 23: I think we'd want `min_features_to_select` and `step` here but not positive on how this section is mean to be set up

thehomebrewnerd: Yeah, good catch. I overlooked that, but will take a closer look and update.

jeremyliweishih: I think we should set up `step` between 0 and 1 so it selects a eliminates a percentage amount (so we can generalize more easily than an integer) but I don't think we need to tune `min_features_to_select` since RFECV is supposed to find the optimal amount of features. I think a default `min_features_to_select=1` is fine unless there's a reason we want RFE to keep more.

jeremyliweishih (changes requested):
Just need to fix up the hyperparameter ranges but LGTM otherwise!

evalml/pipelines/components/transformers/feature_selection/recursive_feature_elimination_selector.py
thehomebrewnerd: I updated to have step between 0.05 and 0.25 and changed the default for `min_features_to_select` to `1`.

jeremyliweishih (approved):
LGTM

gsheni (approved):

PR #3954, opened 2023-01-24
Added demo for prediction intervals

christopherbunn:
Also includes a demo for forecasting that plots the prediction intervals.


docs/source/user_guide/timeseries.ipynb
```diff
@@ -935,6 +1108,11 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.8.6"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "fb5a3afe2d0dd7ad0fd9e1ff89de4e0e95804490c629f36065bf8d930a66d311"
+   }
```
eccabay, line 1115: We should really have an automation or at least a check to strip these out.

docs/source/user_guide/timeseries.ipynb
```diff
@@ -849,6 +849,179 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "While predictions that are generated by EvalML pipelines aim to be accurate as possible, it is very rarely the case that future results are the exact same values as predicted. Prediction intervals can help to contextualize a prediction by showing the range a future prediction is expected to fall within a certain likelihood. \n",
+    "\n",
+    "Given a set of predictions and a **trained** EvalML pipeline, the prediction intervals for this set of predictions is generated by calling `get_prediction_intervals()` on the pipeline's estimator:"
```
eccabay, line 875: This is an important callout, but could you rephrase it slightly to mention that we don't just need the trained pipeline, we specifically need the transformed, ready for prediction features?

I'm also very out of the loop on prediction intervals, why do we need the transformed features?

eccabay (approved):
Looks great! Just a couple small notes.

docs/source/user_guide/timeseries.ipynb
christopherbunn: Updated to clarify that only a fitted estimator is needed and that the example is using the fitted estimator in a trained pipeline. 

We need the transformed features as we're calling `get_prediction_intervals()` from the estimator directly. It expects to take input in that has already been through the preprocessing steps of the pipeline. We currently don't have a way to get prediction intervals at the pipeline level.

jeremyliweishih (approved):
This looks great! Thanks @christopherbunn

PR #3959, opened 2023-01-26
Fix BooleanNullable SimpleImputer bug

eccabay:
Fixes the bug where all-null BooleanNullable columns will break the simple imputer during transform, when fit on nullable data that has a non-null value.

### DIFF
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -85,9 +85,13 @@ class SimpleImputer(Transformer):
         X, _ = drop_natural_language_columns(X)
 
         # Convert any boolean columns to IntegerNullable, but keep track of the columns so they can be converted back
-        self._boolean_cols = X.ww.schema._filter_cols(
-            include=["Boolean", "BooleanNullable"],
+        self._boolean_cols = list(
+            X.ww.select(
+                include=["Boolean", "BooleanNullable"],
+                return_schema=True,
+            ).columns,
         )
+        # import pdb; pdb.set_trace()
         # Make sure we're tracking Categorical columns that should be boolean as well
         self._boolean_cols.extend(
             [
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -91,7 +91,6 @@ class SimpleImputer(Transformer):
                 return_schema=True,
             ).columns,
         )
-        # import pdb; pdb.set_trace()
         # Make sure we're tracking Categorical columns that should be boolean as well
         self._boolean_cols.extend(
             [
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -10,6 +10,7 @@ from evalml.pipelines.components.utils import (
     set_boolean_columns_to_integer,
 )
 from evalml.utils import infer_feature_types
+from evalml.utils.gen_utils import is_categorical_actually_boolean
 
 
 class SimpleImputer(Transformer):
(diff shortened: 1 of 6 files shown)


evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -78,11 +78,25 @@ def fit(self, X, y=None):
...
         nan_ratio = X.isna().sum() / X.shape[0]
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
 
         X, _ = drop_natural_language_columns(X)
-        X = set_boolean_columns_to_categorical(X)
+
+        # Convert any boolean columns to IntegerNullable, but keep track of the columns so they can be converted back
+        self._boolean_cols = X.ww.schema._filter_cols(
+            include=["Boolean", "BooleanNullable"],
+        )
```
chukarsten, line 90: I might be being pedantic, but is the preferred way to call a private function on the schema?  I thought there was a select function on the ww accessor?

eccabay: Certainly. I stole this line directly from `set_boolean_columns_to_integer`, but I can switch both places over to use `select` instead.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -124,11 +134,9 @@ def transform(self, X, y=None):
 
         new_schema = original_schema.get_subset_schema(X_t.columns)
 
-        # TODO: Fix this after WW adds inference of object type booleans to BooleanNullable
```
chukarsten, line 127: Yay!

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -487,11 +487,10 @@ def test_simple_imputer_woodwork_custom_overrides_returned_by_components(
     except ww.exceptions.TypeConversionError:
         return
 
-    impute_strategy_to_use = impute_strategy
-    if logical_type in [NaturalLanguage, Categorical]:
-        impute_strategy_to_use = "most_frequent"
+    if logical_type in [NaturalLanguage, Categorical, Boolean, BooleanNullable]:
+        impute_strategy = "most_frequent"
```
chukarsten, line 491: Not a huge fan of how this was originally done - with `impute_strategy` iterating over a subset of the total `impute_strategy` available and changing it in the test.   But that's not your problem...we might want to think about rewriting this.

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -608,3 +607,19 @@ def test_simple_imputer_errors_with_bool_and_categorical_columns(
     else:
         si = SimpleImputer()
         si.fit(X_df)
+
+
+def test_simple_imputer_boolean_nullable_all_null():
+    X_train = pd.DataFrame({"a": [pd.NA] * 20 + [1.0] + [pd.NA] * 20})
+    y = pd.Series(range(len(X_train)))
+    X_test = pd.DataFrame({"a": [pd.NA] * 10})
```
chukarsten, line 615: Times like these, I think it's helpful to docstring the test to get at what exactly you're testing here.  The test name doesn't seem to match what's going on.  The test case here is that you're train is sparse and your test set happens to not be fully representative of all the classes available in `X`, right?

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -608,3 +607,19 @@ def test_simple_imputer_errors_with_bool_and_categorical_columns(
...
+
+    X_train.ww.init(logical_types={"a": "boolean_nullable"})
+    X_test.ww.init(logical_types={"a": "boolean_nullable"})
+
+    imp = SimpleImputer()
+    imp.fit(X_train, y)
+
+    X_t = imp.transform(X_test)
+    assert not X_t["a"].isna().any()
+    assert X_t.ww.logical_types["a"].type_string == "boolean_nullable"
```
chukarsten, line 625: Is this our preferred methodology for type checking in WW?  Kind of nitting, kind of asking for a friend.

chukarsten (approved):
Pending select change, looks great!

evalml/tests/component_tests/test_simple_imputer.py
jeremyliweishih:
```suggestion
    assert isinstance(X_t.ww.logical_types["a"], BooleanNullable)
```

jeremyliweishih (approved):
Agree with @chukarsten's comments but otherwise LGTM

evalml/tests/component_tests/test_simple_imputer.py
eccabay: A good point - I've taken Jeremy's suggestion which is certainly the better way to check!

evalml/tests/component_tests/test_simple_imputer.py
eccabay: Basically, yeah. It's really just testing having an all-null test set when the training had non-null values. I'll update the test name and add a docstring

PR #3961, opened 2023-01-26
Release v0.66.1

chukarsten:
# v0.66.1 Jan. 26, 2023
### Fixes
- Updated ``LabelEncoder`` to store the original typing information #3960
- Fixed bug where all-null ``BooleanNullable`` columns would break the imputer during transform #3959


jeremyliweishih (approved):
🚢

PR #3964, opened 2023-01-30
Skip featurization for time series native models

eccabay:
Re-adding #3849 after woodwork's schema bug was resolved, now with a test to ensure we shouldn't encounter the same issue again.


jeremyliweishih (approved):
LGTM

Issue #3965, opened 2023-01-30
CatBoost Estimators should be able to handle float categories 


tamargrey:
- As a user, I wish I could pass categorical data whose categories are floats into the CatBoostRegressor and CatBoostClassifier estimators. They currently raise `Invalid type for cat_feature category for [feature_idx=312]=1.0 : cat_features must be integer or string, real number values and NaN values should be converted to string.` with the code snippet below:

#### Code Example

```python
    import woodwork as ww
    import pandas as pd

    X = pd.DataFrame({"double_cats": pd.Series([1.0, 2.0, 3.0, 4.0, 5.0]*20)})
    y = pd.Series(range(100))
    y.ww.init()
    X.ww.init(logical_types={"double_cats": "Categorical"})

    clf = CatBoostRegressor(
        n_estimators=1,
        max_depth=1,
    )
    clf.fit(X, y)

```

In the case above the floats can be converted to ints, but we should also handle the case where we cannot convert the floats to ints without truncating the data (we can consider converting to the `object` dtype in that case)

tamargrey:
See [this comment](https://github.com/alteryx/evalml/pull/3966#discussion_r1091047373) for more info, but for now we will not be handling the case where actual floats (like 1.1, 2.5, etc) are used, as there's not a clean way to convert the values in a way CatBoost will take, and according to their [FAQ](https://catboost.ai/en/docs/concepts/faq#floating-point-values), we might not want to use stringified floats anyway. 

If either https://github.com/catboost/catboost/issues/1965 or https://github.com/pandas-dev/pandas/issues/51074 is completed and we have updated to those CatBoost or Pandas versions, we could update `handle_float_categories_for_catboost` to convert floats as well. Hopefully by the time a need for that arises, one of those is fixed.

PR #3966, opened 2023-01-30
Allow float categories to be passed into CatBoost estimators

tamargrey:
fixes https://github.com/alteryx/evalml/issues/3965

Since catboost cannot handle float categories, converts categories that are really integers to be `int64`

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/catboost_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/catboost_regressor.py
@@ -119,6 +119,19 @@ class CatBoostRegressor(Estimator):
         self._component_obj.fit(X, y, silent=True, cat_features=cat_cols)
         return self
 
+    def predict(self, X):
+        """Make predictions using the fitted CatBoost regressor.
+
+        Args:
+            X (pd.DataFrame): Data of shape [n_samples, n_features].
+
+        Returns:
+            pd.DataFrame: Predicted values.
+        """
+        X = handle_float_categories_for_catboost(X)
+        predictions = super().predict(X)
+        return predictions
+
     @property
     def feature_importance(self):
         """Feature importance of fitted CatBoost regressor."""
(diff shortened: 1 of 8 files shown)


evalml/pipelines/components/estimators/regressors/catboost_regressor.py
```diff
@@ -113,9 +114,25 @@ def fit(self, X, y=None):
         self.input_feature_names = list(X.columns)
         X, y = super()._manage_woodwork(X, y)
         X = downcast_int_nullable_to_double(X)
+
+        X = handle_float_categories_for_catboost(X)
         self._component_obj.fit(X, y, silent=True, cat_features=cat_cols)
         return self
 
+    def predict(self, X):
```
tamargrey, line 122: Figured if I was going to handle categories for the regressor (and I did double check it was necessary for `predict`!) I'd have to add an explicit predict method -- let me know if there's a better way that I missed though!

evalml/tests/component_tests/test_utils.py
```diff
@@ -149,7 +152,9 @@ class NonComponent:
 
 
 def test_scikit_learn_wrapper_invalid_problem_type():
-    evalml_pipeline = MulticlassClassificationPipeline([RandomForestClassifier])
+    evalml_pipeline = MulticlassClassificationPipeline(
+        [RandomForestClassifier],
+    )
```
tamargrey, line 157: I might be in a weird state in VSCode with the new ruff linting, so idk if this and the above change are expected or indicative of my own weird environment. Doesn't seem like I have any linting issues in this PR so I'm fine leaving as is.

chukarsten (approved):

evalml/pipelines/components/estimators/regressors/catboost_regressor.py
jeremyliweishih: looks good!

jeremyliweishih (approved):
LGTM - nice work!

PR #3967, opened 2023-01-30
Update pyproject.toml to correctly specify the data filepaths

gsheni:
- The install test workflow packages evalml (creates whl/tar.gz file), install that whl/tar.gz, and tries to import the datasets (to make sure it is distributed with evalml).
- The commands in meta.yaml also try to import the datasets (to make sure conda has the files).

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -56,8 +56,8 @@ outputs:
         - pyproject.toml
       commands:
         - pytest evalml/tests -n 8 --ignore evalml/tests/automl_tests/parallel_tests/ --is-using-conda
-        - python -c "from evalml.demos.fraud import load_fraud; load_fraud(nrows=10)"
-        - python -c "from evalml.demos.churn import load_churn; load_churn(nrows=10)"
+        - python -c "from evalml.demos import load_fraud; load_fraud(nrows=10)"
+        - python -c "from evalml.demos import load_churn; load_churn(nrows=10)"
 
   - name: evalml
     build:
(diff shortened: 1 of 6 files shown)


.github/meta.yaml
```diff
@@ -56,6 +56,8 @@ outputs:
         - pyproject.toml
       commands:
         - pytest evalml/tests -n 8 --ignore evalml/tests/automl_tests/parallel_tests/ --is-using-conda
+        - python -c "from evalml.demos import load_fraud; load_fraud(nrows=10)"
```
jeremyliweishih, line 59: is this for testing purposes? If so we should remove it 😄

jeremyliweishih (approved):
LGTM!

.github/meta.yaml
gsheni: No it is not for testing purposes. It is to make sure that we distribute the CSV file with evalml.

pyproject.toml
```diff
@@ -126,12 +126,14 @@ namespaces = true
...
-"evalml.tests.data" = [
-    "evalml/tests/data/churn.csv",
-    "evalml/tests/data/daily-min-temperatures.csv",
-    "evalml/tests/data/fraud_transactions.csv.gz",
-    "evalml/tests/data/tips.csv",
-    "evalml/tests/data/titanic.csv",
+"evalml" = [
+    "demo/churn.csv",
+    "demo/daily-min-temperatures.csv",
+    "demo/fraud_transactions.csv.gz",
```
gsheni, line 132:
```suggestion
    "demo/data/churn.csv",
    "demo/data/daily-min-temperatures.csv",
    "demo/data/fraud_transactions.csv.gz",
```

chukarsten (approved):

PR #4024, opened 2023-02-21
Add-handling-utils

### LINKED ISSUE #3990: Add `_handle_nullable_types` and properties indicating incompatibilities to ObjectiveBase and ComponentBase Classes
- As the first step in implementing component-specific handling of nullable types, we will need to add the `_handle_nullable_types` method as well as `_integer_nullable_incompatibilities` and `_boolean_nullable_incompatibilities` to the base objective and component classes. 

This will require us to create and independently test the nullable type utilities we will need. We will not yet begin adding logic to the individual components that need it or begin calling the handling utils.

### Component 

```python
# General component nullable type logic
class ComponentBase(ABC, metaclass=ComponentBaseMeta):
    # Default to components not having any incompatibilities 
    _integer_nullable_incompatibilities = None
    _boolean_nullable_incompatibilities = None
    
    # Method for handling incompatibilities 
    def handle_nullable_types(self, X, y=None):
        # Looks at _integer_nullable_incompatibilities 
        # and _boolean_nullable_incompatibilities to determine what transformations to do
    


```

### Objective Base 
```python
# General objective nullable type logic
class ObjectiveBase(ABC):
    # Default to objectives not having any incompatibilities 
    _integer_nullable_incompatible = False
    _boolean_nullable_incompatible = False
    
    def handle_nullable_types(y):
        ...
```

tamargrey:
### Pull Request Description
closes #3990

Adds the utilities needed for component-specific handling of nullable types and starts adding related properties and methods to the base component and objective classes.


evalml/objectives/binary_classification_objective.py
```diff
@@ -81,3 +88,25 @@ def validate_inputs(self, y_true, y_predicted):
...
+            y_true (pd.Series): Actual class labels of length [n_samples].
+                May contain nullable types.
+
+        Returns:
+            y_true with any incompatible nullable types downcasted to compatible equivalents.
+        """
+        # Since Objective functions dont have the same safeguards around non woodwork inputs,
+        # we'll choose to avoid the downcasting path since we shouldn't have nullable pandas types
+        # without them being set by Woodwork
+        if isinstance(y_true, pd.Series) and y_true.ww.schema is not None:
```
tamargrey, line 105: Wanted to highlight this behavior. At least in tests, we often call the objective functions with inputs that don't actually have woodwork types, and the objective functions don't have much in the way of logic to initialize woodwork. Given how late in the pipeline these are used and the fact that the nulalble type or not doesn't have any downstream implications here, I decided to just not initialize Woodwork if the data didn't already have types. No reason to waste the time with type inference when the non woodwork inputs should always be compatible anyway.

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        X (pd.DataFrame): Input data of shape [n_samples, n_features] whose nullable types will be changed.
+        handle_boolean_nullable (bool, optional): Whether or not to downcast data with BooleanNullable logical types.
+        handle_integer_nullable (bool, optional): Whether or not to downcast data with IntegerNullable or AgeNullable logical types.
+
+
+    Returns:
+        X with any incompatible nullable types downcasted to compatible equivalents.
+    """
+    # --> consider adding param for expecting there to not be any nans present so we're
+    # notified if we're ever unknowingly converting to Double or Categorical when we shouldnt in automl search
```
tamargrey, line 19: Just a small behavior consideration that I am not going to implement unless/until a need for it arises. Likely, it'd happen during the integration into AutoMLSearch.

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Returns:
+        LogicalType string to be used to downcast incompatible nullable logical types.
+    """
+    # --> maybe this can be configurable so we could easily choose different values to downcast to for specific components?
+    downcast_matches = {
+        "BooleanNullable": ("Boolean", "Categorical"),
+        "IntegerNullable": ("Integer", "Double"),
+        # --> age fractional or double? I think AgeFractional to avoid losing info
+        "AgeNullable": ("Age", "AgeFractional"),
```
tamargrey, line 100: We were previously converting to Integer/Double for age nullable columns, but as far as I can tell, there's no reason to not maintain this information when possible!

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Args:
+        nullable_logical_type (str): String representation of the Woodwork LogicalType to downcast
+        data_has_nans (bool): Whether or not nans were present in the data.
+            Determines whether a non nullable LogicalType can be used for downcasting or not.
+
+    Returns:
+        LogicalType string to be used to downcast incompatible nullable logical types.
+    """
+    # --> maybe this can be configurable so we could easily choose different values to downcast to for specific components?
```
tamargrey, line 95: Another idea that might prove useful to implement at some point in the future. Say a component needed to convert from BooleanNullable to IntegerNullable for some reason, it'd be cool if we could leave the actual downcasted logical type decisions to the components. 

This would add a level of complexity, though, that we shouldnt introduce unless necessary.

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    select_param = []
+    if handle_boolean_nullable:
+        select_param.append("BooleanNullable")
+    if handle_integer_nullable:
+        select_param.append("IntegerNullable")
+        select_param.append("AgeNullable")
+
+    if not select_param:
+        return X
```
eccabay, line 28: I think this check is redundant, we can check for the same scenario with `X.ww.select([])` below, right?

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    cols_with_nans = set(
+        nullable_X_to_downcast.columns[nullable_X_to_downcast.isnull().any()],
+    )
+
+    new_ltypes = {
+        col: _get_downcast_logical_type(ltype, col in cols_with_nans)
+        for col, ltype in nullable_X_to_downcast.ww.logical_types.items()
+    }
```
eccabay, line 41: I'm having a bit of trouble understanding what's happening here. Are we leaving nullable types in the dataframe if there are no nans? Is that the desired behavior?

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,103 @@
+import woodwork as ww
+
+
+def _downcast_nullable_X(X, handle_boolean_nullable=True, handle_integer_nullable=True):
```
eccabay, line 4: There's a lot of similar logic here between this and `_downcast_nullable_y`. I think it would make sense to combine these into a single function, and separate the X/y logic based on whether the input is a DataFrame or a Series. 

It feels a bit more flexible to have a single function rather than two separate ones, and it would be able to condense down the number of tests in `test_nullable_type_utils.py`, since there's a lot of repeated logic between X and y tests there as well.

evalml/utils/nullable_type_utils.py
eccabay: What's the underlying dtype of AgeFractional?

eccabay (changes requested):
This is some awesome work! Most of my comments are related to condensing logic - there's a lot that's repeated between X and y, and I think it would be cleaner to combine them a bit.

evalml/objectives/binary_classification_objective.py
```diff
@@ -1,15 +1,22 @@
...
 
 class BinaryClassificationObjective(ObjectiveBase):
     """Base class for all binary classification objectives."""
 
     problem_types = [ProblemTypes.BINARY, ProblemTypes.TIME_SERIES_BINARY]
+
+    # Referring to the pandas nullable dtypes; not just woodwork logical types
+    _integer_nullable_incompatible = False
+    _boolean_nullable_incompatible = False
+
```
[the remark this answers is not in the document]
tamargrey: Good question--0thanks for pointing it out! It's only binary classification objectives that have incompatibilities. Originally, I was going to put it in ObectiveBase, but  decided that it'd be better to go as specific as possible since the call to `_handle_nullable_types` will get called in `objective_function` (not `score` as I'd been planning at one point). 

I think my only holdup with having just BinaryClassificationObjectives have these properties is that then the others don't have any concept of nullable type compatibility. They don't really need to, which is why I decided to get as specific as possible, but I'm very open to changing it if you or someone else feels strongly.

evalml/objectives/binary_classification_objective.py
```diff
@@ -81,3 +88,25 @@ def validate_inputs(self, y_true, y_predicted):
             raise ValueError("y_true contains more than two unique values")
         if len(np.unique(y_predicted)) > 2 and not self.score_needs_proba:
             raise ValueError("y_predicted contains more than two unique values")
+
+    def _handle_nullable_types(self, y_true):
```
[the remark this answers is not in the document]
tamargrey: Yep, it'll be called in `objective_function` and before we we apply the underlying function to the data (the function is what's incompatible with nullable types - https://github.com/scikit-learn/scikit-learn/issues/25578).

evalml/tests/component_tests/test_components.py
```diff
@@ -1804,3 +1807,26 @@ def test_component_parameters_supported_by_list_API(component_class):
...
+
+
+def test_handle_nullable_types(
+    test_classes,
+    nullable_type_test_data,
+    nullable_type_target,
+):
+    MockComponent, _, _ = test_classes
+    y = nullable_type_target(ltype="IntegerNullable", has_nans=False)
+    X = nullable_type_test_data(has_nans=False)
```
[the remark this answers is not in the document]
tamargrey: 100% agree! I will look into the best way to test. Right now, I'm just thinking that it might be better to make a separate fixture that can take inputs instead of further bastardizing the `test_classes` like I am right now, which are used for much much simpler tests elsewhere in this file.

evalml/tests/conftest.py
```diff
@@ -2226,6 +2241,102 @@ def imputer_test_data():
     return X
 
 
+@pytest.fixture
+def nullable_type_test_data():
```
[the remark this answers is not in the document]
tamargrey: 100% agree! My plan is to have it be configurable similar to the `has_nans` param to decide if users want `Age` to be present (since that, and a small change to call `cool col with nan` `cool with nan` to match the format of the other nan cols, are the only differences between the two fixtures.

evalml/tests/conftest.py
```diff
@@ -2226,6 +2241,102 @@ def imputer_test_data():
...
+    def _build_nullable_type_data(has_nans=True):
```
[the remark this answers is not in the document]
tamargrey: Yep! The fixture is `nullable_type_test_data` and instead of returning a ready to use DataFrame or Series, it just returns the function `_build_nullable_type_data`, which is what can take in the param. I haven't had to use `has_nans` in this or the other fixture just yet (its intended purpose is to make it easy to pass nullable data types into components that can't handle nans in the future), but in `nullable_type_target`, I use the `ltype` param quite a lot. 

I took the idea for this from `make_data_type` which follows the same pattern.

evalml/tests/conftest.py
```diff
@@ -2083,6 +2083,21 @@ def objective_function(self, y_true, y_predicted, X=None):
...
+
+    name = "my_objective"
+    greater_is_better = True
+    score_needs_proba = False
+    perfect_score = 1.0
+    is_bounded_like_percentage = True
+    expected_range = [0, 1]
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
+        pass
```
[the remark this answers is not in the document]
tamargrey: oooh, I like that quite a lot! I'll make that change

evalml/utils/nullable_type_utils.py
tamargrey: `float64`! So it's pretty much the same thing as what we're doing to IntegerNullalbe but just with information maintained

(source: https://github.com/alteryx/woodwork/blob/main/woodwork/logical_types.py#L130-L141)

evalml/utils/nullable_type_utils.py
tamargrey: Correct--the benefit is just that we wouldn't have to do `select` since we'd already know there was nothing problematic. This is me maybe over engineering a bit here for performance, thinking about a case where we have a very wide dataset that we'd be happy to avoid the `select` call on.

But I think in reality, we're not going to be calling `_downcast_nullable_X` on data that has no nullable type incompatibilities. This is because of a change I made in `ComponentBase._handle_nullable_types`; originally I had us calling the downcast function on `X` every time even if it had no incompatibilities, but I changed it and forgot to remove this logic. Will remove!

evalml/utils/nullable_type_utils.py
tamargrey: Fair question -- I'll take a look and try to refactor to have this be easier to read.

So we leave nullable types in the dataframe if they are compatible (what `handle_boolean_nullable` and `handle_integer_nullable` decide).

We then gather the incompatible types into `nullable_X_to_downcast` and decide what logical type to convert them to based off of whether any nans are present (`_get_downcast_logical_type`). So for `IntegerNullable`, they can become the non nullable `Integer` if no nans are present but have to become `Double` if nans are present. Either way, though, we're removing all `IntegerNullable` uses.

evalml/utils/nullable_type_utils.py
tamargrey: I don't disagree about the shared logic. I generally like utilities to not do too many things (it helps me with readability of code and how I think about tests but does, indeed, end up with more tests, and I totally see where you're coming from that these tests are very similar.)

I'm gonna play around with different ways of sharing more logic among the separate utils vs combining them and see if I can put something together that doesn't feel as repetitive. One thing may be to test the shared logic separately, so that the tests for the two utils are really only testing the differing apis for series vs dataframes.

evalml/objectives/binary_classification_objective.py
tamargrey: A user could pass them in if they were using the objective functions directly (I don't think we're at risk of them getting passed in with woodwork types from automl search).

The good news is, we can't get pandas nullable types in numpy inputs, so I think we'd cover user inputs by changing this to pass pandas data without woodwork types to the downcast utils.

evalml/tests/objective_tests/test_objectives.py
```diff
@@ -189,3 +190,25 @@ def test_is_defined_for_problem_type():
...
+def test_binary_objective_handle_nullable_types(nullable_type_target, nullable_ltype):
+    y_true = nullable_type_target(ltype=nullable_ltype, has_nans=False)
+
+    obj = CustomBinaryClassificationObjective()
+    y_true_d = obj._handle_nullable_types(y_true)
+    assert str(y_true_d.ww.logical_type) != nullable_ltype
+
+
+@pytest.mark.parametrize("data_type", ["np", "pd"])
+def test_binary_objective_handle_nullable_types_non_ww_input(make_data_type, data_type):
```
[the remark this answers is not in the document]
tamargrey: I was considering this, and we have ways of setting IntegerNullable and BooleanNullalbe via `make_data_type`, but there's not a great way to set AgeNullable without making larger changes to the fixture (it does a ww.init_series call, which would remove the AgeNullable logical type), and I'm worried it'd cause a bunch of unrelated changes. 

I think also with a change from a comment above to also pass pandas series without woodwork types into the downcast util, this will change these tests anyway, so I'll figure out how we can parametrize with all three `data_type`s even if it doesn't make sense to use `make_data_type` with all three of them.

evalml/utils/nullable_type_utils.py
tamargrey: Removed: 1ad54d33c76c681af0e187d7dbca5f3066c63402

evalml/utils/nullable_type_utils.py
tamargrey: @eccabay I pulled out more shared logic in the utils (along with some variable renaming) and refactored some tests to not be so repetitive across X and y downcasters: https://github.com/alteryx/evalml/pull/4024/commits/601535b3826ff0c85592daf9722640639657f688. I still feel like it makes sense to not squish them into one util with an if/else block since the woodwork dataframe and series apis are different enough that the two blocks of code wouldn't change much if they were in a single util at this point. It just puts the onus on anyone trying to understand the util in the future to understand the scope of the util's abilities.

If we decide to go with one util in the end, I would keep the tests the same and just remove the `downcast_util` parameter in fixtures, so making the change to one util will be very quick, and I can definitely be convinced that that's the right path.

I chose not to refactor tests that would need branching logic at both the setup and assertions, as I feel like that is a good indicator of when two checks really deserve their own tests.

evalml/objectives/binary_classification_objective.py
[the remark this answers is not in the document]
eccabay: Makes sense. I'm inclined to go broader rather than more specific, for the reason you mentioned. It's not a guarantee that all of our non-binary objectives will remain compatible with nullable types, so it's safer to make it more broad. I don't think where `_handle_nullable_types` is called should have an impact, since we'd only need to call it whenever we have an incompatible objective.

evalml/utils/nullable_type_utils.py
tamargrey: I attempted to rename things better and simplify some of the logic in https://github.com/alteryx/evalml/pull/4024/commits/601535b3826ff0c85592daf9722640639657f688 that will hopefully make the utils easier to understand. Let me know if anything still feels confusing

evalml/pipelines/components/component_base.py
```diff
@@ -239,3 +243,33 @@ def update_parameters(self, update_dict, reset_fit=True):
...
         if reset_fit:
             self._is_fitted = False
+
+    def _handle_nullable_types(self, X=None, y=None):
+        """Transforms X and y to remove any incompatible nullable types according to a component's needs.
+
+        Args:
+            X (pd.DataFrame, optional): Input data to a component of shape [n_samples, n_features].
+            y (pd.Series): The target of length [n_samples]. May contain nullable types.
+
```
eccabay, line 253: Copypasta nit: I don't think the `optional` only in the X is correct here, and I don't think y is the only argument that can contain nullables!

eccabay (approved):
Approved pending the remaining comments! Overall, this looks very solid and I'm excited for the final result.

evalml/tests/conftest.py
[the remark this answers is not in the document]
tamargrey: Two things that kept me from implementing this suggestion:
1. The codecove issue can be avoided by replacing `pass` with a string (as the mock above this does - I didntknow why it did that, and now I do lol)
2. testing `_handle_nullable_types` via `objective_function` would only work if we were returning the result of `_handle_nullable_types` in that function. Since we are just using the methods outputs to run the objective function, it wouldn't make sense to test it this way. Once we get to actually calling `_handle_nullable_types` inside of the objective functions, I'll make sure to mock it out and confirm it does get called when we need it, but that wouldn't be part of this PR.

evalml/objectives/binary_classification_objective.py
tamargrey: Allowed pandas series without woodwork types in this commit e5e7a529b

evalml/objectives/binary_classification_objective.py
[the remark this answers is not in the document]
tamargrey: Moved nullable type logic to ObjectiveBase 6ddbf4745

evalml/tests/conftest.py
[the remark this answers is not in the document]
tamargrey: Reused available info between the imputer fixture and this one: 5b34fc851

evalml/tests/objective_tests/test_objectives.py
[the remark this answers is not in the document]
tamargrey: After the changes to allow pandas series without woodwork info to be passed into the downcast util (e5e7a529b), I still feel like the tests are different enough that it doesn't make sense to combine them.

evalml/utils/nullable_type_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    Returns:
+        LogicalType string to be used to downcast incompatible nullable logical types.
+    """
+    downcast_matches = {
+        "BooleanNullable": ("Boolean", "Categorical"),
+        "IntegerNullable": ("Integer", "Double"),
+        "AgeNullable": ("Age", "AgeFractional"),
+    }
+
+    no_nans_ltype, has_nans_ltype = downcast_matches[str(col.ww.logical_type)]
```
jeremyliweishih, line 106: very nice!

evalml/tests/utils_tests/test_nullable_type_utils.py
```diff
@@ -0,0 +1,223 @@
...
+
+    (
+        y_compatible_ltypes,
+        y_incompatible_ltypes,
+    ) = split_nullable_logical_types_by_compatibility(
+        int_null_incompatible,
+        bool_null_incompatible,
+    )
+
+    if isinstance(nullable_ltype, tuple(y_compatible_ltypes)):
```
jeremyliweishih, line 214: can't seem to figure it out but lets make sure we're hitting this case before merging!

evalml/tests/component_tests/test_components.py
```diff
@@ -1804,3 +1807,80 @@ def test_component_parameters_supported_by_list_API(component_class):
...
+        y_int_null_incompatible,
+        y_bool_null_incompatible,
+    )
+
+    assert len(X_d.ww.select(X_incompatible_ltypes).columns) == 0
+    assert len(X_d.ww.select(X_compatible_ltypes).columns) == len(
+        X.ww.select(X_compatible_ltypes).columns,
+    )
+
+    if isinstance(nullable_y_ltype, tuple(y_compatible_ltypes)):
```
jeremyliweishih, line 1877: same thing here.

jeremyliweishih (approved):
LGTM other than the test misses. Looks like `nullable_ltype` is never an instance of `y_compatible_ltypes` across all the tests. Should be good to merge once its fixed! Great work!

evalml/tests/utils_tests/test_nullable_type_utils.py
tamargrey: So turns out `nullable_ltype` is a string so it'll never be an instance of the compatible ltypes 🤦‍♀️ 

Fixed that there and in the component and objective tests as well as a couple of other small bugs in these checks. 878cfc24c

chukarsten (approved):

Issue #3999, opened 2023-02-15
Remove Nullable type logic from Imputer Components and Refactor


tamargrey:
As part of handling nullable type incompatibilities, we're moving any needed logic to the components that need it. Currently, the Imputer components manage replacing nullable types for columns that contain null values (since they'll no longer have null values after being imputed), and that has contributed to an extra level of complexity that we should be able to remove once we've implemented component specific handling of nullable types. 

Current logic in imputer components that handle nullable types:
- in imputer.py: Call to downcast_nullable_types in transform. This is a catch all for any remaining nullable types, though may of the child imputers actually attempt to handle nullable types themselves. https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/imputers/imputer.py#L193
- in knn_imputer.py: Special logic to grab nullable types and update them to Boolean and Double after imputing: https://github.com/alteryx/evalml/blob/b220a01e668bab73fa00618ac1b920fe69b5237d/evalml/pipelines/components/transformers/imputers/knn_imputer.py#L94-L99
- In time_series_imputer.py - 
  - We cast numeric data to floats - this is actually a nullable type incompatibility! So we'll have to handle this via the new design
    - https://github.com/alteryx/evalml/blob/b220a01e668bab73fa00618ac1b920fe69b5237d/evalml/pipelines/components/transformers/imputers/time_series_imputer.py#L175-L177
    - https://github.com/alteryx/evalml/blob/b220a01e668bab73fa00618ac1b920fe69b5237d/evalml/pipelines/components/transformers/imputers/time_series_imputer.py#L190-L192
  - We exclude nullable types from the schema we use to reinit woodwork, meaning we're relying on woodwork type inference to remove the nullable types after imputing: https://github.com/alteryx/evalml/blob/b220a01e668bab73fa00618ac1b920fe69b5237d/evalml/pipelines/components/transformers/imputers/time_series_imputer.py#L154-L158
- in target_imputer.py - Turns IntegerNullabe to Integer, which would probably truncate an imputed mean or median 
  - https://github.com/alteryx/evalml/blob/b220a01e668bab73fa00618ac1b920fe69b5237d/evalml/pipelines/components/transformers/imputers/target_imputer.py#L135-L139

There is also other complexity in the imputer components that we should simplify as part of this issue. These are things that were first introduced as a result of previous handling of nullable types, never got removed once we changed how we handled nullable types, and then were built upon as they proved incompatible with our dependencies:
[...]

PR #4031, opened 2023-02-24
Add nullable type incompatibility properties to the components that have them

### LINKED ISSUE #3991: Add component specific nullable type handling logic to necessary components
- Once the base classes have the necessary logic for handling nullable types, we can add the related logic to the individual components that need it. Note - we should not be adding calls to `_handle_nullable_types` to the fit, transform, and predict calls, because we do not yet want to include this handling in AutoMLSearch.

```python
# A component that can't handle any nullable types and uses standard handling  
    _integer_nullable_incompatibilities = ["X", y"]
    _boolean_nullable_incompatibilities = ["X", y"]


# A component with more specific needs 
class MyEstimator(Estimator):
    _integer_nullable_incompatibilities = None
    _boolean_nullable_incompatibilities = ["X"]
    
    def _handle_nullable_types(self, X):
        # Converts boolean nullable to integer nullable 
        ...
```

This issue includes the following:
- Adding needed properties to components
- Adding tests that track the dependencies that need to add support that will fail when support is added
- Adding tests for all components to confirm that they support nullable types - should trigger a failure if a new component is added that needs nulalble type handling in place  
- Testing that when using the properties for the components in conjunction with `_handle_nullable_types` transforms the data correctly and adds support - note this will change to not have the explicit `_handle_nullable_types` call once we integrate into automl search. 
- Testing that there is not yet any change to automl search performance 

Can be completed in multiple independent PRs if needed

tamargrey:
closes #3991

Adds accurate `_integer_nullable_incompatibilities` and `_boolean_nullable_incompatibilities` properties to the components that are known to have nullable type incompatibilities. Tests that those incompatibilities cause errors that are fixed by the handling. Also tests that components without incompatibilities can use the nullable types.

Note - does not actually start calling `_handle_nullable_types` inside fit/predict/transform yet, as that would start using it in automl search. So we do not yet expect to see any impact to search performance or runtimes.


evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -152,8 +155,10 @@ def transform(self, X, y=None):
         X_not_all_null = X.ww.drop(self._all_null_cols)
         X_schema = X_not_all_null.ww.schema
         X_schema = X_schema.get_subset_schema(
+            # --> note that I'm including it right now so I can test fully even though this will go once we integrate into automlsearch
+            # https://github.com/alteryx/evalml/issues/4001#issuecomment-1443894376
             subset_cols=X_schema._filter_cols(
-                exclude=["IntegerNullable", "BooleanNullable"],
+                exclude=["IntegerNullable", "BooleanNullable", "AgeNullable"],
```
tamargrey, line 161: We'll likely remove this logic once we integrate the new nullable type handling into AutoMLSearch, but I wanted to add in `AgeNullable` here so that we could test with it now.

The one reason we may keep is is that nans are present at this stage, so we'll have to convert types to `Double` and `Categorical`, and we may want logic to not keep those types once nans are gone and we can use `Integer` and `Boolean`. I'll put more thought into this when I get to https://github.com/alteryx/evalml/issues/3999

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -472,3 +474,58 @@ def test_arima_regressor_can_forecast_arbitrary_dates(use_covariates, ts_data):
...
+        # TODO: Remove nullable type handling https://github.com/alteryx/evalml/issues/4016
+        sk_arima.fit(y=y_train, X=X_train)
+
+    # But EvalML's ARIMARegressor has a call to downcast_int_nullable_to_double to handle IntegerNullable
+    # so we don't see any errors currently
+    evalml_arima = ARIMARegressor(**arima_params)
+    evalml_arima.fit(X_train, y_train)
+
+    # Confirm that the handle nullable types method fixes the error for AutoARIMA
+    X_train_d, y_train_d = evalml_arima._handle_nullable_types(X_train, y_train)
```
tamargrey, line 528: Once we integrate the new handling into automl search, this call to `_handle_nullable_types` will be inside fit and predict, so we'll remove this. Keeping the error checking for sk_arima will help us know when sktime adds support for integer nullable on arima.

evalml/tests/component_tests/test_components.py
```diff
@@ -1884,3 +1885,86 @@ def test_handle_nullable_types(
...
+            or comp._boolean_nullable_incompatibilities
+        )
+    ],
+)
+@pytest.mark.parametrize(
+    "nullable_y_ltype",
+    ["BooleanNullable", "IntegerNullable", "AgeNullable"],
+)
+# --> this is 180 tests - is it overkill?
+def test_components_support_nullable_types(
```
tamargrey, line 1906: I feel like we want the tests for making sure our compatible components are actually compatible should be exhaustive. This does add ~180 tests, though, so I just want to make sure there's visibility around that.

evalml/tests/component_tests/test_estimators.py
```diff
@@ -441,3 +444,49 @@ def test_estimator_prediction_intervals(
...
+        X = nullable_type_test_data(has_nans=False)
+
+        # Exclude logical types that most estimators cannot handle
+        X = X.ww.select(["numeric", "Boolean", "BooleanNullable"])
+
+        comp = incompatible_estimator()
+
+        with pytest.raises(Exception):
+            comp.fit(X, y)
+            comp.predict(X)
```
tamargrey, line 486: Because we are using the actaul evalml components here, this will have to change once we integrate into automl search to call the underlying `component_obj` so that we can track when support gets added for that component.

evalml/tests/component_tests/test_oversampler.py
```diff
@@ -434,3 +435,40 @@ def test_oversampler_copy(X_y_binary):
...
+        y = nullable_type_target(ltype=nullable_ltype, has_nans=False)
+        oversampler.fit(X, y)
+
+        with pytest.raises(
+            ValueError,
+            match=re.escape(
+                "Unknown label type: 'unknown'",
+            ),
+        ):
+            oversampler.transform(X, y)
```
tamargrey, line 466: similar to the `test_estimators.py` test - we are using the actual evalml component here, so this will have to change once we integrate into automl search

evalml/tests/conftest.py
```diff
@@ -2332,12 +2332,10 @@ def _build_nullable_type_data(has_nans=True):
 @pytest.fixture
 def nullable_type_target():
     def _build_nullable_type_data(ltype="BooleanNullable", has_nans=True):
-        y = pd.Series([1, 0, 1, 1, 0] * 4)
-
         if has_nans:
             y = pd.Series([1, 0, pd.NA, 1, 0] * 4)
         else:
-            y = pd.Series([1, 0, 1, 1, 0] * 4)
+            y = pd.Series([1, 0, 1, 1, 1] * 4)
```
tamargrey, line 2338: Changed so I could test the oversampler better :)

evalml/tests/component_tests/test_components.py
```diff
@@ -1884,3 +1885,85 @@ def test_handle_nullable_types(
...
+        )
+
+        if nullable_y_ltype == "BooleanNullable":
+            y = bool_col
+
+        # Add nullable types
+        X.ww.set_types(logical_types={"feature": "IntegerNullable"})
+        X.ww["bool col"] = bool_col
+    else:
+        y = nullable_type_target(ltype=nullable_y_ltype, has_nans=False)
```
jeremyliweishih, line 1957: why don't we run `y = nullable_type_target(ltype=nullable_y_ltype, has_nans=False)` for components that require the time index?

jeremyliweishih (approved):
LGTM - just one clarifying question

evalml/tests/component_tests/test_components.py
tamargrey: I probably could; I'd just need to change `nullable_type_target` to be the correct length for `ts_data` and also update it to have the same datetime index values as `X`. I felt like it was simpler to just create `bool_col` and set y to be it.

evalml/tests/component_tests/test_estimators.py
```diff
@@ -441,3 +444,49 @@ def test_estimator_prediction_intervals(
...
+        "y" in incompatible_estimator._boolean_nullable_incompatibilities,
+    )
+
+    for y_ltype in incompatible_y_ltypes:
+        # Most estimators cannot handle nans
+        y = nullable_type_target(ltype=y_ltype, has_nans=False)
+        X = nullable_type_test_data(has_nans=False)
+
+        # Exclude logical types that most estimators cannot handle
+        X = X.ww.select(["numeric", "Boolean", "BooleanNullable"])
```
eccabay, line 480: Aren't we including, not excluding, those types?

evalml/tests/component_tests/test_lgbm_classifier.py
```diff
@@ -328,3 +329,46 @@ def test_lgbm_preserves_schema_in_rename(mock_predict_proba, mock_predict, mock_
...
+                lgb.fit(X, y)
+
+            # Confirm using the handle method lets the transform work
+            X_d, y_d = lgb._handle_nullable_types(X, y)
+            lgb.fit(X_d, y_d)
+            lgb.predict(X_d)
+        else:
+            lgb.fit(X, y)
+            lgb.predict(X)
+            lgb.predict_proba(X)
```
eccabay, line 371: Mega nit: can you simplify this a bit by setting `X, y = lgb._handle_nullable_types(X, y)` and calling fit, predict, and predict_proba outside of the conditional block?

eccabay (approved):
Looks pretty good! Mostly just test cleanups and clarifying questions, but nothing blocking.

evalml/tests/component_tests/test_components.py
```diff
@@ -1884,3 +1885,85 @@ def test_handle_nullable_types(
...
+
+    elif (
+        component_class in cannot_handle_boolean_target
+        and nullable_y_ltype == "BooleanNullable"
+    ):
+        pytest.skip(f"Cannot pass boolean y into {component_class}.")
+
+    # Some components require special handling to fit and predict/transform
+    requires_time_index = [
+        ARIMARegressor,
```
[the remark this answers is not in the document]
tamargrey: whoopse, good call

evalml/tests/component_tests/test_estimators.py
tamargrey: I was referring here to types like natural language or categorical that we aren't usually passing directly to estimators. But I see how the wording is confusing. I'll make it more clear

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -558,3 +560,101 @@ def test_imputer_woodwork_custom_overrides_returned_by_components(
...
+def test_imputer_nullable_handling_numeric_interpolate(nullable_type_test_data):
+    X = nullable_type_test_data(has_nans=True)
+    # Only numeric imputing has interpolate as an option
+    X = X.ww.select("numeric")
+
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "Invalid fill method. Expecting pad (ffill) or backfill (bfill). Got linear",
+        ),
```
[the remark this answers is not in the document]
tamargrey: I'll double check, but I think the parentheses were what was giving me trouble here

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -558,3 +560,101 @@ def test_imputer_woodwork_custom_overrides_returned_by_components(
         transformed.ww.logical_types["categorical with nan"]
         == X.ww.logical_types["categorical with nan"]
     )
+
+
+def test_imputer_nullable_handling_numeric_interpolate(nullable_type_test_data):
```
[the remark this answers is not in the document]
tamargrey: This is a fair point, and I recognize that we do not generally want to be testing our dependencies' behaviors. Part of the goal of this, beyond confirming that our handling lets us avoid the problem, is that this check will notify us when support is added. I have a similar check in the arima tests, and am planning to eventually have them for all incompatibilities.

This is just my preference for how to keep track of our dependencies adding support, though. I recognize it means that when a pandas version is released with support interpolating nullable types, for example, we'll get an error that we will need to handle before we can merge any dependency updates. 

That will be a bit of a pain, but it allows us to remove this handling when possible without having to manually keep track of when dependencies add support, which I would argue is more of a pain. Happy to open this up to a larger discussion with the rest of the team.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -194,3 +199,21 @@ def transform(self, X, y=None):
             y_imputed = ww.init_series(y_imputed)
 
         return X_not_all_null, y_imputed
+
+    def _handle_nullable_types(self, X=None, y=None):
+        """Transforms X and y to remove any incompatible nullable types for the time series imputer when the interpolate method is used.
```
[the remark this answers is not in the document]
tamargrey: interpolation is the only one that has any incompatibilities with nullable types, so there's no need to remove nullable types for the other fill methods.

evalml/tests/component_tests/test_time_series_imputer.py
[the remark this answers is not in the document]
tamargrey: yep


> AssertionError: Regex pattern 'Invalid fill method. Expecting pad (ffill) or backfill (bfill). Got linear' does not match 'Invalid fill method. Expecting pad (ffill) or backfill (bfill). Got linear'. Did you mean to `re.escape()` the regex?

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -658,3 +658,20 @@ def test_imputer_nullable_handling_noop_for_non_impute_methods(
...
     else:
         assert y_d.ww.schema != original_y_schema
+
+
+@pytest.mark.parametrize(
+    "nullable_ltype",
+    ["BooleanNullable", "IntegerNullable", "AgeNullable"],
+)
+@pytest.mark.xfail(strict=True, raises=ValueError)
+def test_time_series_imputer_nullable_type_incompatibility(
```
tamargrey, line 668: @chukarsten @eccabay @jeremyliweishih I'm still adding the rest of the xfail tests but wanted to run this by yall. 

- I have the xfail set to only expect `ValueError`, but unfortunately, we can't specify the expected message. Would people prefer I get as specific about what the expected message is? 
- The `strict` parameter means that our test suite will fail when tests start passing - otherwise they just show as `XPASS` which I think would be easy to miss. But if we want to avoid blocking the update checker, we could set this to False and just periodically remember to check. I think that's probably a halfway point between what I'm suggesting and the fully manual checking process.

eccabay: This method of xfail seems like the best to me!

I'm a fan of keeping `strict=True`. It's really easy to miss an `XPASS`, and we don't have any way of flagging them at this point. I'm pretty sure we already have a bunch of `XPASS`ing tests that we might want to revisit...

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -194,3 +200,27 @@ def transform(self, X, y=None):
...
+
+        Returns:
+            X, y with any incompatible nullable types downcasted to compatible equivalents when interpolate is used. Is NoOp otherwise.
+        """
+        if self._impute_target == "interpolate":
+            # For BooleanNullable, we have to avoid Categorical columns
+            # since the category dtype also has incompatibilities with linear interpolate, which is expected
+            # --> i think this is essentially what happens now but won't that make floating point values that cant be turned back into bools?
+            if isinstance(y.ww.logical_type, BooleanNullable):
+                y = ww.init_series(y, Double)
```
tamargrey, line 220: Noticed that `interpolate()` cannot handle `category` dtypes either. But this, I think, is by design, as the default interpolation is "linear" which doesnt make sense for non numeric values. This is problematic if we try to turn BooleanNullable into Categorical ltype when nans are present, so I'm intercepting it and converting to `Double`, which essentially the current behavior.

In the long run, though, I think we don't want to be doing linear interpolation on boolean columns ever - we don't want to end up with `1.5` in a column of all 1s and 0s. I'll open a separate evalml ticket for handling that, though.

jeremyliweishih: makes sense!

jeremyliweishih (approved):
LGTM - great work!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
eccabay: We already don't allow using interpolation with categorical or boolean types. From the docstring:
```
        categorical_impute_strategy (string): Impute strategy to use for string, object, boolean, categorical dtypes.
            Valid values include "backwards_fill" and "forwards_fill". Defaults to "forwards_fill".
        numeric_impute_strategy (string): Impute strategy to use for numeric columns. Valid values include
            "backwards_fill", "forwards_fill", and "interpolate". Defaults to "interpolate".
```
It's enforced in a couple different places, between checking the arguments and how we select splitting between numeric and categorical!

eccabay: Ah, very true - good point!

PR #2476, opened 2021-07-08
Automated Latest Dependency Updates

machineFL:
This is an auto-generated PR with **latest** dependency updates. Please do not delete the `latest-dep-update` branch because it's needed by the auto-dependency bot.

PR #4035, opened 2023-02-27
Update networkx version

### LINKED ISSUE #3972: Support networkx >=2.6
Currently only networkx 2.5 is supported with this package, but this is a fairly old version of networkx.  I would like to request support of newer versions of this library.

eccabay:
Closes #3972 

Networkx changed the topological order of their DAGs with version 2.6, which broke some of our tests that validate the computation order. We [pinned it](https://github.com/alteryx/evalml/pull/2476) and left it there. Changes here unpin the version and update the tests to be more flexible about the order of computations, so long as the data flow is still maintained.

### DIFF
--- a/evalml/tests/pipeline_tests/test_component_graph.py
+++ b/evalml/tests/pipeline_tests/test_component_graph.py
@@ -111,15 +111,11 @@ def test_init(example_graph):
     comp_graph = ComponentGraph(graph)
     assert len(comp_graph.component_dict) == 6
 
-    expected_order = [
-        "Imputer",
-        "OneHot_ElasticNet",
-        "Elastic Net",
-        "OneHot_RandomForest",
-        "Random Forest",
-        "Logistic Regression Classifier",
-    ]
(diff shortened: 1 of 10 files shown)


evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -271,15 +263,11 @@ def test_instantiate_with_parameters(example_graph):
     }
     component_graph.instantiate(parameters)
 
-    expected_order = [
```
jeremyliweishih, line 274: is the component ordering changing? Just wondering why we're changing these tests

eccabay: Yep, that was the change that caused us to pin the version a year and a half ago. For non-linear graphs, the order is now more breadth-first rather than depth-first. It's still a correct ordering, which is why the tests now simply assert that components that must be fit/transformed before others still are.

jeremyliweishih (approved):
Thanks!

PR #4037, opened 2023-02-27
add logic to speed up permutation feature importance

### LINKED ISSUE #4036: _fast_permutation_importance doesn't run if DFS Transformer present, even when dfs feature pre-exist.
If a DFS Transformer is present, we don't run `_fast_permutation_importance` instead running `_slow_permutation_importance`. 
```
    @property
    def _supports_fast_permutation_importance(self):
        has_more_than_one_estimator = (
            sum(isinstance(c, Estimator) for c in self.component_graph) > 1
        )
        _all_components = set(all_components())
        has_custom_components = any(
            c.__class__ not in _all_components for c in self.component_graph
        )
        has_dim_reduction = any(
            isinstance(c, (PCA, LinearDiscriminantAnalysis))
            for c in self.component_graph
        )
        has_dfs = any(isinstance(c, DFSTransformer) for c in self.component_graph)
        return not any(
            [
                has_more_than_one_estimator,
                has_custom_components,
                has_dim_reduction,
                has_dfs,
            ],
        )
```
This is fine for the majority of cases however in the case where we already have all features from DFS present in the pipeline before applying DFS Transformer we really should be running `_fast_permutation_importance`.

Something like this should work
```
    @property
    def _supports_fast_permutation_importance(self):
        has_more_than_one_estimator = (
            sum(isinstance(c, Estimator) for c in self.component_graph) > 1
        )
        _all_components = set(all_components())
        has_custom_components = any(
            c.__class__ not in _all_components for c in self.component_graph
        )
        has_dim_reduction = any(
            isinstance(c, (PCA, LinearDiscriminantAnalysis))
            for c in self.component_graph
        )
        has_dfs = any(isinstance(c, DFSTransformer) for c in self.component_graph)
        if has_dfs:
            dfs_feature_names = [
                name
                for feature in self.component_graph[DFSTransformer.name].features
                for name in feature.get_feature_names()
                if name != self.input_target_name
            ]
            contains_pre_existing_dfs_features = (
                dfs_feature_names == self.input_feature_names[DFSTransformer.name]
            )
        return not any(
            [
                has_more_than_one_estimator,
                has_custom_components,
                has_dim_reduction,
                (has_dfs and not contains_pre_existing_dfs_features),
            ],
        )
```

ozzieD:
- Checks input features with features in the DFS transform component. If features are the same then it will bypass the `has_dfs` check in `_supports_fast_permutation_importance`
- Fixes #4036


evalml/pipelines/pipeline_base.py
```diff
@@ -757,12 +757,22 @@ def _supports_fast_permutation_importance(self):
             for c in self.component_graph
         )
         has_dfs = any(isinstance(c, DFSTransformer) for c in self.component_graph)
+        if has_dfs:
+            dfs_feature_names = [
+                name
+                for feature in self.component_graph[DFSTransformer.name].features
```
tamargrey, line 763: I 100% get why you didn't pull this out into a shared util with the `_handle_partial_dependence_fast_mode` method in featuretools.py (it's not exactly a lot of code), but I want to make sure that if we ever have to change one of these places for some reason that the other gets the same change made (I could see us forgetting they both exist!). Can we either have a comment linking the two places or pull out the shared logic into a util? It could even be a method on the DFS Transformer that just returns the boolean `contains_pre_existing_dfs_features`

evalml/tests/model_understanding_tests/test_permutation_importance.py
```diff
@@ -398,10 +398,25 @@ def test_fast_permutation_importance_matches_slow_output(
...
+    def __init__(self, name) -> None:
+        self.name = name
+
+    def get_feature_names(self):
+        return self.name
+
+
 def pipelines_that_do_not_support_fast_permutation_importance():
-    pipeline_with_dfs = BinaryClassificationPipeline(
+    pipeline_with_dfs_and_no_prexisting_features = BinaryClassificationPipeline(
```
jeremyliweishih, line 410: Can you add the opposite case to `test_fast_permutation_importance_matches_slow_output` so that we have coverage for DFS with prexisting features?

jeremyliweishih (changes requested):
Just requesting one additional coverage but looks great otherwise!

evalml/pipelines/pipeline_base.py
ozzieD: Done

evalml/pipelines/components/transformers/preprocessing/featuretools.py
```diff
@@ -170,3 +166,21 @@ def _handle_partial_dependence_fast_mode(self, pipeline_parameters, X, target):
...
+            dfs_feature_names = [
+                name
+                for feature in dfs_features
+                for name in feature.get_feature_names()
+                if name != target
+            ]
+            return dfs_feature_names == input_feature_names
+
+        else:
+            return False
```
tamargrey, line 186: Seems this line isn't being hit by tests - might be worth a simple test in `test_featuretools` to get this covered

ozzieD: Done

evalml/tests/model_understanding_tests/test_permutation_importance.py
ozzieD: done

jeremyliweishih (approved):
LGTM once tests pass! great work 🦾

tamargrey (approved):
lgtm! Nice work!

PR #3525, opened 2022-05-17
Sklearn 11 upgrade

chukarsten:
### Pull Request Description
(replace this text with your description)

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #3388, opened 2022-03-18
convert url/email features to object before categorical

freddyaboulton:
### Pull Request Description
Fixes #3366 

-----
*After creating the pull request: in order to pass the **release_notes_updated** check you will need to update the "Future Release" section of* `docs/source/release_notes.rst` *to include this pull request by adding :pr:`123`.*

PR #4038, opened 2023-02-28
Refactor imputer components to remove unnecessary logic 

tamargrey:
Related to https://github.com/alteryx/evalml/issues/3999 (Note - will not yet close this issue, as this PR does not remove the nullable type logic; it only refactors the unnecessary logic)

Refactors the following imputers
- imputer.py 
- simple_imputer.py
- time_series_imputer.py 
- target_imputer.py
- knn_imputer.py

### DIFF
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -127,17 +127,21 @@ class Imputer(Transformer):
         nan_ratio = X.isna().sum() / X.shape[0]
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
 
-        X_numerics = X[[col for col in numeric_cols if col not in self._all_null_cols]]
+        X_numerics = X.ww[
+            [col for col in numeric_cols if col not in self._all_null_cols]
+        ]
         if len(X_numerics.columns) > 0:
             self._numeric_imputer.fit(X_numerics, y)
             self._numeric_cols = X_numerics.columns
 
-        X_categorical = X[[col for col in cat_cols if col not in self._all_null_cols]]
+        X_categorical = X.ww[
+            [col for col in cat_cols if col not in self._all_null_cols]
+        ]
         if len(X_categorical.columns) > 0:
             self._categorical_imputer.fit(X_categorical, y)
             self._categorical_cols = X_categorical.columns
 
-        X_boolean = X[[col for col in bool_cols if col not in self._all_null_cols]]
+        X_boolean = X.ww[[col for col in bool_cols if col not in self._all_null_cols]]
         if len(X_boolean.columns) > 0:
             self._boolean_imputer.fit(X_boolean, y)
             self._boolean_cols = X_boolean.columns
--- a/evalml/pipelines/components/transformers/imputers/imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/imputer.py
@@ -5,7 +5,6 @@ from woodwork import init_series
 from evalml.pipelines.components.transformers import Transformer
 from evalml.pipelines.components.transformers.imputers import KNNImputer, SimpleImputer
 from evalml.utils import downcast_nullable_types, infer_feature_types
-from evalml.utils.gen_utils import is_categorical_actually_boolean
 
 
 class Imputer(Transformer):
@@ -124,15 +123,6 @@ class Imputer(Transformer):
             X.ww.select(["BooleanNullable", "Boolean"], return_schema=True).columns,
         )
         numeric_cols = list(X.ww.select(["numeric"], return_schema=True).columns)
-
-        # TODO: Remove this when columns with True/False/NaN are inferred properly as BooleanNullable.
-        # If columns with boolean values and NaN are included with normal categorical columns, columns
-        # with object dtypes are attempted to be cast to float64 with scikit-learn 1.1.  So we separate
-        # boolean and categorical into separate imputers.
-        for col in cat_cols:
-            if is_categorical_actually_boolean(X, col):
-                cat_cols.remove(col)
-                bool_cols.append(col)
 
         nan_ratio = X.isna().sum() / X.shape[0]
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
--- a/evalml/pipelines/components/transformers/imputers/knn_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/knn_imputer.py
@@ -89,7 +89,6 @@ class KNNImputer(Transformer):
         original_index = X.index
         original_schema = X.ww.schema
 
-        # Drop natural language columns and transform the other columns
         if not self._cols_to_impute:
             return X.ww[not_all_null_cols]
 
--- a/evalml/pipelines/components/transformers/imputers/knn_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/knn_imputer.py
@@ -5,7 +5,6 @@ import woodwork
 from sklearn.impute import KNNImputer as Sk_KNNImputer
 
 from evalml.pipelines.components.transformers import Transformer
-from evalml.pipelines.components.utils import drop_natural_language_columns
 from evalml.utils import infer_feature_types
 
 
@@ -51,16 +50,27 @@ class KNNImputer(Transformer):
 
         """
         X = infer_feature_types(X)
-        X, _ = drop_natural_language_columns(X)
 
         nan_ratio = X.isna().sum() / X.shape[0]
+
+        # Keep track of the different types of data in X
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
+        self._natural_language_cols = list(
+            X.ww.select("NaturalLanguage", return_schema=True).columns.keys(),
+        )
+
+        # Only impute data that is not natural language columns or fully null
+        self._cols_to_impute = [
+            col
+            for col in X.columns
+            if col not in self._natural_language_cols and col not in self._all_null_cols
+        ]
 
         # If the Dataframe only had natural language columns, do nothing.
-        if X.shape[1] == 0:
+        if not self._cols_to_impute:
             return self
 
-        self._component_obj.fit(X, y)
+        self._component_obj.fit(X[self._cols_to_impute], y)
         return self
 
     def transform(self, X, y=None):
@@ -77,37 +87,35 @@ class KNNImputer(Transformer):
 
         not_all_null_cols = [col for col in X.columns if col not in self._all_null_cols]
         original_index = X.index
+        original_schema = X.ww.schema
 
         # Drop natural language columns and transform the other columns
-        X_t, natural_language_cols = drop_natural_language_columns(X)
-        if X_t.shape[1] == 0:
-            return X
-        not_all_null_or_natural_language_cols = [
-            col for col in not_all_null_cols if col not in natural_language_cols
-        ]
+        if not self._cols_to_impute:
+            return X.ww[not_all_null_cols]
 
-        X_t = self._component_obj.transform(X_t)
-        X_t = pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)
+        X_t = self._component_obj.transform(X[self._cols_to_impute])
+        X_t = pd.DataFrame(X_t, columns=self._cols_to_impute)
 
-        X_schema = X.ww.schema
+        # Get Woodwork types for the imputed data
+        new_schema = original_schema.get_subset_schema(self._cols_to_impute)
 
-        X_bool_nullable_cols = X_schema._filter_cols(include=["BooleanNullable"])
-        X_int_nullable_cols = X_schema._filter_cols(include=["IntegerNullable"])
+        X_bool_nullable_cols = new_schema._filter_cols(include=["BooleanNullable"])
+        X_int_nullable_cols = new_schema._filter_cols(include=["IntegerNullable"])
         new_ltypes_for_nullable_cols = {col: "Boolean" for col in X_bool_nullable_cols}
         new_ltypes_for_nullable_cols.update(
             {col: "Double" for col in X_int_nullable_cols},
         )
 
         # Add back in natural language columns, unchanged
-        if len(natural_language_cols) > 0:
-            X_t = woodwork.concat_columns([X_t, X[natural_language_cols]])
+        if len(self._natural_language_cols) > 0:
+            X_t = woodwork.concat_columns([X_t, X.ww[self._natural_language_cols]])
 
         X_t.ww.init(
-            schema=X_schema,
+            schema=new_schema,
             logical_types=new_ltypes_for_nullable_cols,
         )
 
-        if not_all_null_or_natural_language_cols:
+        if self._cols_to_impute:
             X_t.index = original_index
 
         return X_t
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -90,13 +90,14 @@ class SimpleImputer(Transformer):
         # If there are no columns to impute, return early
         if not self._cols_to_impute:
             return self
-        elif (X.dtypes == bool).all():
-            self._fit_with_all_bool_dtypes = True
-            return self
-        else:
-            self._fit_with_all_bool_dtypes = False
 
-        self._component_obj.fit(X[self._cols_to_impute], y)
+        X = X[self._cols_to_impute]
+        if (X.dtypes == bool).all():
+            # Ensure that _component_obj still gets fit so that if any of the dtypes are different
+            # at transform, we've fit the component
+            X = X.astype("boolean")
+
+        self._component_obj.fit(X, y)
         return self
 
     def transform(self, X, y=None):
@@ -113,17 +114,14 @@ class SimpleImputer(Transformer):
         original_schema = X.ww.schema
         original_index = X.index
 
+        X_t = X[self._cols_to_impute]
         not_all_null_cols = [col for col in X.columns if col not in self._all_null_cols]
-        if not self._cols_to_impute:
-            # If there are no columns to impute, return the original data without any fully null columns
+        if not self._cols_to_impute or (X_t.dtypes == bool).all():
+            # If there are no columns to impute or all columns to impute are bool dtype, which sklearn errors on,
+            # return the original data without any fully null columns
             return X.ww[not_all_null_cols]
-        elif (X.dtypes == bool).all():
-            return X
-        # If the dtypes are not all bool but it was with with all bool dtypes we need to fit the _component_obj
-        elif self._fit_with_all_bool_dtypes:
-            self._component_obj.fit(X[self._cols_to_impute])
 
-        X_t = self._component_obj.transform(X[self._cols_to_impute])
+        X_t = self._component_obj.transform(X_t)
         X_t = pd.DataFrame(X_t, columns=self._cols_to_impute)
 
         # Get Woodwork types for the imputed data
--- a/evalml/pipelines/components/transformers/imputers/simple_imputer.py
+++ b/evalml/pipelines/components/transformers/imputers/simple_imputer.py
@@ -5,12 +5,7 @@ from sklearn.impute import SimpleImputer as SkImputer
 from woodwork.logical_types import Double
 
 from evalml.pipelines.components.transformers import Transformer
-from evalml.pipelines.components.utils import (
-    drop_natural_language_columns,
-    set_boolean_columns_to_integer,
-)
 from evalml.utils import infer_feature_types
-from evalml.utils.gen_utils import is_categorical_actually_boolean
 
 
 class SimpleImputer(Transformer):
@@ -69,12 +64,7 @@ class SimpleImputer(Transformer):
         if set([lt.type_string for lt in X.ww.logical_types.values()]) == {
             "boolean",
             "categorical",
-        } and not all(
-            [
-                is_categorical_actually_boolean(X, col)
-                for col in X.ww.select("Categorical")
-            ],
-        ):
+        }:
             raise ValueError(
                 "SimpleImputer cannot handle dataframes with both boolean and categorical features.  Use Imputer instead.",
             )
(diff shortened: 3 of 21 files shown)


machineFL:

evalml/pipelines/components/transformers/preprocessing/transform_primitive_components.py
```diff
@@ -73,17 +73,6 @@ def transform(self, X, y=None):
...
-        # Convert to object dtype so that pd.NA is converted to np.nan
-        # until sklearn imputer can handle pd.NA in release 1.1
-        # FT returns these as string types, currently there isn't much difference
-        # in terms of performance between object and string
-        # see https://pandas.pydata.org/docs/user_guide/text.html#text-data-types
-        # "Currently, the performance of object dtype arrays of strings
-        # "and arrays.StringArray are about the same."
-        features = features.astype(object, copy=False)
-        features.index = X_ww.index
-        features.ww.init(logical_types={col_: "categorical" for col_ in features})
```
tamargrey, line 85: This popped up as a problem in this CI run - by removing `is_categorical_actually_boolean` usage in the simple imputer, we were no longer recognizing features made from `IsFreeEmailDomain` as boolean columns and imputing them as such in `test_imputer_can_impute_features_generated_from_null_email_url_features`.

I think we're good to remove this block of code for two reasons
1. Freddy added much of that in [this PR](https://github.com/alteryx/evalml/pull/3388/files). We should be good to remove it now that we’re upgraded past sklearn 1.1
2. The fact that we were converting all columns to `categorical` logical type at all was originally done because of a [change in woodwork inference](https://github.com/alteryx/evalml/pull/2690/files#r699592236), but we were also not using woodwork in featuretools at that time (which you can see from the usage of entity in the featuretools calls from that PR), so we no longer have to rely on woodwork inference to get the types of features, which will have woodwork types specified by the primitives used to make them.

tamargrey: While sklearn 1.1 did fix the problem Freddy described, I found that removing this block of code entirely caused an issue in Catboost that I was seeing during #3966 with string categories. Converting to `object` fixed that problem, so I'm leaving that as is and will open a ticket to remove this once that bug is fixed.

evalml/pipelines/components/transformers/imputers/imputer.py
```diff
@@ -125,29 +124,24 @@ def fit(self, X, y=None):
...
-        X_categorical = X[[col for col in cat_cols if col not in self._all_null_cols]]
+        X_categorical = X.ww[
+            [col for col in cat_cols if col not in self._all_null_cols]
+        ]
         if len(X_categorical.columns) > 0:
             self._categorical_imputer.fit(X_categorical, y)
             self._categorical_cols = X_categorical.columns
 
-        X_boolean = X[[col for col in bool_cols if col not in self._all_null_cols]]
+        X_boolean = X.ww[[col for col in bool_cols if col not in self._all_null_cols]]
```
tamargrey, line 144: When using the imputer in pipelines, we weren't fitting the child imputers on woodwork-initialized data.

I'm surprised this hasn't been noticed before, but I stumbled upon it because of these changes. Now that we're back to returning early if all the columns are the `bool` dtype, this was causing an issue in a `BooleanNullable` column that would get re-inferred as `Boolean` at fit, so we'd return early and then raise an error that the `_component_obj` wasn't fit at transform when the correct `BooleanNullable` type was passed in.

evalml/tests/pipeline_tests/test_component_graph.py
```diff
@@ -2737,7 +2744,7 @@ def test_get_component_input_logical_types():
         "cat": Categorical(),
         "numeric": Integer(),
         "EMAIL_ADDRESS_TO_DOMAIN(email)": Categorical(),
-        "IS_FREE_EMAIL_DOMAIN(email)": Categorical(),
+        "IS_FREE_EMAIL_DOMAIN(email)": BooleanNullable(),
```
tamargrey, line 2747: The change to not automatically convert all features to `Categorical` will result in `BooleanNullable` columns being introduced into `X` via the `EmailFeaturizer` after the `ReplaceNullableTypes` transformer has been used. 

Because of that, this change will have to wait to be introduced with the rest of the nullable type handling. 😢

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -69,42 +64,33 @@ def fit(self, X, y=None):
...
-                include=["Boolean", "BooleanNullable"],
-                return_schema=True,
-            ).columns,
-        )
-        # Make sure we're tracking Categorical columns that should be boolean as well
-        self._boolean_cols.extend(
-            [
-                col
-                for col in X.ww.select("Categorical")
-                if is_categorical_actually_boolean(X, col)
```
tamargrey, line 99: Explanation as to why we can remove `is_categorical_actually_boolean`:
- Was originally added in https://github.com/alteryx/evalml/pull/3525 when we first started imputing boolean columns separately from categorical columns.
- At that time, the current nullable type handling involved converting BooleanNullable columns to Categorical, so this logic was needed to correctly partition columns for imputing. We no longer create these categorical booleans (I caught and removed the last case where we were doing it in the EmailFeaturizer), and I don't think we need to handle the case where a user has created them, so uses of this should be removed.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -69,42 +64,33 @@ def fit(self, X, y=None):
...
-            ],
-        )
-        X = set_boolean_columns_to_integer(X)
```
tamargrey, line 102: Explanation as to why we can stop converting boolean columns to integers:
- This originally began as converting booleans to categorical at `fit` inside the simple imputer in https://github.com/alteryx/evalml/pull/3324. Not 100% clear why this was needed, but I imagine it had to do with a pandas incompatibility with booleans.
- Then, we started converting all boolean columns to categorical at `transform` as well in https://github.com/alteryx/evalml/pull/3892 (which I think was done because we couldn't reinfer the columns as booleans when they were imputed as booleans because sklearn turns them into 1s and 0s)
- Eventually, turning bools to categories became converting to IntegerNullable in https://github.com/alteryx/evalml/pull/3959 to handle a pandas bug with imputing categorical booleans.
- Given that the sklearn simple imputer's behavior is to turn boolean columns to 1s and 0s when imputed, we've now essentially recreated that behavior outside of it, so we should be able to just pass the original boolean and boolean nullable columns into the sklearn imputer and handle turning them back to Booleans at the end like we are now.

tamargrey:
I believe the codecov failures are just because enough code was removed that coverage has lowered, but if you go into the codecov report, none of the changes show uncovered lines.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
tamargrey: I'm slightly worried that part of the original reason for converting boolean columns was because sklearn will not fit or transform its imputer on data that is only the `bool` dtype. Converting to category or int64 doesnt solves that problem because sklearn will happily take in data of only those types. 

At one point we had the logic that I now added back in to return early and not fit the imputer on sklearn's imputer object if only `bool` columns are present, and I havent been able to confirm it, but it's possible that part of the reason we removed that logic is that it's problematic if nans get introduced in the test data - we would have not fit the component, causing errors.

Because of that, I added logic to store the information if the component was fit on fully `bool` data, and if the data at transform isn't fully `bool`, call fit on the component of only the `cols_to_impute`.

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -124,19 +129,11 @@ def transform(self, X, y):
...
-            y_t = y_t.astype(bool)
-
-        new_logical_type = (
-            Integer
-            if isinstance(y_ww.ww.logical_type, IntegerNullable)
-            else y_ww.ww.logical_type
-        )
+        new_logical_type = y_ww.ww.logical_type
+        if isinstance(y_ww.ww.logical_type, IntegerNullable):
+            new_logical_type = Double
```
tamargrey, line 134: Converting imputed IntegerNullable data to Integer truncates the imputed values for mean and median, so I changed this to Double, which matches other imputers' behavior.

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -124,19 +129,11 @@ def transform(self, X, y):
...
+        elif isinstance(y_ww.ww.logical_type, BooleanNullable):
+            new_logical_type = Boolean
```
tamargrey, line 136: Added this for consistency's sake

evalml/pipelines/components/transformers/imputers/knn_imputer.py
```diff
@@ -51,16 +50,27 @@ def fit(self, X, y=None):
...
         nan_ratio = X.isna().sum() / X.shape[0]
+
+        # Keep track of the different types of data in X
         self._all_null_cols = nan_ratio[nan_ratio == 1].index.tolist()
+        self._natural_language_cols = list(
+            X.ww.select("NaturalLanguage", return_schema=True).columns.keys(),
+        )
+
+        # Only impute data that is not natural language columns or fully null
+        self._cols_to_impute = [
```
tamargrey, line 63: This change to use a `_cols_to_impute` variable to manage the fully null and natural language column logic here and at the simple imputer was a change that, to me, makes the imputer logic easier to understand. It makes the handling more explicit rather than a sequence of events to keep track of. It should also have slightly more efficient woodwork usage.

One implication of this is that we never even pass fully null columns to the imputer component_obj, which could be problematic for test data that may not be fully null, but in that case, we'd have no information from the training data to use to impute data, and in the previous call to `pd.DataFrame(X_t, columns=not_all_null_or_natural_language_cols)`, it assumes the fully null columns wouldn't be present anyway, so I'm assuming that it's okay to not pass the fully null columns in at all.

jeremyliweishih: It should be ok as long as we're not passing the fully null or natural language column as input for both fit and transform. It is also ok to pass the full null columns in - I believe the sklearn imputer's drop them automatically.

jeremyliweishih: Screenshot 2023-03-02 at 2 05 39 PM

evalml/pipelines/components/transformers/imputers/simple_imputer.py
jeremyliweishih: Looks like we have to handle this edge case if I understood correctly:
Screenshot 2023-03-02 at 2 13 26 PM

"Given that the sklearn simple imputer's behavior is to turn boolean columns to 1s and 0s when imputed, we've now essentially recreated that behavior outside of it, so we should be able to just pass the original boolean and boolean nullable columns into the sklearn imputer and handle turning them back to Booleans at the end like we are now."

This is only true when the impute strategy is mean. Above we set it to `most_frequent` and we error out. This is the case we need since we only use `most_frequent` with bool columns in `Imputer`.

Could we convert the column to a string categorical or keep the current behavior of converting to a integer just for the `most_frequent` strategy or your suggestion above? I'd still prefer the conversion over refitting and transforming on the component object.

jeremyliweishih (changes requested):
LGTM other than the all bool logic for the simple imputer. Lmk what you think!

evalml/pipelines/components/transformers/imputers/knn_imputer.py
tamargrey: yep! It simplified things a bit to not pass them in at either fit or transform, but agreed it's something we need to be consistent on.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
eccabay: Agreed with Jeremy here - calling fit during transform opens us up to future bugs. We only call fit once, and only ever call it on the training data. With a call to `_component_obj.fit` during transform, we now allow the component to fit on validation/testing data and overwrite earlier component training.
A quick test shows what would happen in the current setup:
Screen Shot 2023-03-03 at 9 23 37 AM

We will be better served converting the booleans to integers rather than categoricals for `most_frequent` to match the behavior for `mean`, and agreed that conversion is preferable to this refitting.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
```diff
@@ -149,11 +137,13 @@ def transform(self, X, y=None):
...
         X_t.ww.init(schema=new_schema)
 
-        # Add back in natural language columns, unchanged
-        if len(natural_language_cols) > 0:
-            X_t = woodwork.concat_columns([X_t, X[natural_language_cols]])
+        # Add back in the unchanged original natural language columns that we want to keep
+        if len(self._natural_language_cols) > 0:
+            X_t = woodwork.concat_columns([X_t, X.ww[self._natural_language_cols]])
+            # reorder columns to match original
+            X_t = X_t.ww[[col for col in original_schema.columns if col in X_t.columns]]
```
eccabay, line 144: 🙏 this will prevent future mystery bugs, thank you!

evalml/pipelines/components/transformers/imputers/target_imputer.py
```diff
@@ -93,9 +99,9 @@ def fit(self, X, y):
             raise TypeError("Provided target full of nulls.")
         y = y.to_frame()
 
-        # Convert all bool dtypes to category for fitting
+        # Return early since bool dtype doesn't support nans and sklearn errors if all cols are bool
```
eccabay, line 102: Coming off of the discussion in the simple imputer, this raises alarm bells for me. I don't think it's actually wrong, we only need the target imputer during pipeline fit, so if the training data is bool with no nans this component is a no-op. I think it would be better here to state the logical reason for why we don't need to fit on a bool column (akin to what I said above) than the technical reason, since a comment like this could be confusing in the future (it sounds like there's a scikit bug we'd have to revisit, which isn't the case)

evalml/tests/component_tests/test_imputer.py
```diff
@@ -724,3 +727,6 @@ def test_imputer_woodwork_custom_overrides_returned_by_components(
         assert {k: type(v) for k, v in transformed.ww.logical_types.items()} == {
             data: Double,
         }
+
+
+# --> add test for ww kept when passsed to child imputers
```
eccabay, line 732: Seems like something needs to be added here 👀

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -626,3 +626,39 @@ def test_simple_imputer_boolean_nullable_valid_train_empty_test():
...
+        },
+    )
+    X_train.ww.init(
+        logical_types={
+            "bools1": "Boolean",
+            "bools2": "Boolean",
+        },
+    )
+
+    imp = SimpleImputer()
```
eccabay, line 646: A lot of our other tests parameterize the imputation strategy - it'd be great (and I think would have caught what Jeremy mentioned above) to do the same here and test all the boolean imputation strategies!

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -626,3 +626,39 @@ def test_simple_imputer_boolean_nullable_valid_train_empty_test():
...
+    imp.fit(X_train)
+
+    # X_test will be BooleanNullable which will be a problem when _component_obj isn't fit
```
eccabay, line 649: I'm confused what this comment is saying here, could you clarify? 

You also explicitly cast the woodwork types to be BooleanNullable, but they're already automatically inferred to be. This isn't a bad thing, it protects us against future changes in woodwork inference, but I'm a bit confused with the context of the comment.

evalml/tests/component_tests/test_target_imputer.py
```diff
@@ -153,7 +153,7 @@ def test_target_imputer_fit_transform_all_nan_empty(y):
 def test_target_imputer_numpy_input():
     y = np.array([np.nan, 0, 2])
     imputer = TargetImputer(impute_strategy="mean")
-    y_expected = pd.Series([1, 0, 2], dtype="int64")
+    y_expected = pd.Series([1, 0, 2], dtype="float64")
```
eccabay, line 156: This change is interesting to me, do you know why it happened?

evalml/tests/component_tests/test_target_imputer.py
```diff
@@ -265,3 +265,25 @@ def test_target_imputer_woodwork_custom_overrides_returned_by_components(
...
+        imputer._component_obj.transform(y.ww.to_frame())[:, 0],
+        index=y.index,
+    )
+
+    if impute_strategy != "most_frequent":
+        # Confirm floating points are present
+        assert not (expected_y_t % 1 == 0).all()
+
+    for i in range(len(y_t)):
+        assert y_t.iloc[i] == expected_y_t.iloc[i]
```
eccabay, line 289: Can we use a `pd.assert_series_equal` here instead?

evalml/tests/component_tests/test_target_imputer.py
```diff
@@ -265,3 +265,25 @@ def test_target_imputer_woodwork_custom_overrides_returned_by_components(
             assert type(y_t.ww.logical_type) == logical_type
         else:
             assert type(y_t.ww.logical_type) == Double
+
+
+@pytest.mark.parametrize("impute_strategy", ["mean", "median", "most_frequent"])
+def test_target_imputer_doesnt_truncate_imputed_ints(impute_strategy):
```
eccabay, line 271: What's the need behind this new test?

evalml/tests/utils_tests/test_nullable_type_utils.py
```diff
@@ -199,7 +199,6 @@ def test_downcast_nullable_y_replaces_nullable_types(
     nullable_ltype,
     has_nans,
 ):
-    # --> remove this comment
```
eccabay, line 202: :joy:

eccabay (changes requested):
Similar to Jeremy's feedback, I think we need to find a better way to handle booleans in the simple imputer.

Other than that though, it's mostly just nitpicks. I really appreciate this work!

evalml/tests/component_tests/test_target_imputer.py
tamargrey: It's related to the `test_target_imputer_doesnt_truncate_imputed_ints` test I added that you ask about below. 

The target imputer was previously turning all `IntegerNullable` columns to `Integer` after imputing, which would truncate floating point values for the mean and median impute strategies. I confirmed that with the test and then switched it to use the Double logical type in those cases, which is why this now becomes `float64`. It was just coincidence that this example imputing `[nan, 0, 2]` has a whole number mean of 1, but we do want this to be a float (this behavior matches what the simple imputer does).

evalml/tests/component_tests/test_simple_imputer.py
tamargrey: This is just the follow-up to the first comment, explaining the test case where we start with `Boolean` types in X_train and then have BooleanNullable types with nans in X_test. Once I switch to convert boolean types instead of calling fit inside of `transform`, this test will stay the same, though maybe I'll add a check that the actual imputed values are the same as the values that would have been imputed if the types were `BooleanNullable` all along

evalml/pipelines/components/transformers/imputers/simple_imputer.py
tamargrey: Thank you both for keeping me honest here! I definitely missed the nuance that sklearn allows `bool` only data for "mean" and "median" strategies. I think there are two main questions to answer:

1. How should we handle "mean" and "median" strategies for the SimpleImputer? We only allow `knn` or `most_frequent` as strategies for boolean columns in the `Imputer` component, but we don't restrict that in our `SimpleImputer` component despite the fact that we actually get a `TypeConversionError` if you attempt to use `median` or `mean` with a boolean nullable column that contains nans and the result contains not whole numbers (the cases becca's examples show - this is on main and my branch). For now, I think we should ignore this behavior since fixing it should be given proper consideration and it's not reachable from automl. I'll make an issue. 
3. How should we handle the case when only `bool` dtypes are passed to the SimpleImputer (independent of impute strategy which only really works for bool cols when its `most_frequent`)? We need to find a way to fit the `_component_obj` at that point, and I think we should keep the changes as minimal as possible, and only convert those columns to `BooleanNullable` (which sklearn is fine with!) when there are only `bool` columns in the data. There's no need to convert other `bool` columns, and I think converting to string or numeric values just opens us up to unexpected behavior now that sklearn has so much support for the `boolean` dtype. I'll double check with all of the different types that there isn't any unexpected behavior, or course.

Really, I just wish sklearn didn't have this weird limitation. Seems to completely miss the exact use case we need it for which is different training and test data.

tamargrey: Created issue for the TypeConversionError: https://github.com/alteryx/evalml/issues/4050

evalml/pipelines/components/transformers/preprocessing/transform_primitive_components.py
tamargrey: Ticket to remove this workaround: https://github.com/alteryx/evalml/issues/4051

evalml/tests/component_tests/test_simple_imputer.py
tamargrey: Given how `mean` and `median` can raise a TypeConversionError for boolean data, I'm going to treat them as incompatible with boolean data for now, so I'm not going to test with them. I will, however, make the use of `most_frequent` explicit.

evalml/pipelines/components/transformers/imputers/target_imputer.py
tamargrey: Good call - I'll update the comment to just say we're returning early because there would never be any null values. In the SimpleImputer, I did end up deciding to mention the sklearn behavior when we convert to `boolean`, because that behavior is explicitly to work around that sklearn error, but I do also mention that we need to fit the `_component_obj` then in case the dtype is different in the training data.

evalml/pipelines/components/transformers/imputers/simple_imputer.py
jeremyliweishih: sounds good for 1. and agree with the `BooleanNullable` transform as long we set it back to `bool` after imputation!

jeremyliweishih (approved):

jeremyliweishih:
LGTM! great work 😄

evalml/pipelines/components/transformers/imputers/simple_imputer.py
tamargrey: Good call out on making sure it goes back to Boolean after. I thiiink that's what's happening but I'll confirm.

tamargrey: ah, we're only converting to `boolean` dtype in `fit`; we just then return early in transform if all the columns are `bool` , so the types will stay the same!

evalml/tests/component_tests/test_simple_imputer.py
```diff
@@ -626,3 +625,80 @@ def test_simple_imputer_boolean_nullable_valid_train_empty_test():
...
+            "bools1": pd.Series([True, False, True, True] * 20),
+            "bools2": pd.Series([True, False, True, False] * 20),
+        },
+    )
+    X_train.ww.init(
+        logical_types={
+            "bools1": "Boolean",
+            "bools2": "Boolean",
+        },
+    )
```
chukarsten, line 684: Is it worth adding the all bools columns to the `imputer_test_data`

tamargrey: Probably! 

I can add in a second all boolean no nulls column into the fixture that way I can get my "all bools" dataframe with a `select("Boolean")` call and then add nans in after fit to create the test data for this test. I can't think of anything that would be inherently problematic about adding a second `Boolean` column other than maybe some tests that only expect there to be one might fail.

chukarsten (approved):
This all looks solid to me.  I just had the one question about roping the small ad-hoc dataframes into the larger fixture, but that's about it.

Issue #3706, opened 2022-09-12
Clean up downcast functions


ParthivNaresh:
We currently have two functions for downcasting Nullable types to non-nullable types, `downcast_int_nullable_to_double`, and `downcast_nullable_types`. These two functions have different use cases.

The former aims to solely downcast `IntegerNullable` and `AgeNullable` to `Double` in order to support components that might not allow nullable integers.
The latter downcasts `BooleanNullable` to `Boolean` *and* `IntegerNullable` and `AgeNullable` to `Double` simultaneously. While `ignore_null_cols` as an argument can be set to `False` in order to downcast nullable numerics to `Double` without an issue, doing so for `BooleanNullable` will throw a type conversion error if nulls exist.

This issue covers creating one function that combines the flexibility of `downcast_int_nullable_to_double` with the coverage of `downcast_nullable_types`.

tamargrey:
just a note about ARIMA's incompatibility with nullable type for handling in the nullable types epic - you have to remove this whole block to see the error 

```python
        if X is not None:
            X = downcast_int_nullable_to_double(X)
            X = X.fillna(X.mean())
```
because the `fillna` removes the nullable types. We should probably keep the fillna call, but nullable type hanlding will happen elsewhere.

PR #4043, opened 2023-03-01
Component specific nullable types handling across EvalML

tamargrey:
closes https://github.com/alteryx/evalml/issues/3706

This is the PR into which all intermediate PRs will be merged for the implementation of nullable type handling.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -4,6 +4,8 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+        * Calculated partial dependence grid values for integer data by rounding instead of truncating fractional values :pr:`4096`
+        * Remove unnecessary logic from imputer components prior to nullable type handling :pr:`4038`
     * Documentation Changes
     * Testing Changes
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,10 +6,10 @@ Release Notes
     * Fixes
     * Changes
         * Calculated partial dependence grid values for integer data by rounding instead of truncating fractional values :pr:`4096`
-        * Remove unnecessary logic from imputer components prior to nullable type handling :pr:`4038`
-        * Added calls to ``_handle_nullable_types`` in component fit, transform, and predict methods when needed :pr:`4046`
-        * Remove existing nullable type handling across AutoMLSearch to just use new handling :pr:`4085`
-        * Handle nullable type incompatibility in ``Decomposer`` :pr:`4105`
+        * Removed unnecessary logic from imputer components prior to nullable type handling :pr:`4038`, :pr:`4043`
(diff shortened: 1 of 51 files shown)


docs/source/release_notes.rst
```diff
@@ -3,11 +3,16 @@ Release Notes
 **Future Releases**
     * Enhancements
         * Updated `pipeline.get_prediction_intervals()` to add trend prediction interval information from STL decomposer :pr:`4093`
+        * Add component-specific nullable type handling :pr:`4043`
```
eccabay, line 6: An alternative (and I'm not sure what's better here so please feel free to ignore if you disagree) would be to simply list this PR as a second PR for all the changes - i.e. `* Handled nullable type incompatibility in ``Decomposer`` :pr:`4105`, :pr:`4043` - there is some precedent for this method earlier in these notes!

I will also continue to be the release notes sticker, can all these comments be past tense? 😁

evalml/pipelines/components/transformers/samplers/oversampler.py
```diff
@@ -72,18 +72,46 @@ def fit(self, X, y):
...
+            pd.DataFrame, pd.Series: Transformed features and target.
+        """
+        X_ww = infer_feature_types(X)
+        original_schema = X_ww.ww.schema
+        if y is None:
+            raise ValueError("y cannot be None")
+        y_ww = infer_feature_types(y)
+        X_d, y_d = self._handle_nullable_types(X_ww, y_ww)
+        X_t, y_t = super().transform(X_d, y_d)
+        X_t.ww.init(schema=original_schema)
```
eccabay, line 106: I must have missed this or forgotten it from a previous PR - why do we need to reinit woodwork here?

eccabay (approved):
🎉 🎉 🎉

evalml/pipelines/components/transformers/samplers/oversampler.py
tamargrey: It's because we change the types of X and y before passing them into `super().transform(X_d, y_d)`, so it doesn't matter that `BaseSampler.transform` maintains the schema of the data passed in, because we don't want that schema! So I'm re-initing with the original schema here to maintain the original types (which should be possible because the data itself shouldn't be changing in a way that would invalidate the original schema)! 

But I'm now seeing that it looks like I forgot to do the same thing for y. It won't because there' no y incompatibility, but I'm gonna change the call to `_handle_nullable_types` to make that explicit. 

I can also add a comment to clarify why this is necessary

docs/source/release_notes.rst
tamargrey: I like that! Would be better at helping people find the PRs where the discussions happened if need be! 

(also sorry about making you be a broken record on the release notes thing. To be fair, I apparently cannot remember, but I also think I chose the wrong note in a couple of rebases, undoing your hard work from the past)

PR #4046, opened 2023-03-02
Use nullable type handling in components' fit, transform, and predict methods

### LINKED ISSUE #3924: LightGBM: Add nullable type handling for nullable X and y
This was first discovered in [this commit](https://github.com/alteryx/evalml/pull/3895/commits/cb4e1ee866d40dd5cebe274904cfd50880d1193b), which forced nullable type usage into AutoMLSearch. It can be reproduced if `feature_1` is set as either integer nullable or boolean nullable directly in the LightGBMClassifier, raising `ValueError: DataFrame.dtypes for data must be int, float or bool. Did not expect the data types in the following fields: 0`

```python
    X = pd.DataFrame(
        {
            "feature_1": [0, 0, 1, 1, 0, 1],
            "feature_2": ["a", "a", "b", "b", "c", "c"],
        },
    )
    X.ww.init(logical_types={"feature_1": "IntegerNullable", "feature_2": "categorical"})
    y = pd.Series([1, 1, 0, 0, 0, 1])
    
    clf = LightGBMClassifier()
    clf.fit(X, y)
```

This will not currently be seen in automl search because of the replace nullable types component, but we should consider adding nullable handling into the LightGBMClassifier class itself so that it can independently support nullable types.

### LINKED ISSUE #3926: ExponentialSmoothingRegressor: Add component specific nullable type handling for nullable y inputs
This was first discovered in [this commit](https://github.com/alteryx/evalml/pull/3895/commits/cb4e1ee866d40dd5cebe274904cfd50880d1193b), which forced nullable type usage into AutoMLSearch. It can be reproduced if nullable types are passed directly into the component:

```python
def test_model_instance(ts_data):
    X, _, y = ts_data()
    y = ww.init_series(y, logical_type="IntegerNullable")

    regressor = ExponentialSmoothingRegressor()
    fitted = regressor.fit(X, y)
```

This will not currently be seen in automl search because of the replace nullable types component, but we should consider adding nullable handling into the component class itself so that it can independently support nullable types.

### LINKED ISSUE #3974: Oversampler can raise ValueError: Unknown label type: when nullable y passed in
The following block of code will raise the `ValueError: Unknown label type: 'unknown'` error.

```python
    import woodwork as ww
    X, y = X_y_binary
    y = ww.init_series(y, logical_type="BooleanNullable")
    sn = Oversampler()
    _ = sn.fit_transform(X, y)

```

This will not currently be seen in automl search because of the replace nullable types component, but we should consider adding nullable handling into the component class itself so that it can independently support nullable types.

Note this is likely related to https://github.com/alteryx/evalml/issues/3923, https://github.com/alteryx/evalml/issues/3922 , and https://github.com/alteryx/evalml/issues/3910 , which all stem from the inability of sklearn's type_of_target to assign a proper type to nullable data

### LINKED ISSUE #3993: Call _handle_nullable_types in fit, transform, predict
The first step in using the component-specific nullable type handling will be to actually call `_handle_nullable_types` for the components that need it. This will involve updating the necessary components and any resulting test changes. 

```python
    # We can handle nullable types automatically if we want to
    def fit(self, X, y):
      ...
      X, y = self._handle_nullable_types(X, y)
      ...

# A component with more specific needs 
class MyEstimator(Estimator):
    # We can handle nullable types automatically if we want to 
    def predict(self, X):
      ...
      X, y = self._handle_nullable_types(X)
      ...

```
As part of this ticket we should also
- Understand the impact to automl search runtimes (expected to be increased, but the nullable types transformer would still be doing the actual transforming)
- Test use of nullable types at the pipeline level

This should not be merged into main until the rest of the needed changes to automl search to remove existing handlings 
have been made. They should all be merged to main at the same time.

### LINKED ISSUE #4001: TimeSeriesImputer: Add nullable type handling for X and y when interpolate is impute method
When `interpolate` is the impute method, we use pandas' `series.impute` method, which cannot handle nullable types: https://github.com/pandas-dev/pandas/issues/40252

As part of component-specific nullable type handling, we should remove the calls to `astype(float)` we currently do:
- https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/imputers/time_series_imputer.py#L175-L177
- https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/imputers/time_series_imputer.py#L191-L192

This will be replaced with separate calls to `_handle_nullable_types` in both X and y. Since we only need to transform when `self._impute_target == "interpolate"`, we can consider implementing a special `_handle_nullable_types` that takes `_self.impute_target` into account.

### LINKED ISSUE #4015: ARIMARegressor: Add component specific nullable type handling for IntegerNullable in X
Currently, we have a call to [downcast_int_nullable_to_double](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/estimators/regressors/arima_regressor.py#L210) that handles ATIMA's nullable type incompatibility.

With the implementation of component specific nullable type handling, we should use the new utils to perform this downcasting by setting the `_integer_nullable_incompatibilities` property to `["X"]`.

tamargrey:
closes #3993 closes #4001, closes #3924, closes #3974, closes #4015, closes #3926

This PR doesnt handle the objective functions (which we're waiting on for the next sklearn release to be released in time to not have to handle ourselves 🤞 )


machineFL:

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -155,35 +165,45 @@ def transform(self, X, y=None):
...
+        # so we save the original schema to recreate it where possible after imputation
+        original_schema = X.ww.schema
+        X, y = self._handle_nullable_types(X, y)
+
         X_not_all_null = X.ww.drop(self._all_null_cols)
-        X_schema = X_not_all_null.ww.schema
-        X_schema = X_schema.get_subset_schema(
-            subset_cols=X_schema._filter_cols(
-                exclude=["IntegerNullable", "BooleanNullable", "AgeNullable"],
-            ),
```
tamargrey, line 163: This change means that we'll no longer rely on woodwork inference for the nullable types after imputing. Instead, we're passing along these logical types even after imputation. 

This is something I've thought a lot on, since it doesn't immediately make sense that we wouldn't either continue to rely on type inference or specify the non nullable types that we know we can convert to after inference. 
- Don't rely on inference - this will take longer since type inference takes time, and it leaves the types up to woodwork type inference which, as we all know, can change at any time. Not leaving these types up to inference I think makes this component more efficient and more consistent.
- Keep the nullable types even after imputing - the goal of this epic is to be able to pass in nullable types wherever and whenever. We shouldn't need to avoid them anymore, so I don't want to add the extra complexity of determining what types to use unless it proves necessary. That might be if our dependencies show worse runtimes when continuing to use nullable types that are proving problematic for automl search times or if we ever choose to use the information we get from non nullable types that no nans can be present to optimize (ex - never try to impute non nullable data. Though, of course, that's at the imputer level where this conversation is happening, so it's a moot point).

docs/source/user_guide/timeseries.ipynb
```diff
@@ -881,13 +878,24 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Prior to using test data, make sure the types match the training data\n",
+    "X_test.ww.init(schema=X_train.ww.schema)\n",
```
tamargrey, line 891: This is now needed for the training and test data's woodwork schemas to match and not raise the `PipelineError` that the input types are different from the types the data was fitted on.

The training and test data's logical types are now different because of the changes to the time series imputer to no longer rely on Woodwork type inference for IntegerNullable input columns after they're imputed, so instead of reinferring to `Integer` to match the test data, we pass in the original `IntegerNullable`. 

Here's a longer explanation as to why the change in types isn't worrying me right now and what the implications of this are:

The training data has nans while the test doesnt because the TimeSeriesRegularizer introduces null values into the `"Numeric"` and `"Categorical"` columns in the training data, turning the logical type of `Numeric` from `Integer` into `IntegerNullable` which we then fill in with the time series imputer. 

Now that the time series imputer won't return the type back to `Integer` and instead keeps the original `IntegerNullable` type, it's as problem for the component graph that sees those as different schemas.

In this PR, I don't think it's a bad thing to maintain the train and test datas' types. It's what we currently have to do if there are discrepancies between whether nans are present in the train vs test data. But in the long run, I think we may want to change `_schema_is_equal` to allow the nullable and non nullable types to be used interchangeably in the input data to pipelines, which would increase the flexibility of our pipelines.

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -159,7 +169,7 @@ def test_categorical_and_numeric_input(imputer_test_data):
     y = pd.Series([0, 0, 1, 0, 1])
     imputer = TimeSeriesImputer()
     imputer.fit(X, y)
-    transformed, _ = imputer.transform(X, y)
+    transformed, _ = imputer.transform(X.ww.copy(), y)
```
tamargrey, line 172: This was needed because `_handle_nullable_types` actually changes the input data, so the second imputer calls were being run without any nullable types. 

I want to triple check that `_handle_nullable_types` mutating the input data isn't problematic. I seem to remember having a conversation in which I learned that evalml as a whole doesn't worry too much about not mutating input data, but I'm still worried this will prove to be problematic at the pipeline level somehow.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -52,7 +59,7 @@ class TimeSeriesImputer(Transformer):
     # Incompatibility: https://github.com/alteryx/evalml/issues/4001
     # TODO: Remove when support is added https://github.com/alteryx/evalml/issues/4014
     _integer_nullable_incompatibilities = ["X", "y"]
-    _boolean_nullable_incompatibilities = ["X", "y"]
```
tamargrey, line 55: Realized we could remove `X` from the boolean nullable incompatibilities because `interpolate` isn't a valid impute strategy for boolean columns in X (keeping y, though, because we don't stop interpolate from being called on boolean data - we currently convert to float, so I'm keeping that behavior for now - we should deal with that separately https://github.com/alteryx/evalml/issues/4053 and it shouldnt need to be part of the nullable types work)

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -194,10 +213,12 @@ def transform(self, X, y=None):
...
             elif self._impute_target == "interpolate":
-                # TODO: Revert when pandas introduces Float64 dtype
-                y_imputed = y.astype(float).interpolate()
+                y_imputed = y.interpolate()
                 y_imputed.bfill(inplace=True)
-            y_imputed = ww.init_series(y_imputed)
+            # --> https://github.com/alteryx/evalml/issues/4053
+            # by using the y.ww.logical_type, we'll get the downcast type as the final, which is expected and
+            # onyl problematic for boolean which will be floats which is nonsensical
+            y_imputed = ww.init_series(y_imputed, logical_type=y.ww.logical_type)
```
tamargrey, line 221: Wanted to highlight this behavior - we're avoiding woodwork inference and just using whatever type we downcast to for y. This will essentially match the current behavior in that we can end up with Boolean columns becoming Doubles with non 1s or 0s floating point values imputed for nans, which isn't great but isn't reachable from AutoML search.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -155,35 +161,45 @@ def transform(self, X, y=None):
...
             imputed.bfill(inplace=True)  # Fill in the first value, if missing
             X_not_all_null[X_interpolate.columns] = imputed
-        X_not_all_null.ww.init(schema=X_schema)
+
+            # Interpolate may add floating point values to integer data,
+            # so we'll want to use the Double logical type
+            # --> this will lose the age ltype and I'm okay with that in this edge case
+            new_ltypes = {
+                col: Double
+                if isinstance(ltype, (IntegerNullable, AgeNullable))
```
tamargrey, line 198: This will lose the AgeNullable logical type info, turning it into Double, but we doon't use the Age logical types a whole lot, and this is one small case where the extra logic didn't feel worth it. We can always add it in if we need it.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -189,8 +189,12 @@ def fit(self, X, y=None):
             self
         """
         X = infer_feature_types(X)
-        X_encoded = self._encode_categories(X, fit=True)
-        y_encoded = self._encode_labels(y)
+        # --> why wasn't this here before? is it a problem to add
+        if y is not None:
+            y = infer_feature_types(y)
```
tamargrey, line 194: we weren't initializing woodwork for y if it wasn't already set, which is problematic for `_handle_nullable_types` which would init woodwork with type inference, which might not be all that different than what ends up happening in `infer_feature_types`, but I figure we should use the common util when we can.

evalml/pipelines/components/estimators/regressors/lightgbm_regressor.py
```diff
@@ -170,8 +169,9 @@ def fit(self, X, y=None):
         X_encoded = self._encode_categories(X, fit=True)
         if y is not None:
             y = infer_feature_types(y)
-        X_encoded = downcast_int_nullable_to_double(X_encoded)
-        self._component_obj.fit(X_encoded, y)
+        # --> dont love calling this after we encode categories - try putting it first?
+        X_d, y_d = self._handle_nullable_types(X_encoded, y)
```
tamargrey, line 173: `_encode_categories` calls `infer_feature_types` for X the lightgbm estimators, and also has logic related to categorical columns that shouldn't apply to any BooleanNullable columns we convert to Categorical, so I think it makes sense to have the handle call occur after `_encode_categories`, but ideally we'd have the handle call before doing any other work.

evalml/pipelines/components/transformers/samplers/base_sampler.py
```diff
@@ -60,27 +59,10 @@ def _prepare_data(self, X, y):
...
-                    },
-                )
-            except ValueError:
-                X = X.astype(
-                    {
-                        null_col: float
-                        for null_col in X.ww.select(IntegerNullable).columns
-                    },
-                )
-            X.ww.init(schema=X.ww.schema)
```
tamargrey, line 79: This is nullable type handling that we can now remove.

Though worth noting that because the above `astype` calls are outside of woodwork, `X` had no ww types at this point, and this init call was actually with `schema=None` and we were just rerunning woodwork inference, which could've resulted in different oversamplers being chosen if inference changed which columns were Categorical (not super likely, which is why it never came up, I'm guessing)

So now we get to avoid the extra woodwork inference as well :)

evalml/tests/component_tests/test_oversampler.py
```diff
@@ -504,3 +526,26 @@ def test_oversampler_nullable_type_incompatibility(
...
+
+    oversampler = Oversampler()
+
+    assert not mock_handle_nullable_types.called
+    mock_handle_nullable_types.return_value = X, y
+
+    # Incompatibility is only when the sampler is actually called in transform,
+    # so we don't need to handle nullable types at fit
+    oversampler.fit(X, y)
+    assert not mock_handle_nullable_types.called
```
tamargrey, line 548: Want to highlight that for many of the components with incompatibilities, it's only in one of the fit/predict/transform methods, so I've chosen to not call `_handle_nullable_types` when it's not needed, which matches behavior I've seen used in the past of only making incompatibility-related changes in the exact context they're needed. 

Let me know if anyone has a problem with this, though. I made sure that none of the components have any logic that would change how the component is fit based on the downcasting of types to handle incompatibilities, but I may be missing something.

evalml/tests/component_tests/test_time_series_imputer.py
tamargrey: The fact that `_handle_nullable_types` mutates the input data was hiding some incompatibilities in predict/transform methods, because my `test_lgbm_with_nullable_types` test was passing in an `X` without nullable types to predict and predict_proba 🙃. 

I still can't think of a reason good enough to copy the data every time. I think we just need to be aware of this in tests that are checking that nullable types can be passed into components, and in transformers, make sure we're maintaining the original types in the transformed data.

eccabay: This is a great point and edge case - could you add some comments to the tests where you call copy for posterity?

evalml/tests/component_tests/test_time_series_imputer.py
```diff
@@ -562,26 +572,53 @@ def test_imputer_woodwork_custom_overrides_returned_by_components(
...
+    assert X.ww.get_subset_schema(
+        cols_expected_to_stay_the_same,
+    ) == X_imputed.ww.get_subset_schema(cols_expected_to_stay_the_same)
+    assert {
+        str(ltype)
+        for col, ltype in X_imputed.ww.logical_types.items()
+        if col in cols_expected_to_change
+    } == {"Double"}
+
+    assert isinstance(y_imputed.ww.logical_type, expected_imputed_ltype)
```
eccabay, line 620: This is interesting to me. Why do we cast `AgeNullable` to `Double` in X, but cast it to `AgeFractional` in y? Why not keep it consistent?

docs/source/user_guide/timeseries.ipynb
eccabay: It's also interesting to me that you had to update the docs because of this change, but none of our tests broke - seems like a testing gap. This may change depending on how we decide to resolve my above comment, but it would be great to have a test outside of the docs to catch this!

eccabay (changes requested):
Once again, awesome work! I'm a bit concerned about our handling of types in the TimeSeriesRegularizer/Imputer, and when we encounter errors.

evalml/tests/component_tests/test_time_series_imputer.py
tamargrey: I mentioned it [here](https://github.com/alteryx/evalml/pull/4046#discussion_r1128357111) - this was me not wanting to add additional complexity. 

After working on removing the nullable type logic from the rest of automl search (a separate, in progress PR), though, I'm now realizing that we'll need the logic to determine a logical type's fractional equivalent in other places as well, so I'm going to make a common util that I'll use there as well. 

I'll to pull out the [downcast_matches](https://github.com/alteryx/evalml/blob/main/evalml/utils/nullable_type_utils.py#L100) dict into a global variable `DOWNCAST_TYPE_DICT` and then to make a separate util in the nullable type utils called `determine_fractional_downcast_type` or something. So ultimately if we ever change logical types in woodwork, we only have to change that one dict.

tamargrey: Also, to get at specifically why `y` was able to keep the logical type when `X` wasn't: I set `y_imputed`'s logical type to be the original downcast logical type from `_handle_nullable_types`. So if it was `AgeNullable` when passed into `transform`, if any nans were present, we'd get `AgeFractional` and then could keep that type after imputing. But if no nans were present, it'd have been downcasted to `Age`, and we can keep that type anyway since no fractional values could have been imputed. 

This kind of breaks the pattern with what we do with `X` which is to reuse the types from the original schema prior to downcasting. If that feels wrong, I should be able to do something similar with saving the original y ltype and changing it to the fractional version explicitly if need be.

evalml/tests/component_tests/test_time_series_imputer.py
tamargrey: added comments: 827305987

evalml/pipelines/components/transformers/samplers/base_sampler.py
tamargrey: removed! fd06f5912

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
```diff
@@ -189,8 +189,11 @@ def fit(self, X, y=None):
             self
         """
         X = infer_feature_types(X)
-        X_encoded = self._encode_categories(X, fit=True)
-        y_encoded = self._encode_labels(y)
+        if y is not None:
+            y = infer_feature_types(y)
+        X_d, y_d = self._handle_nullable_types(X, y)
```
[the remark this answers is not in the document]
tamargrey: Ah, I kept meaning to bring the `_d` thing up - it's a way of indicating the data has been downcasted that I wanted to introduce (like `_t` indicates "has been transformed" and `_ww` "has been made into woodwork". If people dislike it, I can remove it.

In general, I wasn't particularly following any way of choosing to overwrite variable names or not. I was more following what the component itself was already doing. So the lgbm estimators weren't overwriting variable names (like the lgbm estimators chose a different `X_encoded` instead of overwriting `X` while exponental smoothing was doing stuff like `y = self._remove_datetime(y)`, so I kind of just tried to follow patterns. 

We seem to have a good mixture of how we decide to overwrite variable names across our components, so I didn't necessarily feel like this was the time to exert my hand at standardizing that. But also given that I'm attempting to introduce `_d`, I can see how keeping things standard would be extra important. Happy to standardize things if you feel like it'd clarify things up!

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -155,35 +162,44 @@ def transform(self, X, y=None):
...
             imputed = X_backward.bfill()
             imputed.pad(inplace=True)  # Fill in the last value, if missing
             X_not_all_null[X_backward.columns] = imputed
 
         if self._interpolate_cols is not None:
-            X_interpolate = X.ww[self._interpolate_cols]
-            # TODO: Revert when pandas introduces Float64 dtype
-            imputed = X_interpolate.astype(
-                float,
-            ).interpolate()  # Cast to float because Int64 not handled
```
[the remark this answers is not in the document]
tamargrey: yep! And the `test_time_series_imputer_nullable_type_incompatibility` test means we don't need to rely on TODOs as much

evalml/pipelines/components/transformers/samplers/oversampler.py
```diff
@@ -67,17 +67,34 @@ def fit(self, X, y):
...
 
         # get categorical features first, if necessary
         if sampler_name == "SMOTENC":
-            self._get_categorical(X)
-        super().fit(X, y)
+            self._get_categorical(X_ww)
+        super().fit(X_ww, y_ww)
         return self
 
+    def transform(self, X, y=None):
```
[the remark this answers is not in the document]
tamargrey: this is where my 10000 open PRs is biting me in the butt! tl;dr - we'll eventually need the handle method back in the Oversampler's transform, so I didn't feel like deleting this since it's not going into main.

Full explanation:

You're correct that this is doing nothing. Previously, I needed the `_handle_nullable_types` call because of an incompatibility in `y`, but that was resolved by sklearn 1.2.2, so I removed it and must've forgotten to remove this method. 

But we also have a new oversampler incompatibility in `X` ([this PR](https://github.com/alteryx/evalml/pull/4068)) that will eventually mean we need a call to _handle_nullable_types in this method, which'll mean we need this method. But since _this_ is the PR to add the calls to the handle method, we can't yet add it in the other PR.

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -476,11 +477,16 @@ def test_arima_regressor_can_forecast_arbitrary_dates(use_covariates, ts_data):
     )
 
 
-def test_arima_regressor_nullable_handling():
+@pytest.mark.parametrize(
+    "nullable_ltype",
+    ["IntegerNullable", "AgeNullable"],
+)
+# TODO In https://github.com/alteryx/evalml/issues/4009, add BooleanNullable to this
```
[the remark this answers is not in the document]
tamargrey: good call - removing

evalml/tests/component_tests/test_lgbm_regressor.py
```diff
@@ -289,3 +288,23 @@ def test_lgbm_regressor_nullable_type_incompatibility(
...
     sk_lgb.fit(X, y)
     sk_lgb.predict(X)
+
+
+@patch(
+    "evalml.pipelines.components.component_base.ComponentBase._handle_nullable_types",
+)
+def test_lgbm_calls_handle_nullable_types(
+    mock_handle_nullable_types,
+    X_y_binary,
```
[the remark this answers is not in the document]
tamargrey: hmmm, I see your point. 

My one sticking point is just that we have tests that confirm the incompatibilitiy in the dependency still exists and is fixed by our handle method- test_<component>_nullable_type_incompatibility like you said - and tests that confirm our component can actually take in nullable types. So I view these tests as tying those two together - the reason we can pass those types in is because we call `_handle_nullable_types`. But I just checked, and removing any of the handle calls or the incompatibility properties cause tests to fail, so I think that you're right that these tests are overkill at this point.

evalml/pipelines/components/estimators/classifiers/lightgbm_classifier.py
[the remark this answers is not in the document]
eccabay: I see what you mean. Seems like standardizing if/when we copy vs overwrite data within transformers is a great future tech debt item 🥲 

I don't have a strong opinion one way or the other as to whether we standardize the use of `_d` here. I do wonder about downstream impacts though - would this be related to the odd behavior in the docs at all?

evalml/tests/component_tests/test_lgbm_regressor.py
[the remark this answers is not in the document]
tamargrey: removed the tests 537b9f223

evalml/tests/component_tests/test_time_series_imputer.py
tamargrey: Added a common util to keep AgeFractional in X to more closely match the behavior in y, but not exactly f18ce4b1a

docs/source/user_guide/timeseries.ipynb
tamargrey: Responding to your points
1. I hadn't been thinking about the time series imputer that way (I didn't know that the imputer and regulizer only ever get used together, but it makes sense), but you're probably right that if a column was originally integer and it had to become int nullable because we inserted nans ourselves, we should keep that original type after we've fixed the data. I don't have a problem with having the time series imputer be an exception to the rule that our components should maintain the input type where possible--I'd make sure to have clear comments explaining why it's needed.
2. In general, I don't want us to shy away from keeping IntegerNullable columns as such even if no nans are present (whether we impute them ourselves or users input them). Those types aren't really meant to imply the presence of nans, just that the type can support null values, but other than that they're the same as non nullable integers. For example, in Featuretools, we might output types as IntegerNullable from a Primitive so that users could pass nans in and not have it break. Overall, handling any changes to _schema_is_equal is something I'll do in a separate PR to main so that we give it the proper time it deserves, and we can continue this discussion then.
3. I think the need for this call is a side effect of the fact that we change `X_train` and only X_train after splitting but prior to automl search in this doc; we were just lucky before that the types were the same. Even before these changes, the more a user manipulates only one of the train or test data, the higher the chance is that they make the two schemas incompatible. The fact that the incompatibility here came from calling evalml components outside of pipelines and I changed the behavior of one of those components isn't really indicative of a larger problem, imo. Using evalml components by themselves and then passing data into search isn't really our expected usecase, right?  But in a pipeline that contained the regulaizer and then imputer with the exact same original X train and test from after we insert nans into the data, we would have always had a problem with the fact that the types would be different. In this doc, we avoid that problem by fixing the data ourselves with the components. On the other hand, if we had X train and test with no nans but that the regulizer would insert nans into train but not test, the pipeline with the regulizer and imputer wouldn't have a problem. But those cases I 100% agree we need more tests for different variations of train vs test data that you might have. My preference again would be to do that in a PR to main with any `_schema_is_equal` changes we're considering, since I think this can be considered separately from the nullable type changes, which are only relevant because they increase types we support as inputs.
4. This is a very good question. I suspect it has something to do with a slight difference between how `ComponentGraph._transform_features` gets called from predict vs in a direct call to `transform_all_but_final`, but this is not immediately obvious to me, so I'll have to do some digging.

tamargrey: Regarding what's going on to allow `predict` but not `transform_all_but_final`:
- At predict for `TimeSeriesPipelineBase`, we do indeed call `self.transform_all_but_final`, but when the training data is present, as it is in our `predict` calls, we first `self._add_training_data_to_X_Y`. Inside `_add_training_data_to_X_Y`, we concat `X_train` with `X` but use the [logical types from X_train](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/time_series_pipeline_base.py#L136-L148) to set the woodwork types. So It means that the data we pass into transform all but final will always have the same logical types as the original training data. 
- At a direct `transform_all_but_final`, though, we don't pass in the training data, so we are passing in the original `X_test` which has the different types from the training data.

So I think you're right that most predict calls would have raised the same error; we just have slightly different behavior for time series pipelines. I think including proper testing for all of transform all but final, score, and predict with changes in nullable types for all problem types will be an important part of changing `_schema_is_equal` logic.

I'm gonna open an issue to put all of these thoughts into.

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
tamargrey: summarizing a convo from elsewhere: Performance testing to this point is showing the expected runtime improvements with no impact to model results.

docs/source/user_guide/timeseries.ipynb
tamargrey: opened issue: https://github.com/alteryx/evalml/issues/4077

evalml/utils/nullable_type_utils.py
```diff
@@ -97,14 +97,49 @@ def _determine_downcast_type(col):
...
     if col.isnull().any():
         return has_nans_ltype
 
     return no_nans_ltype
+
+
+def _determine_fractional_type(logical_type):
+    """Determines what logical type to use for integer data that has fractional values imputed.
+    - IntegerNullable becomes Double.
+    - AgeNullable becomes AgeFractional.
```
eccabay, line 110: I think booleans should be on this list as well, right? 

But a follow up question about the boolean fractional value, is returning Categorical the desired behavior? How would we get to that scenario, and where is it used?

eccabay (approved):
This looks great, thanks for all your extremely thorough work! I have some extremely nitpicky final comments but it really is great.

evalml/utils/nullable_type_utils.py
tamargrey: nope - this is intended only for IntegerNullable and AgeNullable. We do have one place that I know of that turns booleans into fractional values in the time series imputer for target values, but that's not reachable from automl search, and I put up an issue to change that since it's nonsensical: https://github.com/alteryx/evalml/issues/4053

I was debating if I should make it explicit with an assertion that this shouldn't be used for BooleanNullable values, but it felt a little overkill for an internal util. I will, however, add a sentence to the docstring making it clear why BooleanNullable shouldnt be used.

tamargrey: Okay, now seeing your other comment on just having the `equivalent` util return all types, I think that is a much better plan for this util as well. So I'll jsut update this one to return the same logical type if the type isnt int or age nullable

docs/source/user_guide/timeseries.ipynb
tamargrey: good call -- removing!

evalml/utils/nullable_type_utils.py
```diff
@@ -97,14 +97,49 @@ def _determine_downcast_type(col):
...
+    non_nullable_ltype, _ = DOWNCAST_TYPE_DICT[str(logical_type)]
+
+    return non_nullable_ltype
+
+
+DOWNCAST_TYPE_DICT = {
+    "BooleanNullable": ("Boolean", "Categorical"),
+    "IntegerNullable": ("Integer", "Double"),
+    "AgeNullable": ("Age", "AgeFractional"),
+}
```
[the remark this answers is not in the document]
tamargrey: yep!

evalml/utils/nullable_type_utils.py
tamargrey: made this change: 7a1b9f821

evalml/pipelines/components/transformers/imputers/time_series_imputer.py
```diff
@@ -155,35 +166,56 @@ def transform(self, X, y=None):
...
+        # We will use the non nullable equivalents where possible
+        original_schema = original_schema.get_subset_schema(
+            list(X_not_all_null.columns),
         )
+        new_ltypes = {
+            col: _determine_non_nullable_equivalent(ltype)
+            if isinstance(ltype, (IntegerNullable, AgeNullable, BooleanNullable))
+            else ltype
+            for col, ltype in original_schema.logical_types.items()
+        }
```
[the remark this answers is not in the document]
tamargrey: made this change: 7a1b9f821

evalml/utils/nullable_type_utils.py
```diff
@@ -97,14 +104,46 @@ def _determine_downcast_type(col):
...
+
+    Args:
+        logical_type (ww.LogicalType): The logical type whose non nullable equivalent we are determining.
+
+    Returns:
+        LogicalType to be used instead of nullable type when nans aren't present.
+    """
+    non_nullable_ltype, _ = DOWNCAST_TYPE_DICT.get(str(logical_type)) or (None, None)
+
+    return non_nullable_ltype or logical_type
```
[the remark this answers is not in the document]
tamargrey: oooh I always forget that. good call!

PR #4058, opened 2023-03-08
Dont allow knn as boolean impute strategy

### LINKED ISSUE #4057: Do not use KNN Imputer for boolean data
We currently allow "knn" as an option for `boolean_impute_strategy` in the `Imputer` component. When this is used, we will almost always impute `True` for missing values, as the sklearn knn imputer component converts the values to numeric 1s and 0s and imputes floating point values for the nans. We then convert the columns' types back to the `Boolean` logical type, which turns every numeric value that isn't 0 into `True`. 

This is not desired, so we should remove `knn` as an option for boolean data in the `Imputer` component and disallow boolean values in our `KNNImputer` component. 

```python
X = pd.DataFrame(
    {
        "some_nan": pd.Series([True, np.nan, False, False, False, False, False, False, False, False], dtype="boolean"),
    },
)
y = pd.Series([1, 0, 0, 1, 0]*2)
knn = KNNImputer(3)
X_t = knn.fit_transform(X, y)

sk_knn = Sk_KNNImputer(n_neighbors=3)
sk_X_t = sk_knn.fit_transform(X, y)
```

tamargrey:
closes #4057 

Removes "knn" from `boolean_impute_strategy` in the `Imputer` component and stops accounting for boolean data in the `KNNImputer` component.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -16,6 +16,7 @@ Release Notes
         * Unpinned ``networkx`` and updated minimum version :pr:`4035`
         * Increased ``scikit-learn`` version to 1.2.2 :pr:`4064`
         * Capped max ``holidays`` version to 0.21 :pr:`4064`
+        * Stop allowing ``knn`` as a boolean impute strategy :pr:`4058`
     * Documentation Changes
     * Testing Changes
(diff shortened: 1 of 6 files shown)


jeremyliweishih (approved):
LGTM

christopherbunn (approved):

PR #4059, opened 2023-03-08
Fix pipeline example erroring out on DFS

### LINKED ISSUE #4056: `generate_pipeline_example` does not work with DFSTransformer
Currently `generate_pipeline_example` runs `generate_pipeline_code` which uses the `__repr__` method of pipelines and components to construct the pipeline and its parameters. However, Featuretools features does not use a `__repr__` method that allows us to directly create a feature from its string representation. To fix this we will need to read in the features separately and set it in the pipeline string.

jeremyliweishih:
Fixes #4056.

This PR removes the features parameters in the pipeline example when DFS is present and adds in the features as an optional argument.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -19,6 +19,7 @@ Release Notes
         * Increased ``scikit-learn`` version to 1.2.2 :pr:`4064`
         * Capped max ``holidays`` version to 0.21 :pr:`4064`
         * Stop allowing ``knn`` as a boolean impute strategy :pr:`4058`
+        * Capped ``nbsphinx`` at < 0.9.0 :pr:`4071`
     * Documentation Changes
     * Testing Changes
         * Use ``release.yaml`` for performance tests on merge to main :pr:`4007`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,6 +9,7 @@ Release Notes
         * Add ``get_evalml_requirements_file`` :pr:`4034`
         * Pipelines with DFS Transformers will run fast permutation importance if DFS features pre-exist :pr:`4037`
     * Fixes
+        * Fixed ``generate_pipeline_example`` erroring out for pipelines with a ``DFSTransformer`` :pr:`4059`
     * Changes
         * Uncapped ``pmdarima`` and updated minimum version :pr:`4027`
         * Increase min catboost to 1.1.1 and xgboost to 1.7.0 to add nullable type support for those estimators :pr:`3996`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -24,6 +24,8 @@ Release Notes
     * Testing Changes
         * Use ``release.yaml`` for performance tests on merge to main :pr:`4007`
         * Pin ``github-action-check-linked-issues`` at v1.4.5 :pr:`4042`
+        * Updated tests to support Woodwork's object dtype inference for numeric columns :pr:`4066`
+        * Updated ``TargetLeakageDataCheck`` tests to handle boolean targets properly :pr:`4066`
 
 .. warning::
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,6 +17,7 @@ Release Notes
         * Unpinned ``networkx`` and updated minimum version :pr:`4035`
         * Increased ``scikit-learn`` version to 1.2.2 :pr:`4064`
         * Capped max ``holidays`` version to 0.21 :pr:`4064`
+        * Stop allowing ``knn`` as a boolean impute strategy :pr:`4058`
     * Documentation Changes
     * Testing Changes
         * Use ``release.yaml`` for performance tests on merge to main :pr:`4007`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,11 +9,17 @@ Release Notes
         * Add ``get_evalml_requirements_file`` :pr:`4034`
         * Pipelines with DFS Transformers will run fast permutation importance if DFS features pre-exist :pr:`4037`
     * Fixes
+<<<<<<< HEAD
         * Fixed ``generate_pipeline_example`` erroring out for pipelines with a ``DFSTransformer`` :pr:`4059`
+=======
+        * Remove nullable types handling for ``OverSampler`` :pr:`4064`
+>>>>>>> 2b6a4e010b81103016b5f57711a5a41ad8eccd45
     * Changes
         * Uncapped ``pmdarima`` and updated minimum version :pr:`4027`
         * Increase min catboost to 1.1.1 and xgboost to 1.7.0 to add nullable type support for those estimators :pr:`3996`
         * Unpinned ``networkx`` and updated minimum version :pr:`4035`
+        * Increased ``scikit-learn`` version to 1.2.2 :pr:`4064`
+        * Capped max ``holidays`` version to 0.21 :pr:`4064`
     * Documentation Changes
     * Testing Changes
         * Use ``release.yaml`` for performance tests on merge to main :pr:`4007`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,6 +8,7 @@ Release Notes
         * Add nullable type incompatibility properties to the components that need them :pr:`4031`
         * Add ``get_evalml_requirements_file`` :pr:`4034`
         * Pipelines with DFS Transformers will run fast permutation importance if DFS features pre-exist :pr:`4037`
+        * Add get_prediction_intervals() at the pipeline level :pr:`4052`
     * Fixes
         * Fixed ``generate_pipeline_example`` erroring out for pipelines with a ``DFSTransformer`` :pr:`4059`
         * Remove nullable types handling for ``OverSampler`` :pr:`4064`
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,11 +9,8 @@ Release Notes
         * Add ``get_evalml_requirements_file`` :pr:`4034`
         * Pipelines with DFS Transformers will run fast permutation importance if DFS features pre-exist :pr:`4037`
     * Fixes
-<<<<<<< HEAD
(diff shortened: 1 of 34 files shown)


evalml/pipelines/pipeline_base.py
```diff
@@ -709,6 +709,8 @@ def repr_component(parameters):
         parameters_repr = ", ".join(
             [
                 f"'{component}':{{{repr_component(parameters)}}}"
+                if component != "DFS Transformer"
```
jeremyliweishih, line 712: Need to exclude the `DFS Transformer` since the `__repr__` of a feature is not executable.

evalml/pipelines/utils.py
```diff
@@ -682,22 +696,23 @@ def generate_pipeline_example(
...
+{generate_pipeline_code(pipeline, path_to_features)}
 
 print(pipeline.name)
 print(pipeline.parameters)
 pipeline.describe()
 
-df = pd.read_csv(PATH_TO_TRAIN)
-y_train = df[TARGET]
-X_train = df.drop(TARGET, axis=1)
+df = ww.deserialize.from_disk(PATH_TO_TRAIN)
```
jeremyliweishih, line 705: moved to use woodwork serialization to keep feature origin info.

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -856,59 +857,96 @@ def test_generate_pipeline_example(
...
+    X_train, X_test, y_train, y_test = split_data(
+        X,
+        y,
+        problem_type=automl_type,
+        test_size=0.2,
+    )
+
+    X_train = pd.DataFrame(X_train)
+    X_train.columns = X_train.columns.astype(str)
+    es = ft.EntitySet()
```
jeremyliweishih, line 894: I can split the DFS/non-DFS case as well but I thought that the DFS case covers all the edge cases already so I kept it like this for now. LMK what you all think.

evalml/pipelines/utils.py
```diff
@@ -637,8 +638,19 @@ def generate_pipeline_code(element):
             element.__class__.__name__,
         ),
     )
+    has_dfs_and_features = (
+        "DFS Transformer" in element.component_graph.compute_order and features_path
+    )
+    if has_dfs_and_features:
+        code_strings.append("from featuretools import load_features")
+        code_strings.append(f'features=load_features("{features_path}")')
```
christopherbunn, line 646: Assuming that it is required for DFS, should we have a warning that pops up if `features_path` is none?

evalml/pipelines/utils.py
```diff
@@ -682,22 +696,23 @@ def generate_pipeline_example(
...
 print(pipeline.parameters)
 pipeline.describe()
 
-df = pd.read_csv(PATH_TO_TRAIN)
-y_train = df[TARGET]
-X_train = df.drop(TARGET, axis=1)
+df = ww.deserialize.from_disk(PATH_TO_TRAIN)
+y_train = df.ww[TARGET]
+X_train = df.ww.drop(TARGET)
+# print([ X_train.ww[c].ww.origin for c in X_train.ww.columns])
```
christopherbunn, line 708: Stray comment here

christopherbunn (changes requested):
Nit + question about warning but once addressed can approve!

evalml/pipelines/utils.py
jeremyliweishih: I made it so that DFS becomes a no-op without parameters if no features are fed in. I can def raise a warning though!

docs/source/release_notes.rst
```diff
@@ -9,7 +9,11 @@ Release Notes
         * Add ``get_evalml_requirements_file`` :pr:`4034`
         * Pipelines with DFS Transformers will run fast permutation importance if DFS features pre-exist :pr:`4037`
     * Fixes
+<<<<<<< HEAD
```
jeremyliweishih, line 12: oops need to fix

evalml/pipelines/pipeline_base.py
tamargrey: could this ever be problematic if there's separate `Numeric Pipeline` vs `Categorical Pipeline`? I forget if dfs transformer comes before we split pipelines, though, so I definitely might not have all the info here.

tamargrey: Same question goes for `has_dfs = "DFS Transformer" in element.component_graph.compute_order` below

jeremyliweishih: the `DFS Transformer` always comes first so I don't believe it should be a problem (and it'll always be named `DFS Transformer`

evalml/pipelines/utils.py
```diff
@@ -637,8 +639,22 @@ def generate_pipeline_code(element):
             element.__class__.__name__,
         ),
     )
+    has_dfs = "DFS Transformer" in element.component_graph.compute_order
```
chukarsten, line 642: I think we might want to reference the name attribute of the class here rather than use a string.  If we change the name of the DFS Transformer, which I think is likely, then this error will appear again and it might be unclear why.

evalml/pipelines/utils.py
```diff
@@ -637,8 +639,22 @@ def generate_pipeline_code(element):
...
+    if has_dfs and not features_path:
+        warnings.warn(
+            "This pipeline contains a DFS Transformer but no `features_path` has been specified. Please add a `features_path` or the pipeline code will generate a pipeline that does not run DFS.",
+        )
```
eccabay, line 646: What's our thought/process on using `warnings.warn` vs `logger.warning`? Do we have one? A quick search of the code shows instances of using both.

evalml/pipelines/utils.py
```diff
@@ -610,11 +611,12 @@ def make_pipeline(
     return pipeline
 
 
-def generate_pipeline_code(element):
+def generate_pipeline_code(element, features_path=None):
     """Creates and returns a string that contains the Python imports and code required for running the EvalML pipeline.
 
     Args:
         element (pipeline instance): The instance of the pipeline to generate string Python code.
+        features_path (str): path to features json. Defaults to None.
```
eccabay, line 619: Nit: can we clarify that this should be a json created from featuretools? It could be confusing to users as is.

chukarsten (changes requested):
I think if we just change the check for DFS Transformer to be more robust based on the name attribute, we're good!

evalml/pipelines/pipeline_base.py
```diff
@@ -709,6 +709,8 @@ def repr_component(parameters):
...
+                else f"'{component}':{{}}"
```
eccabay, line 713: I'm worried about explicitly excluding information this way. Without making note of what features we're expecting, we're opening ourselves up to confusion/bugginess if a user provides the wrong features json in `generate_pipeline_code`. I'd be much more comfortable if we were able to add the string names of features, or something similar.

eccabay (changes requested):
Agreed with Karsten about removing reliance on the DFS transformer's name, and I'm worried about the implications of removing necessary repro information from `__repr__`

evalml/pipelines/utils.py
jeremyliweishih:
```
[warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning

[logging.warning()](https://docs.python.org/3/library/logging.html#logging.warning) if there is nothing the client application can do about the situation, but the event should still be noted
```
https://docs.python.org/3/howto/logging.html#when-to-use-logging

I believe this matches the former!

evalml/pipelines/pipeline_base.py
jeremyliweishih: I agree but I cannot do so for a couple reasons:

1. the `__repr__` method on a featuretools feature is not executable - it looks something like `<Feature 1>`. We need the parameters (in this case the feature definitions) to be executable for our pipeline `__repr__` and generating pipeline code so grabbing the feature name won't change the behavior.
2. there isn't an easy way to rebuild features from scratch. Talking to @rwedge there is a good amount of information the feature needs to know (including dataset info) that isn't easily serialized into a string. Maybe something for the future but I think this is a good intermediate measure for now.

So basically I cannot keep the behavior our pipelines `__repr__` method unless featuretools features do as well. Would you be more comfortable if I issue another warning here?

eccabay: Gotcha, I understand the challenge here. I'm still very worried here about the chance for something to go wrong, but throwing a warning will have to do, I guess.

I still wonder if there's a way to reduce risk. Can we offer a util to save and load featuretools features from/to the DFSTransformer directly, or save the feature names as a parameter for a sanity check? That sort of thing would probably be out of scope, but we could file a followup issue.

jeremyliweishih: >  Can we offer a util to save and load featuretools features from/to the DFSTransformer directly

thats what I'm doing in the pipeline example! Loading in the features and setting the parameters of the `DFSTransformer`.

I can give verifying the feature names a try.

eccabay: Ah, true! I suppose I was more focused on the "saving" part. New request: can we make sure we mention this behavior in the docs? We probably don't need a full example, just a mention of what's necessary and maybe the code for how to save the featuretools features.

Looking forward to seeing the results of trying to verify the feature names 😄

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -918,6 +956,48 @@ def test_generate_pipeline_example(
     exec(pipeline_example)
     assert os.path.exists(output_path)
 
+    with pytest.raises(
+        ValueError,
+        match="Provided features in `features_path` do not match pipeline features. There is a different amount of features in the loaded features.",
+    ):
```
eccabay, line 962: This is already a huge test, there's a lot going on here and it may be hard to debug in the future. If it doesn't add too much repeated code, could you split these failure cases off into their own test?

eccabay (approved):
Looks great!

chukarsten (approved):

PR #4064, opened 2023-03-09
Remove oversampling nullable handling

### LINKED ISSUE #4013: Oversampler: Remove nullable type handling when imblearn adds nullable type support
Once https://github.com/scikit-learn-contrib/imbalanced-learn/issues/969 has been implemented and released, we should upgrade imblearn to that version, which will allow us to remove the nullable type handling put in place by https://github.com/alteryx/evalml/issues/3974.

jeremyliweishih:
Fixes #4013.

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -28,7 +28,7 @@ outputs:
         - pandas >=1.5.0
         - dask >=2022.2.0,!=2022.10.1
         - scipy >=1.5.0
-        - scikit-learn >=1.2.1
+        - scikit-learn >=1.2.2
         - scikit-optimize >=0.9.0
         - statsmodels >=0.12.2
         - colorama >=0.4.4
@@ -46,6 +46,7 @@ outputs:
         - tomli >=2.0.1
         - packaging >=23.0
         - black[jupyter] >=22.3.0
+        - holidays <0.21, >=0.13
     test:
       imports:
         - evalml
--- a/evalml/tests/component_tests/test_oversampler.py
+++ b/evalml/tests/component_tests/test_oversampler.py
@@ -3,8 +3,6 @@ import copy
 import numpy as np
 import pandas as pd
 import pytest
-import woodwork as ww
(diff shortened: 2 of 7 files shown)


tamargrey:
These tests are going to fail unless we raise the min sklearn version, I think https://github.com/alteryx/evalml/actions/runs/4376580073/jobs/7658890406#step:8:4233 since sklearn 1.2.1 is going to still have problems with nullable types

evalml/tests/component_tests/test_oversampler.py
```diff
@@ -474,19 +474,8 @@ def test_oversampler_handle_nullable_types(
...
-    "handle_incompatibility",
-    [
-        True,
-        pytest.param(
-            False,
-            marks=pytest.mark.xfail(strict=True, raises=ValueError),
-        ),
-    ],
-)
 def test_oversampler_nullable_type_incompatibility(
```
tamargrey, line 477: How would you feel about removing this test completly? We have the test above this that checks that we can pass in nullable types without erroring (which I guess is passing along with this one because we don't run min dependency tests with it?)

jeremyliweishih: ah I missed that - can remove. They're both failing for the min tests: https://github.com/alteryx/evalml/actions/runs/4376580073/jobs/7658889954

.github/meta.yaml
```diff
@@ -28,7 +28,7 @@ outputs:
         - pandas >=1.5.0
         - dask >=2022.2.0,!=2022.10.1
         - scipy >=1.5.0
-        - scikit-learn >=1.2.1
+        - scikit-learn >=1.2.2
```
tamargrey, line 31: can we add a release note mentioning this and the holidays dependency changes?

jeremyliweishih: good call!

tamargrey (approved):

chukarsten (approved):

PR #4068, opened 2023-03-10
Handle new oversampler nullable type incompatibility in X

tamargrey:
closes https://github.com/alteryx/evalml/issues/3974

This incompatibility only shows up when string categorical columns are present with the nullable pandas dtypes, but I chose not to add a special `_handle_nullable_types` implementation since it's dataset dependent vs the time series imputer that needed a different handling but that would be applied to any dataset. This is a bit of a gray area, though, so I'm open to changing it.

### DIFF
--- a/evalml/tests/component_tests/test_oversampler.py
+++ b/evalml/tests/component_tests/test_oversampler.py
@@ -529,6 +529,8 @@ def test_oversampler_nullable_type_incompatibility(
         evalml_oversampler = Oversampler(sampling_ratio=0.5)
         X, y = evalml_oversampler._handle_nullable_types(X, y)
 
+    # We have to convert to `object` dtype with categorical columns
+    # because of the sklearn bug described by the test below this one
     X["categorical col"] = X["categorical col"].astype("object")
 
     im_oversampler.fit_resample(X, y)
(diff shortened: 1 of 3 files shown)


evalml/tests/component_tests/test_oversampler.py
```diff
@@ -447,14 +450,117 @@ def test_oversampler_handle_nullable_types(
...
+    "handle_incompatibility",
+    [
+        True,
+        pytest.param(
+            False,
+            marks=pytest.mark.xfail(strict=True, raises=ValueError),
+        ),
+    ],
+)
+def test_oversampler_category_dtype_incompatibility(
```
tamargrey, line 549: This is actually the same bug as the nullable type incompatibility, and it's why we convert categorical columns to "object" in [base_sampler](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/samplers/base_sampler.py#L99-L105). There is a fix in the works https://github.com/scikit-learn/scikit-learn/pull/25814 (because of the wonderful @thomasjpfan who has been a key contributor to our ability to support nullable types so widely), so I wanted to add this xfailing test to remind us to also remove that logic at the same time.

tamargrey:
Note - this doesn't need to get in for the upcoming release

chukarsten (approved):
Seems like a solid approach to me.  Thanks for being so thorough in the documentation of the tests and what we need to undo when it succeeds.

evalml/tests/component_tests/test_oversampler.py
tamargrey: @eccabay converting `caegory` to `object` was a bug fix of yours from a [while back](https://github.com/alteryx/evalml/pull/3732), so I just want to make sure you get a chance to take a look at this test in case I'm missing something here.

eccabay (approved):

PR #4072, opened 2023-03-13
Remove nullable handlings where possible from sklearn 1.2.2 upgrade 

### LINKED ISSUE #3992: Objective Functions: Add nullable type handling logic
- Once the base classes have the necessary logic for handling nullable types, we can add the related logic to the individual objective functions that need it. Note - we should not be adding calls to _handle_nullable_types to score calls, because we do not yet want to include this handling in AutoMLSearch.

```python
# An objective that can't handle any nullable types and uses standard handling  
class Objective1(RegressionObjective):
    _integer_nullable_incompatible = True
    _boolean_nullable_incompatible = True
```

This issue includes the following:

- Adding needed properties to objectives
- Adding tests that track the dependencies that need to add support that will fail when support is added
- Adding tests for all objectives to confirm that they support nullable types - should trigger a failure if a new objective is added that needs nulalble type handling in place - note - only for regression and binary classification objectives; multiclass cant have nullable y
- Testing that when using the properties for the objectives in conjunction with _handle_nullable_types transforms the data correctly and adds support - note this will change to not have the explicit _handle_nullable_types call once we integrate into automl search.
- Testing that there is not yet any change to automl search performance
- Can be completed in multiple independent PRs if needed

### LINKED ISSUE #4018: Objective Functions: Remove nullable type handling when sklearn adds support
When https://github.com/scikit-learn/scikit-learn/issues/25578 is completed and released, we can upgrade evalml and remove the nullable type handling put in place by https://github.com/alteryx/evalml/issues/3992

### LINKED ISSUE #4019: confusion_matrix, roc_curve: Use standardized nullable type handling
Currently, we handle nullable types for the `confusion_matrix` and `roc_curve` with the `_convert_ww_series_to_np_array` util, which works around the inability of `Series.to_numpy` to recognize the pandas nullable dtypes, resulting in numpy data with the `object` dtype, which causes errors down the line. 

We should move `_convert_ww_series_to_np_array` to the nullable type handling utils file and rename it to make it celar it has to do with nullable type handling.

### LINKED ISSUE #4020: confusion_matrix: Remove nullable type handling when possible
There are two ways we could avoid having to explicitly set the numpy dtypes of nullable data after a `Series.to_numpy` call. 
- Pandas stops converting nullable dtypes to `object` in `Series.to_numpy` call: https://github.com/pandas-dev/pandas/issues/48891 - then we could just call `to_numpy` and not have to worry about setting the correct dtype
- Sklearn supports the nullable dtypes in `unique_labels` (https://github.com/scikit-learn/scikit-learn/issues/25634) and `sklearn_confusion_matrix` (https://github.com/scikit-learn/scikit-learn/issues/25635) - currently neither do, which is why we handle the dtype changes prior to making those calls. Once support is added, we can stop converting the data to numpy arrays before passing in y true and predicted

If either of the above are completed and released, we can upgrade to that version and make any needed changes to remove nullable type handling from https://github.com/alteryx/evalml/issues/4019.

### LINKED ISSUE #4021: roc_curve: remove nullable type handling when possible
There are two ways we could avoid having to explicitly set the numpy dtypes of nullable data after a Series.to_numpy call.
- Pandas stops converting nullable dtypes to object in Series.to_numpy call: https://github.com/pandas-dev/pandas/issues/48891 - then we could just call to_numpy and not have to worry about setting the correct dtype
- Sklearn supports the nullable dtypes in the `LabelBinarizer` https://github.com/scikit-learn/scikit-learn/issues/25637

If either of the above are completed and released, we can upgrade to that version and make any needed changes to remove nullable type handling from https://github.com/alteryx/evalml/issues/4019.

### LINKED ISSUE #4054: Call _handle_nullable_types in objective_function
```python
# General objective nullable type logic
class MyObjective(BinaryClassificationObjective):
    # Method where incompatibilities are handled
    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
        # Looks at _integer_nullable_incompatible 
        # and _boolean_nullable_incompatible to determine what transformation to do
        ...
        self._handle_nullable_types(y_true)
        ...
```

As part of this ticket we should also
- Understand the impact to automl search runtimes (expected to be increased, but the nullable types transformer would still be doing the actual transforming)
- Test use of nullable types at the pipeline level
This should not be merged into main until the rest of the needed changes to automl search to remove existing handlings
have been made. They should all be merged to main at the same time.

tamargrey:
closes #4021, closes #4020, closes #4018, closes #3992, closes #4019, closes #4054

Removes logic related to model understanding nullable type handling and Objective nullable type handling, since sklearn 1.2.2 allows us to fully support nullable types!

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,7 @@ Release Notes
     * Fixes
     * Changes
         * Added Oversampler nullable type incompatibility in X :pr:`4068`
+        * Removed nullable handling from objective functions, ``roc_curve``, and ``correlation_matrix`` :pr:`4072`
     * Documentation Changes
     * Testing Changes
 
--- a/evalml/model_understanding/metrics.py
+++ b/evalml/model_understanding/metrics.py
@@ -9,31 +9,9 @@ from sklearn.metrics import precision_recall_curve as sklearn_precision_recall_c
 from sklearn.metrics import roc_curve as sklearn_roc_curve
 from sklearn.preprocessing import LabelBinarizer
 from sklearn.utils.multiclass import unique_labels
-from woodwork.logical_types import BooleanNullable, IntegerNullable
 
 from evalml.exceptions import NoPositiveLabelException
 from evalml.utils import import_or_raise, infer_feature_types, jupyter_check
-
-
-def _convert_ww_series_to_np_array(ww_series):
-    """Helper function to properly convert IntegerNullable/BooleanNullable Woodwork series to numpy arrays.
-
-    Args:
-        ww_series: Woodwork init-ed series possibly containing IntegerNullable or BooleanNullable datatype
-
-    Returns:
-        numpy.ndarray: The values of ww_series but in an array.
-    """
-    np_series = ww_series.to_numpy()
-    if isinstance(ww_series.ww.logical_type, BooleanNullable):
-        np_series = np_series.astype("bool")
-    if isinstance(ww_series.ww.logical_type, IntegerNullable):
-        try:
-            np_series = np_series.astype("int64")
-        except TypeError:
-            np_series = ww_series.astype(float).to_numpy()
-
-    return np_series
 
 
 def confusion_matrix(y_true, y_predicted, normalize_method="true"):
@@ -269,6 +247,7 @@ def roc_curve(y_true, y_pred_proba):
 
     Args:
         y_true (pd.Series or np.ndarray): True labels.
+        # --> also pd dataframe for multiclass
         y_pred_proba (pd.Series or np.ndarray): Predictions from a classifier, before thresholding has been applied.
 
     Returns:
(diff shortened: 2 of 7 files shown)


evalml/model_understanding/metrics.py
```diff
@@ -282,27 +258,30 @@ def roc_curve(y_true, y_pred_proba):
                   * `auc_score`: The area under the ROC curve.
     """
     y_true_ww = infer_feature_types(y_true)
-    y_true_np = _convert_ww_series_to_np_array(y_true_ww)
-    y_pred_proba = infer_feature_types(y_pred_proba).to_numpy()
-
-    if len(y_pred_proba.shape) == 1:
-        y_pred_proba = y_pred_proba.reshape(-1, 1)
```
tamargrey, line 289: The `roc_curve` changes are so drastic to allow us to pass in nullable types, because we were manipulating `y_true` and `y_predict_proba` as numpy arrays instead of pandas series/dataframes.

We generally tend to stay in pandas in evalml, so I don't think it's so problematic to stay in pandas to avoid the extra nullable types logic. Though worth noting a slight increase in runtime when we switch to using pandas (the difference was proportional as I increased number of rows and whether or not the dtype was `int64` or `Int64`). Right now, I'm thinking that this is worth being able to remove the nullable type logic.

image

The reason we can't stay in numpy is that a `series.to_numpy()` call from pandas nullable dtypes will always result in data with the `object` dtype, and this is not something numpy is going to change. Previously, we couldn't pass in the nullable types to the `label_binarizer` or `sklearn_roc_curve`, so moving to numpy right away via `_convert_ww_series_to_np_array` made sense, but now there's no need. But now that we can pass in pandas data with nullable dtypes, we should follow the rest of evalml and stay in pandas when possible.

evalml/objectives/standard_metrics.py
```diff
@@ -838,9 +838,9 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
                 "targets contain the value 0.",
             )
         if isinstance(y_true, pd.Series):
-            y_true = y_true.values
```
tamargrey, line 841: this change was needed because `.values` doesn't produce a numpy array when used with nullable. `Int64` would produce an `<IntegerArray>`, which then will be problematic for the `.mean` call below. Changing to `to_numpy` fixed this, so I'm not calling it an integer nullable incompatibility since the workaround doesnt need to worry about the types.

evalml/tests/objective_tests/test_objectives.py
```diff
@@ -194,60 +194,40 @@ def test_get_objectives_all_expected_ranges(obj):
...
+        # FraudCost needs an "amount" column
+        X = pd.DataFrame({"amount": [100, 5, 250, 89] * 5})
+    elif isinstance(obj, MAPE):
+        if isinstance(y_true.ww.logical_type, BooleanNullable):
+            pytest.skip("MAPE doesn't support inputs containing 0")
+        # Replace numeric inputs containing 0
+        y_true = y_true.ww.replace({0: 10})
+        y_pred = y_pred.replace({0: 10})
+
+    obj.score(y_true=y_true, y_predicted=y_pred, X=X)
```
eccabay, line 233: For sanity's sake, should we be checking any equality or non-nullness here?

docs/source/release_notes.rst
```diff
@@ -20,6 +20,7 @@ Release Notes
         * Capped max ``holidays`` version to 0.21 :pr:`4064`
         * Stop allowing ``knn`` as a boolean impute strategy :pr:`4058`
         * Capped ``nbsphinx`` at < 0.9.0 :pr:`4071`
+        * Remove nullable handling from objective functions, ``roc_curve``, and ``correlation_matrix`` :pr:`4072`
```
eccabay, line 23: I am still the release note nitpicker - `Remove` -> `Removed` 😄

eccabay (approved):
LGTM! Just small questions.

evalml/model_understanding/metrics.py
```diff
@@ -282,27 +258,30 @@ def roc_curve(y_true, y_pred_proba):
...
-    nan_indices = np.logical_or(pd.isna(y_true_np), np.isnan(y_pred_proba).any(axis=1))
-    y_true_np = y_true_np[~nan_indices]
+    y_pred_proba = infer_feature_types(y_pred_proba)
+
+    # Standardize data to be a DataFrame even for binary target
+    if isinstance(y_pred_proba, pd.Series):
+        y_pred_proba = pd.DataFrame(y_pred_proba)
+    # Only use one column for binary inputs that are still a DataFrame
+    elif y_pred_proba.shape[1] == 2:
+        y_pred_proba = pd.DataFrame(y_pred_proba.iloc[:, 1])
```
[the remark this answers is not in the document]
tamargrey: yeah, I also struggled to understand this, but it's specifically meant to allow users to pass in the results from binary classification `predict_proba` without having to pull out the positive class, since we don't make any such requirements for the multiclass case. It was suggested by freddy in this comment: https://github.com/alteryx/evalml/pull/1164#discussion_r488910475. We test it here: https://github.com/alteryx/evalml/blob/main/evalml/tests/model_understanding_tests/test_metrics.py#L347-L366

I tried to add some clarity with the comment `Only use one column for binary inputs that are still a DataFrame` but let me know if there's a better way to clarify what's going on here

evalml/tests/objective_tests/test_objectives.py
tamargrey: added a check that the value isnt null: b5ba0a2b9

christopherbunn (approved):

PR #4093, opened 2023-03-20
Updated `pipeline.get_prediction_intervals()` to add trend prediction interval information from STL decomposer

### LINKED ISSUE #4060: `pipeline.inverse_transform()` is not ran when determining prediction intervals for decomposer pipelines
If a pipeline has a decomposer, the residuals are presented when the prediction intervals are calculated. Instead of the residuals, we should use the recomposed prediction value to calculate the interval.

christopherbunn:
Resolves #4060

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
@@ -97,7 +97,9 @@ class STLDecomposer(Decomposer):
             period=self.period,
         )
         stlf_res = stlf.fit()
-        forecast = stlf_res.forecast(units_forward)
+        # forecast = stlf_res.forecast(units_forward)
+        ps = stlf_res.get_prediction(len(self.trend), len(self.trend) + units_forward - 1)
+        forecast = ps.summary_frame(.05)["mean"]
 
         # Handle out-of-sample forecasts.  The forecast will have additional data
         # between the end of the in-sample data and the beginning of the
--- a/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
@@ -97,9 +97,7 @@ class STLDecomposer(Decomposer):
             period=self.period,
         )
         stlf_res = stlf.fit()
-        # forecast = stlf_res.forecast(units_forward)
-        ps = stlf_res.get_prediction(len(self.trend), len(self.trend) + units_forward - 1)
-        forecast = ps.summary_frame(.05)["mean"]
+        forecast = stlf_res.forecast(units_forward)
 
         # Handle out-of-sample forecasts.  The forecast will have additional data
         # between the end of the in-sample data and the beginning of the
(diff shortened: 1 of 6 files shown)


machineFL:

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -382,3 +382,68 @@ def _decompose_target(X, y, fh):
...
+            self.trend,
+            ARIMA,
+            model_kwargs=dict(order=(1, 1, 0), trend="t"),
+            period=self.period,
+        )
+        stlf_res = stlf.fit()
+        ps = stlf_res.get_prediction(
+            len(self.trend),
+            len(self.trend) + units_forward - 1,
+        )
```
chukarsten, line 420: All of this is `.project_trend()`, so it ipossible to just save off the `STLForecast` prediction results and use them later?

christopherbunn: Will go with the first option

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -382,3 +390,54 @@ def _decompose_target(X, y, fh):
...
+                ] = prediction_interval_result[f"{coverage[i]}_upper"][overlapping_ind]
+            else:  # y.index is not datetime (e.g. int)
+                prediction_interval_result[
+                    f"{coverage[i]}_lower"
+                ] = prediction_interval_result[f"{coverage[i]}_lower"][-len(y) :]
+                prediction_interval_result[
+                    f"{coverage[i]}_upper"
+                ] = prediction_interval_result[f"{coverage[i]}_upper"][-len(y) :]
+                prediction_interval_result[f"{coverage[i]}_lower"].index = y.index
+                prediction_interval_result[f"{coverage[i]}_upper"].index = y.index
```
eccabay, line 441: This is a lot of repeated code, which makes it really hard to read. When we chunk the columns one at a time, any rows that are in `prediction_interval_result` but not in `y` end up being null rows. We also shouldn't have to set the index for each column individually, they're always going to be the same. I understand why we might have to go column by column if there are multiple alphas, but we can split that apart with a few more local variables. I haven't tested it, but I think we can replace this whole chunk with something like:
```
intervals = {
    "lower": result["mean_ci_lower"] - result["mean"],
    "upper": result["mean_ci_upper"] - result["mean"],
}
if len(overlapping_ind) > 0:
    intervals = intervals.iloc[overlapping_ind]
else:
    intervals = intervals[-len(y) : ]
    intervals.index = y.index
prediction_interval_result[f"{coverage[i]}_lower"] = intervals["lower"]
prediction_interval_result[f"{coverage[i]}_upper"] = intervals["upper"]
```

evalml/pipelines/time_series_regression_pipeline.py
```diff
@@ -209,15 +209,32 @@ def get_prediction_intervals(
             X_train=X_train,
             y_train=y_train,
         )
-        if self.estimator.model_family in self.NO_PREDS_PI_ESTIMATORS:
+        has_stl = "STL Decomposer" in list(
```
eccabay, line 212: We're probably going to want to make this slightly more robust here and check for `STLDecomposer.name` instead of hardcoding the decomposer name. The same holds below on line 229 as well.

eccabay (changes requested):
Looks pretty good! Just a few comments about code conciseness and expected behavior.

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
```diff
@@ -346,3 +346,47 @@ def test_stl_decomposer_doesnt_modify_target_index(
...
+
+    trend_pred_intervals = stl.get_trend_prediction_intervals(
+        y_validate,
+        coverage=coverage,
+    )
+    assert_pred_interval_coverage(trend_pred_intervals)
+
+    # Set y.index to be non-datetime
+    y_validate = y_validate.reset_index(drop=True)
+    trend_pred_intervals = stl.get_trend_prediction_intervals(
```
chukarsten, line 388: You should just parametrize this with a descriptive name for the parameter like "index_type": ["integer", "datetime"].  We should try to get away from single tests that have multiple, sequential assertions in them as we'll have to sometimes fix the first assertion, rerun the test, fix the second assertion, rerun the test...rather than just have the test failure parameters show us exactly what behavior we broke.

chukarsten (changes requested):
Thanks for refactoring some of that code out of `_project_trend()`, definitely looks more concise.  Left a few comments to address, but looking forward to getting this out!

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -382,3 +390,54 @@ def _decompose_target(X, y, fh):
...
+        Args:
+            y (pd.Series): Target data.
+            coverage (list[float]): A list of floats between the values 0 and 1 that the upper and lower bounds of the
+                prediction interval should be calculated for.
+
+        Returns:
+            dict of pd.Series: dict: Prediction intervals, keys are in the format {coverage}_lower or {coverage}_upper.
+        """
+        if coverage is None:
+            coverage = [0.95]
```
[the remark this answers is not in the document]
eccabay: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

eccabay (approved):

PR #4096, opened 2023-03-21
Explicitly round grid values for integer data in partial dependence

### LINKED ISSUE #4095: Partial Dependence can raise TypeConversionError if non-integer grid values are produced for IntegerNullable column
Currently in partial dependence, if you are calculating the partial dependence for an Integer logical type column, when we get the grid values with _grid_from_X, if there are more unique values in the integer column than the grid resolution, we calculate grid values to use with mscipy.sttats.mquantiles. This can introduce fractional values. With the Integer logical type, we can set a column with fractional values to the int64 dtype and pandas will truncate the value. Our IntegerNullable logical type, which uses Int64 dtype, on the other hand, does not allow that. Because of this, we’ll need to make some change to allow partial dependence on IntegerNullable columns (I don’t view this as an integer nullable incompatibility, since I think pandas is probably right to not just silently truncate your data).

```python
    y = ww.init_series(pd.Series([True, False]*25), logical_type="Boolean")
    X = pd.DataFrame({
        "col": pd.Series(range(len(y)))
    })
    X.ww.init(logical_types={"col": "IntegerNullable"})

    pipeline = logistic_regression_binary_pipeline

    pipeline.fit(X, y)
    partial_dependence(
        pipeline,
        X,
        grid_resolution=10,
        features="col",
    )
```
We will need to handle the IntegerNullable case, and while we're doing that, we should make the handling of fractional values for Integer columns explicit by rounding instead of letting the values get truncated.

tamargrey:
closes #4095 

This PR fixes the above issue by choosing to round the fractional grid values explicitly in `_grid_from_X` rather than leave it up to the integer dtypes to truncate values or raise an error at woodwork initalization.

### DIFF
--- a/evalml/tests/model_understanding_tests/test_partial_dependence.py
+++ b/evalml/tests/model_understanding_tests/test_partial_dependence.py
@@ -3053,13 +3053,10 @@ def test_partial_dependence_with_nullable_types(
         grid_resolution=grid_resolution,
         features=int_col,
     )
-    assert len(part_dep["partial_dependence"]) == min(
+
+    assert len(part_dep) == min(
         grid_resolution,
-        len(X[int_col].dropna().unique()),
-    )
-    assert len(part_dep["feature_values"]) == min(
-        grid_resolution,
-        len(X[int_col].dropna().unique()),
+        len(X[int_col].unique()),
     )
     assert not part_dep.isnull().any(axis=None)
 
@@ -3080,8 +3077,7 @@ def test_partial_dependence_with_nullable_types(
         grid_resolution=grid_resolution,
         features="bool col nullable",
     )
-    assert len(part_dep["partial_dependence"]) == 2
-    assert len(part_dep["feature_values"]) == 2
+    assert len(part_dep) == 2
     assert not part_dep.isnull().any(axis=None)
 
     fast_part_dep = partial_dependence(
@@ -3102,7 +3098,7 @@ def test_partial_dependence_with_nullable_types(
 )
 @pytest.mark.parametrize(
     "grid_resolution",
-    [2, 10, 50],
+    [2, 10, 60],
 )
 def test_partial_dependence_grid_values_for_numeric_data(
     logistic_regression_binary_pipeline,
@@ -3124,7 +3120,8 @@ def test_partial_dependence_grid_values_for_numeric_data(
     grid_values = pd_values["feature_values"].to_list()
 
     # Confirm fractional values are only used as grid values for floating point data types
-    if numeric_ltype in ["Double", "AgeFractional"]:
+    # when there are fewer grid values than the length of X
+    if numeric_ltype in ["Double", "AgeFractional"] and len(X) > grid_resolution:
         assert any(val % 1 for val in grid_values)
     else:
         assert not any(val % 1 for val in grid_values)
--- a/evalml/tests/model_understanding_tests/test_partial_dependence.py
+++ b/evalml/tests/model_understanding_tests/test_partial_dependence.py
@@ -3008,3 +3008,123 @@ def test_partial_dependence_dfs_transformer_does_not_calculate_feature_matrix(
     assert part_dep.feature_values.notnull().all()
     assert part_dep.partial_dependence.notnull().all()
     pd.testing.assert_frame_equal(part_dep, fast_part_dep)
+
+
+@pytest.mark.parametrize(
+    "nullable_y_ltype",
+    ["IntegerNullable", "AgeNullable", "BooleanNullable"],
+)
+@pytest.mark.parametrize(
+    "grid_resolution",
+    [2, 10, 50],
+)
+@pytest.mark.parametrize(
+    "use_int_null_many_values",
+    [True, False],
+)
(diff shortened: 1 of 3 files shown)


evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -3008,3 +3008,123 @@ def test_partial_dependence_dfs_transformer_does_not_calculate_feature_matrix(
...
+    int_col = "int col nullable"
+    if use_int_null_many_values:
+        X.ww[int_col] = ww.init_series(
+            pd.Series(range(len(y))),
+            logical_type="IntegerNullable",
+        )
+
+    pipeline = linear_regression_pipeline
+    if nullable_y_ltype == "BooleanNullable":
+        pipeline = logistic_regression_binary_pipeline
```
eccabay, line 3047: What's the reasoning behind this?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -3008,3 +3008,123 @@ def test_partial_dependence_dfs_transformer_does_not_calculate_feature_matrix(
...
+    pipeline.fit(X, y)
+    part_dep = partial_dependence(
+        pipeline,
+        X,
+        grid_resolution=grid_resolution,
+        features=int_col,
+    )
+    assert len(part_dep["partial_dependence"]) == min(
+        grid_resolution,
+        len(X[int_col].dropna().unique()),
```
eccabay, line 3058: Why do we need `dropna()` here? Are we ever expected to have nulls?

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -3008,3 +3008,123 @@ def test_partial_dependence_dfs_transformer_does_not_calculate_feature_matrix(
...
+    )
+    assert len(part_dep["feature_values"]) == min(
+        grid_resolution,
+        len(X[int_col].dropna().unique()),
+    )
```
eccabay, line 3063: Any reason we need to check the length these columns individually? We should just be able to check the length of the `part_dep` data frame

evalml/tests/model_understanding_tests/test_partial_dependence.py
```diff
@@ -3008,3 +3008,123 @@ def test_partial_dependence_dfs_transformer_does_not_calculate_feature_matrix(
...
+    "grid_resolution",
+    [2, 10, 50],
+)
+def test_partial_dependence_grid_values_for_numeric_data(
+    logistic_regression_binary_pipeline,
+    numeric_ltype,
+    grid_resolution,
+):
+    y = ww.init_series(pd.Series([True, False] * 25), logical_type="Boolean")
+    X = pd.DataFrame({"col": pd.Series(range(len(y)))})
```
eccabay, line 3113: Does this test fail on main? It looks like with a dataset length of 50, none of the parameterized grid resolution values would result in fractions.

eccabay (approved):
Looks great! Just a few small comments, the only important one is with respect to `test_partial_dependence_grid_values_for_numeric_data`

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: The age fraction and double tests pass on main, but all the remaining tests currently fail. Age and Integer fail on the check for fractional values for all of the grid resolution parameters (more on that in a sec), and the AgeNullable and IntegerNullable types fail because of the type conversion error.

Regarding your point about the grid_resolution parameterization as it relates to the length/number of unique values in X being 50 - you're right that that's not what I want, but they're all currently resulting in fractions. When the grid resolution is strictly lower than the number of unique values, we calculate our own grid values, which is where the fractional values come from.

I'm dealing with an off by one error here - I want the final number to be greater than 50. And in that case, I would expect those tests to pass on main, which I'm fine with. This test is both about recording the existing behavior and fixing broken behavior.

I'm going to change the final grid resolution number to be 60.

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: I want to be able to test partial dependence with various y logical types, so setting different pipelines here lets me test boolean nullable (the linear regression pipeline errors with `Regression pipeline can only handle numeric target data` otherwise)

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: oops this is a holdover from when I first implemented this in my nullable types branch when I was testing this with data that actually contains null values, but because of how we currently handle nullable logical types in our imputer components, these pipelines can't actually take in these nullable types with nans on main.

I can get rid of the dropna!

evalml/tests/model_understanding_tests/test_partial_dependence.py
tamargrey: Good call! This is just logic used from other tests in this file, but I agree there's no need.

christopherbunn (approved):

PR #4102, opened 2023-03-22
Fix ensembling for `generate_pipeline_example`

### LINKED ISSUE #4101: Fix ensembling for `generate_pipeline_example`

jeremyliweishih:
Fixes #4101


evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -909,23 +917,42 @@ def test_generate_pipeline_example(
...
         problem_configuration=problem_configuration,
         features=features,
+        ensembling=ensembling,
     )
-    env = AutoMLTestEnv(automl_type)
-    with env.test_context(score_return_value={aml.objective.name: 1.0}):
-        aml.search()
 
-    pipeline = aml.get_pipeline(2)
+    if ensembling:
```
jeremyliweishih, line 935: Ensembling pipelines do not work as each select columns transformer is not fit and will return no columns when using the test env. This also fails for any catboost or xgboost pipeline. Will file an issue to patch this behavior in `AutoMLTestEnv` but wanted to get this out first.

evalml/pipelines/component_graph.py
```diff
@@ -445,7 +450,10 @@ def _transform_features(
         if not fit:
             X_schema = (
                 self._return_non_engineered_features(X).ww.schema
-                if "DFS Transformer" in self.compute_order
```
tamargrey, line 448: does this mean we were getting the input types wrong for stacked ensemble pipeline with the dfs transformer present?

evalml/pipelines/pipeline_base.py
```diff
@@ -709,7 +709,7 @@ def repr_component(parameters):
         parameters_repr = ", ".join(
             [
                 f"'{component}':{{{repr_component(parameters)}}}"
-                if component != DFSTransformer.name
+                if DFSTransformer.name not in component
```
tamargrey, line 712: This is another way of checking for dfs transformers in a pipeline than the `isinstance` checks above, right? When should one be used over the other?

At some point, I feel like it might be worth rethinking how component names are used across automl search. We have multiple ways that a component's name in a component graph doesn't match the actual name of the component itself (numeric vs categorical pipelines and stacked ensemble pipelines are the two that come to mind but there may be more), and it's easy to have bugs like these. 

It'd be nice if there was a more clear way to distinguish between "I want to check something for any DFS Transformer component in a pipeline" vs "I want to check an individual DFS Transformer instance in a pipeline". Not sure what that'd look like, but maybe it's worth making an issue that we can take on at some point in the future.

tamargrey: Maybe something as simple as `element.component_graph.get_components(DFSTransformer)` that returns a list of all components of that type

jeremyliweishih: I definitely agree with you @tamargrey and thought about the same when I was writing up this PR. In this particular case, the component names are strings so I had to do this string check but I followed your example for checking for DFS below using `has_dfs = any(isinstance(c, DFSTransformer) for c in element.component_graph)`. I'll take a look into how we can standardize it.

evalml/pipelines/utils.py
```diff
@@ -676,8 +687,8 @@ def generate_pipeline_code(element, features_path=None):
     pipeline_code = "\n".join(code_strings)
     if has_dfs_and_features:
         pipeline_code = pipeline_code.replace(
-            "'DFS Transformer':{},",
-            "'DFS Transformer':{'features':features},",
+            "DFS Transformer':{},",
+            "DFS Transformer':{'features':features},",
```
christopherbunn, line 691: Was this on purpose? If not the single quote might be misaligned

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -874,9 +881,9 @@ def test_generate_pipeline_example(
     problem_configuration = (
         {
             "time_index": "date",
-            "gap": 1,
-            "max_delay": 1,
-            "forecast_horizon": 3,
+            "gap": 0,
+            "max_delay": 2,
+            "forecast_horizon": 1,
```
christopherbunn, line 886: Might be a dumb question but any reason why we changed the problem configuration to this particular set of numbers?

christopherbunn (approved):
Smaller nits but once the questions Tamar put up are addressed it LGTM

chukarsten (approved):

evalml/pipelines/utils.py
jeremyliweishih: Yes - this is to make it replace any pipelines with subpipelines (ensembling including) as the naming scheme would be "Linear Pipeline - DFS Transformer....'

jeremyliweishih: I'll write a comment about it.

evalml/pipelines/pipeline_base.py
jeremyliweishih: @tamargrey took a look and we use `has_dfs = any(isinstance(c, DFSTransformer) for c in self.component_graph)` across the board except here where I need to use the component name. Does that work for you?

tamargrey: yep, that makes sense to me! Maybe just add a comment here explaining why we have to use the component name instead of the other way

evalml/tests/pipeline_tests/test_pipeline_utils.py
```diff
@@ -909,23 +917,42 @@ def test_generate_pipeline_example(
...
+        env = AutoMLTestEnv(automl_type)
+        with env.test_context(score_return_value={aml.objective.name: 1.0}):
+            aml.search()
+
+    if ensembling:
+        for i in range(len(aml.full_rankings)):
+            pipeline = aml.get_pipeline(i)
+            if isinstance(
+                pipeline.estimator,
+                (StackedEnsembleClassifier, StackedEnsembleRegressor),
```
eccabay, line 947: For simplicity's sake, you can just check `isinstance(pipeline.estimator, StackedEnsembleBase)` here instead of checking for both!

evalml/pipelines/component_graph.py
```diff
@@ -445,7 +450,10 @@ def _transform_features(
...
+                if any(
+                    isinstance(c, DFSTransformer)
+                    for c in self.component_instances.values()
+                )
```
eccabay, line 456: My only thought here - we keep running into this and having to change the code of how we check in all sorts of different places. I don't think it needs to be changed for this PR, but I think we should consider switching this over to either a method or attribute of the component graph itself, so we can easily just check `ComponentGraph.has_dfs` from anywhere.

eccabay (approved):

evalml/tests/pipeline_tests/test_pipeline_utils.py
jeremyliweishih: good idea!

evalml/pipelines/component_graph.py
jeremyliweishih: yeah good call - I'll just add it here.

Issue #4103, opened 2023-03-22
Decomposer: IntegerNulllable incompatibility in `determine_periodicity`


tamargrey:
The Decomposer component has a nullable type incompatibility that we need to handle to support nullable types across evalml. This shows as the STLDecomposer not being added to pipelines when `y` is `IntegerNullable` type but could also have an impact on usage of `set_period`.

PR #4105, opened 2023-03-22
Handle decomposer incompatibility

tamargrey:
closes https://github.com/alteryx/evalml/issues/4103

This handles a small incompatibility in the Decomposer's `determine_periodicity` method where `float64` data containing nans when subtracted from `Int64` dtype data produces `Float64` dtype data with a `np.NaN` values that are not recognized by `dropna`, resulting in a periodicity of `None`. If we used `Float64` as a dtype in Woodwork, we'd also need to handle that here. `boolean` isn't an issue, because this is only used for regression problems.

### DIFF
--- a/evalml/tests/component_tests/decomposer_tests/test_decomposer.py
+++ b/evalml/tests/component_tests/decomposer_tests/test_decomposer.py
@@ -519,9 +519,7 @@ def test_decomposer_determine_periodicity_nullable_type_incompatibility(
         trend_degree=trend_degree,
     )
 
-    # Multiply by 1000 so we can convert to Integer, truncating the rest of the value
-    # while still mainlining trend, period, etc
-    y = y * 1000
+    # Convert to Integer, truncating the rest of the value
     y = ww.init_series(y.astype(int), logical_type=nullable_ltype)
 
     if handle_incompatibility:
@@ -533,6 +531,9 @@ def test_decomposer_determine_periodicity_nullable_type_incompatibility(
 
     subtracted_floats = y - numpy_float_data
 
+    # Pandas will not recognize the np.NaN value in a Float64 subtracted_floats
+    # and will not drop that null value, so calling _handle_nullable_types ensures
+    # that we stay in float64 and properly drop the null value
     dropped_nans = subtracted_floats.dropna()
     assert len(dropped_nans) == len(y) - 1
 
--- a/evalml/tests/component_tests/decomposer_tests/test_decomposer.py
+++ b/evalml/tests/component_tests/decomposer_tests/test_decomposer.py
@@ -526,16 +526,16 @@ def test_decomposer_determine_periodicity_nullable_type_incompatibility(
         dec = decomposer_child_class(degree=trend_degree, period=period)
         X, y = dec._handle_nullable_types(X, y)
 
-    numpy_float_data = pd.Series(range(len(y)), dtype="float64")
-    numpy_float_data.iloc[-1] = np.nan
-
-    subtracted_floats = y - numpy_float_data
+    # Introduce nans like we do in _detrend_on_fly by rolling y
+    moving_avg = 10
+    y_trend_estimate = y.rolling(moving_avg).mean().dropna()
+    subtracted_floats = y - y_trend_estimate
 
     # Pandas will not recognize the np.NaN value in a Float64 subtracted_floats
-    # and will not drop that null value, so calling _handle_nullable_types ensures
-    # that we stay in float64 and properly drop the null value
+    # and will not drop those null values, so calling _handle_nullable_types ensures
+    # that we stay in float64 and properly drop the null values
     dropped_nans = subtracted_floats.dropna()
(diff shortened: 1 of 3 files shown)


evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -462,6 +468,7 @@ def test_decomposer_determine_periodicity(
         period,
         trend_degree=trend_degree,
     )
+    y = ww.init_series(y.astype(int), logical_type=y_logical_type)
```
tamargrey, line 471: I recognize that it feels very wrong that we can just truncate y by converting to int and still have the tests work.

Originally, I was scaling up by a factor of 1000 or so to try and maintain the information necessary to achieve periodicity, but that ended up with the [tests](https://github.com/alteryx/evalml/actions/runs/4495269110/jobs/7908723125?pr=4105) where the period is set to None and the trend_degree is 1 failing. I don't really know why that would cause those tests to fail but assumed it might have something to do with larger values interacting with the `acf_threshold`? Oddly, I found that changing the scaling factor to `10` caused more tests to fail but just truncating the data allowed all of them to pass. 

This is rubbing me the wrong way, but I don't really understand what's going on in `determine_periodicity` well enough to find a smarter way around this. I'm open to any thoughts anyone may have!

jeremyliweishih: @tamargrey which logical types fail if just using the original float information?

jeremyliweishih: I think I'm OK with this - an alternative would be to run the tests with the integer logical types with integer y by changing `generate_seasonal_data` but it would serve the same purpose. Wonder what @eccabay and @chukarsten think!

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -476,6 +483,60 @@ def test_decomposer_determine_periodicity(
...
+    trend_degree = 2
+    period = 7
+    X, y = generate_seasonal_data(real_or_synthetic="synthetic")(
+        period,
+        trend_degree=trend_degree,
+    )
+
+    # Multiply by 1000 so we can convert to Integer, truncating the rest of the value
+    # while still mainlining trend, period, etc
+    y = y * 1000
```
jeremyliweishih, line 524: do we need this scaling if we don't do so above?

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -476,6 +483,60 @@ def test_decomposer_determine_periodicity(
...
+    if handle_incompatibility:
+        dec = decomposer_child_class(degree=trend_degree, period=period)
+        X, y = dec._handle_nullable_types(X, y)
+
+    numpy_float_data = pd.Series(range(len(y)), dtype="float64")
+    numpy_float_data.iloc[-1] = np.nan
+
+    subtracted_floats = y - numpy_float_data
+
+    dropped_nans = subtracted_floats.dropna()
```
jeremyliweishih, line 536: can you explain why we're tracking this here?

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
tamargrey: good call! I can remove this

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
tamargrey: 7b71388f0 - let me know if this makes sense to you!

jeremyliweishih (approved):
I think this LGTM but will leave the final call to the decomp experts!

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
eccabay: I don't think I'm quite following why this is necessary. Is this just because we get errors when trying to convert the output of `generate_seasonal_data` into the correct logical type, or is some part of the actual function failing when we have non-integer data? With the former I'm unconcerned, but the latter would worry me.

tamargrey: The first one - the output of `generate_seasonal_data` is floats, which we can't convert to `Int64` because pandas decided it will no longer quietly truncate your data for you as it does if the intended dtype is `int64`. 

I just needed a way to test `Int64` data with `determine_periodicity`, so rather than reinvent the wheel, I figured I could turn the output of `generate_seasonal_data` into something I could use. 

And to be clear: the incompatibility with `Int64` in `determine_periodicity` is due only to nullable type incompatibilities. It can handle `int64` data just fine.

eccabay: In that case, this feels fine to me. Thanks for the clarification!

PR #4116, opened 2023-03-28
Allow dropping rows for InvalidTargetDataCheck

ParthivNaresh:
Fixes: #4115


evalml/data_checks/invalid_target_data_check.py
```diff
@@ -37,16 +37,21 @@ class InvalidTargetDataCheck(DataCheck):
...
     multiclass_continuous_threshold = 0.05
 
-    def __init__(self, problem_type, objective, n_unique=100):
+    def __init__(self, problem_type, objective, n_unique=100, null_strategy="impute"):
         self.problem_type = handle_problem_types(problem_type)
         self.objective = get_objective(objective)
         if n_unique is not None and n_unique <= 0:
             raise ValueError("`n_unique` must be a non-negative integer value.")
         self.n_unique = n_unique
+        self.null_strategy = null_strategy
```
Cmancuso, line 54: do we need a check here in case an unsupported null_strategy is added?

evalml/data_checks/invalid_target_data_check.py
```diff
@@ -243,6 +248,37 @@ def _check_target_has_nan(self, y, messages):
         elif null_rows.any():
             num_null_rows = null_rows.sum()
             pct_null_rows = null_rows.mean() * 100
+            rows_to_drop = null_rows.loc[null_rows].index.tolist()
+
+            action_options = []
+            impute_action_option = DataCheckActionOption(
```
Cmancuso, line 254: could we add these into the conditionals so we only run them each if the specified strategy is input?

ParthivNaresh: Sure, initially I had planned on having a `both` option but I think we can push that out.

ParthivNaresh: Neither of these are run during initialization of the `DataCheckActionOption`, it's only when they're specified in `action_options` for a `DataCheckError` that they get converted to a dictionary

Cmancuso (approved):

jeremyliweishih (approved):
LGTM

PR #4147, opened 2023-04-13
Add cap to seasonal period to include decomposition

### LINKED ISSUE #4146: Cap seasonal period for inclusion of STL in search
The larger the seasonal period of a dataset is, the longer STL takes to train. There's some wiggle room here, as STL will take longer to fit on a random dataset compared to a perfectly seasonal dataset when given the same period, but the relationship holds in general. 

We have a few example datasets where the detected period is around 1800, but there isn't really a seasonality there at all when inspecting visually. The pipelines that include STL also perform worse on average than those without the decomposer. 

Therefore, we should impose a cap on how large the seasonal period will be in order to include the STL decomposer in pipelines.

eccabay:
Closes #4146 

Justification for this change can be found in the issue directly. I ran performance tests, there were no changes as all our test datasets have a period of <1000.

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -241,7 +241,7 @@ def _get_decomposer(X, y, problem_type, estimator_class, sampler_name=None):
                 y,
                 rel_max_order=order,
             )
-            if seasonal_period is not None:
+            if seasonal_period is not None and seasonal_period <= 1000:
                 components.append(STLDecomposer)
     return components
 
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -72,6 +72,8 @@ from evalml.utils import get_time_index, infer_feature_types
 from evalml.utils.cli_utils import get_evalml_black_config
 from evalml.utils.gen_utils import contains_all_ts_parameters
 
+DECOMPOSER_PERIOD_CAP = 1000
+
 
 def _get_label_encoder(X, y, problem_type, estimator_class, sampler_name=None):
     component = []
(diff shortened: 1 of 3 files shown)


evalml/pipelines/utils.py
```diff
@@ -241,7 +241,7 @@ def _get_decomposer(X, y, problem_type, estimator_class, sampler_name=None):
                 y,
                 rel_max_order=order,
             )
-            if seasonal_period is not None:
+            if seasonal_period is not None and seasonal_period <= 1000:
```
jeremyliweishih, line 244: in the `make_pipeline` tests can you make sure that the decomposer is not added if the seasonal period is > 1000? I think we can just add a case here by mocking `seasonal_period`.

jeremyliweishih (approved):
LGTM - lmk if you agree on the test example. I think it would be good to have coverage for this.

evalml/pipelines/utils.py
eccabay: Looking back, I realized I completely forgot about the `test_make_pipeline_controls_decomposer_time_series` - I added a case in there!

chukarsten: Ugh, my only issue with this is the hardcoded 1000 period.  Particularly the ramifications for if a user does show up with a super high frequency dataset like the one that triggered this.

eccabay: Very fair. I'll drop it over to be a constant.

christopherbunn (approved):

chukarsten (approved):

PR #4156, opened 2023-04-21
Add recommendation score

### LINKED ISSUE #4149: Implement .recommendation_score() for Pipelines

eccabay:
Closes #4149


evalml/automl/automl_search.py
```diff
@@ -529,61 +547,6 @@ def __init__(
         self.data_splitter = data_splitter
         self.optimize_thresholds = optimize_thresholds
         self.ensembling = ensembling
-        if objective == "auto":
```
eccabay, line 532: This section was not deleted, simply moved down lower to be part of the other objective handling. It needed to be lower rather than moving the new handling up here because the new recommendation_objective logic requires several other validation checks to have already happened and `self.X_train` and `self.y_train` to be set.

docs/source/user_guide/automl.ipynb
```diff
@@ -396,6 +396,59 @@
...
+    "    X_train=X_train,\n",
+    "    y_train=y_train,\n",
+    "    X_holdout=X_holdout,\n",
+    "    y_holdout=y_holdout,\n",
+    "    problem_type=\"binary\",\n",
+    "    use_recommendation=True,\n",
+    ")\n",
+    "automl_recommendation.search(interactive_plot=False)\n",
+    "\n",
+    "automl_recommendation.rankings"
```
jeremyliweishih, line 449: Spacing is off [here](https://feature-labs-inc-evalml--4156.com.readthedocs.build/en/4156/user_guide/automl.html#Recommendation-Score) for some reason
Screenshot 2023-04-25 at 11 14 53 AM

jeremyliweishih (approved):
Overall LGTM - great work! Just have some nits and you may need to take a look at the docs for the new rankings!

evalml/objectives/utils.py
```diff
@@ -214,3 +229,158 @@ def get_core_objectives(problem_type):
...
+    """
+    problem_type = handle_problem_types(problem_type)
+    objective_list = [
+        obj.name
+        for obj in DEFAULT_RECOMMENDATION_OBJECTIVES
+        if obj.is_defined_for_problem_type(problem_type)
+    ]
+
+    if problem_type == ProblemTypes.MULTICLASS and imbalanced:
+        objective_list.remove(objectives.AUCMicro.name)
```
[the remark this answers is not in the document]
eccabay: Yeah, I went back and forth on this a lot. I figured this was more concise, but it may also be more unclear. I'm happy to change this around if need be!

docs/source/user_guide/automl.ipynb
eccabay: This is a weird one. It doesn't happen when running the jupyter notebooks directly, and it looks like the same issue comes up with other rankings dataframes as well - this one looks the worst since it has the most columns. I think I'll drop some unnecessary columns so that they aren't so squished, but I'm also open to other suggestions, as Google was unhelpful here.

evalml/objectives/utils.py
```diff
@@ -214,3 +229,158 @@ def get_core_objectives(problem_type):
...
+):
+    """Computes a recommendation score for a model given scores for a group of objectives.
+
+    Args:
+        objectives (dict[str,float]): A dictionary mapping objectives to their values
+        prioritized_objective (str): An optional name of a priority objective that should be given heavier weight
+            than the other objectives contributing to the score. Defaults to None, where all objectives are
+            weighted equally.
+        prioritized_weight (float): The weight to attribute to the prioritized objective, if it exists.
+            Defaults to 0.5.
```
chukarsten, line 355: You might want to specify here that it should be between 0 and 1 and that, probably unnecessary, higher is weightier?

evalml/objectives/utils.py
```diff
@@ -214,3 +229,158 @@ def get_core_objectives(problem_type):
...
+        normalized[objective_name] = normal
+    return normalized
+
+
+def recommendation_score(
+    objectives,
+    prioritized_objective=None,
+    prioritized_weight=0.5,
+):
+    """Computes a recommendation score for a model given scores for a group of objectives.
```
chukarsten, line 347: I think it's also worth here specifying that it's a weighted average.

evalml/objectives/utils.py
```diff
@@ -214,3 +229,158 @@ def get_core_objectives(problem_type):
...
+
+    Args:
+        objectives (dict[str,float]): A dictionary mapping objectives to their values
```
chukarsten, line 350: Not sure how we want to handle this docstring...because here we're expecting the floating point values for the dict to be between 0 and 1, right?

evalml/objectives/utils.py
```diff
@@ -214,3 +229,158 @@ def get_core_objectives(problem_type):
...
+
+    Raises:
+        ValueError: If the objective to prioritize is not in the known objectives, or if the priority weight is not
+            a float between 0 and 1.
+    """
+    objectives = objectives.copy()  # Prevent mutation issues
+    priority_weight = 0
+    default_weight = 1 / len(objectives)
+    if prioritized_objective is not None:
+        if prioritized_objective not in objectives:
```
chukarsten, line 368: `prioritized_objective` allows us to give more (or I guess less) weight to a single objective.  Is it worth designing this function such that it accepts instead an `objectives_weighting (dict[str,float]): Mapping of objectives to their relative weighting to compose the recommendation score` ?

I was kind of thinking that there would existing presets that would function as data scientist selected "blends" of metrics based on what our user was looking for.  This would require weighting of multiple metrics at the same time.

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5480,3 +5490,191 @@ def test_holdout_set_results_and_rankings(caplog, AutoMLTestEnv):
...
+        X_train=X,
+        y_train=y,
+        problem_type="binary",
+        use_recommendation=True,
+        include_recommendation=["F1"],
+    )
+    assert "already one of the default objectives" in caplog.text
+
+
+def test_recommendation_include_non_optimization(X_y_binary):
```
chukarsten, line 5598: What's this test about?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5480,3 +5490,191 @@ def test_holdout_set_results_and_rankings(caplog, AutoMLTestEnv):
...
+        problem_type="binary",
+        use_recommendation=True,
+        include_recommendation=["Recall"],
+    )
+    assert "Recall" in automl.recommendation_objectives
+    assert get_objective("Recall") in automl.additional_objectives
+
+
+@pytest.mark.parametrize("imbalanced_data", [True, False])
+def test_use_recommendation_score_imbalanced(
```
chukarsten, line 5612: I reviewed the code...why are we treating imbalanced data differently?

evalml/tests/automl_tests/test_automl.py
```diff
@@ -5480,3 +5490,191 @@ def test_holdout_set_results_and_rankings(caplog, AutoMLTestEnv):
...
+        use_recommendation=True,
+    )
+    objectives = get_default_recommendation_objectives(
+        "multiclass",
+        imbalanced=imbalanced_data,
+    )
+    assert automl.recommendation_objectives == objectives
+
+
+def test_use_recommendation_score(AutoMLTestEnv, X_y_binary):
```
chukarsten, line 5639: Hahaha, I can't say I'm in love with the architecture of this test....it's basically 4 tests in one, so if it fails, you still need to go and determine which of the internal tests is failing and why.  Additionally, it covers up any successive failures i.e. internal test #2 fails the test but internal test #4 is also failing.  I would challenging you to think of a way to maybe parametrize this so that the individually parameterized tests can fail and that the parameter reveals hints at the test condition that causes the failure.

evalml/automl/automl_search.py
```diff
@@ -1632,10 +1700,94 @@ def full_rankings(self):
             "search_order",
             pd.Series(self._results["search_order"]),
         )  # place search_order after pipeline_name
-        rankings_df.sort_values("ranking_score", ascending=ascending, inplace=True)
+
+        ranking_column = "ranking_score"
+        if self.use_recommendation:
```
chukarsten, line 1705: Is there any benefit to modifying the API of `full_rankings()` to have the ranking column be selectable via the function call rather than the AutoML __init__ setting?  Probably not, I guess.  Wouldn't want the user to be able to select "recommendation_score" with `self.use_recommendation` not having been set...

evalml/objectives/utils.py
eccabay: That was my original design - amended after discussion [here](https://alteryx.atlassian.net/wiki/spaces/PS/pages/1710916950/Model+Recommendation+Score?focusedCommentId=1720058958)

evalml/tests/automl_tests/test_automl.py
eccabay: We get the additional (not main objective) scores through the `additional_objectives` mechanism. By default, this calculates the objectives for all of the optimization metrics, excluding any that are classified as ranking only (and we have a check against it). However, we want to make sure users have full control over which objectives are included - including ones that are ranking only (for example, recall). This test ensures that the logic permitting this works

evalml/tests/automl_tests/test_automl.py
eccabay: This was discussed in the design doc - it's exclusively used in the multiclass classification case, when selecting whether to use AUC Micro or AUC weighted, because micro averages are better for imbalanced datasets.

evalml/objectives/utils.py
eccabay: I added a bit more to the description - hopefully the docstring as a whole is clearer now!

evalml/objectives/utils.py
```diff
@@ -214,3 +229,187 @@ def get_core_objectives(problem_type):
...
+            than 1, and remaining weight will be evenly distributed between the remaining objectives. Should not be used
+            at the same time as prioritized_objective.
+
+    Returns:
+        A value between 0 and 100 representing how strongly we recommend a pipeline given a set of evaluated objectives
+
+    Raises:
+        ValueError: If the objective to prioritize is not in the known objectives, or if the priority weight is not
+            a float between 0 and 1.
+        TODO: add more
```
chukarsten, line 374: Sorry, missed this TODO....

evalml/objectives/utils.py
```diff
@@ -214,3 +229,187 @@ def get_core_objectives(problem_type):
...
+            normalized using the normalize_objectives function.
+        prioritized_objective (str): An optional name of a priority objective that should be given heavier weight
+            than the other objectives contributing to the score. Defaults to None, where all objectives are
+            weighted equally.
+        prioritized_weight (float): The weight (maximum of 1) to attribute to the prioritized objective, if it exists.
+            Should be between 0 and 1. Defaults to 0.5.
+        custom_weights (dict[str,float]): A dictionary mapping objective names to corresponding weights between 0 and 1.
+            If all objectives are listed, should add up to 1. If a subset of objectives are listed, should add up to less
+            than 1, and remaining weight will be evenly distributed between the remaining objectives. Should not be used
+            at the same time as prioritized_objective.
```
chukarsten, line 366: It's my fault for doing this piecemeal - I blame being distracted - but perhaps we should consider an API where these two are put together and that the input is a dict of str:float.  If there is one value, then perhaps that should be interpreted as the prioritized weight/obj.

eccabay: As we've discussed previously, I'm hesitant to completely remove the prioritized objective/weight API. I think the benefit of the simplicity outweighs the slightly larger API. That being said, after thinking about it more I would be open to removing the `prioritized_weight` argument, and just keeping `prioritized_objective`. That allows users to have an easy way to say "I care about this one more", and if they want to be more specific about the weights, there is still `custom_weights` to open up that opportunity. Thoughts?

chukarsten (approved):

PR #4164, opened 2023-04-28
Update Dask install commands in pyproject.toml

### LINKED ISSUE #4158: Update Dask install in pyproject.toml
Since Dask is a requirement for EvalML, Featuretools and Woodwork should also be installed using their `[dask]` extras install command to make sure a Dask version is installed that meets the requirements for all three libraries. Without using the `[dask]` extra install for Featuretools and Woodwork it is possible that a version of Dask could be installed that is incompatible with FT and WW as those requirements would not be accounted for during dependency resolution.

NOTE: This update should be made after Featuretools is released with Dask as an optional dependency.

thehomebrewnerd:
Fixes #4158


jeremyliweishih (approved):

dvreed77 (approved):
lgtm

eccabay (approved):

PR #4186, opened 2023-05-23
Fix small ARIMA issues

### LINKED ISSUE #4185: Harden ARIMA implementation
There are a number of small issues with our ARIMA implementation that cause and hide bugs. Namely:
- When calculating the forecast horizon, we try to calculate the distance forward but fall back to incorrect values because we end up checking `isinstance(np.int64, int)`, which returns False when it should be considered True.
- If X is None, we set `self.last_X_index` to be the entire y index, rather than the last value of y's index

These cause errors in their own right, and if they're fixed, this code block should raise a `ValueError: X array dims (n_rows) != n_periods`:
```
X = pd.DataFrame({"date": pd.date_range("1/1/21", periods=100)})
y = pd.Series([1, 4, 2] * 20 + [3, 1] * 20)

problem_configuration = {
    "max_delay": 52,
    "gap": 0,
    "forecast_horizon": 5,
    "time_index": "date",
}
X_train, X_holdout, y_train, y_holdout = split_data(
    X, y, problem_type="time series regression"
)
automl = AutoMLSearch(
    X_train,
    y_train,
)
automl.search()
arima = automl.get_pipeline(11)
arima.fit(X_train, y_train)
y_forecast = arima.get_forecast_predictions(X=X, y=y)
```

eccabay:
Closes #4185

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -3,6 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
     * Fixes
+        * Fixed small errors in ``ARIMARegressor`` implementation :pr:`4186`
     * Changes
     * Documentation Changes
     * Testing Changes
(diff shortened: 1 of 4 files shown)


jeremyliweishih (approved):
LGTM

christopherbunn (approved):
🚀

PR #4196, opened 2023-05-31
Added `excluded_model_families` parameter to `AutoMLSearch()`

### LINKED ISSUE #4197: Enhance AutoMLSearch API to Exclude Estimators
As an EvalML user, I would like AutoMLSearch() to permit me to disable specific model families on a per-problem type basis so that I can more easily specify which models are run for certain problem types.

christopherbunn:
Resolves #4197

### DIFF
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -131,10 +131,6 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         self.exclude_featurizers = exclude_featurizers
 
         if allowed_component_graphs not in (None, []):
-            if allowed_model_families not in (None, []):
-                raise ValueError(
-                    "Both `allowed_model_families` and `allowed_component_graphs` cannot be set.",
-                )
             if excluded_model_families not in (None, []):
                 raise ValueError(
                     "Both `excluded_model_families` and `allowed_component_graphs` cannot be set.",
--- a/evalml/automl/automl_algorithm/iterative_algorithm.py
+++ b/evalml/automl/automl_algorithm/iterative_algorithm.py
@@ -130,23 +130,20 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         self._set_additional_pipeline_params()
         self.exclude_featurizers = exclude_featurizers
 
-        if allowed_component_graphs not in (None, []):
-            if excluded_model_families not in (None, []):
+        if allowed_component_graphs is not None:
+            if excluded_model_families is not None:
                 raise ValueError(
                     "Both `excluded_model_families` and `allowed_component_graphs` cannot be set.",
                 )
         self.allowed_component_graphs = allowed_component_graphs
 
-        if allowed_model_families not in (None, []) and excluded_model_families not in (
-            None,
-            [],
-        ):
+        if allowed_model_families is not None and excluded_model_families is not None:
             raise ValueError(
                 "Both `allowed_model_families` and `excluded_model_families` cannot be set.",
             )
 
         self.allowed_model_families = allowed_model_families
-        self.excluded_model_families = excluded_model_families or []
+        self.excluded_model_families = excluded_model_families
 
         super().__init__(
             allowed_pipelines=self.allowed_pipelines,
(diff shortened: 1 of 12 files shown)


machineFL:

eccabay:
@christopherbunn what are your thoughts on removing `allowed_model_families` as part of this work? 

If we don't, it may be possible to introduce some weird conflicts using the iterative algorithm where a model family is simultaneously allowed and disallowed. It's been a weird, semi-deprecated argument for a while anyways, since it's only usable in the iterative algorithm case but we still allow users to set it in the default algorithm case. Open to discussion, of course.

evalml/automl/automl_search.py
```diff
@@ -841,6 +844,18 @@ def _is_imbalanced(X, y, problem_type):
                 )
         self.exclude_featurizers = exclude_featurizers or []
 
+        if exclude_model_families:
```
eccabay, line 847: On the same line of thought as the potential issues around `allowed_model_families`, I think we need some handling here for `allowed_component_graphs` - probably that they shouldn't be able to be set at the same time

evalml/pipelines/components/utils.py
```diff
@@ -56,14 +56,15 @@ def allowed_model_families(problem_type):
...
-def get_estimators(problem_type, model_families=None):
+def get_estimators(problem_type, model_families=None, exclude_model_families=[]):
     """Returns the estimators allowed for a particular problem type.
 
     Can also optionally filter by a list of model types.
 
     Args:
         problem_type (ProblemTypes or str): Problem type to filter for.
         model_families (list[ModelFamily] or list[str]): Model families to filter for.
+        exclude_model_families (list[ModelFamily]): A list of model families to exclude from the results.
```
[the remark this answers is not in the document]
christopherbunn: Makes sense to me, will update.

christopherbunn:
@eccabay re:`allowed_model_families`, I think we can keep it and add logic to make sure that both of these parameters are not set. I can see cases where I would want to have this set, even if we don't currently use it anywhere else.

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -68,6 +68,7 @@ class IterativeAlgorithm(AutoMLAlgorithm):
         verbose (boolean): Whether or not to display logging information regarding pipeline building. Defaults to False.
         exclude_featurizers (list[str]): A list of featurizer components to exclude from the pipelines built by IterativeAlgorithm.
             Valid options are "DatetimeFeaturizer", "EmailFeaturizer", "URLFeaturizer", "NaturalLanguageFeaturizer", "TimeSeriesFeaturizer"
+        excluded_model_families (list[ModelFamily]): A list of model families to exclude from the estimators used when building pipelines.
```
eccabay, line 71: Unrelated request - could you add allowed_model_families to this docstring? Unless it's here and I'm just totally not seeing it

evalml/automl/automl_algorithm/iterative_algorithm.py
```diff
@@ -126,10 +127,27 @@ def __init__(
         self.allowed_pipelines = []
 
         self.features = features
-        self.allowed_component_graphs = allowed_component_graphs
         self._set_additional_pipeline_params()
         self.exclude_featurizers = exclude_featurizers
 
+        if allowed_component_graphs not in (None, []):
+            if excluded_model_families not in (None, []):
```
eccabay, line 134: I'm curious about your decision here to include [] as a possible argument, it's not something we do in other places afaik. It feels a bit clunky, I would certainly prefer to keep `None` as the only no-op option until absolutely necessary if possible.

eccabay: There's also some inconsistency with this in `automl_search.py` w.r.t line 603 and 618, we should clean that up to make it consistent as well!

eccabay (approved):
LGTM, just have some nits about our handling of the "not set" case

evalml/automl/automl_algorithm/iterative_algorithm.py
christopherbunn: It's there, just out of order. I'll update.

jeremyliweishih (approved):
LGTM just wondering if we should centralize the validation of allowing and excluding model families. LMK what you think!

PR #4213, opened 2023-06-26
Update API reference

### LINKED ISSUE #4109: Get API reference up to date
As a user, I would like to rely on our [API Reference](https://evalml.alteryx.com/en/stable/api_index.html) page for quick reference to object and function syntax. However, it is very out of date - missing objects such as our decomposers (currently only has the `PolynomialDecomposer`) and Prophet models. 

Nice to have: an automated CI check ensuring that we keep the API up to date when changes are made.

eccabay:
Closes #4109


jeremyliweishih (approved):
Thank you for this! This LGTM but just need to remove the cloud pickle files from running the notebooks locally.

christopherbunn (approved):

Issue #4138, opened 2023-04-11
Unpin sktime


eccabay:
We pinned sktime to version 0.17.0 in #4137 due to [failures](https://github.com/alteryx/evalml/pull/4135) within pmdarima. This issue tracks unpinning once those issues are resolved.

PR #4214, opened 2023-06-27
Unpinned sktime version

remyogasawara:
Resolves: #4138

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -25,7 +25,7 @@ outputs:
         - setuptools ==58.0.4
       run:
         - numpy >=1.21.0
-        - pandas >=1.5.0
+        - pandas >=1.5.0, <2.0.0
         - dask >=2022.2.0, !=2022.10.1
         - scipy >=1.5.0
         - scikit-learn >=1.2.2
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -79,7 +79,7 @@ outputs:
         - lime >=0.2.0.1
         - python >=3.8.*
         - imbalanced-learn >=0.9.1, <0.11.0
-        - sktime ==0.17.0
+        - sktime >=0.17.0
         - pmdarima >=1.8.5
         - vowpalwabbit >=8.11.0
     test:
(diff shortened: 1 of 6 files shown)


machineFL:

evalml/tests/component_tests/test_arima_regressor.py
```diff
@@ -203,7 +203,6 @@ def test_feature_importance(ts_data):
     [
         (True, False, False, False, False, False),
         (False, True, True, False, False, True),
-        (False, True, True, False, False, False),
```
remyogasawara, line 206: We deleted this test case because sktime depreciated integer indexes

christopherbunn:
Might need to be addressed in either this PR or another one but I think we also need to cap the pandas version for conda for now. [Here's ](https://github.com/alteryx/evalml/actions/runs/5401522931/jobs/9811451707#step:5:3126) a passing build with 1.5.3 while [this build](https://github.com/alteryx/evalml/actions/runs/5404484826/jobs/9818701647?pr=4214#step:5:3126) fails with 2.0.2.

christopherbunn:
@jeremyliweishih I agree that we should bump it up, but I think we should it in a follow up issue. It would keep the pandas 2.0 work separate from this PR. Should we open another PR for the pandas fixes only, merge that in, then rebase and get this in?

christopherbunn (approved):

.github/meta.yaml
```diff
@@ -25,7 +25,7 @@ outputs:
         - setuptools ==58.0.4
       run:
         - numpy >=1.21.0
-        - pandas >=1.5.0
+        - pandas >=1.5.0, <2.0.0
```
jeremyliweishih, line 28: Can you file an issue to unpin if we don't already have one?

jeremyliweishih (approved):

.github/meta.yaml
jeremyliweishih: Do mention that we were not supporting it before as pip would not install > 2.0

PR #4221, opened 2023-06-30
Added MASE metric and `y_train` parameter to objectives

### LINKED ISSUE #4217: Time series regression: add MASE/MAPE/SMAPE as metrics
MASE is a metric for time series regression which EvalML supports. MASE and SMAPE are two other commonly used metrics for time series regression.

remyogasawara:
Resolves #4217


evalml/tests/dependency_update_check/latest_dependency_versions.txt
```diff
@@ -14,7 +14,7 @@ ipywidgets==8.0.4
 kaleido==0.2.1
 lightgbm==3.3.5
 lime==0.2.0.1
-matplotlib==3.7.1
+matplotlib==3.7.2
```
remyogasawara, line 17: matplotlib just got updated to 3.7.2 so I changed that here but wasn't sure if I should be merging that to the main branch as well

evalml/objectives/standard_metrics.py
```diff
@@ -813,6 +816,47 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+    name = "Mean Absolute Scaled Error"
+    greater_is_better = False
+    score_needs_proba = False
+    perfect_score = 0.0
+    is_bounded_like_percentage = False  # Range [0, Inf)
+    expected_range = [0, float("inf")]
+
+    def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
+        """Objective function for mean absolute percentage error for time series regression."""
+        y_train = y_true
```
remyogasawara, line 839: MASE takes in training data as a parameter, since we don't use training data we just used the actual data in place of training data for now, but I'm not sure if that's what we should be doing

eccabay: This brings up a very interesting question. A quick test shows that the results do change based on what you pass in to the `y_train` argument, meaning that the results we pass in with this are going to be less correct.

Looking through where we score pipelines, I don't think it would be too much extra lift to add the ability to pass `y_train` through. We have access to it in the upper level time series pipeline scoring functions, as it's pretty common for time series related things in general to require the historical data.

I'm curious what others think though - should we update our objective functions to have access to `y_train`, or should we keep it as is like this?

evalml/objectives/standard_metrics.py
```diff
@@ -813,6 +816,47 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+                "all training targets contain the value 0.",
+            )
+        if isinstance(y_true, pd.Series):
+            y_true = y_true.to_numpy()
+        if isinstance(y_predicted, pd.Series):
+            y_predicted = y_predicted.to_numpy()
+        if isinstance(y_train, pd.Series):
+            y_train = y_train.to_numpy()
+        mase = MeanAbsoluteScaledError()
+        return mase(y_true, y_predicted, y_train=y_train) * 100
```
eccabay, line 852: Why are we multiplying by 100 here?

evalml/objectives/standard_metrics.py
```diff
@@ -813,6 +816,47 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+            raise ValueError(
+                "Mean Absolute Scaled Error cannot be used when "
+                "all training targets contain the value 0.",
+            )
+        if isinstance(y_true, pd.Series):
+            y_true = y_true.to_numpy()
+        if isinstance(y_predicted, pd.Series):
+            y_predicted = y_predicted.to_numpy()
+        if isinstance(y_train, pd.Series):
+            y_train = y_train.to_numpy()
```
eccabay, line 850: If we do end up keeping `y_train == y_true`, we could consolidate this section a bit by holding off on calling `y_train = y_true` until after we've converted y_true to numpy. However, there is also no actual requirement from sktime that the Series need to be converted to numpy. According to the documentation, passing a `pd.Series` into this function works just fine.

evalml/objectives/standard_metrics.py
eccabay, line 852: According to the documentation, another parameter that we may want to consider leveraging is `sp`. With our `Decomposer.determine_periodicity` function (or, if the pipeline has a Decomposer, the decomposer's saved sp parameter), we should have the information necessary to pass that through. The question remains how, since we need information from X and y train, but it could be done. From brief testing, it looks like adding the sp argument gives us lower values (aka "better" scores), and they would once again be considered to be more accurate, since we use sp during training.

I'm curious what other people think about the need to include sp as well.

eccabay (changes requested):
This metric is a tricky one - I think there's a few details we need to discuss before it can be merged in! Curious for other team members' input on this as well.

evalml/objectives/standard_metrics.py
remyogasawara: You're right it shouldn't be multiplied, thanks for catching that

evalml/objectives/standard_metrics.py
```diff
@@ -813,6 +816,47 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+        mase = MeanAbsoluteScaledError()
+        return mase(y_true, y_predicted, y_train=y_train) * 100
+
+    @classproperty
+    def positive_only(self):
+        """If True, this objective is only valid for positive data."""
+        return True
```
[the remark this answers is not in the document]
remyogasawara: Would that be applied to MAPE and SMAPE too? They both use the same function but also take in negative inputs

evalml/objectives/standard_metrics.py
jeremyliweishih: I think we should update it to support MASE properly but lets work on it in a followup issue. @remyogasawara can you file the follow up in EvalML? Just basically stating the requirements that @eccabay wrote above?

evalml/objectives/standard_metrics.py
```diff
@@ -813,6 +816,47 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
         )
 
 
+class MASE(TimeSeriesRegressionObjective):
+    """Mean absolute percentage error for time series regression. Scaled by 100 to return a percentage.
+
+    Only valid for nonzero inputs. Otherwise, will throw a ValueError.
```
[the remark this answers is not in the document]
jeremyliweishih: Ok I see what you mean - can you update to mention if all the input are zero?

evalml/objectives/standard_metrics.py
jeremyliweishih: If there is no requirement can we remove this section?

jeremyliweishih (changes requested):

evalml/objectives/standard_metrics.py
jeremyliweishih: I agree we should support this! @remyogasawara can you file another issue for this so we can prioritize after decomp is added for multiseries?

evalml/objectives/standard_metrics.py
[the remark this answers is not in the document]
eccabay: Yep, after taking a closer look, we shouldn't have `positive_only` set to True for any of these. Can you remove it for MAPE, SMAPE, and MASE?

eccabay:
One other small thing I forgot to request, can you add this (as well as SMAPE, because I forgot before 😅) to the [API reference?](https://github.com/alteryx/evalml/blob/main/docs/source/api_index.rst)

remyogasawara:
Resolves #4224

evalml/objectives/standard_metrics.py
remyogasawara: #4224

jeremyliweishih (approved):
LGTM good work!

evalml/objectives/cost_benefit_matrix.py
```diff
@@ -35,12 +35,20 @@ def __init__(self, true_positive, true_negative, false_positive, false_negative)
...
+        y_train=None,
+        X=None,
+        sample_weight=None,
+    ):
         """Calculates cost-benefit of the using the predicted and true values.
 
         Args:
             y_predicted (pd.Series): Predicted labels.
             y_true (pd.Series): True labels.
+            y_train (pd.Series): Training labels.
```
eccabay, line 51: For this and other metrics where `y_train` isn't used, can you mention that this is ignored?

evalml/tests/objective_tests/test_standard_metrics.py
```diff
@@ -160,15 +161,17 @@ def test_negative_with_log():
 
 
 @pytest.mark.parametrize("objective_class", _all_objectives_dict().values())
+# @pytest.mark.parametrize("obj", [obj for obj in _all_objectives_dict().values()])
```
eccabay, line 164: We don't need this comment any more!

evalml/objectives/standard_metrics.py
```diff
@@ -827,6 +1040,48 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
         )
 
 
+class MASE(TimeSeriesRegressionObjective):
+    """Mean absolute scaled error for time series regression. Scaled by 100 to return a percentage.
```
eccabay, line 1044: Nit: this is no longer scaled by 100!

evalml/objectives/standard_metrics.py
```diff
@@ -827,6 +1040,48 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+        self,
+        y_true,
+        y_predicted,
+        y_train,
+        X=None,
+        sample_weight=None,
+    ):
+        """Objective function for mean absolute scaled error for time series regression."""
+        # [0, 1, 1, 0] -- good
+        # [0, 0, 0, 0] -- bad
```
eccabay, line 1072: We don't need these comments any more either

evalml/objectives/standard_metrics.py
```diff
@@ -827,6 +1040,48 @@ def objective_function(self, y_true, y_predicted, X=None, sample_weight=None):
...
+
+        if (isinstance(y_train, pd.DataFrame) and (y_train.values == 0).all()) or (
+            isinstance(y_train, pd.Series) and (y_train == 0).all()
+        ):
```
eccabay, line 1076: I don't think you need to split it up this way - `y_train.values` should work regardless of whether it's a dataframe or a series!

eccabay (approved):
Looks good, great job being thorough with your changes! Just a few small nitpicks, but otherwise this looks ready to go

PR #4233, opened 2023-07-12
Update regression metrics to handle multiseries problems

### LINKED ISSUE #4231: Ensure that time series regression metrics support MS

eccabay:
Closes #4231 

Our anticipated input from users for multiseries problems has one column that contains all the series targets and one column that delineates the series_id of each row. However, our estimators expect one column per series, so we will be internally converting between these two forms.

Both scikit-learn and sktime already have implementations for multioutput problems expecting the "one column per series" data format, so this PR assumes that input format as well, and just adapts the few metrics that were not already set up to handle this.


jeremyliweishih (approved):
LGTM good work!

christopherbunn (approved):
LGTM

PR #4246, opened 2023-07-20
Add baseline multiseries regressor

### LINKED ISSUE #4241: Baseline Multiseries Time Series Regressor

eccabay:
Closes #4241

Implementation assumes one column per target series, which is not the final state of the input. Changes will probably have to be made once stacking/unstacking functions are implemented and this is integrated into search.

Implementation also assumes integer indices.

### DIFF
--- a/evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
@@ -72,7 +72,7 @@ class MultiseriesTimeSeriesBaselineRegressor(Estimator):
         delay_columns = pd.DataFrame(
             np.zeros((self.start_delay, y.shape[1])),
             columns=self._series_names,
-            index=range(y.index[-1], self.start_delay + y.index[-1]),
+            index=range(y.index[-1] + 1, self.start_delay + y.index[-1] + 1),
         )
         y = pd.concat([y, delay_columns])
 
--- a/evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
+++ b/evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
@@ -1,0 +1,122 @@
+"""Time series estimator that predicts using the naive forecasting approach."""
+import numpy as np
+import pandas as pd
+
+from evalml.model_family import ModelFamily
+from evalml.pipelines.components.estimators import Estimator
+from evalml.problem_types import ProblemTypes
+from evalml.utils import infer_feature_types
+
+
+class MultiseriesTimeSeriesBaselineRegressor(Estimator):
+    """Multiseries time series regressor that predicts using the naive forecasting approach.
+
+    This is useful as a simple baseline estimator for multiseries time series problems.
+
+    Args:
+        gap (int): Gap between prediction date and target date and must be a positive integer. If gap is 0, target date will be shifted ahead by 1 time period. Defaults to 1.
+        forecast_horizon (int): Number of time steps the model is expected to predict.
+        random_seed (int): Seed for the random number generator. Defaults to 0.
+    """
+
+    name = "Time Series Baseline Regressor"
+    hyperparameter_ranges = {}
+    """{}"""
+    model_family = ModelFamily.BASELINE
+    """ModelFamily.BASELINE"""
+    is_multiseries = True
+    supported_problem_types = [
+        ProblemTypes.TIME_SERIES_REGRESSION,
+    ]
+    """[
+        ProblemTypes.TIME_SERIES_REGRESSION,
+    ]"""
+
+    def __init__(self, gap=1, forecast_horizon=1, random_seed=0, **kwargs):
+        self._prediction_value = None
+        self.start_delay = forecast_horizon + gap
+        self._num_features = None
+
+        if gap < 0:
+            raise ValueError(
+                f"gap value must be a positive integer. {gap} was provided.",
+            )
+
+        parameters = {"gap": gap, "forecast_horizon": forecast_horizon}
+        parameters.update(kwargs)
+        super().__init__(
+            parameters=parameters,
+            component_obj=None,
+            random_seed=random_seed,
+        )
+
+    def fit(self, X, y=None):
+        """Fits multiseries time series baseline regressor to data.
+
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features].
+            y (pd.Series): The target training data of length [n_samples].
+
+        Returns:
+            self
+
+        Raises:
+            ValueError: If input y is None.
+        """
+        if y is None:
+            raise ValueError(
+                "Cannot train Multiseries Time Series Baseline Regressor if y is None",
+            )
+        self._series_names = y.columns
+
+        delay_columns = pd.DataFrame(
+            np.zeros((self.start_delay, y.shape[1])),
+            columns=self._series_names,
+            index=range(y.index[-1], self.start_delay + y.index[-1]),
+        )
+        y = pd.concat([y, delay_columns])
+
+        self._delayed_target = y.shift(self.start_delay, fill_value=0)
+
+        return self
+
+    def predict(self, X):
+        """Make predictions using fitted multiseries time series baseline regressor.
+
+        Args:
+            X (pd.DataFrame): Data of shape [n_samples, n_features].
+
+        Returns:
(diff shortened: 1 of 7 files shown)


evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -234,33 +253,28 @@ def _compute_delays(self, X_ww, y):
...
+                self._delay_df(X_ww, cols_to_delay, categorical_columns, X_categorical),
+            )
         # Handle cases where the target was passed in
         if self.delay_target and y is not None:
-            if type(y.ww.logical_type) == logical_types.Categorical:
-                y = self._encode_y_while_preserving_index(y)
-            for t in self.statistically_significant_lags:
-                lagged_features[self.target_colname_prefix.format(t)] = y.shift(t)
+            if isinstance(y, pd.DataFrame):
+                lagged_features.update(self._delay_df(y, y.columns))
```
jeremyliweishih, line 267: thought: should we just run `self._encode_y_while_preserving_index(y)` even though we won't expect categorical columns just yet?

jeremyliweishih (approved):
Good work 😸

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
eccabay: 🤔 interesting point, will we ever expect categorical columns? We're only supporting regression problems for multiseries

jeremyliweishih: potentially sometime in the distant future! Just thought it would make it one step easier for whoever implements that 😄  it'll be a no-op anyways right now

eccabay: I like the idea in theory, but I'm worried about increasing runtime with checking if `y` contains any categorical columns. We'd have to do so in all cases, which feels wasteful when we know we won't be dealing with it.

evalml/tests/conftest.py
```diff
@@ -830,6 +830,22 @@ def X_y_regression():
     return X, y
 
 
+@pytest.fixture
+def X_y_multiseries_regression():
```
christopherbunn, line 834: I think I'm going to utilize this for my VARMAX testing. In that case, does it make sense to have it align more with the inputs we expect for a mutliseries regressor (e.g. a `series_id` column and the column names having `series_id` value suffixes)

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -124,12 +124,15 @@ def fit(self, X, y=None):
...
         if self.time_index is None:
             raise ValueError("time_index cannot be None!")
-        self.statistically_significant_lags = self._find_significant_lags(
-            y,
-            conf_level=self.conf_level,
-            start_delay=self.start_delay,
-            max_delay=self.max_delay,
-        )
+        if isinstance(y, pd.DataFrame):
+            self.statistically_significant_lags = [self.start_delay]
```
christopherbunn, line 128: So for the multiseries case, do we not try to find the significant lags? Does this just use all or none of the lags?

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
```diff
@@ -0,0 +1,112 @@
...
+        return self
+
+    def predict(self, X):
+        """Make predictions using fitted multiseries time series baseline regressor.
+
+        Args:
+            X (pd.DataFrame): Data of shape [n_samples, n_features].
+
+        Returns:
+            pd.Series: Predicted values.
```
christopherbunn, line 86: I'm pretty sure this is correct, but just to double check we're returning the predictions with the predicted values stacked right (i.e. in series form and not as a dataframe)

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
```diff
@@ -0,0 +1,112 @@
...
+            component_obj=None,
+            random_seed=random_seed,
+        )
+
+    def fit(self, X, y=None):
+        """Fits multiseries time series baseline regressor to data.
+
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features].
+            y (pd.Series): The target training data of length [n_samples].
```
christopherbunn, line 58: I think this needs to be updated to take in the data unstacked and `y` in a dataframe right?

christopherbunn (changes requested):
A few questions for my clarification + a suggestion but once answered LGTM

evalml/tests/conftest.py
eccabay: Good callout that this should more closely match the actual expected input - I wrote these before I wrote the stacking and unstacking functions and never updated it 😅 - same with why these tests don't use `multiseries_ts_data_unstacked`. I'll refactor the test fixtures a bit to be consolidated and up to date.

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
eccabay: Lol, this is a copypasta fail. We're returning the unstacked dataframe here

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
```diff
@@ -0,0 +1,112 @@
...
+        """Fits multiseries time series baseline regressor to data.
+
+        Args:
+            X (pd.DataFrame): The input training data of shape [n_samples, n_features * n_series].
+            y (pd.DataFrame): The target training data of shape [n_samples, n_features * n_series].
+
+        Returns:
+            self
+
+        Raises:
```
chukarsten, line 63: nit: update for the second ValueError

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
```diff
@@ -0,0 +1,112 @@
...
+    def feature_importance(self):
+        """Returns importance associated with each feature.
+
+        Since baseline estimators do not use input features to calculate predictions, returns an array of zeroes.
+
+        Returns:
+            np.ndarray (float): An array of zeroes.
+        """
+        importance = np.array([0] * self._num_features)
+        return importance
```
chukarsten, line 112: Probably another nit, but if you're calling out all Baseline Estimators...is it worth putting together a story to add a BaselineEstimator class to the inheritance chain and have them all inherit this prop def?

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
chukarsten: Might want to add an explicit comment here for the case we're splitting on...e.g. if y is a dataframe, we expect it to be multiseries.

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -215,6 +218,22 @@ def _compute_rolling_transforms(self, X, y, original_features):
...
+        categorical_columns=None,
+        X_categorical=None,
+    ):
+        lagged_features = {}
+        for col_name in cols_to_delay:
+            col = data[col_name]
+            if categorical_columns and col_name in categorical_columns:
+                col = X_categorical[col_name]
+            for t in self.statistically_significant_lags:
+                lagged_features[f"{col_name}_delay_{t}"] = col.shift(t)
```
chukarsten, line 234: We're not going to be doing any external matching on this name format, right?  If so, I think we might want to establish a pattern of making this string format like a module level thing or accessible via the class

chukarsten (approved):
Just a few nits, great work

evalml/pipelines/components/estimators/regressors/multiseries_time_series_baseline_regressor.py
eccabay: Filed #4255

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
eccabay: Good call - adjusting!

christopherbunn (approved):

PR #4248, opened 2023-07-20
Pin scikit-learn

### LINKED ISSUE #4247: Resolve scikit-learn 1.3.0 upgrade
Lots of tests failing, we either need to pin or update. Looks like some deprecations that we missed have now been removed, at the very least.

eccabay:
Closes #4247 

The newest version of scikit-learn is incomatible with the version of imbalanced-learn that we're pinned to. Unfortunately, imbalanced-learn must remain pinned due to lack of nullable type handling, so we can't upgrade scikit-learn. Once imbalanced-learn fixes their nullable type issues, we should be able to unpin both.

### DIFF
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -28,7 +28,7 @@ outputs:
         - pandas >=1.5.0, <2.0.0
         - dask >=2022.2.0, !=2022.10.1
         - scipy >=1.5.0
-        - scikit-learn >=1.2.2
+        - scikit-learn >=1.2.2, <1.3.0
         - scikit-optimize >=0.9.0
         - statsmodels >=0.12.2
         - colorama >=0.4.4
--- a/core-requirements.txt
+++ b/core-requirements.txt
@@ -1,7 +1,7 @@
 numpy>=1.21.0
 pandas>=1.5.0, <2.0.0
 scipy>=1.5.0
-scikit-learn>=1.2.1
+scikit-learn>=1.2.1, <1.3.0
 scikit-optimize>=0.9.0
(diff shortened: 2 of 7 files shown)


docs/source/release_notes.rst
```diff
@@ -7,6 +7,7 @@ Release Notes
     * Changes
         * Unpinned sktime version :pr:`4214`
         * Bumped minimum lightgbm version to 4.0.0 for nullable type handling :pr:`4237`
+        * Pinned scikit-learn version due to incompatibility with pinned imbalanced-learn :pr:`4248`
```
jeremyliweishih, line 10: Do we have an issue to track unpinning imbalanced learn?

jeremyliweishih (approved):
Thanks!

docs/source/release_notes.rst
eccabay: Not yet - will file!

eccabay:
In further investigation of the imbalanced-learn error that caused its pinning, I found a workaround since the error is just in one of our tests. 

Pausing review on this PR until I can determine if this workaround allows me to unpin imbalanced-learn as well as upgrade scikit-learn instead of pinning it.

eccabay:
Looks like this is blocked until https://github.com/conda-forge/sktime-feedstock/pull/94 is merged

fkiraly:
> Looks like this is blocked until [conda-forge/sktime-feedstock#94](https://github.com/conda-forge/sktime-feedstock/pull/94) is merged

No longer blocked :-)

chukarsten (approved):
Thanks for doing this!

PR #4253, opened 2023-07-26
Extend STLDecomposer to Support Multiseries

### LINKED ISSUE #4244: Extend STLDecomposer to support Multiseries
Update the STLDecomposer so if passed in a `series_id` column, will decompose all series in the given dataset, by id

remyogasawara:
Resolves #4244


evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -58,14 +61,19 @@ def __init__(
             seasonal_smoother += 1
 
         self.forecast_summary = None
+        parameters = {
+            "degree": degree,
+            "period": period,
+            "seasonal_smoother": seasonal_smoother,
+            "time_index": time_index,
+            "series_id": series_id,
```
remyogasawara, line 69: If we're only passing in unstacked data for multiseries I don't think we need to use the series_id column, but I left it there for now since I wasn't sure

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -158,43 +166,67 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None) -> STLDecomposer:
...
 
         return self
 
     def transform(
         self,
         X: pd.DataFrame,
-        y: pd.Series = None,
-    ) -> tuple[pd.DataFrame, pd.Series]:
+        y: pd.DataFrame = None,
+    ):
```
eccabay, line 229: y can be a Series or a DataFrame, and it looks like the return types got deleted here - they're still useful!

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -158,43 +166,67 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None) -> STLDecomposer:
...
+
+        # Iterate through each id group
+        self.decompositions = {}
+        for id in y.columns:
+            series_y = y[id]
+            self.original_index = series_y.index if series_y is not None else None
+
+            X, series_y = self._check_target(X, series_y)
+
+            self._map_dt_to_integer(self.original_index, series_y.index)
```
eccabay, line 192: Having this index handling in the loop feels redundant - the index will be identical across all series, and performing this computation will add extra runtime and be confusing, giving the impression it's ever going to change (and it keeps getting overwritten anyways). Could you move this outside of the loop?

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -385,50 +454,149 @@ def _decompose_target(X, y, fh):
...
-            prediction_interval_result[f"{coverage[i]}_lower"] = intervals["lower"]
-            prediction_interval_result[f"{coverage[i]}_upper"] = intervals["upper"]
-
-        return prediction_interval_result
+        series_results = {}
+        for id in y.columns:
+            y_series = y[id]
+
+            if coverage is None:
+                coverage = [0.95]
```
eccabay, line 506: This is redundant to put inside the loop, can you pull it out?

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -385,50 +454,149 @@ def _decompose_target(X, y, fh):
...
+                prediction_interval_result[f"{coverage[i]}_upper"] = intervals["upper"]
+            series_results[id] = prediction_interval_result
+
+            # only return the dictionary if single series
+            if len(y.columns) <= 1:
+                return prediction_interval_result
+        return series_results
+
+    # Overload the plot_decomposition fucntion to be able to plot multiple decompositions for multiseries
+    def plot_decomposition(
```
eccabay, line 541: I don't think we need to overload the parent function here - most of this code is identical to `Decomposer.plot_decomposition`. We can update the parent class to handle this case instead, there's no need to copy code in order to keep all the changes to this one file!

I think this also ties in with changing the use of class attributes. The goal here is to update the STLDecomposer to support multiseries, because that's the decomposer implementation that we have that works, but ideally it won't be the only one forever. I think we shouldn't be afraid of updating the parent Decomposer code as well, so that if we add a new decomposer (or get the Polynomial Decomposer working, lol) it will be easier to have them also support multiseries. No need to shoehorn ourselves in here.

eccabay (changes requested):
My review isn't complete, but I wanted to get this out there sooner rather than later - I have some questions about the implementation that I'd love to get more eyes and opinions on!

evalml/tests/conftest.py
```diff
@@ -2529,6 +2529,75 @@ def _return_proper_func(real_or_synthetic):
...
+        seasonal_scale=1,
+        trend_degree=1,
+        freq_str="D",
+        set_time_index=False,
+    ):
+        """Function to generate a sinusoidal signal with a polynomial trend.
+
+        Args:
+            period: The length, in units, of the seasonal signal.
+            step:
```
MichaelFu512, line 2550: Is `step` needed for this function? If so then make sure to fill out what it does here.

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -203,64 +235,104 @@ def transform(
...
+        for id in y.columns:
+            series_y = y[id]
+
+            if len(y.columns) > 1:
+                seasonality = self.seasonalities[id]
+                trend = self.trends[id]
+                residual = self.residuals[id]
+                period = self.periods[id]
+            else:
+                seasonality = list(self.seasonalities.values())[0]
```
remyogasawara, line 265: Had to add this here since I removed the class attributes and replaced them with dictionaries. As a result, in the pipeline i.e. for `test_pipeline_utils` there was a key error when transforming new data when we don't fit before transforming since the keys don't match up in the dictionary. I'm not sure if there's a better way to be doing this or if we need a way to be able to transform on different data in the multiseries case.

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
MichaelFu512: I think keeping both should be fine.

evalml/tests/conftest.py
remyogasawara: Yeah looks like it was just left over from the `generate_seasonal_data` function that I copied but I'll take it out here

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
remyogasawara: I actually took out all the code supporting the ungrouping by series id, since it is the same iteration process as the unstacked data just changing what is being iterated through (i.e. columns instead of groups) but it would probably be really easy to just call the `unstack_multiseries` function if series id is not None, so let me know if you guys think that would be a good thing to add in

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -158,43 +170,63 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None) -> STLDecomposer:
...
                 f"STLDecomposer may perform poorly on data with a high seasonal smoother ({self.seasonal_smoother}).",
             )
+        X, y = self._check_target(X, y)
+
+        if isinstance(y, pd.Series):
+            y = y.to_frame()
+
+        self.original_index = y.index if y is not None else None
+
+        X, y = self._check_target(X, y)
```
christopherbunn, line 185: I think we're running `self._check_target()` twice?

christopherbunn (changes requested):
Some smaller nits. Only reviewed implemetation so far but will review tests later.

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
```diff
@@ -219,38 +334,57 @@ def test_stl_decomposer_inverse_transform(
...
+            freq_str="D",
+            set_time_index=True,
+        )
+        subset_y = y[: 5 * period]
+    elif variateness == "multivariate":
+        X, y = generate_multiseries_seasonal_data(real_or_synthetic="synthetic")(
+            period=period,
+            freq_str="D",
+            set_time_index=True,
+        )
```
eccabay, line 356: Alright, this code block has shown up so many times, I think it needs its own function in `conftest` - something where we just pass in whether the problem is univariate or multivariate, and it returns the correct data for us.

eccabay (commented):
Making awesome progress! I left a few more comments, mostly just condensing code

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
remyogasawara: It was failing in`evalml/tests/pipeline_tests/test_pipeline_utils.py::test_generate_pipeline_example`. I think from what I saw it was fitting and transforming, and then it was transforming again but was returning a key error since the id wasn't the same from the initial fitting

evalml/pipelines/components/transformers/preprocessing/decomposer.py
```diff
@@ -336,24 +337,52 @@ def plot_decomposition(
...
+            axs[1].plot(results[0]["trend"], "b")
+            axs[1].set_title("trend")
+            axs[2].plot(results[0]["seasonality"], "g")
+            axs[2].set_title("seasonality")
+            axs[3].plot(results[0]["residual"], "y")
+            axs[3].set_title("residual")
+
+            # If multiseries, return a dictionary of tuples
+            if len(y.columns) > 1:
+                fig.suptitle("Decomposition for Series {}".format(id))
```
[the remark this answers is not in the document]
remyogasawara: Its just a matplotlib function that gives a title to each series. Apparently its `sup` not `sub` just because the title is at the top of the graph instead of the bottom

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -158,43 +170,62 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None) -> STLDecomposer:
...
+        self.seasonals = {}
+        self.periods = {}
+        self.seasonalities = {}
+        self.trends = {}
+        self.residuals = {}
+        for id in y.columns:
+            series_y = y[id]
+
+            # Determine the period of the seasonal component
+            if id not in self.periods or self.period is None:
```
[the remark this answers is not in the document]
remyogasawara: Yeah, I should have taken `self.period` out. I think I kept it to keep the functionality of `set_period` the same, but is the function used in the `PolynomialDecomposer` at all because I see that it is tested on both decomposers in `test_decomposer_set_period`. Also would the `periods` parameter be same for the single series case or would it be fine for period to be an integer for single series and then take in a dictionary for multiseries?

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -385,50 +480,114 @@ def _decompose_target(X, y, fh):
...
+        series_results = {}
+        # Iterate through each series id
+        for id in y.columns:
+            result_dfs = []
+            if not isinstance(X.index, pd.DatetimeIndex):
+                raise TypeError("Provided X should have datetimes in the index.")
+            if X.index.freq is None:
+                raise ValueError(
+                    "Provided DatetimeIndex of X should have an inferred frequency.",
+                )
```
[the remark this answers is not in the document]
remyogasawara: If I reset the indexes then it causes problems so I think this check `if not isinstance(X.index, pd.DatetimeIndex)` is necessary but removing the frequency didn't do anything so I'm not sure if checking the index frequency is necessary

evalml/tests/component_tests/decomposer_tests/test_decomposer.py
```diff
@@ -393,11 +443,33 @@ def test_decomposer_projected_seasonality_integer_and_datetime(
...
+        if isinstance(decomposer_child_class(), PolynomialDecomposer):
+            pytest.skip(
+                "Skipping Decomposer because multiseries is not implemented for Polynomial Decomposer",
+            )
+        X, y = multiseries_ts_data_unstacked
+        dts = pd.date_range("01-01-2000", periods=len(X))
+        datetime_index = pd.DatetimeIndex(dts)
+        X.index = datetime_index
+        y.index = datetime_index
+        X["date"] = dts
```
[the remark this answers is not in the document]
remyogasawara: most of the tests were using `multiseries_ts_data_unstacked` before but I'm going to switch them to `ts_multiseries_data` and that should simplify things

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
eccabay: Hmm - @jeremyliweishih, any thoughts? This feels like an edge case, but I can't follow the pipeline generation code well enough to understand why the target name isn't matching - it feels like this shouldn't actually be an issue.

evalml/pipelines/components/transformers/preprocessing/decomposer.py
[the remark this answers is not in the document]
eccabay: Huh, TIL!

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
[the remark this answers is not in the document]
eccabay: Cool, let's remove the frequency check while we're here.

The index check is a pet peeve of mine, I run into it so often - can we remove it by manually setting X's index to y's index, since we already manually set y's index to be a DatetimeIndex? This isn't a requirement, just trying to save myself some frustration 😅

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
```diff
@@ -79,49 +107,94 @@ def test_stl_sets_determined_period(
...
-    # Check the trend to make sure STL worked properly
-    pd.testing.assert_series_equal(
-        pd.Series(expected_trend),
-        pd.Series(stl.trend),
-        check_exact=False,
-        check_index=False,
-        check_names=False,
-        atol=0.3,
-    )
+    if variateness == "univariate":
```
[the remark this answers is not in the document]
eccabay: I think so, yep! I'll double check once you've pushed it

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
christopherbunn: After talking it over with @eccabay I think we should keep internal stacking inside of the decomposer if it isn't too much of a lift!

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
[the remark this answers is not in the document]
remyogasawara: In order to allow users to still pass in a period for both multivariate and univariate cases, I was thinking of keeping both `self.period` and `self.periods` as parameters, but that probably isn't the most efficient implementation so I'm open to other suggestions 😅

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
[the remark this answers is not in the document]
remyogasawara: Done! Should this be changed in the `PolynomialDecomposer` as well?

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -342,39 +444,32 @@ def get_trend_dataframe(self, X, y):
...
         # in ForecastingHorizon during decomposition.
         if not isinstance(y.index, pd.DatetimeIndex):
             y = self._set_time_index(X, y)
 
         self._check_oos_past(y)
 
-        result_dfs = []
-
-        def _decompose_target(X, y, fh):
+        def _decompose_target(X, y, fh, trend, seasonal, residual, period, id):
```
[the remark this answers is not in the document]
remyogasawara: I'm not sure I understand what you mean by this. If I just try to take out the variables it causes an error saying they're referenced before assignment.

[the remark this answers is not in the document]
eccabay: Which variable is throwing that error? It may require a renaming or two, but it should work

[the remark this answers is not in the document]
remyogasawara: Oh wait I think it is just because `_decompose_target` is defined before the variables since the variables are set inside the for loop. Would it be better to pass in the variables or take them out and define `_decompose_target` inside of the for loop?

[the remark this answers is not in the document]
remyogasawara: trend, seasonal and residual threw the error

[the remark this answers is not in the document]
eccabay: Ah, ok - I missed a bit of Python technicality here. Since we're reassigning `trend`, `seasonal`, and `residual` within the function, the Python interpreter treats them as local variables and doesn't consider them to necessarily be the same as those in the outer scope. It's fine to keep those as arguments that need passing in!

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
```diff
@@ -279,66 +366,148 @@ def test_stl_decomposer_get_trend_dataframe(
...
 def test_unsupported_frequencies(
     bad_frequency,
     generate_seasonal_data,
+    variateness,
 ):
     """This test exists to highlight that even though the underlying statsmodels STL component won't work
     for minute or annual frequencies, we can still run these frequencies with automatic period detection.
     """
-    X, y = generate_seasonal_data(real_or_synthetic="synthetic")(
+    # period = 7 if variateness == "univariate" else {}
```
eccabay, line 450: Extra comment 👀

eccabay (approved):
A few final comments, but nothing blocking. Awesome work!

evalml/tests/component_tests/decomposer_tests/test_stl_decomposer.py
remyogasawara: 🤦🏻‍♀️

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -158,43 +173,81 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None) -> STLDecomposer:
...
+                    X,
+                    series_y,
+                    acf_threshold=0.01,
+                    rel_max_order=5,
+                )
+                if self.period is None and len(y.columns) == 1:
+                    self.period = period
+                    self.update_parameters({"period": self.period})
+                elif self.period is not None and len(y.columns) == 1:
+                    period = self.period
```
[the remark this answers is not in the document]
remyogasawara: Since everything is accessing the period through `self.periods` I'm thinking we might not even need to update `self.period` at all, instead just use it to set `period` if its given by a user

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
```diff
@@ -385,50 +504,105 @@ def _decompose_target(X, y, fh):
...
         if isinstance(y, pd.Series):
-            result_dfs.append(_decompose_target(X, y, None))
-        elif isinstance(y, pd.DataFrame):
-            for colname in y.columns:
-                result_dfs.append(_decompose_target(X, y[colname], None))
-        return result_dfs
+            y = y.to_frame()
+        series_results = {}
+        # Iterate through each series id
+        for id in y.columns:
```
eccabay, line 510: This is so late in the process to realize this, but do we actually need to be calling `_decompose_target` in a loop? Since the decomposer handles multiseries generally, it should be able to handle multiseries here too, we can pass `self.periods` through instead of `period=period`, and switch around the logic to return the data in the format we need? That will prevent us from calling `Decomposer.fit()` too many times, which can be slow

remyogasawara: I modified this so that `get_trend_dataframe` returns a dictionary for multiseries, but for single series it is still returning a list of dataframes. In the `PolynomialDecomposer`, `get_trend_dataframe` returns a list of dataframes. I think this is because there is some multivariate implementation written here. Since they're both using `plot_decomposition` for single series, I wanted to keep the indexing the same so I left it as a list for `STLDecomposer` for now (even though it can be modified later to just return a dataframe). This is pretty inconsistent so I wanted to see what others think about it and if I should consider changing the return type of `get_trend_dataframe` for the multiseries and/or single series `STLDecomposer`.

christopherbunn (approved):
LGTM once the test case Becca mentioned is covered!

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
remyogasawara: I updated `get_trend_dataframe()` in the `STLDecomposer` to return `list(pd.DataFrame)` for single series and `dict(list(pd.DataFrame))` for multiseries, but it should be updated in the future to no longer be in a list #4294

evalml/pipelines/components/transformers/preprocessing/stl_decomposer.py
remyogasawara: Not a blocking issue, but tracking it in #4295

PR #4270, opened 2023-08-11
Integrate multiseries into AutoMLSearch

### LINKED ISSUE #4266: Integrate MSTS into AutoMLSearch (EvalML)

eccabay:
Closes #4266


evalml/pipelines/components/estimators/regressors/varmax_regressor.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         trend: Optional[str] = "c",
         random_seed: Union[int, float] = 0,
         maxiter: int = 10,
-        use_covariates: bool = True,
+        use_covariates: bool = False,
```
eccabay, line 64: Flipped this default for the sake of speed. My test example did not train within a 5 minute window when `use_covariates` was True, but it ran in <10 seconds when `use_covariates` was False.

evalml/pipelines/components/utils.py
```diff
@@ -56,7 +56,19 @@ def allowed_model_families(problem_type):
     return list(set([e.model_family for e in estimators]))
 
 
-def get_estimators(problem_type, model_families=None, excluded_model_families=None):
+def _filter_multiseries_estimators(estimators, is_multiseries):
+    if is_multiseries:
+        return [estimator for estimator in estimators if estimator.is_multiseries]
```
jeremyliweishih, line 61: nit and maybe for a follow up: could `estimator.is_multiseries` be something like `estimator.supports_multiseries`? Think its more clear especially since we're passing `is_multiseries` everywhere for now.

evalml/pipelines/components/estimators/regressors/varmax_regressor.py
jeremyliweishih: I think it would be nice to see performance tests for `use_covariates` turned on or off. Could possibly turn it off only for tests if performance is greatly improved with covariates!

jeremyliweishih (approved):
LGTM and agreed with @chukarsten on potentially refactoring `is_multiseries` into a separate problem type!

jeremyliweishih (approved):
LGTM again

evalml/automl/automl_search.py
```diff
@@ -651,6 +653,14 @@ def __init__(
                     f"Dataset size is too small to create holdout set. Minimum dataset size is {self._HOLDOUT_SET_MIN_ROWS} rows, X_train has {len(X_train)} rows. Holdout set evaluation is disabled.",
                 )
 
+        # For multiseries problems, we need to mke sure that the data is primarily ordered by the time_index rather than the series_id
```
MichaelFu512, line 656: Nit: we need to `mke` sure -> we need to `make` sure

MichaelFu512 (approved):
Looks good to me

PR #4274, opened 2023-08-15
Fixed Data Checks errors

### LINKED ISSUE #4212: ReadTheDocs - Data Checks errors
The following errors were notices in the [docs](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html):

The [initial](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html#Missing-Data) cell is throwing the error `The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.`

The target leakage [cell](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html#Missing-Data) is throwing `One or more pairs of columns did not share enough rows of non-null data to measure the relationship.  The measurement for these columns will be NaN.  Use 'extra_stats=True' to get the shared rows for each pair of columns.`

The multicollinearity [cell](https://evalml.alteryx.com/en/stable/user_guide/data_checks.html#Multicollinearity) is throwing `One or more pairs of columns did not share enough rows of non-null data to measure the relationship.  The measurement for these columns will be NaN.  Use 'extra_stats=True' to get the shared rows for each pair of columns.`

remyogasawara:
Resolves #4212


docs/source/release_notes.rst
```diff
@@ -3,6 +3,7 @@ Release Notes
 **Future Releases**
     * Enhancements
     * Fixes
+        * Fixed Data Checks errors :pr:`4274`
```
eccabay, line 6: This is a bit vague for a release note, users should be able to look at this and understand what changed. Can you move this to the `Documentation Changes` section and rephrase it to something along the lines of `Removed erroneous warnings from Data Checks User Guide page`?

eccabay (commented):
Awesome! Just a note about the release note, but otherwise looks good

docs/source/disable-warnings.py
```diff
@@ -1,3 +1,6 @@
+# flake8: noqa 401 imported to force console mode for tqdm in jupyter notebooks
+from tqdm.auto import tqdm
```
remyogasawara, line 2: Also added this import to get rid of this error `Using `tqdm.autonotebook.tqdm` in notebook mode. Use `tqdm.tqdm` instead to force console mode (e.g. in jupyter console)` that appears in all the notebooks. Let me know if you think we should add it to a different file or anything

jeremyliweishih (approved):

MichaelFu512 (approved):

christopherbunn (approved):

PR #4275, opened 2023-08-16
Fixed error when stacking data with no exogenous variables

### LINKED ISSUE #4276: Error when stacking X dataframes without exogenous variables

christopherbunn:
Resolves #4276

### DIFF
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -1490,7 +1490,8 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
         starting_index (int): The starting index to use for the stacked DataFrame. If None, the starting index
             will match that of the input data. Defaults to None.
         series_id_values (set): The unique values of a series ID, used to generate the index. If None, values will
-            be generated from X column values. Defaults to None.
+            be generated from X column values. Required if X only has time index values and no exogenous values.
+            Defaults to None.
 
     Returns:
         pd.DataFrame: The restacked features.
@@ -1507,7 +1508,7 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
 
     if len(series_ids) == 0:
         raise ValueError(
-            "X has no exogenous variables and `series_id_values` is None.",
+            "Series ID values needs to be passed in X column values or as a set with the `series_id_values` parameter.",
         )
 
     time_index_col = X[time_index].repeat(len(series_ids)).reset_index(drop=True)
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -1489,7 +1489,7 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
         time_index (str): The name of the time index column.
         starting_index (int): The starting index to use for the stacked DataFrame. If None, the starting index
             will match that of the input data. Defaults to None.
-        series_id_values (set): The unique values of a series ID, used to generate the index. If None, values will
+        series_id_values (set, list): The unique values of a series ID, used to generate the index. If None, values will
             be generated from X column values. Required if X only has time index values and no exogenous values.
             Defaults to None.
 
@@ -1508,7 +1508,7 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
 
     if len(series_ids) == 0:
         raise ValueError(
-            "Series ID values needs to be passed in X column values or as a set with the `series_id_values` parameter.",
+            "Series ID values need to be passed in X column values or as a set with the `series_id_values` parameter.",
         )
 
     time_index_col = X[time_index].repeat(len(series_ids)).reset_index(drop=True)
--- a/evalml/pipelines/utils.py
+++ b/evalml/pipelines/utils.py
@@ -1401,6 +1401,7 @@ def unstack_multiseries(
     # Perform the unstacking
     X_unstacked_cols = []
     y_unstacked_cols = []
+    new_time_index = None
     for s_id in series_id_unique:
         single_series = full_dataset[full_dataset[series_id] == s_id]
 
@@ -1417,8 +1418,11 @@ def unstack_multiseries(
                 X_unstacked_cols.append(new_column)
 
     # Concatenate all the single series to reform dataframes
-    X_unstacked = pd.concat(X_unstacked_cols, axis=1)
     y_unstacked = pd.concat(y_unstacked_cols, axis=1)
+    if len(X_unstacked_cols) == 0:
+        X_unstacked = pd.DataFrame(index=y_unstacked.index)
+    else:
+        X_unstacked = pd.concat(X_unstacked_cols, axis=1)
 
     # Reset the axes now that they've been unstacked, keep time info in X
     X_unstacked = X_unstacked.reset_index()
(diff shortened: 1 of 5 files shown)


machineFL:

evalml/pipelines/utils.py
```diff
@@ -1466,21 +1470,29 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
...
         original_columns.add("_".join(separated_name[:-1]))
-        series_ids.add(separated_name[-1])
+        if series_id_values is None:
+            series_ids.add(separated_name[-1])
 
     restacked_X = []
 
+    if len(series_ids) == 0:
+        raise ValueError(
+            "Unable to stack X as X had no exogenous variables and `series_id_values` is None.",
```
jeremyliweishih, line 1493: I think this would be more accurate if it was `no exogenous variables or `series_id_values` is None`

evalml/pipelines/utils.py
```diff
@@ -1493,10 +1505,25 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
                 starting_index=starting_index,
             ),
         )
-    restacked_X = pd.concat(restacked_X, axis=1)
-
     time_index_col = X[time_index].repeat(len(series_ids)).reset_index(drop=True)
-    time_index_col.index = restacked_X.index
-    restacked_X[time_index] = time_index_col
+
+    if len(restacked_X) == 0:
```
jeremyliweishih, line 1510: can you help me understand why we need this logic block here? If `restacked_X` has length 0 why don't we error out?

jeremyliweishih (commented):
some questions

evalml/pipelines/utils.py
christopherbunn: Changed it to `"X has no exogenous variables and `series_id_values` is None."` as it's a bit more succinct

christopherbunn:
@jeremyliweishih @eccabay yep, at a high level we need to handle the case where the unstacked data is only composed of a time index column. This case arises when the stacked data is solely composed of a time index column and a series ID column. When the data is unstacked, we essentially only retain the unique time stamps and drop any series ID information from X (and the series ID values essentially goes to y).

When stacking, we need to know how many series ID values there are so that we can repeat the number of timestamp values for each series ID value as seen [here](https://github.com/alteryx/evalml/blob/aa73702810dd99ffa4059a902f366b6dcffc6a95/evalml/pipelines/utils.py#L1498). Since `stack_X()` only takes in X (which does not have series ID info), we must somehow pass it in from an external source. We could do one of the following:
1. We take in a set of the current series ID values as the `series_id_values` parameter. The length of this set is then used to repeat the timestamps an appropriate number of times. This is our current approach.
2. Alternatively, we can just take in just the number of unique series IDs as an integer with a `n_series_ids` parameter.
3. We can expand out `stack_X()` to optionally take in the `y` dataframe as an optional parameter. If passed in, we can use the column names of `y` to generate the series ID values. 

Open to either of these approaches or any alternatives!

evalml/pipelines/utils.py
```diff
@@ -1466,21 +1470,29 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
...
     original_columns = set()
-    series_ids = set()
+    series_ids = series_id_values or set()
     for col in X.columns:
         if col == time_index:
             continue
         separated_name = col.split("_")
         original_columns.add("_".join(separated_name[:-1]))
-        series_ids.add(separated_name[-1])
+        if series_id_values is None:
```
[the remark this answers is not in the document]
christopherbunn: I can refactor this for loop to only run if `series_id_values` is not set.

evalml/pipelines/utils.py
christopherbunn: `restacked_X` can be zero in the case where we only have the time_index columns in the unstacked `X` dataframe. We want to be able to still stack in this case, which is why we have this logic block.

evalml/pipelines/utils.py
```diff
@@ -1381,6 +1381,7 @@ def unstack_multiseries(
     # Perform the unstacking
     X_unstacked_cols = []
     y_unstacked_cols = []
+    new_time_index = None
```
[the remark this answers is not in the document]
christopherbunn: Agreed, deleted.

evalml/pipelines/utils.py
christopherbunn: That being said, with the refactor I just put in, I ended up condition to be if len(`original_columns`) == 0.

eccabay:
@christopherbunn gotcha, I understand the challenge now. I think your new implementation makes sense, although I still have to do a full review. If possible, I think we should be as explicit as possible that `series_id_values` is a required parameter if the only column in X is the time index - maybe specifically checking and calling that out as the root error?

christopherbunn:
@eccabay makes sense to me, I updated the docstring and error message that is raised to clarify this. Open to tweaking the wording of either though

evalml/pipelines/utils.py
```diff
@@ -1486,37 +1489,61 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
         time_index (str): The name of the time index column.
         starting_index (int): The starting index to use for the stacked DataFrame. If None, the starting index
             will match that of the input data. Defaults to None.
+        series_id_values (set): The unique values of a series ID, used to generate the index. If None, values will
```
eccabay, line 1492: Can you add a check to make sure `series_id_values` is a set? I have a feeling it'll be a common error to pass in a list instead. Unless it's not actually necessary for it to be a set instead of a list, in which case we could mention both here.

evalml/pipelines/utils.py
```diff
@@ -1486,37 +1489,61 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
...
+        for col in X.columns:
+            if col == time_index:
+                continue
+            separated_name = col.split("_")
+            original_columns.add("_".join(separated_name[:-1]))
+            series_ids.add(separated_name[-1])
+
+    if len(series_ids) == 0:
+        raise ValueError(
+            "Series ID values needs to be passed in X column values or as a set with the `series_id_values` parameter.",
```
eccabay, line 1511: Nit: `values needs` -> `values need`

evalml/pipelines/utils.py
```diff
@@ -1486,37 +1489,61 @@ def stack_X(X, series_id_name, time_index, starting_index=None):
...
-    time_index_col.index = restacked_X.index
-    restacked_X[time_index] = time_index_col
+
+    if len(original_columns) == 0:
+        start_index = starting_index or X.index[0]
+        stacked_index = pd.RangeIndex(
+            start=start_index,
+            stop=start_index + len(time_index_col),
+        )
+        time_index_col.index = stacked_index
```
eccabay, line 1522: I don't think we need this line

eccabay (approved):
Thanks! Just a few nits

jeremyliweishih (approved):
LGTM but agree with Becca's nits!

evalml/pipelines/utils.py
christopherbunn: sure, updated to clarify it can also be a list as well.

evalml/pipelines/utils.py
christopherbunn: We need it in the case where the `time_index_col` has a different starting index value other than 0. Without it, the test cases where `starting_index` is not None will fail.

PR #4284, opened 2023-08-21
Add multiseries support to graph_prediction_vs_actual_over_time

### LINKED ISSUE #4285: `graph_prediction_vs_actual_over_time` supports visualizing multiseries time series performance.
### Description
As a user and developer of multiseries time series in EvalML, I would like our `graph_prediction_vs_actual_over_time` function to support visualizing multiseries time series performance.

There are two options for how to update the function:
- An argument passed in to the function delineating which single series of the multiseries dataset to plot
- The function automatically generates separate plot for each dataset
- 1 & 2 combined, the function will automatically plot all series, unless a single series is passed in in which case it will plot only the one

### Acceptance Criteria (AC)
-  `graph_prediction_vs_actual_over_time` takes in stacked data
-  `graph_prediction_vs_actual_over_time` provides the option to visualize multiseries problems, as described above

### Jira link (TML-8241)
https://alteryx.atlassian.net/browse/TML-8241

MichaelFu512:
### Pull Request Description
Closes #4285

### A small snippet of what `graph_prediction_vs_actual_over_time` looks like for multiseries:
Screen Shot 2023-08-21 at 9 24 40 AM


machineFL:

evalml/model_understanding/visualizations.py
```diff
@@ -419,6 +446,53 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, X_train, y_train, dates
...
+                    name=f"Series {id}: Prediction",
+                ),
+                row=curr_count + 1,
+                col=1,
+            )
+            fig.update_xaxes(title_text="Time")
+            fig.update_yaxes(title_text=y.name)
+        if single_series is not None:
+            fig.update_layout(
+                height=600,
```
jeremyliweishih, line 484: is there any autosizing we can take advantage of? Also not sure if we need the single_series case: can it just match what we already had before for time series?

evalml/model_understanding/visualizations.py
```diff
@@ -419,6 +446,53 @@ def graph_prediction_vs_actual_over_time(pipeline, X, y, X_train, y_train, dates
         dates,
     )
 
+    if is_multiseries(pipeline.problem_type):
```
jeremyliweishih, line 449: nit: can we wrap the original non-multiseries case in the else block and only return at the end of the method? Helpful for code clarity.

jeremyliweishih (changes requested):
One small testing change but LGTM otherwise.

evalml/model_understanding/visualizations.py
MichaelFu512: > Is there any autosizing we can take advantage of?

- For python plotly, it doesn't look like there's any autosizing besides doing what I did in line 490 (though correct me if I'm wrong). If I left the parameters blank for the multiseries graph (i.e. don't define x and y) and let it use the default values the graph looks really squished.
Screen Shot 2023-08-21 at 11 33 29 AM

- For single series though I can probably leave the parameters blank since it is only one graph.

> Also not sure if we need the single_series case: can it just match what we already had before for time series?

- The issue is that for single_series case I want the data that corresponds to that single series. 
    - In the before case for time series, y is taken from `data["target"]` whereas for single series I want to take it from `curr_df["target"]` where `curr_df = data[data["series_id"] == single_series]`
    - If I wanted to do that with the before code, I'd probably have to add more `if` statements and such so it made more sense to me to have it included with the multiseries case.
    - Also I'd like for the single series plot to have a different title than the time series plot.

evalml/utils/__init__.py
```diff
@@ -41,3 +41,4 @@
     _determine_downcast_type,
     _get_incompatible_nullable_types,
 )
+from evalml.utils.lazy_loader import LazyLoader
```
jeremyliweishih, line 44: is this needed?

evalml/model_understanding/visualizations.py
```diff
@@ -373,25 +374,44 @@ def get_prediction_vs_actual_over_time_data(pipeline, X, y, X_train, y_train, da
...
     Args:
         pipeline (TimeSeriesRegressionPipeline): Fitted time series regression pipeline.
-        X (pd.DataFrame): Features used to generate new predictions.
-        y (pd.Series): Target values to compare predictions against.
+        X (pd.DataFrame): Features used to generate new predictions. If problem is multiseries, X should be stacked.
+        y (pd.Series): Target values to compare predictions against. If problem is multiseries, y should be stacked.
         X_train (pd.DataFrame): Data the pipeline was trained on.
         y_train (pd.Series): Target values for training data.
         dates (pd.Series): Dates corresponding to target values and predictions.
+        single_series (str): The single series that will be plotted from multiseries. Defaults to None
```
eccabay, line 414: Can you update this a little bit to make it clearer what this is referencing? i.e. along the lines of "a single series id value to plot just one series in a multiseries dataset"

eccabay (approved):
Thanks! Just a docstring comment

PR #4287, opened 2023-08-24
Updated `excluded_model_families` data type

### LINKED ISSUE #4259: "allowed families" argument ignores LightGBM
Hello everyone! Unfortunately when I use `allowed_model_families`  in the `AutoMLSearch` class, it is absolutely ignored and other models like Random Forest are used. I want to use only LightGBM  and it is really annoying to see other models during the search. Also there is funny difference between `allowed_model_families`  and `excluded_model_families`. First one uses `str` data type as an element of a list which goes into the argument, and latter one uses only `ModelFamily` instances. What's the reason behind this difference? 

#### The command I use to run

```python
automl = AutoMLSearch(
    X_train=X_train,
    y_train=y_train,
    data_splitter=cv,
    problem_type="binary",
    objective='f1',
    additional_objectives=["auc", "precision"],
    allowed_model_families=["lightgbm"],
    max_batches=3,
    optimize_thresholds=True,
    verbose=True,
)

automl.search()

```
image

As you can see Random Forest are used somehow, but I do not need it. Only LightGBM. Thanks in advance.

remyogasawara:
Resolves #4259 

Updated `excluded_model_families` to allow both `str` and `ModelFamily` data types and updated corresponding documentation.


evalml/tests/automl_tests/test_automl.py
```diff
@@ -5356,7 +5356,7 @@ def test_excluded_model_families(
         y_train=y,
         problem_type=ProblemTypes.BINARY,
         automl_algorithm=automl_algorithm,
-        excluded_model_families=[ModelFamily.RANDOM_FOREST],
+        excluded_model_families=[ModelFamily.RANDOM_FOREST, "linear_model"],
```
jeremyliweishih, line 5359: Let's also assert than any linear models are not created as well.

eccabay (approved):
Thank you!

PR #4288, opened 2023-08-25
Unpinned ipywidgets

### LINKED ISSUE #4099: Unpin ipywidgets
After #4097 is merged, we want to revisit the errors that led to us pinning ipywidgets so we can unpin the dependency.

See comments in issue #4098

remyogasawara:
Resolves #4099


.github/meta.yaml
```diff
@@ -72,7 +72,7 @@ outputs:
         - python-kaleido >=0.1.0
         - matplotlib-base >=3.3.3
         - seaborn >=0.11.1
-        - ipywidgets >=7.5, <8.0.5
+        - ipywidgets >=7.5
```
eccabay, line 75: Looks like ipywidgets must have fixed whatever was causing the errors we saw back in March, if we're able to unpin and have no tests fail. However, I think if a user happened to have ipywidgets v8.0.5 installed in the environment where they're installing evalml, the issues would still exist. 

Could you install ipywidgets 8.0.5 and the following versions locally, to see if they still error? If so, we should exclude those specific versions.

jeremyliweishih (approved):

.github/meta.yaml
remyogasawara: The tests passed for v.8.0.5-7 as well!

eccabay: 👀 surprising, but glad it'll work 🚢

eccabay (approved):

PR #4312, opened 2023-09-18
Update split_data to call split_multiseries_data

### LINKED ISSUE #4311: Holdout data splitter creates NaN values with multiseries datasets

eccabay:
Closes #4311

### DIFF
--- a/evalml/preprocessing/utils.py
+++ b/evalml/preprocessing/utils.py
@@ -4,7 +4,12 @@ from sklearn.model_selection import ShuffleSplit, StratifiedShuffleSplit
 
 from evalml.pipelines.utils import stack_data, stack_X, unstack_multiseries
 from evalml.preprocessing.data_splitters import TrainingValidationSplit
-from evalml.problem_types import is_classification, is_regression, is_time_series
+from evalml.problem_types import (
+    is_classification,
+    is_multiseries,
+    is_regression,
+    is_time_series,
+)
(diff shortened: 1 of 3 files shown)


evalml/preprocessing/utils.py
```diff
@@ -144,6 +149,23 @@ def split_data(
         1    9
         dtype: int64
     """
+    if is_multiseries(problem_type) and isinstance(y, pd.Series):
+        series_id = problem_configuration.get("series_id")
```
jeremyliweishih, line 153: Should check if problem_configuration exists and key exists before getting and if not also raise the same value error.

evalml/tests/preprocessing_tests/test_split_data.py
```diff
@@ -29,6 +30,8 @@ def test_split_data(
         X, y = X_y_regression
     problem_configuration = None
     if is_time_series(problem_type):
+        if is_multiseries(problem_type):
+            pytest.skip("Multiseries time series is tested separately")
```
jeremyliweishih, line 34: Instead of skipping could we mock out the multi series split and asset that it's called?

jeremyliweishih (changes requested):
Some quick comments but otherwise LGTM

MichaelFu512 (approved):

christopherbunn (approved):

PR #4315, opened 2023-09-19
Pin pandas version

### LINKED ISSUE #4313: Pin pandas under 2.1.0
Pandas 2.1.0 causes test failures with a large number of warnings, among a few other issues. We should pin pandas until we resolve these issues

eccabay:
Closes #4313

### DIFF
--- a/.github/workflows/linux_unit_tests_with_latest_deps.yaml
+++ b/.github/workflows/linux_unit_tests_with_latest_deps.yaml
@@ -15,6 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - python_version: "3.9"
+            command: 'git-test-automl'
           - python_version: "3.8"
             command: 'git-test-automl'
           - python_version: "3.8"
--- a/.github/meta.yaml
+++ b/.github/meta.yaml
@@ -25,7 +25,7 @@ outputs:
         - setuptools ==58.0.4
       run:
         - numpy >=1.21.0
-        - pandas >=1.5.0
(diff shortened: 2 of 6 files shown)


.github/workflows/linux_unit_tests_with_latest_deps.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - python_version: "3.9"
+            command: 'git-test-automl'
```
christopherbunn, line 19: Do we need to test on 3.9? If so, should we just bump up the Python versions to 3.9 across all tests?

christopherbunn (approved):
One question but otherwise LGTM

.github/workflows/linux_unit_tests_with_latest_deps.yaml
eccabay: Testing on 3.9 is how we can double check that the pinning worked, since pandas 2.1.0 isn't installed on 3.8. I don't think we should just immediately bump everything up, but I do agree that arbitrarily just running `git-test-automl` for 3.9 and not the others doesn't make much sense.

Thoughts on testing 3.9 alongside 3.8 for all PR tests? I don't think it'll add any time, since the tests all run in parallel

jeremyliweishih: I think adding one additional test for 3.9 is sufficient.

jeremyliweishih (approved):

PR #4357, opened 2023-10-25
Changed target name/series ID divider and added ability to return series ID column with predictions

### LINKED ISSUE #4359: Further refine multiseries stacking utilities
As an EvalML user, I noticed that when using datasets with underscores in the name, stacking and unstacking multiseries data occasionally causes the column names to disappear.

In addition, given that the series ID order can be inconsistent within a series, I would like to be able to return the series ID column with an outputted prediction.

christopherbunn:
Resolves #4359

### DIFF
--- a/evalml/pipelines/multiseries_regression_pipeline.py
+++ b/evalml/pipelines/multiseries_regression_pipeline.py
@@ -161,17 +161,15 @@ class MultiseriesRegressionPipeline(TimeSeriesRegressionPipeline):
             objective,
             calculating_residuals,
         )
-        if include_series_id:
-            stacked_predictions = stack_data(
-                unstacked_predictions,
-                include_series_id=True,
-                series_id_name=self.series_id,
-            )
-        else:
-            stacked_predictions = stack_data(unstacked_predictions)
+        stacked_predictions = stack_data(
+            unstacked_predictions,
+            include_series_id=include_series_id,
+            series_id_name=self.series_id,
+        )
 
         # Index will start at the unstacked index, so we need to reset it to the original index
         stacked_predictions.index = X.index
+        stacked_predictions = infer_feature_types(stacked_predictions)
         return stacked_predictions
 
     def get_forecast_period(self, X):
--- a/evalml/pipelines/multiseries_regression_pipeline.py
+++ b/evalml/pipelines/multiseries_regression_pipeline.py
@@ -128,7 +128,9 @@ class MultiseriesRegressionPipeline(TimeSeriesRegressionPipeline):
             self.time_index,
             self.input_target_name,
         )
+
         # Order series columns to be same as expected input feature names
+        # and filter to only include features in `X_unstacked`.
         input_features = list(self.input_feature_names.values())[0]
         X_unstacked = X_unstacked[
             [feature for feature in input_features if feature in X_unstacked.columns]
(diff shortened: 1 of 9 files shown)


machineFL:

evalml/pipelines/utils.py
```diff
@@ -1418,7 +1418,7 @@ def unstack_multiseries(
         for column_name in full_dataset.columns.drop([time_index, series_id]):
             new_column = single_series[column_name]
             new_column.index = new_time_index
-            new_column.name = f"{column_name}_{s_id}"
+            new_column.name = f"{column_name}|{s_id}"
```
chukarsten, line 1421: Yea, I think seeing this change here and later on makes me wonder if we should have a an importable constant like `SEPARATOR_SYMBOL` so that they can all be changed in the same place?

evalml/pipelines/utils.py
```diff
@@ -1511,8 +1516,8 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
         for col in X.columns:
             if col == time_index:
                 continue
-            separated_name = col.split("_")
-            original_columns.add("_".join(separated_name[:-1]))
+            separated_name = col.split("|")
+            original_columns.add("".join(separated_name[:-1]))
```
chukarsten, line 1520: Do we really want to join with no separator?

evalml/pipelines/multiseries_regression_pipeline.py
```diff
@@ -133,7 +161,14 @@ def predict_in_sample(
...
         )
-        stacked_predictions = stack_data(unstacked_predictions)
+        if include_series_id:
+            stacked_predictions = stack_data(
+                unstacked_predictions,
+                include_series_id=True,
+                series_id_name=self.series_id,
+            )
+        else:
+            stacked_predictions = stack_data(unstacked_predictions)
```
eccabay, line 171: We shouldn't need to split these into separate logical calls - if we pass in the `series_id_name` when `include_series_id=False`, nothing will happen. This can be condensed to
```
stacked_predictions = stack_data(
    unstacked_predictions,
    include_series_id=include_series_id,
    series_id_name=self.series_id
)
```

eccabay (changes requested):
Just a few comments, and I think we're missing a test!

evalml/pipelines/multiseries_regression_pipeline.py
```diff
@@ -125,6 +128,31 @@ def predict_in_sample(
...
+                if feature in X_train_unstacked.columns
+            ]
+        ]
+        y_overlapping_features = [
+            feature
+            for feature in y_train_unstacked.columns
+            if feature in y_unstacked.columns
+        ]
+        y_unstacked = y_unstacked[y_overlapping_features]
+        y_train_unstacked = y_train_unstacked[y_overlapping_features]
```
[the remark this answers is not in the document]
christopherbunn: - Will push an additional test case
- This covers the case when we're forecasting. When we're forecasting, we only pass in the dates + the series IDs. If there's extra features, we can't add them in.  
- The goal is to do both for the reason described above.

jeremyliweishih (approved):
LGTM after recent revisions

evalml/pipelines/multiseries_regression_pipeline.py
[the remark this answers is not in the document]
christopherbunn: Added `test_multiseries_pipeline_predict_in_sample_series_out_of_order()` which evaluates this case.

eccabay (approved):

PR #4385, opened 2024-01-26
Added support for additional estimators for multiseries datasets

### LINKED ISSUE #4386: Add support for additional estimators for multiseries problems
Currently, our model performance is relatively limited with just the VARMAX regressor. We can generate better results if we included additional non-time-series-native estimators for multiseries problems.

christopherbunn:
Resolves #4386

### DIFF
--- a/evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
@@ -1,4 +1,5 @@
 """Transformer to drop rows specified by row indices."""
+import pandas as pd
 from woodwork import init_series
 
 from evalml.pipelines.components.transformers import Transformer
@@ -43,12 +44,25 @@ class DropNaNRowsTransformer(Transformer):
         y_t = infer_feature_types(y) if y is not None else None
 
         X_t_schema = X_t.ww.schema
+        y_t_logical = None
+        y_t_semantic = None
         if y_t is not None:
-            y_t_logical = y_t.ww.logical_type
-            y_t_semantic = y_t.ww.semantic_tags
+            if isinstance(y_t, pd.DataFrame):
+                y_t_logical = y_t.ww.logical_types
+                y_t_semantic = y_t.ww.semantic_tags
+            else:
+                y_t_logical = y_t.ww.logical_type
+                y_t_semantic = y_t.ww.semantic_tags
 
         X_t, y_t = drop_rows_with_nans(X_t, y_t)
         X_t.ww.init_with_full_schema(X_t_schema)
         if y_t is not None:
-            y_t = init_series(y_t, logical_type=y_t_logical, semantic_tags=y_t_semantic)
+            if isinstance(y_t, pd.DataFrame):
+                y_t.ww.init(logical_types=y_t_logical, semantic_tags=y_t_semantic)
+            else:
+                y_t = init_series(
+                    y_t,
+                    logical_type=y_t_logical,
+                    semantic_tags=y_t_semantic,
+                )
         return X_t, y_t
--- a/evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
@@ -49,10 +49,9 @@ class DropNaNRowsTransformer(Transformer):
         if y_t is not None:
             if isinstance(y_t, pd.DataFrame):
                 y_t_logical = y_t.ww.logical_types
-                y_t_semantic = y_t.ww.semantic_tags
             else:
                 y_t_logical = y_t.ww.logical_type
-                y_t_semantic = y_t.ww.semantic_tags
+            y_t_semantic = y_t.ww.semantic_tags
 
         X_t, y_t = drop_rows_with_nans(X_t, y_t)
         X_t.ww.init_with_full_schema(X_t_schema)
--- a/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
@@ -129,7 +129,16 @@ class TimeSeriesFeaturizer(Transformer):
 
         # For the multiseries case, where we only want the start delay lag for the baseline
         if isinstance(y, pd.DataFrame):
-            self.statistically_significant_lags = [self.start_delay]
+            self.statistically_significant_lags = {}
+            for column in y.columns:
+                self.statistically_significant_lags[
+                    column
+                ] = self._find_significant_lags(
+                    y[column],
+                    conf_level=self.conf_level,
+                    start_delay=self.start_delay,
+                    max_delay=self.max_delay,
+                )
         else:
             self.statistically_significant_lags = self._find_significant_lags(
                 y,
@@ -234,7 +243,7 @@ class TimeSeriesFeaturizer(Transformer):
             col = data[col_name]
             if categorical_columns and col_name in categorical_columns:
                 col = X_categorical[col_name]
-            for t in self.statistically_significant_lags:
+            for t in self.statistically_significant_lags[col_name]:
                 lagged_features[self.df_colname_prefix.format(col_name, t)] = col.shift(
                     t,
                 )
--- a/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
@@ -127,7 +127,7 @@ class TimeSeriesFeaturizer(Transformer):
         if self.time_index is None:
             raise ValueError("time_index cannot be None!")
 
-        # For the multiseries case, where we only want the start delay lag for the baseline
+        # For the multiseries case, each series ID has individualized lag values
         if isinstance(y, pd.DataFrame):
             self.statistically_significant_lags = {}
             for column in y.columns:
--- a/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
@@ -243,7 +243,25 @@ class TimeSeriesFeaturizer(Transformer):
             col = data[col_name]
             if categorical_columns and col_name in categorical_columns:
                 col = X_categorical[col_name]
-            for t in self.statistically_significant_lags[col_name]:
+            # Lags are stored in a dict for multiseries problems
+            # Returns the lags corresponding to the series ID value
+            if isinstance(self.statistically_significant_lags, dict):
+                from evalml.pipelines.utils import MULTISERIES_SEPARATOR_SYMBOL
+
+                col_series_id = (
+                    MULTISERIES_SEPARATOR_SYMBOL
+                    + col_name.split(MULTISERIES_SEPARATOR_SYMBOL)[-1]
+                )
+                for (
+                    series_id_target_name,
+                    lag_list,
+                ) in self.statistically_significant_lags.items():
+                    if series_id_target_name.endswith(col_series_id):
+                        lags = lag_list
+                        break
+            else:
+                lags = self.statistically_significant_lags
+            for t in lags:
                 lagged_features[self.df_colname_prefix.format(col_name, t)] = col.shift(
                     t,
                 )
--- a/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
@@ -135,8 +135,9 @@ class TimeSeriesFeaturizer(Transformer):
             )
         else:
             # For the multiseries case, each series ID has individualized lag values
-            if isinstance(y, pd.Series):
-                y = y.to_frame()
+            if isinstance(y, pd.Series) or isinstance(y, np.ndarray):
+                y = pd.DataFrame(y)
+
             self.statistically_significant_lags = {}
             for column in y.columns:
                 self.statistically_significant_lags[
--- a/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
+++ b/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
@@ -127,8 +127,16 @@ class TimeSeriesFeaturizer(Transformer):
         if self.time_index is None:
             raise ValueError("time_index cannot be None!")
 
-        # For the multiseries case, each series ID has individualized lag values
-        if isinstance(y, pd.DataFrame):
+        if y is None:
+            # Set lags to all possible lag values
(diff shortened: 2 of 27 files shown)


machineFL:

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -98,19 +97,20 @@ def test_iterative_algorithm_init(
     assert algo.batch_number == 0
     assert algo.default_max_batches == 1
     estimators = get_estimators(problem_type)
-    decomposer = [STLDecomposer] if is_regression(problem_type) else []
+    decomposer = [True, False] if is_regression(problem_type) else [True]
```
jeremyliweishih, line 100: Can you add a comment clarifying why you're using true false instead of the decomposer name?

jeremyliweishih (approved):
LGTM pending performance tests

MichaelFu512 (approved):

evalml/pipelines/utils.py
```diff
@@ -1508,22 +1503,28 @@ def stack_X(X, series_id_name, time_index, starting_index=None, series_id_values
...
         for col in X.columns:
             if col == time_index:
                 continue
             separated_name = col.split(MULTISERIES_SEPARATOR_SYMBOL)
             original_columns.add(MULTISERIES_SEPARATOR_SYMBOL.join(separated_name[:-1]))
-            series_ids.add(separated_name[-1])
+            series_ids.append(separated_name[-1])
+        # Remove duplicates
+        seen = set()
+        series_ids = [val for val in series_ids if not (val in seen or seen.add(val))]
```
eccabay, line 1527: We can simplify this to `series_ids = list(set(series_ids))` - converting to a set automatically removes duplicates

evalml/preprocessing/utils.py
```diff
@@ -79,7 +79,7 @@ def split_multiseries_data(X, y, series_id, time_index, **kwargs):
 
     # Get unique series value from X if there is only the time_index column
     # Otherwise, this information is generated in `stack_X` from the column values
-    series_id_values = set(X[series_id]) if len(X_unstacked.columns) == 1 else None
+    series_id_values = X[series_id].unique() if len(X_unstacked.columns) == 1 else None
```
eccabay, line 82: What's the difference with this change? Is it just list vs set data format?

evalml/tests/automl_tests/test_iterative_algorithm.py
eccabay: I think it would be better to just parametrize the `include_decomposer` argument here - this and the below section are confusing to read out of context

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
```diff
@@ -129,7 +129,16 @@ def fit(self, X, y=None):
...
+            self.statistically_significant_lags = {}
+            for column in y.columns:
+                self.statistically_significant_lags[
+                    column
+                ] = self._find_significant_lags(
+                    y[column],
+                    conf_level=self.conf_level,
+                    start_delay=self.start_delay,
+                    max_delay=self.max_delay,
+                )
```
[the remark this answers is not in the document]
christopherbunn: Good point, I just consolidated it.

evalml/preprocessing/utils.py
christopherbunn: This was done to maintain the input series ID order. I can add a clarifying comment here.

evalml/tests/automl_tests/test_iterative_algorithm.py
christopherbunn: For this test, we are basically only checking the number of pipelines matches up. Before, we only needed to add the decomposer one once since there was one estimator type (VARMAX).

Now, since we have multiple estimator types, each estimator type will have one pipeline with a decomposer and another without a decomposer. As such, we need to have this [True, False] and to iterate through it in order to generate the correct number of pipelines.

evalml/pipelines/utils.py
eccabay: Ah I see, makes sense. Do we care about insertion order? (I assume we do, but I'm not sure why) It makes sense to do it this way then, but a comment calling this out is certainly helpful

evalml/pipelines/components/transformers/preprocessing/drop_nan_rows_transformer.py
```diff
@@ -43,12 +44,25 @@ def transform(self, X, y=None):
...
+        y_t_semantic = None
         if y_t is not None:
-            y_t_logical = y_t.ww.logical_type
-            y_t_semantic = y_t.ww.semantic_tags
+            if isinstance(y_t, pd.DataFrame):
+                y_t_logical = y_t.ww.logical_types
+                y_t_semantic = y_t.ww.semantic_tags
+            else:
+                y_t_logical = y_t.ww.logical_type
+                y_t_semantic = y_t.ww.semantic_tags
```
eccabay, line 55: Super tiny nit: we can move `y_semantic` out of the logical branches since the call is the same for both

evalml/pipelines/multiseries_regression_pipeline.py
```diff
@@ -163,16 +163,45 @@ def predict_in_sample(
             objective,
             calculating_residuals,
         )
+        unstacked_predictions = unstacked_predictions[
+            [
+                series_id_target
+                for series_id_target in y_train_unstacked.columns
```
eccabay, line 169: Looks like we'd want to use `self.series_id_target_names` instead of `y_train_unstacked.columns` to mirror above, right?

evalml/tests/automl_tests/test_iterative_algorithm.py
```diff
@@ -98,19 +97,21 @@ def test_iterative_algorithm_init(
...
+                include_decomposer=include_decomposer,
                 parameters=search_parameters,
             )
             for estimator in estimators
-        ]
-        + decomposer,
+            for include_decomposer in (
+                [True, False] if is_regression(problem_type) else [False]
+            )
+        ],
```
eccabay, line 114: Sorry, I should have clarified with my request for parametrization - I meant that we should parametrize it over the whole function, like
```
@pytest.mark.parametrize("problem_type", ["binary", "multiseries time series regression"])
@pytest.mark.parametrize("include_decomposer", [True, False])
def test_iterative_algorithm_init(
    problem_type,
    include_decomposer
    X_y_binary,
    multiseries_ts_data_stacked,
):
if include_decomposer and problem_type=="binary":
    pytest.skip("Decomposers are not used in binary problems")
...
```
It makes it much easier to debug this way!

evalml/pipelines/utils.py
christopherbunn: Yep, when we restack the X columns we need it to be the correct order. It's used in line 1547 to generate the stacked X series ID values.

evalml/pipelines/multiseries_regression_pipeline.py
christopherbunn: When unstacked, the column names include the target name with the series ID values (`target|series_a`, `target|series_b`) rather than just the series ID values (`series_a`, `series_b`)

evalml/tests/automl_tests/test_iterative_algorithm.py
jeremyliweishih: I think a clarifying comment would be useful here 👍

evalml/pipelines/multiseries_regression_pipeline.py
```diff
@@ -163,16 +163,45 @@ def predict_in_sample(
...
+        ]
+        unstacked_predictions.index = X_unstacked[self.time_index]
         stacked_predictions = stack_data(
             unstacked_predictions,
-            include_series_id=include_series_id,
+            include_series_id=True,
             series_id_name=self.series_id,
         )
-
+        stacked_predictions = stacked_predictions.reset_index()
```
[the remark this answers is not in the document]
christopherbunn: The goal of this snippet is to set the index as the time index column, stack the data (and thus using the dates in the time index column to generate new stacked dates) and then resetting the index so that the resulting time index column can be used when we `pd.merge` later on in line 193. 

While it's possible to just copy over the time_index column from `X` after stacking, I think it's safer to just generate it from the `X_unstacked` index like this as we know for sure that the X_unstacked time_index aligns  with the unstacked_predictions whereas it's technically possible to have an `X` time_index that's out of order (and thus would be incorrect if we simply copied over this column). I'm open to suggestions for a cleaner implementation!

evalml/tests/automl_tests/test_iterative_algorithm.py
christopherbunn: To clarify, here's the intended behavior. 

- Binary problems have 6 estimators: LightGBM, Extra Trees, Elastic Net, XGBoost, RF, and Logistic Regression. Decomposers cannot be used with binary problems. As such, we want six pipelines. 
   - LightGBM Pipeline
   - Extra Trees Pipeline
   - Elastic Net Pipeline
   - XGBoost Pipeline
   - RF Pipeline
   - Logistic Regression Pipelines 

For a total of 5 pipelines. When we evaluate what we generate with `len(algo.allowed_pipelines)`, this evaluates to true. 

- For multiseries time series problems, there are 5 estimators: VARMAX, Extra Trees, XGBoost, RF, and Elastic Net. Since decomposers can be used with multiseries problems, we want to generate one pipeline with a decomposer and one pipeline without a decomposer. 
   - VARMAX w/ Decomposer Pipeline
   - VARMAX w/o Decomposer Pipeline 
   - Extra Trees w/ Decomposer Pipeline
   - Extra Trees w/o Decomposer Pipeline 
   - XGBoost w/ Decomposer Pipeline
   - XGBoost w/o Decomposer Pipeline 
   - RF w/ Decomposer Pipeline
   - RF w/o Decomposer Pipeline 
   - Elastic Net w/ Decomposer Pipeline
   - Elastic Net w/o Decomposer Pipeline 

For a total of 10 pipelines. When we evaluate what we generate with `len(algo.allowed_pipelines)`, this evaluates to true. 

Thus, I don't think making `include_decomposer` a test parameter makes sense since we're using this variable to make the pipelines with and without the decomposer when checking for the `mutliseries time series regression` problem type. We only need to test two cases here, which is (binary pipelines without decomposer) and (multiseries pipelines with and without decomposers).

As always, open to including clarifying comments or a cleaner implementation.

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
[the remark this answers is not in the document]
christopherbunn: We handle the y being null in `self._find_significant_lags` since we calculate all lags in that function (and just set significant lags to be all_lags if y is None). Should I put it off into it's own separate branch, even though the code would be identical to the case where y is a series? e.g. 

```
        # For the multiseries case, each series ID has individualized lag values
        if isinstance(y, pd.DataFrame):
            self.statistically_significant_lags = {}
            for column in y.columns:
                self.statistically_significant_lags[
                    column
                ] = self._find_significant_lags(
                    y[column],
                    conf_level=self.conf_level,
                    start_delay=self.start_delay,
                    max_delay=self.max_delay,
                )
        elif y is None:
            self.statistically_significant_lags = self._find_significant_lags(
                y,
                conf_level=self.conf_level,
                start_delay=self.start_delay,
                max_delay=self.max_delay,
            )
        else:
            self.statistically_significant_lags = self._find_significant_lags(
                y,
                conf_level=self.conf_level,
                start_delay=self.start_delay,
                max_delay=self.max_delay,
            )
```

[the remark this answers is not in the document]
eccabay: Ok, sorry for drilling into this so much, but I think I understand now. My new potentially hot take proposal is something like:
```
if y is None:
    self.statistically_significant_lags = np.arange(self.start_delay, self.start_delay +self. max_delay + 1)
else:
    if isinstance(y, pd.Series): 
        y = y.to_frame()
    for column in y.columns:
        self.statistically_significant_lags = ...
```
And then we can remove the handling of y being None from the static function. My argument for doing this is that calling all lags the statistically significant lags is a misnomer, since we didn't actually check statistical significance. This is me getting *very* into the weeds though, so I very much understand if you would rather keep things closer to the way they are 😅 

Regardless, even with your new proposal, we'd still be able to combine the two non y=None cases by casting the series to a dataframe

evalml/pipelines/multiseries_regression_pipeline.py
eccabay: I must be missing something then - don't we save `self.series_id_target_names = y_unstacked.columns` (where y_unstacked is the training y) above? Or does it change somewhere without me noticing?

evalml/pipelines/multiseries_regression_pipeline.py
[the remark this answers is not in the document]
eccabay: Ok, I think I understand now! I think a comment would be great. I also wonder if it would be useful to explicitly say `reset_index(drop=False)`, so that even if pandas changes their defaults we don't get screwed. 

My motivation here is that this is something that might be confusing to someone looking back at it in the future, since the goal isn't clear from the code itself. I hope that makes sense!

evalml/tests/automl_tests/test_iterative_algorithm.py
eccabay: Ooh I see. Makes sense! Sorry to drag you down this false path with me haha

eccabay (approved):
Looks solid! I think with a final few code comments this is all set 😎

evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py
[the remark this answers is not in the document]
christopherbunn: Your example makes sense to me, I don't see our behavior for `y is None` changing anytime soon so I'm comfortable with pulling that out and changing the function.

evalml/tests/automl_tests/test_iterative_algorithm.py
christopherbunn: Added a short one to this test

evalml/pipelines/multiseries_regression_pipeline.py
christopherbunn: Oh true, this one is on me 🤦‍♂️. Somehow I misread this as just the series ID values. I'll update it to use `self.series_id_target_names` instead for parity.

evalml/pipelines/multiseries_regression_pipeline.py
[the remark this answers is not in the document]
christopherbunn: Good call on the reset index parameter, I'll add that in. I'll add a clarifying comment or two so that it's clear what's going on here.

Your motivation makes sense! I feel like I've been so lost in the weeds of this implementation for a while now so it's good to have multiple pairs of eyes on this to highlight what's intuitive and what isn't 😅

Issue #4394, opened 2024-02-13
Bump up minimum kaleido version to 0.2.1


christopherbunn:
Currently, there are no arm64 wheels for kaleido 0.1.0. We should bump up this version to 0.2.0 and add additional flexibility for new versions in the requirements in the future.

MichaelFu512:
Closed from merging of #4423

PR #4423, opened 2024-05-16
Fix install test and uncap kaleido + various others

### LINKED ISSUE #4420: Fix install tests
The install tests haven't been working due to issues with the `install_test.yaml` file.

MichaelFu512:
### Pull Request Description
Closes #4420 and #4394
-----


evalml/tests/utils_tests/test_gen_utils.py
```diff
@@ -37,10 +37,14 @@ def in_container_arm64():
...
-        import plotly.io as pio
+    # if os.getenv("DOCKER_ARM", None):
+    import plotly.io as pio
 
-        pio.kaleido.scope.chromium_args += ("--single-process",)
+    pio.kaleido.scope.chromium_args += (
+        "--single-process",
+        "--headless",
+        "--disable-gpu",
+    )
```
MichaelFu512, line 47: Originally these were only meant for mac m1, but windows also needs these settings for testing, else some tests hang forever, causing workers to fail.

.github/workflows/install_test.yaml
```diff
@@ -37,16 +34,16 @@ jobs:
...
       - name: Build evalml package
         run: make package
       - name: Install evalml from sdist (not using cache)
-        if: steps.cache.outputs.cache-hit != 'true'
         run: |
           python -m pip install "unpacked_sdist/."
-      - name: Install evalml from sdist (using cache)
-        if: steps.cache.outputs.cache-hit == 'true'
-        run: |
-          python -m pip install "unpacked_sdist/." --no-deps
```
MichaelFu512, line 49: Whenever the tests would use cache, the test would fail.

pyproject.toml
```diff
@@ -78,9 +78,9 @@ dependencies = [
 [project.optional-dependencies]
 test = [
     "pytest == 7.1.2",
-    "pytest-xdist == 2.1.0",
-    "pytest-timeout == 1.4.2",
-    "pytest-cov == 2.10.1",
+    "pytest-xdist >= 2.1.0",
+    "pytest-timeout >= 1.4.2",
+    "pytest-cov >= 2.10.1",
```
MichaelFu512, line 83: Originally I uncapped them to see if they would fix the failing tests. They didn't fix them, but nothing broke because of them so just kept it.

eccabay (approved):

.github/workflows/install_test.yaml
```diff
@@ -37,16 +34,16 @@ jobs:
         with:
           path: ${{ env.pythonLocation }} 
           key: ${{ matrix.os- }}-${{ matrix.python_version }}-install-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml') }}-v01
+      - name: Set up Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+      - name: Set up cmake and libomp
+        run: brew install cmake libomp
```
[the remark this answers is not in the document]
MichaelFu512: In commit [ce422b5](https://github.com/alteryx/evalml/pull/4423/commits/ce422b5944e1a29541ff2eb673c05ac4f41039cb), the install test for MacOS unilaterally fail because:
>  ERROR: Could not build wheels for lightgbm, which is required to install pyproject.toml-based projects

.github/workflows/install_test.yaml
christopherbunn: Interesting, do you think it's noticably slower since we're not using the cache? If so, by how much?

christopherbunn (approved):
Single (non-blocking) question, LGTM

.github/workflows/install_test.yaml
MichaelFu512: Interestingly the time it takes varies.

The latest commit ([2b11a69](https://github.com/alteryx/evalml/pull/4423/commits/2b11a6918d3620df5eee8426390626714cb3bd47)) has the time that ubuntu takes for this step to be 8 seconds. For MacOS it takes 40 seconds for python3.9 and 3 minutes for python3.11.

Conversely, the second latest commit ([cd3f4f8](https://github.com/alteryx/evalml/pull/4423/commits/cd3f4f8110eb2d1053ab4443f993a947c344e9a3)) has the time that ubuntu takes hover around 1 minute. MacOS hovers at around 40 seconds.

I can't say for sure how much slower it is without cache because it didn't work with cache enabled. I can go test to see if I can get cache working (hopefully it doesn't erase the approvals).

thehomebrewnerd: FWIW, I was having trouble with caches in other OS repos recently and ended up removing them because they were failing too often.

PR #4431, opened 2024-05-22
Unpin dependencies from [project.optional-dependencies]

### LINKED ISSUE #4430: Unpin dependencies from [project.optional-dependencies]
There are a lot of dependencies under [project.optional-dependencies] that are pinned.

MichaelFu512:
### Pull Request Description
Closes #4430 

-----


pyproject.toml
```diff
@@ -77,35 +77,35 @@ dependencies = [
...
     "alteryx-open-src-update-checker >= 2.1.0"
 ]
 docs = [
-    "docutils >=0.15.2, < 0.17",
+    "docutils >= 0.15.2, < 0.17",
     "pydata-sphinx-theme >= 0.3.1",
     "astroid <= 2.6.6",
     "Sphinx >= 5.0.0",
     "nbconvert >= 6.5.0",
     "nbsphinx >= 0.8.5, < 0.9.0",
```
MichaelFu512, line 104: A combination of nbsphinx, astroid, and docutils cause the `readthedocs` pipeline to fail. I'll need to open a separate ticket to uncap those.

.github/meta.yaml
```diff
@@ -52,7 +52,7 @@ outputs:
       imports:
         - evalml
       requires:
-        - pytest ==6.2.5
+        - pytest >=6.2.5
```
VascoSch92, line 55: This can be dangerous :) at least make sure it cannot be bigger then 7.0.0, i.e, <7.0.0

MichaelFu512: Thanks for the warning! I'm curious, what exactly makes 7.0.0 dangerous for pytest?

VascoSch92: Nothing. But is a good things to do. Let's say that tomorrow they piblish a new version of pytest which break everything. If you bound the version above, then you are protect against that 😃 

I think it is always a good idea to bound versions (not to pin this is way too strict)

But just an advice ;) 

Pytest will be for sure not a problem but other packages could give this problem.

MichaelFu512: Sounds good, thanks for the explanation!

thehomebrewnerd: @MichaelFu512 If you look at the CI logs you can see we are already running with pytest `8.2.1`, so there is definitely no need to restrict this to `<7.0.0`.

Example snippet:
```
Successfully installed attrs-23.2.0 coverage-7.5.3 debugpy-1.8.1 evalml-0.83.0 execnet-2.1.1 fastjsonschema-2.19.1 iniconfig-2.0.0 ipykernel-6.29.4 jsonschema-4.22.0 jsonschema-specifications-2023.12.1 jupyter-client-8.6.2 jupyter-core-5.7.2 nbformat-5.10.4 nbval-0.11.0 nest-asyncio-1.6.0 pluggy-1.5.0 pytest-8.2.1 pytest-cov-5.0.0 pytest-timeout-2.3.1 pytest-xdist-3.6.1 referencing-0.35.1 rpds-py-0.18.1
```

evalml/tests/automl_tests/test_automl.py
```diff
@@ -1530,7 +1530,7 @@ def test_results_getter(AutoMLTestEnv, X_y_binary):
 
     assert automl.results["pipeline_results"][0]["mean_cv_score"] == 1.0
 
-    with pytest.raises(AttributeError, match="set attribute"):
+    with pytest.raises(AttributeError):
```
thehomebrewnerd, line 1533: Do you know why the changes lines in this file and in `test_pipeline_search_plots.py` are showing as uncovered by tests? That seems a bit strange to me.

thehomebrewnerd (approved):
Seems fine overall, but I'm a bit concerned/confused about the changed lines that show up as uncovered by codecov.

jeremyliweishih (approved):

Issue #4014, opened 2023-02-17
Remove nullable type handling from TimeSeriesImputer when pandas adds nullable type support for interpolate


tamargrey:
Once https://github.com/pandas-dev/pandas/issues/41565 has been implemented and released, we should upgrade pandas to that version, which will allow us to remove the nullable type handling put in place by https://github.com/alteryx/evalml/issues/4001.

PR #4432, opened 2024-05-22
Unpin dependencies blocked on Python 3.8

### LINKED ISSUE #4314: Re-evaluate pandas and Python versions
With Pandas version 2.1.0 released on August 30th, the package no longer supports Python versions below 3.9. 

We missed this for a long time because the majority of our testing (including all tests that run for PRs, as well as the latest dependency checker) only runs on Python 3.8. This was only discovered now because the pandas upgrade happened on our nightly 3.9 tests after sktime upgraded to version 0.23.0.

Our current plan is to pin pandas under version 2.1.0, but this is not a good long term solution. This issue tracks the discussion around what to do with our Python and pandas versioning in the longer term, as this will only become more of a problem as time goes on and Python 3.8 becomes more out of date.

eccabay:
Closes #4314 and #4014


eccabay:
Closing, will unpin shap and scipy in #4436 and continue to track pandas unpinning once the sktime and pandas incompatibility (issue [here](https://github.com/sktime/sktime/issues/6499)) is resolved.

